Issue #8348, opened 2023-09-02
Modular BDD Test Epic


rbair23:
Improvements to the BDD Test suite (`test-clients`) when used with the modular services to support multiple concurrent nodes, restart, reconnect, upgrade, and reboot scenarios.

The general idea is to update the `HapiTestEngine` to start 4 nodes instead of 1, and to have each node live in its own ClassLoader (thus avoiding shared static state) and with its own data & output directories. Start these nodes before all tests. Also, add to the HapiSpecOp set operations that can be used to stop, start, upgrade, delay (reconnect), or reboot these nodes. Then write tests specifically uses these capabilities to better test these scenarios in an efficient and easier to debug way.

```[tasklist]
### Tasks
- [ ] Factor `HapiTestEngine` nodes into a classloader-scoped unit with distinct data and log directories
- [ ] Create a `restart` HapiSpecOp that will freeze and start all `HapiTestEngine` nodes
- [ ] Create a `disconnect` HapiSpecOp that will disconnect a given node from communicating with the others
- [ ] Create a `reconnect` HapiSpecOp to reconnect a given node to communicate with the others
- [ ] Create a `reboot` HapiSpecOp to "reboot" a specific node
- [ ] Create a `upgrade` HapiSpecOp to freeze and upgrade all nodes
```

The last task, for `upgrade` is unknown how to handle it exactly. The software artifacts have to be uploaded but we may use "fake" software artifacts, and we may just play with the version numbers used (i.e. if we are on release 0.42.0, then maybe we start the 3 nodes in 0.41.0, put something fake into the upgrade file directory, download it on the node directory, check the directory that it contains the right bits, then restart the nodes in 0.42.0). But even this is finicky. Maybe we will need a completely different migration test plan, or for the upgrade tests we have to start a different network from the normal one in the previous release. Real migration tests also use some previous state, so we would have to work out what the "prior" state is to use.

PR #8632, opened 2023-09-14
Remove `ReadableRunningHashLeafStore` from network admin

rbair23:
Related to #8348, found while working on restart.


mhess-swl:
Is the running hash store the code we were mentioning to @lukelee-sl that he could use to get running hashes in contracts within the past few weeks? I remember we decided it would be good to access the N-3 hash through a store, but can't remember if the code deleted in this PR exactly applies

jasperpotts (approved):

PR #8659, opened 2023-09-15
File test passing ProtectedFilesUpdateSuite, PermissionSemanticsSpec, FileUpdateSuite (6 out of 16)

### LINKED ISSUE #8055: FileUpdateSuite (6 out of 16)
```[tasklist]
### Tasks
- [x] vanillaUpdateSucceeds
- [x] updateFeesCompatibleWithCreates
- [ ] apiPermissionsChangeDynamically
- [x] cannotUpdateExpirationPastMaxLifetime
- [x] optimisticSpecialFileUpdate
- [ ] associateHasExpectedSemantics
- [x] notTooManyFeeScheduleCanBeCreated
- [ ] allUnusedGasIsRefundedIfSoConfigured
- [ ] maxRefundIsEnforced
- [ ] gasLimitOverMaxGasLimitFailsPrecheck
- [ ] kvLimitsEnforced
- [ ] serviceFeeRefundedIfConsGasExhausted
- [ ] chainIdChangesDynamically
- [ ] entitiesNotCreatableAfterUsageLimitsReached
- [ ] rentItemizedAsExpectedWithOverridePriceTiers
- [x] messageSubmissionSizeChange
```

povolev15:
Fix File update handler to support correctly . In addition 31 additional tests are passing.

**Related issue(s)**:

Fixes #8055 #8020 #8056 8056


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/positive/SysDelSysUndelSpec.java
```diff
@@ -70,13 +73,15 @@ private HapiSpec distinguishesAdminPrivileges() {
                 .then(
                         systemFileDelete("misc")
                                 .payingWith(SYSTEM_UNDELETE_ADMIN)
-                                .hasPrecheck(NOT_SUPPORTED),
+                                .hasPrecheckFrom(NOT_SUPPORTED, OK)
+                                .hasKnownStatus(UNAUTHORIZED),
```
rbair23, line 77: Is UNAUTHORIZED or NOT_SUPPORTED the expected code? It looks like the existing code expected NOT_SUPPORTED from the precheck, but the new code expects UNAUTHORIZED to be determined at handle time. But NOT_SUPPORTED and UNAUTHORIZED are quite different in meeting. Should we be changing the expectation from NOT_SUPPORTED to UNAUTHORIZED?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/negative/AppendFailuresSpec.java
```diff
@@ -55,21 +54,12 @@ private HapiSpec handleRejectsOversized() {
...
                         fileAppend("file").content(BYTES_1).hasKnownStatus(MAX_FILE_SIZE_EXCEEDED))
-                .then(
-                        fileUpdate(APP_PROPERTIES)
-                                .payingWith(ADDRESS_BOOK_CONTROL)
-                                .overridingProps(Map.of("maxFileSize", "1024"))
-                                .droppingUnmentioned(),
-                        fileUpdate(APP_PROPERTIES)
-                                .payingWith(ADDRESS_BOOK_CONTROL)
-                                .path("tmp-application.properties"));
+                .then(overriding("files.maxSizeKb", "1024"));
```
rbair23, line 62: Why do these lines need to be removed? What caused them to fail? I'm worried about whether modifying the tests either diminishes the test coverage or if it breaks mono testing.

tinker-michaelj: Those `fileUpdate(APP_PROPERTIES)` operations were very early versions of the `overriding()` utility; and until we switched genesis schema to have empty `0.0.121`, they prevented this test from running at all.

Semantics of this test haven't changed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/positive/SysDelSysUndelSpec.java
povolev15: This test confirms that the system delete admin is not authorized to do system delete and the previous status is less accurate

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/negative/AppendFailuresSpec.java
povolev15: We need to use the overriding operation in case 121 file containing all properties otherwise, you cannot use the single file update.

tinker-michaelj (dismissed):

tinker-michaelj (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
```diff
@@ -225,21 +225,18 @@ public HapiSpec notTooManyFeeScheduleCanBeCreated() {
         final var denom = "fungible";
         final var token = "token";
         return defaultHapiSpec("OnlyValidCustomFeeScheduleCanBeCreated")
-                .given(fileUpdate(APP_PROPERTIES)
-                        .payingWith(GENESIS)
-                        .overridingProps(Map.of(MAX_CUSTOM_FEES_PROP, "1")))
+                .given(overriding(MAX_CUSTOM_FEES_PROP, "1"))
```
Neeharika-Sompalli, line 228: Any reason why we should change `payingWith(GENESIS)`?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
```diff
@@ -225,21 +225,18 @@ public HapiSpec notTooManyFeeScheduleCanBeCreated() {
...
                         tokenCreate(denom),
                         tokenCreate(token)
                                 .treasury(DEFAULT_PAYER)
                                 .withCustom(fixedHbarFee(1, DEFAULT_PAYER))
                                 .withCustom(fixedHtsFee(1, denom, DEFAULT_PAYER))
                                 .hasKnownStatus(CUSTOM_FEES_LIST_TOO_LONG))
-                .then(fileUpdate(APP_PROPERTIES)
-                        .payingWith(GENESIS)
-                        .overridingProps(Map.of(MAX_CUSTOM_FEES_PROP, DEFAULT_MAX_CUSTOM_FEES)));
+                .then(overriding(MAX_CUSTOM_FEES_PROP, DEFAULT_MAX_CUSTOM_FEES));
```
Neeharika-Sompalli, line 236: same here

povolev15: `overriding()` always uses `GENESIS`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
povolev15: `overriding()` always uses `GENESIS`

Neeharika-Sompalli (approved):
LGTM ! Seems changes for test-client files are intentional to make file 121 size smaller.

qnswirlds (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/positive/SysDelSysUndelSpec.java
rbair23: But does this mean mono-service returns one status and modular is returning a different status? I thought our goal was to maintain compatibility on status codes?

povolev15: Do we wants a special case to return NOT_SUPPORTED in this specific case (the SystemUndelete admin trying to use SystemDelete) because AUTHORIZATION_FAILED is thrown [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L463)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/negative/AppendFailuresSpec.java
tinker-michaelj: `overriding()` is the new version of `fileUpdate(APP_PROPERTIES).overridingProps()`---it is resilient to file `0.0.121` having a size >6kb.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/positive/SysDelSysUndelSpec.java
povolev15: we reverted the test back as it was with @tinker-michaelj and applied thee right logic in the workflow.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -457,6 +458,9 @@ private ValidationResult validate(
 
         // Check if the payer has the required permissions
         if (!authorizer.isAuthorized(payerID, functionality)) {
+            if (functionality == HederaFunctionality.SYSTEM_DELETE) {
```
netopyr, line 461: I understand it is "time to commit crimes," but seeing this one hurts. I will create a ticket so we remember to revisit this later.

povolev15: Thank you. Can you post it here? I done it because we want to match the logic as it is in Mono and Richard prefer not to change the test even that system delete and undelete never used in prod enviroments.

iwsimon (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM !

iwsimon (approved):

tinker-michaelj (approved):
LGTM

PR #8662, opened 2023-09-15
08629 platform builder

### LINKED ISSUE #8629: Integrate platform builder into Browser.java
The class `PlatformBuilder.java` duplicates a lot of logic from `Browser.java`. Integrate these two classes.

cody-littley:
Closes #8629


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -62,7 +64,9 @@ public InProcessHapiTestNode(@NonNull final Path workingDir, final long nodeId,
 
     @Override
     public void start() {
-        if (th != null) throw new IllegalStateException("Node is not stopped, cannot start it!");
+        if (th != null) {
+            throw new IllegalStateException("Node is not stopped, cannot start it!");
+        }
```
rbair23, line 69: Single line if statements are accepted as per our coding guidelines

cody-littley: reverted formatting change

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -212,237 +164,100 @@ private static void launchUnhandled(@NonNull final CommandLineArgs commandLineAr
...
-                    mainClassName,
-                    swirldName,
-                    appVersion,
-                    softwareUpgrade,
-                    initialState.get(),
-                    addressBookInitializer.getPreviousAddressBook(),
-                    emergencyRecoveryManager);
-            new InfoMember(infoSwirld, platform);
-            return platform;
+        for (Thread startThread : startThreads) {
```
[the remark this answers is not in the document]
cody-littley: fixed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -143,6 +145,17 @@ public final class Hedera implements SwirldMain {
...
+    /**
+     * The application name from the platform's perspective. This is currently locked in at the old main class name and
+     * requires data migration to change.
+     */
+    public static final String APP_NAME = "com.hedera.services.ServicesMain";
+
+    /**
+     * The swirld name. Currently there is only one swirld.
+     */
+    public static final String SWIRLD_NAME = "0"; // TODO waiting to hear back from devops on the current configuration
```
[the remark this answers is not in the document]
cody-littley: TODO resolved

alittley (dismissed):

tinker-michaelj (approved):
LGTM!

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -185,4 +186,18 @@ private <T extends Record> void addConfigDataType(@NonNull final Class<T> type)
...
+     *
+     * @param propertyName name of the property
+     * @param value        the value
+     * @return the {@link ConfigurationBuilder} instance (for fluent API)
+     */
+    @NonNull
+    public ConfigurationBuilder withValue(@NonNull final String propertyName, @NonNull final Object value) {
+        Objects.requireNonNull(propertyName, "propertyName must not be null");
+        Objects.requireNonNull(value, "value must not be null");
+        return withSource(new SimpleConfigSource(propertyName, value.toString()));
```
hendrikebbers, line 201: That implementation is not good from my point of view. If you provide multiple values it would end in multiple sources that all have the same ordinal. That will be hard to handle. Please store a map in the impl class and create a single SimpleConfigSource with very low ordinal once build() is called.

cody-littley: I didn't realize that was possible, I was following the pattern in `TestConfigBuilder.java`, which creates a new source each time. I agree that the way you suggest is a lot better. Change made.

edward-swirldslabs (approved):

tinker-michaelj (dismissed):
LGTM, thanks @cody-littley !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -174,6 +174,17 @@ public final class Hedera implements SwirldMain {
...
+    /**
+     * The application name from the platform's perspective. This is currently locked in at the old main class name and
+     * requires data migration to change.
+     */
+    public static final String APP_NAME = "com.hedera.services.ServicesMain";
+
+    /**
+     * The swirld name. Currently there is only one swirld.
+     */
+    public static final String SWIRLD_NAME = "123";
```
hendrikebbers, line 186: I do not understand what the swirldName is. Documentation in PlatformBuilder does not help. Independent of that I assume that "123" is not a good value for our real application

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -174,6 +174,17 @@ public final class Hedera implements SwirldMain {
     private HandleThrottleAccumulator handleThrottling;
     private MonoMultiplierSources monoMultiplierSources;
 
+    /**
+     * The application name from the platform's perspective. This is currently locked in at the old main class name and
+     * requires data migration to change.
+     */
+    public static final String APP_NAME = "com.hedera.services.ServicesMain";
```
hendrikebbers, line 181: Do we have an issue to change the name?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -90,10 +91,16 @@ public void run() {
...
+        final Hedera hedera = new Hedera(registry);
+        final NodeId selfId = args != null && args.length > 0 ? new NodeId(Integer.parseInt(args[0])) : new NodeId(0);
+
+        final PlatformBuilder builder = new PlatformBuilder(
+                Hedera.APP_NAME, Hedera.SWIRLD_NAME, hedera.getSoftwareVersion(), hedera::newState, selfId);
+
+        final SwirldsPlatform platform = builder.build();
+        hedera.init(platform, selfId);
+        platform.start();
+        hedera.run();
```
hendrikebbers, line 104: I still hate this code block. It is way too complicated. Either hedera.run() or platform.start() should call the other 2 methods. No need to do this in the PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -90,10 +91,16 @@ public void run() {
...
-                .withNodeId(args != null && args.length > 0 ? Integer.parseInt(args[0]) : 0)
-                .buildAndStart();
+
+        final Hedera hedera = new Hedera(registry);
+        final NodeId selfId = args != null && args.length > 0 ? new NodeId(Integer.parseInt(args[0])) : new NodeId(0);
+
+        final PlatformBuilder builder = new PlatformBuilder(
+                Hedera.APP_NAME, Hedera.SWIRLD_NAME, hedera.getSoftwareVersion(), hedera::newState, selfId);
+
+        final SwirldsPlatform platform = builder.build();
```
hendrikebbers, line 101: Why does it needs to be a SwirldsPlatform? I would assume Platform should be used here and services have no idea about the implementation

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -88,6 +96,9 @@ public Configuration build() {
             if (initialized.get()) {
                 throw new IllegalStateException("Configuration already initialized");
             }
+            if (!properties.isEmpty()) {
+                withSource(new SimpleConfigSource(properties));
```
hendrikebbers, line 100: We should define a specific ordinal for that source

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -185,4 +196,19 @@ private <T extends Record> void addConfigDataType(@NonNull final Class<T> type)
...
+     *
+     * @param propertyName name of the property
+     * @param value        the value
+     * @return the {@link ConfigurationBuilder} instance (for fluent API)
+     */
+    @NonNull
+    public ConfigurationBuilder withValue(@NonNull final String propertyName, @NonNull final Object value) {
+        Objects.requireNonNull(propertyName, "propertyName must not be null");
+        Objects.requireNonNull(value, "value must not be null");
+        properties.put(propertyName, value.toString());
```
hendrikebbers, line 211: That is not working. The configuration has Converters to define how a string value is transformed to an object of a specific type. The conversion from object to string is currently not defined but it won't be just `toString()`. That won't work for any `java.time` classes for example. For now the best would be to have `withValue(@NonNull final String propertyName, @NonNull final String value)`. Everything else will make that PR really way more complexe.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/LegacyFileConfigSource.java
```diff
@@ -47,16 +47,14 @@ public class LegacyFileConfigSource extends AbstractConfigSource {
...
      * @throws IOException if settings.txt can not be loaded
      */
-    public static LegacyFileConfigSource ofSettingsFile() throws IOException {
-        final Path path = FileSystems.getDefault()
-                .getPath("settings.txt")
-                .toAbsolutePath()
-                .normalize();
-        return new LegacyFileConfigSource(path, ConfigSourceOrdinalConstants.LEGACY_PROPERTY_FILE_ORDINAL_FOR_SETTINGS);
+    @NonNull
+    public static LegacyFileConfigSource ofSettingsFile(@NonNull final Path settingsPath) throws IOException {
```
hendrikebbers, line 55: why not have both methods?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/SimpleConfigSource.java
```diff
@@ -41,6 +43,13 @@ public SimpleConfigSource() {
         this.internalProperties = new HashMap<>();
     }
 
+    /**
+     * Creates an instance without any config properties
+     */
+    public SimpleConfigSource(@NonNull final Map<String, String> properties) {
+        this.internalProperties = Objects.requireNonNull(properties);
```
hendrikebbers, line 50: Please create a copy of the map

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/SwirldMain.java
```diff
@@ -16,67 +16,57 @@
...
-    }
-
-    /**
-     * Update the configuration builder with any additional configuration that is required by this swirld.
-     *
-     * @param configurationBuilder
-     * 		the configuration builder to update
-     */
-    default void updateConfigurationBuilder(@NonNull final ConfigurationBuilder configurationBuilder) {
+    @Nullable
```
hendrikebbers, line 34: I do not like to have a method that returns List<> to allow a nullable return. That should never be the case. Just return List.of() as default.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /**
+     * Setup static utilities. If running multiple platforms in the same JVM, this should be called when constructing
+     * the first platform only.
+     *
+     * @param configuration the configuration for this node
+     * @return true if this is the first time this method has been called, false otherwise
+     */
+    private boolean setupStaticUtilities(@NonNull final Configuration configuration) {
```
hendrikebbers, line 207: I ask my self if we can externalize that to have the builder not use static. Maybe a `StaticPlatformBuilder` as a singleton that is just called once by this builder class.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -0,0 +1,375 @@
...
+        return legacyConfig.getAddressBook();
+    }
+
+    /**
+     * Build a platform.
+     *
+     * @param start if true then start the platform
+     * @return a new platform instance
+     */
+    public SwirldsPlatform build(final boolean start) {
```
hendrikebbers, line 290: Why? Can't we just define the builder to return a platform that is never started? That would make it way more easy from my point of view.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -0,0 +1,375 @@
...
+        ConfigurationHolder.getInstance().setConfiguration(configuration);
+
+        // Setup logging, using the configuration to tell us the location of the log4j2.xml
+        final PathsConfig pathsConfig = configuration.getConfigData(PathsConfig.class);
+        final Path log4jPath = pathsConfig.getLogPath();
+        try {
+            Log4jSetup.startLoggingFramework(log4jPath).await();
+        } catch (final InterruptedException e) {
+            CommonUtils.tellUserConsole("Interrupted while waiting for log4j to initialize");
+            Thread.currentThread().interrupt();
```
hendrikebbers, line 223: We should throw an exception in that point from my point of view

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -0,0 +1,375 @@
...
+     *
+     * @return the configuration
+     */
+    @NonNull
+    private Configuration buildConfiguration() {
+        if (configurationBuilder == null) {
+            configurationBuilder = ConfigurationBuilder.create();
+        }
+
+        if (!Files.exists(configPath)) {
```
hendrikebbers, line 257: Can we move that checks to BootstrapUtils.setupConfigBuilder or even better to the config source that really uses the path?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
```diff
@@ -56,8 +56,7 @@ public final class LegacyConfigPropertiesLoader {
     private static final String SWIRLD_PROPERTY_NAME = "swirld";
     private static final String GENESIS_FREEZE_TIME_PROPERTY_NAME = "genesisfreezetime";
 
-    public static final String ERROR_CONFIG_TXT_NOT_FOUND =
-            "ERROR: Browser.startPlatforms called on non-existent config.txt";
+    public static final String ERROR_CONFIG_TXT_NOT_FOUND = "ERROR: Browser.startPlatforms called on non-existent %s";
```
hendrikebbers, line 59:
```suggestion
    public static final String ERROR_CONFIG_TXT_NOT_FOUND = "ERROR: Configuration file not found: %s";
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
```diff
@@ -72,13 +71,13 @@ public final class LegacyConfigPropertiesLoader {
...
-    public static LegacyConfigProperties loadConfigFile(Path configPath) throws ConfigurationException {
+    public static LegacyConfigProperties loadConfigFile(@NonNull final Path configPath) throws ConfigurationException {
         CommonUtils.throwArgNull(configPath, "configPath");
 
         // Load config.txt file, parse application jar file name, main class name, address book, and parameters
         if (!Files.exists(configPath)) {
-            logger.error(EXCEPTION.getMarker(), ERROR_CONFIG_TXT_NOT_FOUND);
-            throw new ConfigurationException(ERROR_CONFIG_TXT_NOT_FOUND);
+            logger.error(EXCEPTION.getMarker(), ERROR_CONFIG_TXT_NOT_FOUND.formatted(configPath.toString()));
+            throw new ConfigurationException(ERROR_CONFIG_TXT_NOT_FOUND.formatted(configPath.toString()));
```
hendrikebbers, line 80: Do we really need to log and throw?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/editor/StateEditor.java
```diff
@@ -62,7 +63,8 @@ public class StateEditor {
     @SuppressWarnings("java:S106")
     public StateEditor(final Path statePath) throws IOException {
 
-        final Configuration configuration = DefaultConfiguration.buildBasicConfiguration();
+        final Configuration configuration =
+                DefaultConfiguration.buildBasicConfiguration(getAbsolutePath("settings.txt"));
```
hendrikebbers, line 67: We use `DefaultConfiguration.buildBasicConfiguration(getAbsolutePath("settings.txt"))` all over the place. That should be a direct method in `DefaultConfiguration`.

platform-sdk/swirlds-unit-tests/common/swirlds-test-framework/src/main/java/com/swirlds/test/framework/config/TestConfigBuilder.java
```diff
@@ -154,14 +93,16 @@ public TestConfigBuilder withValue(@NonNull final String propertyName, final boo
      */
     @NonNull
     public TestConfigBuilder withValue(@NonNull final String propertyName, @NonNull final Object value) {
```
hendrikebbers, line 95: I missed that we have the method with Object here, too. Again I would totally avoid it by now (even if I might have created this one).

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
```diff
@@ -54,22 +54,23 @@ private DefaultConfiguration() {
      * @throws IOException if there is an error reading the configuration files
      */
     @NonNull
-    public static Configuration buildBasicConfiguration() throws IOException {
-        return buildBasicConfiguration(Collections.emptyList());
+    public static Configuration buildBasicConfiguration(@NonNull final Path settingsPath) throws IOException {
```
alittley, line 57: missing param doc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -184,22 +139,18 @@ public static void launch(@NonNull final CommandLineArgs commandLineArgs, final
...
     private static void launchUnhandled(@NonNull final CommandLineArgs commandLineArgs, final boolean pcesRecovery)
             throws Exception {
         Objects.requireNonNull(commandLineArgs);
 
-        StartupTime.markStartupTime();
-        logger.info(STARTUP.getMarker(), "\n\n" + STARTUP_MESSAGE + "\n");
-        logger.debug(STARTUP.getMarker(), () -> new NodeStartPayload().toString());
-
         // This contains the default PathsConfigs values, since the overrides haven't been loaded in yet
         final PathsConfig pathsConfig = loadPathsConfig();
```
alittley, line 149: `pathsConfig` is no longer used

alittley (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
cody-littley: It's currently an unused feature, more or less. But it's something that is currently used in file system paths, so we have to carry it forward until we do a data migration. Right now all hedera deployments use the swirld name "123". This is data that currently gets passed in via the `config.txt` file. Storing it as a constant that gets passed into the constructor will eliminate that dependency.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
cody-littley: I'm not aware of an issue. It will require data migration since this is currently used for file paths on disk. It will probably also need tooling updates, since I'm assuming there is probably a lot of tooling that expects the files to be in a specific location (e.g. NMT).

This is definitely something that would be really nice to fix. But the effort required vs. the immediate benefit means that we always end up procrastinating. 🙃

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
cody-littley: The only reason it is `SwirldsPlatform` is because there is no `start()` method on the interface `Platform`. Of the following options, which of the following do you prefer?

- keep it the way it is
- `Platform extends Startable`
- `((Startable) platform).start()`
- Create a new interface `StartablePlatform extends Startable, Platform`

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
cody-littley: This now uses the following ordinal:

```
    /**
     * Ordinal for configuration passed to the platform builder via withConfig().
     */
    public static final int PLATFORM_BUILDER_ORDINAL = 50;
```

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
cody-littley: Removed this method

cody-littley: Changed to `withValue(@NonNull final String propertyName, @NonNull final String value)`

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/SimpleConfigSource.java
cody-littley: done

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/LegacyFileConfigSource.java
cody-littley: Mostly just because we don't have any current use cases for the zero arg constructor, and since this is a legacy type I thought we'd try to avoid new use cases in the future. I can add it back in if you feel strongly about it.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/SwirldMain.java
cody-littley: change made

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
cody-littley: I've moved the static stuff to a helper class.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
cody-littley: Mostly just for convenience, but since you don't like it I will remove it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
cody-littley: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
cody-littley: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
cody-littley: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
cody-littley: Not really, now it will just throw

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/editor/StateEditor.java
cody-littley: method with default location added

platform-sdk/swirlds-unit-tests/common/swirlds-test-framework/src/main/java/com/swirlds/test/framework/config/TestConfigBuilder.java
cody-littley: I reverted changes to this file in this PR. Perhaps we can update it in a future PR.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
cody-littley: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
cody-littley: removed

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/ConfigSourceOrdinalConstants.java
```diff
@@ -47,6 +47,12 @@ public final class ConfigSourceOrdinalConstants {
 
     @Deprecated(forRemoval = true)
     public static final int LEGACY_PROPERTY_FILE_ORDINAL_FOR_CONFIG = LEGACY_PROPERTY_FILE_ORDINAL + 20;
+
+    /**
+     * Ordinal for configuration passed to the platform builder via withConfig().
+     */
+    public static final int PLATFORM_BUILDER_ORDINAL = 50;
```
hendrikebbers, line 54: Please do not add platform dependent content to common. Can you move that const to platform-core?

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
```diff
@@ -119,6 +119,16 @@ public interface ConfigurationBuilder {
     @NonNull
     ConfigurationBuilder withConfigDataTypes(@NonNull Class<? extends Record>... types) throws IllegalStateException;
 
+    /**
+     * Sets the value for the config.
```
hendrikebbers, line 123: Please change to:
`Adds a property. In the created {@link Configuration} the property will be defined by an internal config source that has the defined ordinal of FOO.`
or similar.

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
```diff
@@ -119,6 +119,16 @@ public interface ConfigurationBuilder {
...
+     *
+     * @param propertyName name of the property
+     * @param value        the value
```
hendrikebbers, line 126:
```suggestion
     * @param value        the value of the property
```

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -88,6 +98,9 @@ public Configuration build() {
             if (initialized.get()) {
                 throw new IllegalStateException("Configuration already initialized");
             }
+            if (!properties.isEmpty()) {
+                withSource(new SimpleConfigSource(properties).withOrdinal(PLATFORM_BUILDER_ORDINAL));
```
hendrikebbers, line 102: Ahh, no I see where `PLATFORM_BUILDER_ORDINAL` is used. Please rename it and place it here :)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -184,22 +138,15 @@ public static void launch(@NonNull final CommandLineArgs commandLineArgs, final
...
-        logger.info(STARTUP.getMarker(), "\n\n" + STARTUP_MESSAGE + "\n");
-        logger.debug(STARTUP.getMarker(), () -> new NodeStartPayload().toString());
-
-        // This contains the default PathsConfigs values, since the overrides haven't been loaded in yet
-        final PathsConfig pathsConfig = loadPathsConfig();
-
         // Load config.txt file, parse application jar file name, main class name, address book, and parameters
         final ApplicationDefinition appDefinition =
-                ApplicationDefinitionLoader.loadDefault(pathsConfig.getConfigPath());
+                ApplicationDefinitionLoader.loadDefault(getAbsolutePath("./config.txt"));
```
hendrikebbers, line 149: isn't `"./config.txt"` the same as `"config.txt"` ? If yes I would prefer the second one, but maybe that is just me :)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -209,240 +156,108 @@ private static void launchUnhandled(@NonNull final CommandLineArgs commandLineAr
...
-                    nodeId,
-                    configAddressBook,
-                    emergencyRecoveryManager);
-        } catch (final SignedStateLoadingException e) {
-            exitSystem(SystemExitCode.FATAL_ERROR, "unable to load initial state");
-            throw new IllegalStateException(); // unreachable
+    private static void startPlatforms(
+            @NonNull final List<SwirldsPlatform> platforms, @NonNull final Map<NodeId, SwirldMain> appMains) {
+
+        if (platforms.size() == 1) {
```
hendrikebbers, line 220: Question: Why do we not start 1 platform in an external thread like we would do if there are 2? The current code makes a hard difference between having 1 or N platforms. Maybe not for this PR but from my point of view we should not have 2 different behaviors.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+    private final String swirldName;
+
+    private ConfigurationBuilder configurationBuilder;
+
+    private static final String SWIRLDS_PACKAGE = "com.swirlds";
+
+    /**
+     * The path to config.txt.
+     */
+    private Path configPath = getAbsolutePath("config.txt");
```
hendrikebbers, line 86: Here you do it without `"./"` so we should have the path written in same style everywhere. I prefer this one.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/StaticPlatformBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.nio.file.Path;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Static setup code for a platform. Logic in this class should gradually disappear as we move away from the use of
+ * static stateful objects.
+ */
+final class StaticPlatformBuilder {
```
hendrikebbers, line 42: From my point of view the class should directly be deprecated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
```diff
@@ -47,29 +49,43 @@ private DefaultConfiguration() {
     }
 
     /**
-     * Build a basic configuration with the default configuration sources and paths.
-     * And register the configuration to the {@link ConfigurationHolder}.
+     * Build a basic configuration with the default configuration sources and paths. Reads configuration form
+     * "./settings.txt". Registers the configuration to the {@link ConfigurationHolder}.
```
hendrikebbers, line 53:
```suggestion
     * "settings.txt". Registers the configuration to the {@link ConfigurationHolder}.
```

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
cody-littley: done

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
cody-littley: done

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/sources/ConfigSourceOrdinalConstants.java
cody-littley: removed

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
cody-littley: Since we want to reference this value in the javadocs on the public interface `ConfigurationBuilder`, I moved this constant to `ConfigurationBuilder.java` and renamed it to `CUSTOM_PROPERTY_ORDINAL`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
cody-littley: They are equivalent. I'm ok leaving out the `./`. Change made.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
cody-littley: I've switched this over to always using a thread. Mildly inefficient, but I'm ok making that sacrifice for easier to understand code.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/StaticPlatformBuilder.java
cody-littley: There is a very small amount of logic that I do not think we will ever remove.

- starting the thread dump generator
- starting the JVM pause detector
- writing the "hello world" log message (this may change with logging contexts)
- system health checks

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
cody-littley: done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
hendrikebbers: Does it makes sense to create an issue for that to have a more meaningful name in future?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
hendrikebbers: `Platform extends Startable` <- Maybe that is the best idea for now, what do you think?

alittley (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
cody-littley: Created issue: https://github.com/hashgraph/hedera-services/issues/9029

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
cody-littley: I'm ok with this solution. Done.

edward-swirldslabs (dismissed):

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/CleanCommand.java
```diff
@@ -66,7 +66,7 @@ public static void clean(@NonNull final Path sdkPath) throws IOException {
         Objects.requireNonNull(sdkPath);
 
         final Configuration configuration = DefaultConfiguration.buildBasicConfiguration(
-                List.of(sdkPath.resolve("config.txt"), sdkPath.resolve("settings.txt")));
+                sdkPath.resolve("settings.txt"), List.of(sdkPath.resolve("config.txt")));
```
lpetrovic05, line 69: Can we put these file names in a constant somewhere and use that constant? I see these strings all over the place

lpetrovic05 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/CleanCommand.java
cody-littley: Created constants.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
```diff
@@ -128,26 +127,27 @@ public PlatformBuilder withConfigurationBuilder(@Nullable final ConfigurationBui
...
         return this;
     }
 
     /**
-     * The path to the config file. Default is "./config.txt" where "./" is the current working directory of the JVM.
+     * The path to the config file (i.e. the file with the address book. Traditionally named
+     * {@link #DEFAULT_CONFIG_FILE_NAME}.
      *
-     * @param path the path to config.txt
+     * @param path the path the config file
```
alittley, line 147: the path *to* the config file

alittley (approved):

hendrikebbers (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
cody-littley: fixed

tinker-michaelj (approved):
LGTM!

edward-swirldslabs (approved):

Issue #8077, opened 2023-08-17
Start testing `TokenCreateSpecs`


Neeharika-Sompalli:
- [ ] validateNewTokenAssociations(will be fixed in https://github.com/hashgraph/hedera-services/issues/8757)
- [x] createsFungibleInfiniteByDefault
- [x] worksAsExpectedWithDefaultTokenId
- [x] cannotCreateWithExcessiveLifetime
- [ ] autoRenewValidationWorks(will be fixed in https://github.com/hashgraph/hedera-services/issues/8758)
- [x] creationYieldsExpectedToken
- [x] creationSetsExpectedName
- [x] creationWithoutKYCSetsCorrectStatus
- [ ] baseCreationsHaveExpectedPrices(will be fixed in https://github.com/hashgraph/hedera-services/issues/8808)
- [x] creationHappyPath
- [x] creationSetsCorrectExpiry
- [x] creationValidatesExpiry
- [x] creationValidatesFreezeDefaultWithNoFreezeKey
- [x] creationValidatesMemo
- [x] creationValidatesNonFungiblePrechecks
- [x] creationValidatesMaxSupply
- [ ] onlyValidCustomFeeScheduleCanBeCreated(will be fixed in https://github.com/hashgraph/hedera-services/issues/8811)
- [x] feeCollectorSigningReqsWorkForTokenCreate
- [x] creationValidatesName
- [x] creationValidatesSymbol
- [x] creationRequiresAppropriateSigs
- [x] creationRequiresAppropriateSigsHappyPath
- [x] creationValidatesTreasuryAccount
- [x] initialSupplyMustBeSane
- [x] treasuryHasCorrectBalance
- [x] prechecksWork

PR #8672, opened 2023-09-18
Fix TokenCreateSpecs

vtronkov:
**Description**:
This PR fixes the failing tests in TokenCreateSpecs. 3 of them will be fixed in separate issues.

**Related issue(s)**:

Fixes [https://github.com/hashgraph/hedera-services/issues/8077](https://github.com/hashgraph/hedera-services/issues/8077)

**Notes for reviewer**:
* Added missing validations
* Changing the expected statuses
* Fixed the INVALID_EXPIRATION_TIME logic

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -188,6 +188,26 @@ private void associateAccounts(
         for (final var customFee : requireCollectorAutoAssociation) {
             // This should exist as it is validated in validateSemantics
             final var collector = accountStore.get(customFee.feeCollectorAccountIdOrThrow());
+            // TODO: probably this is not the fix but I leave it here to start a discussion.
```
vtronkov, line 191: I'm open for discussions on that one

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -323,14 +334,16 @@ private void addCustomFeeCollectorKeys(
...
-                final var royaltyFee = customFee.royaltyFeeOrThrow();
-                var alwaysAdd = false;
-                if (royaltyFee.hasFallbackFee()) {
-                    final var fFee = royaltyFee.fallbackFeeOrThrow();
-                    alwaysAdd = fFee.hasDenominatingTokenId()
-                            && fFee.denominatingTokenIdOrThrow().tokenNum() == 0;
-                }
-                addAccount(context, collector, alwaysAdd);
+                // TODO: the suite onlyValidCustomFeeScheduleCanBeCreated is failing because there is no custom fee set.
+                // do we need to require it?
```
vtronkov, line 338: Thoughts?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImpl.java
```diff
@@ -91,7 +91,7 @@ public void validateExpiry(long expiry) {
                 context.configuration().getConfigData(EntitiesConfig.class).maxLifetime();
         final var now = context.consensusNow().getEpochSecond();
         final var expiryGivenMaxLifetime = now + maxEntityLifetime;
-        validateTrue(expiry > now && expiry <= expiryGivenMaxLifetime, INVALID_EXPIRATION_TIME);
+        validateTrue(expiry > now && expiry < expiryGivenMaxLifetime, INVALID_EXPIRATION_TIME);
```
vtronkov, line 94: I think Iris had a PR that changed this one but it was rejected. Does anyone knows if this change is correct? As far as I can see the old code works the same way:
`return then.isAfter(consensusNow) && then.isBefore(expiryGivenMaxLifetime);`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -771,7 +788,7 @@ public HapiSpec onlyValidCustomFeeScheduleCanBeCreated() {
                 .then(getTokenInfo(token)
                         .hasCustom(fixedHbarFeeInSchedule(hbarAmount, hbarCollector))
                         .hasCustom(fixedHtsFeeInSchedule(htsAmount, feeDenom, htsCollector))
-                        .hasCustom(fixedHtsFeeInSchedule(htsAmount, token, htsCollector))
+                        .hasCustom(fixedHtsFeeInSchedule(htsAmount, SENTINEL_VALUE, htsCollector))
```
vtronkov, line 791: Seems that the above tokeCreate was updated but the getTokenInfo was not.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -279,17 +282,18 @@ public HapiSpec autoRenewValidationWorks() {
                         cryptoCreate(deletingAccount).balance(0L))
                 .when(
                         cryptoDelete(deletingAccount),
-                        tokenCreate(PRIMARY)
-                                .autoRenewAccount(deletingAccount)
-                                .hasKnownStatus(INVALID_AUTORENEW_ACCOUNT),
+                        tokenCreate(PRIMARY).autoRenewAccount(deletingAccount).hasKnownStatus(ACCOUNT_DELETED),
```
vtronkov, line 285: We have a `TokenHandlerHelper.getIfUsable` method - this is a general method that checks if the account was deleted - if so we return the general error code _ACCOUNT_DELETED_(without checking for additional details and return specific response codes). This fails in `creationValidatesTreasuryAccount`, `validateNewTokenAssociations`, and `autoRenewValidationWorks` where we expect a concrete deleted response(_INVALID_TREASURY_ACCOUNT_FOR_TOKEN_ and _INVALID_AUTORENEW_ACCOUNT_). For example, _INVALID_TREASURY_ACCOUNT_FOR_TOKEN_, when the account is deleted **AND** the account is treasury -  we return _INVALID_TREASURY_ACCOUNT_FOR_TOKEN_. Do we need to add the specific checks or is it okay to return the general _ACCOUNT_DELETED_ in these cases?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -242,18 +262,9 @@ private Token buildToken(
...
-                ? consensusTime + op.autoRenewPeriod().seconds()
-                : consensusTime + maxEntityLifetime;
-
+    private ExpiryMeta getExpiryMeta(@NonNull final TokenCreateTransactionBody op) {
         return new ExpiryMeta(
-                impliedExpiry,
+                op.expiry().seconds(),
                 op.hasAutoRenewPeriod() ? op.autoRenewPeriod().seconds() : NA,
                 // Shard and realm will be ignored if num is NA
                 op.autoRenewAccount());
```
vtronkov, line 270: I see that in the old code, we are just passing the expiry and then we do:
`return then.isAfter(consensusNow) && then.isBefore(expiryGivenMaxLifetime);`
Why do we need this additional calculation?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -200,6 +200,8 @@ private HapiSpec validateNewTokenAssociations() {
                                         selfDenominatedFixedCollector,
                                         otherSelfDenominatedFixedCollector)
                                 .via(creationTxn),
+                        // TODO: Here it throws that we expect INVALID_TREASURY_ACCOUNT_FOR_TOKEN but it was OK.
+                        // When I change it to expect OK it now returns PLATFORM_NOT_ACTIVE
```
vtronkov, line 204: Created a story for this: https://github.com/hashgraph/hedera-services/issues/8757

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImpl.java
netopyr: @tinker-michaelj wrote the original code in `StandardizedAttributeValidator`. We should check with him.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
netopyr: I assume this change will break the tests for the current code base. If we agree to change the expectation, we must preserve the old expectation.

A more specific error message is preferred IMO. A new token can be created with several keys, and knowing which one is invalid is helpful.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
netopyr: I am not sure I understand. Why was this change necessary?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -919,13 +951,12 @@ public HapiSpec creationRequiresAppropriateSigsHappyPath() {
...
 
+    @HapiTest
     public HapiSpec creationValidatesTreasuryAccount() {
         return defaultHapiSpec("CreationValidatesTreasuryAccount")
                 .given(cryptoCreate(TOKEN_TREASURY).balance(0L))
                 .when(cryptoDelete(TOKEN_TREASURY))
-                .then(tokenCreate("shouldntWork")
-                        .treasury(TOKEN_TREASURY)
-                        .hasKnownStatus(INVALID_TREASURY_ACCOUNT_FOR_TOKEN));
+                .then(tokenCreate("shouldntWork").treasury(TOKEN_TREASURY).hasKnownStatus(ACCOUNT_DELETED));
```
netopyr, line 959: I think we should try to preserve the old behavior.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
```diff
@@ -84,12 +89,14 @@ public Set<CustomFee> validateForCreation(
...
+            final var isSpecified = fee.hasFixedFee() || fee.hasFractionalFee() || fee.hasRoyaltyFee();
+            validateTrue(isSpecified, CUSTOM_FEE_NOT_FULLY_SPECIFIED);
+
             switch (fee.fee().kind()) {
                 case FIXED_FEE -> validateFixedFeeForCreation(
                         tokenType, fee, createdToken, tokenRelationStore, tokenStore, fees);
-                case FRACTIONAL_FEE -> validateTrue(
-                        isFungibleCommon(tokenType), CUSTOM_FRACTIONAL_FEE_ONLY_ALLOWED_FOR_FUNGIBLE_COMMON);
-                case ROYALTY_FEE -> validateRoyaltyFeeForCreation(tokenType, fee, tokenRelationStore, tokenStore);
+                case FRACTIONAL_FEE -> validateFractionalFeeForCreation(tokenType, fee, fees);
```
netopyr, line 98: I find it surprising that methods called `validateXYZ()` change the value of `fees`. Maybe there is an easy solution to separate the check and the modification of `fees`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImpl.java
tinker-michaelj: I agree with @vtronkov, the [corresponding check](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/validation/ContextOptionValidator.java#L145) in mono-service is a strict inequality, `<` not `<=`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -188,6 +188,26 @@ private void associateAccounts(
         for (final var customFee : requireCollectorAutoAssociation) {
             // This should exist as it is validated in validateSemantics
             final var collector = accountStore.get(customFee.feeCollectorAccountIdOrThrow());
+            // TODO: probably this is not the fix but I leave it here to start a discussion.
+            // onlyValidCustomFeeScheduleCanBeCreated was failing because of this:
```
Neeharika-Sompalli, line 192: I think what you did is correct. But we should use similar logic like mono-service using set of accountIds in a [set](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/token/process/NewRels.java#L36) like this ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
Neeharika-Sompalli: Agreed can change name of the function. Since we are adding all custom fees based on validation, I am not sure how it can be separated.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
Neeharika-Sompalli: I think what you did is correct. But we should use similar logic like mono-service using set of accountIds in a set like this ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
```diff
@@ -228,20 +237,48 @@ private void validateFixedFeeForCreation(
         }
     }
 
+    private void validateFractionalFeeForCreation(TokenType tokenType, CustomFee fee, Set<CustomFee> fees) {
```
Neeharika-Sompalli, line 240: Wow nice catch ! 💯 Did not realize these from constructor of FractionalFeeSpec

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
```diff
@@ -228,20 +237,48 @@ private void validateFixedFeeForCreation(
...
+
         if (royaltyFee.hasFallbackFee()) {
+            validateTrue(royaltyFee.fallbackFee().amount() > 0, CUSTOM_FEE_MUST_BE_POSITIVE);
             final var fallbackFee = royaltyFee.fallbackFeeOrThrow();
             if (fallbackFee.hasDenominatingTokenId()) {
                 final var denominatingTokenId = fallbackFee.denominatingTokenIdOrThrow();
                 validateTrue(denominatingTokenId.tokenNum() != 0, CUSTOM_FEE_DENOMINATION_MUST_BE_FUNGIBLE_COMMON);
                 validateExplicitTokenDenomination(
                         fee.feeCollectorAccountId(), denominatingTokenId, tokenRelationStore, tokenStore);
+                fees.add(fee);
```
Neeharika-Sompalli, line 281: Why are we adding this to fees ? The list of fees is only needed for collector auto association when token mentioned is a sentinel value.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
```diff
@@ -228,20 +237,48 @@ private void validateFixedFeeForCreation(
...
+        final var minimumUnitsToCollect = fractionalFee.minimumAmount();
+        final var nominalMax = fractionalFee.maximumAmount();
+        final var maximumUnitsToCollect = nominalMax == 0 ? Long.MAX_VALUE : nominalMax;
+
+        validateTrue(denominator != 0, FRACTION_DIVIDES_BY_ZERO);
+        validateTrue(numerator > 0 && denominator > 0, CUSTOM_FEE_MUST_BE_POSITIVE);
+        validateTrue(minimumUnitsToCollect >= 0 && maximumUnitsToCollect >= 0, CUSTOM_FEE_MUST_BE_POSITIVE);
+        validateTrue(maximumUnitsToCollect >= minimumUnitsToCollect, FRACTIONAL_FEE_MAX_AMOUNT_LESS_THAN_MIN_AMOUNT);
+
+        fees.add(fee);
```
Neeharika-Sompalli, line 255: same here, we should not be adding this to fees, since that list is only needed for collector auto association when token mentioned is a sentinel value.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
Neeharika-Sompalli: `PLATFORM_NOT_ACTIVE` could be because of some error or because state locally is not cleaned. This might not be the root cause for this test failure.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
Neeharika-Sompalli: We should preserve mono-service behavior. In this case if we need specific error codes, we can add a new method with different signature of `getIfUsable` and provide the error response to be returned.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
```diff
@@ -213,6 +220,8 @@ private void validateFixedFeeForCreation(
             @NonNull final Set<CustomFee> feesWithCollectorsToAutoAssociate) {
         final var fixedFee = fee.fixedFeeOrThrow();
         if (fixedFee.hasDenominatingTokenId()) {
+            validateTrue(fixedFee.amount() > 0, CUSTOM_FEE_MUST_BE_POSITIVE);
```
Neeharika-Sompalli, line 223: This check should be outside `if (fixedFee.hasDenominatingTokenId())` on line 222 as I see in mono-service

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
vtronkov: We can but we have a different structure than the mono. As far as I can understand to have the same effect we need to:
1. Create a new Set<AccountID> associatedSoFar variable in `associateAccounts`
2. Pass it to `createAndLinkTokenRels` to add the linked _accountId_
3. Pass it to `validateAssociation` to verify that the current _accountId_ has not already been linked.

It seems a bit cumbersome. But if you think that it adds additional value let me know and I will implement it 👍

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
Neeharika-Sompalli: +1

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -868,7 +894,13 @@ public HapiSpec creationValidatesSymbol() {
...
                 .when()
                 .then(
-                        tokenCreate("missingSymbol").symbol("").hasPrecheck(MISSING_TOKEN_SYMBOL),
+                        tokenCreate("missingSymbol")
+                                .symbol("")
+                                // Either fail pre-check with MISSING_TOKEN_SYMBOL (mono)
+                                // or post-consensus (modular). When we only have modular code left,
+                                // we can just check hasKnownStatus.
+                                .hasPrecheckFrom(OK, MISSING_TOKEN_SYMBOL)
+                                .hasKnownStatus(MISSING_TOKEN_SYMBOL),
```
Neeharika-Sompalli, line 903: Instead of this you can add `MISSING_TOKEN_SYMBOL` to this property `spec.streamlinedIngestChecks`. This will avoid modifying tests

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
Neeharika-Sompalli: +1

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -849,7 +867,14 @@ public HapiSpec creationValidatesName() {
...
                 .then(
-                        tokenCreate(PRIMARY).name("").logged().hasPrecheck(MISSING_TOKEN_NAME),
+                        tokenCreate(PRIMARY)
+                                .name("")
+                                .logged()
+                                // Either fail pre-check with MISSING_TOKEN_NAME (mono)
+                                // or post-consensus (modular). When we only have modular code left,
+                                // we can just check hasKnownStatus.
+                                .hasPrecheckFrom(OK, MISSING_TOKEN_NAME)
+                                .hasKnownStatus(MISSING_TOKEN_NAME),
```
Neeharika-Sompalli, line 877: Same here. Instead of this you can add MISSING_TOKEN_NAME to this property `spec.streamlinedIngestChecks`. This will avoid modifying tests

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -539,11 +547,18 @@ public HapiSpec creationSetsCorrectExpiry() {
...
                 .given()
                 .when()
-                .then(tokenCreate(PRIMARY).expiry(1000).hasPrecheck(INVALID_EXPIRATION_TIME));
+                .then(tokenCreate(PRIMARY)
+                        .expiry(1000)
+                        // Either fail pre-check with INVALID_EXPIRATION_TIME (mono)
+                        // or post-consensus (modular). When we only have modular code left,
+                        // we can just check hasKnownStatus.
+                        .hasPrecheckFrom(OK, INVALID_EXPIRATION_TIME)
+                        .hasKnownStatus(INVALID_EXPIRATION_TIME));
```
Neeharika-Sompalli, line 561: here as well

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
vtronkov: Maybe I don't completely understand the business case here. So we need to add to the fees only if the token is a sentinel value(0.0.0) **and** if the custom fee is of type **FIXED_FEE**? Maybe if you can explain a bit or give me resources for that case that would be great 😄 

I will remove the addition of the fees in the two places you mentioned 👍

vtronkov: I compare the business logic from `NewRels.listFrom`. In it, we are traversing each custom fee and we check if we should require auto association:
```
for (final var customFee : provisionalToken.getCustomFees()) {
      if (customFee.requiresCollectorAutoAssociation()) {
          final var collector = customFee.getValidatedCollector();
          associateGiven(provisionalToken, collector, tokenStore, associatedSoFar, newRels, dynamicProperties);
      }
  }
```
And `requiresCollectorAutoAssociation` is the following:
```
public boolean requiresCollectorAutoAssociation() {
        switch (feeType) {
            case FRACTIONAL_FEE:
                return true;
            case FIXED_FEE:
                return fixedFeeSpec.usedDenomWildcard();
            case ROYALTY_FEE:
                if (royaltyFeeSpec.hasFallbackFee()) {
                    return royaltyFeeSpec.fallbackFee().usedDenomWildcard();
                }
        }
        return false;
    }
```

* For the **FRACTIONAL_FEE** it's always true - that's why I always add it
* For the **ROYALTY_FEE** when `hasFallbackFee()` is true - that's why I added it inside the `royaltyFee.hasFallbackFee()` (to be more precise maybe I should take it out of the `fallbackFee.hasDenominatingTokenId` if statement)

Does this make sense? Or maybe I am looking at something else 😄

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
vtronkov: Thank you! 🎉  I didn't know about this property.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
vtronkov: I had the wrong impression of what the test did. I will give it another go and try to fix it 👍

vtronkov: Okay, I reverted the change and created an issue for that since I can't find what's wrong with this test. https://github.com/hashgraph/hedera-services/issues/8811

vtronkov:
Note: I will fix the unit test when the code is finalized

Neeharika-Sompalli (commented):
Seems some checks are failing in CI

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -999,7 +1013,11 @@ private HapiSpec prechecksWork() {
                                 .hasPrecheck(ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS),
                         tokenAssociate(FIRST_USER, A_TOKEN),
                         cryptoTransfer(moving(0, A_TOKEN).between(TOKEN_TREASURY, FIRST_USER))
-                                .hasPrecheck(OK),
+                                // Either fail pre-check with EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS (mono)
+                                // or post-consensus (modular). When we only have modular code left,
+                                // we can just check hasKnownStatus.
+                                .hasPrecheckFrom(OK, EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS)
+                                .hasKnownStatus(EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS),
```
Neeharika-Sompalli, line 1020: We should not be changing a passing transaction in mono-service to a failing one. Should preserve mono-service behavior

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenCreateSpecs.java
```diff
@@ -286,10 +289,13 @@ public HapiSpec autoRenewValidationWorks() {
...
-                                .autoRenewAccount(AUTO_RENEW)
-                                .autoRenewPeriod(Long.MAX_VALUE)
-                                .hasPrecheck(INVALID_RENEWAL_PERIOD),
+                        // TODO: for some reason setting the autoRenewPeriod does not change the value. We always get
+                        // the default value(7776000). Not sure why
+                        // Other then that the validation itself seems to be in place.
+                        //                        tokenCreate(PRIMARY)
+                        //                                .autoRenewAccount(AUTO_RENEW)
+                        //                                .autoRenewPeriod(Long.MAX_VALUE)
+                        //                                .hasPrecheck(INVALID_RENEWAL_PERIOD),
```
Neeharika-Sompalli, line 298: Can we uncomment this as in mono-service ? If we are not adding @HapiTest annotation

netopyr:
Unit tests are failing

vtronkov:
> Unit tests are failing

Yes, I'm waiting to finalize the code and then I will fix them

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImpl.java
iwsimon: Some tests were failing so I made that change to pass those tests. If you change this back, need to double check those tests.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
mhess-swl: I agree, that does seem cumbersome. I'm not sure if this code is equivalent to that code or not; as long as it has the same end result I would think this is OK, but not sure. @Neeharika-Sompalli?

Neeharika-Sompalli: yeah that should be fine.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
Neeharika-Sompalli: I need to look into this comment. But overall makes sense to me. Approving this. I will definitely look into this today.

Neeharika-Sompalli (dismissed):
LGTM ! I just need to re-verify this [comment](https://github.com/hashgraph/hedera-services/pull/8672/files#r1332853414) . Thanks @vtronkov 👏

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
```diff
@@ -113,7 +113,8 @@ public void pureChecks(@NonNull final CryptoTransferTransactionBody op) throws P
             validateFalsePreCheck(nftIds.size() < nftTransfers.size(), TOKEN_ID_REPEATED_IN_TOKEN_LIST);
 
             // Verify that one and only one of the two types of transfers (fungible or non-fungible) is present
-            validateFalsePreCheck(!nonZeroFungibleValueFound && nftIds.isEmpty(), EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS);
+            validateFalsePreCheck(
+                    uniqueTokenAcctIds.isEmpty() && nftIds.isEmpty(), EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS);
```
mhess-swl, line 117: Don't we need both of these checks? It's ok for `uniqueTokenAcctIds` and `nftIds` to be empty if there's a non-zero fungible value found

vtronkov: I'm not sure about that. And I don't see such a check in mono(PureTransferSemanticChecks.checkTokenTransfersList). But I might be missing something. @Neeharika-Sompalli thoughts?

Neeharika-Sompalli: yes @mhess-swl @vtronkov . We are okay if they are having zero amount. But both should not be set. So, agree with this change to be valid.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
```diff
@@ -113,7 +113,8 @@ public void pureChecks(@NonNull final CryptoTransferTransactionBody op) throws P
...
             validateFalsePreCheck(nonZeroFungibleValueFound && !nftIds.isEmpty(), INVALID_ACCOUNT_AMOUNTS);
```
Neeharika-Sompalli, line 118: Even here I think we should use `!uniqueTokenAcctIds.isEmpty()` @vtronkov  ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CustomFeesValidator.java
Neeharika-Sompalli: Agree with this change. Thanks for fixing it @vtronkov 🙌

Neeharika-Sompalli (approved):
Looks Great ! 💯  Great job for the fixes @vtronkov, Thank you!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
Neeharika-Sompalli: Can be done after this PR is merged

vtronkov: @Neeharika-Sompalli when I change it `prechecksWork` test starts to fail.
Looking at the original code(`PureTransferSemanticChecks.checkTokenTransfersList`) we have the following logic:
```
final var ownershipChangesHere = scopedTransfers.getNftTransfersCount();
final var transferCounts = scopedTransfers.getTransfersCount();
if (ownershipChangesHere > 0) {
    ...
} else {
    if (transferCounts == 0) {
        return EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS;
    }
    count += transferCounts;
}
```
So we are returning `EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS` only when _ownershipChangesHere_ and _transferCounts_ are 0. This seems to correlate to both `uniqueTokenAcctIds` and `nftIds` lists being empty in the new code base. What do you think?

Neeharika-Sompalli: Thanks @vtronkov . 
`transferCounts` is just seeing if the transferList has any items. But `uniqueTokenAcctIds` is checking if they are valid.  I will look at that change and `prechecksWork`

PR #8679, opened 2023-09-18
Fix `TokenManagementSpecs` tests

### LINKED ISSUE #8598: Resolve `TokenManagementSpecs` errors
### Problem

We have the following tests suite `TokenManagementSpecs` that has problems when it's running against modularization codebase.

### Solution

### Alternatives

_No response_

natanasow:
**Description**:

We have the following test suite TokenManagementSpecs that has problems when it's running against the modularization codebase. Resolve issues with the test suite. When ready add @hapitest annotations to the passing tests.

**Related issue(s)**:

Fixes #8598

**Notes for reviewer**:

**Checklist**


netopyr:
I am by no means an expert on the TokenService (yet).

@Neeharika-Sompalli or @mhess-swl: can you please answer the questions?

mhess-swl:
@natanasow

>[This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R121) checks if we provide a zero amount, the tx will be executed, but t[here](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-8e5fdf4373f4013734da47ae4b2d8f302a35a2a679218eaf91baf6882627191aR116) is an explicit check in the handler that throws an exception if the amount is zero - here. I've commented out the check till I found out what's the right behavior. Will we accept 0 amount or we will throw an exception?

Running this on my machine I get `Wrong actual status EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS, expected SUCCESS` on the first `cryptoTransfer(...)` of the test's `.then(...)` statement. According to `PureTransferSemanticChecks` (from the mono code), `EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS` should only be the status when the number of NFT transfers and the number of fungible token transfers are 0. Not sure how this is passing in the mono code... @Neeharika-Sompalli any thoughts?

> [This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R279) makes cryptoDelete, but [there is a check](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-84e777e6cf739208677117f43251e2db4ad8daf633fc46b11f14282001eef567R176) if the deleted account is associated to any tokens. Here I also commented out the check till I found out what is the right behavior. Should we permit cryptoDelete if the account has associated tokens or we should make dissociations from the tokens first?

Running this on my machine I get `Wrong status! Expected SUCCESS, was TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES!` on the `cryptoDelete('misc')` line. In the mono code, `CryptoDeleteTransitionLogic.doStateTransition(...)` checks `ledger.allTokenBalancesVanish(id)`, which does _NOT_ validate token associations. I think the test is correct here and the modularized code has a bug. The crypto delete shouldn't make sure there is no token association – only that the token balance is zero. @Neeharika-Sompalli can you confirm that my reasoning here is correct?

> The last one is [this test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R200). TokenRelation doesn't have kycStatus, it has only isKycGranted which is bool. In the test we revoked the kyc and the response from getAccountInfo for kycStatus is 0 (KYC_NOT_APPLICABLE) when it should be 2 (REVOKED). Do you have some hints on how to debug where the problem is? Maybe we should add a new property in TokenRelation for kycStatus?

Running this on my machine I get `Wrong status! Expected ACCOUNT_KYC_NOT_GRANTED_FOR_TOKEN, was SUCCESS!` on the `mintToken(...)` line, which is different than what you described here. First I would recommend updating your copy of the `develop` branch, and if you still get a different error, write down which branch this is happening on (and any other changes made) so we can recreate the different error. Otherwise, it looks like the token mint handler should be checking that a kyc is granted for the token, but it isn't, and is therefore allowing the operation (when the operation should not be allowed). 

CC @netopyr @Neeharika-Sompalli

Neeharika-Sompalli:
Agree with @mhess-swl on the outputs we are seeing in new code. We should definitely make sure new code behaves the same as mono-service code. 

> @netopyr there are several misunderstandings IMO
> 
> * [This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R121) checks if we provide a zero amount, the tx will be executed, but there is an explicit check in the handler that throws an exception if the amount is zero - [here](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-8e5fdf4373f4013734da47ae4b2d8f302a35a2a679218eaf91baf6882627191aR116). I've commented out the check till I found out what's the right behavior. Will we accept 0 amount or we will throw an exception?

We should allow zero amount transfers. In mono-service we fail this way when [number of transfers](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/PureTransferSemanticChecks.java#L134) in `transferList` is zero. Please modify the logic in handler to do so, in new code. 

> * [This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R279) makes cryptoDelete, but [there is a check](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-84e777e6cf739208677117f43251e2db4ad8daf633fc46b11f14282001eef567R176) if the deleted account is associated to any tokens. Here I also commented out the check till I found out what is the right behavior. Should we permit cryptoDelete if the account has associated tokens or we should make dissociations from the tokens first?

Test is correct. It seems there is a bug the way we are setting `numPositiveBalances` to a negative number [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAccountWipeHandler.java#L174). It should be `updatedAcctBuilder.numberPositiveBalances(Math.max(acct.numberPositiveBalances() - 1, 0));` 

> * The last one is [this test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R200). `TokenRelation` doesn't have `kycStatus`, it has only `isKycGranted` which is bool. In the test we revoked the kyc and the response from `getAccountInfo` for `kycStatus` is 0 (KYC_NOT_APPLICABLE) when it should be 2 (REVOKED). Do you have some hints on how to debug where the problem is? Maybe we should add a new property in `TokenRelation` for `kycStatus`?

Neeharika-Sompalli:
@natanasow

Agree with @mhess-swl on the outputs we are seeing in new code. We should definitely make sure new code behaves the same as mono-service code. 

> @netopyr there are several misunderstandings IMO
> 
> * [This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R121) checks if we provide a zero amount, the tx will be executed, but there is an explicit check in the handler that throws an exception if the amount is zero - [here](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-8e5fdf4373f4013734da47ae4b2d8f302a35a2a679218eaf91baf6882627191aR116). I've commented out the check till I found out what's the right behavior. Will we accept 0 amount or we will throw an exception?

We should allow zero amount transfers. In mono-service we fail this way when [number of transfers](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/PureTransferSemanticChecks.java#L134) in `transferList` is zero. Please modify the logic in handler to do so, in new code. 

> * [This test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R279) makes cryptoDelete, but [there is a check](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-84e777e6cf739208677117f43251e2db4ad8daf633fc46b11f14282001eef567R176) if the deleted account is associated to any tokens. Here I also commented out the check till I found out what is the right behavior. Should we permit cryptoDelete if the account has associated tokens or we should make dissociations from the tokens first?

Test is correct. It seems there is a bug the way we are setting `numPositiveBalances` to a negative number [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAccountWipeHandler.java#L174). It should be `updatedAcctBuilder.numberPositiveBalances(Math.max(acct.numberPositiveBalances() - 1, 0));` 

> * The last one is [this test](https://github.com/hashgraph/hedera-services/pull/8679/files#diff-0653dd5d13d746d05ac8c6de8b7d674e76ddfbf5a11a15456b93cafa32eedfc8R200). `TokenRelation` doesn't have `kycStatus`, it has only `isKycGranted` which is bool. In the test we revoked the kyc and the response from `getAccountInfo` for `kycStatus` is 0 (KYC_NOT_APPLICABLE) when it should be 2 (REVOKED). Do you have some hints on how to debug where the problem is? Maybe we should add a new property in `TokenRelation` for `kycStatus`?

cc : @netopyr @mhess-swl

natanasow:
@mhess-swl @Neeharika-Sompalli Thank you! I have all the answers now :pray:.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
```diff
@@ -115,6 +115,12 @@ public void pureChecks(@NonNull final CryptoTransferTransactionBody op) throws P
             // Verify that one and only one of the two types of transfers (fungible or non-fungible) is present
             validateFalsePreCheck(!nonZeroFungibleValueFound && nftIds.isEmpty(), EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS);
             validateFalsePreCheck(nonZeroFungibleValueFound && !nftIds.isEmpty(), INVALID_ACCOUNT_AMOUNTS);
+
+            // Verify that if there is only one fungible transfer and no non-fungible transfers,
+            // the fungible transfer should not be empty
+            if (fungibleTransfers.size() == 1 && nftIds.isEmpty()) {
```
natanasow, line 121: @Neeharika-Sompalli  @mhess-swl  

Without a newly added check, this [test](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/CryptoTransferHandlerPureChecksTest.java#L289) was failing. Is it valid or we should edit the test and let the validator be as it was before?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
```diff
@@ -92,7 +92,7 @@ public void pureChecks(@NonNull final CryptoTransferTransactionBody op) throws P
                 validateTruePreCheck(acctAmount.hasAccountID(), INVALID_TRANSFER_ACCOUNT_ID);
                 uniqueTokenAcctIds.add(acctAmount.accountIDOrThrow());
                 netTokenBalance += acctAmount.amount();
-                if (!nonZeroFungibleValueFound && acctAmount.amount() != 0) {
+                if (!nonZeroFungibleValueFound && acctAmount.amount() >= 0) {
```
mhess-swl, line 95: @Neeharika-Sompalli shouldn't we allow a negative account amount here (i.e. `!= 0` is correct)? I thought we did allow negative amounts, but can't remember for sure

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
mhess-swl: This check should've been taken care of by the lines directly above, shouldn't they @Neeharika-Sompalli?

i.e.
```
validateFalsePreCheck(!nonZeroFungibleValueFound && nftIds.isEmpty(), EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS);
validateFalsePreCheck(nonZeroFungibleValueFound && !nftIds.isEmpty(), INVALID_ACCOUNT_AMOUNTS);
```

At the very least, thes new line should be `fungibleTransfers.size() > 0` instead of ` == 1`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenManagementSpecs.java
```diff
@@ -499,10 +516,10 @@ public HapiSpec supplyMgmtFailureCasesWork() {
                         burnToken(RIGID, 1).signedBy(GENESIS).hasKnownStatus(TOKEN_HAS_NO_SUPPLY_KEY),
                         mintToken(SUPPLE, Long.MAX_VALUE).hasKnownStatus(INVALID_TOKEN_MINT_AMOUNT),
                         mintToken(SUPPLE, 0).hasPrecheck(OK),
-                        mintToken(SUPPLE, -1).hasPrecheck(INVALID_TOKEN_MINT_AMOUNT),
+                        mintToken(SUPPLE, -1).hasKnownStatus(INVALID_TOKEN_MINT_AMOUNT),
                         burnToken(SUPPLE, 2).hasKnownStatus(INVALID_TOKEN_BURN_AMOUNT),
                         burnToken(SUPPLE, 0).hasPrecheck(OK),
-                        burnToken(SUPPLE, -1).hasPrecheck(INVALID_TOKEN_BURN_AMOUNT));
+                        burnToken(SUPPLE, -1).hasKnownStatus(INVALID_TOKEN_BURN_AMOUNT));
```
mhess-swl: I think these statuses need to be added to the list of special ingest exceptions (`spec.streamlinedIngestChecks` in `spec-default.properties`) instead of changing from `hasPrecheck` to `hasKnownStatus`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGrantKycToAccountHandler.java
```diff
@@ -68,6 +69,8 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         if (tokenMeta == null) throw new PreCheckException(INVALID_TOKEN_ID);
         if (tokenMeta.hasKycKey()) {
             context.requireKey(tokenMeta.kycKey());
+        } else {
+            throw new PreCheckException(TOKEN_HAS_NO_KYC_KEY);
```
mhess-swl, line 73: I don't think this is correct. The `validate(...)` method [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/token/GrantKycLogic.java#L57) suggests this is not required in precheck

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -114,6 +114,9 @@ protected void changeSupply(
         requireNonNull(token);
         requireNonNull(treasuryRel);
         requireNonNull(invalidSupplyCode);
+        if (token.hasKycKey()) {
+            validateTrue(treasuryRel.kycGranted(), ACCOUNT_KYC_NOT_GRANTED_FOR_TOKEN);
+        }
```
mhess-swl, line 119: I don't see this added logic anywhere in [the old code](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/models/Token.java#L359). I'm not sure this is correct

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenManagementSpecs.java
natanasow: Done.

mhess-swl (approved):

mhess-swl (changes requested):
Whoops, shouldn't have approved just yet. We still need to resolve the remaining PR comments, e.g. https://github.com/hashgraph/hedera-services/pull/8679#discussion_r1339515293

natanasow:
@mhess-swl @Neeharika-Sompalli @netopyr 

Just updated with the latest develop and most of the fixes are already implemented in another PRs, so could you review it from scratch one last time :pray:

mhess-swl (approved):

Issue #5424, opened 2023-03-06
Logging Architecture - Base Logger


hendrikebbers:
## Custom logging facade

The design outlined is based on the development of a bespoke logging facade, tailored specifically to meet our unique requirements. Recognizing that components from this repository may eventually be incorporated into different software, we must consider this during development. 
Consequently, it is essential to implement an adapter capable of redirecting all events from our custom facade to SLF4J. While this adapter may not be utilized directly within this repository, it will be available for any projects that rely on modules from our repository.

### Why we need a custom logging facade

To effectively aggregate logging events and messages from both our modules and third-party libraries, the use of a logging facade is necessary. In the Java ecosystem, [SLF4J](https://www.slf4j.org/index.html) is widely recognized as the predominant logging facade, primarily due to its ability to consolidate logs from various logging libraries. Another crucial aspect of a logging facade is the distinction between its API and its implementation. In this context, Log4J2's API module stands out as it can function as a facade with a custom logging implementation. Log4J2 has certain advantages over SLF4J, including [additional features](https://logging.apache.org/log4j/2.x/#api-separation). Additionally, Java offers a basic logging facade through `java.lang.System.Logger`.

However, each of these options comes with its own set of limitations, leading us to the conclusion that a custom logging facade is the most suitable solution for our needs:

- SLF4J introduced a fluent API supporting lambdas (Supplier for messages) from version 2 onwards. However, this API is distinct from the classic one to avoid disrupting the standard SLF4J API. This separation results in a less than ideal API definition.
- The maintenance of SLF4J is primarily shouldered by a single individual, raising concerns about its sustainability and responsiveness to evolving needs.
- The `java.lang.System.Logger` offers a very basic API, lacking several functionalities that are essential for our purposes.
- While the Log4J2 API is extensive, encompassing over 100 classes, it presents challenges in implementation due to its complexity. Moreover, many of its functionalities are surplus to our requirements.

Considering these factors, a tailor-made logging facade emerges as the optimal approach, ensuring that our specific needs are addressed effectively.

## Collecting all logging messages and events

We must collect all logging messages from all loggers on the classpath at runtime. There are several patterns that support collecting log messages:
- `java.lang.System.Logger` is just a facade that forwards all messages to an underlying logger
- SLF4J provides Adapters to collect the logging from several logging frameworks.
- SLF4J is often used directly and all messages that are created in SLF4 or passed to SLF4J can easily be forwarded to any concrete logging implementation

[...]

PR #8680, opened 2023-09-18
Emergency logger + needed dependencies added

hendrikebbers:
Part of #5424 
This adds first parts of the API (LogEvent) and the emergency logger that is used internally in the logger to log events.

In that PR the `Level` enum and the `Marker` class are the only 2 classes that are part of the private API (that will be exposed to all libs). 

The classes below `com.swirlds.logging.v2.extensions` will be available for plugins like a "Write-To-Console-Handler", a "Write-To-File-Handler" or a "Log4j2-Provider".


platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/extensions/emergency/EmergencyLogger.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    void log(@NonNull LogEvent event);
+
+    /**
+     * Logs a message.
+     *
+     * @param level   the level of the message
+     * @param message the message to log
+     */
+    default void log(@NonNull Level level, @NonNull String message) {
```
cody-littley, line 51:
```suggestion
    default void log(@NonNull final Level level, @NonNull final String message) {
```

cody-littley (approved):

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/extensions/event/LogEventConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+public interface LogEventConsumer extends Consumer<LogEvent> {
+
+    /**
+     * Checks if the consumer is enabled for the given name and level.
+     *
+     * @param name  the name
+     * @param level the level
+     * @return true if the consumer is enabled, false otherwise
+     */
+    default boolean isEnabled(@NonNull String name, @NonNull Level level) {
```
lpetrovic05, line 35: Should this have a default implementation? Seems like something its easy to forget to override

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/extensions/event/ParameterizedLogMessage.java
```diff
@@ -0,0 +1,129 @@
...
+ * @see LogMessage
+ */
+public record ParameterizedLogMessage(@NonNull String messagePattern, @NonNull Object... args) implements LogMessage {
+
+    static final char DELIM_START = '{';
+    static final String DELIM_STR = "{}";
+    private static final char ESCAPE_CHAR = '\\';
+
+    @Override
+    public String getMessage() {
```
lpetrovic05, line 40: In my mind, a getter should not be doing so much work, seems counterintuitive

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/internal/emergency/EmergencyLoggerImpl.java
```diff
@@ -0,0 +1,263 @@
...
+    public void logNPE(@NonNull final String nameOfNullParam) {
+        log(
+                Level.ERROR,
+                "Null parameter: " + nameOfNullParam,
+                new NullPointerException("Null parameter: " + nameOfNullParam));
+    }
+
+    @Override
+    public void log(@NonNull final Level level, @NonNull final String message, @Nullable final Throwable thrown) {
+        if (level == null && message == null) {
```
lpetrovic05, line 127: level and message as marked with `@NonNull`, then checked for `null`

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/extensions/event/ParameterizedLogMessage.java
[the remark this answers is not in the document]
hendrikebbers: good question: It is defined in the interface. Do we than define it in each implementation?

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/internal/emergency/EmergencyLoggerImpl.java
hendrikebbers: The annotation does not forbid to call it with null. It is only checked in sonar.

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/v2/extensions/event/ParameterizedLogMessage.java
hendrikebbers: Well, it is mentioned in the java doc of the interface that calling the method can create some workload. java.util.function.Supplier uses get() as function name, too. Based on that I would disagree here. 
But: That need to be changed in future anyway since the value need to be cached and should not calculated every time the message is called. Based on that it might look more like a getter in future.

cody-littley (approved):

PR #8692, opened 2023-09-18
Error handling for penalty payments

### LINKED ISSUE #8686: Many tests fail with UNKNOWN status
Many tests fail with UNKNOWN status. The main issue is that we do not handle failed penalty charges correctly.

netopyr:
This change introduces proper error handling while we charge penalties in the `HandleWorkflow`.

Closes #8686


vtronkov (approved):

georgi-l95 (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -348,12 +352,22 @@ private void handleUserTransaction(
...
+                try {
+                    final var penaltyPayerID =
+                            validationResult.status() == NODE_DUE_DILIGENCE_FAILURE ? creator.accountId() : payer;
+                    final var penaltyFee = new Fees(fees.nodeFee(), fees.networkFee(), 0L);
+                    feeAccumulator.charge(penaltyPayerID, penaltyFee);
+                } catch (HandleException ex) {
+                    final var identifier = validationResult.status == NODE_DUE_DILIGENCE_FAILURE
+                            ? "node " + creator.nodeId()
+                            : "account " + payer;
+                    logger.error(
```
povolev15, line 365: why do we log it and not throw

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -381,7 +395,15 @@ private void handleUserTransaction(
             logger.error("An unexpected exception was thrown during handle", e);
             rollback(ResponseCodeEnum.FAIL_INVALID, stack, recordListBuilder);
             if (payer != null && fees != null) {
-                feeAccumulator.charge(payer, fees);
+                try {
+                    feeAccumulator.charge(payer, fees);
+                } catch (HandleException chargeException) {
+                    logger.error(
+                            "Unable to charge account {} a penalty after an unexpected exception {}. Cause of the failed charge:",
+                            payer,
```
povolev15, line 403: same here

povolev15 (approved):
other than the 2 comments LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
povolev15: What does stop us from logging and throwing it?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
netopyr: The `TransactionReceipt` would contain the status `FAIL_INVALID` instead of the more specific error code that is available here. What would we gain if we threw an `Exception` here?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
netopyr: As written above, an `Exception` would skip creating and storing the `TransactionReceipt`, i.e., a user would not be able to see that the transaction failed.

rbair23 (approved):

PR #8700, opened 2023-09-18
Skip `0.0.0` auto-renew account id in synthetic `ContractCreate`

### LINKED ISSUE #8475: Sentinel missing id (`0.0.0`) should not be stored explicitly
### Description

Contract accounts without auto-renew accounts are setting `Entity.MISSING_ID` as their auto-renew account.

If such a contract does an internal `CONTRACT_CREATION`, the synthetic `ContractCreateTransactionBody` will appear to set the new child's auto-renew account id to `0.0.0`.

We don't want this, so we should stop using the sentinel `0.0.0` anywhere in state; and ensure that synthetic `ContractCreateTransactionBody`'s don't include it either.

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

mainnet

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #8475

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/HederaAccountCustomizer.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/HederaAccountCustomizer.java
@@ -81,7 +81,10 @@ public final class HederaAccountCustomizer
             op.setDeclineReward((boolean) changes.get(AccountProperty.DECLINE_REWARD));
         }
         if (changes.containsKey(AccountProperty.AUTO_RENEW_ACCOUNT_ID)) {
-            op.setAutoRenewAccountId(((EntityId) changes.get(AccountProperty.AUTO_RENEW_ACCOUNT_ID)).toGrpcAccountId());
+            final var autoRenewAccountId = (EntityId) changes.get(AccountProperty.AUTO_RENEW_ACCOUNT_ID);
+            if (!EntityId.MISSING_ENTITY_ID.equals(autoRenewAccountId)) {
+                op.setAutoRenewAccountId(autoRenewAccountId.toGrpcAccountId());
+            }
         }
         if (changes.containsKey(AccountProperty.MAX_AUTOMATIC_ASSOCIATIONS)) {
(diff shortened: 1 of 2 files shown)


mhess-swl (approved):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/HederaAccountCustomizer.java
```diff
@@ -81,7 +81,10 @@ public void customizeSynthetic(final ContractCreateTransactionBody.Builder op) {
             op.setDeclineReward((boolean) changes.get(AccountProperty.DECLINE_REWARD));
         }
         if (changes.containsKey(AccountProperty.AUTO_RENEW_ACCOUNT_ID)) {
-            op.setAutoRenewAccountId(((EntityId) changes.get(AccountProperty.AUTO_RENEW_ACCOUNT_ID)).toGrpcAccountId());
+            final var autoRenewAccountId = (EntityId) changes.get(AccountProperty.AUTO_RENEW_ACCOUNT_ID);
```
Neeharika-Sompalli, line 84: Is this probably because we don't validate auto_renew_account_id is valid before customizing ?

iwsimon (approved):
LGTM

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/HederaAccountCustomizer.java
tinker-michaelj: Afaict it is only to have `EntityId.MISSING_ID` passed into this method when:
 - The "parent" contract is a `MerkleAccount` (**not** `OnDiskAccount`) whose `autoRenewId` field is set to `EntityId.MISSING_ID`.

And **that** is only possible when the parent contract is a `MerkleAccount` that:
 - Originally had an `autoRenewId`; and then,
 - Had that cleared to `0.0.0` via a `ContractUpdate`.

Since we plan to move accounts to disk in 0.42, the first `MerkleAccount` condition will no longer be met. So this check is just an extra defense measure, in case there is some other possibility I am not seeing. 🤷

Neeharika-Sompalli (approved):

Issue #8463, opened 2023-09-08
Implement `calculateFees` in all handlers of token service


Neeharika-Sompalli:
- [x] #8833
- [x] #8748

Tokens: 
- [x] #8725
- [x] #8726
- [x] #8806

PR #8708, opened 2023-09-18
Fix fees for some of Token operations

Neeharika-Sompalli:
Fixes some operations in #8463 
Fixes https://github.com/hashgraph/hedera-services/issues/8725

- TokenMint
- TokenBurn
- TokenPause
- TokenUnpause
- TokenFreeze
- TokenUnfreeze
- TokenCreate
Enabled TokenPause fee specs


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
```diff
@@ -239,4 +264,53 @@ private Nft buildNewlyMintedNft(
...
+
+        final var calculator = feeContext.feeCalculator(subType);
+        if (SubType.TOKEN_NON_FUNGIBLE_UNIQUE.equals(subType)) {
+            // The price of nft mint should be increased based on number of signatures.
+            // The first signature is free and is accounted in the base price, so we only need to add
+            // the price of the rest of the signatures.
+            calculator.addVerificationsPerTransaction(Math.max(0, numSimpleKeys(payerKey) - 1L));
+        }
+
+        final var meta = TOKEN_OPS_USAGE_UTILS.tokenMintUsageFrom(fromPbj(feeContext.body()), fromPbj(subType), 76000L);
```
iwsimon, line 290: What's this 7600L number? Please add comments here.

Neeharika-Sompalli: yes , that parameter is not used. So added a default value. Will add comments

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
```diff
@@ -239,4 +264,53 @@ private Nft buildNewlyMintedNft(
...
+            for (final var k : keys) {
+                numSimpleKeys(k);
+            }
+        } else if (key.hasKeyList()) {
+            final var keys = key.keyListOrElse(KeyList.DEFAULT).keysOrElse(Collections.emptyList());
+            for (final var k : keys) {
+                numSimpleKeys(k);
+            }
+        } else {
+            count.incrementAndGet();
```
tinker-michaelj, line 312: Was going to leave a comment about how `ContractID` keys don't make sense to count here; then checked [mono equivalent `calculateKeysMetadata()`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/fee/FeeBuilder.java#L231) and realized this is the behavior we need to emulate. So well played. 😇 

We need a shorthand comment for "not ideal but required for differential testing."

tinker-michaelj (approved):
LGTM, thx @Neeharika-Sompalli !

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
Neeharika-Sompalli: Added comment

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
Neeharika-Sompalli: Added a comment. Thanks @tinker-michaelj

tinker-michaelj (approved):
LGTM

iwsimon (approved):
LGTM!

Issue #8481, opened 2023-09-09
Implement "classic" HTS view calls


tinker-michaelj:
### Problem

We need to implement the non-ERC, "classic" static calls in the `HtsSystemContract`:
  - [x] #8638 
  - [x] #8639 
  - [x] #8640 
  - [x] #8642 
  - [x] #8643 
  - [x] #8644 
  - [x] #8645 
  - [x] #8646 
  - [x] #8647 
  - [x] #8648 
  - [x] #8649 
  - [x] #8650 

### Solution

Implement that above.

PR #8712, opened 2023-09-19
Implement some of the classic HTS view functions

lukelee-sl:
**Description**:
Implements:
IS_FROZEN
IS_KYC
IS_TOKEN
TOKEN_TYPE
DEFAULT_FREEZE_STATUS
DEFAULT_KYC_STATUS

The remaining classic view function calls will be implemented in a separate pr

**Related issue(s)**: #8481

Fixes #

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/defaultfreezestatus/DefaultFreezeStatusCall.java
```diff
@@ -0,0 +1,49 @@
...
+    public DefaultFreezeStatusCall(
+            @NonNull final HederaWorldUpdater.Enhancement enhancement, @Nullable final Token token) {
+        super(enhancement, token);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected @NonNull HederaSystemContract.FullResult resultOfViewingToken(@Nullable final Token token) {
```
tinker-michaelj, line 40: Here I think we may need a new variant of `AbstractTokenViewCall` that instead of reverting [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractTokenViewCall.java#L45) on a null `token`, does e.g.
```
if (token == null) {
    return gasOnly(viewCallResultWith(INVALID_TOKEN_ID, 0L));
}
```
where there's a method
```
abstract FullResult viewCallResultWith(ResponseCodeEnum status);
```
that subclasses need to override.

(Or maybe we could keep just `AbstractTokenViewCall` but add methods,
```
protected boolean revertsOnMissingToken() {
    return true;
}

protected FullResult viewCallResultWith(ResponseCodeEnum status) {
    throw new UnsupportedOperationException();
}
```

?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/isfrozen/IsFrozenCall.java
```diff
@@ -0,0 +1,63 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    protected @NonNull HederaSystemContract.FullResult resultOfViewingToken(@NonNull final Token token) {
+        requireNonNull(token);
+        // TODO - gas calculation
+
+        final var accountNum = accountNumberForEvmReference(account, nativeOperations());
+        if (accountNum < 0) {
+            return revertResult(INVALID_ACCOUNT_ID, 0L);
```
tinker-michaelj, line 56: As above, the classic view calls (c.f. `isFrozen()` signature [here](https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L606)) don't revert on a failure of this kind; instead they run to `COMPLETED_SUCCESS` from EVM perspective and return the `INVALID_ACCOUNT_ID` failure encoded in their output `Bytes.`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/isfrozen/IsFrozenCall.java
```diff
@@ -0,0 +1,75 @@
...
+        requireNonNull(token);
+        // TODO - gas calculation
+
+        final var accountNum = accountNumberForEvmReference(account, nativeOperations());
+        if (accountNum < 0) {
+            return fullResultsFor(INVALID_ACCOUNT_ID, 0L, false);
+        }
+        var tokenRel = nativeOperations()
+                .getTokenRelation(accountNum, token.tokenIdOrThrow().tokenNum());
+        var result = tokenRel == null ? false : tokenRel.frozen();
```
david-bakin-sl, line 61: (interesting it is not an error to ask if a token you don't hold is frozen)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/iskyc/IsKycCall.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    protected @NonNull FullResult viewCallResultWith(@NonNull ResponseCodeEnum status, long gasRequirement) {
+        return fullResultsFor(status, gasRequirement, false);
+    }
+
+    private @NonNull FullResult fullResultsFor(@NonNull ResponseCodeEnum status, long gasRequirement, boolean isKyc) {
+        return successResult(
+                DEFAULT_FREEZE_STATUS.getOutputs().encodeElements(status.protoOrdinal(), isKyc), gasRequirement);
```
david-bakin-sl, line 72: 🪳 `DEFAULT_FREEZE_STATUS` -> `IS_KYC`? 

(Maybe others should be checked too ... I'm not going backwards to find out, sorry)

hedera-node/hedera-app/src/xtest/java/contract/ClassicViewsXTestConstants.java
```diff
@@ -0,0 +1,47 @@
...
+import com.hedera.hapi.node.base.FileID;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+
+public class ClassicViewsXTestConstants {
+    static final FileID CLASSIC_VIEWS_INITCODE_FILE_ID = new FileID(0, 0, 1029L);
+    static final ContractID CLASSIC_QUERIES_X_TEST_ID =
+            ContractID.newBuilder().contractNum(1030L).build();
+    static final String SUCCESS_RESPONSE_CODE = "0000000000000000000000000000000000000000000000000000000000000016";
+    static final Bytes TOKEN_IS_FROZEN =
+            Bytes.fromHex(SUCCESS_RESPONSE_CODE + "0000000000000000000000000000000000000000000000000000000000000000");
```
david-bakin-sl, line 30: (looking strictly at the hex it's a mystery why sometimes success response is `0` and other times it is `1` - maybe a comment to explain that? like, what's the difference between is frozen and is kyc?)

hedera-node/hedera-app/src/xtest/java/contract/ClassicViewsXTestConstants.java
```diff
@@ -0,0 +1,47 @@
...
+
+package contract;
+
+import com.esaulpaugh.headlong.abi.Function;
+import com.hedera.hapi.node.base.ContractID;
+import com.hedera.hapi.node.base.FileID;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+
+public class ClassicViewsXTestConstants {
+    static final FileID CLASSIC_VIEWS_INITCODE_FILE_ID = new FileID(0, 0, 1029L);
```
david-bakin-sl, line 25: (Where does 1029 here and 1030 on the next line come from? Is it that we know in what order we create things, and we always start at 1028, or what?)

david-bakin-sl (changes requested):
1) Didn't review unit tests
2) Did not do the work of verifying revert/non-revert against mono service the way you did

One (minor) bug, otherwise fine.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractNonRevertibleTokenViewCall.java
```diff
@@ -0,0 +1,70 @@
...
+    @NonNull
+    protected abstract HederaSystemContract.FullResult resultOfViewingToken(@NonNull Token token);
+
+    /**
+     * Returns the result of viewing the given {@code token} given the {@code status}.
+     * @param status - ResponseCodeEnum status
+     * @return the results to return to the caller
+     */
+    @NonNull
+    protected abstract FullResult viewCallResultWith(ResponseCodeEnum status, long gasRequirement);
```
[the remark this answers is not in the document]
lukelee-sl: Yes in every known usage thus far but one can image that this function could be used to create SUCCESS results as well.  I'll try to make the current usage more clear in the javadocs

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/iskyc/IsKycCall.java
lukelee-sl: yikes - thanks for catching this.  I'll double check the other call functions.  Too much copy/pasting of code.

hedera-node/hedera-app/src/xtest/java/contract/ClassicViewsXTestConstants.java
lukelee-sl: The xtest initialization methods create a bunch of entities including the file for the contract bytecode.  It is assumed that 30 entities are enough because `MiscViewsXTestConstants.NUM_ENTITY_NUM = 1030L` and this demarks the first entity number created during the test.  The pattern is just to use 29 for the initcode file as I think this is the last entity that needs to be pre-created.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenAndTypeCheckSuite.java
```diff
@@ -124,7 +122,7 @@ private HapiSpec checkTokenAndTypeNegativeCases() {
         final AtomicReference<TokenID> vanillaTokenID = new AtomicReference<>();
         final var notAnAddress = new byte[20];
 
-        return defaultHapiSpec("checkTokenAndTypeNegativeCases")
+        return onlyDefaultHapiSpec("checkTokenAndTypeNegativeCases")
```
david-bakin-sl, line 125: probably not intended to be `only...`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenAndTypeCheckSuite.java
```diff
@@ -138,18 +136,18 @@ private HapiSpec checkTokenAndTypeNegativeCases() {
...
                         spec,
-                        contractCall(
-                                        TOKEN_AND_TYPE_CHECK_CONTRACT,
-                                        IS_TOKEN,
-                                        HapiParserUtil.asHeadlongAddress(notAnAddress))
-                                .via("FakeAddressTokenCheckTx")
-                                .payingWith(ACCOUNT)
-                                .gas(GAS_TO_OFFER),
-                        contractCall(
+                        //                        contractCallLocal(
```
david-bakin-sl, line 139: should reference that one issue (in comment), or not?

david-bakin-sl (approved):
LGTM (modulo that `onlyDefaultHapiSpec` thing)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/isfrozen/IsFrozenCall.java
david-bakin-sl: not a problem, it's just hard to draw the line which things are worth failing and which are worth ignoring

david-bakin-sl (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/defaultfreezestatus/DefaultFreezeStatusCall.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    protected @NonNull FullResult viewCallResultWith(
+            @NonNull final ResponseCodeEnum status, final long gasRequirement) {
+        return fullResultsFor(status, gasRequirement, false);
+    }
+
+    private @NonNull FullResult fullResultsFor(
+            @NonNull final ResponseCodeEnum status, final long gasRequirement, final boolean freezeStatus) {
+        // @Future remove to revert #9067 after modularization is completed
+        if (isStaticCall && status != SUCCESS) {
```
tinker-michaelj, line 64: Not so bad I guess... 😵‍💫

tinker-michaelj (approved):
LGTM, tyvm @lukelee-sl !

Issue #8637, opened 2023-09-15
Fix ContractCreateSuite tests


MiroslavGatsanoga:
Fix `ContractCreateSuite` for modularized codebase.

PR #8713, opened 2023-09-19
Fix tests from ContractCreateSuite

MiroslavGatsanoga:
**Description**:

Fixes for some of the `ContractCreateSuite` failing tests.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8637

**Notes for reviewer**:

I did some debugging for the rest of the tests left to be fixed and here is some of the things I have found that can give a direction for their fixes:

- `rejectsInsufficientFee`: looks like it would need calculateFees() implementation for ContractCreateHandler
- `delegateContractIdRequiredForTransferInDelegateCall`: spent some time debugging it but it does not look like something straightforward to fix, seems to be failing due to the initial transfer to the beneficiary in the contract function for some reason
- `childCreationsHaveExpectedKeysWithOmittedAdminKey`: fails due to incorrect implementation for getCreatedContractIDs
- `vanillaSuccess`: failure seems to be caused by same reason as in `childCreationsHaveExpectedKeysWithOmittedAdminKey`
- `createContractWithStakingFields`: staking id logic from ContractCreateTransitionLogic.validate() should be implemented in mod codebase
- `blockTimestampChangesWithinFewSeconds`: seems like it's caused by contract call local being handled as transaction instead of query
- `cannotCreateTooLargeContract`: failing due to too large file and after that is fixed it will fail with FILE_CONTENT_EMPTY a couple of lines later in the test

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
@@ -208,7 +208,8 @@ public class HandleHederaOperations implements HederaOperations {
         dispatchAndMarkCreation(
                 number,
                 synthAccountCreationFromHapi(
-                        ContractID.newBuilder().contractNum(number).build(), evmAddress, impliedContractCreation));
+                        ContractID.newBuilder().contractNum(number).build(), evmAddress, impliedContractCreation),
+                null);
     }
 
     /**
@@ -221,7 +222,8 @@ public class HandleHederaOperations implements HederaOperations {
         dispatchAndMarkCreation(
                 number,
                 synthAccountCreationFromHapi(
-                        ContractID.newBuilder().contractNum(number).build(), evmAddress, body));
+                        ContractID.newBuilder().contractNum(number).build(), evmAddress, body),
+                body.autoRenewAccountId());
     }
 
     /**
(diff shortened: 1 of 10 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -208,7 +208,8 @@ public void createContract(final long number, final long parentNumber, @Nullable
         dispatchAndMarkCreation(
                 number,
                 synthAccountCreationFromHapi(
-                        ContractID.newBuilder().contractNum(number).build(), evmAddress, impliedContractCreation));
+                        ContractID.newBuilder().contractNum(number).build(), evmAddress, impliedContractCreation),
+                null);
```
tinker-michaelj, line 212: Can we set this to `parent.autoRenewAccountId()`?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
```diff
@@ -55,6 +55,10 @@ public void handle(@NonNull final HandleContext context) throws HandleException
                 .contractCallResult(outcome.result())
                 .contractID(outcome.recipientIdIfCalled())
                 .status(outcome.status());
+
+        if (!outcome.isSuccess()) {
```
tinker-michaelj, line 59: Great catch, that is also fixed in [this PR](https://github.com/hashgraph/hedera-services/pull/8635) but at the rate I'm getting approvals, you will merge first. 😂

tinker-michaelj (approved):
Nice work @MiroslavGatsanoga 🥂

PR #8733, opened 2023-09-19
Support "isolated" flag on HapiTestSuite

### LINKED ISSUE #8687: Add `isolated` as a boolean option to the `HapiTestSuite` annotation
### Problem

Some tests, such as those related to testing "freeze", leave the network in a state where it cannot be reused for subsequent tests. Yet for efficiency, we create a single network at the start of all test suites, and reuse that same network for all tests.

### Solution

We propose adding an `isolated` boolean to the `HapiTestSuite` annotation. If true, then a new network is created for the duration of that suite and the old one thrown away, and at the end of the suite the network is thrown away and a new fresh one created.

### Alternatives

We could create a new network at the start of each suite, but creating the network takes a lot of time. The network (currently by design) even has a 10second wait at startup. When multiplied by many tests, these startup and teardown operations just take a large amount of time.

We could make such tests manual, but then we don't get as much test coverage since they cannot be run on each PR.

We could add to the test a call to "bounce" the nodes (when that HapiOp is merged) and have it bring the network back into a working state, although that is more of a restart test.

We could remove the freeze test completely and just use the restart tests for it.

rbair23:
**Description**:
Adds an 'isolated' flag to `HapiTestSuite`. This flag, when used, causes a fresh genesis restart of the cluster for the sake of the test suite.

To cause this to work reliably, I had to disable running "Alice" in the JUnit JVM instance. Instead, all four nodes are run in a separate VM. A future PR could enhance this so that we normally run a single out of process Alice node and only run multiple nodes on demand with another attribute on `HapiTestSuite`.

**Related issue(s)**:

Fixes #8687


hedera-node/configuration/dev/settings.txt
```diff
@@ -41,7 +41,6 @@ reconnect.asyncStreamTimeoutMilliseconds,      60000
 metrics.csvFileName,                           MainNetStats
 metrics.csvOutputFolder,                       data/stats
 showInternalStats,                             true
-prometheus.endpointEnabled,                    false                        # differs from mainnet
```
rbair23, line 44: I want the prometheus endpoint to run in dev, so we can poll it to know the status of the node.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -242,6 +242,14 @@ public boolean isActive() {
                 && daggerApp.grpcServerManager().isRunning();
     }
 
+    /**
+     * Indicates whether this node is FROZEN.
+     * @return True if the platform is frozen
+     */
+    public boolean isFrozen() {
+        return platformStatus == PlatformStatus.FREEZE_COMPLETE;
+    }
```
rbair23, line 251: Added so that the node control code in the test system can know if the node is frozen. Only applies to an in-process "Alice", which we are not using in this PR. So technically, this method wasn't needed in this PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -435,26 +443,33 @@ public void init(@NonNull final Platform platform, @NonNull final NodeId nodeId)
...
+                        logger.info("Hederanode#{} is CHECKING", nodeId);
+                        if (wasActive) shutdownGrpcServer();
                     }
-                    case CHECKING -> logger.info("Hederanode#{} is CHECKING", nodeId);
                     case OBSERVING -> logger.info("Hederanode#{} is OBSERVING", nodeId);
-                    case FREEZING -> logger.info("Hederanode#{} is FREEZING", nodeId);
+                    case FREEZING -> {
+                        logger.info("Hederanode#{} is FREEZING", nodeId);
+                        if (wasActive) shutdownGrpcServer();
+                    }
```
rbair23, line 472: Basic cleanup -- On each transaction from/to active, we have to shutdown/startup the grpc server.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -92,6 +94,8 @@ public static void main(final String... args) throws Exception {
         final var registry = ConstructableRegistry.getInstance();
         new SwirldsPlatformBuilder()
                 .withName("Hedera")
+                .withConfigSource(SystemEnvironmentConfigSource.getInstance())
+                .withConfigSource(SystemPropertiesConfigSource.getInstance())
```
rbair23, line 98: Allow system properties and environment variables to override other properties. Needed so various config properties can be specified in the hapi test code and take effect on the test VMs.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/NettyGrpcServerManager.java
```diff
@@ -170,17 +170,19 @@ public synchronized void start() {
...
         }
 
-        if (tlsServer != null) {
+        if (tlsServer != null && !tlsServer.isTerminated()) {
             logger.info("Shutting down TLS gRPC server on port {}", tlsServer.getPort());
             terminateServer(tlsServer);
-            tlsServer = null;
+        } else {
+            logger.info("Cannot shut down an already stopped gRPC server");
         }
```
rbair23, line 186: Basic cleanup

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -235,6 +248,37 @@ public Type getType() {
...
+
+        @Override
+        public void after(HapiTestEngineExecutionContext context) throws Exception {
+            // If we are isolated, then stop and destroy everything to prepare for a new genesis startup.
+            final var env = context.getEnv();
+            if (env != null && env.started() && isolated) {
+                env.terminate();
+                env.getNodes().forEach(HapiTestNode::clearState);
+            }
+        }
```
rbair23, line 281: This is the essence of the change. All other changes just support this.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
rbair23, line 1: As of this change, this class is not currently used. I wish it could be. The problem is we cannot effectively do classloader isolation and we cannot effectively shut it down. I have some ideas on how we may be able to make classloader isolation work, but will need to be part of a subsequent PR, if at all possible.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -105,28 +184,27 @@ public void start() {
...
+                    connection.connect();
+                    return;
+                } catch (MalformedURLException e) {
+                    throw new RuntimeException("Should never happen", e);
+                } catch (IOException ignored) {
+                    // This is expected, the node is not up yet.
                 }
-            } catch (Exception e) {
-                throw new RuntimeException("node " + nodeId + ": Unable to read from the hgcaa log file " + log, e);
             }
```
rbair23, line 209: Use the prometheus endpoint instead of scraping logs to make this more reliable.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -188,4 +327,34 @@ private String getClasspath() {
...
+                    // This line always comes AFTER the # HELP line.
+                    statusKey = matcher.group(1);
+                    break;
+                }
+            }
+        } catch (IOException | InterruptedException ignored) {
+        }
+
+        return statusMap.get(statusKey);
+    }
```
rbair23, line 359: Parses the prometheus endpoint results for the platform status information.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
jsync-swirlds: Interesting.  This is a significant change in how the platform can be configured (which is currently configured _exclusively_ via two text files).

It might be wise to mention this to the platform team, devops, and release engineering so it is not a surprise to anyone.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -111,7 +112,6 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
     private static void addChildToEngineDescriptor(
             Class<?> javaClass, EngineDiscoveryRequest discoveryRequest, EngineDescriptor engineDescriptor) {
         if (IS_HAPI_TEST_SUITE.test(javaClass)) {
-            discoveryRequest.getConfigurationParameters().keySet().forEach(System.out::println);
```
jsync-swirlds, line 114: As verbose as it may be, this was pretty useful at times...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -65,40 +68,72 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster) {
...
     public void start() {
+        started = true;
         for (final var node : nodes) {
             node.start();
         }
 
         for (final var node : nodes) {
-            node.waitForActive(60);
+            try {
+                node.waitForActive(60);
```
jsync-swirlds, line 99: This needs to be changed to a larger value (as done in a previously merged PR) constant or keyed off of the `secondsWaitingServerUp` property you're setting elsewhere in this PR.
Otherwise the servers simply do not have enough time to reach an ACTIVE state in the CI environment.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -65,40 +68,72 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster) {
...
+            } catch (TimeoutException e) {
+                throw new RuntimeException(e);
```
jsync-swirlds, line 101: Why not just declare the TimeoutException?  Everything that calls this either catches or declares `Exception`, so it won't create any cascade code changes...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -74,12 +123,24 @@ public void start() {
...
 
     @Override
-    public void waitForActive(long seconds) {
+    public void waitForActive(long seconds) throws TimeoutException {
         final var waitUntil = System.currentTimeMillis() + (seconds * 1000);
-        while (th == null || th.hedera == null || !th.hedera.isActive()) {
-            if (System.currentTimeMillis() > waitUntil) {
-                throw new RuntimeException(
-                        "node " + nodeId + ": Waited " + seconds + " seconds, but node did not become active!");
+        while (System.currentTimeMillis() < waitUntil) {
```
jsync-swirlds, line 128: Might be good to put in a sleep of `(seconds/10)` seconds before this, just to give the node some chance to come up before querying it.
A further sleep on each loop might smooth this out, particularly in the more constrained CI environments.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -90,21 +151,123 @@ public void waitForActive(long seconds) {
...
-    public void stop() {
-        if (th != null) {
+    public void shutdown() {
+        if (th != null && (th.hedera.isFrozen() || th.hedera.isActive())) {
             if (th.hedera != null) th.hedera.shutdown();
             th.interrupt();
-            th = null;
+
+            // This is a hack, but it's the best I can do without classloader isolation and without a systematic
+            // way to shut down a node. Normally, nodes are shutdown by existing the JVM. However, we don't want to
```
jsync-swirlds, line 167: `existing` ==> `exiting`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -90,21 +151,123 @@ public void waitForActive(long seconds) {
...
+            // do that here, because we're in-process! So what I'm going to do is:
+            //  1. Search through all the threads in the JVM to find all of them with a callstack with "WorkerThread"
+            //     in the stack. This means they were called by the WorkerThread. Only objects that are part of the
+            //     node will be in this callstack.
+            //  2. For each such thread, stop it. Thread.stop is deprecated and discouraged, because it is almost
```
jsync-swirlds, line 172: Note, we **should** be able to call `Thread.interrupt()` on the thread and get a clean shutdown.  If we do not, then the platform code is incorrect and we need to fix the bugs preventing a proper shutdown in response to interrupt.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -86,9 +158,16 @@ public void start() {
...
                     .command(
                             javaCmd,
+                            // You can attach to any node. Node 0 at 5005, node 1 at 5006, etc. But if you need the
+                            // node to stop at startup, you can change the below line so nodeId == the node you want
+                            // to suspend at startup and the first "n" to "y".
+                            "-agentlib:jdwp=transport=dt_socket,server=y,suspend=" + (nodeId == 0 ? "n" : "n")
+                                    + ",address=*:" + (5005 + nodeId),
                             "-classpath",
                             classPath,
+                            "-Dfile.encoding=utf-8",
```
jsync-swirlds, line 168: This is basically ignored, because the node checks *native* encoding, but it cannot hurt.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/lifecycle/LifecycleOp.java
```diff
@@ -0,0 +1,63 @@
...
+    protected final boolean submitOp(HapiSpec spec) throws Throwable {
+        final var nodes = spec.getNodes().stream().filter(selector).collect(Collectors.toList());
+
+        if (nodes.isEmpty()) {
+            log.warn("Unable to find any node with criteria {}", selector);
+            return false;
+        }
+
+        // Run the op for each node, and if ANY node returns "true", then the test is over and we should also
+        // return true. If they all return false, then we also return false.
```
jsync-swirlds, line 53: Should this be majority, rather than any 1?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/lifecycle/LifecycleOp.java
```diff
@@ -0,0 +1,63 @@
...
+        return nodes.stream().anyMatch(this::run);
+    }
+
+    /**
+     * Executes the operation on the given node.
+     * @param node The node to operate on.
+     * @return true if the test is over and should be terminated, false if the test should continue.
```
jsync-swirlds, line 60: True is fail and false is continue.
That is going to produce some confusion in the future...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -33,8 +35,9 @@ public class HapiTestEnv {
     private static final int CAPTIVE_NODE_STARTUP_TIME_LIMIT = 300;
     private final List<HapiTestNode> nodes = new ArrayList<>();
     private final List<String> nodeHosts = new ArrayList<>();
+    private boolean started = false;
 
-    public HapiTestEnv(@NonNull final String testName, final boolean cluster) {
+    public HapiTestEnv(@NonNull final String testName, final boolean cluster, final boolean useInProcessAlice) {
```
vtronkov, line 40: What will be the process/idea of using the 3rd argument(useInProcessAlice)? Currently, it's always **false** so we are always using SubProcessHapiTestNode. Is the idea to manually set it to **true** when we debug and use it only for that?

vtronkov: Oh, I saw your comment in InProcessHapiTestNode

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -156,16 +157,18 @@ public HapiTestEngineExecutionContext before(HapiTestEngineExecutionContext cont
                 return context;
             }
 
-            env = new HapiTestEnv("HAPI Tests", true);
-            env.start();
+            // For now, switching to non-in process servers, because in process doesn't work for the
+            // restart and reconnect testing.
+            env = new HapiTestEnv("HAPI Tests", true, false);
```
vtronkov, line 162: Why the second argument is always **true**? I also saw that you removed the cluster variable from the HapiTestSuite annotation. How do we plan to control that?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -35,27 +52,82 @@
  * The {@code stdout} and {@code stderr} files will be written into the working directory.
  */
 final class SubProcessHapiTestNode implements HapiTestNode {
+    private static final Pattern PROM_PLATFORM_STATUS_HELP_PATTERN =
+            Pattern.compile("# HELP platform_PlatformStatus (.*)");
+    private static final Pattern PROM_PLATFORM_STATUS_PATTERN =
+            Pattern.compile("platform_PlatformStatus\\{.*\\} (\\d+)\\.\\d+");
     private static final String[] EMPTY_STRING_ARRAY = new String[0];
     /** The Hedera instance we are testing */
     private ProcessHandle handle;
```
vtronkov, line 61: Where do we assign this variable? I can only see the assignment in the `start` method(`handle = builder.start().toHandle())` but before that we have the following check `if (handle != null) throw new IllegalStateException("Node is not stopped, cannot start it!");` which means that it will never go there.

vtronkov (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
rbair23: Hmmm. We weren't actually making use of the "cluster" flag anywhere. But we could. We could default to false and then in special cases (where we expect to take a node down) set `cluster` to true. But I thought this was maybe less useful, for two reasons. First, there is value in running a cluster all the time, because it helps to find ISS issues. Second, if one node happens to die for some random non-code reason, the network keeps running.

On the other hand, if a node dies for some code reason, then we won't catch it.

It would be almost trivial to support the flag and have "cluster" default to false.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/lifecycle/LifecycleOp.java
rbair23: Yes, this is defined by the BDD system itself. And maybe it doesn't mean fail or pass but "end the test" or something else (but the only reason to end the test explicitly is if there was a failure?). I just documented my understanding.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/lifecycle/LifecycleOp.java
rbair23: I'm not sure TBH.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
rbair23: Nice! I'll just let TimeoutException get thrown.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
rbair23: I agree, that would be preferable. Really I want to add some API to platform to "shutdown" and let it handle all the details.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
rbair23: It is set to null in terminate

kimbor (dismissed):

jeromy-cannon (dismissed):

qnswirlds (dismissed):

jsync-swirlds:
The HapiTest run on this PR spent 6 hours trying to connect to a node for each individual test.  It appears the nodes all failed at some point, and never restarted.  Do we want to fix that in this PR, or a separate one?  It appears we might need some level of watchdog on the started nodes to restart any that exit (and avoid 6-hour-long failing runs)...

iwsimon (dismissed):
LGTM

jsync-swirlds (dismissed):

netopyr (dismissed):
LGTM - just some minor comments that can be handled later

jeromy-cannon (dismissed):

timo0 (dismissed):

iwsimon (dismissed):

Neeharika-Sompalli (dismissed):
LGTM !

jsync-swirlds:
Any chance of getting this moving again soon?  Having the entire test run go into a fatal death-spiral (because nothing restarts a failed node) for one error (which happens quite frequently when run locally) is extremely frustrating.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -90,21 +151,123 @@ public void waitForActive(long seconds) {
...
+    @Override
+    public void waitForShutdown(long seconds) throws TimeoutException {
+        final var waitUntil = System.currentTimeMillis() + (seconds * 1000);
+        while (th != null && th.hedera != null && th.hedera.isActive()) {
+            if (System.currentTimeMillis() > waitUntil) {
+                throw new TimeoutException(
+                        "node " + nodeId + ": Waited " + seconds + " seconds, but node did not shut down!");
+            }
+
+            try {
```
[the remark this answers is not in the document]
jsync-swirlds: That would most likely be a condition variable.  It's more complicated than a loop, however, and might not be worth that complexity in test code.

An alternative would be for `Hedera.shutdown` to return a CompletableFuture and wait for that to complete (with timeout).  That might be a cleaner approach, as it could cleanly compose the different waits behind the server shutdown into a single item.

vtronkov:
Note: I will create a follow-up story addressing Michael's comments. I think it's better to merge this PR to unblock suites and after that polish the changes.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -163,11 +163,13 @@ public void start() {
...
+                            "-Dfile.encoding=UTF-8",
                             // You can attach to any node. Node 0 at 5005, node 1 at 5006, etc. But if you need the
                             // node to stop at startup, you can change the below line so nodeId == the node you want
                             // to suspend at startup and the first "n" to "y".
                             "-agentlib:jdwp=transport=dt_socket,server=y,suspend=" + (nodeId == 0 ? "n" : "n")
                                     + ",address=*:" + (5005 + nodeId),
+                            "-Dhedera.workflows.enabled=true",
                             "-classpath",
                             classPath,
                             "-Dfile.encoding=utf-8",
```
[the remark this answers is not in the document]
vtronkov: Good catch. Fixed 👍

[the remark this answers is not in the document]
jsync-swirlds: Uhhh....  That removed the wrong line.
We do still need the workflows enabled flag; just not two instances of file.encoding.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -92,11 +164,16 @@ public void start() {
...
+                            // to suspend at startup and the first "n" to "y".
                             "-agentlib:jdwp=transport=dt_socket,server=y,suspend=" + (nodeId == 0 ? "n" : "n")
                                     + ",address=*:" + (5005 + nodeId),
-                            "-Dhedera.workflows.enabled=true",
                             "-classpath",
                             classPath,
+                            "-Dfile.encoding=utf-8",
+                            "-Dhedera.workflows.enabled=true",
+                            "-Dprometheus.endpointPortNumber=" + (10000 + nodeId),
                             "com.hedera.node.app.ServicesMain",
```
jsync-swirlds, line 177: Suggested replacement for lines 166-177 (github doesn't allow change suggestions if any line in the section is deleted)...
```java
                            // You can attach to any node. Node 0 at 5005, node 1 at 5006, etc. But if you need the
                            // node to stop at startup, you can change the below line so nodeId == the node you want
                            // to suspend at startup and the first "n" to "y".
                            "-agentlib:jdwp=transport=dt_socket,server=y,suspend=" + (nodeId == 0 ? "n" : "n")
                                    + ",address=*:" + (5005 + nodeId),
                            "-classpath",
                            classPath,
                            "-Dfile.encoding=UTF-8",
                            "-Dhedera.workflows.enabled=true",
                            "-Dprometheus.endpointPortNumber=" + (10000 + nodeId),
                            "com.hedera.node.app.ServicesMain",

```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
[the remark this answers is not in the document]
vtronkov: That should do it :)

jsync-swirlds (dismissed):

jeromy-cannon (dismissed):

jeromy-cannon (approved):

vtronkov:
Follow up story to address open comments: https://github.com/hashgraph/hedera-services/issues/9766

netopyr (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -33,9 +35,10 @@ public class HapiTestEnv {
     private static final int CAPTIVE_NODE_STARTUP_TIME_LIMIT = 300;
     private final List<HapiTestNode> nodes = new ArrayList<>();
     private final List<String> nodeHosts = new ArrayList<>();
+    private boolean started = false;
     public static final int CLUSTER_SIZE = 4;
 
-    public HapiTestEnv(@NonNull final String testName, final boolean cluster) {
+    public HapiTestEnv(@NonNull final String testName, final boolean cluster, final boolean useInProcessAlice) {
```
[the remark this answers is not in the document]
vtronkov: Added to https://github.com/hashgraph/hedera-services/issues/9766 👍

Neeharika-Sompalli (approved):
LGTM !

PR #8740, opened 2023-09-19
Improved stale event metrics.

### LINKED ISSUE #8739: Create metrics for stale transactions
We currently have metrics for tracking stale events, but we do not have metrics for tracking the number of transactions in the events that become stale. That could be useful information.

cody-littley:
Closes #8739


alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/EventIntakeMetrics.java
```diff
@@ -68,44 +69,68 @@ public class EventIntakeMetrics implements StaleEventObserver {
...
+                    INTERNAL_CATEGORY, "staleEventsPerSecond")
+            .withDescription("number of stale events per second");
     private final SpeedometerMetric staleEventsPerSecond;
 
+    private static final SpeedometerMetric.Config STALE_SELF_EVENTS_PER_SECOND_CONFIG = new SpeedometerMetric.Config(
+                    INTERNAL_CATEGORY, "staleSelfEventsPerSecond")
+            .withDescription("number of stale self events per second");
+    private final SpeedometerMetric staleSelfEventsPerSecond;
+
+    private static final SpeedometerMetric.Config STALE_APP_TRANSACTIONS_PER_SECOND_CONFIG =
```
poulok, line 82: I think these metrics will be better as counters. It is simple to render them as changes (spikes) in Grafana instead of just an ever-increasing maximum. I suggest this because I think the counters are more performant that the speedometer metric, and we plan on eliminating it at some point.

timo0 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/LongGauge.java
```diff
@@ -89,17 +105,14 @@ final class Config extends MetricConfig<LongGauge, LongGauge.Config> {
 
         /**
          * Constructor of {@code LongGauge.Config}
+         * <p>
```
alittley, line 108: duplicate paragraph tag

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/EventIntakeMetrics.java
```diff
@@ -69,35 +70,35 @@ public class EventIntakeMetrics implements StaleEventObserver {
...
+    private static final LongGauge.Config STALE_EVENTS_PER_SECOND_CONFIG = new LongGauge.Config(
+                    INTERNAL_CATEGORY, "staleEvents")
+            .withDescription("number of stale events since the node booted up");
+    private final LongGauge staleEventCount;
 
-    private static final SpeedometerMetric.Config STALE_SELF_EVENTS_PER_SECOND_CONFIG = new SpeedometerMetric.Config(
-                    INTERNAL_CATEGORY, "staleSelfEventsPerSecond")
-            .withDescription("number of stale self events per second");
-    private final SpeedometerMetric staleSelfEventsPerSecond;
+    private static final LongGauge.Config STALE_SELF_EVENTS_PER_SECOND_CONFIG = new LongGauge.Config(
```
alittley, line 78: Your config variable names still say `PER_SECOND`, but these aren't per second anymore

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/EventIntakeMetrics.java
lpetrovic05: To me it seems that these metrics are per second counters as opposed to "since forever"

cody-littley: > Your config variable names still say PER_SECOND, but these aren't per second anymore

Doh! Fixed.

> To me it seems that these metrics are per second counters as opposed to "since forever"

Kelly actually requested I switch these away from being per second. She said that we are planning on deprecating the speedometer metric because it's inefficient, and that we can configure grafana to display a "since forever" metric as just diffs, making it look pretty similar to what a speedometer would show. This makes sense to me... in general I think it's a good idea to do any data processing after the fact and not directly on the node.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/LongGauge.java
cody-littley: this class is now deleted

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/EventIntakeMetrics.java
cody-littley: Ok, after further discussion, these are now accumulators. This means they are cleared out every time they are read, so it will be easy to interpret.

alittley (approved):

edward-swirldslabs (approved):

PR #8742, opened 2023-09-19
Resolve `TokenUpdateSpecs` errors

georgi-l95:
**Description**:
This PR fixes all remaining tests in the TokenUpdateSpec, but two, as the last two require more time for debugging.
Changes include:
 - Correct response code, according to the tests for renewal period
 - Changes in the order of check, to be compliant with the tests in the spec
 - Changes to token create expiry setting
 - Add treasury as owner for specific nft, if owner is null

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


netopyr (commented):
In general, this PR looks good. But it changes validation methods that are used in several places. Did you double-check that we have no regressions?

georgi-l95:
@netopyr Actually fixes couple of tests in the TokenUpdateSpec, but breaks only 1 from TopicUpdateSpec (`validateMultipleFields()`) . So I'm not sure which is the correct logic, they both require different ResponseCodes for the same operation. Would like someone to confirm what is the correct logic.
Changes here fix more than 4-5 tests, but break 1.

netopyr:
I wouldn't be surprised if both behaviors are correct. 🙂 The returned error codes are not always consistent in our API.

As a quick fix, we can add a second parameter to `validateAutoRenewPeriod()`, which allows us to specify the error code.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -240,9 +240,7 @@ private ExpiryMeta getExpiryMeta(
             @NonNull final HandleContext context) {
         final var maxEntityLifetime =
                 context.configuration().getConfigData(EntitiesConfig.class).maxLifetime();
-        final var impliedExpiry = op.hasAutoRenewPeriod()
-                ? consensusTime + op.autoRenewPeriod().seconds()
-                : consensusTime + maxEntityLifetime;
+        final var impliedExpiry = op.hasExpiry() ? op.expiry().seconds() : consensusTime + maxEntityLifetime;
```
Neeharika-Sompalli, line 243: Awesome. 👍

Neeharika-Sompalli (commented):
As @netopyr mentioned, both are correct. Since we have to keep mono-service behavior for differential testing, we can make `TopicUpdateHandler`  return `AUTORENEW_DURATION_NOT_IN_RANGE` when it fails. and add a TODO comment to change in future.

Neeharika-Sompalli (commented):
Need to merge back develop and see none of HAPI tests are failing.

netopyr:
Unfortunately, there is a conflict now.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImplTest.java
```diff
@@ -117,7 +117,7 @@ void rejectsBriefAutoRenewPeriod() {
 
         assertThatThrownBy(() -> subject.validateAutoRenewPeriod(55L))
                 .isInstanceOf(HandleException.class)
-                .has(responseCode(AUTORENEW_DURATION_NOT_IN_RANGE));
+                .has(responseCode(INVALID_RENEWAL_PERIOD));
```
mhess-swl, line 120: Why are these tests being changed?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -228,7 +230,17 @@ private ExpiryMeta resolvedUpdateMetaFrom(
...
         if (updatesExpiryMeta(op)) {
             final var updateMeta = new ExpiryMeta(effExpiryOf(op), effAutoRenewPeriodOf(op), op.autoRenewAccount());
-            return expiryValidator.resolveUpdateAttempt(currentMeta, updateMeta);
+            try {
+                return expiryValidator.resolveUpdateAttempt(currentMeta, updateMeta);
+            } catch (final HandleException e) {
+                if (e.getStatus() == INVALID_RENEWAL_PERIOD) {
+                    // Tokens throw INVALID_EXPIRATION_TIME, but for topic it's expected currently to throw
+                    // AUTORENEW_DURATION_NOT_IN_RANGE
+                    // TODO: We need to change this in the future.
```
mhess-swl, line 239: `future('<github issue number>')` is preferable to `TODO`

georgi-l95: done.

mhess-swl (dismissed):
LGTM!

netopyr (dismissed):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGetNftInfoHandler.java
```diff
@@ -123,19 +125,22 @@ public Response findResponse(@NonNull final QueryContext context, @NonNull final
     private Optional<TokenNftInfo> infoForNft(
             @NonNull final NftID nftId,
             @NonNull final ReadableNftStore readableNftStore,
+            @NonNull final ReadableTokenStore readableTokenStore,
```
rbair23, line 128: Can you also update the javadoc comment with the missing param?

rbair23 (dismissed):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGetNftInfoHandler.java
georgi-l95: Ah yes, forgot about javadoc, thank you. Updated.

Neeharika-Sompalli (approved):
LGTM !

PR #8747, opened 2023-09-19
Create intake phase timer

### LINKED ISSUE #8129: Replace intake cycle metrics with fractional timer

alittley:
- closes #8129


cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/EventIntake.java
```diff
@@ -147,14 +151,17 @@ public EventIntake(
      * @param event the event
      */
     public void addUnlinkedEvent(final GossipEvent event) {
```
edward-swirldslabs, line 153: Is all usage of the phase timer on the same thread?  The same timer is used in 3 different methods.  `addUnlinkedEvent`, `addEvent` and `EventValidator.validateEvent()`.    These methods are sent through SwirldsPlatform into a variety of places and I am not confident they are all on the same thread.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/EventIntake.java
```diff
@@ -198,31 +206,34 @@ public void addEvent(final EventImpl event) {
...
         }
+
         if (consensus().getMinGenerationNonAncient() > minGenNonAncientBeforeAdding) {
             // consensus rounds can be null and the minNonAncient might change, this is probably because of a round
             // with no consensus events, so we check the diff in generations to look for stale events
+            phaseTimer.activatePhase(EventIntakePhase.HANDLING_STALE_EVENTS);
             handleStale(minGenNonAncientBeforeAdding);
-            stats.dispatchedStale();
         }
-        stats.doneIntakeAddEvent();
```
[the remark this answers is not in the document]
alittley: I was relying on the set to `IDLE` in `addUnlinkedEvent`, but you're correct and that isn't sufficient, since self events don't follow that pathway

I'll add a set to `IDLE` here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/EventIntake.java
alittley: Yes, all usages are on the intake thread.

alittley: The access from different threads occurs because PCES replay uses the same code pathways as the standard event intake thread. I talked this over with @cody-littley, and his opinion is that even if these metrics were to be skewed during PCES replay, it wouldn't be a big deal.

Also worth noting is that even though PCES replay and intake are on separate threads, they should not be operating simultaneously, since a node replaying events won't be gossiping.

edward-swirldslabs (approved):

cody-littley (approved):

PR #8749, opened 2023-09-19
Fix genesis records creation order

mhess-swl:
The genesis account records were being created in different orders, causing an ISS. Ordering by account ID fixes the issue.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
@@ -33,6 +33,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.time.Instant;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import javax.inject.Singleton;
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
```diff
@@ -132,16 +133,18 @@ private void createAccountRecordBuilders(
             @NonNull final TokenContext context,
             @Nullable final String recordMemo,
             @Nullable final Long overrideAutoRenewPeriod) {
-        for (Map.Entry<Account, CryptoCreateTransactionBody.Builder> entry : map.entrySet()) {
+        final var orderedAccts = map.keySet().stream()
+                .sorted(Comparator.comparingLong(acct -> acct.accountId().accountNum()))
```
jsync-swirlds, line 137: This will only work as long as we have no shards or realms (it will be unspecified ordering for same-number accounts once we have multiple shards and/or realms).
OK for now, but making most of the protobuf ID classes comparable is probably a good idea long-term so we don't need adhoc comparators like this.

jsync-swirlds (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
mhess-swl: Good call. I don't see any `Comparable` in the `AccountID` class, any idea where it is?

iwsimon (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
jsync-swirlds: Not there yet.  We need to work with Lubo to have a comment directive added to the PBJ compiler so it can generate the compareTo method and add the Comparable interface on selected classes; then add that directive to ID classes and a few others.
That is not going to happen tomorrow, so this approach will have to do for now.

kimbor (approved):

Neeharika-Sompalli (approved):

Issue #8469, opened 2023-09-08
New VirtualMap metric: virtual node cache size, in Mb


artemananiev:
A new metric would be very helpful, current virtual node cache size in Mb. There is already a number of metrics related to flushes, but they are only useful after data is flushed to disk. What we have been observing recently is there are no flushes, which means data is accumulated in the node cache up until its size reaches the configured threshold. However, there is no way to see what is the current (estimated) cache size, there should be a new metric for this purpose.

PR #8755, opened 2023-09-20
8469: New VirtualMap metric: virtual node cache size, in Mb

artemananiev:
Fix summary:

* A new VirtualMap metric is added, `nodeCacheSizeMb`
* The metric is updated on every virtual pipeline thread iteration (e.g. after flush, merge, etc.)
* A new unit test is provided
* A few related issues are fixed, see my comments in the files

Fixes: https://github.com/hashgraph/hedera-services/issues/8469

### DIFF
--- a/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
+++ b/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
@@ -473,7 +473,6 @@ public class VirtualPipeline {
         for (PipelineListNode<VirtualRoot> node = copies.getFirst(); node != null; node = node.getNext()) {
             final VirtualRoot copy = node.getValue();
             if (!copy.isImmutable()) {
-                assert node.getNext() == null;
                 break;
             }
             final long estimatedSize = copy.estimatedSize();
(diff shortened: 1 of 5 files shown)


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
```diff
@@ -473,7 +473,6 @@ private long currentTotalSize() {
         for (PipelineListNode<VirtualRoot> node = copies.getFirst(); node != null; node = node.getNext()) {
             final VirtualRoot copy = node.getValue();
             if (!copy.isImmutable()) {
-                assert node.getNext() == null;
```
artemananiev, line 476: This is a thread race. Right after `copy.isImmutable()` check, the copy may become mutable, and a new copy may become its "next" here. I noticed this issue when one of VirtualPipelineTests cases failed a couple times in a row.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
```diff
@@ -562,8 +561,10 @@ private void hashFlushMerge() {
                 logger.debug(VIRTUAL_MERKLE_STATS.getMarker(), "Merge {}", copy.getFastCopyVersion());
                 merge(next);
                 copies.remove(next);
-                statistics.setPipelineSize(copies.getSize());
```
artemananiev, line 565: This code is moved two lines below, outside of the block, so it's now executed both for "flush" and "merge" cases

jsync-swirlds (approved):

PR #8761, opened 2023-09-20
Fix fee calculation for TokenWipe and TokenBurn

### LINKED ISSUE #8760: Tests failing with `feeData` null error
### Problem

When trying to execute tests which include `TokenBurn` or `TokenWipe` operations, they all fail with `feeData` null exception

### Solution

Fix errors by adding logic for fee calculation to check for tokenType

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds logic to the fee calculation for TokenWipe and TokenBurn to check for token type, which will be then used for  selection from the feeDataMap

**Related issue(s)**:

Fixes #8760 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAccountWipeHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAccountWipeHandler.java
@@ -194,9 +194,15 @@ public final class TokenAccountWipeHandler implements TransactionHandler {
     @Override
     public Fees calculateFees(@NonNull final FeeContext feeContext) {
         final var op = feeContext.body();
+        final var readableTokenStore = feeContext.readableStore(ReadableTokenStore.class);
+        final var tokenType =
+                readableTokenStore.get(op.tokenWipeOrThrow().tokenOrThrow()).tokenType();
         final var meta = TOKEN_OPS_USAGE_UTILS.tokenWipeUsageFrom(fromPbj(op));
         return feeContext
-                .feeCalculator(SubType.DEFAULT)
+                .feeCalculator(
+                        tokenType.equals(TokenType.FUNGIBLE_COMMON)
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAccountWipeHandler.java
```diff
@@ -194,9 +194,15 @@ public void handle(@NonNull final HandleContext context) throws HandleException
     @Override
     public Fees calculateFees(@NonNull final FeeContext feeContext) {
         final var op = feeContext.body();
+        final var readableTokenStore = feeContext.readableStore(ReadableTokenStore.class);
+        final var tokenType =
+                readableTokenStore.get(op.tokenWipeOrThrow().tokenOrThrow()).tokenType();
         final var meta = TOKEN_OPS_USAGE_UTILS.tokenWipeUsageFrom(fromPbj(op));
         return feeContext
-                .feeCalculator(SubType.DEFAULT)
+                .feeCalculator(
```
netopyr, line 202: Do we have to take into account if the token has custom fees and eventually use `TOKEN_FUNGIBLE_COMMON_WITH_CUSTOM_FEES` and `TOKEN_NON_FUNGIBLE_UNIQUE_WITH_CUSTOM_FEES`?

georgi-l95: In the feeDataMap for both `TOKEN_BURN` and `TOKEN_ACCOUNT_WIPE` only subTypes available are `TOKEN_FUNGIBLE_COMMON` and `TOKEN_NON_FUNGIBLE_UNIQUE`. So judging by this I'd say we don't need to worry about custom fees. But won't hurt for someone else to confirm, just to be sure.

netopyr (approved):
LGTM

Neeharika-Sompalli (commented):
Approving post merge. Thanks @georgi-l95

PR #8802, opened 2023-09-20
Create Genesis Blocklist Accounts

### LINKED ISSUE #8711: Implement blocklist accounts in token service genesis migration

mhess-swl:
Closes #8711


hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
```diff
@@ -216,16 +217,27 @@ public void migrate(
...
             final var newStates = new FilteredWritableStates(writeableStates, remainingStates);
 
-            final var migrationContext =
-                    new MigrationContextImpl(previousStates, newStates, config, networkInfo, genesisRecordsBuilder);
+            // For any changes to state that depend on other services outside the current service, we need a reference
+            // to the overall state that we can pass into the context. This reference to overall state will be strictly
+            // controlled via the MigrationContext API so that only changes explicitly specified in the interface can be
+            // made (instead of allowing any arbitrary change to overall state). As above, we won't commit anything
+            // until after this service's migration
+            final var migrationStack = new SavepointStackImpl(hederaState);
```
netopyr, line 225: This looks suspicious. Changes in `writableStates` will not appear in `migrationStack` and vice versa. They are accessing different parts of the `HederaState`, but this is not enforced. Maybe we can find a solution that makes this more explicit.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
```diff
@@ -216,16 +217,27 @@ public void migrate(
...
             } else {
                 schema.migrate(migrationContext);
             }
+            // Now we commit all the service-specific changes made during this service's update or migration
             if (writeableStates instanceof MerkleHederaState.MerkleWritableStates mws) {
                 mws.commit();
             }
+            // And commit any changes that were made to any state belonging to other services
+            migrationStack.createSavepoint();
+            migrationStack.commit();
```
netopyr, line 240: This will have no effect. You do not need to create a save-point and you probably want to use `migrationStack.commitFullStack()`. `commit()` will only commit the top-most frame on the stack.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
```diff
@@ -33,13 +37,15 @@
  * @param newStates The new states, preloaded with any new state definitions.
  * @param configuration The configuration to use
  * @param genesisRecordsBuilder The instance responsible for genesis records
+ * @param savepointStack todo
```
netopyr, line 40: todo? 😄

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
```diff
@@ -48,5 +54,12 @@ public record MigrationContextImpl(
         requireNonNull(configuration);
         requireNonNull(networkInfo);
         requireNonNull(genesisRecordsBuilder);
+        requireNonNull(savepointStack);
+    }
+
+    @Override
+    public long newEntityNum() {
+        final var entityIdsFactory = new WritableStoreFactory(savepointStack, EntityIdService.NAME);
+        return entityIdsFactory.getStore(WritableEntityIdStore.class).incrementAndGet();
```
netopyr, line 63: This will create a new factory whenever somebody calls `newEntityNum()`. What do you think about creating the factory in the constructor?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
```diff
@@ -0,0 +1,145 @@
...
+        final List<String> fileLines = readFileLines(blocklistResourceName);
+        if (fileLines.isEmpty()) return Collections.emptyList();
+
+        return parseBlockList(fileLines);
+    }
+
+    private List<String> readFileLines(@NonNull final String blocklistResourceName) {
+        final List<String> fileLines;
+        try {
+            fileLines = readPrivateKeyBlocklist(blocklistResourceName);
```
netopyr, line 56: You can do `return readPrivateKeyBlocklist(blocklistResourceName);`. Or do you prefer to have the `return` statement at the end?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
```diff
@@ -0,0 +1,145 @@
...
+        } catch (IllegalArgumentException iae) {
+            log.error("Failed to parse blocklist", iae);
+            return Collections.emptyList();
+        }
+        return blocklist;
+    }
+
+    @NonNull
+    private static List<String> readPrivateKeyBlocklist(@NonNull final String fileName) {
+        final var inputStream = BlocklistParser.class.getClassLoader().getResourceAsStream(fileName);
```
netopyr, line 82: You are not closing the streams. Maybe you can put everything in a try-with-resources statement.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
```diff
@@ -0,0 +1,145 @@
...
+        return fileLines;
+    }
+
+    private static List<BlockedInfo> parseBlockList(final List<String> fileLines) {
+        final List<BlockedInfo> blocklist;
+        try {
+            final var columnHeaderLine = fileLines.get(0); // Assume that the first line is the header
+            final var blocklistLines = fileLines.subList(1, fileLines.size());
+            final var columnCount = columnHeaderLine.split(",").length;
+            blocklist = blocklistLines.stream()
```
netopyr, line 70: It is really unfortunate that we turn the `Stream` into a `List` in `readPrivateKeyBlocklist`, and here we are turning it back into a `Stream`. I wonder if we can avoid this.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * The resulting blocked account info record contains the EVM address derived from the private key, and the memo.
+     *
+     * @param line line from the blocklist resource
+     * @param columnCount number of comma-separated values in a line
+     * @return blocked account info record
+     */
+    @NonNull
+    private static BlockedInfo parseCSVLine(final @NonNull String line, final int columnCount) {
+        final var parts = line.split(",", -1);
```
netopyr, line 104: Did you copy this code from somewhere? I am too tired to review it thoroughly right now. 🙂

netopyr (commented):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -219,6 +219,7 @@ public Hedera(@NonNull final ConstructableRegistry constructableRegistry) {
                         new UtilServiceImpl(),
                         new RecordCacheService(),
                         new BlockRecordService(),
+                        // todo there should be a problem with this being after token service...???
```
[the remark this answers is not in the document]
mhess-swl: Old note to myself that I forgot to erase 😊

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
mhess-swl: This is precisely what I was worried about, using the savepoint stack incorrectly. Let's discuss when we're both available

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
mhess-swl: Whoops 🤦

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
mhess-swl: I'm fine with that. I got this code from the handle context IIRC, so I wasn't sure about the nuances

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: Yes, everything in `BlocklistParser` was copied. I can spend the time to improve it now if needed, but I thought the time would be better spent elsewhere.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
mhess-swl: Instead of using the savepoint stack, we decided to use a writable store directly to generate entity IDs. It's not pretty, but it will work for now.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
mhess-swl: This is no longer relevant. We aren't going to use a `WritableStoreFactory`, instead using an entity ID writable store directly, which will be passed in.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: This was copied over from the mono service code verbatim

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/MigrationContextImpl.java
mhess-swl: Fixed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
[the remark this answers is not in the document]
mhess-swl: Deleted

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -412,19 +416,52 @@ private void onMigrate(
...
+
             registry.migrate(
                     state,
                     previousVersion,
                     currentVersion,
                     configProvider.getConfiguration(),
                     networkInfo,
-                    backendThrottle);
+                    backendThrottle,
+                    requireNonNull(entityIdStore));
```
netopyr, line 459: I wonder if it is sufficient to pass `state` into `migrate()` and create the store there. It feels awkward to require `state` and a specific store here. The `registry` can create the `entityIdStore` only when needed.

I believe when you call a contructor, the result can never be `null`. Either it returns an instance or throws.

netopyr (changes requested):
Most of the comments are minor issues but closing the stream in `BlocklistParser` is mandatory.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: We could, but it's awkward. We have to differentiate the first line from the rest of the lines in order to parse the column count.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
mhess-swl: This is definitely possible, but it makes MerkleRegistrySchemaTest really ugly. The functionality inside of the `Hedera` class – where the EntityIdService is migrated first, followed by all the other services – has to be applied to  ~20 tests, which otherwise throw an exception:
```
java.lang.IllegalArgumentException: There are no singleton states
```
What would you prefer? Keeping the entity ID store where it is, or going the test route?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: Done

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: Done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -412,19 +416,52 @@ private void onMigrate(
...
+                state,
+                previousVersion,
+                currentVersion,
+                configProvider.getConfiguration(),
+                networkInfo,
+                backendThrottle,
+                // We call with null here because we're migrating the entity ID service itself
+                null);
+        // Now that the Entity ID Service is migrated, migrate the remaining services
+        for (final var registration : servicesRegistry.registrations().stream()
```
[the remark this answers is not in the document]
mhess-swl: Done

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
netopyr: You also have to close `inputStream`. Luckily, `try-with-resource` provides support for multiple resources.

netopyr (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/BlocklistParser.java
mhess-swl: Good catch. Done

netopyr (approved):

netopyr (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
mhess-swl: Keeping it as is for now..

nathanklick (approved):

PR #8812, opened 2023-09-21
Fix tests from ContractCallLocalSuite

MiroslavGatsanoga:
**Description**:

Fixes for some of the `ContractCallLocalSuite` failing tests.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8722

**Notes for reviewer**:

I did some debugging for the rest of the tests left to be fixed and I have created one additional issue to be addressed separately:
- `insufficientFeeFails` & `lowBalanceFails`: we need calculateFees() implementation for CryptoTransferHandler, this seems it would be handled as part of https://github.com/hashgraph/hedera-services/issues/8463
- `htsOwnershipCheckWorksWithAliasAddress`: https://github.com/hashgraph/hedera-services/issues/8813

**Checklist**


tinker-michaelj (dismissed):
LGTM, thanks @MiroslavGatsanoga !

If I understand, your fixes were:
  1. Add logic accounting for the possibility that an address refers to a token.
  2. To actually _set_ the `ContractCallLocal` status in the response header. 😁

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java
```diff
@@ -101,7 +101,7 @@ public ContractFunctionResult asProtoResultOf(
      */
     public ContractFunctionResult asQueryResultOf() {
         if (haltReason != null) {
-            throw new AssertionError("Not implemented");
+            return null;
```
Nana-EC, line 104: Q: why null vs a ContractFunctionResult with a revert error or details?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java
```diff
@@ -279,7 +279,18 @@ public long getIdNumber(@NonNull Address address) {
...
-        final var account = validatedAccount(number);
+        final var account = nativeOperations.getAccount(number);
+        if (account == null) {
+            final var token = nativeOperations.getToken(number);
+            if (token != null) {
+                // If the token is deleted or expired, the system contract executed by the redirect
+                // bytecode will fail with a more meaningful error message, so don't check that here
+                return asLongZeroAddress(number);
+            } else {
+                throw new IllegalArgumentException("No account has number " + number);
```
Nana-EC, line 290: nit:
```suggestion
                throw new IllegalArgumentException("No account or token has number " + number);
```

Nana-EC (commented):
Looks good.
1 question and 1 nit

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java
MiroslavGatsanoga: Fixed.

Nana-EC (approved):

netopyr (approved):
LGTM in regards to hedera-services

PR #8834, opened 2023-09-22
Check for deleted accounts in crypto allowances

mhess-swl:
Part of #8465


iwsimon (dismissed):
LGTM

iwsimon (dismissed):

iwsimon (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoDeleteAllowanceSuite.java
```diff
@@ -504,28 +509,30 @@ private HapiSpec exceedsTransactionLimit() {
...
                 .when(
                         cryptoApproveAllowance()
                                 .payingWith(owner)
                                 .addNftAllowance(owner, nft, spender, false, List.of(1L, 2L)),
                         cryptoDeleteAllowance()
                                 .payingWith(owner)
-                                .addNftDeleteAllowance(owner, nft, List.of(1L, 2L, 3L, 3L, 3L))
-                                .hasPrecheck(MAX_ALLOWANCES_EXCEEDED),
+                                .addNftDeleteAllowance(owner, nft, List.of(1L, 1L, 1L, 1L, 1L))
+                                .hasPrecheck(OK),
```
mhess-swl, line 525: Usually I wouldn't change the test conditions, but I think this change makes sense. I'm guessing this test wasn't passing before because the code was counting all the duplicate nft IDs (i.e. it counted ID 3 three separate times). The new implementation checks for duplicates and doesn't add them to the total, so I changed this expected precheck status to OK. In other words:
`.addNftDeleteAllowance(owner, nft, List.of(1L, 1L, 1L, 1L, 1L))` is the equivalent of `.addNftDeleteAllowance(owner, nft, List.of(1L))`

iwsimon: hm, I would ask @Neeharika-Sompalli to double check this.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkGetAccountDetailsHandler.java
```diff
@@ -98,6 +100,7 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
         final var accountStore = context.createStore(ReadableAccountStore.class);
         final var accountMetadata = accountStore.getAccountById(op.accountIdOrThrow());
```
Neeharika-Sompalli, line 101: nit : It is not `accountMetadata` . Instaed it is `account. It was metadata before changing all stores to return account object 🙂

mhess-swl: Whoops, you're right!

mhess-swl: Done

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoDeleteAllowanceHandler.java
```diff
@@ -61,9 +62,10 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         final var txn = context.body();
         pureChecks(txn);
         final var op = txn.cryptoDeleteAllowanceOrThrow();
-        // Every owner whose allowances are being removed should sign, if the owner is not payer
+        // Every owner whose allowances are being removed should sign (or the payer, if there is no owner)
+        final var payerId = txn.transactionID().accountID();
         for (final var allowance : op.nftAllowancesOrElse(emptyList())) {
-            context.requireKeyOrThrow(allowance.ownerOrElse(AccountID.DEFAULT), INVALID_ALLOWANCE_OWNER_ID);
+            context.requireKeyOrThrow(allowance.ownerOrElse(payerId), INVALID_ALLOWANCE_OWNER_ID);
```
Neeharika-Sompalli, line 68: Payer would have already signed in workflow right ? Why to sign again ? It should just be  

```
for (final var allowance : op.nftAllowancesOrElse(emptyList())) {
            if(allowance.hasOwner()){
                context.requireKeyOrThrow(allowance.ownerOrThrow(), INVALID_ALLOWANCE_OWNER_ID); 
            }
        }
``` ?

mhess-swl: The payer would have signed already, you're right

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoDeleteAllowanceSuite.java
Neeharika-Sompalli: If we are making it a set, then the implementation is wrong. We should comply with mono-service. We should not be changing this error code.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoDeleteAllowanceHandler.java
mhess-swl: Fixed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoDeleteAllowanceSuite.java
mhess-swl: This seems like a bug in mono service code, but I'll go ahead and recreate it the same way

Neeharika-Sompalli: I believe it is mentioned in HIP. Need to re-check

mhess-swl: Fixed

Neeharika-Sompalli (approved):

iwsimon (dismissed):

Neeharika-Sompalli (approved):

Neeharika-Sompalli (commented):
Seems one HAPI test is failing to be fixed

iwsimon (dismissed):

Neeharika-Sompalli (approved):
LGTM ! Thanks @mhess-swl

PR #8840, opened 2023-09-22
Fix metadata digest management during PCES replay with old files on disk

### LINKED ISSUE #8738: Metadata hash wrong on first block of PCES replay w/ old files on disk
### Problem

When doing PCES replay from a saved state with pre-restart record files still on disk, the metadata hash digest will be contaminated with prefix metadata from all record files before the first "net-new" block of the replay.

The root cause is that the metadata hash digest [is not reset](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/stream/RecordStreamFileWriter.java#L241) when skipping the `closeCurrentAndSign()` step for a block whose record file is already on disk.

### Solution

Instead of initializing the metadata hash digest with prefix metadata when we _start_ a new block, simply hash the entire metadata at the point were are closing and writing the block's record file.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #8738 
 - Moves the entire metadata hash computation into `closeCurrentAndSign()`, since all necessary data is available there; this eliminates the need for the `dosMeta` member.
 - For extra safety calls `MessageDigest.reset()` before writing digest contents for both stream file and its metadata.
 - _Note to reviewer:_ Unfortunately the `SAVED_STATE_RUNNING_HASH` field was set inconsistently with the other replay assets in the `RecordStreamRecoveryAndReplayTest`. This had no effect on the recovery tests in that file (since during recovery the node resets its initial hash to the start hash of the "golden" record file on disk that overlaps the recovery start time); but took some effort to notice when adding this replay test. 😵‍💫


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/stream/RecordStreamFileWriter.java
```diff
@@ -541,19 +504,41 @@ private HashObject toProto(final byte[] hash) {
...
         final var signatureFile =
                 SignatureFile.newBuilder().setFileSignature(fileSignature).setMetadataSignature(metadataSignature);
-
-        // create signature file
         final var sigFilePath = relatedRecordStreamFile + "_sig";
         try (final var fos = new FileOutputStream(sigFilePath)) {
             // version in signature files is 1 byte, compared to 4 in record files
             fos.write(streamType.getSigFileHeader()[0]);
-            signatureFile.build().writeTo(fos);
+            fos.write(serialize(signatureFile.build()));
```
mhess-swl, line 541: Why this change? Just curious

mhess-swl (dismissed):

mhess-swl:
Nice work, LGTM! 🥂

rbair23:
Do we have an issue to implement these tests on the modular service? Do we have PCES tests in test-clients, or only in unit tests?

tinker-michaelj:
> Do we have an issue to implement these tests on the modular service? Do we have PCES tests in test-clients, or only in unit tests?

Created an [issue](https://github.com/hashgraph/hedera-services/issues/8878) to port these tests. I think PCES replay always happens when restarting with a state + event files on disk, so reconnect tests will exercise it to some degree afaik.

I don't know if we have a test to simulate the network-wide kill switch that hit testnet that day.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/stream/RecordStreamFileWriter.java
```diff
@@ -249,18 +244,10 @@ public void closeCurrentAndSign() {
                     // write endRunningHash
                     final var endRunningHash = runningHash.getFutureHash().get();
                     recordStreamFileBuilder.setEndObjectRunningHash(toProto(endRunningHash.getValue()));
-                    dosMeta.write(endRunningHash.getValue());
                     LOG.debug(
```
Neeharika-Sompalli, line 247: Do we know if modular code has similar logic ? I guess a question for Jasper. But just curious

Neeharika-Sompalli (approved):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/stream/RecordStreamFileWriter.java
tinker-michaelj: @Neeharika-Sompalli I need to investigate this actually; almost certain this exact issue won't exist, but do need to validate the replay/recovery scenarios.

PR #8842, opened 2023-09-22
7905 Schedule Service Unit Tests - Part 2

jsync-swirlds:
* Changed handler unit tests to use MapReadbleKVStates and MapWritableKVStates instead of mocks. * This actually requires more code, but it's less brittle. * It also highlighted that the real KVStates cache far more aggressively and narcissistically than expected.
* Added whitelist unit test to ensure the schedule whitelist is enforced
   * Noticed, and fixed, a missing required payer key (for explicit payer) backstop.
* Removed use of deprecated methods on HandleContext
   * Also removed the deprecated methods, as no other code uses them
* Merged the ScheduleStoreTestBase and ScheduleHandlerTestBase to reduce duplication
   * Renamed ScheduleStoreTestBase to ScheduleTestBase, which is a superclass of ScheduleHandlerTestBase
* Major clean up of ScheduleServiceStateTranslatorTest, which was all over the map with test input and output
* Added Handler tests for Schedule Create Handler
* Added Handler tests for Schedule Delete Handler
* Added Handler tests for Schedule Sign Handler
* Added tests for pureChecks for all handlers

This will complete #7905.


tinker-michaelj (approved):
LGTM, thanks @jsync-swirlds !

PR #8860, opened 2023-09-25
fix: change contractCallLocalMethod to Query instead of transaction

vtronkov:
**Description**:
Fixing CallOperationSuite tests

**Related issue(s)**:
Changing the SmartContractServiceDefinition of contractCallLocalMethod to query instead of transaction. Same fix as in https://github.com/hashgraph/hedera-services/pull/8812

Fixes https://github.com/hashgraph/hedera-services/issues/8854


iwsimon (commented):
HapiTest failed

vtronkov:
> HapiTest failed

The tests are flaky. I ran them twice and every time a different error appeared. It seems like the third time is a charm.

Nana-EC (approved):

PR #8879, opened 2023-09-25
Retry failing queries during HAPI tests

netopyr:
This PR introduces a small change to retry failing queries. This does not fix all possible issues that make the tests unstable, but many of them.


tinker-michaelj (approved):
LGTM

Issue #8884, opened 2023-09-26
Implement methods in `ContractGetBytecodeHandler`


thenswan:
**The Context**
Issues [#8867](https://github.com/hashgraph/hedera-services/issues/8867), [#8868](https://github.com/hashgraph/hedera-services/issues/8868), and [#8869](https://github.com/hashgraph/hedera-services/issues/8869) are failing with the same error as [#8804](https://github.com/hashgraph/hedera-services/issues/8804) and [#8848](https://github.com/hashgraph/hedera-services/issues/8848) on `develop` branch.
If I try to run them using branch `08722-contract-call-local-suite-fixes` (PR [#8812](https://github.com/hashgraph/hedera-services/pull/8812)) — I get `java.lang.UnsupportedOperationException: Not implemented` in `ContractGetBytecodeHandler#validate`.

**The Problem**
`ContractGetBytecodeHandler#validate` does not contain validation logic and throws new `UnsupportedOperationException("Not implemented")`.
Also `ContractGetBytecodeHandler#findResponse` throws `UnsupportedOperationException("Not implemented")` as well.

PR #8905, opened 2023-09-27
Fix `ContractGetBytecodeSuite` tests

### LINKED ISSUE #8897: Fix ContractGetBytecodeSuite tests

thenswan:
**Description**:
Fix `ContractGetBytecodeSuite` tests and implement methods in `ContractGetBytecodeHandler`.

**Related issue(s)**:
Fixes #8897 #8884


tinker-michaelj (approved):
LGTM, ty @thenswan !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractGetBytecodeHandler.java
```diff
@@ -57,13 +68,36 @@ public Response createEmptyResponse(@NonNull final ResponseHeader header) {
...
+        validateFalsePreCheck(contractFrom(context) == null, INVALID_CONTRACT_ID);
     }
 
     @Override
     public Response findResponse(@NonNull final QueryContext context, @NonNull final ResponseHeader header) {
         requireNonNull(context);
         requireNonNull(header);
-        throw new UnsupportedOperationException("Not implemented");
+        final var contractGetBytecode = ContractGetBytecodeResponse.newBuilder().header(header);
+        if (header.nodeTransactionPrecheckCode() == OK && header.responseType() == ANSWER_ONLY) {
```
Nana-EC, line 79: nit: comments here would have been valuable.
Does this imply only for the answer only case are the actual bytes returned? In all other cases just the response header is sent?

Q: Where does this show up in practice?

Nana-EC (commented):
Looks good. Question for me

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractGetBytecodeHandler.java
thenswan: @Nana-EC I just followed the example of `ContractGetInfoHandler` (it's logic) in order to implement `ContractGetBytecodeHandler`, so `ContractGetBytecodeSuite` and a couple of other suites would pass.

tinker-michaelj: > Does this imply only for the answer only case are the actual bytes returned? In all other cases just the response header is sent?

This is correct @Nana-EC; although unfortunately the [`ResponseType` enum](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/query_header.proto#L39) includes an unsupported field `ANSWER_STATE_PROOF`, the "answer" to the query is returned **ONLY** when both,
  1. The requested response type is `ANSWER_ONLY`; and,
  2. The [`ResponseHeader#nodeTransactionPrecheckCode`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/response_header.proto#L41) is `OK`.

Nana-EC (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractGetBytecodeHandler.java
Nana-EC: @thenswan I'm fine with this. 
Assuming there's a follow up PR you can add such a comment to the file later.
Thanks

Issue #8910, opened 2023-09-28
Fix ContractDeleteSuite tests


vtronkov:

vtronkov:
This will be fixed in https://github.com/hashgraph/hedera-services/issues/9124

PR #8911, opened 2023-09-28
Fix ContractDeleteSuite tests 

vtronkov:
**Description**:
Adding MAX_CHILD_RECORDS_EXCEEDED validation

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8910


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/ActionSidecarContentTracer.java
```diff
@@ -51,4 +51,9 @@ public interface ActionSidecarContentTracer extends OperationTracer {
      * @param type the type of precompile called; expected values are {@code PRECOMPILE} and {@code SYSTEM}
      */
     void tracePrecompileResult(@NonNull MessageFrame frame, @NonNull ContractActionType type);
+
+    /**
+     * TODO: Add documentation
+     */
```
vtronkov, line 57: Since I'm not sure about the implementation - I will add the documentation when the code is finalized.

tinker-michaelj (changes requested):
We only want to limit the number of preceding/following synthetic `TransactionBody`'s; not the number of actions in the `EvmActionTracer`. 🤔 

I suspect we need at least a bit more clarity on how records are managed before we tackle this one.

vtronkov:
> We only want to limit the number of preceding/following synthetic `TransactionBody`'s; not the number of actions in the `EvmActionTracer`. 🤔
> 
> I suspect we need at least a bit more clarity on how records are managed before we tackle this one.

Ohh, I see what you are saying. Okay, I will leave it for now. Maybe I've bitten off more than I can chew 😄 I will close this PR for now then.

PR #8916, opened 2023-09-28
Fix Auto Create with Alias 

georgi-l95:
**Description**:
Fixes dispatching of synthetic transaction upon cryptoTransfer for auto creating an account with alias.
Alias have length of 35 as well as starts with certain characters.

All tests in those suites are failing with `FAIL_INVALID`, this PR fixes this error and gets them closer to passing.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
+++ b/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
@@ -51,7 +51,7 @@ public record ContractsConfig(
         @ConfigProperty(defaultValue = "5000000") @NetworkProperty long maxNumber,
         // CHAINID returns 295 (0x0127) for mainnet, 296 (0x0128) for testnet, and 297 (0x0129) for previewnet.
         // c.f. https://hips.hedera.com/hip/hip-26 for reference
-        @ConfigProperty(defaultValue = "298") @NetworkProperty int chainId,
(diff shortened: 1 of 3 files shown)


hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -51,7 +51,7 @@ public record ContractsConfig(
         @ConfigProperty(defaultValue = "5000000") @NetworkProperty long maxNumber,
         // CHAINID returns 295 (0x0127) for mainnet, 296 (0x0128) for testnet, and 297 (0x0129) for previewnet.
         // c.f. https://hips.hedera.com/hip/hip-26 for reference
-        @ConfigProperty(defaultValue = "295") @NetworkProperty int chainId,
+        @ConfigProperty(defaultValue = "298") @NetworkProperty int chainId,
```
Neeharika-Sompalli, line 54: I think we should have defaults same as mainnet.  I believe 298 is for dev. 
If we have to change tests for this or override to 298 in tests, I think we should.

georgi-l95: Gotcha, i wasn't sure about this change.

iwsimon (approved):
LGTM

povolev15 (approved):

Neeharika-Sompalli (approved):

PR #8940, opened 2023-10-02
Remove hapi test suite annotation from suite that not need to be fixed

MiroslavGatsanoga:
**Description**:

We have some suites that do not look like a regular suite that should be run on it’s own, but rather as an additional step after the execution of all suites to check the behaviour after them:
- `BalanceValidation`
- `TokenBalanceValidation`
- `TransactionBodyValidation`
- `RecordStreamValidation`

Also `PrngSeedOperationSuite` is broken and is not being run against mono, removing the annotation from it as well as won't fix for now.
  
**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


netopyr (dismissed):
LGTM

netopyr (approved):
LGTM

I created issues to handle the suites later appropriately.

PR #8941, opened 2023-10-02
Fix TxnReceiptRegression suite

### LINKED ISSUE #8917: FIx TxnReceiptRegression

stoyanov-st:
**Description**:
Fixed some of `TxnReceiptRegression` tests.

**Related issue(s)**:

Fixes #8917 
**Notes for reviewer**:
Added `@HapiTest` annotation to `returnsInvalidForUnspecifiedTxnId()`, `receiptAvailableWithinCacheTtl()` and `receiptNotFoundOnUnknownTransactionID()`. I had to extend the if condition in `validate` in the `NetworkTransactionGetReceiptHandler` in order to throw the correct exceptions and change the expected return of `receiptNotFoundOnUnknownTransactionID ()` to `INVALID_TRANSACTION_ID` as that is the implemented logic.

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnReceiptRegression.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnReceiptRegression.java
@@ -43,7 +43,7 @@ import org.apache.logging.log4j.Logger;
 public class TxnReceiptRegression extends HapiSuite {
     static final Logger log = LogManager.getLogger(TxnReceiptRegression.class);
 
-    public static void main(String... args) {
+    public static void main(final String... args) {
         new TxnReceiptRegression().runSuiteSync();
     }
 
@@ -65,6 +65,7 @@ public class TxnReceiptRegression extends HapiSuite {
         });
     }
 
+    @HapiTest
     private HapiSpec returnsInvalidForUnspecifiedTxnId() {
         return defaultHapiSpec("ReturnsInvalidForUnspecifiedTxnId")
                 .given()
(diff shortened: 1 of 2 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnReceiptRegression.java
```diff
@@ -113,13 +115,14 @@ private HapiSpec receiptUnavailableIfRejectedInPrecheck() {
...
+    @HapiTest
     private HapiSpec receiptNotFoundOnUnknownTransactionID() {
         return defaultHapiSpec("receiptNotFoundOnUnknownTransactionID")
                 .given()
                 .when()
                 .then(withOpContext((spec, ctxLog) -> {
-                    HapiGetReceipt op =
-                            getReceipt(spec.txns().defaultTransactionID()).hasAnswerOnlyPrecheck(RECEIPT_NOT_FOUND);
+                    final HapiGetReceipt op = getReceipt(spec.txns().defaultTransactionID())
+                            .hasAnswerOnlyPrecheck(INVALID_TRANSACTION_ID);
```
Neeharika-Sompalli, line 125: We should not be changing the response and comply with mono-service

rbair23: That is correct, the test logic should not change. We need to align the behavior in the modular code to match the mono service behavior, and the tests are the guard to ensure we do it correctly.

stoyanov-st: Understood. Changes are applied.

Neeharika-Sompalli (approved):
LGTM !

PR #8947, opened 2023-10-02
Fix TxnRecordRegression

### LINKED ISSUE #8943: Fix TxnRecordRegression

stoyanov-st:
**Description**:
Fix one of `TxnRecordRegression` tests.

**Related issue(s)**:

Fixes #8943 

**Notes for reviewer**:
Added @HapiTest annotation to `recordNotFoundIfNotInPayerState()` test, had to change the expected return to `INVALID_TRANSACTION_ID` as that is the implemented logic.

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnRecordRegression.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnRecordRegression.java
@@ -29,6 +29,7 @@ import static com.hedera.services.bdd.spec.utilops.UtilVerbs.usableTxnIdNamed;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.ACCOUNT_DELETED;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.INSUFFICIENT_PAYER_BALANCE;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.INVALID_ACCOUNT_ID;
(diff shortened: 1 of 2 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnRecordRegression.java
```diff
@@ -101,13 +102,14 @@ private HapiSpec returnsInvalidForUnspecifiedTxnId() {
...
 
+    @HapiTest
     private HapiSpec recordNotFoundIfNotInPayerState() {
         return defaultHapiSpec("RecordNotFoundIfNotInPayerState")
                 .given(
                         cryptoCreate("misc").via("success"),
                         usableTxnIdNamed("rightAccountWrongId").payerId("misc"))
                 .when()
-                .then(getTxnRecord("rightAccountWrongId").hasCostAnswerPrecheck(RECORD_NOT_FOUND));
+                .then(getTxnRecord("rightAccountWrongId").hasCostAnswerPrecheck(INVALID_TRANSACTION_ID));
```
Neeharika-Sompalli, line 112: We should not be changing the response we are returning and comply with mono-service

rbair23: Correct, we need the test to stay the same, and for the modular node behavior to match mono service.

stoyanov-st: Understood. Changes are applied.

Neeharika-Sompalli (approved):
LGTM!

PR #8986, opened 2023-10-02
Fix `createContractWithStakingFields()`

### LINKED ISSUE #8719: Implement staking id validation logic for `ContractCreate` transactions
We have the `createContractWithStakingFields` test from `ContractCreateSuite` failing due to this missing.
I think we should implement the equivalent logic from `ContractCreateTransitionLogic.validate()` in mono-service, in particular [this](https://github.com/hashgraph/hedera-services/blob/c672967964359c08a8b21c4a17563a1b3e599f51/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java#L335).

tinker-michaelj:
**Description**:
 - Closes #8719 
   * The remaining issue with the `createContractWithStakingFields()` spec was that a `_TO_ACCOUNT` stake change scenario would cause a `ConcurrentModificationException` in the `for` loop [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/staking/StakingRewardsHandlerImpl.java#L105), since the logic _inside_ the loop would add the "stakee" account to the modified set---and cause the iterator to detect a concurrent modification.
- Adds a spec validating system accounts cannot be used as initcode.


Neeharika-Sompalli (approved):
Makes sense to me!

Nana-EC (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/staking/StakingRewardsHandlerImpl.java
```diff
@@ -100,7 +101,10 @@ public Map<AccountID, Long> applyStakingRewards(final FinalizeContext context) {
      * @param writableStore The store to write to for updated values
      */
     public void adjustStakedToMeForAccountStakees(@NonNull final WritableAccountStore writableStore) {
-        final var modifiedAccounts = writableStore.modifiedAccountsInState();
+        // If there is a FROM_ACCOUNT_ or _TO_ACCOUNT stake change scenario, the set of modified
+        // accounts returned by writableStore.modifiedAccountsInState(); so we only iterate through
```
mhess-swl, line 105: > the set of modified accounts returned by writableStore.modifiedAccountsInState();

I expected something to complete this statement, did I misread?

mhess-swl (approved):

kimbor (approved):

iwsimon (approved):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/staking/StakingRewardsHandlerImpl.java
tinker-michaelj: Ah fixed, oops.

iwsimon (dismissed):

mhess-swl (dismissed):

Nana-EC (approved):

mhess-swl (approved):

PR #8991, opened 2023-10-03
Differential testing analytic engine: State file file dumper now dumps special files

### LINKED ISSUE #8990: Differential testing analytical engine: Dumper for special file store
System upgrade files are put in a special file store (`MerkleSpecialFiles`, a specialization of `MerkleLeaf`) and need to be dumped with the other files.

david-bakin-sl:
**Description**:

State file file dumper now dumps special files

* And for system files it tells you what file it is. 
* And you can ask for the report _without_ the file contents (much smaller file, suitable for putting into a spreadsheet.

**Related issue(s)**:

Fixes #8990 

**Checklist**

- [x] Tested (manual)

### DIFF
--- a/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
+++ b/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
@@ -499,7 +499,7 @@ public class DumpFilesSubcommand {
 
         final var r = new HashMap<Integer, HederaFile>();
         final int HIGHEST_NON_USER_ACCOUNT = 1000;
-        for (int id = 0; id < HIGHEST_NON_USER_ACCOUNT; id++) {
+        for (int id = 0; id <= HIGHEST_NON_USER_ACCOUNT; id++) {
             final var fileId = FileID.newBuilder().setFileNum(id).build();
             if (specialFileStore.contains(fileId)) {
                 final var contents = specialFileStore.get(fileId);
@@ -513,6 +513,10 @@ public class DumpFilesSubcommand {
         return Pair.of(fileSummary, r);
     }
 
+    /** Merge two (or more) maps.
+     *
+     * (Seems like it should be provided by the `Map` class, but maybe not.)
+     */
     @SafeVarargs
     @NonNull
     static <K, V> Map<K, V> merge(@NonNull final BinaryOperator<V> mergeFunction, @NonNull final Map<K, V>... maps) {
@@ -521,6 +525,10 @@ public class DumpFilesSubcommand {
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, mergeFunction));
     }
 
+    /** Pick `a` or `b` arbitrarily.
+     *
+     *  Not usually a good idea, but in this case it is really never called, as it a merge function for disjoint sets
+     */
     <T> T arb(@NonNull final T a, @NonNull final T b) {
         final var choice = new Random().nextBoolean();
         return choice ? a : b;
--- a/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
+++ b/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
@@ -17,6 +17,7 @@
 package com.hedera.services.cli.signedstate;
 
 import static com.hedera.services.cli.utils.ThingsToStrings.quoteForCsv;
+import static com.hedera.services.cli.utils.ThingsToStrings.squashLinesToEscapes;
 import static com.hedera.services.cli.utils.ThingsToStrings.toStringOfByteArray;
 import static com.hedera.services.cli.utils.ThingsToStrings.toStructureSummaryOfJKey;
 import static com.swirlds.common.threading.manager.AdHocThreadManager.getStaticThreadManager;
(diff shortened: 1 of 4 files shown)


hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
```diff
@@ -154,19 +233,28 @@ void reportSummary(
...
 
-        final var nMetadataFiles = fileSummary.typeOccurrences().get(VirtualBlobKey.Type.FILE_METADATA);
+        final var nMetadataFiles = fileSummary.typeOccurrences().getOrDefault(VirtualBlobKey.Type.FILE_METADATA, 0);
         final var nActiveDataFiles =
                 allFiles.values().stream().filter(HederaFile::isActive).count();
-        final var nActiveDataFilesWithMetadata =
-                allFiles.values().stream().filter(HederaFile::isActive).count();
+        final var nActiveDataFilesWithMetadata = allFiles.values().stream()
+                .filter(HederaFile::isActive)
+                .filter(hf -> null != hf.metadata())
```
mhess-swl, line 245:
```suggestion
                .filter(hf -> hf.metadata() != null)
```
Personally I'd flip these to make it more readable, but that's a completely subjective suggestion

mhess-swl (approved):

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
```diff
@@ -383,7 +488,49 @@ Pair<FileSummary, Map<Integer, HederaFile>> collectFiles(
...
+        }
+
+        final var typeMap = Map.of(VirtualBlobKey.Type.FILE_DATA, r.size());
+        final var fileSummary = new FileSummary(typeMap, Map.of(), 0);
+        return Pair.of(fileSummary, r);
+    }
+
+    /** Merge two (or more) maps.
+     *
+     * (Seems like it should be provided by the `Map` class, but maybe not.)
```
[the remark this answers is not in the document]
david-bakin-sl: Yes but why not a `Map.putAll` overload that takes a remapping function like `Map.merge`?  (Which is what I wanted here.)

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
```diff
@@ -114,32 +132,93 @@ void doit() {
...
-            return new HederaFile(contractId, contents, metadata);
+        static HederaFile of(@NonNull final FileStore fileStore, final int fileId, @NonNull final byte[] contents) {
+            return new HederaFile(fileStore, fileId, contents, null, SystemFileType.byId.get(fileId));
         }
 
         boolean isActive() {
-            return null != metadata && !metadata.isDeleted();
+            if (null != systemFileType) return true;
+            if (null != metadata) return !metadata.isDeleted();
+            return false;
```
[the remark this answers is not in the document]
david-bakin-sl: I get you, but many times - like here - I prefer the straight "ladder" of condition->return followed by the final return true/false.  And, TBH, a limitation of my brain is that your suggested conjunction is more difficult to follow that the simpler conditional followed by the bare return.

(Java really could use a null-coalescing or null-conditional operator like C#.)

Will leave this as-is.

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
david-bakin-sl: So you don't like Yoda conditions?  It's an acquired taste to be sure.  (But you didn't complain about it up at L219-220?)

I just now tried `.filter(hf -> nonNull(hf.systemFileType())` (using `Objects::nonNull`) - I think I like it better and will use it.

(Actually what `Stream.filter` really needs is an overload `filter(Function<T,U> mapper, Predicate<U> predicate)` that does `e -> predicate(mapper(e))` so you could write, e.g., `filter(HederaFile::fileStore, isEqual(FileStore.SPECIAL))` - filtering on a condition of a field (or function) of the stream element but passing through the stream element.)

mhess-swl (approved):

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
mhess-swl: This sparked an idea – a YouTube channel where Yoda reads source code as listeners fall asleep. You're welcome

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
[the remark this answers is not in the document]
jsync-swirlds: Map is already a rather large interface, and adding every useful option would make it unmanageable.  The assumption is that you will compose what's there (e.g. forEach with merge or replaceAll on the source map, then putAll to merge).
Using the Streams API is a reasonable approach, and mostly matches "recent" java practice (as well as being more "functional").

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpFilesSubcommand.java
```diff
@@ -318,7 +424,7 @@ Pair<FileSummary, Map<Integer, HederaFile>> collectFiles(
 
         final var nType = new ConcurrentHashMap<VirtualBlobKey.Type, Integer>();
         final var nNullValues = new ConcurrentHashMap<VirtualBlobKey.Type, Integer>();
-        final var nNulLMetadataValues = new AtomicInteger();
+        final var nNullMetadataValues = new AtomicInteger();
```
jsync-swirlds, line 427: What is the `n` prefix?  It's easy to miss if one doesn't read slowly.

david-bakin-sl: `n` == "number" (aka count) -

PR #9012, opened 2023-10-03
08663 event creation cleanup

### LINKED ISSUE #7922: Remove `StartUpEventFrozenManager`
With the new platform status system, the `StartUpEventFrozenManager` is no longer necessary. The platform status prevents events from being created at startup

### LINKED ISSUE #8663: Make tipset event creation the only event creation algorithm
Remove non-tipset event creation logic.

alittley:
- closes #8663 
- closes #7922


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/SingleNodeSyncGossip.java
cody-littley, line 1: Just to be safe, please verify that a one node network still works. A local test by hand is sufficient.

jsync-swirlds (approved):
Approved with respect to the one file owned by "architects".
Not sure how the removal of FreezeManager might interact with NetworkAdmin service "Freeze" transaction.  If there is any interaction or impact in this regard, please communicate that to the services team.

cody-littley:
> Approved with respect to the one file owned by "architects".
Not sure how the removal of FreezeManager might interact with NetworkAdmin service "Freeze" transaction. If there is any interaction or impact in this regard, please communicate that to the services team.

Freeze behavior is not expected to change. The FreezeManager is a class that we used to use to implement freeze, but we have since replaced it.

PR #9015, opened 2023-10-04
Fix `smallContractCreate()` and enable `ethereumCallWithCalldataBiggerThanMaxSucceeds()`

tinker-michaelj:
**Description**:
 - Avoids ISS in `smallContractCreate()` by removing special treatment of `node0` in `HapiTestEnv`.
 - When un-aliasing an `0.0.<alias>` form `AccountID` in `ReadableAccountStoreImpl`, also considers "indirect" references where an ECDSA key has an implied EVM address that is an in-use alias.
 - When answering a `CryptoGetInfo` query, ensures the returned `AccountID` is a numeric `0.0.X` id.
 - Also enables a passing `ethereumCallWithCalldataBiggerThanMaxSucceeds()` spec (have not checked others).

### DIFF
--- a/hedera-node/configuration/dev/application.properties
+++ b/hedera-node/configuration/dev/application.properties
@@ -1,4 +1,4 @@
 balances.exportDir.path=data/accountBalances/
-ledger.autoRenewPeriod.minDuration=2592000
+ledger.autoRenewPeriod.minDuration=10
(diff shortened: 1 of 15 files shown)


georgi-l95 (approved):
LG, we can merge into `fix-ethereum-suites` and I'll finish it up. Thank you 🙏

PR #9035, opened 2023-10-05
Issue 8989 - Design ScheduleGetInfo query handler

### LINKED ISSUE #8989: Design ScheduleInfo query

jsync-swirlds:
* Added implementaiton of the ScheduleGetInfo query handler
* Added unit test for ScheduleGetInfo query handler
* Fixed a minor oversight in logging config that caused confusing output in gradle run.

Closes #8989

### DIFF
--- a/hedera-node/log4j2.xml
+++ b/hedera-node/log4j2.xml
@@ -198,9 +198,9 @@
     </Logger>
 
     <!-- Send transaction state logs to their own appender   -->
-    <AsyncLogger name="com.hedera.node.app.state.logging.TransactionStateLogger" level="trace" additivity="false">
+    <Logger name="com.hedera.node.app.state.logging.TransactionStateLogger" level="trace" additivity="false">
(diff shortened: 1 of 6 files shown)


hedera-node/log4j2.xml
```diff
@@ -198,9 +198,9 @@
     </Logger>
 
     <!-- Send transaction state logs to their own appender   -->
-    <AsyncLogger name="com.hedera.node.app.state.logging.TransactionStateLogger" level="trace" additivity="false">
+    <Logger name="com.hedera.node.app.state.logging.TransactionStateLogger" level="trace" additivity="false">
```
povolev15, line 201: I thought it was supposed to be asyncLogger. why do we change it to Logger?

jsync-swirlds: Because It is not supposed to be async yet, as it pulls in a dependency we don't want to bring in yet.
Jasper asked for this change via a slack thread.

PR #9043, opened 2023-10-05
Fix SignedTransactionBytesRecordsSuite

### LINKED ISSUE #9019: SignedTransactionBytesRecordsSuite
Fix SignedTransactionBytesRecordsSuite

MrValioBg:
**Description**:

Enable tests in `SignedTransactionBytesRecordsSuite`.

Remove validation logic for valid receipt/record in both `NetworkTransactionGetRecordHandler` and `NetworkTransactionGetReceiptHandler`
 - In mono we only return RECORD/RECEIPT_NOT_FOUND for ANSWER_ONLY flow, validating it prior the code in `findResponse` will make it throw this precheck in COST_ANSWER.
 - We already have a flow in the `findResponse` method which handles the cases for RECORD/RECEIPT_NOT_FOUND and returns it as the precheck for ANSWER_ONLY
 
 

**Related issue(s)**:

Fixes #9019 

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnRecordRegression.java
```diff
@@ -108,16 +108,17 @@ private HapiSpec recordNotFoundIfNotInPayerState() {
                         cryptoCreate("misc").via("success"),
                         usableTxnIdNamed("rightAccountWrongId").payerId("misc"))
                 .when()
-                .then(getTxnRecord("rightAccountWrongId").hasCostAnswerPrecheck(RECORD_NOT_FOUND));
+                .then(getTxnRecord("rightAccountWrongId").hasAnswerOnlyPrecheck(RECORD_NOT_FOUND));
```
mhess-swl, line 111: @Neeharika-Sompalli @tinker-michaelj is this change correct?

iwsimon: I am wondering this too. We normally should not change the test itself as it also needs to run in mono code successfully.

Neeharika-Sompalli: Not sure too. Why do we need this change @MrValioBg  ?

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetReceiptHandler.java
```diff
@@ -74,11 +73,6 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
         if (!op.hasTransactionID() || !op.transactionID().hasAccountID()) {
             throw new PreCheckException(INVALID_TRANSACTION_ID);
         }
-
-        // The receipt must exist for that transaction ID
-        final var recordCache = context.recordCache();
-        final var receipt = recordCache.getReceipt(op.transactionIDOrThrow());
-        mustExist(receipt, RECEIPT_NOT_FOUND);
```
Neeharika-Sompalli, line 81: Why should we remove this ?

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetRecordHandler.java
```diff
@@ -83,10 +82,6 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
         if (accountID == null || accountID.equals(AccountID.DEFAULT)) {
             throw new PreCheckException(INVALID_ACCOUNT_ID);
         }
-
-        final var recordCache = context.recordCache();
-        final var receipt = recordCache.getReceipt(txnId);
-        mustExist(receipt, RECORD_NOT_FOUND);
```
Neeharika-Sompalli, line 89: same here, are we moving this check somewhere else ?

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetReceiptHandler.java
MrValioBg: In mono, we only return RECORD/RECEIPT_NOT_FOUND for the ANSWER_ONLY flow. If, in the modularized code, we validate this here before the findResponse method in either `NetworkTransactionGetRecordHandler` or `NetworkTransactionGetReceiptHandler`, it will cause this precheck to be thrown in COST_ANSWER, leading to a different behavior.

We already have a flow within the findResponse method that handles the cases for RECORD/RECEIPT_NOT_FOUND and returns it as the precheck for ANSWER_ONLY.

Neeharika-Sompalli (approved):
LGTM ! Thanks @MrValioBg

iwsimon (approved):
LGTM

Issue #8971, opened 2023-10-02
Add Fee Computation to remaining Network Service queries


kimbor:
This issue covers the following services:
- [ ] NetworkGetByKeyHandler
- [x] NetworkTransactionGetFastRecordHandler
- [x] NetworkTransactionGetReciptHandler
- [x] NetworkTransactionGetRecordHandler

All other network admin services are covered in issue #8683

PR #9052, opened 2023-10-05
Fix Paid query handler base class to not permit subclasses to be free.

jsync-swirlds:
The query handler inferface has a default implementation for fee calculation that reutrns `FREE`.  This is clearly incorrect for paid queries.
   The paid query abstract base class does not override this default, however, and six (6) paid queries do not implement fee calculation as a result.
   This change adds an abstract declaration of the fee calculation method to the PaidQueryHandler interface so that _paid_ query subclasses must implement fee calculation.
* Made fee calculation for base class PaidQueryHandler an abstract declaration instead of defaulting to FREE (from interface).
   * Updated four(4) contract query handlers and two(2) network admin query handlers with a partial (just calculate the fee from workflow) non-free implementation of fee calculation to avoid compile issues.
   * All six handlers will need corrected implementations (or to not extend the PaidQueryHandler base class if they should actually be free queries).

Related issues:
#8971 
#9054 

Preceding PR:
https://github.com/hashgraph/hedera-services/pull/9035


rbair23 (approved):

povolev15 (approved):

kimbor (approved):
lgtm!

tinker-michaelj (approved):
LGTM

Issue #9005, opened 2023-10-03
Prototype wiring framework


cody-littley:
Iterate on the proposed wiring framework.

PR #9107, opened 2023-10-06
09005 Wiring Framework

cody-littley:
Adds metrics and backpressure capability to wiring.

Related to #9005


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/internal/MeteredSequentialWire.java
```diff
@@ -0,0 +1,207 @@
...
+            send();
+        }
+
+        /**
+         * Execute this task.
+         */
+        @Override
+        public boolean exec() {
+            counter.offRamp();
+            busyTimer.activate();
```
OlegMazurov, line 85: `SequentialTask`s are executed serially by design. Paying high price for updating atomics is not justified here (applies to `busyTimer` but not `counter`).

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,150 @@
...
+
+        if (sleepDuration == null) {
+            sleepNanos = -1;
+        } else {
+            sleepNanos = sleepDuration.toNanos();
+        }
+
+        increment = count -> {
+            if (count >= capacity) {
+                throw new NoCapacityException();
```
lpetrovic05, line 62: throwing an exception is expensive, better to use `compareAndSet`

cody-littley: I actually think this may be faster in the standard case we want to optimize, i.e. when there is available capacity and we don't need to apply back pressure.

If I use compare and set, I'd have to perform two operations for each on-ramping, even when capacity is not reached.
1) read the current value
2) do a compare and set operation

With the throwing approach, I only have to do one atomic long operation per on-ramping. When we are at capacity there is additional overhead from the exception we are throwing.

In my opinion, we should optimize the performance of the scenario when there is available capacity, even if there are costs for when capacity is scarce.

lpetrovic05: If you look inside of `updateAndGet`, there is a loop, it has to do the same thing basically. So I don't think there is any benefit of this.

You could go with `getAndAccumulate`, that way you only call the method once

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean block() throws InterruptedException {
+        if (sleepNanos > 0) {
+            try {
+                NANOSECONDS.sleep(sleepNanos);
```
lpetrovic05, line 65: I thought we were not going to sleep...

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/WireBuilder.java
```diff
@@ -0,0 +1,331 @@
...
+    @NonNull
+    public Wire<O> build() {
+        final Counters counters = buildCounters();
+        final FractionalTimer busyFractionTimer = buildBusyTimer();
+
+        if (metricsBuilder != null) {
+            metricsBuilder.registerMetrics(name, counters.onRamp());
+        }
+
+        // TODO add a way to signal back pressure that is external to the wire
```
lpetrovic05, line 305: TODO

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/Wire.java
```diff
@@ -0,0 +1,183 @@
...
+     * <p>
+     * Warning: this will appease the compiler, but it is possible to cast a wire into a data type that will cause
+     * runtime exceptions. Use with appropriate caution.
+     *
+     * @param <X> the type to cast to
+     * @return this, cast into whatever type is requested
+     */
+    @NonNull
+    @SuppressWarnings("unchecked")
+    public final <X> Wire<X> cast() {
```
lpetrovic05, line 64: I'm not sure this is needed

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/model/ModelTests.java
```diff
@@ -0,0 +1,933 @@
...
+class ModelTests {
+
+    /**
+     * For debugging with a human in the loop.
+     */
+    private static final boolean printMermaidDiagram = false;
+
+    //    @BeforeAll
+    //    public static void beforeAll() throws InterruptedException {
+    //        // TODO remove
```
lpetrovic05, line 41: todo

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
OlegMazurov: Throwing an exception is very expensive in performance sensitive code like this (costs CPU and generates a lot of garbage). Especially when it's done in an implicit busy wait loop: `ForkJoinPool.managedBlock()` -> `BackpressureBlocker.isReleasable()`.
We may consider two blocking implementations: busy wait (attempt to increment until successful) and a thread blocking/unblocking one.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
cody-littley: Sleeping is ok if and only if it is coordinated with the fork join pool. This class implements `ManagedBlocker`. It's designed specifically to block in a way that works properly with a fork join pool workflow.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/Wire.java
```diff
@@ -0,0 +1,183 @@
...
+
+import com.swirlds.common.wiring.counters.ObjectCounter;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.function.Consumer;
+
+/**
+ * Wires two or more components together.
+ *
+ * @param <O> the output time of the wire (use {@link Void}) for a wire with no output type)
```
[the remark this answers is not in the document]
cody-littley: We discussed this, and I ended up adding secondary output channels. Return type always goes to the primary output to keep things simple for the most common use case.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/Wire.java
cody-littley: This is actually super useful, ping me and I'll show you examples.

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/model/ModelTests.java
cody-littley: resolved

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputChannel.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Get the wiring model containing this output channel.
+     *
+     * @return the wiring model
+     */
+    @NonNull
+    protected WiringModel getModel() { // TODO is this needed?
+        return model;
+    } // TODO necessary?
```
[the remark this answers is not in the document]
cody-littley: resolved

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: I went with the two-implementation approach. An exception for the first time we run out of capacity, and compare-and-set for the busy loop.

cody-littley:
This PR is superseded by https://github.com/hashgraph/hedera-services/pull/9482

PR #9132, opened 2023-10-09
08138 Increased `reconnect.asyncStreamTimeout` property to prevent genesis reconnect from failure

### LINKED ISSUE #8138: VMerkleDual-NDReconnect-500-60m failing while transmitting virtual map
Originally documented in https://github.com/swirlds/swirlds-platform-regression/issues/3527

imalygin:
Increased `reconnect.asyncStreamTimeout` property to prevent genesis reconnect from failure

**Description**:

In case the tree gets too big, hashing may take more that 60s and in this case the learner node doesn't send the lesson confirmation on time, and the teacher node assumes that a connection between nodes should be closed. After that the nodes are not able to finish the reconnect properly. 

**Related issue(s)**:

Fixes #8138 

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/test/resources/signedState/MHS/settingsUsed.txt
+++ b/hedera-node/hedera-app/src/test/resources/signedState/MHS/settingsUsed.txt
@@ -96,7 +96,6 @@ data/threadDump = threadDumpLogDir
            true = reconnect.active
              -1 = reconnect.reconnectWindowSeconds
             0.5 = reconnect.fallenBehindThreshold
-          60000 = reconnect.asyncStreamTimeoutMilliseconds
             100 = reconnect.asyncOutputStreamFlushMilliseconds
           10000 = reconnect.asyncStreamBufferSize
            true = reconnect.asyncStreams
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/config/ReconnectConfig.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/config/ReconnectConfig.java
@@ -54,7 +54,7 @@ public record ReconnectConfig(
         @ConfigProperty(defaultValue = "true") boolean active,
         @ConfigProperty(defaultValue = "-1") int reconnectWindowSeconds,
         @ConfigProperty(defaultValue = "0.50") double fallenBehindThreshold,
-        @ConfigProperty(defaultValue = "60s") Duration asyncStreamTimeout,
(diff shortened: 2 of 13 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/config/ReconnectConfig.java
```diff
@@ -54,7 +54,7 @@ public record ReconnectConfig(
         @ConfigProperty(defaultValue = "true") boolean active,
         @ConfigProperty(defaultValue = "-1") int reconnectWindowSeconds,
         @ConfigProperty(defaultValue = "0.50") double fallenBehindThreshold,
-        @ConfigProperty(defaultValue = "60s") Duration asyncStreamTimeout,
+        @ConfigProperty(defaultValue = "180s") Duration asyncStreamTimeout,
```
artemananiev, line 57: This is not enough. The value is explicitly set per environment, e.g. in `hedera-node/configuration/mainnet/settings.txt` They should be changed, too (or better removed, since we usually don't have overrides, which are identical to default values)

poulok: in the `configuration/*/settings.txt` files, the old setting name is specified (`reconnect.asyncStreamTimeoutMilliseconds, 60000`), so i dont think that value is even being used. It should be removed from all the `settings.txt` files.

imalygin: fair enough. I'll fix it

imalygin: Done

artemananiev (approved):

poulok (approved):

rbair23 (approved):

hedera-node/hedera-app/src/test/resources/signedState/MHS/settingsUsed.txt
cody-littley, line 1: I'm not sure there is a need to update the `settingsUsed.txt` files. This is just data that was generated a long time ago. IMO it probably shouldn't have ever been committed to the repo in the first place.

This is not a merge blocker if the timeline needs a very fast merge. It won't cause problems, it's just not necessary.

cody-littley (approved):

PR #9139, opened 2023-10-09
Fix fee calculations to not throw exceptions

### LINKED ISSUE #9134: Handle thrown exceptions in `calculateFees()`
There are exceptions thrown in `calculateFees()`. But, they were not caught correctly in Ingest and Handle Workflows. Either change the method to not throw or catch the exception correctly in workflows.

Neeharika-Sompalli:
Fixes #9134 

Fixes fee calculations to not throw any errors. 
Stopped throwing exception in mono-service from `usageGiven` because only a few handlers were following the pattern to throw when transactionBody is not null. 
This is non-reachable code path, since that check should already be done before it reaches fee calculation stage, otherwise we should have failed with `FAIL_INVALID` in fees.


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/fees/calculation/consensus/txns/UpdateTopicResourceUsage.java
```diff
@@ -86,9 +80,8 @@ public FeeData usageGivenExplicit(
                         merkleTopic.hasAutoRenewAccountId(),
                         expiry,
                         txnBody.getConsensusUpdateTopic());
-            } catch (final InvalidKeyException illegal) {
-                log.warn("Usage estimation unexpectedly failed for {}!", txnBody, illegal);
-                throw new InvalidTxBodyException(illegal);
+            } catch (final InvalidKeyException ignore) {
+                // The key was validated before setting on MerkleTopic, so this should never happen
```
mhess-swl, line 84: I agree, this shouldn't happen, but in case it does, would it still be worth logging the warning so it doesn't fail silently?

hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/fees/calculation/UsageBasedFeeCalculatorTest.java
```diff
@@ -346,21 +345,16 @@ void throwsIseOnBadScheduleInFcfs() {
...
         given(txnUsageEstimators.get(CryptoCreate)).willReturn(List.of(correctOpEstimator));
-        given(correctOpEstimator.usageGiven(
-                        argThat(accessor.getTxn()::equals),
-                        argThat(factory.apply(expectedSigUsage)),
-                        argThat(view::equals)))
-                .willThrow(InvalidTxBodyException.class);
 
         // when:
-        assertThrows(IllegalArgumentException.class, () -> subject.computeFee(accessor, payerKey, view, consensusNow));
+        assertThrows(NullPointerException.class, () -> subject.computeFee(accessor, payerKey, view, consensusNow));
```
mhess-swl, line 357: Is this correct? This looks more like a mock failed to return something

hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/fees/calculation/consensus/txns/UpdateMerkleTopicResourceUsageTest.java
```diff
@@ -96,66 +85,6 @@ void recognizesApplicableQuery() {
...
-                null,
-                null);
-        given(topics.get(EntityNum.fromTopicId(topicId))).willReturn(null);
-
-        final var feeData = subject.usageGiven(txBody, sigValueObj, view);
-
-        checkServicesFee(feeData, 0);
-        checkNetworkFee(feeData, 120, 0);
-        checkNodeFee(feeData, 120);
-    }
```
mhess-swl, line 157: Why are we getting rid of this test? Is it somehow related to the exceptions?

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/fees/calculation/consensus/txns/UpdateTopicResourceUsage.java
Neeharika-Sompalli: Makes sense. Done

hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/fees/calculation/consensus/txns/UpdateMerkleTopicResourceUsageTest.java
Neeharika-Sompalli: Reverted. Thanks

hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/fees/calculation/UsageBasedFeeCalculatorTest.java
Neeharika-Sompalli: Added a new mock. But NPE is expected

mhess-swl (approved):

tinker-michaelj (approved):
LGTM!

Issue #8899, opened 2023-09-27
Resolve CreateOperationSuite errors


stoyanov-st:
We have the following test suite `CreateOperationSuite` which has problems running against the modularized codebase. The `HapiSpecRegistry` entry for the contract has null value for `ContractID` thus producing `INVALID_CONTRACT_ID` errors.

PR #9142, opened 2023-10-10
Resolve CreateOperationSuite errors

vtronkov:
**Description**:
* Fixing failing E2E tests in CreateOperationSuite

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8899

**Notes for reviewer**:
* In `ProxyWorldUpdater.getHederaAccount`(the caller of `evmFrameState.getAddress`) we are returning null if the account/token is not found **AND** if the account is deleted. We do not differentiate between both. That's why instead of returning `null` when the account is deleted(in `evmFrameState.getAddress`) I directly throw an error with the correct `ReponseCode`
* The other change is to add a missing validation in `ContractGetBytecodeHandler`


iwsimon (dismissed):
LGTM. Although Unit Tests failed. You will have to check those.

vtronkov:
> LGTM. Although Unit Tests failed. You will have to check those.

The branch was based on 08718 and the failing tests came from there. Now that the PR is merged and I merged with develop it should be fine 🤞

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java
```diff
@@ -292,7 +294,7 @@ public long getIdNumber(@NonNull Address address) {
         }
 
         if (account.deleted()) {
```
tinker-michaelj, line 296: Here we definitely want to just return `null` (i.e. to the calling code a deleted account appears to be just missing---as it would be if the `selfdestruct` had been executed in Ethereum).

tinker-michaelj (changes requested):
The `ContractGetBytecode` check looks great, but we definitely don't want to throw `HandleException` in`getIdNumber`. 😁

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java
vtronkov: @tinker-michaelj I agree with the `HandleException` I throw it by default 😄

I think we should throw some error or have a way to differentiate between "deleted account" and "not found account/token"(see the PR "Notes for reviewer" section for details).

Maybe in ProxyWorldUpdated.getHederaAccount when we catch the `IllegalArgumentException` we shouldn't return null?

In the current implementation `factoryAndSelfDestructInConstructorContract` test is failing because it expects `CONTRACT_DELETED` but we return `INVALID_CONTRACT_ID`

tinker-michaelj: Ah got it; we should then add a top-level check in `HevmTransactionFactory.assertValidCall()` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java#L239) that fails "explicitly" on the deleted contract.

But afaict this method needs to have Besu EVM semantics; that is, return `null` when a usable contract does not exist, but don't throw unchecked exceptions.

vtronkov: Makse sense 👍 Fixed.

tinker-michaelj (approved):
Awesome, lgtm!

PR #9151, opened 2023-10-10
Fix create2 operation suite tests

MiroslavGatsanoga:
**Description**:

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8862

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.jpms-modules.gradle.kts
+++ b/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.jpms-modules.gradle.kts
@@ -14,11 +14,7 @@
  * limitations under the License.
  */
 
-import com.hedera.hashgraph.gradlebuild.rules.IoGrpcDependencyMetadataRule
-import com.hedera.hashgraph.gradlebuild.rules.IoGrpcMetadataRule
-import com.hedera.hashgraph.gradlebuild.rules.IoNettyNativeEpollMetadataRule
-import com.hedera.hashgraph.gradlebuild.rules.IoPrometheusSimpleclientMetadataRule
-import com.hedera.hashgraph.gradlebuild.rules.RemoveFindbugsAnnotationsMetadataRule
+import com.hedera.hashgraph.gradlebuild.rules.*
 
 plugins {
     id("org.gradlex.java-ecosystem-capabilities")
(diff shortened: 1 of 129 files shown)


rbair23 (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
```diff
@@ -221,4 +247,11 @@ private SystemPrivilege checkCryptoUpdate(
     private SystemPrivilege checkEntityDelete(final long entityNum) {
         return isSystemEntity(entityNum) ? IMPERMISSIBLE : UNNECESSARY;
     }
+
+    // move to somewhere else
```
lukelee-sl, line 251: Should this be moved elsewhere per the comment?

lukelee-sl (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
```diff
@@ -83,9 +86,21 @@ public SystemPrivilege hasPrivileges(
                     payerId, txBody.fileUpdateOrThrow().fileIDOrThrow().fileNum());
             case FILE_APPEND -> checkFileChange(
                     payerId, txBody.fileAppendOrThrow().fileIDOrThrow().fileNum());
-            case CONTRACT_UPDATE -> checkFileChange(
-                    payerId,
-                    txBody.contractUpdateInstanceOrThrow().contractIDOrThrow().contractNumOrThrow());
+            case CONTRACT_UPDATE -> {
+                var contractId = txBody.contractUpdateInstanceOrThrow().contractIDOrThrow();
+
+                if (contractId.hasEvmAddress()) {
```
tinker-michaelj, line 92: There are no system contracts (below `0.0.1000`) to check for special update and delete privileges---and almost certainly never will be such system contracts.

(If there _were_, I don't think we would want to assume they only had long-zero EVM addresses.)

I'd suggest just deleting the `CONTRACT_UPDATE` and `CONTRACT_DELETE` branches in this switch; created https://github.com/hashgraph/hedera-services/issues/9383 for that work since all checks are passing here.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -216,7 +216,7 @@ public void createContract(final long number, final long parentNumber, @Nullable
                 number,
                 synthAccountCreationFromHapi(
                         ContractID.newBuilder().contractNum(number).build(), evmAddress, impliedContractCreation),
-                null);
+                parent.autoRenewAccountId());
```
tinker-michaelj, line 219: Nice catch. 🙌

tinker-michaelj (approved):
LGTM!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
tinker-michaelj: We should just delete this entire block imo @lukelee-sl , c.f. https://github.com/hashgraph/hedera-services/issues/9383

Issue #8466, opened 2023-09-08
Test CryptoTransferSuite


Neeharika-Sompalli:
- [x] aliasKeysAreValidated
- [x] allowanceTransfersWithComplexTransfersWork
- [x] allowanceTransfersWorkAsExpected
- [x] autoAssociationRequiresOpenSlots
- [x] baseCryptoTransferFeeChargedAsExpected
- [x] cannotTransferFromImmutableAccounts
- [x] dissociatedRoyaltyCollectorsCanUseAutoAssociation
- [x] hapiTransferFromForFungibleTokenWithCustomFeesWithAllowance
- [x] hapiTransferFromForNFTWithCustomFeesWithAllowance
- [x] hbarAndFungibleSelfTransfersRejectedBothInPrecheckAndHandle
- [x] nftSelfTransfersRejectedBothInPrecheckAndHandle
- [x] nftTransfersCannotRepeatSerialNos
- [x] okToRepeatSerialNumbersInBurnList
- [x] okToRepeatSerialNumbersInWipeList
- [x] okToSetInvalidPaymentHeaderForCostAnswer
- [x] royaltyCollectorsCanUseAutoAssociation
- [x] specialAccountsBalanceCheck
- [x] tokenTransferFeesScaleAsExpected
- [x] transferToNonAccountEntitiesReturnsInvalidAccountId
- [x] transferWithMissingAccountGetsInvalidAccountId
- [x] vanillaTransferSucceeds
- [x] https://github.com/hashgraph/hedera-services/issues/9345

mhess-swl:
Some of these tests depend on #9286 (e.g. `complexKeyAcctPaysForOwnTransfer()`)

mhess-swl:
`allowanceTransfersWorkAsExpected` depends on #9321

Neeharika-Sompalli:
Closing this issue . Remaining tests are in this issue https://github.com/hashgraph/hedera-services/issues/9345

PR #9171, opened 2023-10-10
Add checks for immutable keys

mhess-swl:
Part of #8466.

This fixes one test in the crypto transfer suite, testing various attempts at using immutable accounts.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
@@ -16,13 +16,14 @@
 
 package com.hedera.node.app.service.token.impl.handlers;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_AUTORENEW_ACCOUNT;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CUSTOM_FEE_COLLECTOR;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TRANSACTION_BODY;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TREASURY_ACCOUNT_FOR_TOKEN;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.MAX_ENTITIES_IN_PRICE_REGIME_HAVE_BEEN_CREATED;
 import static com.hedera.node.app.hapi.fees.usage.crypto.CryptoOpsUsage.txnEstimateFactory;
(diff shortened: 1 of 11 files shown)


hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenCreateHandleParityTest.java
```diff
@@ -137,7 +137,7 @@ void tokenCreateMissingTreasuryKey() throws PreCheckException {
         final var txn = txnFrom(TOKEN_CREATE_WITH_MISSING_TREASURY);
 
         final var context = new FakePreHandleContext(accountStore, txn);
-        assertThrowsPreCheck(() -> subject.preHandle(context), INVALID_TREASURY_ACCOUNT_FOR_TOKEN);
+        assertThrowsPreCheck(() -> subject.preHandle(context), INVALID_ACCOUNT_ID);
```
mhess-swl, line 140: Looking at `SigRequirementsTest`, the missing treasury test [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/sigs/order/SigRequirementsTest.java#L3588) shows that `INVALID_TREASURY_ACCOUNT_FOR_TOKEN` is not the correct error code. This test was ported to `TokenCreateHandleParityTest` incorrectly, and is corrected here.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -319,6 +318,15 @@ private void addCustomFeeCollectorKeys(
         for (final var customFee : customFeesList) {
             final var collector = customFee.feeCollectorAccountIdOrElse(AccountID.DEFAULT);
 
+            // Verify that the collector either has a non-empty alias OR a mutable key
+            final var acctStore = context.createStore(ReadableAccountStore.class);
+            final var collectorAcct = acctStore.getAccountById(collector);
+            if (collectorAcct != null
```
tinker-michaelj, line 324: If this is necessary, it suggests we need to circle back to how we are adding hollow account signing requirements. 🤔 But probably better to do that in a dedicated issue.

mhess-swl: Sure, I'll open a new issue. For this I followed the mono code as best I could, but there could well be a better way to work the signing requirements.

Neeharika-Sompalli: Agree with @tinker-michaelj . There were some mono-service enhancements happening when implementing this. It is possible we missed some signing requirements for hollow accounts.

mhess-swl: Opened #9195 to address this

PR #9181, opened 2023-10-11
Resolve one more test in `TokenTransactSpec`

georgi-l95:
**Description**:
This PR fixes one more test in the TokenTransactSpec by enabling autoAssociation for contacts

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


tinker-michaelj (dismissed):
Mildly annoying that we can't just stick an `overriding("contracts.allowAutoAssociations", "true")` here since `TokenTransactSpecs` runs concurrently; but all of mono-service CI runs with this flag enabled anyways, so. 🤷

kimbor (approved):

hedera-node/configuration/dev/application.properties
```diff
@@ -3,3 +3,4 @@ ledger.autoRenewPeriod.minDuration=10
 ledger.autoRenewPeriod.maxDuration=1000000000
 upgrade.artifacts.path=data/upgrade
 contracts.chainId=298
+contracts.allowAutoAssociations=true
```
netopyr, line 6: Is it possible to specify this differently, maybe directly in the test? Getting approval for this may take some time.

georgi-l95: I think it's possible by overriding in the tests, we have to do it for each test that needs it, but also it says that only your approval will be sufficient.

mhess-swl (approved):

mhess-swl (dismissed):

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/transfer/AutoAccountCreatorTest.java
```diff
@@ -83,8 +88,8 @@ void happyPathECKeyAliasWorks() {
 
         assertThat(writableAccountStore.sizeOfAliasesState()).isEqualTo(2);
         assertThat(writableAccountStore.modifiedAccountsInState()).isEmpty();
-        assertThat(writableAccountStore.get(asAccount(hbarReceiver))).isNull();
-        assertThat(writableAccountStore.get(asAccount(tokenReceiver))).isNull();
+        assertThat(writableAccountStore.get(asAccount(hbarReceiver))).isNotNull();
+        assertThat(writableAccountStore.get(asAccount(tokenReceiver))).isNotNull();
```
mhess-swl, line 92: @Neeharika-Sompalli this change (and the same change in other places in this file) looks right – seems we would want to verify that the account is not null – but just want to check this is correct since this is a test change?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/UtilScalePricingCheck.java
```diff
@@ -54,7 +53,7 @@ public List<HapiSpec> getSpecsInSuite() {
         return List.of(nftPriceScalesWithUtilization());
     }
 
-    @HapiTest
+    // Can only be run against an isolated network since it assumes the state begins with 0 NFTs minted
```
mhess-swl, line 56: Is there any action item for this test, or do we just leave it?

tinker-michaelj:
Superceded by https://github.com/hashgraph/hedera-services/pull/9227

PR #9186, opened 2023-10-11
Fix noisy log warnings

MiroslavGatsanoga:
**Description**:

The warning logs `"There are 5 active throttles, but 0 usage snapshots from saved state. Not performing a reset!"` should be occurring only if what we have in memory for the throttles differs from what we have in state e.g. at genesis. But that would be possible for the first handle transaction because after handling it we should [save what we have in memory to state](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L437) and for any subsequent transactions we should not be seeing such logs.

It looked like we didn't see the writes to state because we should have used a  `SavepointStack` instead of `HederaState` instance in `HandleWorkflow` ([first commit](https://github.com/hashgraph/hedera-services/pull/9186/commits/9e00360492c9e6352be037db36ea5a1d6a2a6e20) fixes that).

After that the number of such warning logs is reduced dramatically but we still will see the single log for the first transaction after genesis. In order to fix that we should be initializing the state during genesis based on the throttle configuration (second commit addresses that [here](https://github.com/hashgraph/hedera-services/pull/9186/commits/07e101cba1fff79a9eebef6f13be5ad80dd65a44#diff-450761d714d8e3095ec92a9186dc4aaf8ab2c56cd9d29c6e2c3f0e1097c4340fR89)).

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9080

**Notes for reviewer**:

**Checklist**


MiroslavGatsanoga:
> I'm confused by this PR, it's linked to #9080 about log flooding, but the changes mostly seem to be throttle changes??

@mhess-swl Sorry for not providing a description, updated it now based on the Slack discussion related to this issue.

mhess-swl:
> It looked like we didn't see the writes to state because we should have used a SavepointStack instead of HederaState instance in HandleWorkflow ([first commit](https://github.com/hashgraph/hedera-services/pull/9186/commits/9e00360492c9e6352be037db36ea5a1d6a2a6e20) fixes that).

@netopyr just want to check this is correct?

tinker-michaelj (approved):
Very precise @MiroslavGatsanoga , LGTM!

netopyr:
> > It looked like we didn't see the writes to state because we should have used a SavepointStack instead of HederaState instance in HandleWorkflow ([first commit](https://github.com/hashgraph/hedera-services/pull/9186/commits/9e00360492c9e6352be037db36ea5a1d6a2a6e20) fixes that).
> 
> @netopyr just want to check this is correct?

Yes, this is correct. The code in question is part of the `HandleWorkflow` where using the `SavepointStack` is the right thing to do.

PR #9193, opened 2023-10-11
Apply latest config changes to modularized code base

netopyr:
This PR ensures recent changes are also applied to config classes in modularized code.
Also, the overrides currently used in mainnet are made the defaults.

### DIFF
--- a/hedera-node/configuration/mainnet/application.properties
+++ b/hedera-node/configuration/mainnet/application.properties
@@ -3,20 +3,3 @@
 # hedera-config), this requirement will no longer be valid.
 # It's used by modular code for property overrides, taking hedera-config/ as the base, 
 # with overrides from this file (configuration/mainnet/application.properties).
-
-autoRenew.targetTypes=
-# Of historical interest only---block 33476932 was the "synchronization block" used to align the consensus
-# node block numbers with mirror node blocks in the 0.26 Services release
-contracts.knownBlockHash=055fac8eb788872957dfee45b7d40f7c94d66ea0a75ec703b993d57e441de574@33476932
-contracts.localCall.estRetBytes=4096
-hedera.allowances.isEnabled=true
-scheduling.whitelist=CryptoTransfer,ConsensusSubmitMessage,TokenBurn,TokenMint,CryptoApproveAllowance
-staking.fees.stakingRewardPercentage=10
-staking.maxStakeRewarded=650000000000000000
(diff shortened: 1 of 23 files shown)


hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/SchedulingConfig.java
```diff
@@ -30,6 +30,6 @@ public record SchedulingConfig(
         @ConfigProperty(defaultValue = "10000000") @NetworkProperty long maxNumber,
         @ConfigProperty(defaultValue = "5356800") @NetworkProperty long maxExpirationFutureSeconds,
         @ConfigProperty(defaultValue =
-                "ConsensusSubmitMessage,CryptoTransfer,TokenMint,TokenBurn,CryptoApproveAllowance")
+                "CryptoTransfer,ConsensusSubmitMessage,TokenBurn,TokenMint,CryptoApproveAllowance")
```
jsync-swirlds, line 33: Order is irrelevant here, so why change it?

hedera-node/hedera-file-service-impl/src/test/java/com/hedera/node/app/service/file/impl/test/FileTestBase.java
```diff
@@ -98,7 +98,7 @@ public class FileTestBase {
 
     protected final String beneficiaryIdStr = "0.0.3";
     protected final long paymentAmount = 1_234L;
-    protected final Bytes ledgerId = Bytes.wrap(new byte[] {3});
+    protected final Bytes ledgerId = Bytes.wrap(new byte[] {0});
```
jsync-swirlds, line 101: This is only for unit testing, and is deliberately intended to be different from config.

hedera-node/configuration/mainnet/application.properties
```diff
@@ -3,20 +3,3 @@
 # hedera-config), this requirement will no longer be valid.
 # It's used by modular code for property overrides, taking hedera-config/ as the base, 
 # with overrides from this file (configuration/mainnet/application.properties).
-
```
jsync-swirlds, line 6: Out of curiosity, what is the reasoning for removing these properties from both bootstrap and application files, but only for mainnet?

iwsimon (approved):
LGTM

Nana-EC (approved):

Neeharika-Sompalli (approved):

PR #9197, opened 2023-10-11
07672 Run Hapi Tests

jsync-swirlds:
* Fixed extremely bad exception design that resulted in massive runaway and time expansion when the hapi node fails.

Basically, we had methods declaring `Throwable` in signature, which makes correct exception handling practically impossible, so I cleaned up a few of those that impacted the runaway code path.  Then I changed the catch blocks to retry when possible, but fail completely when no amount of retry could fix the problem.  Now, instead of burning 6 hours at 100% CPU retrying connection to dead nodes, the HAPI test clients just fail all subsequent tests.

If PR #8733 revives, then a node restart can be added after that to restart the nodes when they fail, instead of just failing all the tests, but for now failing quickly is far preferrable to hours of useless retry.


iwsimon (approved):
LGTM. Catching a specific exception and handling it is the correct way!

povolev15 (approved):

PR #9215, opened 2023-10-12
Dispatch prehandle

### LINKED ISSUE #9213: Add checks to  HandleContextImpl.dispatchSyntheticTxn()
Add checks to `HandleContextImpl.dispatchSyntheticTxn()` right before we call `dispatchHandle()`

mustafauzunn:
**Description**:

PreHandleContext for dispatchSyntheticTxn

The goal of this PR is to address system contract key verification.
Example: We have wipe system contract call and we need to check the tokens wipe key and if the caller has this key to wipe the amount of the given token from the specified Hedera account. This transaction must be signed by the token's Wipe Key

**Related issue(s)**:

Fixes #9213 

**Notes for reviewer**:

_After initial review tests will be adapted_

**Checklist**


netopyr:
I am unsure which issue you address with this PR, but executing the preHandle workflow for synthetic transactions is the wrong approach. The primary purpose of the (poorly named) `preHandle()`-method is to collect keys whose signatures need to be validated asynchronously. However, child transactions do not use the general functionality for signature checks. It is up to the parent transaction to define how this is done.

PR #9222, opened 2023-10-12
Ensure SetApprovalForAll system contract matches mono behaviour.

### LINKED ISSUE #9214: Approve_operator(SetApprovalForAllPrecompile)

MrValioBg:
**Description**:

Ensure the SetApprovalForAll system contract matches mono behaviour, by adding additional checks.
Add XTests to ensure the wanted behaviour
Also add check in AllowanceValidator for the SetApprovalForAll case, where we validate the spender address correctly.

**Related issue(s)**:

Fixes #9214 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/setapproval/SetApprovalForAllCall.java
```diff
@@ -0,0 +1,75 @@
...
+                .getOutputs()
+                .encodeElements(BigInteger.valueOf(status.protoOrdinal()));
+
+        if (status != ResponseCodeEnum.SUCCESS) {
+            // This checks ensure mono behaviour
+            if (status.equals(INVALID_ALLOWANCE_SPENDER_ID)) {
+                return reversionWith(TOKEN_NOT_ASSOCIATED_TO_ACCOUNT, 0L);
+            }
+            if (status.equals(INVALID_TOKEN_ID)) {
+                return gasOnly(successResult(encodedOutput, 0L));
```
stoyanov-st, line 68: You might want to use `completionWith()` to return the status code here

stoyanov-st (approved):
LGTM

stoyanov-st:
There is a HAPI test failing. Can you check that?
Also, run spotlessApply.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/setapproval/SetApprovalForAllTranslator.java
```diff
@@ -60,8 +59,10 @@ public boolean matches(@NonNull final HtsCallAttempt attempt) {
      */
     @Override
     public HtsCall callFrom(@NonNull final HtsCallAttempt attempt) {
-        return new DispatchForResponseCodeHtsCall<>(
-                attempt, bodyForClassic(attempt), SingleTransactionRecordBuilder.class);
+        final var result = bodyForClassic(attempt);
+        // @Future remove to revert #9214 after modularization is completed
+        return new SetApprovalForAllCall(attempt, result);
+        //  return new DispatchForResponseCodeHtsCall<>(attempt, result, SingleTransactionRecordBuilder.class);
```
lukelee-sl, line 65: do we need this commented out line?

lukelee-sl (dismissed):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
```diff
@@ -181,6 +181,10 @@ private void validateNftAllowances(
             final var spenderAccount = accountStore.getAccountById(spender);
             validateNFTSpender(serialNums, spenderAccount);
 
+            if (Boolean.TRUE.equals(allowance.approvedForAll())) {
```
tinker-michaelj, line 184: Nice catch, this definitely seems right. 🙌 Was there a `@HapiTest` failing, or you just noticed this by inspection?

tinker-michaelj (approved):
LGTM!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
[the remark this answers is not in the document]
david-bakin-sl: is the logic actually (I _think_ this is equivalent, but clearer):

```
if (allowance.approvedForAll()) {
   validateNFTSpender(spenderAccount); 
} else {
   validateNFTSpender(serialNums, spenderAccount); 
}
```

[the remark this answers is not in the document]
MrValioBg: Thanks for the feedback. Unfortunately, we cannot simplify it like that because the return type is the "Boolean" wrapper class. Thus, it could produce a NullPointerException if it's unboxed and is null. I also agree that we can optimize it with an else statement. :) I've modified it to:

     if (Boolean.TRUE.equals(allowance.approvedForAll())) {
                validateNFTSpender(spenderAccount);
            } else {
                validateNFTSpender(serialNums, spenderAccount);
            }
            
            This way, we don't call validateNFTSpender(serialNums, spenderAccount); when we don't have to.

MrValioBg: Thanks! Something didn't seem right. When I gave an input with an invalid account ID, the response was always 'SUCCESS' for setApprovalForAll, but for Grant approval it was returning correctly INVALID_ALLOWANCE_SPENDER_ID. So, I inspected the flow and found out that a similar check exists in mono, but is missing here.

david-bakin-sl (approved):
LGTM

tinker-michaelj (approved):
LGTM

PR #9227, opened 2023-10-12
Simulate mono-service custom fee failure codes in mod-service

### LINKED ISSUE #9226: Align mod-service custom fee failure codes with mono-service
### Problem

The mono-service,
 - Fails on charging a custom fee to an auto-created account with `INSUFFICIENT_PAYER_BALANCE_FOR_CUSTOM_FEE`; and,
 - Fails on charging a custom fee to an existing account in an unassociated token denomination with `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT`.

We need to simulate this behavior in mod-service, but _without_ failing prematurely in the case of royalty custom fee payments that will be deducted from a NFT sender that is receiving fungible value in new token types via auto-association.

### Solution

Add special cases to mod-service to simulate this behavior.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9226 
 - Aligns custom fee failure codes for auto-created accounts and unassociated accounts to match mono-service.
 - Enables


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/UtilScalePricingCheck.java
```diff
@@ -54,7 +53,7 @@ public List<HapiSpec> getSpecsInSuite() {
         return List.of(nftPriceScalesWithUtilization());
     }
 
-    @HapiTest
+    // Can only be run against an isolated network since it assumes the state begins with 0 NFTs minted
```
mhess-swl, line 56: Does anything need to be done about this, or do we leave it as is?

mhess-swl (dismissed):

Neeharika-Sompalli (dismissed):
LGTM!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/UtilScalePricingCheck.java
tinker-michaelj: We could consider adding an `int order() default 0` (lower number is higher precedence) to the `@HapiTest` annotation and then mark this `@HapiTest(order=Integer.MIN_VALUE)`.

mhess-swl: Hmm, yeah something to think about

mhess-swl (approved):

PR #9232, opened 2023-10-13
task: Update Dockerfile to reduce failure rates in CI

jsync-swirlds:
* Updated the docker file to use a more recent LTS base image, instead of using a PPA.
* Consolidated multiple RUN commands to reduce layers in the final image.

Ideally this will end the frequent failures to build the container image for CI test actions.
Future tasks will modify this further to use a more specifically build base image in the Swirlds Labs internal container repository.

### DIFF
--- a/hedera-node/docker/Dockerfile
+++ b/hedera-node/docker/Dockerfile
@@ -4,24 +4,18 @@
 ##  * Configuration from /opt/hedera/services/config-mount; and,
 ##  * Logs at /opt/hedera/services/output; and,
 ##  * Saved states under /opt/hedera/services/output
-FROM ubuntu:20.04 AS base-runtime
+## Ideally we'd use a much lighter (perhaps CoreOS) base, but this will do for now.
+FROM ubuntu:22.04 AS base-runtime
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y dos2unix openssl libsodium23 bind9-host
-
-# JDK
-RUN apt-get install -y software-properties-common && \
-    add-apt-repository -y ppa:openjdk-r/ppa && \
-    apt-get install -y openjdk-17-jdk
-
-# Services runtime
-RUN mkdir -p /opt/hedera/services/data/lib
-RUN mkdir /opt/hedera/services/data/apps
-RUN mkdir /opt/hedera/services/data/config
-RUN mkdir /opt/hedera/services/data/saved
-RUN mkdir /opt/hedera/services/data/onboard
-RUN mkdir /opt/hedera/services/output
-RUN mkdir /opt/hedera/services/config-mount
+    apt-get install -y dos2unix openssl libsodium23 bind9-host software-properties-common openjdk-17-jdk \
+ && mkdir -p /opt/hedera/services/data/lib \
+ && mkdir /opt/hedera/services/data/apps \
+ && mkdir /opt/hedera/services/data/config \
+ && mkdir /opt/hedera/services/data/saved \
+ && mkdir /opt/hedera/services/data/onboard \
+ && mkdir /opt/hedera/services/output \
+ && mkdir /opt/hedera/services/config-mount
 
 ## Finishes by copying the Services JAR to the base runtime
 FROM base-runtime AS final-image
@@ -29,14 +23,10 @@ FROM base-runtime AS final-image
 WORKDIR /opt/hedera/services
 
 COPY start-services.sh /opt/hedera/services/start-services.sh
-
-COPY .env /opt/hedera/services
-RUN for PIECE in $(cat .env | head -1 | tr '=' ' '); do \
-  if [ "$IS_VERSION" = "true" ]; then echo $PIECE >> .VERSION ; else IS_VERSION=true; fi done
-
 COPY --from=services-data lib /opt/hedera/services/data/lib
-RUN ls -al /opt/hedera/services/data/lib
-COPY --from=services-data onboard/StartUpAccount.txt /opt/hedera/services/data/onboard
 COPY --from=services-data apps /opt/hedera/services/data/apps
-RUN dos2unix start-services.sh
+
+RUN ls -al /opt/hedera/services/data/lib \
+ && dos2unix start-services.sh
+
 CMD ["/bin/sh", "-c", "./start-services.sh"]


hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+FROM ${IMAGE_PREFIX}network-node-base:${IMAGE_TAG} AS final-image
+
+# Create services runtime locations and pre-touch logs
+RUN mkdir -p /opt/hgcapp/services-hedera/HapiApp2.0/data/lib \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/apps \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/config \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/saved \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/onboard \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/output \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/config-mount \
```
nathanklick, line 18: Drop these as base image already creates the folder structure. Config mount is not necessary either.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+ && mkdir -p "/opt/hgcapp/services-hedera/HapiApp2.0/output" \
+ && touch "/opt/hgcapp/services-hedera/HapiApp2.0/stdout.log" \
+
+WORKDIR /opt/hgcapp/services-hedera/HapiApp2.0
+
+# Add SDK components
+ADD sdk/data/apps/* /opt/hgcapp/services-hedera/HapiApp2.0/data/apps/
+ADD sdk/data/lib/* /opt/hgcapp/services-hedera/HapiApp2.0/data/lib/
```
nathanklick, line 26: The source directory needs to be changed to the correct paths.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+
+# Copy configuration and scripts
+COPY --from=services-data onboard/StartUpAccount.txt /opt/hgcapp/services-hedera/HapiApp2.0/data/onboard
+COPY start-services.sh /opt/hgcapp/services-hedera/HapiApp2.0/start-services.sh
```
nathanklick, line 30: this should be switched to use the local node `entrypoint.sh`

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+COPY .env /opt/hgcapp/services-hedera/HapiApp2.0/
```
nathanklick, line 31: The `.env` file is completely unnecessary.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
 
-COPY --from=services-data lib /opt/hedera/services/data/lib
-RUN ls -al /opt/hedera/services/data/lib
-COPY --from=services-data onboard/StartUpAccount.txt /opt/hedera/services/data/onboard
-COPY --from=services-data apps /opt/hedera/services/data/apps
-RUN dos2unix start-services.sh
+# Set Final Working Directory, User, and Entrypoint
+USER 2000
+WORKDIR "/opt/hgcapp"
 CMD ["/bin/sh", "-c", "./start-services.sh"]
```
nathanklick, line 48: This should be matched to the local node image entrypoint

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+# Copy configuration and scripts
+COPY --from=services-data onboard/StartUpAccount.txt /opt/hgcapp/services-hedera/HapiApp2.0/data/onboard
+COPY start-services.sh /opt/hgcapp/services-hedera/HapiApp2.0/start-services.sh
+COPY .env /opt/hgcapp/services-hedera/HapiApp2.0/
+
+# Ensure proper file permissions, fill in version file, list libraries, and
+# ensure the start script has correct line endings
 RUN for PIECE in $(cat .env | head -1 | tr '=' ' '); do \
-  if [ "$IS_VERSION" = "true" ]; then echo $PIECE >> .VERSION ; else IS_VERSION=true; fi done
+  if [ "$IS_VERSION" = "true" ]; then echo $PIECE >> .VERSION ; else IS_VERSION=true; fi done \
```
nathanklick, line 36: These lines are unnecessary, the `VERSION` file is only to satisfy devops requirements in the production deployments and have no merit being included in the docker container image.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+ && ls -al /opt/hgcapp/services-hedera/HapiApp2.0/data/lib \
+ && chown -R 2000:2000 /opt/hgcapp/services-hedera/HapiApp2.0 \
+ && chmod -R +x /opt/hgcapp/services-hedera/HapiApp2.0/entrypoint.sh \
+ && dos2unix start-services.sh
```
nathanklick, line 40: This line is unnecessary when using the script.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+ && touch "/opt/hgcapp/services-hedera/HapiApp2.0/stdout.log" \
+
+WORKDIR /opt/hgcapp/services-hedera/HapiApp2.0
+
+# Add SDK components
+ADD sdk/data/apps/* /opt/hgcapp/services-hedera/HapiApp2.0/data/apps/
+ADD sdk/data/lib/* /opt/hgcapp/services-hedera/HapiApp2.0/data/lib/
+
+# Copy configuration and scripts
+COPY --from=services-data onboard/StartUpAccount.txt /opt/hgcapp/services-hedera/HapiApp2.0/data/onboard
```
nathanklick, line 29: this file is no longer used by the services software.

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
-COPY start-services.sh /opt/hedera/services/start-services.sh
-
-COPY .env /opt/hedera/services
+##  * Configuration from /opt/hgcapp/services-hedera/HapiApp2.0/config-mount; and,
+##  * Logs at /opt/hgcapp/services-hedera/HapiApp2.0/output; and,
+##  * Saved states under /opt/hgcapp/services-hedera/HapiApp2.0/output
+ARG IMAGE_PREFIX="gcr.io/hedera-registry/"
+ARG IMAGE_TAG="0.43.0-alpha.6"
+
+FROM ${IMAGE_PREFIX}network-node-base:${IMAGE_TAG} AS final-image
```
nathanklick, line 9: the `AS` clause can be dropped since this is no longer a multi-stage build

```suggestion
FROM ${IMAGE_PREFIX}network-node-base:${IMAGE_TAG}
```

hedera-node/docker/Dockerfile
```diff
@@ -1,42 +1,48 @@
...
+# Create services runtime locations and pre-touch logs
+RUN mkdir -p /opt/hgcapp/services-hedera/HapiApp2.0/data/lib \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/apps \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/config \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/saved \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/data/onboard \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/output \
+ && mkdir /opt/hgcapp/services-hedera/HapiApp2.0/config-mount \
+ && mkdir -p "/opt/hgcapp/services-hedera/HapiApp2.0/output" \
+ && touch "/opt/hgcapp/services-hedera/HapiApp2.0/stdout.log" \
```
nathanklick, line 20: these lines are still required.

hedera-node/docker/Dockerfile
jsync-swirlds: local node entrypoint is missing critical items to run the tests.
It looks like this isn't the correct change anyway (draft PR is just for testing).
It looks like the actual problem is in a completely different place (verifying currently).

hedera-node/docker/Dockerfile
jsync-swirlds: This is half done and not at all ready for review.  It's pushed to draft PR just to test in CI.

nathanklick (approved):

PR #9234, opened 2023-10-13
Fix `TokenManagementSpecs` KYC Tests

georgi-l95:
**Description**:
This PR resolves the last two tests in the TokenManagementSpecs, by adding additional validations. 

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGrantKycToAccountHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGrantKycToAccountHandler.java
@@ -18,6 +18,8 @@ package com.hedera.node.app.service.token.impl.handlers;
 
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TOKEN_ID;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TRANSACTION_BODY;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_HAS_NO_KYC_KEY;
 import static com.hedera.node.app.hapi.fees.usage.crypto.CryptoOpsUsage.txnEstimateFactory;
 import static com.hedera.node.app.service.mono.pbj.PbjConverter.fromPbj;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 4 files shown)


tinker-michaelj (approved):
Looks exactly right to me!

Neeharika-Sompalli (approved):

Neeharika-Sompalli (approved):
LGTM !

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGrantKycToAccountHandler.java
```diff
@@ -64,9 +65,8 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         final var tokenStore = context.createStore(ReadableTokenStore.class);
         final var tokenMeta = tokenStore.getTokenMeta(op.tokenOrElse(TokenID.DEFAULT));
         if (tokenMeta == null) throw new PreCheckException(INVALID_TOKEN_ID);
-        if (tokenMeta.hasKycKey()) {
-            context.requireKey(tokenMeta.kycKey());
-        }
+        if (!tokenMeta.hasKycKey()) throw new PreCheckException(TOKEN_HAS_NO_KYC_KEY);
+        context.requireKey(tokenMeta.kycKey());
```
mhess-swl, line 69: Nit: there's a helper function that will do this:
```suggestion
        validateTruePreCheck(tokenMeta.hasKycKey(), TOKEN_HAS_NO_KYC_KEY);
        context.requireKey(tokenMeta.kycKeyOrThrow());   // using ...orThrow() gets rid of a null check
```
This way we match other code

georgi-l95: Added suggestion, although tokenmetadata record does not have kycKeyOrThrow() method, but not needed as it will fail on previous line if there is no kyc key, correct ?

Neeharika-Sompalli (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenGrantKycToAccountHandler.java
mhess-swl: ok no worries, I can never remember which ones have `OrThrow`

mhess-swl (approved):

PR #9239, opened 2023-10-13
Use dummy transaction instead of empty bytes for accessor

MiroslavGatsanoga:
**Description**:

Use dummy transaction instead of empty bytes for creating `SignedTxnAccessor` in `MonoMultiplierSources` in order to stop redundant throwing & catching of exceptions [here](https://github.com/hashgraph/hedera-services/blob/0c76929438ec49a46a4c58115ed130afacda68d7/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/utils/MiscUtils.java#L260).

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/congestion/MonoMultiplierSources.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/congestion/MonoMultiplierSources.java
@@ -31,6 +31,29 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.time.Instant;
 
 public class MonoMultiplierSources {
+    private static final TxnAccessor accessor;
+
+    static {
+        try {
+            // only accessor.congestionExempt() is used in the mono multiplier implementation and that seems like could
+            // be only set for triggered transactions
+            // so using just a dummy accessor
+            accessor = SignedTxnAccessor.from(Transaction.PROTOBUF
+                    .toBytes(Transaction.newBuilder()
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/congestion/MonoMultiplierSources.java
```diff
@@ -45,7 +48,17 @@ public void updateMultiplier(@NonNull final Instant consensusTime) {
...
             // be only set for triggered transactions
             // so using just a dummy accessor
-            accessor = SignedTxnAccessor.from(Bytes.EMPTY.toByteArray());
+            final var dummySignedTxn = SignedTransaction.newBuilder()
+                    .bodyBytes(TransactionBody.PROTOBUF.toBytes(TransactionBody.newBuilder()
+                            .cryptoTransfer(
+                                    CryptoTransferTransactionBody.newBuilder().build())
+                            .build()))
+                    .build();
+            final var dummyTxn = Transaction.newBuilder()
```
netopyr, line 57: I believe this can be a static property, and we do not need to generate it for each call.

netopyr: Ah, maybe better not. `SingedTxnAccessor` is mutable. I do not think it matters here, but it always feels weird to create a static member that is mutable... 🤔

MiroslavGatsanoga: Updated.

rbair23 (approved):

PR #9241, opened 2023-10-13
08670 Non linking orphan buffer

### LINKED ISSUE #8670: Non-linking orphan buffer
Create an implementation of the orphan buffer that does not link events together and that does not rely on the shadow graph. The purpose of this is to further disentangle the shadowgraph from other steps in the intake pipeline.

alittley:
- closes #8670


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/orphan/OrphanBuffer.java
```diff
@@ -0,0 +1,254 @@
...
+            if (nonOrphan.getGeneration() < minimumGenerationNonAncient) {
+                // Although it doesn't cause harm to pass along ancient events, it is unnecessary to do so.
+                intakeEventCounter.eventExitedIntakePipeline(event.getSenderId());
+                currentOrphanCount--;
+                continue;
+            }
+
+            eventConsumer.accept(nonOrphan);
+            eventsWithParents.add(nonOrphanDescriptor);
+            currentOrphanCount--;
```
cody-littley, line 227: If you move `currentOrphanCount--` to the top of the while loop and remove it from the current locations, the code is logically equivalent I think.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/orphan/OrphanBuffer.java
```diff
@@ -0,0 +1,254 @@
...
+     * <p>
+     * Accounts for events potentially becoming un-orphaned as a result of this event not being an orphan.
+     *
+     * @param event the event that is not an orphan
+     */
+    private void eventIsNotAnOrphan(@NonNull final GossipEvent event) {
+        final Deque<GossipEvent> nonOrphanStack = new LinkedList<>();
+        nonOrphanStack.push(event);
+
+        while (!nonOrphanStack.isEmpty()) {
```
cody-littley, line 214: I suggest adding a comment something like this for future engineers who might lack some context:

```
// When a missing parent is found, there may be many descendants of that parent who end up
// being un-orphaned. This loop frees all such orphans non-recursively (recursion yields pretty
// code but can thrash the stack).
```

cody-littley (approved):

PR #9247, opened 2023-10-13
Write standalone event deduplicator

### LINKED ISSUE #9196: Event deduplication without validation
When discussing dynamic address book design, I realized that we will need to alter the way we do event deduplication.

Currently, validation and deduplication is tightly coupled. We keep a set of the hashes of all received valid events. When we get a new event, we check if the hash is in the set, and discard it if it is already present. Then we validate the event, then if valid we add the hash to the set.

The problem arises when we receive "time traveling" events, i.e. events that were created with an address book in our future. We can't validate the event until local consensus advances and we get the appropriate address book. But it's still important to deduplicate as soon as the data comes in over the wire.

The solution to this problem is to maintain a set containing the tuple `(event hash, event signature)`, and discarding events where we already have a `(hash, signature)` pair matching the new event.

alittley:
- closes #9196


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/deduplication/EventDeduplicator.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+
+/**
+ * Deduplicates events.
+ * <p>
+ * A duplicate event is defined as an event with an identical descriptor, and identical signature to an event that has
+ * already been observed.
```
cody-littley, line 37: Nitpick. In general I think a list of two things doesn't deserve a comma in the middle (although technically speaking it's not grammatically incorrect).

```suggestion
 * A duplicate event is defined as an event with an identical descriptor and identical signature to an event that has
 * already been observed.
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/deduplication/EventDeduplicator.java
```diff
@@ -0,0 +1,137 @@
...
+
+        // a duplicate descriptor alone isn't sufficient to constitute a duplicate event
+        // we still need to check the signature
+        final boolean duplicateDescriptor = observedEvents.containsKey(eventDescriptor);
+
+        if (!duplicateDescriptor) {
+            observedEvents.put(eventDescriptor, new HashSet<>());
+        }
+
+        if (observedEvents.get(eventDescriptor).add(event.getUnhashedData().getSignature())) {
```
cody-littley, line 113: Define this constant (to avoid the creation of a bunch of lambdas):

```
private static final Function<EventDescriptor, Set<byte[]>> BUILD_NEW_HASH_SET = k -> new HashSet<>();
```

Then, you can do the following:

```
final Set<byte[]> signatures = observedEvents.computeIfAbsent(eventDescriptor, BUILD_NEW_HASH_SET);

if (signatures.isEmpty() {
   // event descriptor is unique
} else {
    if (signatures.add(event.getUnhashedData().getSignature()) {
       // signature is unique, descriptor is not
    } else {
       // duplicate descriptor and signature
    }
}
```

This ends up only using a single map operation, as opposed to 3.

cody-littley (commented):
As a part of this PR, I suggest creating an async version as well. This way it's fully ready to wire up in a follow up PR.

edward-swirldslabs (approved):

cody-littley (approved):

PR #9251, opened 2023-10-13
Fix unreliable behavior in Unit Test

jsync-swirlds:
ScheduleSignHandlerTest `handleExecutesImmediateIfPossible` was not behaving deterministically.  This was traced to a HashSet that was allowing duplicate PBJ values, somehow, for reference-equal objects.  The set implementation was changed to LinkedHashSet, and appears to be behaving correctly now.


hedera-node/hedera-schedule-service-impl/src/test/java/com/hedera/node/app/service/schedule/impl/handlers/ScheduleSignHandlerTest.java
mhess-swl, line 1: Are we able to add a test to verify the scenario this fixes doesn't happen again?

jsync-swirlds: Not really, as it isn't a problem in the code, it's a problem in the *test*.
Also, it's not deterministic, so any test to verify that the test is correct would fail sometimes.

mhess-swl (approved):

Issue #9244, opened 2023-10-13
Address PR review comments for #8374


tinker-michaelj:
### Problem

There were a few final tweaks to address with the frontend throttle implementation done in #8374 ; but all checks were passing so it seemed advisable to merge that PR immediately.

### Solution

- [ ] Replace uses of `Pair` with a dedicated `record ScaledThrottle(DeterministicThrottle throttle, int logicalCapacityRequired)` type. (Or similar.)
- [x] Update `ThrottleAccumulator` query throttling to detect exempt payer accounts; re-enable `callLocalDoesNotCheckSignaturesNorPayer()`.
- [x] Inject a `string throttleType` into `ThrottleAccumulator` so we can log what type of throttle is being re-created, etc.
- [x] Remove misleading `@Singleton` annotation from `ThrottleAcumulator`.
- [x] https://github.com/hashgraph/hedera-services/pull/8374#discussion_r1358228339 - address this comment

PR #8374, opened 2023-09-05
Implement frontend throttle

vtronkov:
**Description**:
Implementing the FE throttle

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8355

**Notes for reviewer**:
* Branching off of the BE throttle story
* Reusing the throttle login from it
* Wrapping it into a HapiThrottle class and using _synchronize_ for the methods
* Deleting not needed classes

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #9266, opened 2023-10-16
Address PR review comments for #8374

vtronkov:
**Description**:
Addressing PR comments from 8374

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9244

**Notes for reviewer**:
About _"Replace uses of Pair with a dedicated record ScaledThrottle(DeterministicThrottle throttle, int logicalCapacityRequired) type. (Or similar.)"_ - I started implementing the change but it turns out that we are re-using the `ThrottleReqsManager` from the mono. In order to refactor that I will need to refactor mono code - `ThrottleReqsManager` and  `DeterministicThrottling` classes. So I'm not sure if it would be better if I create a tech debt story to refactor that after we retire the mono code. What do you think @tinker-michaelj ?


hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -126,11 +130,25 @@ public boolean shouldThrottle(
...
-            @NonNull final HederaFunctionality queryFunction, @NonNull final Instant now, @NonNull final Query query) {
+            @NonNull final HederaFunctionality queryFunction,
+            @NonNull final Instant now,
+            @NonNull final Query query,
+            @Nullable final AccountID queryPayerId) {
         final var configuration = configProvider.getConfiguration();
+
+        if (queryPayerId == null) {
+            return false;
+        }
```
vtronkov, line 145: Is this okay? If not how can I get the **payerId** in the `QueryWorkflowImpl.handleQuery` without calling `ingestChecker.runAllChecks(...)`?

netopyr: `queryPayerId` is `null` if the query is free. There is a throttle defined for free queries. Thus, I think returning `false` is wrong. I guess we should only skip the `throttleExempt()`-check.

vtronkov: Makes sense. Fixed.

mhess-swl (commented):
LGTM, but someone who better understands throttles (e.g. @tinker-michaelj) should look at this

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -79,24 +78,29 @@
...
     private static final int UNKNOWN_NUM_IMPLICIT_CREATIONS = -1;
+
+    private EnumMap<HederaFunctionality, ThrottleReqsManager> functionReqs = new EnumMap<>(HederaFunctionality.class);
     private boolean lastTxnWasGasThrottled;
     private GasLimitDeterministicThrottle gasThrottle;
     private List<DeterministicThrottle> activeThrottles = Collections.emptyList();
 
+    private final ConfigProvider configProvider;
+    private final IntSupplier capacitySplitSource;
+    private final String throttleType;
```
netopyr, line 95: I think it is better to use the enum here.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -531,7 +542,8 @@ public void applyGasConfig() {
         gasThrottle = new GasLimitDeterministicThrottle(capacity);
 
         log.info(
-                "Resolved consensus gas throttle -\n {} gas/sec (throttling {})",
+                "Resolved {} gas throttle -\n {} gas/sec (throttling {})",
+                throttleType,
```
netopyr, line 546: Is this the only place we use the `ThrottleType.type()`? If so, I think we can remove it and use the enum's name directly. It is just us who are reading the logs.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -758,9 +760,9 @@ private void restart(
         logger.info("Initializing ThrottleManager");
         this.throttleManager = new ThrottleManager();
 
-        this.backendThrottle = new ThrottleAccumulator(() -> 1, configProvider);
-        this.frontendThrottle =
-                new ThrottleAccumulator(() -> platform.getAddressBook().getSize(), configProvider);
+        this.backendThrottle = new ThrottleAccumulator(() -> 1, configProvider, BACKEND_THROTTLE.type());
```
netopyr, line 763: `() -> 1` can be `static`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -543,7 +555,8 @@ private ThrottleGroup<HederaFunctionality> hapiGroupFromPbj(
     }
 
     private void logResolvedDefinitions(final int capacitySplit) {
-        var sb = new StringBuilder("Resolved handle throttles (after splitting capacity ")
+        var sb = new StringBuilder("Resolved " + throttleType + " ")
```
netopyr, line 558: Concatenating a String and passing the result to a StringBuilder is inefficient. It is better to use `append()` already here.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
vtronkov: We use it for another log as well but it makes sense to use the enum's name, yes 👍

tinker-michaelj (approved):
LGTM!

tinker-michaelj:
>In order to refactor that I will need to refactor mono code - ThrottleReqsManager and DeterministicThrottling classes. So I'm not sure if it would be better if I create a tech debt story to refactor that after we retire the mono code. What do you think @tinker-michaelj ?

Ah I missed that, I think we can leave as written here, ty @vtronkov !

mhess-swl (approved):

PR #9298, opened 2023-10-17
Ensure Wipe operations match mono behaviour

### LINKED ISSUE #9272: WipeAmount(WipeFungiblePrecompile)

MrValioBg:
**Description**:

Implement additional XTests to ensure mono behaviour.

**Related issue(s)**:

Fixes #9272 #9273 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/wipe/WipeCall.java
```diff
@@ -0,0 +1,76 @@
...
+import com.hedera.hapi.node.transaction.TransactionBody;
+import com.hedera.node.app.service.contract.impl.exec.scope.VerificationStrategy;
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.AbstractHtsCall;
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.HtsCallAttempt;
+import com.hedera.node.app.spi.workflows.record.SingleTransactionRecordBuilder;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.math.BigInteger;
+
+// @Future remove to revert #9272 after modularization is completed
+public class WipeCall extends AbstractHtsCall {
```
stoyanov-st, line 37: I think that the default `DispatchForResponseCodeHtsCall` behaviour covers the mono responses so you don't need to implement the `HtsCall`.

MrValioBg: You are absolutely right. Thanks for spotting this.

stoyanov-st (approved):
LGTM!

lukelee-sl (approved):
LGTM

tinker-michaelj (approved):
LGTM!

david-bakin-sl (approved):
Two comments to consider:

1) In `WipeXTest`, consider breaking `doScenarioOperations` into two sub-methods, one for fungible only and the other for NFT only.
2) Also consider running all the fungible tests for _both_ V1 and V2 signatures.  (There should of course be no difference, but why not, and the current scheme of sometimes picking the V1 signature and sometimes the V2 signature isn't really explainable.)

MrValioBg:
> Two comments to consider:
> 
> 1. In `WipeXTest`, consider breaking `doScenarioOperations` into two sub-methods, one for fungible only and the other for NFT only.
> 2. Also consider running all the fungible tests for _both_ V1 and V2 signatures.  (There should of course be no difference, but why not, and the current scheme of sometimes picking the V1 signature and sometimes the V2 signature isn't really explainable.)

@david-bakin-sl. Thank you for your feedback! I like the idea and I agree that these changes would enhance clarity. As I'm also working on adding key verification tests for wipe operations, I will incorporate your suggestions into the XTest :) .

PR #9308, opened 2023-10-17
Validations for dispatchSyntheticTxn

### LINKED ISSUE #9213: Add checks to  HandleContextImpl.dispatchSyntheticTxn()

mustafauzunn:
**Description**:

Add checks to HandleContextImpl.dispatchSyntheticTxn() right before we call dispatchHandle()

**Related issue(s)**:

Fixes #9213 

**Notes for reviewer**:

**Checklist**


Neeharika-Sompalli (commented):
Some unit tests and HAPI tests are failing on this PR 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -612,7 +612,7 @@ private ValidationResult validate(
         try {
             transactionKeys = allKeysForTransaction(txBody, payer);
         } catch (PreCheckException e) {
-            return new ValidationResult(PRE_HANDLE_FAILURE, KEY_NOT_PROVIDED);
+            return new ValidationResult(PRE_HANDLE_FAILURE, e.responseCode());
```
MrValioBg, line 615: Maybe we can directly use here the logic we need from allKeysForTransaction, since we will be double dispatching         `dispatcher.dispatchPureChecks(nestedTxn); ` as its called earlier in the flow.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -584,6 +597,66 @@ private void dispatchSyntheticTxn(
...
+        final var privileges = authorizer.hasPrivilegedAuthorization(payer, function, txBody);
+        if (privileges == SystemPrivilege.UNAUTHORIZED) {
+            return new ValidationResult(PRE_HANDLE_FAILURE, ResponseCodeEnum.AUTHORIZATION_FAILED);
+        }
+        if (privileges == SystemPrivilege.IMPERMISSIBLE) {
+            return new ValidationResult(PRE_HANDLE_FAILURE, ResponseCodeEnum.ENTITY_NOT_ALLOWED_TO_DELETE);
+        }
+
+        // Check all signature verifications. This will also wait, if validation is still ongoing.
+        final var payerKeyVerification = verifier.verificationFor(payerKey);
```
tinker-michaelj, line 637: We need an option here to skip verification of the payer key when dispatching EVM child transactions.

If we don't have this option, then consider:
 - A contract `0.0.C` has a cryptographic admin key `K`.
 - A user calls contract `0.0.C`, and it e.g. calls the `tokenTransfer()` system contract to transfer some tokens it controls.

Now, we do **NOT** want to require key `K` signed the top-level HAPI `ContractCall`; because authorization in this case is granted by the logic in the smart contract itself, and not a HAPI signature.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -611,6 +624,66 @@ private void dispatchSyntheticTxn(
...
+            @NonNull PreHandleResult.Status status, @NonNull ResponseCodeEnum responseCodeEnum) {}
+
+    private ValidationResult validate(
+            @NonNull final HandleContextVerifier verifier,
+            HederaFunctionality function,
+            TransactionBody txBody,
+            AccountID payer,
+            Key payerKey) {
+
+        PreHandleContextImpl preHandleContext;
```
netopyr, line 637: `HandleWorkflow.validate()` contains a few additional checks that need to be included here. Do you want to add them in a separate PR?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -578,6 +584,13 @@ private void dispatchSyntheticTxn(
                 return;
             }
         }
+
+        final var validationResult = validate(verifier, function, txBody, payer, payerKey);
```
netopyr, line 588: Are you sure these are the right parameters? I would have expected `childVerifier`, `syntheticPayer`, and `childPayerKey`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -605,12 +619,76 @@ private void dispatchSyntheticTxn(
...
+        final var privileges = authorizer.hasPrivilegedAuthorization(payer, function, txBody);
+        if (privileges == SystemPrivilege.UNAUTHORIZED) {
+            return new ValidationResult(PRE_HANDLE_FAILURE, ResponseCodeEnum.AUTHORIZATION_FAILED);
+        }
+        if (privileges == SystemPrivilege.IMPERMISSIBLE) {
+            return new ValidationResult(PRE_HANDLE_FAILURE, ResponseCodeEnum.ENTITY_NOT_ALLOWED_TO_DELETE);
+        }
+
+        // Skip payer verification when dispatching a child transaction
+        if (!txCategory.equals(CHILD)) {
```
stoyanov-st, line 666: @tinker-michaelj  is this going to be enough to skip the payer key verification?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
stoyanov-st: Which checks exactly? I'm guessing the payer status check?

netopyr: Right. If a payer is defined, we must check his status and solvency.

In addition, we need to check if the child transaction is a duplicate. I guess we also need to check the time box.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -584,6 +598,64 @@ private void dispatchSyntheticTxn(
         }
     }
 
+    private record ValidationResult(
+            @NonNull PreHandleResult.Status status, @NonNull ResponseCodeEnum responseCodeEnum) {}
+
+    private ValidationResult validate(
```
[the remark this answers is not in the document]
stoyanov-st: Noted.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
tinker-michaelj: @stoyanov-st This should work for the `ContractService`, and is good for now I think. 🤔

tinker-michaelj (approved):
Since the handle workflow will ultimately assign a guaranteed-unique `TransactionID` to a child transaction, I think the duplicate check is superfluous. But overall this makes sense to me, ty @mustafauzunn !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -599,18 +627,99 @@ private void dispatchSyntheticTxn(
...
+        preHandleContext = new PreHandleContextImpl(readableStoreFactory(), txBody, payer, configuration(), dispatcher);
+        dispatcher.dispatchPreHandle(preHandleContext);
+
+        // Check for duplicate transactions. It is perfectly normal for there to be duplicates -- it is valid for
+        // a user to intentionally submit duplicates to multiple nodes as a hedge against dishonest nodes, or for
+        // other reasons. If we find a duplicate, we *will not* execute the transaction, we will simply charge
+        // the payer (whether the payer from the transaction or the node in the event of a due diligence failure)
+        // and create an appropriate record to save in state and send to the record stream.
+        final var duplicateCheckResult = recordCache.hasDuplicate(txBody.transactionID(), nodeID);
+        if (duplicateCheckResult != NO_DUPLICATE && duplicateCheckResult != SAME_NODE) {
```
netopyr, line 664: Why is `SAME_NODE` not throwing an Exception?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -599,18 +627,99 @@ private void dispatchSyntheticTxn(
...
-        } catch (HandleException e) {
+        } catch (final HandleException e) {
             childRecordBuilder.status(e.getStatus());
             recordListBuilder.revertChildrenOf(childRecordBuilder);
         }
     }
 
+    private void validate(
+            @NonNull final KeyVerifier verifier,
+            final HederaFunctionality function,
```
netopyr, line 645: We should use different names here because these are already used for the fields.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -599,18 +627,99 @@ private void dispatchSyntheticTxn(
...
+            throw new PreCheckException(DUPLICATE_TRANSACTION);
+        }
+
+        // Check the status and solvency of the payer
+        try {
+
+            final var fee = dispatchComputeFees(txBody, payer);
+            final var payerAccount = solvencyPreCheck.getPayerAccount(readableStoreFactory(), payer);
+            solvencyPreCheck.checkSolvency(txBody, payer, functionality, payerAccount, fee, true);
+        } catch (final InsufficientServiceFeeException | InsufficientNonFeeDebitsException e) {
```
netopyr, line 674: `InsufficientServiceFeeException` and `InsufficientNonFeeDebitsException` are `PreCheckException`s. I don't think we need to catch them here.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
netopyr: Does this also work for `ScheduleService`?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
stoyanov-st: Noted

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
stoyanov-st: Noted

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
stoyanov-st: Done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
jsync-swirlds: This might help with some of the issues in schedule transactions.  Most of the time Scheduled transactions should follow a very similar pattern to a synthetic transaction; although there are some important differences (e.g. top level record instead of child, transaction ID matched to create rather than the triggering transaction, etc...).
I'll try implementing the same in the paths scheduled transactions exercise and see how it goes.

netopyr: This code path should be executed for scheduled transactions, too. (The method name is misleading.) We should not implement separate dispatch mechanisms and validations for each possible use case we will have.

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (approved):
I think we should merge and then urgently fix `XTest`'s; c.f. https://github.com/hashgraph/hedera-services/issues/9776

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
jsync-swirlds: It's not about verify, it's more about whether this will fix some issues in the schedule execution specs.
I'll follow up separately from this PR.

Neeharika-Sompalli: Agreed.. @stoyanov-st  Please change them in follow up PR as Joseph suggested.

PR #9318, opened 2023-10-17
`SystemContract` gas calculations

### LINKED ISSUE #9190: Finish gas calculation for system contracts
### Problem

We need to take advantage of the now-implemented fee calculators in the `TokenService` handlers to complete the system contract gas calculations.

### Solution

Make `dispatchComputeFees()` visible in the `HandleContext` and call it from the system contract implementations with the appropriate synthetic `TransactionBody`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9190 

1️⃣ _Core gas calculation changes_
 - Introduces a `SystemContractGasCalculator` that encapsulates what is needed to the gas requirement for a system contract call; namely,
     1. The `TinybarValues` of `gas` and `rbh` resource prices; along with the exchange rate.
     2. The `CanonicalDispatchPrices` that come from the [_canonical-prices.json_](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hapi-fees/src/main/resources/canonical-prices.json#L1) and are used to set minimum gas requirements for the system contracts.
     3. A fee calculator that, given a `TransactionBody` to dispatch and the synthetic payer id for that body, computes its fee in tinybar if done through HAPI.
 - Adds the in-scope `SystemContractGasCalculator` to the initial `MessageFrame` in `FrameBuilder`.
 - Updates all `HtsCall` implementations to use this gas calculator.

2️⃣ _Misc fixes_
 - Updates the burn precompiles to correctly return `newTotalSupply` along with the `ResponseCodeEnum` ordinal.
 - Fixes `ClassicViewsXTest` to override `AbstractContractXTest.configuration()` to set a custom `ledger.id`.
 - Updates `TokenBurnHandler` and `TokenMintHandler` to set the `newTotalSupply` in their records.


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/TinybarValuesTest.java
```diff
@@ -27,46 +26,73 @@
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
 class TinybarValuesTest {
     private static final ExchangeRate RATE_TO_USE =
```
david-bakin-sl, line 33: Please add a comment here pointing out this is where the magic number `7`, used throughout this test, comes from.  Otherwise it is somewhat mysterious.

Alternatively, perhaps `7` should be a named constant and used throughout _including_ in this line as `...newBuilder().hbarEquiv(2).centEquiv(2 * HBAR_CENT_RATIO)...`

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/TinybarValuesTest.java
```diff
@@ -27,46 +26,73 @@
...
             ExchangeRate.newBuilder().hbarEquiv(2).centEquiv(14).build();
     private static final long RBH_FEE_SCHEDULE_RICE = 77_000L;
```
david-bakin-sl, line 35: what is a `_RICE`?  Some amount of tinybar clumped together into a not-so-tiny _grain_?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/SystemContractGasCalculatorTest.java
```diff
@@ -0,0 +1,54 @@
...
+    private CanonicalDispatchPrices canonicalDispatchPrices;
+
+    private SystemContractGasCalculator subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new SystemContractGasCalculator(tinybarValues, canonicalDispatchPrices, feeCalculator);
+    }
+
+    @Test
```
david-bakin-sl, line 50: For some reason I don't like this test.  It's not wrong: you know and I know that this method`viewGasRequirement` simply returns a constant.  But seeing it right under a `setUp` that creates a `SystemContractGasCalculator` with every single argument mocked is ... I dunno, unit tests are _not_ black box testing so you can expect to know how the unit operates when reading them ... but ... hmmm.  

No suggestion, just whining again.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/CanonicalDispatchPricesTest.java
```diff
@@ -0,0 +1,57 @@
...
+import com.hedera.node.app.service.contract.impl.exec.gas.DispatchType;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
```
david-bakin-sl, line 33: Maybe a comment to indicate that this test class is very incomplete as it does not test when there _are_ canonical dispatch prices in the asset loader?  And the constructor of `CanonicalDispatchPrices` is quite complex.  

(There isn't a test annotation that indicates an "incomplete" test needing work later, is there? I don't think JUnit has one, and I also don't think we have one, though we could.)

Or the necessary tests for that constructor could be added now?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/TransactionModuleTest.java
```diff
@@ -132,6 +143,19 @@ void providesNullEthTxDataIfNotEthereumTransaction() {
...
+    void providesSystemGasContractCalculator() {
+        given(context.dispatchComputeFees(TransactionBody.DEFAULT)).willReturn(new Fees(1, 2, 3));
+        given(tinybarValues.childTransactionServiceGasPrice()).willReturn(2L);
+        given(canonicalDispatchPrices.canonicalPriceInTinycents(DispatchType.APPROVE))
+                .willReturn(66L);
+        given(tinybarValues.asTinybars(66L)).willReturn(7L);
+        final var calculator =
+                TransactionModule.provideSystemContractGasCalculator(context, canonicalDispatchPrices, tinybarValues);
+        final var result = calculator.gasRequirement(TransactionBody.DEFAULT, DispatchType.APPROVE);
+        assertEquals(4L, result);
```
david-bakin-sl, line 156: I guess the idea here is that whoever is supposed to look at this test when it fails is one who is deeply steeped in fees.  I can make a guess that we've got a `7 - (1+2)` or maybe a `7 - 3` but can't choose between them and it may be something else too.  A comment, maybe?  Or at a bare minimum choose values for the fees (`(1,2,3)`) such that there's no ambiguity about what the calculation is supposed to be, once you reverse-engineer it.

david-bakin-sl (commented):
Initial comments looking only at unit tests.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CanonicalDispatchPrices.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * A component providing the canonical prices in tinycents for each type of transaction dispatch.
+ */
+@Singleton
+public class CanonicalDispatchPrices {
+    private final Map<DispatchType, Long> pricesMap = new EnumMap<>(DispatchType.class);
+
+    @Inject
+    public CanonicalDispatchPrices(@NonNull final AssetsLoader assetsLoader) {
```
david-bakin-sl, line 44: as noted in a previous comment there's no unit test for this constructor (which would be tested indirectly via `canonicalPriceInTinycents`)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/mint/MintTranslator.java
```diff
@@ -54,8 +57,27 @@ public boolean matches(@NonNull final HtsCallAttempt attempt) {
 
     @Override
     public HtsCall callFrom(@NonNull final HtsCallAttempt attempt) {
+        final var body = bodyForClassic(attempt);
+        final var isFungibleMint = body.tokenMintOrThrow().metadata().isEmpty();
```
david-bakin-sl, line 61: It's one thing to encode semantics (e.g., the "subtype" of something) the presence-or-absence of a field as long as it stays internal to the class.  Why can't we have a proper predicate on a `TokenMintTransactionBody` that exposes fungible or non-fungible? 

I know why, it's a protobuf, and for some reason (probably lack of time) we're stuck with having _only_ the API that Google's protobufs give us.  But protobufs are a communication protocol message, not an object-oriented domain type, and it's unfortunate the inability to specify domain-specific knowledge (predicates, in this case) is leaking through to us out of the communication layer.  (In our case it is also a storage layer, but same difference.)

Can this test be extracted into a nearby method so at least _here_ it looks like a proper test against a subtype?

(Remember all those questions we had earlier this year about how to distinguish fungible and non-fungible given the _amount_ being positive, zero, or negative, and the trouble that caused?  That's what I'm thinking of.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/DispatchGasCalculator.java
```diff
@@ -0,0 +1,41 @@
...
+import com.hedera.node.app.service.contract.impl.hevm.HederaWorldUpdater.Enhancement;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+@FunctionalInterface
+public interface DispatchGasCalculator {
+    /**
+     * Given a transaction body to be dispatched, along with the {@link SystemContractGasCalculator} and
+     * {@link Enhancement} in the dispatch context, returns the gas requirement for the dispatch.
+     *
+     * <p>If it wasn't for the complexity in {@link ClassicTransfersCall}
```
lukelee-sl, line 30: finish this thought?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/wipe/WipeTranslator.java
```diff
@@ -61,8 +64,27 @@ public boolean matches(@NonNull final HtsCallAttempt attempt) {
      */
     @Override
     public HtsCall callFrom(@NonNull final HtsCallAttempt attempt) {
+        final var body = bodyForClassic(attempt);
+        final var isFungibleWipe = body.tokenWipeOrThrow().serialNumbers().isEmpty();
```
david-bakin-sl, line 68: oh here's the same thing only this time it's serial numbers.  at least _this_ seems directly tied to NFT or not, more than "has a memo".  Still, not the best way to express semantics.

david-bakin-sl (commented):
Tranche #2.  At this time I've seen the structure and burned through all the plumbing.  What's left is the meat (apparently): Burn and Transfer.  And Burn might be straightforward.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
```diff
@@ -16,6 +16,69 @@
...
+     * @param verificationStrategy the verification strategy
+     * @param gasCalculator the gas calculator
+     * @param spender the spender
+     * @param systemContractOperations the system contract operations
+     * @param invalidTokenReversion the invalid token reversion
+     * @return the result of the call
+     */
 
-public interface BurnCall extends HtsCall {}
+    // too may parameters
```
david-bakin-sl, line 55: "may" -> "many" but thanks for the comment! otherwise these sonar suppressions are completely opaque and you can't even tell when they're no longer necessary (when the code gets rewritten, they just tend to hang around)

(Plus this is one of the mildly stupid ones, and it should be configurable as to N to adapt to local coding styles for specific projects.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
```diff
@@ -16,6 +16,69 @@
...
+            return invalidTokenReversion.apply(gasCalculator.canonicalPriceInTinybars(dispatchType));
+        }
+        final var spenderId = addressIdConverter.convert(asHeadlongAddress(spender.toArrayUnsafe()));
+        final var body = bodySupplier.get();
+        final var gasRequirement = gasCalculator.gasRequirement(body, dispatchType);
+        final var recordBuilder =
+                systemContractOperations.dispatch(body, verificationStrategy, spenderId, TokenBurnRecordBuilder.class);
+        if (recordBuilder.status() != ResponseCodeEnum.SUCCESS) {
+            return gasOnly(revertResult(recordBuilder.status(), gasRequirement));
+        } else {
```
david-bakin-sl, line 77: (Personally I'd leave off the `else` and just put the next 4 lines at the main level of the method.  To set off the error return from the normal path. But maybe that's just me.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/DispatchType.java
```diff
@@ -0,0 +1,74 @@
...
+    GRANT_KYC(HederaFunctionality.TOKEN_GRANT_KYC_TO_ACCOUNT, DEFAULT),
+    REVOKE_KYC(HederaFunctionality.TOKEN_REVOKE_KYC_FROM_ACCOUNT, DEFAULT),
+    PAUSE(HederaFunctionality.TOKEN_PAUSE, DEFAULT),
+    UNPAUSE(HederaFunctionality.TOKEN_UNPAUSE, DEFAULT),
+    FREEZE(HederaFunctionality.TOKEN_FREEZE_ACCOUNT, DEFAULT),
+    UNFREEZE(HederaFunctionality.TOKEN_UNFREEZE_ACCOUNT, DEFAULT),
+    WIPE_FUNGIBLE(HederaFunctionality.TOKEN_ACCOUNT_WIPE, TOKEN_FUNGIBLE_COMMON),
+    WIPE_NFT(HederaFunctionality.TOKEN_ACCOUNT_WIPE, TOKEN_NON_FUNGIBLE_UNIQUE),
+    UPDATE(HederaFunctionality.TOKEN_UPDATE, DEFAULT),
+    PRNG(HederaFunctionality.UTIL_PRNG, DEFAULT);
```
lukelee-sl, line 57: technically isn't PRNG its own system contract?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
```diff
@@ -16,6 +16,69 @@
...
+            final var encodedOutput = BurnTranslator.BURN_TOKEN_V1
```
david-bakin-sl, line 78: Why do we not need to know here about the difference between V1 and V2?  The return types happen to be the same, yes, but still?

(I do see that this is just code you moved here from `FungibleBurnCall` and `NonFungibleBurnCall, but still...)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
```diff
@@ -16,6 +16,69 @@
...
+import com.hedera.node.app.service.contract.impl.exec.scope.VerificationStrategy;
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.AddressIdConverter;
 import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.HtsCall;
+import com.hedera.node.app.service.token.records.TokenBurnRecordBuilder;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.function.LongFunction;
+import java.util.function.Supplier;
+
+public interface BurnCall extends HtsCall {
```
david-bakin-sl, line 39: Just out of curiosity - now that we have default methods in interfaces how do you choose between using them, as here, vs the traditional abstract class?  (In this case this could have been an abstract class that extended `AbstractHtsCall`)

david-bakin-sl (dismissed):
Looks good, with a bunch of comments on alternate things that could have been done (optional, as usual) and some questions I had.

Oh! And I still have a compile error.  What's up with that?  I gave the checkmark having forgotten about that.  So, if it's my error: checkmark ok.  If the code doesn't compile that's bad of course.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/CanonicalDispatchPricesTest.java
tinker-michaelj: `liveAssetsContainPricesForAllDispatchTypes()` does assert that each dispatch type has a price.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
```diff
@@ -64,7 +92,10 @@ public long serviceGasPrice() {
      *
      * @return the tinybar-denominated price of a rbh for the current operation
      */
-    public long serviceRbhPrice() {
-        return asTinybars(resourcePrices.basePrices().servicedataOrThrow().rbh() / FEE_SCHEDULE_UNITS_PER_TINYCENT);
+    public long topLevelServiceRbhPrice() {
```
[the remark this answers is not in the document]
tinker-michaelj: The `rbh` cost applies to any `log` events that the top-level HIP transaction stores in its record.

The HTS child transactions won't have any such `log` events; but we _do_ need to know the price of `gas` in tinybar so we can convert the child's HAPI cost from tinybar into `gas`.

(You may wonder why we have an intermediate step of converting from tinycents to tinybars; why not convert the HAPI tinycent price directly into gas? The answer is just that `TransactionHandler.calculateFees()` returns fees in tinybar.)

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/TinybarValuesTest.java
tinker-michaelj: Done.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/TinybarValuesTest.java
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/TransactionModuleTest.java
tinker-michaelj: Added comments clarifying intent.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/DispatchGasCalculator.java
tinker-michaelj: Removed fragment.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/fees/ChildFeeContextImplTest.java
```diff
@@ -0,0 +1,134 @@
...
+                        0,
+                        0,
+                        NOW,
+                        SubType.TOKEN_FUNGIBLE_COMMON_WITH_CUSTOM_FEES))
+                .willReturn(feeCalculator);
+        assertSame(feeCalculator, subject.feeCalculator(SubType.TOKEN_FUNGIBLE_COMMON_WITH_CUSTOM_FEES));
+    }
+
+    @Test
+    void propagatesInvalidBodyAsIse() {
```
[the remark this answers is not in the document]
tinker-michaelj: `propagatesInvalidBodyAsIllegalStateException()`

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/HandleContextImplTest.java
```diff
@@ -581,6 +587,38 @@ void exposesGivenNetworkInfo() {
...
+            when(stack.createWritableStates(TokenService.NAME))
+                    .thenReturn(MapWritableStates.builder()
+                            .state(MapWritableKVState.builder("ACCOUNTS").build())
+                            .state(MapWritableKVState.builder("ALIASES").build())
+                            .build());
+
+            context = createContext(defaultTransactionBody());
+        }
+
+        @SuppressWarnings("ConstantConditions")
```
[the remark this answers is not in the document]
tinker-michaelj: Dunno. 😵‍💫

hedera-node/hedera-app/src/xtest/java/contract/AbstractContractXTest.java
```diff
@@ -212,7 +212,8 @@ private void runHtsCallAndExpect(
         final var context = component.txnContextFactory().apply(PLACEHOLDER_CALL_BODY);
         final var tinybarValues = new TinybarValues(
                 context.exchangeRateInfo().activeRate(Instant.now()),
-                context.resourcePricesFor(HederaFunctionality.CONTRACT_CALL, SubType.DEFAULT));
+                context.resourcePricesFor(HederaFunctionality.CONTRACT_CALL, SubType.DEFAULT),
+                contractCallResourcePrices);
```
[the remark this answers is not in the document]
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
```diff
@@ -22,18 +22,26 @@
...
+    @Nullable
+    private final FunctionalityResourcePrices childTransactionResourcePrices;
 
     public TinybarValues(
-            @NonNull final ExchangeRate exchangeRate, @NonNull final FunctionalityResourcePrices resourcePrices) {
-        this.exchangeRate = exchangeRate;
-        this.resourcePrices = resourcePrices;
+            @NonNull final ExchangeRate exchangeRate,
+            @NonNull final FunctionalityResourcePrices topLevelResourcePrices,
+            @Nullable final FunctionalityResourcePrices childTransactionResourcePrices) {
```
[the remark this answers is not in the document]
tinker-michaelj: Switch to static factories and made this constructor private.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param body the transaction body to be dispatched
+     * @param minimumPriceInTinybars the minimum price in tinybars
+     * @return the gas requirement for the transaction to be dispatched
+     */
+    public long gasRequirement(@NonNull final TransactionBody body, final long minimumPriceInTinybars) {
+        final var nominalPriceInTinybars = feeCalculator.applyAsLong(body);
+        final var priceInTinybars = Math.max(minimumPriceInTinybars, nominalPriceInTinybars);
+        final var gasPrice = tinybarValues.childTransactionServiceGasPrice();
+        final var gasRequirement = (priceInTinybars + gasPrice - 1) / gasPrice;
```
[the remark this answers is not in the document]
tinker-michaelj: Added comment.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CanonicalDispatchPrices.java
tinker-michaelj: I believe it does exist now.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/DispatchType.java
tinker-michaelj: True, deleted this pointless `enum` value.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
tinker-michaelj: Done.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
tinker-michaelj: In this case since the parent `interface` already existed I extracted to here.

Not sure how much it helped to eliminate this duplication, but Sonar was complaining. 🤷

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenBurnHandler.java
```diff
@@ -150,6 +152,7 @@ public void handle(@NonNull final HandleContext context) throws HandleException
 
             // Remove the nft objects
             nftSerialNums.forEach(serialNum -> nftStore.remove(tokenId, serialNum));
+            context.recordBuilder(TokenBurnRecordBuilder.class).newTotalSupply(newTotalSupply);
```
Neeharika-Sompalli, line 155: 🙌

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
tinker-michaelj: Fixed. 👍

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/totalsupply/TotalSupplyCall.java
```diff
@@ -31,8 +32,11 @@
  */
 public class TotalSupplyCall extends AbstractRevertibleTokenViewCall {
 
-    public TotalSupplyCall(@NonNull final HederaWorldUpdater.Enhancement enhancement, @Nullable final Token token) {
-        super(enhancement, token);
+    public TotalSupplyCall(
+            @NonNull final SystemContractGasCalculator gasCalculator,
+            @NonNull final HederaWorldUpdater.Enhancement enhancement,
+            @Nullable final Token token) {
+        super(gasCalculator, enhancement, token);
```
david-bakin-sl, line 39: Gas `0L` at L49 ok because viewing?

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -418,6 +419,19 @@ <T> T dispatchChildTransaction(
...
+     *
+     * <p>The override payer id still matters for this purpose, because a transaction can add
+     * state whose lifetime is scoped to a payer account (the main current example is a
+     * {@link HederaFunctionality#CRYPTO_APPROVE_ALLOWANCE} dispatch).
+     *
+     * @param txBody the {@link TransactionBody} of the child transaction to compute fees for
+     * @param payerId the child payer
+     * @return the calculated fees
+     */
+    Fees dispatchComputeFees(@NonNull TransactionBody txBody, @NonNull AccountID payerId);
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Ah got it. Thank you

Neeharika-Sompalli (approved):
LGTM  as per TokenService changes! 🙌

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersCall.java
```diff
@@ -74,7 +76,7 @@ public ClassicTransfersCall(
             @Nullable ApprovalSwitchHelper approvalSwitchHelper,
             @NonNull final VerificationStrategy verificationStrategy,
             @NonNull final SystemAccountCreditScreen systemAccountCreditScreen) {
-        super(enhancement);
+        super(gasCalculator, enhancement);
```
david-bakin-sl, line 79: Commont `// TODO - gas calculation` at L95 and thus there are several `0L` gas charges in there.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/Erc721TransferFromCall.java
```diff
@@ -61,8 +62,9 @@ public Erc721TransferFromCall(
             @NonNull final VerificationStrategy verificationStrategy,
             @NonNull final org.hyperledger.besu.datatypes.Address spender,
             @NonNull final HederaWorldUpdater.Enhancement enhancement,
+            @NonNull final SystemContractGasCalculator gasCalculator,
             @NonNull final AddressIdConverter addressIdConverter) {
-        super(enhancement);
+        super(gasCalculator, enhancement);
```
david-bakin-sl, line 67: Comment `// TODO - gas calculation` at L83 and as a result there are several `0L` gas charges in there.

Maybe you just haven't done gas charges for the transfers yet, that's coming?

david-bakin-sl (commented):
(Skipping unit tests this time.)  Several locations where gas charges are `0L` and probably (maybe?) shouldn't be.  Maybe gas charges for transfers simply aren't there yet but are coming in this PR?

I do suggest making sure there are no `0L`s in this contracts directory - replace remaining ones with a proper named constant.  The problem is that a `0L` may be the actual value you want to use **or** it may be a placeholder for future code-to-be-written, but the reader can't tell them apart.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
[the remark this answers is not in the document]
david-bakin-sl: (Stupidly I didn't even see the common ` + N -1 ) / N` idiom for rounding right there in front of me, sorry.)

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/gas/CanonicalDispatchPricesTest.java
david-bakin-sl: ok, i completely read that method wrong, sorry

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/HandleContextImplTest.java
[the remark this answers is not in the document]
david-bakin-sl: sad. defeated by an IDE ... (admittedly it has about 10000 "inspections" to look through ...)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/burn/BurnCall.java
david-bakin-sl: (not actually fixed - but IDC)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -462,6 +471,21 @@ public <T> T dispatchChildTransaction(
...
+    @Override
+    public @NonNull Fees dispatchComputeFees(@NonNull final TransactionBody txBody, @NonNull final AccountID payerId) {
+        var bodyToDispatch = txBody;
+        if (!txBody.hasTransactionID()) {
+            // Legacy mono fee calculators frequently estimate an entity's lifetime using the epoch second of the
+            // transaction id/ valid start as the current consensus time; ensure those will behave sensibly here
+            bodyToDispatch = txBody.copyBuilder()
+                    .transactionID(TransactionID.newBuilder()
+                            .transactionValidStart(Timestamp.newBuilder()
+                                    .seconds(consensusNow().getEpochSecond())))
```
david-bakin-sl, line 483: Down at `AbstractContractXTest` L137 `nanos()` is set as well as `seconds()` - intentional difference?

david-bakin-sl (commented):
You pushed a new commit (or more) while I was commenting - some of these comments may be out of date - I'll check that now).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
tinker-michaelj: Entity lifetimes are only computed relative to second, but to reduce possible confusion, added `nanos()` here.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -418,6 +419,19 @@ <T> T dispatchChildTransaction(
             @NonNull Predicate<Key> callback,
             @NonNull AccountID syntheticPayer);
 
+    /**
+     * Dispatches the fee calculation for a child transaction (that might then be dispatched).
+     *
+     * <p>The override payer id still matters for this purpose, because a transaction can add
```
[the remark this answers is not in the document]
tinker-michaelj: Made all these `syntheticPayerId`.

hedera-node/hedera-app/src/xtest/java/contract/TestApproverXTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+public class TestApproverXTest extends AbstractContractXTest {
+    public static final long APPROVAL_TOKEN_BALANCE = 1000L;
+    public static final long CREATE_GAS = 5_000_000L;
+    private static final FileID TEST_APPROVER_INITCODE_FILE_ID = new FileID(0, 0, 1003);
+    private static final TupleType INPUTS = TupleType.of("address", "address");
+
+    @Override
+    protected void doScenarioOperations() {
+        handleAndCommitSingleTransaction(CONTRACT_SERVICE.handlers().contractCreateHandler(), synthCreateTxn());
```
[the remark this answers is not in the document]
tinker-michaelj: This `handleAndCommitSingleTransaction()` implicitly asserts the result is `SUCCESS`; I just needed this test to figure out what the heck was going on with gas calculation on this test. 😓

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/DispatchType.java
```diff
@@ -0,0 +1,73 @@
...
+    FREEZE(HederaFunctionality.TOKEN_FREEZE_ACCOUNT, DEFAULT),
+    UNFREEZE(HederaFunctionality.TOKEN_UNFREEZE_ACCOUNT, DEFAULT),
+    WIPE_FUNGIBLE(HederaFunctionality.TOKEN_ACCOUNT_WIPE, TOKEN_FUNGIBLE_COMMON),
+    WIPE_NFT(HederaFunctionality.TOKEN_ACCOUNT_WIPE, TOKEN_NON_FUNGIBLE_UNIQUE),
+    UPDATE(HederaFunctionality.TOKEN_UPDATE, DEFAULT);
+
+    private final HederaFunctionality functionality;
+    private final SubType subtype;
+
+    DispatchType(final HederaFunctionality functionality, final SubType subtype) {
```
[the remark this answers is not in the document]
tinker-michaelj: Added `@NonNull`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param body the transaction body to be dispatched
+     * @param dispatchType the type of dispatch whose minimum cost should be respected
+     * @param payer the payer of the transaction
+     * @return the gas requirement for the transaction to be dispatched
+     */
+    public long gasRequirement(
+            @NonNull final TransactionBody body,
+            @NonNull final DispatchType dispatchType,
+            @NonNull final AccountID payer) {
+        return gasRequirement(body, payer, canonicalPriceInTinybars(dispatchType));
```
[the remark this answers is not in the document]
tinker-michaelj: Added

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractHtsCall.java
```diff
@@ -24,18 +24,24 @@
...
 import com.hedera.node.app.service.contract.impl.exec.scope.SystemContractOperations;
 import com.hedera.node.app.service.contract.impl.hevm.HederaWorldUpdater;
 import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
- * Minimal implementation support for an {@link HtsCall} that needs an {@link HederaWorldUpdater.Enhancement}.
+ * Minimal implementation support for an {@link HtsCall} that needs an {@link HederaWorldUpdater.Enhancement}
+ * and {@link SystemContractGasCalculator}.
  */
 public abstract class AbstractHtsCall implements HtsCall {
```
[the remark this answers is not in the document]
tinker-michaelj: Added javadoc.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractHtsCall.java
```diff
@@ -24,18 +24,24 @@
...
+    protected final SystemContractGasCalculator gasCalculator;
     protected final HederaWorldUpdater.Enhancement enhancement;
 
-    protected AbstractHtsCall(@NonNull final HederaWorldUpdater.Enhancement enhancement) {
+    protected AbstractHtsCall(
+            @NonNull final SystemContractGasCalculator gasCalculator,
+            @NonNull final HederaWorldUpdater.Enhancement enhancement) {
+        this.gasCalculator = gasCalculator;
```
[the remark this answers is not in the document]
tinker-michaelj: Added.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/ClassicCreatesCall.java
```diff
@@ -45,12 +46,13 @@ public class ClassicCreatesCall extends AbstractHtsCall {
...
 
     public ClassicCreatesCall(
+            @NonNull final SystemContractGasCalculator systemContractGasCalculator,
             @NonNull final HederaWorldUpdater.Enhancement enhancement,
             @NonNull final TransactionBody syntheticCreate,
             @NonNull final VerificationStrategy verificationStrategy,
             @NonNull final org.hyperledger.besu.datatypes.Address spender,
             @NonNull final AddressIdConverter addressIdConverter) {
-        super(enhancement);
+        super(systemContractGasCalculator, enhancement);
```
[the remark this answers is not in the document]
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/getapproved/GetApprovedCall.java
```diff
@@ -55,24 +57,23 @@ public GetApprovedCall(
...
+            return revertResult(INVALID_TOKEN_NFT_SERIAL_NUMBER, gasCalculator.viewGasRequirement());
         }
         final var spenderNum = nft.spenderId().accountNumOrThrow();
         final var spender = nativeOperations().getAccount(spenderNum);
         return isErcCall
                 ? successResult(ERC_GET_APPROVED.getOutputs().encodeElements(headlongAddressOf(spender)), 0L)
                 : successResult(
                         HAPI_GET_APPROVED.getOutputs().encodeElements(SUCCESS.getNumber(), headlongAddressOf(spender)),
-                        0L);
+                        gasCalculator.viewGasRequirement());
```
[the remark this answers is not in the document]
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersCall.java
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/Erc20TransfersCall.java
```diff
@@ -69,7 +71,7 @@ public Erc20TransfersCall(
             @NonNull final VerificationStrategy verificationStrategy,
             @NonNull final org.hyperledger.besu.datatypes.Address spender,
             @NonNull final AddressIdConverter addressIdConverter) {
-        super(enhancement);
+        super(gasCalculator, enhancement);
```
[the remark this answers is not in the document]
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/Erc721TransferFromCall.java
tinker-michaelj: Fixed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersCall.java
```diff
@@ -111,7 +121,88 @@ public ClassicTransfersCall(
...
+     * Simulates the mono-service gas calculation for a classic transfer, which is significantly complicated by our
+     * current strategy for setting the minimum tinybar price based on the canonical prices of various operations.
+     *
+     * @param body the transaction body to be dispatched
+     * @param systemContractGasCalculator the gas calculator to use
+     * @param enhancement the enhancement to use
+     * @param payerId the payer of the transaction
+     * @return the gas requirement for the transaction to be dispatched
+     */
+    public static long classicTransferGasRequirement(
```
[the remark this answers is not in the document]
tinker-michaelj: Copied the comments from mono-service here.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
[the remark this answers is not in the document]
david-bakin-sl: Still missing check for `payer` here and down at L72-73 - though only really needed down there.

david-bakin-sl:
I've resolved all of my comments that have been addressed - and a couple of Luke's too - and the only comments of mine left unresolved are those I think could still be addressed.

Will continue review later today for all of the latest changes.

Regardless of my comments - which you know mainly address readability of this code _in the future_ - this is a pretty nice way to do gas calculations, which are notoriously finicky. Right up there with fee calculations.  (Hmm, there's probably a connection, huh?)

Neeharika-Sompalli (approved):
LGTM !

PR #9319, opened 2023-10-17
Allowance Transfer Fixes

mhess-swl:
This PR fixes some errors in `CryptoTransferSuite.allowanceTransfersWorkAsExpected()`. Due to an issue with determining repeated accounts, however, this test can't be enabled yet.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
@@ -21,9 +21,8 @@ import static com.hedera.hapi.node.base.ResponseCodeEnum.ACCOUNT_KYC_NOT_GRANTED
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_NFT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.NO_REMAINING_AUTOMATIC_ASSOCIATIONS;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.SENDER_DOES_NOT_OWN_NFT_SERIAL_NO;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.SPENDER_DOES_NOT_HAVE_ALLOWANCE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_NOT_ASSOCIATED_TO_ACCOUNT;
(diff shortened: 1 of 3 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoTransferSuite.java
```diff
@@ -887,7 +887,13 @@ private HapiSpec allowanceTransfersWorkAsExpected() {
                                 .payingWith(SPENDER)
                                 .signedBy(SPENDER)
                                 .fee(ONE_HBAR)
-                                .hasKnownStatus(INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE),
+                                // INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE is the expected mono service
+                                // outcome here, but the modularized implementation doesn't provide enough information
```
Neeharika-Sompalli, line 891: Are we not enabling any tests now ?

mhess-swl: No, it's blocked by the repeated accounts issue

Neeharika-Sompalli: Ah okayy! 😞

mhess-swl: Repeated accounts issue: #9321

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -88,10 +91,34 @@ public void doIn(@NonNull final TransferContext transferContext) {
...
+                    }
+                }
+
+                // If no allowance or spender is found, check the NFT's owner
+                if (!allowanceFound) {
+                    if (nft.hasOwnerId()) {
+                        validateTrue(nft.ownerIdOrThrow().equals(senderId), SENDER_DOES_NOT_OWN_NFT_SERIAL_NO);
+                    } else {
+                        // If there's no owner, this NFT's only authorized sender is the token treasury
+                        validateTrue(token.treasuryAccountId().equals(senderId), SENDER_DOES_NOT_OWN_NFT_SERIAL_NO);
```
Neeharika-Sompalli, line 120: I think this logic is already there in `NFTOwnersChangeStep`. Not sure if we need these here

mhess-swl: I refactored the code so the same check is being done in both places, without duplicating the logic. Good call! 👍

mhess-swl:
Closing in favor of #9334

PR #9323, opened 2023-10-17
09030 save all records

### LINKED ISSUE #9030: Record cache should store child records too
### Description

We found during testing on the modular code that transactions with child records did not have those child records stored in the record cache, so they were unavailable during queries.

This issue is NOT IN PRODUCTION CODE.

### Steps to reproduce

Run the AutoAccountCreationSuite. Tests will fail due to not finding enough records.

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

rbair23:
**Description**:
Captures child records and preceding records as part of the handle workflow, storing them in the record cache. This PR also fixes a few issues, such as adjusting the consensus timestamp in the same way that mono service does (moving the user transaction to -1000 + 3 ns from the consensus timestamp given by the platform). It also assigns the nonce values correctly, and the consensus timestamps of preceding and following child records.

**Related issue(s)**:

Fixes #9030


hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/HandleWorkflowTest.java
rbair23, line 1: This change is needed to the tests to account for the timestamp shenanigans done by handle workflow.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -149,13 +150,28 @@ public class SingleTransactionRecordBuilderImpl
...
 
+    // Used for some child records builders.
+    private final boolean removable;
+
     /**
      * Creates new transaction record builder.
      *
      * @param consensusNow the consensus timestamp for the transaction
      */
     public SingleTransactionRecordBuilderImpl(@NonNull final Instant consensusNow) {
```
iwsimon, line 161: SingleTransactionRecordBuilderImpl(consensusNow, false) could replace this.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SubProcessHapiTestNode.java
```diff
@@ -92,6 +92,8 @@ public void start() {
             builder.command(
                             javaCmd,
                             "-Dfile.encoding=UTF-8",
+                            "-agentlib:jdwp=transport=dt_socket,server=y,suspend=" + (nodeId == 0 ? "n" : "n")
```
jsync-swirlds, line 95: Worth noting that, similar to the three test profiles now prompting any test run in IDE, we could modify the debug profile in the build to set a property to control the suspend parameter here.
Future work, but might make debug for out-of-process a little smoother.

hedera-node/hedera-network-admin-service-impl/src/test/java/com/hedera/node/app/service/networkadmin/impl/test/handlers/NetworkAdminHandlerTestBase.java
```diff
@@ -402,15 +446,15 @@ protected void givenNonFungibleTokenRelation() {
                 .build();
     }
 
-    private TransactionID transactionID(int nonce) {
-        return transactionID(0, nonce);
+    private TransactionID transactionID(Instant validStartTime, int nonce) {
+        return transactionID(validStartTime, 0, nonce);
```
jsync-swirlds, line 450: Why not use the nanos in the Instant?
```suggestion
        return transactionID(validStartTime, validStartTime.getNano(), nonce);
```
It is a bit odd that the base overload separates them out to begin with, but at least here we can simplify that.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -250,23 +311,17 @@ public List<TransactionRecord> getRecords(@NonNull final AccountID accountID) {
         for (final var transactionID : transactionIDs) {
             final var history = histories.get(transactionID);
             if (history != null) {
-                final var recs = history.records();
+                final var recs = history.orderedRecords();
                 records.addAll(recs.size() > maxRemaining ? recs.subList(0, maxRemaining) : recs);
                 maxRemaining -= recs.size();
                 if (maxRemaining <= 0) break;
```
jsync-swirlds, line 317: Not sure if it matters, but these may not be fully ordered if there are more than the maximum limit of transactions to be returned.
The transactions never added to the `records` list could have been anywhere in the final ordering.
If we promise any kind of final ordering in the return value, then we would have to build the **full** list, sort it, *then* trim to length.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordCache.java
```diff
@@ -37,79 +45,117 @@
...
-     * the transaction.
+     * <p>The {@code childRecords} list contains a list of all child transactions of the original user transaction.
+     * Duplicate transactions never have child transactions.
      *
-     * @param accountID The accountID of the Payer of the transactions
-     * @return The {@link TransactionRecord}s, if any, or else an empty list.
+     * @param nodeIds The IDs of every node that submitted a transaction with the txId that came to consensus and was
+     *                handled. This is an unordered set, since deterministic ordering is not required for this in-memory
+     *                data structure
+     * @param records Every {@link TransactionRecord} handled for every user transaction that came to consensus
```
jsync-swirlds, line 61: Missing `@param` for childRecords...

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordCache.java
```diff
@@ -37,79 +45,117 @@
...
+         * transaction ID), but we do not yet have a record for it.
+         */
+        private static final TransactionReceipt PENDING_RECEIPT =
+                TransactionReceipt.newBuilder().status(UNKNOWN).build();
+
+        /**
+         * Create a new {@link History} instance with empty lists.
+         */
+        public History() {
+            this(new HashSet<>(), new ArrayList<>(), new ArrayList<>());
```
jsync-swirlds, line 79: Probably doesn't matter, but we *could* be a little defensive here if we used a `TreeSet` or `ConcurrentSkipListSet` to impose a sort order on the nodeIds.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordCache.java
```diff
@@ -37,79 +45,117 @@
...
+        /**
+         * Returns a list of all records, ordered by consensus timestamp. Some elements of {@link #childRecords} may
+         * come before those in {@link #records}, while some may come after some elements in {@link #records}.
+         * @return The list of all records, ordered by consensus timestamp.
+         */
+        public List<TransactionRecord> orderedRecords() {
+            final var ordered = new ArrayList<>(records);
+            ordered.addAll(childRecords);
+            ordered.sort((a, b) -> TIMESTAMP_COMPARATOR.compare(
+                    a.consensusTimestampOrElse(Timestamp.DEFAULT), // Comparator doesn't currently handle nulls
```
jsync-swirlds, line 133: Given how easy it is to write a class implementing the `Comparator` interface (it's just one method), it might be worth doing that in a follow-up PR to handle nulls properly and remove the `somethingOrElse` calls that happen in dense loops.

PR #9325, opened 2023-10-17
9225: Serialization/deserialization overhead in MerkleDb during reconnects

### LINKED ISSUE #9225: Serialization/deserialization overhead in MerkleDb during reconnects
### Problem

When a `VirtualMap` is synchronized during reconnects, on the teacher side, leaves and hashes are deserialized when read from disk only to be serialized again, albeit with different methods, before they are sent over the wire.
On the learner side, the received bytes are deserialized in order to be inserted into `VirtualMap` and serialized again, albeit with different methods, for hashing and separately for recording in `MerkleDb`.

### Solution

There are two levels of solution:
* implement a shortcut on the teacher side, so that bytes are sent over the wire in the same form as they are read from disk; the equivalence of different serde methods needs to be verified/established; this solution alone can greatly reduce memory pressure on the teacher side.
* have `VirtualMap`/`MerkleDb` work with immutable `byte[]` arrays for keys/values and move all serde up the stack to consumers.

### Alternatives

_No response_

artemananiev:
Fix summary:

* A new method is introduced to `VirtualDataSource` interface, `loadAndWriteLeafRecord()`, to find a leaf record in the database and write it to the given output stream
* By default, this method is implemented as reading the record and then serializing it to the stream. MerkleDb data source provides more efficient implementation: it reads virtual leaf record bytes and writes them to the stream, bypassing (de)serialization
* This new method is used by `VirtualTeacherTreeView`

Fixes: https://github.com/hashgraph/hedera-services/issues/9225

### DIFF
--- a/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/DataFileCollectionBench.java
+++ b/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/DataFileCollectionBench.java
@@ -21,6 +21,7 @@ import com.swirlds.merkledb.collections.LongListOffHeap;
 import com.swirlds.merkledb.files.DataFileCollection;
 import com.swirlds.merkledb.files.DataFileReader;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.List;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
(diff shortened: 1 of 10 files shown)


imalygin (approved):

artemananiev:
Further investigation showed that this idea is really hard to implement. Despite SelfSerializable and MerkleDb binary formats are the same for virtual keys and values, virtual leaf record formats are quite different. If a leaf record is loaded from disk and sent to learner as is, learner will fail to parse it. A workaround could be to partially parse the record, extract a path from it, key (as bytes, without deserialization), and value (as bytes). It would kill the whole idea, and I'm not sure if it can be implemented at all, as MerkleDb stores keys and values together with no clear boundary. This is why I'm closing this PR and the ticket.

PR #9326, opened 2023-10-18
Match `MintToken` mono behaviour

### LINKED ISSUE #9271: `mintToken` - match mono service behavior
### Problem

Match mono service behavior on incorrect input

### Solution

Match the output for `mintToken` from [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json) document for 

1. non-static context 
a. incorrect token_id INVALID_TOKEN_ID

### Alternatives

_No response_

dikel:
**Description**:
This PR adds `Mint` X test where we expect INVALID_TOKEN_ID upon passing an invalid token address

**Related issue(s)**:

Fixes #9271


agadzhalov (approved):
LGTM

petreze (approved):

agadzhalov (approved):
LGTM

rbair23 (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
```diff
@@ -282,11 +282,7 @@ private Nft buildNewlyMintedNft(
     @Override
     public Fees calculateFees(@NonNull final FeeContext feeContext) {
         final var op = feeContext.body().tokenMintOrThrow();
-        final var readableTokenStore = feeContext.readableStore(ReadableTokenStore.class);
-        final var tokenType = readableTokenStore.get(op.tokenOrThrow()).tokenType();
-        final var subType = tokenType == TokenType.FUNGIBLE_COMMON
-                ? SubType.TOKEN_FUNGIBLE_COMMON
-                : SubType.TOKEN_NON_FUNGIBLE_UNIQUE;
+        final var subType = op.amount() > 0 ? SubType.TOKEN_FUNGIBLE_COMMON : SubType.TOKEN_NON_FUNGIBLE_UNIQUE;
```
david-bakin-sl, line 285: I don't get it.  Old code was a good way to determine fungible/non-fungible: just look directly at the token type, it'll tell you.  This code goes back to the way we see all over the mono service - figuring it out indirectly from other cues, like amount > 0.  Which in the past has led to problems, which made it to production.  

Why make this change?

david-bakin-sl (commented):
Change in `TokenMintHandler` does not seem necessary or an improvement.  Please explain what I'm missing!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
agadzhalov: Because when we pass [INVALID_TOKEN_ADDRESS](https://github.com/hashgraph/hedera-services/blob/119c1cc180ab7e1ec1a0143c68dd1d49275f42b5/hedera-node/hedera-app/src/xtest/java/contract/MintsXTest.java#L128) in the X test the old code throws null pointer exception and can not get [tokenType()](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java#L288)

david-bakin-sl: Then maybe that case should be explicitly tested for and handled?  So it is _explicit_ in the code what is happening, that there's a necessary check for an invalid token address?  Rather than just papering it over so it's hidden?

Issue #9321, opened 2023-10-17
Fix Repeated Account Amount Approvals


mhess-swl:

mhess-swl:
When running `CryptoTransferSuite.allowanceTransfersWorkAsExpected()`, we get the following error:

```
2023-10-17 10:04:21.410 INFO  301  HapiTxnOp - 'AllowanceTransfersWorkAsExpected' -  HapiCryptoTransfer{sigs=1, payer=spender, node=0.0.3, transfers=[0.0.1002 -> -500000000, 0.0.1004 <- +100000000, 0.0.1004 <- +500000000, 0.0.1003 -> -100000000], tokenTransfers=} Wrong actual status ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS, expected SUCCESS
2023-10-17 10:04:21.410 WARN  262  HapiSpecOperation - 'AllowanceTransfersWorkAsExpected' - HapiCryptoTransfer{sigs=1, payer=spender, node=0.0.3, transfers=[0.0.1002 -> -500000000, 0.0.1004 <- +100000000, 0.0.1004 <- +500000000, 0.0.1003 -> -100000000], tokenTransfers=} failed - Wrong status! Expected SUCCESS, was ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS!
```

The error happens close to [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoTransferSuite.java#L969) (approximately line 969):
```java
cryptoTransfer(
            allowanceTinyBarsFromTo(OWNER, RECEIVER, 5 * ONE_HBAR),
            tinyBarsFromTo(SPENDER, RECEIVER, ONE_HBAR))
    .payingWith(SPENDER)
    .signedBy(SPENDER),
```

Instead of a successful transfer, the modularized code claims there are repeated account amounts. The [modularized code checks](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java#L69) for repeated account amounts in `CryptoTransferValidator.pureChecks(...)`, which has the following:
```java
        final var acctAmounts = op.transfersOrElse(TransferList.DEFAULT).accountAmountsOrElse(emptyList());
        final var uniqueAcctIds = new HashSet<AccountID>();
[...]

PR #9334, opened 2023-10-18
Allowance transfers fixes 

### LINKED ISSUE #9284: Fix allowance tests in CryptoTransferSuite
Fix the following tests in `CryptoTransferSuite`

- [x] allowanceTransfersWithComplexTransfersWork
- [x] allowanceTransfersWorkAsExpected

Neeharika-Sompalli:
Fixed allowance transfers tests in `CryptoTransferSuite`

Fixes #9284 
Fixes https://github.com/hashgraph/hedera-services/issues/9321

- Added missing validation that is_approval=true NFT transfer has owner's approval
- Fixed false positive detection of repeated AccountAmount's by including the is_approval value in AccountAmount "identity"

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
@@ -204,8 +204,7 @@ public class CustomFeeAssessmentStep {
             throw new IllegalStateException("Custom fee charging exceeded max recursion depth");
         }
 
-        // If there are no assessed custom fees, as long as there are hbar transfers
-        // or token transfers in the transaction, add it to the list of assessed transactions for further steps.
+        // If the last charging level assessed fees, we should include them for further steps of CryptoTransfer
         if (!hbarTransfers.isEmpty() || !tokenTransfers.isEmpty()) {
             assessedTxns.add(txnToAssess);
         }
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
@@ -204,6 +204,8 @@ public class CustomFeeAssessmentStep {
             throw new IllegalStateException("Custom fee charging exceeded max recursion depth");
         }
 
+        // If there are no assessed custom fees, as long as there are hbar transfers
+        // or token transfers in the transaction, add it to the list of assessed transactions for further steps.
         if (!hbarTransfers.isEmpty() || !tokenTransfers.isEmpty()) {
             assessedTxns.add(txnToAssess);
         }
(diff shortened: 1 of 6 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
```diff
@@ -215,13 +220,17 @@ private record CustomFeeAssessmentResult(
             List<CryptoTransferTransactionBody> assessedTxns, List<AssessedCustomFee> assessedCustomFees) {}
 
     private void validateTotalAdjustments(final CryptoTransferTransactionBody op, final int maxTransfersDepth) {
-        final var hbarTransfers = op.transfersOrElse(TransferList.DEFAULT)
-                .accountAmountsOrElse(emptyList())
+        final var hbarTransfers = op.transfersOrElse(TransferList.DEFAULT).accountAmountsOrElse(emptyList()).stream()
+                .map(AccountAmount::accountID)
+                .collect(Collectors.toSet())
```
mhess-swl, line 225: It shouldn't matter because this is a set, but will this code's implementation deterministically map to equivalent set objects every time? Or is this another subtle case where we might get an occasional ISS?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/NFTOwnersChangeStep.java
```diff
@@ -123,7 +123,7 @@ public void doIn(final TransferContext transferContext) {
      * @param tokenId token id of the nft
      * @param nft nft to be transferred
      */
-    private void validateSpenderHasAllowance(
+    public static void validateSpenderHasAllowance(
```
mhess-swl, line 126: Does this need to be `public`? I thought package private would work
```suggestion
   static void validateSpenderHasAllowance(
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
```diff
@@ -199,7 +204,7 @@ public CustomFeeAssessmentResult assessFees(
             throw new IllegalStateException("Custom fee charging exceeded max recursion depth");
         }
 
-        if (!hbarTransfers.isEmpty()) {
+        if (!hbarTransfers.isEmpty() || !tokenTransfers.isEmpty()) {
```
mhess-swl, line 207: Good catch!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java
Neeharika-Sompalli: It just needs size. So it doesn't matter if they are in order

mhess-swl: oh I missed that, nevermind thanks!

tinker-michaelj (approved):
LGTM! 🙌

PR #9351, opened 2023-10-19
Fix handling of complex keys during ingest

netopyr:
Key verification was not correctly implemented in the ingest workflow. This PR adds support for complex keys (key-lists and threshold-keys).


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
```diff
@@ -184,7 +186,7 @@ public TransactionInfo runAllChecks(
         }
 
         // 6. Verify payer's signatures
-        verifyPayerSignature(txInfo, payerKey);
+        verifyPayerSignature(txInfo, payer, payerKey, configuration);
```
Neeharika-Sompalli, line 189: nit: Do we need both payer and payerKey as inputs or just payer should be fine?

Neeharika-Sompalli (dismissed):
LGTM !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
```diff
@@ -195,43 +197,30 @@ public TransactionInfo runAllChecks(
...
+        final var hederaConfig = configuration.getConfigData(HederaConfig.class);
+        final var sigPairs = txInfo.signatureMap().sigPairOrElse(List.of());
+
         // Expand the signatures
         final var expandedSigs = new HashSet<ExpandedSignaturePair>();
-        signatureExpander.expand(payerKey, txInfo.signatureMap().sigPairOrThrow(), expandedSigs);
+        signatureExpander.expand(sigPairs, expandedSigs);
+        if (!isHollow(payer)) {
+            signatureExpander.expand(payerKey, sigPairs, expandedSigs);
+        }
```
rbair23, line 214: I'm not sure about this part either, as to why we check for hollow. In the other file we also check for null payer. It just isn't clear to me what the expectation is

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -678,12 +679,15 @@ private PreHandleResult addMissingSignatures(
 
         // prepare signature verification
         final var verifications = new HashMap<Key, SignatureVerificationFuture>();
-        final var payerKey = previousResult.payerKey();
-        verifications.put(payerKey, previousResult.verificationResults().get(payerKey));
+        final var payer = solvencyPreCheck.getPayerAccount(storeFactory, previousResult.payer());
+        final var payerKey = payer.key();
```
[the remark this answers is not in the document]
netopyr: Right. There were several issues. Reusing the payerKey from prehandle was one of them.

iwsimon (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -678,12 +679,15 @@ private PreHandleResult addMissingSignatures(
...
 
         // expand all keys
         final var expanded = new HashSet<ExpandedSignaturePair>();
         signatureExpander.expand(sigPairs, expanded);
+        if (payerKey != null && !HapiUtils.isHollow(payer)) {
+            signatureExpander.expand(payerKey, sigPairs, expanded);
+        }
```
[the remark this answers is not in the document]
netopyr: I plan to do a follow-up PR that makes sure we handle hollow accounts and invalid transactions without a payer key correctly. In this upcoming PR, I also want to remove the `payerKey` from `PreHandleResult`.

This fix is about expanding the payer key instead of using it directly, as this works only for primitive keys.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
netopyr: Same here. I will go thoroughly through this next week.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
netopyr: Good catch! Fixed.

Neeharika-Sompalli (approved):
LGTM!

rbair23 (approved):

Issue #9277, opened 2023-10-16
`createNonFungibleToken` - match mono service behavior


agadzhalov:
### Problem

Match mono service behavior on incorrect input

### Solution

Match the output for `createNonFungibleToken` from [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json) document for 

1. non-static context 
a. incorrect account_id -> `CONTRACT_REVERT_EXECUTED `

PR #9361, opened 2023-10-19
Match `TokenCreate` mono behaviour 

### LINKED ISSUE #9275: `createFungibleToken` - match mono service behavior
### Problem

Match mono service behavior on incorrect input

### Solution

Match the output for `createFungibleToken` from [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json) document for 

1. non-static context 
a. incorrect account_id -> `CONTRACT_REVERT_EXECUTED`

### Alternatives

_No response_

petreze:
**Description**:
This PR adds TokenCreate X tests to check and match the behaviour from mono. However, as we discussed, the call should revert and the proper message for the child status should be `INVALID_TREASURY_ACCOUNT_FOR_TOKEN` instead of `null` as pointed out inside the [following file](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json)

This applies for `createFungibleToken()`, `createFungibleTokenWithCustomFees()`, `createNonFungibleToken()` and `createNonFungibleTokenWithCustomFees()`

**Related issue(s)**:

Fixes #9275 #9276 #9277 #9278 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app/src/xtest/java/contract/CreatesXTest.java
```diff
@@ -257,6 +270,51 @@ protected void doScenarioOperations() {
                         .encodeCallWithArgs(DEFAULT_HEDERA_TOKEN, new Tuple[] {FIXED_FEE}, new Tuple[] {ROYALTY_FEE})
                         .array()),
                 assertSuccess());
+
+        // should revert with `INVALID_TREASURY_ACCOUNT_FOR_TOKEN` as we are passing an invalid address for the treasury
```
agadzhalov, line 274: can we shorten those comments to be one line only. Something like:

> // should revert with INVALID_TREASURY_ACCOUNT_FOR_TOKEN when passing invalid address for the treasury account

PR #9388, opened 2023-10-21
Invoke `leakCapacityForNOfUnscaled(numImplicitCreations, CryptoCreate)` on failed self-submitted auto-creation

tinker-michaelj:
**Description**:
 - Upon a failed auto-creation in `TransferLogic`, makes the node that submitted the failed auto-creation leak the reserved `CryptoCreate` capacity from its frontend throttle.

### DIFF
--- a/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
+++ b/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
@@ -251,7 +251,6 @@ class DeterministicThrottleTest {
         subject.allow(1, originalDecision);
         final var preLeakState = subject.usageSnapshot();
         subject.leakCapacity(preLeakState.used() / 2);
-        ;
         final var postLeakState = subject.usageSnapshot();
         assertEquals(preLeakState.used() / 2, postLeakState.used());
     }
--- a/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
+++ b/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
@@ -241,6 +241,22 @@ class DeterministicThrottleTest {
     }
 
     @Test
+    void canLeakSpecificAmounts() {
+        final int mtps = 333;
+        final int burstPeriod = 6;
+        final var originalDecision = Instant.ofEpochSecond(1_234_567L, 0);
+        final var subject = DeterministicThrottle.withMtpsAndBurstPeriod(mtps, burstPeriod);
+        final var capacity = subject.capacity();
+
+        subject.allow(1, originalDecision);
+        final var preLeakState = subject.usageSnapshot();
+        subject.leakCapacity(preLeakState.used() / 2);
+        ;
+        final var postLeakState = subject.usageSnapshot();
+        assertEquals(preLeakState.used() / 2, postLeakState.used());
+    }
+
+    @Test
     void returnsExpectedInstantaneousState() {
         final int mtps = 333;
         final int burstPeriod = 666;
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java
@@ -53,6 +53,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.function.IntConsumer;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.lang3.tuple.Pair;
@@ -73,6 +74,7 @@ public class TransferLogic {
     private final TransactionContext txnCtx;
     private final AliasManager aliasManager;
     private final FeeDistribution feeDistribution;
+    private final IntConsumer cryptoCreateThrottleReclaimer;
 
     @Inject
     public TransferLogic(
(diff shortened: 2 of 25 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java
```diff
@@ -139,6 +146,9 @@ public void doZeroSum(final List<BalanceChange> changes) {
                 break;
             }
         }
+        if (failedAutoCreation && txnCtx.isSelfSubmitted()) {
+            cryptoCreateThrottleReclaimer.accept(txnCtx.numImplicitCreations());
+        }
```
Neeharika-Sompalli, line 151: Makes sense. 💯

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/InfrastructureFactory.java
```diff
@@ -240,7 +240,10 @@ public TransferLogic newTransferLogic(
                 recordsHistorian,
                 txnCtx,
                 aliasManager,
-                feeDistribution);
+                feeDistribution,
+                // No-op, we won't have used any frontend throttle capacity for an auto-creation
+                // attempted during an EVM transaction
+                n -> {});
```
Neeharika-Sompalli, line 246: Why is this not possible, can it not trigger a `CryptoTransfer`? Is that because we reduce gas instead?

mhess-swl:
It's not clear to me whether this should also be implemented in the modularized code, or if the modularized code is calling through to mono throttles, and thus this functionality is included in the modularized app by default. If it's not ported to the modularized code, is there a ticket for doing so?

Neeharika-Sompalli (commented):
LGTM ! Have a question though. I think for modularized code, the only change needed is in 
`CryptoTransferHandler` since we are re-using throttles code.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
```diff
@@ -240,6 +240,22 @@ void returnsExpectedState() {
...
+        final int mtps = 333;
+        final int burstPeriod = 6;
+        final var originalDecision = Instant.ofEpochSecond(1_234_567L, 0);
+        final var subject = DeterministicThrottle.withMtpsAndBurstPeriod(mtps, burstPeriod);
+        final var capacity = subject.capacity();
+
+        subject.allow(1, originalDecision);
+        final var preLeakState = subject.usageSnapshot();
+        subject.leakCapacity(preLeakState.used() / 2);
+        ;
```
david-bakin-sl, line 254: typo

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java
david-bakin-sl, line 151: I almost think L149-151 should be moved to between L172-L173 because L168-L174 that's where the actions happen for invalid txns.  (Except for the creation of aliases which did happen up at L125 and is reverted at L171.)

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/throttling/ThrottleReqsManager.java
```diff
@@ -41,6 +42,29 @@ public boolean allReqsMetAt(Instant now, int nTransactions, ScaleFactor scaleFac
...
+     *     <li>Passes the frontend throttle check, using capacity in the buckets there; and,</li>
+     *     <li>Fails at consensus, thus <I>not</I> using network capacity for the expensive operation in the end.</li>
+     * </ol>
+     *
+     * @param nTransactions the number of transactions to undo
+     */
+    public void undoClaimedReqsFor(int nTransactions) {
+        for (int i = 0; i < passedReq.length; i++) {
+            var req = allReqs.get(i);
+            var opsRequired = req.getRight();
```
david-bakin-sl, line 62: `final`s?

david-bakin-sl (approved):
LGTM with nits.

Neeharika-Sompalli (dismissed):
LGTM , just one question. 🙌

tinker-michaelj:
> It's not clear to me whether this should also be implemented in the modularized code, or if the modularized code is calling through to mono throttles, and thus this functionality is included in the modularized app by default. If it's not ported to the modularized code, is there a ticket for doing so?

There's an issue [here](https://github.com/hashgraph/hedera-services/issues/9389); we will probably want a bit of discussion about the best place to put this. 🤔

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/InfrastructureFactory.java
tinker-michaelj: You're correct this _could_ do a `CryptoTransfer` with an auto-creation; but since the frontend `ContractCall` throttle won't have taken that into account, we don't want to refund any corresponding capacity to the frontend throttle bucket.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottleTest.java
tinker-michaelj: Deleted.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java
tinker-michaelj: Agreed, moved.

david-bakin-sl (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM ! 🙌

PR #9413, opened 2023-10-23
Enable `AutoAccountUpdateSuite` test

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/9434

Enable `AutoAccountUpdateSuite` updateKeyOnAutoCreatedAccount test. 
Fixes an issue in client side while generating signatureMaps, and add all the public keys that are needed by the transaction which may or may not have signatures for public key prefix matching.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoUpdateHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoUpdateHandler.java
@@ -409,7 +409,6 @@ public class CryptoUpdateHandler extends BaseCryptoHandler implements Transactio
                 Math.max(explicitAutoAssocSlotLifetime, oldLifetime),
                 newSlotsUsage,
                 Math.max(explicitAutoAssocSlotLifetime, newLifetime));
-        System.out.println("rbsDelta = " + rbsDelta + ", slotRbsDelta = " + slotRbsDelta + "baseSize = " + baseSize);
         return feeCalculator
                 .addBytesPerTransaction(baseSize)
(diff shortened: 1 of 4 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoUpdateHandler.java
```diff
@@ -401,17 +401,19 @@ private Fees cryptoUpdateFees(
...
-                ? op.maxAutomaticTokenAssociations() * UPDATE_SLOT_MULTIPLIER
+                ? op.maxAutomaticTokenAssociations().longValue() * UPDATE_SLOT_MULTIPLIER
                 : oldSlotsUsage;
         // If given an explicit auto-assoc slot lifetime, we use it as a lower bound for both old and new lifetimes
         final long slotRbsDelta = ESTIMATOR_UTILS.changeInBsUsage(
                 oldSlotsUsage,
                 Math.max(explicitAutoAssocSlotLifetime, oldLifetime),
                 newSlotsUsage,
                 Math.max(explicitAutoAssocSlotLifetime, newLifetime));
+        System.out.println("rbsDelta = " + rbsDelta + ", slotRbsDelta = " + slotRbsDelta + "baseSize = " + baseSize);
```
tinker-michaelj, line 412: `println` here.

Neeharika-Sompalli: Removed. Thanks 🙂

tinker-michaelj (approved):
LGTM!

iwsimon (approved):
LGTM! Thanks @Neeharika-Sompalli

PR #9416, opened 2023-10-23
Compress data sent over the network.

### LINKED ISSUE #9414: Compress gossip
Add optional gzip compression to the gossip communication channels.

cody-littley:
Closes #9414

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/SyncInputStream.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/SyncInputStream.java
@@ -57,18 +57,20 @@ public class SyncInputStream extends SerializableDataInputStream {
                 .getConfigData(SocketConfig.class)
                 .gzipCompression();
 
+        final InputStream meteredStream = extendInputStream(in, syncCounter);
+
         final InputStream wrappedStream;
         if (compress) {
             try {
-                wrappedStream = new GZIPInputStream(in, bufferSize);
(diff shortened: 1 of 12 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/SyncInputStream.java
```diff
@@ -18,59 +18,67 @@
...
+
+        final boolean compress = platformContext
+                .getConfiguration()
+                .getConfigData(SocketConfig.class)
+                .gzipCompression();
+
+        final InputStream meteredStream = extendInputStream(in, syncCounter);
+
+        final InputStream wrappedStream;
+        if (compress) {
```
edward-swirldslabs, line 63: Everyone on the network has to be configured the same way?  Either everyone uses gzip compression or everyone does not?   Is there any value in handshaking or negotiating on connection to use gzip?

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/SyncInputStream.java
cody-littley: Some settings can theoretically be varied from node to node, others can not. For example, if the rounds until ancient setting is not the same on a node it will ISS.

For this particular configuration, I don't think we need to worry about supporting varied configuration. If compression improves the performance of the network as a whole, I'd prefer not to give nodes the option of ignoring it.

PR #9417, opened 2023-10-23
09412 Create standalone in-order linker

### LINKED ISSUE #9412: Create standalone `InOrderLinker`
- create an in order linker that doesn't depend on shadowgraph, or the removed event mapper

alittley:
- closes #9412


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/linking/InOrderLinker.java
```diff
@@ -16,99 +16,161 @@
...
+     * <p>
+     * This map is needed in addition to the sequence map, since we need to be able to look up parent events based on
+     * hash. Elements are removed from this map when the window of the sequence map is shifted.
+     */
+    private final Map<Hash, EventImpl> parentHashMap = new HashMap<>(INITIAL_CAPACITY);
+
+    /**
+     * Linked events are passed to this consumer.
+     */
+    private final Consumer<EventImpl> eventConsumer;
```
lpetrovic05, line 74: Passing data onto the next consumer in line is a pattern we decided not to go with in the new wiring. The preference is that things work the way they used to, linked events are returned by the component.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
```diff
@@ -0,0 +1,254 @@
...
+     *
+     * @param selfHash              the hash of the event
+     * @param selfGeneration        the generation of the event
+     * @param selfParentHash        the hash of the self parent of the event
+     * @param selfParentGeneration  the generation of the self parent of the event
+     * @param otherParentHash       the hash of the other parent of the event
+     * @param otherParentGeneration the generation of the other parent of the event
+     * @return the mock event
+     */
+    private GossipEvent generateMockEvent(
```
lpetrovic05, line 79: there is a GossipEventBuilder which is better suited for this

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/linking/InOrderLinker.java
alittley: I've been temporarily continuing with this event consumer model in these components since it allows me to easily wire things up before having access to the new framework, to expose any logic flaws in the planned refactor.

This way even if the framework takes a while to finally merge, I can continue making progress

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
alittley: The internal logic of `GossipEventBuilder` is complex and undocumented. Since this component only cares about a small subset of event values, I think it's actually simpler in this case to just mock those few values.

cody-littley (approved):

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
lpetrovic05: Generally, I think its a lot better to build up test tools then having each test do some variation of the same work. If we did that, then GossipEventBuilder would have all the functionality needed for building events for tests. Doesn't have to be for this PR, but I think its a good general rule to follow

alittley: I agree with that, but what gives me pause here is just Ed's current efforts with modifying events.

With event structure undergoing significant changes, I fear investing to expand a utility to emulate the *current* structure may be wasted effort

edward-swirldslabs: My PR will absorb your changes and compensate.

PR #9418, opened 2023-10-23
9046 Enable ScheduleCreateSpecs HAPI tests

jsync-swirlds:
* Enabled most of the HapiTest methods for Schedule Create and Schedule Delete
 * Modifications to schedule handlers to increase passing tests
 * Added a `KeyComparator` class to compare Key objects and impose a total order on collections of Keys.
   * This is necessary to have deterministic order of key collections (particularly sets and maps) across nodes.

### DIFF
--- a/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyComparator.java
+++ b/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyComparator.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.spi.key;
+
+import com.hedera.hapi.node.base.ContractID;
+import com.hedera.hapi.node.base.Key;
+import com.hedera.hapi.node.base.Key.KeyOneOfType;
+import com.hedera.hapi.node.base.KeyList;
+import com.hedera.hapi.node.base.ThresholdKey;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import java.util.Comparator;
+import java.util.List;
+
+/**
+ * Comparator used to impose a deterministic ordering on collections of Key objects.
+ * <br>These include maps, sets, lists, arrays, etc...
+ * <br>The methods in this class are used in hot spot code, so allocation must be kept to a bare
+ * minimum, and anything likely to have performance questions should be avoided.
+ * <br>Note that comparing keys is unavoidably costly.  We try to exit as early as possible throughout
+ * this class, but worst case we're comparing every simple key byte-by-byte for the entire tree, which
+ * may be up to 15 levels deep with any number of keys per level.  We haven't seen a key with
+ * several million "simple" keys included, but that does not mean nobody will create one.
+ */
+public class KeyComparator implements Comparator<Key> {
+    @Override
+    public int compare(final Key first, final Key second) {
+        if (first == second) return 0;
+        else if (first == null) return -1;
+        else if (second == null) return 1;
+        else if (first.key() == null) return second.key() == null ? 0 : -1;
+        else if (second.key() == null) return 1;
+
+        final KeyOneOfType firstKeyType = first.key().kind();
+        final KeyOneOfType secondKeyType = second.key().kind();
+        if (firstKeyType != secondKeyType) return compareCrossType(firstKeyType, secondKeyType);
+        else {
+            // both keys are the same type; so compare the details.
+            return switch (firstKeyType) {
+                case UNSET -> 0; // both unset compares equal.
+                case CONTRACT_ID -> compareContractId(first, second);
+                case DELEGATABLE_CONTRACT_ID -> compareDelegateable(first, second);
+                case ED25519 -> compareEdwards(first, second);
+                case ECDSA_SECP256K1 -> compareSecp256k(first, second);
+                case THRESHOLD_KEY -> compareThreshold(first, second);
+                case KEY_LIST -> compareKeyList(first, second);
+                    // The next two are not currently supported key types.
+                case RSA_3072 -> compareRsa(first, second);
+                case ECDSA_384 -> compareEcdsa(first, second);
+            };
+        }
+    }
+
+    private int compareContractId(final Key first, final Key second) {
+        final ContractID lhs = first.contractID();
+        final ContractID rhs = second.contractID();
+        if (lhs == rhs) return 0;
+        else if (lhs == null) return -1;
+        else if (rhs == null) return 1;
+        else return compareId(lhs, rhs);
+    }
+
+    private int compareDelegateable(final Key first, final Key second) {
+        final ContractID lhs = first.delegatableContractId();
+        final ContractID rhs = second.delegatableContractId();
+        if (lhs == rhs) return 0;
+        else if (lhs == null) return -1;
+        else if (rhs == null) return 1;
+        else return compareId(lhs, rhs);
+    }
+
+    private int compareId(final ContractID leftId, final ContractID rightId) {
+        final long realmOne = leftId.realmNum();
+        final long realmTwo = rightId.realmNum();
+        final long shardOne = leftId.shardNum();
+        final long shardTwo = rightId.shardNum();
+        final Bytes evmOne = leftId.evmAddress();
+        final Bytes evmTwo = rightId.evmAddress();
+        final Long leftNum = leftId.contractNum();
+        final Long rightNum = rightId.contractNum();
+        final long firstId = leftNum != null ? leftNum.longValue() : 0L;
+        final long secondId = rightNum != null ? rightNum.longValue() : 0L;
+        if (realmOne == realmTwo) {
+            if (shardOne == shardTwo) {
+                if (firstId == secondId) {
+                    return compareBytes(evmOne, evmTwo);
+                } else {
+                    return firstId > secondId ? -1 : 1;
+                }
+            } else {
+                return shardOne > shardTwo ? -1 : 1;
+            }
+        } else {
+            return realmOne > realmTwo ? -1 : 1;
+        }
+    }
+
+    private int compareEdwards(final Key first, final Key second) {
+        final Bytes lhs = first.ed25519();
+        final Bytes rhs = second.ed25519();
+        return compareBytes(lhs, rhs);
+    }
+
+    private int compareSecp256k(final Key first, final Key second) {
+        final Bytes lhs = first.ecdsaSecp256k1();
+        final Bytes rhs = second.ecdsaSecp256k1();
+        return compareBytes(lhs, rhs);
+    }
+
+    private int compareThreshold(final Key first, final Key second) {
+        final ThresholdKey lhs = first.thresholdKey();
+        final ThresholdKey rhs = second.thresholdKey();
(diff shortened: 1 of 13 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleCreateSpecs.java
```diff
@@ -203,12 +208,13 @@ private HapiSpec onlyBodyAndMemoCreation() {
...
     }
 
+    @HapiTest
     private HapiSpec bodyAndPayerCreation() {
         return defaultHapiSpec("BodyAndPayerCreation")
                 .given(cryptoCreate(PAYER))
                 .when(scheduleCreate(
                                 ONLY_BODY_AND_PAYER,
-                                cryptoTransfer(tinyBarsFromTo(DEFAULT_PAYER, GENESIS, 1))
+                                cryptoTransfer(tinyBarsFromTo(PAYER, GENESIS, 1))
```
iwsimon, line 217: Why is PAYER insteadof DEFAULT_PAYER here? Didn't check the code in detail. Are those two different values?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleExecutionSpecs.java
```diff
@@ -1571,11 +1598,11 @@ public HapiSpec executionWithDefaultPayerButNoFundsFails() {
                             Assertions.assertEquals(
                                     INSUFFICIENT_PAYER_BALANCE,
                                     triggeredTx.getResponseRecord().getReceipt().getStatus(),
-                                    SCHEDULED_TRANSACTION_SHOULD_NOT_BE_SUCCESSFUL);
+                                    SCHEDULED_TRANSACTION_MUST_NOT_SUCCEED);
```
mhess-swl, line 1601: Is there a difference between the mono and mod code that makes these changes necessary? (There are quite a few  of them in this file) 
And, does mono service also pass with these changes?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleCreateSpecs.java
jsync-swirlds: Yes, different values, and default payer doesn't match the created payer on line 214.  It *sometimes* worked due to ordering assumptions; this makes it correctly work regardless of execution order.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleExecutionSpecs.java
```diff
@@ -2064,7 +2091,7 @@ private HapiSpec scheduledFreezeWorksAsExpected() {
         return defaultHapiSpec("ScheduledFreezeWorksAsExpected")
                 .given(
                         cryptoCreate(PAYING_ACCOUNT),
-                        overriding(SCHEDULING_WHITELIST, FREEZE),
+                        overriding(SCHEDULING_WHITELIST, "Freeze"),
```
mhess-swl, line 2094: Is this a readability change? Curious why this would be an issue. I see a few of these changes in this file

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyComparator.java
```diff
@@ -0,0 +1,212 @@
...
+     * The "natural" order, in this case, is just the type order in the proto file.
+     * @param firstKeyType The OneOfType for the first (left hand) key.
+     * @param secondKeyType The OneOfType for the second (right hand) key.
+     * @return a value less than, equal to, or greater than 0 indicating if the first key type should be
+     *   sorted before, the same, or after, the second key type.
+     */
+    private int compareCrossType(final KeyOneOfType firstKeyType, final KeyOneOfType secondKeyType) {
+        return firstKeyType.protoOrdinal() - secondKeyType.protoOrdinal();
+    }
+}
```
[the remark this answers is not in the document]
jsync-swirlds: Mono does not sort keys.  Mono uses lists and assumes insertion order is fully deterministic, which is has been so far, but it's a little fragile.  The new code uses sorted structures to be certain the order in state is always the same for all nodes.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleCreateSpecs.java
```diff
@@ -220,12 +226,14 @@ private HapiSpec bodyAndPayerCreation() {
...
     }
 
+    @HapiTest
     private HapiSpec bodyAndSignatoriesCreation() {
         var scheduleName = "onlyBodyAndSignatories";
         var scheduledTxn = cryptoTransfer(tinyBarsFromTo(SENDER, RECEIVER, 1));
 
         return defaultHapiSpec("BodyAndSignatoriesCreation")
                 .given(
+                        overriding(SCHEDULING_WHITELIST, "CryptoTransfer,CryptoCreate"),
```
mhess-swl, line 236: Was this override not necessary for mono code? Does it have to do with the comment further up about the weird required ordering of the tests?

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyComparator.java
```diff
@@ -0,0 +1,212 @@
...
+ * Comparator used to impose a deterministic ordering on collections of Key objects.
+ * <br>These include maps, sets, lists, arrays, etc...
+ * <br>The methods in this class are used in hot spot code, so allocation must be kept to a bare
+ * minimum, and anything likely to have performance questions should be avoided.
+ * <br>Note that comparing keys is unavoidably costly.  We try to exit as early as possible throughout
+ * this class, but worst case we're comparing every simple key byte-by-byte for the entire tree, which
+ * may be up to 15 levels deep with any number of keys per level.  We haven't seen a key with
+ * several million "simple" keys included, but that does not mean nobody will create one.
+ */
+public class KeyComparator implements Comparator<Key> {
```
mhess-swl, line 38: A thorough unit test for this class would be helpful

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/AbstractScheduleHandler.java
```diff
@@ -267,44 +275,52 @@ protected boolean validationOk(final ResponseCodeEnum validationResult) {
...
-     * Implementation Note: we use ConcurrentSkipListSet that orders the entries by hash code.  This means that
-     *   object order will be very different from a reasonable natural ordering, but it should (if hashCode is defined
-     *   properly) be consistent with equals, and have sufficient diffusion to work well with the random
-     *   functions inherent in the Skip List data structure. Order is not important, so this approach is reasonable.
-     *   The alternative is a CopyOnWriteArraySet, which has significant GC performance issues.
      * @param keyCollection an Iterable of Key values.
-     * @return a {@link Set<Key>} containing the same contents as the input.  Duplicates and null values are excluded
-     * from this Set.  This Set is always a modifiable set.
+     * @return a {@link SortedSet<Key>} containing the same contents as the input.
+     * Duplicates and null values are excluded from this Set.  This Set is always a modifiable set.
```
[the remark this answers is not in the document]
jsync-swirlds: The calling code assumes the returned set can be used in the key lookup and verification, which require the set to be modifiable so we can gather newly signed keys, and reduce the remaining required keys.
Also, Mono filters the keys in state to remove no-longer-required, but previously signed, keys; which requires a modifiable set.
By having this guarantee in this method, it documents that there is no need to copy sets for modification.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/ScheduleCreateHandler.java
```diff
@@ -165,13 +175,34 @@ public void handle(@NonNull final HandleContext context) throws HandleException
         }
     }
 
+    /*
+     Note: This method should not be needed, but HapiTests don't work properly in this regard
+      (genesis doesn't fill in the next entity state).  This ensures the next entity is not a system entity before
+      returning the value.
+      This is temporary until we can fix the genesis state handling for newEntityNum...
```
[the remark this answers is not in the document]
jsync-swirlds: It's already fixed, I think.  My next PR will remove it.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/ScheduleCreateHandler.java
```diff
@@ -238,11 +270,29 @@ private void checkSchedulableWhitelistHandle(final Schedule provisionalSchedule,
         }
     }
 
+    private void validatePayerAndScheduler(
+            final PreHandleContext context, final ScheduleCreateTransactionBody scheduleBody) throws PreCheckException {
+        final ReadableAccountStore accountStore = context.createStore(ReadableAccountStore.class);
+        final AccountID payerForSchedule = scheduleBody.payerAccountID();
+        if (payerForSchedule != null) {
+            final Account payer = accountStore.getAccountById(payerForSchedule);
+            if (payer == null) {
```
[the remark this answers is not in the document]
jsync-swirlds: Here we just need to check that the payer exists.  Currently, we should not check more until execution, because mono assumes nearly all checks wait until then.
In the future, after differential testing is completed, we should definitely consider making this check more robust.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleExecutionSpecs.java
jsync-swirlds: 1. Yes, there are differences in Mono, lots of weirdness around execution
2. Yes, Mono service passes with these changes
3. (not asked) There are some additional changes needed, mostly because the tests in this class are extremely order-dependent and fail in HapiTest because the order is random (or at least not exactly what the old suite expects).  Whitelist management is a big issue, and some of these tests unintentionally assume a prior test "set up" some balances (for instance).  The next PR, which focuses on execution tests, will make quite a few changes to make these tests reliable and not order-dependent.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleExecutionSpecs.java
jsync-swirlds: Mostly readability, yes.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyComparator.java
jsync-swirlds: True.  I'll add that to the todo list.  Might not be the next PR (that needs to get at least the majority of execution tests working), but definitely the one after that.

mhess-swl (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/schedule/ScheduleCreateSpecs.java
iwsimon: Not sure why DEFAULT_PAYER was used here before. Was this test passed in mono code run?

jsync-swirlds: It passed in mono (and still does); not entirely clear why (possibly due to the prior tests setting the DEFAULT_PAYER into just the right state), but it failed in Hapi and the mismatch was clearly unintentional (possibly an ancient copy/paste that just happened to succeed?).
It's surprising how much of the E2E and Integration tests pass almost due to luck rather than intent.

iwsimon (approved):
LGTM

povolev15 (approved):
LGTM

mhess-swl (approved):

rbair23 (approved):

PR #9441, opened 2023-10-24
Maintain doubly-linked list storage on contact commit

### LINKED ISSUE #9346: Update contract storage after transaction execution
### Problem

Contract storage is implemented as a key/value pair where the value is implemented as a doubly linked list of bytes.  After successful contract transaction execution, the value needs to be updated to reflect any changes as a result.  

### Solution

Implement changes to update the storage linked list for a contract to reflect current state after successful contract transaction execution. 

### Alternatives

_No response_

lukelee-sl:
**Description**:
Maintain as necessary the doubly-linked list storage of smart contracts when committed via the RootProxyWorldUpdater

**Related issue(s)**:

Fixes #9346 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +60,178 @@ public IterableStorageManager() {
...
+                            tuweniToPbjBytes(access.writtenValue()),
+                            contractAccesses.contractNumber(),
+                            tuweniToPbjBytes(access.key()));
+                } else if (access.isUpdate()) {
+                    updateAccessedValue(
+                            store,
+                            tuweniToPbjBytes(access.writtenValue()),
+                            contractAccesses.contractNumber(),
+                            tuweniToPbjBytes(access.key()));
+                }
```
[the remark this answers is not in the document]
lukelee-sl: I believe it is a closed set.  I modified the code make this a little clearer.

david-bakin-sl (commented):
A bunch of nullability things that need adjusting, plus some questions on logging.  

But also: It may be that the unit tests for this low-level manipulation are sufficient, but do you think an xtest could be useful too? It too might be kind of gnarly without additional debugging help from token service, e.g., a variant of `summarizeContractChanges` that also returned changed firstContractKeys.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +60,178 @@ public IterableStorageManager() {
...
+                var firstContractKey = contractFirstKeyOf(enhancement, contractAccesses.contractNumber());
+
+                if (access.isRemoval()) {
+                    firstContractKey = removeAccessedValue(
+                            store, firstContractKey, contractAccesses.contractNumber(), tuweniToPbjBytes(access.key()));
+                } else if (access.isInsertion()) {
+                    firstContractKey = insertAccessedValue(
+                            store,
+                            firstContractKey,
+                            tuweniToPbjBytes(access.writtenValue()),
```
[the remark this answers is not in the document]
david-bakin-sl: Oh yes you're right.  That's too convoluted (non-local) for IntelliJ to figure out, sorry!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +60,178 @@ public IterableStorageManager() {
...
+            final var newSlotKey = new SlotKey(contractNumber, newKey);
+            final var newSlotValue = SlotValue.newBuilder()
+                    .value(newValue)
+                    .previousKey(null)
+                    .nextKey(firstContractKey)
+                    .build();
+            store.putSlot(newSlotKey, newSlotValue);
+            return newKey;
+        } catch (Exception irreparable) {
+            log.error("Failed link management when inserting {}", newKey, irreparable);
```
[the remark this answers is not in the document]
lukelee-sl: I agree that the message seems wrong but I think I might want to keep the scope of the try where it is because if any of the lines that could throw exceptions leading up to the `putSlot` are actually issues (null `key` given or `slotKey` does not map to a valid `slotValue`) are not things which can be mitigated so that the `putSlot` would have the correct value.  Best thing to do is bail and log in that case in my opinion.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +60,178 @@ public IterableStorageManager() {
...
+    private Bytes removeAccessedValue(
+            @NonNull final ContractStateStore store,
+            @Nullable final Bytes firstContractKey,
+            long contractNumber,
+            @NonNull final Bytes key) {
+        try {
+            Objects.requireNonNull(store);
+            Objects.requireNonNull(key);
+            final var slotKey = new SlotKey(contractNumber, key);
+            final var slotValue = Objects.requireNonNull(store.getSlotValue(slotKey), () -> "Missing key " + slotKey);
```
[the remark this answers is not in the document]
lukelee-sl: The message was in mono service so wanted to match.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +60,178 @@ public IterableStorageManager() {
...
+     * @return the new first key in the linked list of storage for the given contract
+     */
+    @NonNull
+    private Bytes removeAccessedValue(
+            @NonNull final ContractStateStore store,
+            @Nullable final Bytes firstContractKey,
+            long contractNumber,
+            @NonNull final Bytes key) {
+        try {
+            Objects.requireNonNull(store);
```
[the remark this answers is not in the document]
david-bakin-sl: TY IMO looks much more focused on the linked list manipulation now!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +61,200 @@ public IterableStorageManager() {
...
+
+                // Create new previous value and put into the store
+                final var newPrevValue =
+                        prevValue.copyBuilder().nextKey(nextKey).build();
+                store.putSlot(prevSlotKey, newPrevValue);
+            }
+            store.removeSlot(slotKey);
+            return key.equals(firstContractKey) ? slotValue.nextKey() : firstContractKey;
+        } catch (Exception irreparable) {
+            log.error(
```
david-bakin-sl, line 170: Just occurred to me: Here (+ update + insert) should the exception be rethrown?  (Or wrapped in another exception and thrown?)

If so, two issues occur at first glance:
a) would caller `persistChanges` let it fly through (stopping then at first error) or would it collect possibly multiple such exceptions and wrap them together and throw a single exception for a batch?  (Way overkill I imagine)
b) way upstream (I think) is `TransactionProcessor.safeCommit` which _does_ handle a `ResourceExhaustedException` specifically but no other.  So I don't know _what_ happens if a different exception is thrown when trying to commit.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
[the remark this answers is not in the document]
david-bakin-sl: ok

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +61,200 @@ public IterableStorageManager() {
...
+        return new SlotKey(contractNumber, key);
+    }
+
+    @NonNull
+    private SlotValue newSlotValueFor(
+            @NonNull final ContractStateStore store,
+            final boolean forModify,
+            @NonNull final SlotKey slotKey,
+            @NonNull final String msgOnError,
+            @NonNull final Bytes key) {
```
david-bakin-sl, line 255: A nit: I think the `key` argument here is always `slotKey.key()` so isn't strictly necessary?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +61,200 @@ public IterableStorageManager() {
...
+        return firstContractKey;
+    }
+
+    @NonNull
+    private SlotKey newSlotKeyFor(long contractNumber, @NonNull final Bytes key) {
+        return new SlotKey(contractNumber, key);
+    }
+
+    @NonNull
+    private SlotValue newSlotValueFor(
```
david-bakin-sl, line 250: This isn't a `newSlotValueFor`, it's the _existing_ `SlotValue` ... (i.e., looking at the name `new...`)

david-bakin-sl (dismissed):
LGTM with a couple of questions

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
lukelee-sl: actually calling this with key = `nextKey` and key = `prevKey`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
lukelee-sl: changed to `slotValueFor`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
lukelee-sl: Mono service does not seem to re-throw the exception.  Doesn't mean that is correct.  The way things are now, an exception in one change does not affect the processing of the rest of the changes which might be the right behavior but certainly open for discussion.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
david-bakin-sl: I said that badly.  I'm looking at 4 uses of this method and in each case I think the `key` argument is the key of the `slotKey` argument (where in each case the slot key is now created on the previous line: by the way, that regularity in your updated code is one of the reasons the updated code is easier to read)

david-bakin-sl (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
lukelee-sl: ah, I see. changed.

david-bakin-sl (dismissed):
LGTM for sure!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/IterableStorageManager.java
```diff
@@ -51,29 +61,199 @@ public IterableStorageManager() {
...
+                switch (StorageAccessType.getAccessType(access)) {
+                    case REMOVAL -> firstContractKey = removeAccessedValue(
+                            store, firstContractKey, contractAccesses.contractNumber(), tuweniToPbjBytes(access.key()));
+                    case INSERTION -> firstContractKey = insertAccessedValue(
+                            store,
+                            firstContractKey,
+                            tuweniToPbjBytes(access.writtenValue()),
+                            contractAccesses.contractNumber(),
+                            tuweniToPbjBytes(access.key()));
+                    case UPDATE -> updateAccessedValue(
```
tinker-michaelj, line 91: The `SavepointStackImpl` will already be committing the modified slot values as part of committing the storage `WritableKVState` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/WritableContractStateStore.java#L35) (note we used its pending modifications to construct the `List<StorageAccesses>` [in the first place](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java#L134)).

So afaict this branch is doing redundant work, given it doesn't change `prev`/`next` pointers.

lukelee-sl: makes sense.  removed the update case.

david-bakin-sl (approved):

tinker-michaelj (approved):
Looks great to me @lukelee-sl, tyvm! 🙌

PR #9443, opened 2023-10-24
Fix `mintUniqueTokenReceiptCheck` in `UniqueTokenManagementSpecs`

georgi-l95:
**Description**:
This PR fix `mintUniqueTokenReceiptCheck` in `UniqueTokenManagementSpecs` by setting the treasury as owner, if ownerId is null.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
@@ -180,6 +180,8 @@ public class RecordFinalizerBase {
                         persistedNft.hasOwnerId() && !persistedNft.ownerId().equals(AccountID.DEFAULT);
                 // If the NFT did not have an owner before set it to the treasury account
                 senderAccountId = hasOwnerId ? persistedNft.ownerId() : token.treasuryAccountId();
+            } else {
+                senderAccountId = ZERO_ACCOUNT_ID;
             }
 
             // If the NFT has been burned or wiped, modifiedNft will be null. In that case the receiverId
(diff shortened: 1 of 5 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
```diff
@@ -185,7 +186,10 @@ protected Map<TokenID, List<NftTransfer>> nftChangesFrom(
             // will be explicitly set as 0.0.0
             final var builder = NftTransfer.newBuilder();
             if (modifiedNft != null) {
-                builder.receiverAccountID(modifiedNft.ownerId());
+                // If ownerId is null, we assume that the receiver is the treasury
```
Neeharika-Sompalli, line 189: Great catch! Thanks @georgi-l95 ! This looks correct , but we also need else case on line 184
when `persistedNft == null` set the `senderAccountId` to  `ZERO_ACCOUNT_ID`

Neeharika-Sompalli: Which HAPI test is failing? @georgi-l95

netopyr:
HAPI tests are failing

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
Neeharika-Sompalli: @georgi-l95  This test is failing in mono-service also. We should add the change I suggested and update test. I see @mhess-swl  is also working on this ..

georgi-l95: Yes, I will close this one as Matt already fixed it in his.

PR #9453, opened 2023-10-25
Populate evm function result on failing eth transaction

### LINKED ISSUE #9149: Failing Ethereum transaction doesn't populate contract function result in the record stream
Contract function result is missing in the transaction records created for failing transactions. 
This leads to some inconsistencies in the Mirror node - https://github.com/hashgraph/hedera-mirror-node/issues/6831 and the Relay - https://github.com/hashgraph/hedera-json-rpc-relay/issues/1711.

The problem comes from the fact that some validations on the transaction fail (e.g. [here](https://github.com/hashgraph/hedera-services/blob/06b8d8ae123ece1ab72c35f43c1097d1dfb0dc9b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L147)), which throws an exception and the logic that adds information about the contract function result to the record stream (e.g. [this line](https://github.com/hashgraph/hedera-services/blob/06b8d8ae123ece1ab72c35f43c1097d1dfb0dc9b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/ethereum/EthereumTransitionLogic.java#L129)) is skipped altogether. 

Steps to reproduce:
1. Execute the following EET:
```
HapiSpec callToNonExistingContractFailsGracefully() {

        return defaultHapiSpec("callToNonExistingContractFailsGracefully")
                .given(
                        withOpContext((spec, ctxLog) -> spec.registry().saveContractId("invalid", asContract("1.1.1"))),
                        newKeyNamed(SECP_256K1_SOURCE_KEY).shape(SECP_256K1_SHAPE),
                        cryptoCreate(RELAYER).balance(6 * ONE_MILLION_HBARS),
                        cryptoCreate(TOKEN_TREASURY),
                        cryptoTransfer(tinyBarsFromAccountToAlias(GENESIS, SECP_256K1_SOURCE_KEY, ONE_HUNDRED_HBARS)))
                .when(withOpContext((spec, opLog) -> allRunFor(
                        spec,
                        ethereumCallWithFunctionAbi(
                                        false,
                                "invalid",
                                        getABIFor(Utils.FunctionType.FUNCTION, "totalSupply", ERC20_ABI))
                                .type(EthTxData.EthTransactionType.EIP1559)
                                .signingWith(SECP_256K1_SOURCE_KEY)
                                .payingWith(RELAYER)
                                .via("invalidContractCallTxn")
                                .nonce(0)
                                .gasPrice(0L)
                                .gasLimit(1_000_000L)
                                .hasKnownStatusFrom(ResponseCodeEnum.SUCCESS, INVALID_CONTRACT_ID))))
                .then(
                        getTxnRecord("invalidContractCallTxn").andAllChildRecords().logged()
                );
    }
```
2. Observe the transaction record. There is no contract function result. 

Expected behavior: A contract function result to be present in the record stream.

stoqnkpL:
**Description**:
Contract function result is missing in the transaction records created for transactions that the consensus node deems "invalid".
This leads to some inconsistencies in the Mirror node - https://github.com/hashgraph/hedera-mirror-node/issues/6831 and the Relay - https://github.com/hashgraph/hedera-json-rpc-relay/issues/1711.

More info in the issue. 

**Related issue(s)**:

Fixes #9149 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/ethereum/EthereumTransitionLogic.java
```diff
@@ -120,13 +125,20 @@ public void doStateTransition() {
...
-            delegateToCallTransition(callerNum.toId(), synthTxn, relayerId, maxGasAllowance, userOfferedGasPrice);
-        } else {
-            delegateToCreateTransition(callerNum.toId(), synthTxn, relayerId, maxGasAllowance, userOfferedGasPrice);
+        try {
+            if (synthTxn.hasContractCall()) {
+                delegateToCallTransition(callerNum.toId(), synthTxn, relayerId, maxGasAllowance, userOfferedGasPrice);
+            } else {
+                delegateToCreateTransition(callerNum.toId(), synthTxn, relayerId, maxGasAllowance, userOfferedGasPrice);
+            }
+        } catch (InvalidTransactionException e) {
```
lukelee-sl, line 134: Is `InvalidTransactionException` the only one that should be caught here?

stoqnkpL: hmm, my thinking is we want to catch all _validation_ exceptions (thrown for example [here](https://github.com/hashgraph/hedera-services/blob/06b8d8ae123ece1ab72c35f43c1097d1dfb0dc9b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L147)) and populate the record stream with the message they contain. I'm not sure we would want to catch _any_ exception here and possibly include information in the record stream that's more sensitive. 
@tinker-michaelj wdyt?

tinker-michaelj (approved):
LGTM!

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/ethereum/EthereumTransitionLogic.java
Nana-EC: My only concern is the bugs/system error part.
If i submit an execution I should be able to get a hint of why it failed and the result details regardless.
We want to avoid submitting a transaction and not knowing details around gas consumed and result details.
If this PR doesn't address bugs/system errors we've definitely improved the issue but haven't resolved it.
I'm fine with a follow up in 0.45 if necessary but I'm not certain we should leave as is in the improved state.
@stoqnkpL if we don't cover the rest please open a ticket to describe what's left. It's possible we triage and say won't fix but let's still capture it

Nana-EC (approved):
LG
Please capture the non covered flow in a ticket for reference

PR #9463, opened 2023-10-25
Move where internal consistency check takes place

### LINKED ISSUE #9449: Move internal data consistency checks to a new DTO
Currently, we are checking in the `StaticValidators` class that an event's purported parents are internally consistent, with respect to generation and hash.

There is no reason that this shouldn't be done immediately upon deserialization when receiving events through gossip, prior to hashing.

Additionally, we should be checking that hashed and unhashed data are not null, and that max transaction size isn't being exceeded.

Create a new DTO that can eventually be slotted in between gossip and hashing, to perform basic data validity checks.

alittley:
- closes #9449

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
@@ -16,6 +16,9 @@
 
 package com.swirlds.platform.event;
 
+import static com.swirlds.platform.consensus.GraphGenerations.FIRST_GENERATION;
+
+import com.swirlds.common.crypto.Hash;
 import com.swirlds.common.io.streams.SerializableDataInputStream;
 import com.swirlds.common.io.streams.SerializableDataOutputStream;
 import com.swirlds.common.system.NodeId;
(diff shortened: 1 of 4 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
```diff
@@ -83,6 +113,10 @@ public void deserialize(final SerializableDataInputStream in, final int version)
             in.readLong(); // roundCreated
         }
         timeReceived = Instant.now();
+
+        if (hashedData == null || unhashedData == null || !isHashedDataInternallyConsistent(hashedData)) {
+            throw new IOException("Invalid event: " + this);
+        }
```
cody-littley, line 119: It would be nice if the message included the reason why the event is invalid. Not very useful if an attacker is sending a bad event, but perhaps helpful when debugging.

alittley: Logging improved

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/events/BaseEventHashedData.java
```diff
@@ -235,8 +230,25 @@ public void deserialize(
...
+                hasUserTransactions = true;
+            }
+
+            totalTransactionSize += transaction.getSerializedLength();
+        }
+
+        if (totalTransactionSize > transactionConfig.maxTransactionBytesPerEvent()) {
+            throw new IOException("Total size of transactions in this event %s exceeds the maximum allowed size"
+                    .formatted(totalTransactionSize));
+        }
```
cody-littley, line 251: Can this be moved to a helper function? Might reduce probably of merge conflicts with @edward-swirldslabs's pending event changes.

edward-swirldslabs: Given that my merge has to wait for HIP approval, I am expecting merge conflicts to accumulate.  I don't mind absorbing them.

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/events/BaseEventHashedData.java
```diff
@@ -235,8 +230,25 @@ public void deserialize(
         otherParentHash = in.readSerializable(false, Hash::new);
         timeCreated = in.readInstant();
         in.readInt(); // read serialized length
-        transactions = in.readSerializableArray(ConsensusTransactionImpl[]::new, maxTransactionCount, true);
-        checkUserTransactions();
+
+        final TransactionConfig transactionConfig = ConfigurationHolder.getConfigData(TransactionConfig.class);
```
lpetrovic05, line 234: ConfigurationHolder is deprecated so we want to avoid using it. Events should at some point move to protobuf, which I think will not have the option of validating them while being deserialized. So maybe this change does not make sense unfortunately 😞

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/events/BaseEventHashedData.java
alittley: I split this logic off into a helper method (with an unfortunately awkward name)

cody-littley (approved):

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/events/BaseEventHashedData.java
lpetrovic05: The current validation logic will need to be preserved, but if we move it to a validator that performs it on a DTO, this will be compatible with protobuf. But if its part of the serialization, we have to refactor it again afterwards

Issue #9479, opened 2023-10-25
Add more logs to debug virtual map reconnect issues


artemananiev:
To help investigate #9477, more information about reconnects / virtual maps / data sources should be logged

PR #9481, opened 2023-10-25
9479: Add more logs to debug virtual map reconnect issues

artemananiev:
More logs and more try-catch blocks.

Fixes: https://github.com/hashgraph/hedera-services/issues/9479


cody-littley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/LearningSynchronizer.java
```diff
@@ -149,12 +150,15 @@ public void synchronize() throws InterruptedException {
...
+        logger.warn(
+                RECONNECT.getMarker(),
+                "Deleting partially constructed tree:\n{}",
+                new MerkleTreeVisualizer(newRoot).setDepth(5).render());
         try {
             if (newRoot != null) {
                 newRoot.release();
             }
-        } catch (final Exception ex) {
+        } catch (final Throwable ex) {
```
jsync-swirlds, line 161: This is a bit dangerous.  Might be better to catch RuntimeException and not Throwable, so Errors still exit the JVM (as they should).

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -794,29 +794,34 @@ public void close() throws IOException {
...
+                    if (longKeyToPath != null) {
+                        longKeyToPath.close();
+                    }
+                    if (objectKeyToPath != null) {
+                        objectKeyToPath.close();
+                    }
+                    pathToKeyValue.close();
+                    // Store metadata
+                    saveMetadata(dbPaths.metadataFile);
+                } catch (final Throwable t) {
```
jsync-swirlds, line 817: Same here, Throwable includes Errors (which almost never should be caught), so perhaps catch RuntimeException instead?

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/VirtualRootNode.java
```diff
@@ -965,7 +965,7 @@ public void onShutdown(final boolean immediately) {
         if (dataSource != null) {
             try {
                 dataSource.close();
-            } catch (final IOException e) {
+            } catch (final Throwable e) {
```
jsync-swirlds, line 968: Here as well.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
artemananiev: I should probably re-throw it after it's logged.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/LearningSynchronizer.java
artemananiev: I will probably revert this change and will not include it to the PR. More important is to print the tree as a part of warn() call right above.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
jsync-swirlds: The only issue with re-throwing Throwable is that it's checked, so it kind of cascades Throwable all the way up, so might need to wrap in a RuntimeException to avoid that.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/VirtualRootNode.java
artemananiev: Will change it to `Exception`

imalygin (dismissed):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/VirtualRootNode.java
jsync-swirlds: This method is expected to throw IOException, so if that's caught it should rethrow.
Might be best to split as IOException and RuntimeException so both can be rethrown...

imalygin (dismissed):

jsync-swirlds (dismissed):
Still a little concerned with the catch structure, but should at least be safe.

imalygin (approved):

jsync-swirlds (approved):

PR #9482, opened 2023-10-25
Wiring Framework

### LINKED ISSUE #9005: Prototype wiring framework

cody-littley:
Closes #9005


platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/benchmark/WiringBenchmark.java
```diff
@@ -0,0 +1,144 @@
...
+                    System.out.println("Uncaught exception in thread " + t.getName());
+                    e.printStackTrace();
+                },
+                true);
+
+        final PlatformContext platformContext =
+                TestPlatformContextBuilder.create().build();
+        final WiringModel model = WiringModel.create(platformContext, Time.getCurrent());
+
+        // Step 1: construct wires
```
alittley, line 61: `// step` comments still reference old terminology

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/EventSignatureValidationWire.java
```diff
@@ -0,0 +1,96 @@
...
+import com.swirlds.common.wiring.TaskScheduler;
+import com.swirlds.common.wiring.WiringModel;
+import com.swirlds.platform.event.GossipEvent;
+import com.swirlds.platform.event.validation.EventValidator;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Wiring for the event signature validator.
+ */
+public class EventSignatureValidationWire {
```
alittley, line 30: no longer a "wire"

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/OrphanBufferWire.java
```diff
@@ -0,0 +1,105 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.List;
+
+// Future work: it may actually make more sense to colocate the wire classes with the implementations.
+//              This decision can be delayed until we begin migration in earnest.
+
+/**
+ * Wiring for the {@link OrphanBuffer}.
+ */
+public class OrphanBufferWire {
```
alittley, line 34: no longer a "wire"

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+import java.util.concurrent.ForkJoinPool.ManagedBlocker;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * This class is used to implement backpressure in a {@link java.util.concurrent.ForkJoinPool} friendly way.
+ */
+class BackpressureBlocker implements ManagedBlocker {
+
+    private final AtomicLong count;
```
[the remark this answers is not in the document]
cody-littley:
```
    /**
     * The current count. This object will attempt to block until this count can be increased by one without exceeding
     * the maximum capacity.
     */
    private final AtomicLong count;

    /**
     * The maximum desired capacity. It is possible that the current count may exceed this capacity (i.e. if
     * {@link ObjectCounter#forceOnRamp()} is used to bypass the capacity).
     */
    private final long capacity;

    /**
     * The amount of time to sleep while waiting for capacity to become available, or 0 to not sleep.
     */
    private final int sleepNanos;
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onRamp() {
+        try {
+            count.updateAndGet(increment);
+            // Best case scenario: capacity is was immediately available.
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/EmptyBlocker.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+class EmptyBlocker implements ManagedBlocker {
+
+    private final AtomicLong count;
+    private final int sleepNanos;
+
+    /**
+     * Constructor.
+     *
+     * @param count                     the counter to use
```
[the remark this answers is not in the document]
cody-littley: odd, fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/EmptyBlocker.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param sleepNanos                the number of nanoseconds to sleep while blocking, or -1 to not sleep
```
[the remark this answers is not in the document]
cody-littley: No reason, changed to 0

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.common.wiring.counters;
+
+/**
+ * A class that counts the number of objects in a various part of the pipeline.
```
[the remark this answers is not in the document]
cody-littley: Copilot loves to propagate spelling mistakes from one part of the code to the other. 🙃 Searched for this phrase and fixed all occurrences.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public abstract class ObjectCounter {
```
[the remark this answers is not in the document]
cody-littley: Oleg says that interfaces decrease performance, he suggested that I avoid interfaces for core parts of this framework. Normally not a big deal, but possibly a big deal in very performance critical parts of the code.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/ModelVertex.java
```diff
@@ -0,0 +1,130 @@
...
+
+    private final List<ModelEdge> outgoingEdges = new ArrayList<>();
+
+    /**
+     * Constructor.
+     *
+     * @param name                the name of the vertex
+     * @param insertionIsBlocking true if the insertion of this vertex may block until capacity is available
+     */
+    public ModelVertex(@NonNull String name, boolean insertionIsBlocking) {
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/ModelVertex.java
```diff
@@ -0,0 +1,130 @@
...
+    public int hashCode() {
+        return name.hashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(Object obj) {
+        return (obj instanceof ModelVertex) && name.equals(((ModelVertex) obj).name);
```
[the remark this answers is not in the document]
cody-littley:
```
if (obj instanceof final ModelVertex that) {
    return name.equals(that.name);
}
return false;
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringFlowchart.java
```diff
@@ -0,0 +1,237 @@
...
+                    throw new IllegalStateException("Vertex " + vertex.getName() + " is in multiple groups.");
+                }
+            }
+        }
+
+        return new ArrayList<>(uniqueVertices);
+    }
+
+    /**
+     * For all verticies that are in collapsed groups, we want to draw edges to the collapsed group instead of to the
```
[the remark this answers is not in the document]
cody-littley: fixed all incidents

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/AbstractTask.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    private final ForkJoinPool pool;
+
+    /**
+     * Constructor.
+     *
+     * @param dependencyCount the number of dependencies that must be satisfied before this task is eligible for
+     *                        execution
+     */
+    protected AbstractTask(@NonNull final ForkJoinPool pool, final int dependencyCount) {
```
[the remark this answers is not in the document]
cody-littley: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTask.java
```diff
@@ -0,0 +1,66 @@
...
+    private final UncaughtExceptionHandler uncaughtExceptionHandler;
+
+    /**
+     * Constructor.
+     *
+     * @param handler the method that will be called when this task is executed
+     * @param data    the data to be passed to the consumer for this task
+     */
+    protected ConcurrentTask(
+            @NonNull final ForkJoinPool pool,
```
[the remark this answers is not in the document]
cody-littley: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTask.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    protected boolean exec() {
+        try {
+            handler.accept(data);
+        } catch (final Throwable t) {
+            uncaughtExceptionHandler.uncaughtException(Thread.currentThread(), t);
+        } finally {
+            offRamp.offRamp();
+        }
+        return true;
```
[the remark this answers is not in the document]
cody-littley: This has to do with the weirdness of the fork join pool API. We don't want the fork join pool to have anything to do with error handling, so we just always tell it that there were no errors.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTaskScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+import com.swirlds.common.wiring.counters.ObjectCounter;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.lang.Thread.UncaughtExceptionHandler;
+import java.util.Objects;
+import java.util.concurrent.ForkJoinPool;
+import java.util.function.Consumer;
+
+/**
+ * A {@link TaskScheduler} that permits parallel execution of tasks. Similar to {@link ConcurrentTaskScheduler} but with
```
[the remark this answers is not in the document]
cody-littley: A relic of a previous iteration of this branch. Previously there were two implementations of a concurrent scheduler. I've removed this sentence.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTaskScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+ * extra metering.
+ *
+ * @param <O> the output time of the wire (use {@link Void}) for a task scheduler with no output type)
```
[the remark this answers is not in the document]
cody-littley: fixed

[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTaskScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+    private final ObjectCounter onRamp;
+    private final ObjectCounter offRamp;
+    private final UncaughtExceptionHandler uncaughtExceptionHandler;
+    private final ForkJoinPool pool;
+
+    /**
+     * Constructor.
+     *
+     * @param model                    the wiring model containing this wire
+     * @param name                     the name of the wire
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTaskScheduler.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param pool                     the fork join pool that will execute tasks on this wire
+     * @param uncaughtExceptionHandler the handler for uncaught exceptions
+     * @param onRamp                   an object counter that is incremented when data is added to the wire, ignored if
+     *                                 null
```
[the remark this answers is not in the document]
cody-littley: Relic of a previous iteration. Fixed.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/ConcurrentTaskScheduler.java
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/SequentialTaskScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Objects;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+/**
+ * A {@link TaskScheduler} that guarantees that tasks are executed sequentially in the order they are received.
+ *
+ * @param <O> the output time of the wire (use {@link Void}) for a task scheduler with no output type)
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/SequentialTaskScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param model                    the wiring model containing this wire
+     * @param name                     the name of the task scheduler
+     * @param pool                     the fork join pool that will execute tasks on this scheduler
+     * @param uncaughtExceptionHandler the uncaught exception handler
+     * @param onRamp                   an object counter that is incremented when data is added to the task scheduler,
+     *                                 ignored if null
```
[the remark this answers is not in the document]
cody-littley: Annotation is correct, fixed javadoc

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/SequentialTaskScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Schedule a task to be handled. This should only be called after successfully on-ramping (one way or another).
+     *
+     * @param handler the method that will be called when this task is executed
+     * @param data    the data to be passed to the consumer for this task
+     */
+    private void scheduleTask(@NonNull final Consumer<Object> handler, @Nullable final Object data) {
+        // This method may be called by may threads, but it must serialize the results a sequence of tasks that are
```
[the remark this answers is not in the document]
cody-littley:
```
        // This method may be called by many threads, but actual execution is required to happen serially. This method 
        // organizes tasks into a linked list. Tasks in this linked list are executed one at a time in order. 
        // When execution of one task is completed, execution of the next task is scheduled on the pool.
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/SequentialTaskScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.function.Consumer;
+
+/**
+ * A {@link TaskScheduler} that guarantees that tasks are executed sequentially in the order they are received.
+ *
+ * @param <O> the output time of the wire (use {@link Void}) for a task scheduler with no output type)
+ */
+public class SequentialTaskScheduler<O> extends TaskScheduler<O> {
+
+    private final AtomicReference<SequentialTask> lastTask;
```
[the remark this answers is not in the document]
cody-littley:
```
    /**
     * The next task to be scheduled will be inserted into this placeholder task. When that happens, a new task will be
     * created and inserted into this placeholder.
     */
    private final AtomicReference<SequentialTask> nextTaskPlaceholder;
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/SequentialTask.java
```diff
@@ -0,0 +1,106 @@
...
+     *                                 for execution. The first task in a sequence has a dependency count of 1 (data
+     *                                 must be provided), and subsequent tasks have a dependency count of 2 (the
+     *                                 previous task must be executed and data must be provided).
+     */
+    SequentialTask(
+            @NonNull final ForkJoinPool pool,
+            @NonNull final ObjectCounter offRamp,
+            @NonNull final FractionalTimer busyTimer,
+            @NonNull final UncaughtExceptionHandler uncaughtExceptionHandler,
+            final int dependencyCount) {
```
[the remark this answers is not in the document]
cody-littley: changed to a boolean

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+ * An object that can insert work to be handled by a {@link TaskScheduler}.
+ *
+ * @param <I> the type of data that passes into the wire
+ * @param <O> the type of the data that comes out of the parent {@link TaskScheduler}'s primary output wire
+ */
+public class InputWire<I, O> {
+
+    private final TaskScheduler<O> taskScheduler;
+    private Consumer<Object> handler;
+    private String name;
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param taskScheduler the wire to insert data into
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,235 @@
...
+     * Forward output data to any wires/consumers that are listening for it.
+     *
+     * @param data the output data to forward
+     */
+    protected void forward(@NonNull final O data) {
+        for (final Consumer<O> destination : forwardingDestinations) {
+            try {
+                destination.accept(data);
+            } catch (final Exception e) {
+                // Future work: if we ever add a name to output wires, include that name in this log message.
```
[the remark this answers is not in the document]
cody-littley: removed comment

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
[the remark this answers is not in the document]
alittley: Very interesting

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,235 @@
...
+            }
+        }
+    }
+
+    /**
+     * Specify an input wire where output data should be passed. This forwarding operation respects back pressure.
+     *
+     * <p>
+     * Soldering is the act of connecting two wires together, usually by melting a metal alloy between them. See
+     * https://en.wikipedia.org/wiki/Soldering
```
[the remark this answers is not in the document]
cody-littley: fixed, forgot to copy-paste the better formatting.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,235 @@
...
+    public OutputWire<O> buildFilter(@NonNull final String name, @NonNull final Predicate<O> predicate) {
+        final WireFilter<O> filter =
+                new WireFilter<>(model, Objects.requireNonNull(name), Objects.requireNonNull(predicate));
+        solderTo(filter.getName(), filter);
+        return filter;
+    }
+
+    /**
+     * Build a {@link WireListSplitter} that is soldered to the output of this wire. Creating a splitter for wires
+     * without a list output type will cause to runtime exceptions.
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,173 @@
...
+
+import com.swirlds.common.wiring.counters.ObjectCounter;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.function.Consumer;
+
+/**
+ * Schedules tasks for a component.
+ *
+ * @param <O> the output type of the primary output wire (use {@link Void}) if no output is needed)
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,173 @@
...
+    protected abstract void put(@NonNull Consumer<Object> handler, @Nullable Object data);
+
+    /**
+     * Add a task to the scheduler. If backpressure is enabled and there is not immediately capacity available, this
+     * method will not accept the data.
+     *
+     * @param data the data to be processed by the scheduler
+     * @return true if the data was accepted, false otherwise
+     */
+    protected abstract boolean offer(@NonNull Consumer<Object> handler, @Nullable Object data);
```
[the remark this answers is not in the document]
cody-littley: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.ForkJoinPool;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * A builder for {@link TaskScheduler}s.
+ *
+ * @param <O> the output time of the primary output wire for this task scheduler (use {@link Void}) for a scheduler with
```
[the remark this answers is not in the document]
cody-littley: fixed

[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+     *             alphanumeric characters and underscores.
+     */
+    TaskSchedulerBuilder(@NonNull final WiringModel model, @NonNull final String name) {
+        this.model = Objects.requireNonNull(model);
+
+        // The reason why wire names have a restricted character set is because downstream consumers of metrics
+        // are very fussy about what characters are allowed in metric names.
+        if (!name.matches("^[a-zA-Z0-9_]*$")) {
+            throw new IllegalArgumentException(
+                    "Task Schedulers name must only contain alphanumeric characters, underscores, and hyphens");
```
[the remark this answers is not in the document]
cody-littley: Fixed error string. Prometheus doesn't deal well with hyphens, so it's better to disallow them in anything that could be included in a metric name.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+     * @return this
+     */
+    @NonNull
+    public TaskSchedulerBuilder<O> withUnhandledTaskCapacity(final long unhandledTaskCapacity) {
+        this.unhandledTaskCapacity = unhandledTaskCapacity;
+        return this;
+    }
+
+    /**
+     * Set whether the task scheduler should enable flushing. Default false. Flushing a wire with this disabled will
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+    @NonNull
+    public TaskSchedulerBuilder<O> withOffRamp(@NonNull final ObjectCounter offRamp) {
+        this.offRamp = Objects.requireNonNull(offRamp);
+        return this;
+    }
+
+    /**
+     * If true then the framework will assume that back pressure is being applied via external mechanisms.
+     * <p>
+     * This method does not change the runtime behavior of the wiring framework. But it does effect cyclical back
```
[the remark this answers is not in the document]
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+     * Figure out which counters to use for this task scheduler (if any), constructing them if they need to be
+     * constructed.
+     */
+    @NonNull
+    private Counters buildCounters() {
+        final ObjectCounter innerCounter;
+
+        if (unhandledTaskCapacity != UNLIMITED_CAPACITY) {
+            innerCounter = new BackpressureObjectCounter(unhandledTaskCapacity, sleepDuration);
+        } else if (metricsBuilder != null && metricsBuilder.isUnhandledTaskMetricEnabled()
```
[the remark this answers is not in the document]
cody-littley:
```
        // If we need to enforce a maximum capacity, we have no choice but to use a backpressure object counter.
        // 
        // If we don't need to enforce a maximum capacity, we need to use a standard object counter if any 
        // of the following conditions are true:
        //  - we have unhandled task metrics enabled
        //  - the scheduler is concurrent and flushing is enabled. This is because the concurrent scheduler's
        //    flush implementation requires a counter that is not a no-op counter.
        //
        // In all other cases, better to use a no-op counter. Counters have overhead, and if we don't need one
        // then we shouldn't use one.
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/WiringModel.java
```diff
@@ -0,0 +1,129 @@
...
+
+import com.swirlds.base.time.Time;
+import com.swirlds.common.context.PlatformContext;
+import com.swirlds.common.wiring.model.StandardWiringModel;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * A wiring model is a collection of wires. It can be used to analyze the wiring of a system and to generate diagrams.
```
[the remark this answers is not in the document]
cody-littley: Yes, I'm pretty sure that's what happened.

```
/**
 * A wiring model is a collection of task schedulers and the wires connecting them. It can be used to analyze the wiring
 * of a system and to generate diagrams.
 */
public abstract class WiringModel {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/WiringModel.java
```diff
@@ -0,0 +1,129 @@
...
+    @NonNull
+    public static WiringModel create(@NonNull final PlatformContext platformContext, @NonNull final Time time) {
+        return new StandardWiringModel(platformContext, time);
+    }
+
+    /**
+     * Get a new task scheduler builder.
+     *
+     * @param name the name of the task scheduler. Used for metrics and debugging. Must be unique (not enforced by
+     *             framework). Must only contain alphanumeric characters, underscores, and hyphens (enforced by
```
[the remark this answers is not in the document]
cody-littley: Fixed javadoc.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/EventSignatureValidationWire.java
cody-littley: renamed to `EventSignatureValidationScheduler`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/OrphanBufferWire.java
cody-littley: renamed to `OrphanBufferScheduler`

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+     * @param <X>       the input type of the input wire
+     * @return this
+     */
+    @SuppressWarnings("unchecked")
+    public <X> InputWire<X, O> withInputType(@NonNull final Class<X> inputType) {
+        return (InputWire<X, O>) this;
+    }
+
+    /**
+     * Bind this input wire to a handler. A handler must be bound to this inserter prior to inserting data via any
```
poulok, line 100:
```suggestion
     * Bind this input wire to a handler. A handler must be bound to this input wire prior to inserting data via any
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param handler the handler to bind to this input wire
+     * @return this
+     * @throws IllegalStateException if a handler is already bound and this method is called a second time
+     */
+    @SuppressWarnings("unchecked")
+    @NonNull
+    public InputWire<I, O> bind(@NonNull final Consumer<I> handler) {
+        if (this.handler != null) {
+            throw new IllegalStateException("Handler already bound");
```
poulok, line 111:
```suggestion
            throw new IllegalStateException("Handler '" + name  + "' is already bound");
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * An object that can insert work to be handled by a {@link TaskScheduler}.
+ *
+ * @param <I> the type of data that passes into the wire
+ * @param <O> the type of the data that comes out of the parent {@link TaskScheduler}'s primary output wire
+ */
+public class InputWire<I, O> {
```
poulok, line 31: I understand why you chose "I" and "O", but in general the letters "O", "I", and "L" should never be used for single letter names because different fonts make then difficult to distinguish from "0", "1", and "l".

I suggest using `IN` and `OUT` or different letters.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
```diff
@@ -0,0 +1,172 @@
...
+        return this;
+    }
+
+    /**
+     * Add a task to the task scheduler. May block if back pressure is enabled.
+     *
+     * @param data the data to be processed by the task scheduler
+     */
+    public void put(@Nullable final I data) {
+        taskScheduler.put(handler, data);
```
poulok, line 148: it would be nice to throw if the input wire is unbound when this is invoked.

```
throwIfUnbound();
taskScheduler.put(handler, data);
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * <p>
+     * Forwarding should be fully configured prior to data being inserted into the wire. Adding forwarding destinations
+     * after data has been inserted into the wire is not thread safe and has undefined behavior.
+     *
+     * @param inputWire the inputWire to forward output data to
+     * @return this
+     */
+    @NonNull
+    public OutputWire<O> injectionSolderTo(@NonNull final InputWire<O, ?> inputWire) {
```
lpetrovic05, line 148: maybe its better if we overload `solderTo` with another parameter for injecting

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
cody-littley: updated

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
cody-littley: renamed to `IN` and `OUT`

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/InputWire.java
cody-littley: I intentionally didn't add nullity checks like this to the framework due to the performance critical nature it will fill. If you aren't convinced, perhaps we can discuss.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
cody-littley: done

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,236 @@
...
+            @NonNull final WiringModel model,
+            @NonNull final String name,
+            final boolean registerWithModel,
+            final boolean insertionIsBlocking) {
+
+        this.model = Objects.requireNonNull(model);
+        this.name = Objects.requireNonNull(name);
+
+        if (registerWithModel) {
+            model.registerVertex(name, insertionIsBlocking);
```
poulok, line 67: Conceptually, it does not make sense to call an output wire a vertex in a graph. The task scheduler + component combo is the vertex.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.function.Consumer;
+
+/**
+ * Schedules tasks for a component.
+ *
+ * @param <OUT> the output type of the primary output wire (use {@link Void} if no output is needed)
+ */
+public abstract class TaskScheduler<OUT> extends OutputWire<OUT> {
```
poulok, line 29: I don't think `TaskScheduler` should extend `OutputWire`. The proper relationship is that a Task Schedule has an output wire, not is an output wire.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
```diff
@@ -0,0 +1,236 @@
...
+     *
+     * @param name      the name of the filter
+     * @param predicate the predicate that filters the output of this wire
+     * @return the filter
+     */
+    @NonNull
+    public OutputWire<OUT> buildFilter(@NonNull final String name, @NonNull final Predicate<OUT> predicate) {
+        final WireFilter<OUT> filter =
+                new WireFilter<>(model, Objects.requireNonNull(name), Objects.requireNonNull(predicate));
+        solderTo(filter.getName(), filter);
```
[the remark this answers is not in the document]
poulok: It may be more clear to users if this method did not exist on its own, but was a variant of `solderTo`, such as `filterAndSolderTo`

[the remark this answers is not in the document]
poulok: Another option is to only support soldering to a single output wire, and make it up to the caller to combine output wires (filters, splitters, transformers) into a single output wire.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+     * <ul>
+     * <li>{@link TaskSchedulerMetricsBuilder#withUnhandledTaskMetricEnabled(boolean)} is called with the value true</li>
+     * <li>{@link TaskSchedulerBuilder#withUnhandledTaskCapacity(long)} is passed a positive value</li>
+     * <li>{@link TaskSchedulerBuilder#withOnRamp(ObjectCounter)} is passed a counter that is not a no op counter</li>
+     * </ul>
+     */
+    public abstract long getUnprocessedTaskCount();
+
+    /**
+     * Flush all data in the task scheduler. Blocks until all data currently in flight has been processed.
```
poulok, line 94: Is this correct?
```suggestion
     * Flush all data in the task scheduler. Blocks until all data currently in flight has been scheduled.
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+     * @return this, cast into whatever type is requested
+     */
+    @NonNull
+    @SuppressWarnings("unchecked")
+    public final <X> TaskScheduler<X> cast() {
+        return (TaskScheduler<X>) this;
+    }
+
+    /**
+     * Get an input wire for passing data to this task scheduler. In order to use this wire, a handler must be bound via
```
poulok, line 69:
```suggestion
     * Build an input wire for passing data to this task scheduler. In order to use this wire, a handler must be bound via
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+     * Get the number of unprocessed tasks. Returns -1 if this task scheduler is not monitoring the number of
+     * unprocessed tasks. Schedulers do not track the number of unprocessed tasks by default. This method will always
+     * return -1 unless one of the following is true:
+     * <ul>
+     * <li>{@link TaskSchedulerMetricsBuilder#withUnhandledTaskMetricEnabled(boolean)} is called with the value true</li>
+     * <li>{@link TaskSchedulerBuilder#withUnhandledTaskCapacity(long)} is passed a positive value</li>
+     * <li>{@link TaskSchedulerBuilder#withOnRamp(ObjectCounter)} is passed a counter that is not a no op counter</li>
+     * </ul>
+     */
+    public abstract long getUnprocessedTaskCount();
```
poulok, line 91: The term "processed" in this context is not defined, and I'm unsure if it means "unscheduled" or if that term applies more to the task based implementation and not the queue based implementation.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+     * <p>
+     * Warning: some implementations of flush may block indefinitely if new work is continuously added to the scheduler
+     * while flushing. Such implementations are guaranteed to finish flushing once new work is no longer being added.
+     * Some implementations do not have this restriction, and will return as soon as all of the in flight work has been
+     * processed, regardless of whether or not new work is being added.
+     *
+     * @throws UnsupportedOperationException if {@link TaskSchedulerBuilder#withFlushingEnabled(boolean)} was set to
+     *                                       false (or was unset, default is false)
+     */
+    public abstract void flush();
```
poulok, line 108: What do you expect this to be used for?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param name the name of the task scheduler. Used for metrics and debugging. Must be unique. Must only contain
+     *             alphanumeric characters and underscores.
+     */
+    TaskSchedulerBuilder(@NonNull final WiringModel model, @NonNull final String name) {
+        this.model = Objects.requireNonNull(model);
+
```
poulok, line 74: Need a non-null check here

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /**
+     * Set whether the scheduler should be concurrent or not. Default false.
+     *
+     * @param concurrent true if the task scheduler should be concurrent, false otherwise
+     * @return this
+     */
+    @NonNull
+    public TaskSchedulerBuilder<O> withConcurrency(boolean concurrent) {
```
poulok, line 94:
```suggestion
    public TaskSchedulerBuilder<O> withConcurrency(final boolean concurrent) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+     *
+     * @return the uncaught exception handler
+     */
+    @NonNull
+    private UncaughtExceptionHandler buildUncaughtExceptionHandler() {
+        if (uncaughtExceptionHandler != null) {
+            return uncaughtExceptionHandler;
+        } else {
+            return (thread, throwable) ->
+                    logger.error(EXCEPTION.getMarker(), "Uncaught exception in wire {}", name, throwable);
```
poulok, line 229:
```suggestion
                    logger.error(EXCEPTION.getMarker(), "Uncaught exception in task scheduler {}", name, throwable);
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: "processed" is actually correct here I think. We will actually wait for all processing to be fully complete before this method returns. Waiting for data to be scheduled implies to me that it may be processed at a future time.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: Added the following to the javadoc:

```
     * Get the number of unprocessed tasks. A task is considered to be unprocessed until the data has been passed to the
     * handler method (i.e. the one given to {@link InputWire#bind(Consumer)}) and that handler method has returned.
```

This is how I would define the lifecycle of a task:

1. unscheduled: task has not yet been added to a scheduler via `inputWire.put(data)` (or equivalent)
2. scheduled but unprocessed: task has been added to the logical queue via `inputWire.put(data)` (or equivalent), but the handle method has not yet been called
3. processed: task has been passed to the handle method and the handle method has returned

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
cody-littley: I dereference `name` via `name.matches("^[a-zA-Z0-9_]*$")`, so if you pass in null it will throw an informative error before we update any data structures.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /**
+     * Set the maximum number of permitted scheduled tasks. Default is unlimited.
+     *
+     * @param unhandledTaskCapacity the maximum number of permitted unhandled tasks
+     * @return this
+     */
+    @NonNull
+    public TaskSchedulerBuilder<O> withUnhandledTaskCapacity(final long unhandledTaskCapacity) {
```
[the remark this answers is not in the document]
cody-littley: We should discuss this, I don't think `unscheduled` is accurately descriptive.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: I will work on this change, not yet completed in the branch.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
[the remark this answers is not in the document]
cody-littley: I don't understand your first comment. I'm not sure how the ordering matters. We should discuss.

> Another option is to only support soldering to a single output wire or a list of wires, and make it up to the caller to combine output wires

The reason for this is unclear to me. When writing tests, it was quite convenient to allow the framework to manage this for me. Let's also discuss this.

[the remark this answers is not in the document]
lpetrovic05: I think the confusing part is that the methods in this class return `this` sometimes, and sometimes they return  `filter` or `splitter`. I feel like `solderTo` should return `void` to avoid this confusion.

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/schedulers/SequentialTaskSchedulerTests.java
```diff
@@ -0,0 +1,1966 @@
...
+            if (x % 2 == 0) {
+                aOutBoolean.forward(x % 3 == 0);
+            }
+
+            if (x % 5 == 0) {
+                aOutString.forward(Integer.toString(x));
+            }
+
+            return x;
+        });
```
lpetrovic05, line 1716: I don't understand this approach for alternative outputs fully. This puts the delegation of tasks inside the business logic. I think a better approach would be to optionally specify the output wire when calling `bind`

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public abstract void offRamp();
+
+    /**
+     * Get the number of objects in the part of the system that this object is being used to monitor.
+     */
+    public abstract long getCount();
+
+    /**
+     * Blocks until the number of objects off-ramped is equal to the number of objects on-ramped.
```
poulok, line 54: If one thread calls this method and is waiting, could another thread on ramp more objects and cause the first thread to wait longer? In other words, if a thread is blocking in this method, does it block other threads from on ramping objects?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/StandardObjectCounter.java
```diff
@@ -0,0 +1,101 @@
...
+    private final AtomicLong count = new AtomicLong(0);
+    private final ManagedBlocker waitUntilEmptyBlocker;
+
+    /**
+     * Constructor.
+     *
+     * @param sleepDuration when a method needs to block, the duration to sleep while blocking
+     */
+    public StandardObjectCounter(@NonNull final Duration sleepDuration) {
+        final int sleepNanos = (int) sleepDuration.toNanos();
```
poulok, line 39: why cast to int?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
```diff
@@ -0,0 +1,90 @@
...
+        this.count = Objects.requireNonNull(count);
+        this.capacity = capacity;
+        this.sleepNanos = sleepNanos;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean block() throws InterruptedException {
```
poulok, line 66: I do not fully understand the documentation on this method and the `isReleasable()` method, but it seems that this method should check if there is capacity, and return `true` if there is.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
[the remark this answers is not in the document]
cody-littley: > I think the confusing part is that the methods in this class return this sometimes, and sometimes they return filter or splitter. I feel like solderTo should return void to avoid this confusion.

Makes sense, change made. `solderTo()` is now a void method.

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/schedulers/SequentialTaskSchedulerTests.java
cody-littley: We discussed this in slack, repeating the conclusion of that discussion here.

I agreed to add this feature on the condition that we first find a use case in our current code or in planned code, and Lazar agreed that this work could be done in a follow up PR.

cody-littley: https://github.com/hashgraph/hedera-services/issues/9598

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/WiringModel.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param groups optional groupings of vertices
+     * @return a mermaid style wiring diagram
+     */
+    @NonNull
+    public abstract String generateWiringDiagram(@NonNull final Set<ModelGroup> groups);
+
+    /**
+     * Reserved for internal framework use. Do not call this method directly.
+     * <p>
+     * Register a vertex in the wiring model. These are either output task schedulers or wire transformers. Vertexes
```
poulok, line 105: The plural of "vertex" is "vertices". There are couple instances of "vertexes"

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
cody-littley: Correct, this is what happens. It's possible to cause this method to block indefinitely if work is continuously on-ramped.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/StandardObjectCounter.java
cody-littley: No particular reason. I've changed this into a `long` type.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
cody-littley: Let's discuss this, it took me a long time to wrap my head around the semantics of how this interface works.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/WiringModel.java
cody-littley: fixed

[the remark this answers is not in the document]
cody-littley: It's the result of a name change partly executed with find and replace. 🙃 This should just be "task scheduler".

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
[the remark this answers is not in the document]
poulok: Sure. There are a few terms being used in different places that are not defined, like "handled" and "processed". I'm unsure if these are relics of old naming or not.

alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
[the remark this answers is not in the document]
poulok: Per my discussion with @Cody, it would also be helpful to document that the `builderFilter` and similar methods for transformer and splitter actually return the output wire of the fitler/transformer/splitter which is a newly created output wire.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
poulok: It would be nice to define these in the class level javadoc

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+        // In all other cases, better to use a no-op counter. Counters have overhead, and if we don't need one
+        // then we shouldn't use one.
+
+        if (unhandledTaskCapacity != UNLIMITED_CAPACITY) {
+            innerCounter = new BackpressureObjectCounter(unhandledTaskCapacity, sleepDuration);
+        } else if ((metricsBuilder != null && metricsBuilder.isUnhandledTaskMetricEnabled())
+                || (concurrent && flushingEnabled)) {
+            innerCounter = new StandardObjectCounter(sleepDuration);
+        } else {
+            innerCounter = null;
```
poulok, line 285: Would we ever want this? An unlimited number of items in flight with no way of tracking the number?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * The maximum desired capacity. It is possible that the current count may exceed this capacity (i.e. if
+     * {@link ObjectCounter#forceOnRamp()} is used to bypass the capacity).
+     */
+    private final long capacity;
+
+    /**
+     * The amount of time to sleep while waiting for capacity to become available, or 0 to not sleep.
+     */
+    private final long sleepNanos;
```
poulok, line 46: which thread does the sleeping?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public void onRamp() {
+        try {
+            count.updateAndGet(increment);
+            // Best case scenario: capacity was immediately available.
+        } catch (final NoCapacityException e) {
+            // Slow case. Capacity wasn't available, so we need to block.
+            try {
+                // This will block until capacity is available and the count has been incremented.
+                ForkJoinPool.managedBlock(onRampBlocker);
```
poulok, line 75: What is actually going on here? I think this is the part you added to ensure we don't get a deadlock if there are more task schedulers than threads. If so, I think it's important to explain this.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,138 @@
...
+            } catch (final InterruptedException ex) {
+                // This should be impossible.
+                Thread.currentThread().interrupt();
+                throw new IllegalStateException("Interrupted while blocking on an onRamp()");
```
poulok, line 79: If we ever do see this error, or the similar one below, it would be extremely useful to know which task scheduler this backpressure object is for. It would be very easy to pass the task scheduler name in here to allow that.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * A utility for counting the number of objects in various parts of the pipeline. Will apply backpressure if the number
+ * of objects exceeds a specified capacity.
+ */
+public class BackpressureObjectCounter extends ObjectCounter {
+
+    private final AtomicLong count = new AtomicLong(0);
+    private final LongUnaryOperator increment;
+
+    private final ManagedBlocker onRampBlocker;
```
poulok, line 35: These two member variables could use javadocs

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,138 @@
...
+    private final ManagedBlocker waitUntilEmptyBlocker;
```
poulok, line 36: These two member variables could use javadocs

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/AbstractTask.java
```diff
@@ -0,0 +1,82 @@
...
+
+package com.swirlds.common.wiring.schedulers;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A unit of work that is handled by a task scheduler.
```
poulok, line 25: ?
```suggestion
 * A unit of work that is processed by a task scheduler.
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
cody-littley: Added javadocs and inline comments.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskSchedulerBuilder.java
cody-littley: This was an ask from Oleg. The idea is that there may be use cases where we don't want to apply back pressure (for example, when we do merkle hashing), and so we may prefer not to pay the overhead of updating the atomic variable.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureBlocker.java
cody-littley: If by thread you mean the OS level concept, which thread is not really defined. The fork join pool may do this on any thread it chooses, and may even change threads in the middle of the block.

From the logical perspective, the blocking behavior IS experienced by the caller/inserter. If scheduler `A` wants to push a task into the logical queue for scheduler `B` but `B` is at maximum capacity, `A` will be forced to wait, and this mechanism will enforce that wait.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley:
```
                // This will block until capacity is available and the count has been incremented.
                // 
                // This is logically equivalent to the following pseudocode.
                // Note that the managed block is thread safe when onRamp() is being called from multiple threads,
                // even though this pseudocode is not.
                //
                //                while (count >= capacity) {
                //                    Thread.sleep(sleepNanos);
                //                }
                //                count++;
                //
                // The reason why we use the managedBlock() strategy instead of something similar has to do with
                // the fork join pool paradigm. Unlike traditional thread pools where we have more threads than
                // CPUs, blocking (e.g. Thread.sleep()) on a fork join pool may monopolize an entire CPU core.
                // The managedBlock() pattern allows us to block while yielding the physical CPU core to other
                // tasks.
                ForkJoinPool.managedBlock(onRampBlocker);
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley:
```
    /**
     * When back pressure needs to be applied due to lack of capacity, this object is used to efficiently sleep on the
     * fork join pool.
     */
    private final ManagedBlocker onRampBlocker;

    /**
     * When waiting for the count to reach zero, this object is used to efficiently sleep on the fork join pool.
     */
    private final ManagedBlocker waitUntilEmptyBlocker;
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/AbstractTask.java
cody-littley: changed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/OutputWire.java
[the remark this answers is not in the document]
cody-littley: added requested documentation

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
cody-littley: change made

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -0,0 +1,167 @@
...
+            @NonNull final String name, final long capacity, @NonNull final Duration sleepDuration) {
+        if (capacity <= 0) {
+            throw new IllegalArgumentException("Capacity must be greater than zero");
+        }
+
+        this.name = Objects.requireNonNull(name);
+
+        increment = count -> {
+            if (count >= capacity) {
+                throw new NoCapacityException();
```
OlegMazurov, line 66: There is no benefit in using an exception here and a lot of cost (both heap/GC and CPU).
It ends up in a busy wait loop anyway (`ForkJoinPool.managedBlock()` -> `onRampBlocker.isReleasable()` -> `count.compareAndSet()`), so, why add costly complexity that will be on every radar screen for real code.
The ideal situation is when the back-pressure mechanism is not invoked (all covered consumers are faster then the producer) but it should not exacerbate the condition, by creating pressure on the GC for one thing, when that's not the case.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/TaskScheduler.java
```diff
@@ -0,0 +1,176 @@
...
```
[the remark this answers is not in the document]
cody-littley: Let's discuss this on a call. This is a surprisingly deep topic.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
poulok: So invoking this method does NOT prevent new items from being added. It may be worth explicitly stating this in the javadoc.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: I've changed my implementation as suggested.

alittley (dismissed):

alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/ObjectCounter.java
cody-littley:
```
    /**
     * Blocks until the number of objects off-ramped is equal to the number of objects on-ramped. Does not prevent
     * new objects from being on-ramped. If new objects are continuously on-ramped, it is possible that this method
     * may block indefinitely.
     */
```

alittley (approved):

lpetrovic05 (approved):

poulok (approved):

Issue #9436, opened 2023-10-24
Support for @Tag in HapiTestEngine


netopyr:
We need to be able to filter our HAPI tests. For instance, we may want to execute long-running tests less frequently or separate flaky tests and make them optional.

Usually, this is achieved by tagging particular test methods with the `@Tag` annotation and filtering based on these tags. We need to add support to our test engine to allow this with our HAPI tests.

PR #9490, opened 2023-10-26
Support for @Tag in HapiTestEngine

vtronkov:
**Note:** - https://github.com/hashgraph/hedera-services/pull/9510 was merged into this PR

**Description**:
- Adding support for `@Tag` annotation in `HapiTestEngine`

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9436

**Notes for reviewer**:
To filter by `@Tag` you need to:
- Annotate a test/tests with `@Tag("TAG_NAME")` annotation
- Open `test-clients/build.gradle.kts` file and add the following config:
```
tasks.register<Test>("hapiTestSlow") {
    testClassesDirs = sourceSets.main.get().output.classesDirs
    classpath = sourceSets.main.get().runtimeClasspath

    useJUnitPlatform {
        includeTags("slow")
        excludeTags("performance")
    }

    // Do not yet run things on the '--module-path'
    modularity.inferModulePath.set(false)
}
```
_(change the `includeTags` and `excludeTags` according to your tests)_

Run the **hapiTestSlow** gradle task from IntelliJ:
Screenshot 2023-11-01 at 13 44 54

or execute `./gradlew hapiTestSlow ` in the console

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -137,11 +143,12 @@ private void appendTestsInClasspathRoot(
     @Override
     protected HapiTestEngineExecutionContext createExecutionContext(ExecutionRequest request) {
         // Populating the data needed for the context
-        return new HapiTestEngineExecutionContext(Path.of("data"), Path.of("eventstreams"));
+        return new HapiTestEngineExecutionContext();
```
vtronkov, line 146: This is no longer needed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngineExecutionContext.java
```diff
@@ -26,21 +25,4 @@
...
-    }
-
-    public Path getSavedStateDirectory() {
-        return savedStateDirectory;
-    }
-
-    public Path getEventsLogDir() {
-        return eventsLogDir;
-    }
-}
```
vtronkov, line 46: This is no longer needed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -243,33 +250,67 @@ public SkipResult shouldBeSkipped(HapiTestEngineExecutionContext context) {
...
+            final var isHapiTestAnnotated = AnnotationSupport.isAnnotated(testMethod, HapiTest.class);
+            final var disabledAnnotation = AnnotationSupport.findAnnotation(testMethod, Disabled.class);
+
+            if (!isHapiTestAnnotated) {
                 return SkipResult.skip(testMethod.getName() + " No @HapiTest annotation");
-            } else if (annotation.isPresent()) {
-                final var msg = annotation.get().value();
+            } else if (disabledAnnotation.isPresent()) {
+                final var msg = disabledAnnotation.get().value();
                 return SkipResult.skip(msg == null || msg.isBlank() ? "Disabled" : msg);
```
vtronkov, line 302: A bit of refactoring here.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -243,33 +250,67 @@ public SkipResult shouldBeSkipped(HapiTestEngineExecutionContext context) {
...
+            // but the current implementation does not support this.
+            final var tagsAnnotation = testMethod.getAnnotation(Tags.class);
+            final var tagAnnotation = testMethod.getAnnotation(Tag.class);
+
+            final var tags = new HashSet<TestTag>();
+            if (tagsAnnotation != null) {
+                tags.addAll(Arrays.stream(tagsAnnotation.value())
+                        .map(t -> TestTag.create(t.value()))
+                        .toList());
+            } else if (tagAnnotation != null) {
```
netopyr, line 282: Does it make sense to support both annotations on a method?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -243,33 +250,67 @@ public SkipResult shouldBeSkipped(HapiTestEngineExecutionContext context) {
...
+                final var msg = disabledAnnotation.get().value();
                 return SkipResult.skip(msg == null || msg.isBlank() ? "Disabled" : msg);
             }
-            return SkipResult.doNotSkip();
+
+            final var tagFilter = System.getenv("TAG_FILTER");
+            TestTag tag = tagFilter != null ? TestTag.create(tagFilter) : null;
+            if (tag == null && testTags.isEmpty()) {
+                return SkipResult.doNotSkip();
+            } else if (testTags.contains(tag)) {
```
netopyr, line 309: I think we will need to support more complex logic. At a minimum, we need to be able to specify "run all tests except the ones that are marked as slow". Ideally we support standard expressions as defined in the [documentation](https://junit.org/junit5/docs/current/user-guide/#running-tests-tag-expressions)

Maybe we can take advantage of a default implementation like [TagFilter](https://junit.org/junit5/docs/5.2.0/api/org/junit/platform/launcher/TagFilter.html)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -243,33 +250,67 @@ public SkipResult shouldBeSkipped(HapiTestEngineExecutionContext context) {
...
                 return SkipResult.skip(testMethod.getName() + " No @HapiTest annotation");
-            } else if (annotation.isPresent()) {
-                final var msg = annotation.get().value();
+            } else if (disabledAnnotation.isPresent()) {
+                final var msg = disabledAnnotation.get().value();
                 return SkipResult.skip(msg == null || msg.isBlank() ? "Disabled" : msg);
             }
-            return SkipResult.doNotSkip();
+
+            final var tagFilter = System.getenv("TAG_FILTER");
```
netopyr, line 305: Does this approach work with the default gradle behavior as described in the [documentation](https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle), e.g.
```
test {
    useJUnitPlatform {
        includeTags("fast", "smoke & feature-a")
        // excludeTags("slow", "ci")
    }
}
```

vtronkov:
Okay, I played a bit with a randomly picked gradle task:
```
tasks.register<Test>("hammerTest") {
    testClassesDirs = sources.output.classesDirs
    classpath = sources.runtimeClasspath

    shouldRunAfter(tasks.test)

    useJUnitPlatform { includeTags("HAMMER") }
    maxHeapSize = "8g"
    jvmArgs("-XX:ActiveProcessorCount=16")
}
```

If I annotate `EndToEndPackageRunner` with _@HapiTestSuite_ then create a test method in it:
```
@Tag("HAMMER")
@Tag("contract.traceability")
@TestFactory
@HapiTest
Collection<DynamicContainer> hammer() {
    return List.of(extractSpecsFromSuite(ContractCreateSuite::new));
}
```

I can see that out of the box, we are calling the `HapiTestEngine` and it looks like we have the tags support. However, immediately after that we call `JupiterTestEngine` and we execute the test the old way. This is because the Gradle task is using the _jUnit_ instead of our custom implementation.

I think we should have a story to first have the Gradle task to use the new HapiTestEngine in place and then I can test our engine and see what we support out of the box and what we should add.

Maybe @jjohannes or @nathanklick work on that?(I see you guys worked on these gradle tasks)

What do you guys think? @netopyr

jsync-swirlds:
The trial you ran is flawed.
Hammer tests already have specific logic for running them, and whole suite extract methods (like the hammer method) cannot be annotated HapiTest because they won't work correctly with that engine (the engine won't find any valid tests to run).


A better trial would involve adding tags to existing HapiTest methods and copying the `hapiTest` block in gradle to exclude/include tags.

Perhaps try this setup (after tagging a few individual tests with "slow" and "unstable"):
```
// Existing hapiTest task
tasks.register<Test>("hapiTest") {
    testClassesDirs = sourceSets.main.get().output.classesDirs
    classpath = sourceSets.main.get().runtimeClasspath

    useJUnitPlatform {
        excludeTags("slow", "unstable")
    }

    // Do not yet run things on the '--module-path'
    modularity.inferModulePath.set(false)
}

// New task for slow/unstable tests
tasks.register<Test>("hapiSlowTest") {
    testClassesDirs = sourceSets.main.get().output.classesDirs
    classpath = sourceSets.main.get().runtimeClasspath

    useJUnitPlatform {
        includeTags("slow", "unstable")
    }

    // Do not yet run things on the '--module-path'
    modularity.inferModulePath.set(false)
}
```

vtronkov:
@jsync-swirlds I misled you a bit. I was not actually using the HAMMER task - I just copied it and created a new one(HAMMER2). Nonetheless thanks for pointing out the existing hapiTest task - it helped debugging the issue.
The main issue was that I didn't override the `getTags` method and the default one was called(which returns an empty set).

Not it should work 😄

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
vtronkov: I think it makes sense, yes. For example we might have a test that we can annotate with **slow** and **performance** tags at the same time

netopyr (approved):

Issue #9431, opened 2023-10-24
Conditional, supported records of internal `CONTRACT_CREATION`'s (including finalizations of existing hollow accounts as contracts via `CREATE2`).


tinker-michaelj:
Validate the below happy path scenarios:
 - [ ] A `CREATE1` record matches the corresponding mono-service record.
 - [ ] A `CREATE2` record matches the corresponding mono-service record.
 - [ ] A `CREATE2` hollow account completion as contract matches the corresponding mono-service record.

And the failure scenarios:
 - [ ] A `CREATE1` in the dynamic context of a reverted `MessageFrame` generates no record.
 - [ ] A `CREATE2` in the dynamic context of a reverted `MessageFrame` generates no record.
 - [ ] A `CREATE2` hollow account completion in the dynamic context of a reverted `MessageFrame` generates no record.

JivkoKelchev:
Validated path scenarios:
	•	A CREATE1 record matches the corresponding mono-service record.
	CreateOperationSuite.simpleFactoryWorks()


	•	A CREATE2 record matches the corresponding mono-service record. 
	CreateOperationSuite.create2FactoryWorksAsExpected()


	•	A CREATE2 hollow account completion as contract matches the corresponding mono-service record.
	CreateOperationSuite.canMergeCreate2ChildWithHollowAccount() 


And the failure scenarios:
	•	A CREATE1 in the dynamic context of a reverted MessageFrame generates no record.
	CreateOperationSuite.resetOnFactoryFailureWorks()

	•	A CREATE2 in the dynamic context of a reverted MessageFrame generates no record.
	CreateOperationSuite.create2FactoryWorksAsExpected() This scenario first test failure, then success

	•	A CREATE2 hollow account completion in the dynamic context of a reverted MessageFrame generates no record.
	CreateOperationSuite.canMergeCreate2ChildWithHollowAccount() This scenario first test failure, then success

**All tests are passing, but still needs adjustment to pass FUZZY_MATCH_AGAINST_HAPI_TEST_STREAMS**



Validated child sidecars scenarios:

	•	A CREATE2 record matches the corresponding mono-service sidecars.
	TraceabilitySuite.traceabilityE2EScenario15() This scenario use CREATE2 DEPLOY

	•	A CREATE2 hollow account completion as contract matches the corresponding mono-service sidecars.
	TraceabilitySuite.hollowAccountCreate2MergeExportsExpectedSidecars()

**TODO:**
[...]

PR #9498, opened 2023-10-26
09394 fix failing tests in create2 operation suite

### LINKED ISSUE #9394: Fix any failing tests left in Create2OperationSuite

JivkoKelchev:
**Description**:

This PR enables most of the tests that are left in `Create2OperationSuit`. It resolves problems that are related to missing or incorrect child records.

There are three significant modifications:
- Update contract create child record, to match record fields in mono.
- Add `ContractCreateRecordBuilder` when hollow account is merged.
- Add a way to revert child records on frame message revert.

**Related issue(s)**:

Fixes #9394
This modification also overlaps with #9431 

**Notes for reviewer**:

The first thing I tried was to build `ContractCreateTransactionBody` instead of `CryptoCreateTransactionBody`, and dispatch a child transaction. But this way I had problems with initCode and charging the parent contract instead of the payer, even if the payer was set correctly in the context of the child transaction.
So I decided to stick with `CryptoCreateTransaction` and add the missing fields in the recordBuilder.

**Checklist**


netopyr (approved):
LGTM in regards to `hedera-app`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -298,4 +318,18 @@ private void dispatchAndMarkCreation(
 
         tokenServiceApi.markAsContract(accountId, autoRenewAccountId);
     }
+
+    public void externalizeHollowAccountMerge(@NonNull ContractID contractId, @Nullable Bytes evmAddress) {
+        var recordBuilder = context.addRemovableChildRecordBuilder(ContractCreateRecordBuilder.class);
+        recordBuilder
+                .contractID(contractId)
+                // add dummy transaction, because SingleTransactionRecord require NonNull on build
+                .transaction(Transaction.newBuilder()
```
tinker-michaelj, line 327: I think this will need to be a synthetic `ContractUpdate`; but that will be much easier to test once [this PR](https://github.com/hashgraph/hedera-services/pull/9513) is merged.

tinker-michaelj (approved):
LGTM, very nice work @JivkoKelchev 🙏

Issue #9427, opened 2023-10-24
Unconditional, enriched records of read-only HTS system contract calls.


tinker-michaelj:
Validate the below happy path scenarios:
 - [ ] For a respectable sampling of (5-6 different) read-only system contracts, the synthetic `ContractCallResult`s found on their success records match those from mono-service.

And the failure scenarios:
 - [ ] For a respectable sampling of (5-6 different) of read-only system contracts, the synthetic `ContractCallResult`s found on their unsuccessful records match those from mono-service.

MiroslavGatsanoga:
https://github.com/hashgraph/hedera-services/pull/9509 covered the following scenarios:
- TokenInfoHTSSuite
    - Happy path scenarios covered
      - HAPI_GET_TOKEN_INFO
      - HAPI_GET_FUNGIBLE_TOKEN_INFO
      - HAPI_GET_NON_FUNGIBLE_TOKEN_INFO
      - HAPI_GET_TOKEN_CUSTOM_FEES
    - No failure scenarios asserted
 - DefaultTokenStatusSuite
   - Happy path scenarios covered
     - GET_TOKEN_DEFAULT_FREEZE_STATUS
     - GET_TOKEN_DEFAULT_KYC_STATUS
   - No failure scenarios asserted

ilko-iliev-lime:
Read-only calls covered: 

- TypeCheckCall: Issues fixed in the [following PR](https://github.com/hashgraph/hedera-services/pull/9844).
- IsTokenCall: Issues fixed in the [following PR](https://github.com/hashgraph/hedera-services/pull/9844).
- IsKycCall: No issues found
- IsFrozenCall: No issues found   
- GetApprovedCall: No issues found 
- GetAllowanceCall: Missing the INVALID_ACCOUNT_ID status on both a static and non-static calls. Mod output doesn't match with mono.

PR #9509, opened 2023-10-26
Fix TokenInfoHTSSuite tests

MiroslavGatsanoga:
**Description**:

Fix child record issues related to token info precompile,  making `TokenInfoHTSSuite` passing as a result.

Also enable a couple of tests in `DefaultTokenStatusSuite` since changes to read-only precompiles are related to them.

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/9427

**Notes for reviewer**:

**Checklist**


tinker-michaelj (changes requested):
Looks great @MiroslavGatsanoga, very cool to see these tests passing. 🙌  

I don't think we actually need to pass in the `MessageFrame` to these `HtsCall` implementations, however.
  1. The `ProxyWorldUpdater`'s method to externalize system contract result just [directly calls an `Enhancement` method](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java#L452); and the `HtsCall` implementations already have access to the `Enhancement`.
  2. For mono-service compatibility, we don't want to skip externalizing the result for _any_ `static` frame, but _only_ if we are in a `ContractCallLocalQuery` instead of a HAPI transaction; compare [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java#L208) in mono-service.

I think we just need to make the `QuerySystemContractOperations` method [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/QuerySystemContractOperations.java#L108) a no-op; and then remove the `MessageFrame` from the PR. 🤔

MiroslavGatsanoga:
> Looks great @MiroslavGatsanoga, very cool to see these tests passing. 🙌
> 
> I don't think we actually need to pass in the `MessageFrame` to these `HtsCall` implementations, however.
> 
> 1. The `ProxyWorldUpdater`'s method to externalize system contract result just [directly calls an `Enhancement` method](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java#L452); and the `HtsCall` implementations already have access to the `Enhancement`.
> 2. For mono-service compatibility, we don't want to skip externalizing the result for _any_ `static` frame, but _only_ if we are in a `ContractCallLocalQuery` instead of a HAPI transaction; compare [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java#L208) in mono-service.
> 
> I think we just need to make the `QuerySystemContractOperations` method [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/QuerySystemContractOperations.java#L108) a no-op; and then remove the `MessageFrame` injections. 🤔

1. Ah, missed that I can use `Enhancement` directly instead of `ProxyWorldUpdater`, will update it with that and remove the injected `MessageFrame`s.
2. I think I see why this will work, just want to check my understanding: Making `QuerySystemContractOperations` method a no-op (which basically means to skip it) instead of checking something like `proxyUpdater.isInTransaction()` as in mono will work because of the `@QueryScope` annotation which is meant for the `ContractCallLocalQuery` case. Right?

Will push the updates shortly, tyvm @tinker-michaelj!

tinker-michaelj (approved):
LGTM, ty @MiroslavGatsanoga

Issue #9437, opened 2023-10-24
Support for @Order in HapiTestEngine


netopyr:
Some suites require that test methods be executed in a particular order. To enforce this, the junit framework provides the `@Order` annotation in combination with the `@TestMethodOrder` annotation. To use this functionality, we need to add support to our test engine.

vtronkov:
Merged in - https://github.com/hashgraph/hedera-services/pull/9490

PR #9510, opened 2023-10-26
Support for @Order in HapiTestEngine

vtronkov:
**Description**:
- Add support for `@Order` annotation in `HapiTestEngine`

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9437

**Notes for reviewer**:
At first, I wanted to integrate it somehow with jUnit and make it work out of the box. Looking at how it was implemented in `JupiterTestEngine` they are using `DiscoverySelectorResolver` and inside it, there is a `MethodOrderingVisitor`. Unfortunately, these methods are internal and I can't reuse them. So I had to make our custom implementation.

As far as I can understand our needs we don't need to support all the **MethodOrderer** types so for now we are supporting only `MethodOrderer.OrderAnnotation` - that should be sufficient for now.

This is **before**:
image

This is **after**:
image

_(don't mind the failing tests, they are unrelated)_

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -212,6 +216,18 @@ public ClassTestDescriptor(Class<?> testClass, TestDescriptor parent, EngineDisc
             this.testClass = testClass;
             setParent(parent);
 
+            // Currently we support only ASC MethodOrderer.OrderAnnotation sorting
+            final var sort = testClass.getAnnotation(TestMethodOrder.class) != null
+                    && testClass.getAnnotation(TestMethodOrder.class).value() == MethodOrderer.OrderAnnotation.class;
+
+            Comparator<Method> noSorting = (m1, m2) -> 0;
```
netopyr, line 223: I believe these `Comparator`s can be declared `static`.

netopyr (approved):
LGTM

PR #9519, opened 2023-10-26
use Bytes.EMPTY instead of null to indicate empty keys/values

### LINKED ISSUE #9517: Slot values should never be null
### Description

Replace all slot values and keys to use Bytes.EMPTY to indicate empty values

### Steps to reproduce

See above

### Additional context

_No response_

### Hedera network

other

### Version

v0.43

### Operating system

None

lukelee-sl:
**Description**:
Fix broken HAPI tests by using Bytes.EMPTY instead of null

**Related issue(s)**:

Fixes #9517

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
@@ -30,7 +30,7 @@ import org.apache.tuweni.units.bigints.UInt256;
  *
  * @param key the key of the access
  * @param value the value read or overwritten
- * @param writtenValue if not null, the overwriting value
+ * @param writtenValue if not Bytes.EMPTY, the overwriting value
  */
 public record StorageAccess(@NonNull UInt256 key, @NonNull UInt256 value, @Nullable UInt256 writtenValue) {
     public StorageAccess {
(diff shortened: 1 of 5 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
```diff
@@ -30,7 +30,7 @@
  *
  * @param key the key of the access
  * @param value the value read or overwritten
- * @param writtenValue if not null, the overwriting value
+ * @param writtenValue if not Bytes.EMPTY, the overwriting value
```
david-bakin-sl, line 33: but also the `@Nullable` below and be sure to add the `requiresNonNull` check!

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/api/TokenServiceApi.java
```diff
@@ -140,10 +140,10 @@ void assertValidStakingElection(
      * Updates the storage metadata for the given contract.
      *
      * @param accountId the id of the contract
-     * @param firstKey       the first key in the storage linked list, null if the storage is empty
+     * @param firstKey       the first key in the storage linked list, empty if the storage is empty
```
david-bakin-sl, line 143: possibly prefer explicit `Bytes.EMPTY` to "empty"

david-bakin-sl (changes requested):
Initial look

tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

david-bakin-sl (approved):
Leaving comments to be addressed in (immediately) subsequent PR in order to unblock build

PR #9521, opened 2023-10-26
address comments from pr #9519

### LINKED ISSUE #9517: Slot values should never be null

lukelee-sl:
**Description**:
There were a couple of code review comments from pr #9519 that need to be addressed

**Related issue(s)**:

Fixes #9517

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
@@ -30,7 +30,7 @@ import org.apache.tuweni.units.bigints.UInt256;
  *
  * @param key the key of the access
  * @param value the value read or overwritten
- * @param writtenValue if not Bytes.EMPTY, the overwriting value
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/StorageAccess.java
```diff
@@ -30,7 +30,7 @@
  *
  * @param key the key of the access
  * @param value the value read or overwritten
- * @param writtenValue if not Bytes.EMPTY, the overwriting value
+ * @param writtenValue if not null, the overwriting value
```
lukelee-sl, line 33: undoing this change from the previous pr as the `writtenValue` _can_ be null if the `StorageAccess` was for a read.

david-bakin-sl (approved):

tinker-michaelj (approved):
LGTM

kimbor (approved):

PR #9535, opened 2023-10-26
Enable `HollowAccountFinalizationSuite`

### LINKED ISSUE #9362: Start testing HollowAccountFinalizationSuite
- [x] hollowAccountCompletionWithCryptoTransfer
- [x] hollowAccountCompletionWithContractCreate
- [x] hollowAccountCompletionWithContractCall
- [x] hollowAccountCompletionWithTokenAssociation
- [x] hollowAccountCompletionWithTokenTransfer
- [x] hollowAccountCompletionViaNonReqSigIsNotAllowed
- [x] hollowAccountCompletionWhenHollowAccountSigRequiredInOtherReqSigs
- [x] tooManyHollowAccountFinalizationsShouldFail
- [x] completedHollowAccountsTransfer
- [x] hollowAccountFinalizationWhenAccountNotPresentInPreHandle
- [x] hollowAccountFinalizationOccursOnlyOnceWhenMultipleFinalizationTensComeInAtTheSameTime
- [x] txnWith2CompletionsAndAnother2PrecedingChildRecords
- [x] hollowPayerAndOtherReqSignerBothGetCompletedInASingleTransaction
- [x] hollowAccountCompletionIsPersistedEvenIfTxnFails
- [x] precompileTransferFromHollowAccountWithNeededSigFailsAndDoesNotFinalizeAccount

PR #9419

Neeharika-Sompalli:
Fixes #9362 

- Enabled all tests except 2 tests in `HollowAccountFinalizationSuite`

**Child Records Issues:** 
- Fixes the way preceding child records are recognized. Since they don't have parentConsensusTimestamp set, we need to be able to add to the history of current user transaction based on [the nonce > 0](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java#L237)
- Sorted the child records to be [stored as per increasing value of nonce](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java#L187), so they are returned same way for `getTxnRecord` query.
- Since in mono-service preceding child transaction are interpreted only by the way records are arranged, the check to see if state has changed causes all preceding transactions to fail. [So commented them out](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java#L450) ( Open for discussion on this change)

**Hollow Account Finalization:** 
- When any account is required to sign, [check if the account is hollow and add them to ](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java#L258)`PreHandlContext.requiredHollowAccounts` which will be available in preHandleResult. [Same for the payer](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleWorkflowImpl.java#L236)
- In the [Handle Workflow finalize all required hollow accounts](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L433) that were required to sign for the transaction. Gets the key corresponding to the alias from key verifier and dispatches a synthetic  _preceding_ `CryptoUpdate` to update the key.
- In CryptoTransfer logic when there is a debit and if the debited account is hollow, it is possible that the account has signed the transaction. If so [add it](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java#L302) to the `requiredHollowAccounts` and finalize in `HandleWorkflow`

**Misc:** 

- Found that we fail everytime we run any transaction on genesis start with `MAX_CHILD_RECORDS_EXCEEDED` because we are adding ~700 child records for system account creation. Added an enum [PrecedingTransactionCategory](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java#L646) to add exception for the case. This will be an exception for genesis records and staking update synthetic transaction that happen every midnight. 
- Parsing an evmAlias using `Key.PROTOBUF.parse()` causing `BufferUnderFlowException`. [Fixed it by catching it ](https://github.com/hashgraph/hedera-services/blob/5d5a34ae97334d102aad49d19de638b66a99654b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/ReadableAccountStoreImpl.java#L157)and also not parsing evmAliases

**NOTE:** 
Two tests are not enabled `hollowAccountCompletionIsPersistedEvenIfTxnFails`, `txnWith2CompletionsAndAnother2PrecedingChildRecords` because in mono-service the hollow account finalizations are committed even if the transaction fails.  Do we want to retain that in mod-code ?


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
```diff
@@ -71,6 +71,8 @@
  * multiple handlers.
  */
 public class TokenHandlerHelper {
+    private static final int STAKING_REWARD_ACCOUNT = 800;
```
netopyr, line 74: I believe this must be read from the configuration.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -444,17 +445,21 @@ public <T> T dispatchPrecedingTransaction(
...
         }
-        if (stack.depth() > 1) {
-            throw new IllegalStateException(
-                    "Cannot dispatch a preceding transaction when a savepoint has been created");
-        }
-
-        if (current().isModified()) {
-            throw new IllegalStateException("Cannot dispatch a preceding transaction when the state has been modified");
-        }
+        // In mono-service preceding records is just the way we arrange the records,
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Because we charge fees before this and the ACCOUNTS state has modifications. 🙂 . So we fail here when we see `current().isModified()`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
Neeharika-Sompalli: Ah yes agreed. Will change

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -225,7 +231,10 @@ private void addToInMemoryCache(
         // And all transactions, regardless of the type, are added to the payer-reverse-index, so that queries of
         // the payer account ID will return all transactions they paid for.
         final var txId = transactionRecord.transactionIDOrThrow();
-        final var isChildTx = transactionRecord.hasParentConsensusTimestamp();
+        // For the preceding child records parentConsensusTimestamp is not set, but the nonce will be greater than 1
+        // For the following child records parentConsensusTimestamp is also set. So to differentiate child records
+        // from user records, we check if the nonce is greater than 0.
+        final var isChildTx = transactionRecord.hasParentConsensusTimestamp() || txId.nonce() > 0;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: 😄

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -181,9 +182,14 @@ public void add(
         final var queue = getQueue();
         final var firstRecord = transactionRecords.get(0);
         removeExpiredTransactions(queue, firstRecord.transactionRecord().consensusTimestampOrElse(Timestamp.DEFAULT));
-
+        // Sort the transaction records based on nonce, so that the user transaction is always first,
+        // followed by any preceding transactions, followed by any child transactions.
+        final var sortedRecords = transactionRecords.stream()
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Noo, the preceding records order was different. Like nonce 3 was given first when I do `getTxnRecord`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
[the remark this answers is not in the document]
Neeharika-Sompalli: And if I move the `finalizeHollowAccounts` all the way up. Then because we dispatch hollow accounts finalizations before autoCreations, it fails while dispatching preceding auto-creation transaction. 😞

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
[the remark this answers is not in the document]
tinker-michaelj: @Neeharika-Sompalli I believe `RecordListBuilder` orders the list by consensus time [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java#L293)---which unfortunately is not the same as increasing `nonce`. 😞 

However, I would lean toward putting the sort [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetRecordHandler.java#L125) in the query response handler, since _functionally_ that's the only place we need it to be ordered by `nonce`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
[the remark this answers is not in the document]
tinker-michaelj: In Services we have always referred to both preceding and following as "child records"; I think mirror node is a bit more rigorous in thinking of children as **only** "records having a [non-empty `parent_consensus_timestamp` field](https://github.com/hashgraph/hedera-protobufs/blob/main/services/transaction_record.proto#L116)".

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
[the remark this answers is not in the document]
Neeharika-Sompalli: I see. Thanks @tinker-michaelj ! 👍

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -642,4 +647,9 @@ public SavepointStack savepointStack() {
     public void revertChildRecords() {
         recordListBuilder.revertChildrenOf(recordBuilder);
     }
+
+    public enum PrecedingTransactionCategory {
+        GENESIS,
```
tinker-michaelj, line 652: Since this is also used by `EndOfStakingPeriodUpdater`, and is used to decide whether we enforce the child records limit, maybe enum values like `LIMITED_CHILD_RECORDS` and `UNLIMITED_CHILD_RECORDS` would be more informative? 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -690,6 +747,8 @@ private PreHandleResult addMissingSignatures(
         signatureExpander.expand(sigPairs, expanded);
         if (payerKey != null && !isHollow(payer)) {
             signatureExpander.expand(payerKey, sigPairs, expanded);
+        } else if (isHollow(payer)) {
```
tinker-michaelj, line 750: Nice... 🙌 Surprised we missed this before, haha 😵

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -690,6 +747,8 @@ private PreHandleResult addMissingSignatures(
...
+            context.requireSignatureForHollowAccount(payer);
         }
         signatureExpander.expand(currentRequiredPayerKeys, sigPairs, expanded);
```
tinker-michaelj, line 753: Just noticed there's a typo in this name `currentRequiredPayerKeys`---it's actually the required **non**-payer keys haha (c.f. [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L675)).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -476,6 +488,51 @@ private void handleUserTransaction(
...
+        final var maxRecords = consensusConfig.handleMaxPrecedingRecords();
+        // If the hollow accounts that need to be finalized is greater than the max preceding
+        // records allowed throw an exception
+        if (precedingHollowAccountRecords >= maxRecords) {
+            throw new HandleException(MAX_CHILD_RECORDS_EXCEEDED);
+        } else {
+            for (final var hollowAccount : accounts) {
+                // get the verified key for this hollow account
+                final var verification = verifier.verificationFor(hollowAccount.alias());
+                if (verification.key() != null) {
```
tinker-michaelj, line 517: Wouldn't it be a catastrophic invariant failure if we got this far with a required hollow account having `null` verification?

Imo we can replace line 516 with,
```
final var verification = Objects.requireNonNull(verifier.verificationFor(hollowAccount.alias()), "Required hollow account verified signature did not exist");
```
...because in any case this should be impossible?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -278,6 +289,14 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
         for (int i = count - 1; i >= into; i--) {
             childRecordBuilders.remove(i);
         }
+
+        // If there are preceding child records somehow are added that are more than allowed preceding records,
+        // then we need to revert them as well.
+        // TODO: need to discuss if this is needed with team.
+        precedingTxnRecordBuilders.stream()
```
tinker-michaelj, line 296: I feel like we don't need the `skip(3)` here; and actually want destructive filtering that deletes the `removable()` record builders like,
```
precedingTxnRecordBuilders  = precedingTxnRecordBuilders
    .filter(Predicate.not(SingleTransactionRecordBuilderImpl::removable))
    .map(child -> child.status(child.status() == SUCCESS ? REVERTED_SUCCESS : child.status()))
    .toList();
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
Neeharika-Sompalli: Agreed. Changed those. Thanks @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -290,7 +300,7 @@ public PreHandleContext requireKeyOrThrow(
         }
 
         // Verify this key isn't for an immutable account
-        verifyIsNotImmutableAccount(key, responseCode);
+        verifyNotStakingAccounts(account.accountIdOrThrow(), responseCode);
```
tinker-michaelj, line 303: Now that you have correct checks for hollow accounts (🙌), can we do this check **only** inside the `if (isHollow(account))` blocks? (Since any non-hollow account is definitely not `0.0.800` or `0.0.801`.)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -476,6 +488,51 @@ private void handleUserTransaction(
...
+                    if (!IMMUTABILITY_SENTINEL_KEY.equals(hollowAccount.keyOrThrow())) {
+                        logger.error("Hollow account {} has a key other than the sentinel key", hollowAccount);
+                        return;
```
iwsimon, line 520: Should we throw a HandleException instead of just return? We would have to check the log to find out what went wrong.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
Neeharika-Sompalli: haha yes 😄 Changed. Thank you!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
Neeharika-Sompalli: Agreed. Changed as suggested

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
Neeharika-Sompalli: Makes sense ...Fixed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Agreed 🙌 . Fixed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
Neeharika-Sompalli: yes makes sense. Not sure on the response code 🤔

iwsimon: Maybe KEY_REQUIRED?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
```diff
@@ -680,11 +690,14 @@ private HapiSpec completedHollowAccountsTransfer() {
...
 
+    //    @HapiTest We need a way to differentiate between hollow created records and other preceding records.
+    // In mono-service we commint hollow accounts even if transaction fails, but in mod service we don't.
+    // So we can't enable this test
     private HapiSpec txnWith2CompletionsAndAnother2PrecedingChildRecords() {
         final var ecdsaKey2 = "ecdsaKey2";
         final var recipientKey = "recipient";
         final var recipientKey2 = "recipient2";
-        return defaultHapiSpec("txnWith2CompletionsAndAnother2PrecedingChildRecords")
+        return onlyDefaultHapiSpec("txnWith2CompletionsAndAnother2PrecedingChildRecords")
```
iwsimon, line 700: debug change?

Neeharika-Sompalli: yes, fixed. Thanks @iwsimon

iwsimon (approved):
LGTM

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ConsensusConfig.java
```diff
@@ -24,7 +24,7 @@
 public record ConsensusConfig(
         @ConfigProperty(value = "message.maxBytesAllowed", defaultValue = "1024") @NetworkProperty
                 int messageMaxBytesAllowed,
-        @ConfigProperty(value = "message.maxPrecedingRecords", defaultValue = "950") @NetworkProperty
+        @ConfigProperty(value = "message.maxPrecedingRecords", defaultValue = "3") @NetworkProperty
```
mhess-swl, line 27: Won't this throw an exception for any genesis boot? (Since system accts have to be created)

Neeharika-Sompalli: That is the correct value in mono-service.  I have added a condition `PrecedingTransactionCategory` enum to differentiate genesis case in the same PR

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
```diff
@@ -802,6 +813,9 @@ private HapiSpec hollowPayerAndOtherReqSignerBothGetCompletedInASingleTransactio
                 }));
     }
 
+    @HapiTest
+    //    This is possible in mono-service because we mutate state before handle,
+    //    but can't be possible in modular-services
```
iwsimon, line 818: You enabled the test, should this comment be removed?

tinker-michaelj (dismissed):
LGTM!

iwsimon (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -353,18 +363,21 @@ public PreHandleContext requireKeyIfReceiverSigRequired(
         if (!account.receiverSigRequired()) {
             return this;
         }
-
+        // If it is hollow account, and we require this to sign, we need to finalize the account
```
mhess-swl, line 366: It would be nice to collapse as much as possible of `requireKeyOrThrow ` and `requireKeyIfReceiverSigRequired` into a common method so we don't have to duplicate any of the logic across both methods, but not a blocker for this PR

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -495,14 +496,19 @@ public <T> T doDispatchPrecedingTransaction(
...
         if (stack.depth() > 1) {
             throw new IllegalStateException(
                     "Cannot dispatch a preceding transaction when a savepoint has been created");
         }
 
-        if (current().isModified()) {
-            throw new IllegalStateException("Cannot dispatch a preceding transaction when the state has been modified");
-        }
+        // This condition fails, because for auto-account creation we charge fees, before dispatching the transaction,
+        // and the state will be modified.
```
mhess-swl, line 506:
```suggestion
        // This condition fails, because for auto-account creation we charge fees, before dispatching the transaction, 
        // and the state will be modified. FUTURE: revisit this design
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -249,6 +259,9 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
         if (childRecordBuilders == null) {
             childRecordBuilders = new ArrayList<>();
         }
+        if (precedingTxnRecordBuilders == null) {
+            precedingTxnRecordBuilders = new ArrayList<>();
+        }
```
mhess-swl, line 264: Seems we could instantiate `precedingTxnRecordBuilders` as a list on object creation, but this is fine

mhess-swl (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
Neeharika-Sompalli: yes added a TODO

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Agreed. Could not find a way that time.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
Neeharika-Sompalli: Agreed

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/AliasUtils.java
```diff
@@ -200,11 +206,21 @@ public static Key asKeyFromAliasPreCheck(@NonNull final Bytes alias) throws PreC
     @Nullable
     public static Key asKeyFromAliasOrElse(@NonNull final Bytes alias, @Nullable final Key def) {
         requireNonNull(alias);
+        // If the alias is an evmAddress we don't need to parse with Key.PROTOBUF.
+        // This will cause BufferUnderflowException
+        if (!isAliasSizeGreaterThanEvmAddress(alias)) {
+            return null;
```
mhess-swl, line 212: Should this be `return def` instead? Since the method name is `...OrElse`?

iwsimon (dismissed):
LGTM

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/AliasUtils.java
Neeharika-Sompalli: yes agreed Matt. Thanks . Fixed

iwsimon (dismissed):
LGTM!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
tinker-michaelj: DevOps has alerts on all `ERROR` logs so we would be notified of this.

But imo we would throw an `IllegalStateException` here because this should be impossible ([`isHollow()`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/HapiUtils.java#L92) only classifies an account as "hollow" if its key is the `IMMUTABILITY_SENTINEL_KEY`).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -290,17 +290,21 @@ public PreHandleContext requireKeyOrThrow(
         if (account == null) {
             throw new PreCheckException(responseCode);
         }
-
+        // If it is hollow account, and we require this to sign, we need to finalize the account
+        // with the corresponding ECDSA key in handle
+        if (isHollow(account)) {
+            // Verify this key isn't for an immutable account
+            verifyNotStakingAccounts(account.accountIdOrThrow(), responseCode);
```
tinker-michaelj, line 297: ⚠️ [`isHollow()`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/HapiUtils.java#L94) has been changed to require account num > 1000,
```
        return (account.accountIdOrThrow().accountNum() > 1000
                && account.keyOrElse(EMPTY_KEY_LIST).equals(EMPTY_KEY_LIST)
                ...
```
which means now `isHollow(0.0.800) == false` and this `verifyNotStakingAccounts()` is a no-op.

So it seems we should move `verifyNotStakingAccounts()` below this `if` block.
```
if (isHollow(account)) {
    requiredHollowAccounts.add(account);
    return this;
}
verifyNotStakingAccounts(account.accountIdOrThrow(), responseCode);

```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -318,17 +322,20 @@ public PreHandleContext requireKeyOrThrow(
...
```
tinker-michaelj, line 329: As above.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -353,7 +360,14 @@ public PreHandleContext requireKeyIfReceiverSigRequired(
         if (!account.receiverSigRequired()) {
             return this;
         }
-
+        // If it is hollow account, and we require this to sign, we need to finalize the account
+        // with the corresponding ECDSA key in handle
+        if (isHollow(account)) {
+            // Verify this key isn't for an immutable account
+            verifyNotStakingAccounts(account.accountIdOrThrow(), responseCode);
```
tinker-michaelj, line 367: As above.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -390,18 +400,21 @@ public PreHandleContext requireKeyIfReceiverSigRequired(
...
```
tinker-michaelj, line 407: As above.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Yes, makes sense. Thanks @tinker-michaelj . Changed.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: Done

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleWorkflowImpl.java
```diff
@@ -234,6 +222,21 @@ private PreHandleResult expandAndVerifySignatures(
...
+        // 2. Expand the Payer signature
+        final Key payerKey;
+        if (!isHollow(payerAccount)) {
+            // If the account IS a hollow account, then we will discover all such possible signatures when expanding
+            // all "full prefix" keys above, so we already have it covered. We only need to do this if the payer is
+            // NOT a hollow account (which is the common case).
+            payerKey = payerAccount.keyOrThrow();
+            signatureExpander.expand(payerKey, originals, expanded);
+        } else {
+            payerKey = null;
```
tinker-michaelj, line 234: Since `0.0.800` payer will now make it into this block, does it make sense to add,
```
if (EMPTY_KEY_LIST.equals(payerAccount.keyOrThrow())) {
    return preHandleFailure(payer, EMPTY_KEY_LIST, INVALID_PAYER_ACCOUNT_ID, txInfo, Set.of(), Set.of(), Map.of());
}
```
? Or I guess we're safe because the `EMPTY_KEY_LIST` will never have an OK `Verifiaction`?

Neeharika-Sompalli: If someone mentioned 800 staking reward account as payer

tinker-michaelj (approved):
LGTM!

iwsimon (approved):
LGTM

Issue #9514, opened 2023-10-26
Reduce Java allocations when sending internal node hashes during reconnect


artemananiev:
This is similar to #9225, but that ticket was focused on (virtual) leaf nodes, while this one is about internal nodes. Leafs are serialized differently on disk and in sockets, but internal nodes are just hashes. Using the same approach it should be possible to serve hashes from disk directly to the socket output stream without allocating new `Hash` objects.

PR #9538, opened 2023-10-27
9514: Reduce Java allocations when sending internal node hashes during reconnect

artemananiev:
Fix summary:

* `getChildHashes()` method in `TeacherTreeView` is replaced with another method, `writeChildHashes()`. This new method is expected to write child hashes to the specified output stream in binary format, identical to how list of `Hash` objects was previously written
* In `StandardTeacherTreeView` this method is implemented using the old `getChildHashes()` and serializing them with `SerializableDataOutputStream.writeSerializableList()`
* In `VirtualTeacherTreeView` this new method delegates to the underlying data source
* In `VirtualDataSource` a new method is added for this purpose, `loadAndWriteHash()`
* This new method in `MerkleDbDataSource` loads hash bytes from disk and serializes them to the output stream, bypassing deserialization to `Hash`

Testing:

* Existing virtual map reconnect tests in `swirlds-merkle` are used
* I found that these tests use MerkleDb, but the data source is not actually used, all nodes are loaded from virtual node cache. Fixed that
* Then I had to set hashes RAM to disk threshold to 0 to load hashes from disk, not from data source cache
* Then I found that one of the reconnect tests take 5 mins to complete because of recent async timeout update. Changed the test to use custom timeout (5s)

Fixes: https://github.com/hashgraph/hedera-services/issues/9514

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCollection.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCollection.java
@@ -126,14 +126,14 @@ public class DataFileCollection<D> implements Snapshotable {
      * before compaction is complete. In the end of compaction, all the compacted files are removed
      * from this list.
      *
-     * The list is used to read data items and to make snapshots. Reading from the file, which is
+     * <p>The list is used to read data items and to make snapshots. Reading from the file, which is
      * being written to during compaction, is possible because both readers and writers use Java
      * file channel APIs. Snapshots are an issue, though. Snapshots must be as fast as possible, so
      * they are implemented as to make hard links in the target folder to all data files in
      * collection. If compaction is in progress, the last file in the list isn't fully written yet,
      * so it can't be hard linked easily. To solve it, before a snapshot is taken, the current
      * compaction file is flushed to disk, and compaction is put on hold using {@link
-     * DataFileCompactor#snapshotCompactionLock} and then resumed after snapshot is complete.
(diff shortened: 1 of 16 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCollection.java
```diff
@@ -463,22 +452,49 @@ protected D readDataItem(final long dataLocation) throws IOException {
     }
 
     /**
-     * Read a data item from any file that has finished being written. Uses a LongList that maps
-     * key-&gt;dataLocation, this allows for multiple retries going back to the index each time. The
-     * allows us to cover the cracks where threads can slip though.
+     * Read data item bytes at a given location (file index + offset). If the file is not found or
+     * alread closed, this method returns {@code null}. This may happen, if the index, where the data
```
imalygin, line 456: typo: `alread` -> `already`

imalygin (dismissed):

imalygin (approved):

OlegMazurov (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCollection.java
artemananiev: Fixed

cody-littley (approved):

qnswirlds (approved):

nathanklick (approved):

PR #9539, opened 2023-10-27
Fix issues related to auto-account creation flows

### LINKED ISSUE #8044: Start testing AutoAccountCreationSuite

rbair23:
**Description**:
This PR addresses issues with `AutoAccountCreationSuite`. The approach I took was to review all uses of the alias, and streamline those uses. There is a lot more work that can be done to streamline. Before this PR, we would perform many redundant tasks between `pureChecks`, `preHandle`, and `handle`, and within `handle` there was a lot of redundancy too.

I have not removed all of the redundancies. To really do so, we would need to make some broader changes, such as caching some computation between stages (where, for example, we can inspect the alias to see if it is a key-alias and parse it, and if it is an ECDSA(secp256k1) key convert it to an EVM address and cache that). Such work is very much worth doing, but I wanted to cut short before going that far, for the sake of time.

Some changes will therefore appear to be extraneous or unnecessary, because I only got part of the way to goal, and didn't go all the way.

**Related issue(s)**:

Fixes #8044 

**Notes for reviewer**:
We had (and still have) multiple places doing very similar things. There are many places where we do checks for whether the alias is 20 bytes long, and have different constants for doing this. I started down the path of consolidating this. I removed from `KeyUtils` the methods related to aliases, and put them into `AliasUtils`. I then moved `AliasUtils` from the implementation module to the API module for token service, so it can be reused by other modules (such as the contract service).

I'm really annoyed by `PreCheckException` and `HandleException`. One is checked, one is not. But when moving code from `handle` to `pureChecks`, it broke all kinds of things. We have validation methods for throwing `PreCheckException` and others for throwing `HandleException` and woe be upon you if you use the wrong one in preCheck or in handle! It actually leads to brittle code. We should combine these into a single exception type. This would lead to much safer code. The problem is, if you throw `HandleException` from `pureChecks`, then the catch block in `PreHandleWorkflowImpl` isn't expecting it, it expects a `PreCheckException`, so you fall into the generic exception block. The same is true for `HandleWorkflow` if you throw a `PreCheckException`. This is just bad. We need to combine these.

I added a simple `containsKey` method so to check whether an account has a key, we don't have to go to all the trouble of deserializing the object just to check that it exists. VirtualMap has a contains method we can end up using instead.

I added `pureChecks` to the `PreHandleWorkflowImpl` where we had a `Future` statement.

I renamed `deleteAndMaybeUnaliasContract` to `deleteContract`. This can be argued as a vanity change, but I find `deleteContract` to say what it does, and `deleteAndMaybeUnaliasContract` to be overly descriptive and verbose. There are many things beyond just maybe unaliasing a contract that will happen when the contract is deleted, but we don't include all of those in the name too. So I prefer the simpler name.

I made looking up a leaf in the account store much more efficient, and got rid of a lot of duplicate code between the readable and writable stores.

I collapsed and removed a lot of redundant code. I moved many checks from handle in to `pureChecks`. This is the right thing for us to do everywhere -- `pureChecks` are now called in `preHandle`, which means they will be called on background threads before `handle`, which is a much better place for us to do all the syntactic checks of the transaction. By the time we get to `preHandle` and `handle`, we can therefore assume the transactions are valid. So for example, I don't have to check in `CryptoCreateHandler` that `op.key()` is not null, because this is done in `pureChecks`. That saves time during handle. We **must** optimize these code paths to get really great TPS. I didn't streamline everything due to time constraints, but I started putting us on the path.

As a result, there are unit tests failing that need to be fixed, and some that I commented out that need to be restored. Checks that were in `handle` are now in `pureChecks`. Or checks in some validator was moved to the handler. Some of this is vanity -- I don't particularly see the point of the validation classes, they can just as easily be private methods in the handler, and we can test all their variations by crafting transactions to pass to `pureChecks` or `preHandle` or `handle` methods. But if the code owners want the validation classes, they may want to restructure some of what I did in later PRs. Personally, I prefer to keep the logic close to the handlers as I find that is easier to see and reason about. But I understand that is totally subjective.

Of course, I also added lots of comments.

**Checklist**


rbair23:
I'm looking for contributions to help fix the unit tests. It is kind of a slog due to the heavy use of mocking. If each of these tests had been fulfilled by constructing a transaction  / transaction body and calling `pureChecks`, `preHandle`, or `handle`, then all the tests would be passing. But since they're digging into the implementation details, there is a lot of (IMO unnecessary) work that now has to be done to fix them up. Help wanted!

tinker-michaelj (dismissed):
LGTM!

Neeharika-Sompalli (approved):
LGTM ! 🙌

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
```diff
@@ -95,22 +96,35 @@ public PreHandleContextImpl(
...
+    }
+
+    /** Create a new instance */
+    private PreHandleContextImpl(
+            @NonNull final ReadableStoreFactory storeFactory,
+            @NonNull final TransactionBody txn,
+            @NonNull final AccountID payer,
+            @NonNull final Configuration configuration,
+            @NonNull final TransactionDispatcher dispatcher,
+            final boolean isUserTx)
```
netopyr, line 120: We should not use a `boolean` here IMO. An enum with two describing names is easier to understand later. Or we can use `TransactionCategory`.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/PreHandleContext.java
```diff
@@ -73,6 +73,26 @@ public interface PreHandleContext extends TransactionKeys {
...
     Configuration configuration();
 
+    /**
+     * Returns true if the transaction associated with this context was created by the user, false if the transaction
+     * was created by another transaction, or by the system.
+     *
+     * @return {@code true} this transaction was created by a user, {@code false} if it was a system or child
+     *         transaction.
+     */
+    boolean isUserTransaction();
```
netopyr, line 83: Where is this needed?

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/PreHandleContext.java
```diff
@@ -73,6 +73,26 @@ public interface PreHandleContext extends TransactionKeys {
...
+
+    /**
+     * Returns the opposite of {@link #isUserTransaction()}. A transaction is either a user transaction, or a
+     * synthetic transaction, but not both.
+     *
+     * @return {@code false} this transaction was created by a user, {@code true} if it was a system or child
+     *         transaction.
+     */
+    default boolean isSyntheticTransaction() {
```
netopyr, line 92: I cannot find where this is needed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
```diff
@@ -95,112 +105,178 @@ public CryptoCreateHandler(
...
+            if (isOfEvmAddressSize(alias)) {
+                accountStore.putAlias(alias, createdAccountID);
+                recordBuilder.evmAddress(alias);
+            } else {
+                // The only other kind of alias it could be is a key-alias. And in that case, it could be an ED25519
+                // protobuf-encoded key, or it could be an ECDSA_SECP256K1 protobuf-encoded key. In this latter case,
+                // we will actually store two things in the map. We'll store the raw alias, and we'll also extract the
+                // EVM address from the key and store a second mapping from that EVM address to account ID. This makes
+                // it trivial at look up time to see if an EVM address is already in use, or to look up the account ID
+                // for an EVM address, and leads to simpler, more correct code.
```
mhess-swl, line 259: 👍

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (approved):
LGTM!

nathanklick (approved):

jeromy-cannon (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleContextImpl.java
Neeharika-Sompalli: @netopyr Makes sense. I will open a PR to fix this once this PR is merged.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/PreHandleContext.java
Neeharika-Sompalli: It is used in `isSyntheticTransaction`

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/PreHandleContext.java
Neeharika-Sompalli: It is used in `CryptoCreateHandler`

PR #9541, opened 2023-10-27
Enable leftover tests from TokenAssociateSpecs

### LINKED ISSUE #9500: Enable leftover tests from TokenAssociateSpecs
### Problem

There are some remaining tests that need to be enabled in modularization:

- dissociateHasExpectedSemantics
- dissociateHasExpectedSemanticsForDeletedTokens
- 

### Solution

Fix the issues and enable the tests

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
- Enables the remaining tests from the suite.
- Disables `dissociationFromExpiredTokensAsExpected`. It used to pass because there was no active logic for throwing `TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES`. 

**Related issue(s)**:

Fixes #9500 

**Notes for reviewer**:

**Checklist**


tinker-michaelj (changes requested):
Hmm I think we should _only_ be touching record transfer lists in a [`RecordFinalizerBase`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java#L43) subclass (agreed @Neeharika-Sompalli ?)

As for the test that dissociates an account `0.0.A` from an expired token `0.0.T`, _even when `0.0.A` has a non-zero balance of `0.0.T`...interesting question.

I'll ask about what people think about this.

Neeharika-Sompalli:
> Hmm I think we should _only_ be touching record transfer lists in a [`RecordFinalizerBase`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java#L43) subclass (agreed @Neeharika-Sompalli ?)
> 
> _[Edit - that is, we shouldn't have [this change](https://github.com/hashgraph/hedera-services/pull/9541/files#diff-52840e1a3a9777e856a6ae82e892d7500a3a5d0f3c6fe7a75f25b5044db1ec59R143) imo.]_
> 
> As for the test that dissociates an account `0.0.A` from an expired token `0.0.T`, _even when `0.0.A` has a non-zero balance of `0.0.T`...interesting question.
> 
> I'll ask about what people think about this.

@tinker-michaelj  Agreed. We should build transferLists only by looking at the changes of state.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
```diff
@@ -94,8 +94,9 @@ protected Map<EntityIDPair, Long> fungibleChangesFrom(
             final var relAcctId = modifiedRel.accountId();
             final var relTokenId = modifiedRel.tokenId();
             final var token = tokenStore.get(relTokenId);
-            // Add this to fungible token transfer list only if this token is a fungible token
-            if (!token.tokenType().equals(TokenType.FUNGIBLE_COMMON)) {
+            // Add this to fungible token transfer list only if this token is a fungible token or if the token is
+            // deleted
+            if (!token.deleted() && !token.tokenType().equals(TokenType.FUNGIBLE_COMMON)) {
```
Neeharika-Sompalli, line 99: Why do we need this check ?  `!token.deleted()`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java
```diff
@@ -144,21 +146,25 @@ public void handle(@NonNull final HandleContext context) {
...
-                        final var updatedTreasuryBalanceTokenRel = treasuryTokenRel.balance() + tokenRelBalance;
-                        treasuryBalancesToUpdate.add(treasuryTokenRel
-                                .copyBuilder()
-                                .balance(updatedTreasuryBalanceTokenRel)
-                                .build());
-                    }
+
+                    // Remove when token expiry is implemented
+                    throw new HandleException(TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES);
+                }
```
Neeharika-Sompalli, line 152: `validateFalse` throws a `HandleException`. Why do we need to throw here ?

Neeharika-Sompalli: you mean to add this `validateTrue(tokenIsExpired,TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES);`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
Ivo-Yankov: This hapi test `dissociateHasExpectedSemanticsForDeletedTokens` requires the receipt for the `uniqDissoc` dissociate transaction to contain the a `tokenTransferList` like this:

```
tokenTransferLists {
  token {
    tokenNum: 1111
  }
  transfers {
    accountID {
      accountNum: 1109
    }
    amount: -3
  }
}
```

Which is built in `fungibleChangesFrom` if the `continue` statement is skipped. I think it does make more sense to add the transfers as `nftTransfers` since the token is a `NON_FUNGIBLE_UNIQUE`, but then the behaviour would differ from the mono service.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java
Ivo-Yankov: The logic for `tokenIsExpired` is not implemented yet, and it is always `false` for now. I didn't use a `validateTrue(balance  == 0, TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES)` here because this is already in a IF statement with that condition.

A few lines below there is a comment with `validateTrue(tokenIsExpired,TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES);`. It should be uncommented when `tokenIsExpired` is implemented and the `throw` should be deleted.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
tinker-michaelj: I think it's very concerning to have record-construction logic depend on **anything** but the difference between the values returned by the `getOriginalValue()` and `get()` methods on the stores. (In this case, the `WritableTokenRelationStore`).

So even if this fixes a `@HapiTest`, I am not comfortable committing it. What do you think @Neeharika-Sompalli ?

tinker-michaelj (changes requested):
I don't have the correct fix ready offhand, but I don't think we should make `RecordFinalizerBase` depend on anything but the differences between `get()` and `getOriginalValue()`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
Neeharika-Sompalli: Yes agreed @tinker-michaelj

Ivo-Yankov: I removed the `token.deleted()` condition and disabled the HapiTest `dissociateHasExpectedSemanticsForDeletedTokens` as I am not sure how to properly fix it.

tinker-michaelj (approved):
LGTM!

PR #9557, opened 2023-10-27
enaled one more test and remove the other one that not really in use

### LINKED ISSUE #9531: RandomOps

povolev15:
Enable another test and remove one test that is not needed anymore

**Related issue(s)**:

Fixes #9531


jsync-swirlds (dismissed):

mhess-swl (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/RandomOps.java
```diff
@@ -211,6 +147,8 @@ private HapiSpec retryLimitDemo() {
                         cryptoTransfer(tinyBarsFromTo(GENESIS, FUNDING, 7L)));
     }
 
+    // Since the freeze tests all shut down the network, they can’t be part of the regular hapi test suite. It was
+    // decided to exclude them from the hapi suite.
```
jsync-swirlds, line 151: Possibly add the following here in the next PR...:
`// @todo('8687') Revisit once isolated tests are enabled`

povolev15: done thanks

jsync-swirlds (approved):

mhess-swl (approved):

Issue #9424, opened 2023-10-23
Conditional records of account auto-creation via HTS "classic transfer" variants.


tinker-michaelj:
Validate the below happy path scenarios:
 - [x] A "classic transfer" sending an NFT to an absent `receiver` address (outside long-zero range) generates a preceding child with a synthetic CryptoCreate and TransactionRecord identical that created by mono-service when it resolves to SUCCESS.
 - [ ] An ERC-20 `transferFrom()` sending fungible value to an absent `receiver` address (outside long-zero range) generates a preceding child with a synthetic CryptoCreate and TransactionRecord identical that created by mono-service when it resolves to SUCCESS.

And validate the below failure scenarios:
 - [ ] After two internal `MESSAGE_CALL`s perform hollow account creations, a "classic transfer" sending fungible value to two absent `receiver` addresses (outside long-zero range) reverts with a `MAX_CHILD_RECORDS_EXCEEDED` reason; when the top-level HAPI `ContractCall` resolves to `SUCCESS` we external two `SUCCESS` preceding children for the internal hollow creations; and one more `REVERTED_SUCCESS` for the first auto-creation done by the `CryptoTransfer` before it fails.

PR #9570, opened 2023-10-30
Fix Preceding Child Records for Auto Account Creation

georgi-l95:
**Description**:
This PR adds new method in the Handle Context class, which takes care of creating removable preceding child records. This is needed for CryptoTransfer for example towards non existing account.
Added new test in the AutoCreationSuite which covers this new case and passes in mono codebase as well.
With this change we cover this happy path from issue https://github.com/hashgraph/hedera-services/issues/9424:

- [x] A "classic transfer" sending an NFT to an absent receiver address (outside long-zero range) generates a preceding child with a synthetic CryptoCreate and TransactionRecord identical that created by mono-service when it resolves to SUCCESS.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


netopyr (changes requested):
This PR does not fit to the current design of `RecordListBuilder`. What do you want to accomplish?

georgi-l95:
> This PR does not fit to the current design of `RecordListBuilder`. What do you want to accomplish?

You can checkout more in #9424, but basically some child records needs to be preceding. I've added test which check those records and passes on mono, but not here. Therefore I've added the needed changes to adhere to the needed business logic, so that the same tests case can pass on modularization as well.

PR #9581, opened 2023-10-31
Increase contract kv/pairs storage allowed by 100x

lukelee-sl:
**Description**:


poulok (changes requested):
The defaults in the code must also be updated for the modular code. They are located in `ContractsConfig.java`

david-bakin-sl (dismissed):
LGTM.  There are a few more `163_?840` in the source - mainly in JMH though one is in `ContractAutoExpirySpecs` - don't need to be changed but it might be worth asking if they should be changed in the future?  (Esp. the JMH ones.)

poulok (dismissed):

Nana-EC (approved):

rbair23 (changes requested):
The PR appears to be missing changes to the Config records in `hedera-config` for the modular code.

rbair23 (dismissed):
Nope. you got it.

povolev15 (approved):

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -46,7 +46,7 @@ public record ContractsConfig(
         @ConfigProperty(defaultValue = "15000000") @NetworkProperty long maxGasPerSec,
         @ConfigProperty(value = "maxKvPairs.aggregate", defaultValue = "500000000") @NetworkProperty
                 long maxKvPairsAggregate,
-        @ConfigProperty(value = "maxKvPairs.individual", defaultValue = "163840") @NetworkProperty
+        @ConfigProperty(value = "maxKvPairs.individual", defaultValue = "16384000") @NetworkProperty
```
jsync-swirlds, line 49: This might interact poorly with the default on line 31, which appears to be _possibly_ incorrect... (it has `0til100M,2000til450M`, which might be intended as `0til100M,200Mtil450M` (note 200M instead of 2000, not entirely sure)

lukelee-sl: According to the comments [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/fees/charging/ContractStoragePriceTiers.java#L54), the value in line 31 looks to be correct and probably not related to configuration value that I changed.

Nana-EC (approved):
LGTM

netopyr (approved):

Issue #9589, opened 2023-11-01
Property files should handle `_` as a numeric separator


david-bakin-sl:
### Problem

A couple of us lost several hours today because we didn't realize that you couldn't use `_` as a numeric separator in property files.  A PR had added some numeric values that way and lots of HAPI tests failed.

There was a `NumberFormatException` in the logs pointing to one of those values. Took out the `_` and all the tests succeeded. 

_Strange thing is that there were **already** a few values with underscores in the property files_ - they didn't cause any tests to fail.  Don't know why.  Also don't know how that impacted the nodes in production, if it did.  But it's a trap waiting for more victims.

### Solution

Make the property file parsers accept `_` in numeric literals just like Java does.

### Alternatives

Doing nothing.  Would be a solution but not be ideal: Someone else will get bit and learn the hard way that property files don't accept `_`.

poulok:
Completed in https://github.com/hashgraph/hedera-services/pull/9682

PR #9591, opened 2023-11-01
Cherry-pick #9581 into release/0.43

lukelee-sl:
**Description**:


hedera-node/configuration/mainnet/bootstrap.properties
```diff
@@ -19,3 +19,4 @@ accounts.maxNumber=20_000_000
 #Overrides that differ based on the network
 ledger.id=0x00
 contracts.chainId=295
+contracts.maxKvPairs.individual=16384000
```
netopyr, line 22: I don't think we need to change the override files. Updating the defaults is sufficient.

netopyr (dismissed):

Nana-EC (approved):

povolev15 (approved):

hedera-node/hedera-app/src/test/resources/bootstrap.properties
```diff
@@ -79,8 +79,8 @@ contracts.keys.legacyActivations=1058134by[1062784]
 contracts.knownBlockHash=
 contracts.localCall.estRetBytes=32
 contracts.maxGasPerSec=15000000
-contracts.maxKvPairs.aggregate=500_000_000
-contracts.maxKvPairs.individual=163_840
+contracts.maxKvPairs.aggregate=500000000
```
poulok, line 82: why remove the underscores in this value? They make it easier to read and verify that the correct value is set

hedera-node/hedera-app/src/test/resources/bootstrap.properties
```diff
@@ -79,8 +79,8 @@ contracts.keys.legacyActivations=1058134by[1062784]
...
+contracts.maxKvPairs.individual=16384000
```
poulok, line 83: Please add underscores

hedera-node/hedera-app/src/test/resources/bootstrap.properties
lukelee-sl: The underscores were causing some of the HAPI tests to fail in the modularized code with `NumberFormatException`. Apparently, the logic for parsing the config values with the underscores has not been implemented in some cases (most likely the new config system)? We probably went overboard on removing them but we wanted to be sure that the tests would pass.

rbair23 (changes requested):

rbair23 (approved):

hedera-node/hedera-app/src/test/resources/bootstrap.properties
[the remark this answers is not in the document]
lukelee-sl: Restored the original value `500_000_000`

lukelee-sl: > Can you file an issue for Hendrik to add underscore support to number properties in the new config system? I agree, without underscores, we're likely to make a mistake.

Will do.

lukelee-sl: an issue was already created by @david-bakin-sl 
https://github.com/hashgraph/hedera-services/issues/9589

Nana-EC (approved):

tinker-michaelj (approved):
LGTM

povolev15 (approved):
LGTM

hedera-node/hedera-app/src/test/resources/bootstrap.properties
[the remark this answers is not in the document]
Nana-EC: Under scores restored

hedera-node/hedera-app/src/test/resources/bootstrap.properties
Nana-EC: Under scores restored

hedera-node/hedera-app/src/test/resources/bootstrap.properties
Nana-EC: Under scores restored and ticket opened

PR #9592, opened 2023-11-01
Cherry-pick #9581 into release/0.42

lukelee-sl:
**Description**:


netopyr (commented):
I wonder if it makes sense to update the code if we do not plan to deploy it. It may be confusing in the future if the file for the overridden properties contains the default values.

Nana-EC (dismissed):

lukelee-sl:
> I wonder if it makes sense to update the code if we do not plan to deploy it. It may be confusing in the future if the file for the overridden properties contains the default values.

For clarity, are you suggesting that I remove all changes to the *.java files or something else?

hedera-node/hedera-app/src/test/resources/bootstrap.properties
```diff
@@ -78,8 +78,8 @@ contracts.keys.legacyActivations=1058134by[1062784]
 contracts.knownBlockHash=
 contracts.localCall.estRetBytes=32
 contracts.maxGasPerSec=15000000
-contracts.maxKvPairs.aggregate=500_000_000
-contracts.maxKvPairs.individual=163_840
+contracts.maxKvPairs.aggregate=500000000
```
poulok, line 81: I'm not sure why the underscores were removed. They are helpful when verifying the correct number is set.

lukelee-sl: The underscores were causing some of the HAPI tests to fail in the modularized code with `NumberFormatException`.  Apparently, the logic for parsing the config values with the underscores has not been implemented?  We probably went overboard on removing them but we wanted to be sure that the tests would pass.

Nana-EC: @lukelee-sl can we open a ticket to fix that in modularization. 
FYI @hendrikebbers

Nana-EC (approved):
LG
Values are update and overrides for TN and MN envs are added

hedera-node/hedera-app/src/test/resources/bootstrap.properties
lukelee-sl: looks like an issue was already created by @david-bakin-sl 
https://github.com/hashgraph/hedera-services/issues/9589

rbair23 (approved):

tinker-michaelj (approved):
LGTM!

povolev15 (approved):

PR #9594, opened 2023-11-01
Reversible preceding transactions

netopyr:
Added support for reversible preceding transactions.
Commit state changes if non-reversible preceding transactions immediately.


Neeharika-Sompalli (approved):
These changes LGTM . Thanks @netopyr ! 🙌  I still have issues with the two conditions we expect state to be not changed

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr !

PR #9609, opened 2023-11-01
Make scheduled txn records queryable from `ScheduleCreate` payer account

tinker-michaelj:
**Description**:
As above.

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/records/TxnAwareRecordsHistorian.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/records/TxnAwareRecordsHistorian.java
@@ -325,10 +325,16 @@ public class TxnAwareRecordsHistorian implements RecordsHistorian {
 
     private void save(
             final ExpirableTxnRecord baseRecord,
-            final AccountID effPayer,
+            AccountID effPayer,
             final TransactionID txnId,
             final long submittingMember,
             final long consensusSecond) {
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/records/TxnAwareRecordsHistorian.java
```diff
@@ -325,10 +325,16 @@ private void save(
...
+            AccountID effPayer,
             final TransactionID txnId,
             final long submittingMember,
             final long consensusSecond) {
+        if (txnId.getScheduled() && txnCtx.isPayerSigKnownActive()) {
+            // Give the record of executing this scheduled transaction to the ScheduleCreate payer account,
+            // since the override payer account (if any) is not trivial to derive from just the record; and
+            // the record is all the information we will have when purging it at expiry
+            effPayer = txnId.getAccountID();
+        }
```
Neeharika-Sompalli, line 337: Just curious , what will be `txnCtx.effectivePayer()` here ?

Neeharika-Sompalli: Discussed offline. The issue was , we are storing the record under override payer account. But when purging we are using `ScheduleCreate` payer account. So the solution was to store it under `ScheduleCreate` payer account

Neeharika-Sompalli (approved):
LGTM ! Even though it would have been nice if we had used the override payer in the scheduled transaction's transactionID 😇

PR #9612, opened 2023-11-01
Make scheduled txn records queryable from `ScheduleCreate` payer account

### LINKED ISSUE #9611: Scheduled transaction records should be queryable via `ScheduleCreate` payer
### Problem

There are awkward consequences of trying to make the record of a scheduled transaction queryable from its override payer account via the HAPI `getAccountRecords()`.

### Solution

Since in practice such records are always retrieved from mirror nodes, we should simply make the scheduled transaction record queryable from the `ScheduleCreate` payer.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9611


Neeharika-Sompalli (approved):
LGTM ! Even though it would have been nice if we had used the override payer in the scheduled transaction's transactionID 😇

PR #9625, opened 2023-11-02
Fix E2E tests in TokenUpdatePrecompileSuite

### LINKED ISSUE #9566: Fix E2E tests in TokenUpdatePrecompileSuite
Fix E2E tests in `TokenUpdatePrecompileSuite`

petreze:
**Description**:
This PR fixes `TokenUpdatePrecompileSuite` e2e tests. Some changes were also needed in how we attach the child records and some additional handles inside `TokenKeyCall` in order to catch some of the corner cases

**Related issue(s)**:

Fixes #9566 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractNonRevertibleTokenViewCall.java
```diff
@@ -64,7 +65,8 @@ protected AbstractNonRevertibleTokenViewCall(
                 .systemOperations()
                 .externalizeResult(
                         contractFunctionResultSuccessFor(gasRequirement, output, contractID),
-                        SystemContractUtils.ResultStatus.IS_SUCCESS);
+                        SystemContractUtils.ResultStatus.IS_SUCCESS,
+                        SUCCESS);
```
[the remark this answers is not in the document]
petreze: Because previously it was hardcoded to be `FAIL_INVALID` in [HandleSystemContractOperations](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleSystemContractOperations.java#L122) and in some cases like [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java#L294) and [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java#L298) we expect the specific reason why the contract failed

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyCall.java
```diff
@@ -69,10 +82,51 @@ public TokenKeyCall(
...
-        if (isStaticCall && status != SUCCESS) {
+        if ((isStaticCall && status != SUCCESS) || status == INVALID_TOKEN_ID || status == KEY_NOT_PROVIDED) {
             return revertResult(status, 0);
         }
         return successResult(
                 TOKEN_KEY.getOutputs().encodeElements(status.protoOrdinal(), keyTupleFor(key)), gasRequirement);
     }
+
+    @Override
+    public @NonNull PricedResult execute() {
```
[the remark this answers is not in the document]
petreze: In some suite test cases we compare the returned status represented in string with the expected string representation like [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/CreatePrecompileSuite.java#L260) (more occurrences in the same file) and also in those two cases we would have to fail, so `contractFunctionResultFailedFor()` should be called instead of `contractFunctionResultSuccessFor()`

kimbor (approved):

agadzhalov (dismissed):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyCall.java
```diff
@@ -69,10 +81,51 @@ public TokenKeyCall(
     private @NonNull FullResult fullResultsFor(
             @NonNull final ResponseCodeEnum status, final long gasRequirement, @NonNull final Key key) {
         // @Future remove to revert #9069 after modularization is completed
-        if (isStaticCall && status != SUCCESS) {
+        if ((isStaticCall && status != SUCCESS) || status == INVALID_TOKEN_ID || status == KEY_NOT_PROVIDED) {
```
MrValioBg, line 84: Why do we need to add INVALID_TOKEN_ID here? 
For non static calls to match mono behaviour in [call-results-snapshot.json](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json) Invalid token ID should result in top-level SUCCESS status with INVALID_TOKEN_ID without revert

MrValioBg (changes requested):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyCall.java
petreze: True but the suite tests expect the call to revert and since the task here is to adapt the logic so that the suites result in success. So basically the expected result from the suite test and the one from [call-results-snapshot.json](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json) contradict each other

MrValioBg (dismissed):
LGTM! Great work 💯

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyCall.java
MrValioBg: I tested with this branch locally and removed the INVALID_TOKEN_ID from this if check and the test are still passing.

From what I saw from the suite it seems like this check that requires REVERT with INVALID_TOKEN_ID is for Update_TXN and not for tokenKey. Could you double-check, please?

petreze: Yes, they do indeed. I've removed the mentioned checks. Good catch!

dikel (approved):
LGTM

agadzhalov (approved):

MrValioBg (approved):

tinker-michaelj (approved):
Very nice work, LGTM!

PR #9628, opened 2023-11-02
feat: removal of address validity checks for top-level EVM calls

### LINKED ISSUE #9459: Removal of address validity checks for top-level and internal EVM calls
### Abstract / Motivation

### Overview

### Solution

1. Create a new EVM version.
2. Add a Feature flag.

### Tasks
```[tasklist]
### Tasks
- [ ] #8373 
- [ ] #9120 
- [ ] #4672 
```

### Resources

natanasow:
### Abstract / Motivation

As part of the effort to achieve full Ethereum equivalence, a new EVM version v046 is created with relaxed validations for the address used as:
- receiver in *top-level* and *internal* EVM calls
- address of contract to load code for 
- and balance operations

### Behavioural Changes Overview

1. Scenario: `EOA` -call→ `Contract`, where `Contract` doesn’t exist on the address that’s called
    
    In EVM versions < v045 fails with `INVALID_CONTRACT_ID`, no gas is charged and the transaction is not successful.
    In EVM versions >= v045 the transaction succeeds and performs a no-op (loads empty bytes as contract code).
    
2. Scenario: `EOA` -call→ `Caller` (contract) -call→ `Receiver`, where `Receiver` doesn’t exist on the address that's called
    
    In EVM versions < v046 the transaction is reverted with an `INVALID_SOLIDITY_ADDRESS`
    In EVM versions >= v046 the internal call succeeds and performs a no-op.
    
⚠️ Note: the calls in both scenarios are not transferring value. Value transfers to non-existing contracts/accounts activate a different flow, where a hollow account creation attempt occurs
⚠️ Note 2: the examples above specify the `call` EVM operation as an example, but other operations like all `code` operations, the `balance` operation and other variants of the `call` operation are updated as well

### Code Changes Overview

1. Create a [new dagger EVM version](https://github.com/hashgraph/hedera-services/blob/9459-removal-of-address-validity-checks-for-top-level-evm-calls/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/ContractsV_0_45Module.java) and update ContractsModule
2. Add EVM v045 logic in [HederaOperationUtil](https://github.com/hashgraph/hedera-services/blob/c93989a07e85e8f30467120f581b2e277f60f266/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaOperationUtilV038.java#L82) instead of creating new versions of the EVM Operations that would make this PR _very_ difficult to review
3. Override the AbstractLedgerEvmWorldUpdater [getOrCreate](https://github.com/hashgraph/hedera-services/blob/c93989a07e85e8f30467120f581b2e277f60f266/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/store/contracts/AbstractLedgerEvmWorldUpdater.java#L98) method that Besu [calls](https://github.com/hyperledger/besu/blob/be5cc68b6069ca2f718d08c95994e1a764576391/evm/src/main/java/org/hyperledger/besu/evm/processor/MessageCallProcessor.java#L109) internally for each receiver when it starts executing a new message frame
  a. In [AbstractLedgerWorldUpdater](https://github.com/hashgraph/hedera-services/blob/c93989a07e85e8f30467120f581b2e277f60f266/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/AbstractLedgerWorldUpdater.java#L135), add logic to create a _ghost_ account in case we know we won't be persisting the account 
5. Refactor the [ContractCallTransitionLogic](https://github.com/hashgraph/hedera-services/blob/c93989a07e85e8f30467120f581b2e277f60f266/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L250) receiver extraction and validation logic
6. Update [HederaTracer](https://github.com/hashgraph/hedera-services/blob/c93989a07e85e8f30467120f581b2e277f60f266/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java#L208) with logic to handle calls to non-existing addresses
7. Create a new [Evm45ValidationSuite](https://github.com/hashgraph/hedera-services/blob/9459-removal-of-address-validity-checks-for-top-level-evm-calls/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/validation/Evm45ValidationSuite.java)

### ToDo list

- [x] Create a new EVM version
- [x] Update the `Call` operations
- [x] Update the `Balance` operation
- [x] Update the `Code` operations
- [x] Add a Feature flag to enable/disable the removal of address validity checks
- [x] Add a Feature flag for support of grandfathering contract behavior if needed
- [x] Update legacy tests to be executed with the legacy EVM v035
- [x] Create new tests that validate the EVM v045 behavior
- [x] Fix in both the mono- and the modularized version of services - **NOTE:** _this PR will fix only the mono repo, there will be another subsequent PR for the modularized code_ 

### Tasks
```[tasklist]
### Tasks
- [ ] #8373 
- [ ] #9120 
- [ ] #4672 
```

### Resources

**Related issue(s)**:

Contributes to #9459 (the mono part, the modular part will be separate)

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/store/contracts/AbstractCodeCache.java
```diff
@@ -64,6 +64,11 @@ public Code getIfPresent(final Address address) {
             cache.put(cacheKey, code);
         }
 
+        if (code == null) {
+            code = CodeFactory.createCode(Bytes.EMPTY, 0, false);
+            cache.put(cacheKey, code);
+        }
+
```
stoqnkpL, line 71: We should make sure that actually creating a contract on the address would update the code cache with the contract's code and add such an EET.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/AccountStore.java
```diff
@@ -99,6 +99,14 @@ public Account loadContract(final Id id) {
         return loadEntityOrFailWith(id, null, INVALID_CONTRACT_ID, CONTRACT_DELETED);
     }
 
+    public boolean isContractUsable(final Id id) {
+        // check existence
+        // check expired
+        // check deleted
```
stoqnkpL, line 105: for expired and deleted contracts we should confirm what the expected behavior is. There is no such concept in Ethereum so this would be purely a Hedera decision.

stoqnkpL (commented):
Great start! I noted a couple of things we should not forget but all of this makes sense to me.

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmMessageCallProcessorV045.java
```diff
@@ -0,0 +1,158 @@
...
+            frame.setState(COMPLETED_SUCCESS);
+        } else {
+            frame.setState(EXCEPTIONAL_HALT);
+        }
+    }
+
+    protected void executeLazyCreate(final MessageFrame frame, final OperationTracer operationTracer) {
+        // no-op
+    }
+}
```
lukelee-sl, line 158: As far as I can tell, there are no differences between this file and the V38 version.  Is this true?  Also the same question for the V45 versions of the operations.  I understand why we need to create these files for the new evm version but for easier review, it would be nice to know if anything has changed in these files.

natanasow: The key difference is [here](https://github.com/hashgraph/hedera-services/pull/9628/files/2f1b95f71a58298c7a37b620176aff1d76f7e2fe#diff-f6b7d8a2463cb67f670dc2693b00f0e1223a6bef6e0271e4ba0a4ef5e7f30bb4R114). And most of the v45 files are same as v38 but using [HederaOperationUtilV045](https://github.com/hashgraph/hedera-services/pull/9628/files/2f1b95f71a58298c7a37b620176aff1d76f7e2fe#diff-c7199b2acd7fd4402b6310305c8b7f9ee1fdbc94551a2a853561b722a2cb1e82R42) where the `INVALID_SOLIDITY_ADDRESS` check had been removed. I would say that the main changes has been done here:
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-f6b7d8a2463cb67f670dc2693b00f0e1223a6bef6e0271e4ba0a4ef5e7f30bb4R114
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-e10c913db9f6cd627bf87ae17a80510dca59e8792eb0ac2b59d0ef967604b658R134
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-fdd7b93ba105216deef18b34063222d969680bfd5a50cab9b7726a34705358f5R91
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-276868df7723ffbcb87082bf04260f1ad6925821098a6e4f1541f9dc7e5b3a49R177
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-c7199b2acd7fd4402b6310305c8b7f9ee1fdbc94551a2a853561b722a2cb1e82R117
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-096e95e56d8349fd27a7a2c7ff113184b9fab4de0459a5c365cb790a866aea8cR73
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-f4a9c9a79bad422339d5033268ac44c97c68d2971863c8aedaf3e855bc6f3c76R80
- https://github.com/hashgraph/hedera-services/pull/9628/files#diff-bef7a524d6ec9a8a34a10f0965b760c92a2c4ed87dddbb4db2580a843dcdc7d0R140

stoqnkpL (commented):
A lot of *Operation classes are copied into version 45s of themselves with the sole purpose of changing the HederaOperationUtil class they're using. 
IMO it would be the lesser of two evils if we add a check for EVM v45 in HederaOperationUtilV038 and update the logic there, even though it sounds wrong to have V45 logic in a V38 class. My arguments for this are that:
1. this will make this PR much smaller and easier to review
2. this logic is already refactored in the modularized codebase so we don't need to overcomplicate our lives for the mono-service change

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
```diff
@@ -174,13 +174,19 @@ private void trackNewAction(final MessageFrame messageFrame, final Consumer<Soli
...
+                    // code can be empty when calling precompiles too, but we handle
+                    // that in tracePrecompileCall, after precompile execution is completed
+                    action.setRecipientAccount(recipient);
+                } else {
+                    action.setRecipientContract(recipient);
+                }
+            } catch (IllegalArgumentException e) {
+                // handle non-existing to/receiver address
+                action.setRecipientAccount(null);
+                action.setRecipientContract(null);
```
stoqnkpL, line 189: In what scenario does this code block throw an exception?

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmMessageCallProcessorV045.java
stoqnkpL: I suggest adding these key code changes as a list in the PR description with a small description of what each change is for, that will make reviewing this PR a lot easier

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
```diff
@@ -236,19 +247,22 @@ private void finalizeActionFor(final SolidityAction action, final MessageFrame f
...
-                                EntityId.fromAddress(asMirrorAddress(frame.getContractAddress(), frame)));
-                        syntheticInvalidAction.setTargetedAddress(
-                                Words.toAddress(frame.getStackItem(1)).toArray());
-                        syntheticInvalidAction.setError(
-                                INVALID_SOLIDITY_ADDRESS.name().getBytes(StandardCharsets.UTF_8));
-                        syntheticInvalidAction.setCallOperationType(
-                                toCallOperationType(frame.getCurrentOperation().getOpcode()));
-                        allActions.add(syntheticInvalidAction);
-                    }
+                    //                    if (action.getCallType().equals(CALL) &&
```
lukelee-sl, line 250: should we leave this commented out block?

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/AccountStore.java
```diff
@@ -99,6 +99,16 @@ public Account loadContract(final Id id) {
         return loadEntityOrFailWith(id, null, INVALID_CONTRACT_ID, CONTRACT_DELETED);
     }
 
+    public boolean isContractUsable(final Id id) {
+        try {
+            loadContract(id);
+            return true;
+        } catch (InvalidTransactionException e) {
+            validateFalse(e.getResponseCode() == CONTRACT_DELETED, CONTRACT_DELETED);
```
lukelee-sl, line 107: If this validateFalse throws an exception, is it handled?

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/operations/HederaEvmOperationsUtilV045.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param systemAccountDetector  Hedera system account detector
+     * @param systemAccountExecutionSupplier  Supplier for system account execution
+     * @return The operation result of the execution
+     */
+    static Operation.OperationResult addressCheckExecution(
+            @NonNull final MessageFrame frame,
+            @NonNull final Supplier<Bytes> supplierAddressBytes,
+            @NonNull final LongSupplier supplierHaltGasCost,
+            @NonNull final Supplier<Operation.OperationResult> supplierExecution,
+            @NonNull final BiPredicate<Address, MessageFrame> addressValidator,
```
lukelee-sl, line 53: What is the purpose of passing in `addressValidator`?  It does not seem to be used.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/crypto/EvmAutoCreationLogic.java
```diff
@@ -57,6 +60,9 @@ protected void trackAlias(final ByteString alias, final AccountID newId) {
         if (alias.size() != EVM_ADDRESS_LEN) {
             throw new UnsupportedOperationException("Stacked alias manager cannot link aliases with size != 20.");
         }
+        if (isMirror(alias.toByteArray())) {
+            throw new InvalidTransactionException(ResponseCodeEnum.INVALID_ALIAS_KEY);
```
lukelee-sl, line 64: Can you explain why the response code here is INVALID_ALIAS_KEY?

jeromy-cannon (approved):

jeromy-cannon (approved):

hendrikebbers (approved):
No platform-base files are part of that PR. Since a review is needed I just approve it ;)

jeromy-cannon (approved):

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/operations/HederaEvmOperationsUtilV038.java
```diff
@@ -53,15 +55,22 @@ static Operation.OperationResult addressCheckExecution(
...
             final var address = Words.toAddress(supplierAddressBytes.get());
             if (systemAccountDetector.test(address)) {
                 return systemAccountExecutionSupplier.get();
             }
-            if (Boolean.FALSE.equals(addressValidator.test(address, frame))) {
-                return new Operation.OperationResult(
-                        supplierHaltGasCost.getAsLong(), HederaExceptionalHaltReason.INVALID_SOLIDITY_ADDRESS);
+            // skip target entity existing check when
+            // evm >= 0.46 or FF is enabled or the target is grandfather contract
+            if (!evmProperties.evmVersion().equals(EVM_VERSION_0_46)
```
tinker-michaelj, line 67: This three-part condition seems duplicated in man places, would be good to encapsulate as a single method on `EvmProperties`.

tinker-michaelj (dismissed):
LGTM, tyvm @natanasow and @stoqnkpL !

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/operations/HederaEvmOperationsUtilV038.java
natanasow: Done.

stoqnkpL (approved):
LGTM!

jeromy-cannon (dismissed):

lukelee-sl (dismissed):
LGTM

tinker-michaelj (dismissed):
LGTM! 🙌

lukelee-sl (approved):
LGTM

jeromy-cannon (approved):

tinker-michaelj (approved):
LGTM!

PR #9630, opened 2023-11-02
Fix tests from GasLimitThrottlingSuite

MiroslavGatsanoga:
**Description**:

It looks like we haven't run the tests against mono because we have the suite commented out [here](https://github.com/hashgraph/hedera-services/blob/28afe85cc71231502f51f407039060449fb79693/hedera-node/test-clients/src/eet/java/EndToEndTests.java#L445).

The actual test definitions seemed to be broken:

1. Using single byte instead of 32 byte values e.g. Bytes.fromHexString("0x05"), I just padded them with zeros.
2. Updated the `txOverGasLimitThrottled` to instead of using gas value for the transaction which is exactly equal to the one in the gas bucket to use value that is greater than the one in the gas bucket (e.g. `MAX_GAS_PER_SECOND + 1L`).

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/8777

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/throttling/GasLimitThrottlingSuite.java
```diff
@@ -106,10 +113,12 @@ private HapiSpec txOverGasLimitThrottled() {
...
                                         "twoSSTOREs",
-                                        Bytes.fromHexString("0x05").toArray())
-                                .gas(1_000_001)
+                                        Bytes.fromHexString(
+                                                        "0x0000000000000000000000000000000000000000000000000000000000000005")
+                                                .toArray())
+                                .gas(MAX_GAS_PER_SECOND + 1L)
                                 .payingWith(PAYER_ACCOUNT)
-                                .hasPrecheck(BUSY),
+                                .hasPrecheck(MAX_GAS_LIMIT_EXCEEDED),
```
mhess-swl, line 121: Is this response code change ok for both mono and modular code? @tinker-michaelj

MiroslavGatsanoga: Actually I think I should revert it to `BUSY` and have one more transaction before it that should be succeeding and should drain the gas bucket. Not sure if that would not be flaky though. 🤔

netopyr: No, the frontend throttle check has to be run during ingest. Can the test be changed so that it does not throw `BUSY`?

MiroslavGatsanoga: As agreed with @netopyr I reverted the changes in modularized and updated the test to expect `.hasPrecheckFrom(MAX_GAS_LIMIT_EXCEEDED, BUSY)`.  cc @tinker-michaelj

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/throttling/GasLimitThrottlingSuite.java
```diff
@@ -106,10 +114,12 @@ private HapiSpec txOverGasLimitThrottled() {
...
                                         "twoSSTOREs",
-                                        Bytes.fromHexString("0x05").toArray())
-                                .gas(1_000_001)
+                                        Bytes.fromHexString(
+                                                        "0x0000000000000000000000000000000000000000000000000000000000000005")
+                                                .toArray())
+                                .gas(MAX_GAS_PER_SECOND + 1L)
                                 .payingWith(PAYER_ACCOUNT)
-                                .hasPrecheck(BUSY),
+                                .hasPrecheckFrom(MAX_GAS_LIMIT_EXCEEDED, BUSY),
```
tinker-michaelj, line 122: Won't the _only_ response be `MAX_GAS_LIMIT_EXCEEDED`?

tinker-michaelj (approved):
LGTM!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/throttling/GasLimitThrottlingSuite.java
MiroslavGatsanoga: No, in modularized we do the frontend throttles check first so we'll get `BUSY` but in mono we do the semantic check before the throttle one, so there we'll get `MAX_GAS_LIMIT_EXCEEDED`.

Unless we call `dispatcher.dispatchPureChecks()` during ingest in modularized we'll be getting `BUSY` there, but as per @netopyr's comment from above we don't want to do that.

Issue #9398, opened 2023-10-23
Fix createTokenWithInvalidFeeCollector, createTokenWithInvalidFixedFeeWithERC721Denomination tests


yosiftouma:
- Overriding CONTRACTS_MAX_NUM_WITH_HAPI_SIGS_ACCESS to “10_000_000" that contains underscores seems to throw an exception
- We get SUCCESS instead of the expected INVALID_CUSTOM_FEE_COLLECTOR / CUSTOM_FEE_DENOMINATION_MUST_BE_FUNGIBLE_COMMON status in child record

PR #9632, opened 2023-11-02
Fix handling of custom fees in token create transaction builder

MrValioBg:
**Description**:

This PR fixes the builder logic for adding customFees to the TransactionBuilder. As well as improves code readability in CreateSyntheticTxnFactory.

In the TokenCreateWrapper we initialize the `fixedFees`, `fractionalFees` & `royaltyFees` in the constructor to List.of(); So its not possible for them to be null. Also customFees method in `TokenCreateTransactionBody` directly overrides the Collection rather than adding the elements of the newly passed ones to it.

This PR addresses those issues.

**Checklist**


tinker-michaelj (changes requested):
I'm not certain I understand the point of the change; can you add a unit test that fails on `develop` and passes here to clarify?

MrValioBg:
Hey @tinker-michaelj 

I was working on [9398](https://github.com/hashgraph/hedera-services/issues/9398), the tests were failing, producing a SUCCESS status instead of the expected INVALID_CUSTOM_FEE_COLLECTOR. Upon investigating the workflow, I discovered an issue with the way custom fees are added in the transaction builder for the create token operation, which I have addressed in this PR.

The if checks in `CreateSyntheticTxnFactory` that I removed were redundantly checking if specific fees are null - these fees are initialized in the constructor of TokenCreateWrapper as List.of(), so they are never null.

Additionally, the way we were adding the fees was incorrect. The customFees(List<CustomFee>) method in TokenCreateTransactionBody simply assigns/overrides the local field for customFees in the TokenCreateTransactionBody.

Consequently, the following issue occurred:

- Since the null checks are redundant and the Fees Collections are never null, we always end up passing every if check. 
- Furthermore, because txnBodyBuilder.customFees() overrides the custom fees in TokenCreateTransactionBody, all fees were being replaced by the royalty fees, as they are the last if check to be processed. 
This issue became apparent when I noticed that the CustomFees in TokenCreateTransactionBody were always zero, which was due to the absence of royalty fees, resulting in the collection being overridden with an empty list.

I have not enabled any tests in this PR since after this fix the tests still fails, but now due to the missing child record logic.

tinker-michaelj (approved):
LGTM, thanks for the explanation @MrValioBg !

Issue #9639, opened 2023-11-02
Fix NPE while using only alias for account in `TokenWipeAccount`


Neeharika-Sompalli:
When using an alias form of account for `TokenWipe`, it fails with a NPE if the alias doesn't exist in state. 
Fix the NPE by returning `EntityNum.MISSING_NUM` when alias is not found [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/utils/accessors/SignedTxnAccessor.java#L457).

poulok:
@Neeharika-Sompalli can this ticket be closed?

Neeharika-Sompalli:
> @Neeharika-Sompalli can this ticket be closed?

Closed this after all PRs to release branches and develop are merged @poulok

PR #9640, opened 2023-11-02
Fix NPE in `TokenWipe` when using missing alias for an account

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/9639


tinker-michaelj (approved):
LGTM!

povolev15 (approved):

PR #9649, opened 2023-11-03
Fix ApproveAllowanceSuite tests

### LINKED ISSUE #9357: Fix ApproveAllowanceSuite tests

thenswan:
**Description**:
Fix ApproveAllowanceSuite tests

**Related issue(s)**:
Fixes #9357


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/grantapproval/ClassicGrantApprovalCall.java
```diff
@@ -70,7 +76,35 @@ public PricedResult execute() {
...
+            } else {
+                frame.addLog(getLogForNftAdjustAllowance(tokenAddress));
+            }
+
             return gasOnly(FullResult.successResult(encodedOutput, gasRequirement));
         }
     }
+
+    private Log getLogForFungibleAdjustAllowance(final Address logger) {
+        return EncodingFacade.LogBuilder.logBuilder()
```
tinker-michaelj, line 92: ⚠️ We don't want to leave any dependency on the `mono-service` module floating around; so we need to re-create this utility somewhere in the `contract-service-impl` module.

tinker-michaelj (changes requested):
Overall great work! 🙌 

Two items:
  1. _(Blocking)_ We need to re-create `EncodingFacade` utilities in `contract-service-impl` so we don't have lingering dependencies on mono-service.
  2. _(Nice-to-have)_ Can we add to the `HtsCall` interface a default method,
 
```
    default PricedResult execute(MessageFrame frame) {
       return execute(); 
    }
```

so this PR doesn't have to change all the files for `HtsCall` implementations that don't actually _need_ the `MessageFrame`?

thenswan:
Hey @tinker-michaelj, I've addressed your points 🙂

tinker-michaelj (approved):
Awesome, tyvm @thenswan

PR #8561, opened 2023-09-13
Cleanup Module patching

jjohannes:
**Description**:
- Remove duplications and unused entires from `com.hedera.hashgraph.jpms-modules.gradle.kts`
- Patch all Modules that end up in `data/libs` to be **real** Modules
- Remove transitive dependencies that are unused to reduce what ends up in `data/lib`
- Add some check code that makes sure everything in `data/lib` has a `module-info.class`. This can be extended later to check more precisely what we expect/allow in `data/lib`.

The patched modules might not yet be ready to really run on the _module path_. That is, they might miss certain "requires" at runtime and there could be issue with Jars containing native code. Hence this PR does not change how we start/package the application. However, in a follow up this can be addressed and then also `jlink` and `jpackage` could be used.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #9651, opened 2023-11-03
Enable Gradle Configuration Cache

jjohannes:
**Description**:

https://docs.gradle.org/current/userguide/configuration_cache.html

For this to work, all external processes that are part of the Gradle configuration need to run through `providers.exec`. Many plugins do not yet do this, hence this PR removes all plugins that call `git` under the hood. Instead we call `git` directly using `providers.exec` where needed. 

Also, the `com.google.cloud.artifactregistry.gradle-plugin` is not compatible with the _configuration cache_ yet (https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools/issues/85). But the functionality is only needed for publishing. Until the plugin is fixed, we apply it conditionally when we run without cache for publishing.

### DIFF
--- a/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.application.gradle.kts
+++ b/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.application.gradle.kts
@@ -17,12 +17,9 @@
 plugins {
     id("application")
     id("com.hedera.hashgraph.java")
-    id("com.gorylenko.gradle-git-properties")
 }
 
 group = "com.swirlds"
-
-gitProperties { keys = listOf("git.build.version", "git.commit.id", "git.commit.id.abbrev") }
 
 // Find the central SDK deployment dir by searching up the folder hierarchy
 fun sdkDir(dir: Directory): Directory =
@@ -38,9 +35,11 @@ val copyLib =
 // Copy built jar into `data/apps` and rename
 val copyApp =
     tasks.register<Copy>("copyApp") {
+        inputs.property("projectName", project.name)
+
         from(tasks.jar)
         into(sdkDir(layout.projectDirectory).dir("data/apps"))
-        rename { "${project.name}.jar" }
+        rename { "${inputs.properties["projectName"]}.jar" }
     }
 
 tasks.assemble {
--- a/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
+++ b/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
@@ -70,6 +70,33 @@ sourceSets.all {
     }
 }
 
+val writeGitProperties =
+    tasks.register<WriteProperties>("writeGitProperties") {
+        property("git.build.version", project.version)
+        @Suppress("UnstableApiUsage")
+        property(
+            "git.commit.id",
+            providers
+                .exec { commandLine("git", "rev-parse", "HEAD") }
+                .standardOutput
+                .asText
+                .map { it.trim() }
+        )
+        @Suppress("UnstableApiUsage")
+        property(
+            "git.commit.id.abbrev",
+            providers
+                .exec { commandLine("git", "rev-parse", "--short", "HEAD") }
+                .standardOutput
+                .asText
+                .map { it.trim() }
+        )
+
+        destinationFile.set(layout.buildDirectory.file("generated/git/git.properties"))
+    }
+
+tasks.processResources { from(writeGitProperties) }
+
 tasks.withType<AbstractArchiveTask>().configureEach {
     isPreserveFileTimestamps = false
     isReproducibleFileOrder = true
--- a/hedera-node/hapi/build.gradle.kts
+++ b/hedera-node/hapi/build.gradle.kts
@@ -26,19 +26,26 @@ description = "Hedera API"
 val hapiProtoBranchOrTag = "add-pbj-types-for-state"
 val hederaProtoDir = layout.projectDirectory.dir("hedera-protobufs")
 
-@Suppress("UnstableApiUsage")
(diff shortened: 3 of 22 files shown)


hedera-node/hapi/build.gradle.kts
```diff
@@ -23,6 +23,39 @@ plugins {
...
+
+@Suppress("UnstableApiUsage")
+providers
+    .exec {
+        workingDir = hederaProtoDir.asFile
+        commandLine("git", "checkout", hapiProtoBranchOrTag, "-q")
+    }
+    .result
+    .get()
+
```
jjohannes, line 58: This will most likely be replaced with some other solution for how to integrate `hedera-protobufs` soon. Doing this, instead of using `me.champeau.includegit`, makes the current approach fully configuration cache compatible (and we did not use any other features of that plugin).

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.application.gradle.kts
```diff
@@ -38,9 +35,11 @@ val copyLib =
 // Copy built jar into `data/apps` and rename
 val copyApp =
     tasks.register<Copy>("copyApp") {
+        inputs.property("projectName", project.name)
+
         from(tasks.jar)
         into(sdkDir(layout.projectDirectory).dir("data/apps"))
-        rename { "${project.name}.jar" }
+        rename { "${inputs.properties["projectName"]}.jar" }
```
jjohannes, line 42: This runs during Gradle's execution phase. With config cache enabled, it is no longer allowed to access the "configuration time state" during execution directly. Everything a task needs, has to be declared as `input`. Hence the changes to our custom task setups.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
```diff
@@ -70,6 +70,33 @@ sourceSets.all {
...
+                .standardOutput
+                .asText
+                .map { it.trim() }
+        )
+
+        destinationFile.set(layout.buildDirectory.file("generated/git/git.properties"))
+    }
+
+tasks.processResources { from(writeGitProperties) }
+
```
jjohannes, line 99: Moved this up from `com.hedera.hashgraph.sdk.conventions.gradle.kts` where the `com.gorylenko.gradle-git-properties` plugin was doing this before.
I am not sure if this is still needed/desried at all, but if we have it, I think **all** Modules in the project should have these properties added to the Jar they produce consistently.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/utils/ActionsHelperTest.java
```diff
@@ -53,7 +53,7 @@ class ActionsHelperTest {
     @Test
     void prettyPrintsAsExpected() {
         final var expected =
-                "SolidityAction(callType: CALL, callOperationType: OP_CALL, value: 0, gas: 500000, gasUsed: 0, callDepth: 0, callingAccount: <null>, callingContract: ContractID[shardNum=0, realmNum=0, contract=OneOf[kind=CONTRACT_NUM, value=666]], recipientAccount: <null>, recipientContract: ContractID[shardNum=0, realmNum=0, contract=OneOf[kind=CONTRACT_NUM, value=666]], invalidSolidityAddress (aka targetedAddress): <null>, input: Bytes[1,2,3,4,5,6,7,8,9], output: Bytes[9,8,7,6,5,4,3,2,1], revertReason: <null>, error: <null>)";
+                "SolidityAction(callType: CALL, callOperationType: OP_CALL, value: 0, gas: 500000, gasUsed: 0, callDepth: 0, callingAccount: <null>, callingContract: ContractID[shardNum=0, realmNum=0, contract=OneOf[kind=CONTRACT_NUM, value=666]], recipientAccount: <null>, recipientContract: ContractID[shardNum=0, realmNum=0, contract=OneOf[kind=CONTRACT_NUM, value=666]], invalidSolidityAddress (aka targetedAddress): <null>, input: 010203040506070809, output: 090807060504030201, revertReason: <null>, error: <null>)";
```
jjohannes, line 56: Change on behavior in PBJ: `0.7.4` -> `0.7.6` (see https://github.com/hashgraph/pbj/pull/99)
Checked by @imalygin

hedera-node/hedera-app/build.gradle.kts
```diff
@@ -146,25 +145,6 @@ val copyLib =
...
-                            null
-                        }
-                    }
-                    .sorted()
-            if (nonModulalJars.isNotEmpty()) {
-                throw RuntimeException(
-                    "Jars without 'module-info.class' in 'data/lib'\n${nonModulalJars.joinToString("\n")}"
-                )
-            }
-        }
```
jjohannes, line 167: The way I did this (in #8561), is not compatible with the configuration cache. I will re-add a similar check in a follow up. (By now the `extra-java-module-info` has some new functionality –  https://github.com/gradlex-org/extra-java-module-info/pull/75 – that may help us.)

nathanklick (dismissed):

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/utils/ActionsHelperTest.java
jjohannes: @quiet-node can you approve this PR for the above change? Thanks.

nathanklick (approved):

lukelee-sl (approved):
LGTM

PR #9661, opened 2023-11-03
Stabilize fuzzy matching

### LINKED ISSUE #9660: Stabilize `SnapshotModeOp` fuzzy-matching for additional non-deterministic fields
### Problem

Not all fields currently fuzzy-matched are deterministic between test runs.

### Solution

Update `SnapshotModeOp` to address this.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9660 
 - Allows `transactionFee` fields to vary by a tinybar by default.
 - Skips `expirationTime` and `parent_consensus_timestamp` by default.
 - Permits skipping `contractCallResult` and `functionParameters` when these vary by addresses.
 - Tries to read from all 4 _recordStreams/_ in a HAPI Test network output, in case at least node successfully flushes its streams.
 - Adds mono-service snapshots for two `Create2OperationSuite`'s; these cannot be matched against `@HapiTest` streams yet because we are missing `transactionFee` fields in a `ContractCreate` records. I.e. the current failure is,
```
        2023-11-03 12:38:16.560 ERROR  539  HapiSpec - Record snapshot fuzzy-match failed
       ... Mismatched field counts  (expected but not find [transactionFee]) between expected receipt {
```

To fix this for now, we probably need a `HandleContext` method a `TransactionHandler` can call to report a non-standard fee charge. 🤔 See https://github.com/hashgraph/hedera-services/issues/9663.


mhess-swl (approved):

iwsimon (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM !

PR #9672, opened 2023-11-03
Add comments to failing record regression tests

mhess-swl:
The three remaining tests in `TxnRecordRegression` fail in both the modular and the mono code, even with a 3 second TTL. This PR just adds comments to revisit them later. 

Closes #9622

Issue #8813, opened 2023-09-21
Implement payer account exemption logic


MiroslavGatsanoga:
We have `htsOwnershipCheckWorksWithAliasAddress` test from `ContractCallLocalSuite` failing with `INSUFFICIENT_TX_FEE` due to failure in a token mint transaction.
The reason for that seems to be due to transaction fee being set to 0 in `test-clients`. In mono we have[ exemption logic in place for the payer account](https://github.com/hashgraph/hedera-services/blob/f832b996cef64bf557305fd0f07708590c1520c5/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/submission/SolvencyPrecheck.java#L104) for this case to pass, we need the same implemented in modularized.

netopyr:
The payer exempt logic was recently added. The test still fails, but with a different error message: 
```
2023-10-05 10:32:38.798 WARN   262  HapiSpecOperation - 'htsOwnershipCheckWorksWithAliasAddress' - CustomSpecAssert failed - Sender address should match the owner address. ==> expected: <0x7c1E698315a7cAF0c3EFdEF02B7eaE2f2D597c78> but was: <0x0000000000000000000000000000000000000000>!
```

Can you please take another look?

MiroslavGatsanoga:
> The payer exempt logic was recently added. The test still fails, but with a different error message:
> 
> ```
> 2023-10-05 10:32:38.798 WARN   262  HapiSpecOperation - 'htsOwnershipCheckWorksWithAliasAddress' - CustomSpecAssert failed - Sender address should match the owner address. ==> expected: <0x7c1E698315a7cAF0c3EFdEF02B7eaE2f2D597c78> but was: <0x0000000000000000000000000000000000000000>!
> ```
> 
> Can you please take another look?

Alright, I will take a look, thanks.

MiroslavGatsanoga:
It looks like now the test fails [here](https://github.com/hashgraph/hedera-services/blob/fb2c82b5acf55c25ac5904589e3aded6db34e284/hedera-node/test-clients/src/main/resource/contract/contracts/OwnershipCheck/OwnershipCheck.sol#L11), after calling the `getNonFungibleTokenInfo‎` precompile we expect `tokenInfo.ownerId` and `msg.sender` to be equal in the test.
Additionally tried running it against mono and there we also get `responseCode` `SUCCESS(22)` but in modularized we get `UNKNOWN(21)`.

[...]

PR #9681, opened 2023-11-06
08813: ContractCallLocalSuiteTests (#9672)

ilko-iliev-lime:
Description:

Purpose of the PR is fixing the htsOwnershipCheckWorksWithAliasAddress in the ContractCallLocalSuite. It fixes some problems in the NftTokenInfoCall and covers some unhandled cases.

Related issue(s):
https://github.com/hashgraph/hedera-services/issues/8813

Fixes #9617 

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
@@ -33,6 +33,7 @@ import com.hedera.hapi.node.transaction.FixedFee;
 import com.hedera.hapi.node.transaction.FractionalFee;
 import com.hedera.hapi.node.transaction.RoyaltyFee;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.hapi.node.state.token.Account;
 import com.hederahashgraph.api.proto.java.TokenSupplyType;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.math.BigInteger;
@@ -60,7 +61,7 @@ public class TokenTupleUtils {
         return Tuple.of(
                 token.expirationSecond(),
                 headlongAddressOf(token.autoRenewAccountIdOrElse(ZERO_ACCOUNT_ID)),
-                token.autoRenewSeconds());
+                (token.autoRenewSeconds() > 0 ? token.autoRenewSeconds() : 0));
     }
 
     /**
@@ -230,14 +231,15 @@ public class TokenTupleUtils {
             @NonNull final Token token,
             @NonNull final Nft nft,
             final long serialNumber,
-            @NonNull final String ledgerId) {
+            @NonNull final String ledgerId,
+            Account ownerAccount) {
 
         final var nftMetaData = nft.metadata() != null ? nft.metadata().toByteArray() : Bytes.EMPTY.toByteArray();
 
         return Tuple.of(
                 tokenInfoTupleFor(token, ledgerId),
                 serialNumber,
-                headlongAddressOf(nft.ownerIdOrElse(ZERO_ACCOUNT_ID)),
+                headlongAddressOf(ownerAccount),
                 nft.mintTimeOrElse(new Timestamp(0, 0)).seconds(),
                 nftMetaData,
                 headlongAddressOf(nft.spenderIdOrElse(ZERO_ACCOUNT_ID)));
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.nfttokeninfo;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.SUCCESS;
 import static com.hedera.node.app.service.contract.impl.exec.systemcontracts.HederaSystemContract.FullResult.revertResult;
 import static com.hedera.node.app.service.contract.impl.exec.systemcontracts.HederaSystemContract.FullResult.successResult;
@@ -24,7 +25,9 @@ import static com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts
 import static com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.nfttokeninfo.NftTokenInfoTranslator.NON_FUNGIBLE_TOKEN_INFO;
 import static java.util.Objects.requireNonNull;
 
+import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.ResponseCodeEnum;
+import com.hedera.hapi.node.state.token.Account;
 import com.hedera.hapi.node.state.token.Nft;
 import com.hedera.hapi.node.state.token.Token;
 import com.hedera.node.app.service.contract.impl.exec.gas.SystemContractGasCalculator;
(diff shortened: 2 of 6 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
```diff
@@ -61,40 +65,54 @@ public NftTokenInfoCall(
...
                 gasRequirement);
     }
+
+    private Account getOwnerAccount(Nft nft, Token token) {
+        final var explicitId = nft.ownerIdOrElse(AccountID.DEFAULT);
+        if (explicitId.account().kind() == AccountID.AccountOneOfType.UNSET) {
+            return null;
+        }
+        final long ownerNum;
+        if (explicitId.accountNumOrElse(TREASURY_OWNER_NUM) == TREASURY_OWNER_NUM) {
```
MrValioBg, line 111: Maybe we can directly use if(!explicitId.hasAccountNum()) and then we wouldn't need the constant TREASURY_OWNER_NUM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
```diff
@@ -61,40 +65,54 @@ public NftTokenInfoCall(
...
+                .nativeOperations()
+                .getNft(token.tokenIdOrElse(ZERO_TOKEN_ID).tokenNum(), serialNumber);
         // @Future remove to revert #9074 after modularization is completed
-        if (isStaticCall && (status != SUCCESS || nft == null)) {
+        if ((isStaticCall && (status != SUCCESS)) || nft == null) {
             return revertResult(status, gasCalculator.viewGasRequirement());
         }
 
-        final var nonNullNft = nft != null ? nft : Nft.DEFAULT;
+        Account ownerAccount = getOwnerAccount(nft, token);
```
MrValioBg, line 92: Just a small thing, we can use "var" here in order to be consistent.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
```diff
@@ -61,40 +65,54 @@ public NftTokenInfoCall(
...
         requireNonNull(token);
 
         final var ledgerConfig = configuration.getConfigData(LedgerConfig.class);
         final var ledgerId = Bytes.wrap(ledgerConfig.id().toByteArray()).toString();
+        final var nft = enhancement
+                .nativeOperations()
+                .getNft(token.tokenIdOrElse(ZERO_TOKEN_ID).tokenNum(), serialNumber);
         // @Future remove to revert #9074 after modularization is completed
-        if (isStaticCall && (status != SUCCESS || nft == null)) {
+        if ((isStaticCall && (status != SUCCESS)) || nft == null) {
```
MrValioBg, line 88: I think this would result in CONTRACT_REVERT for non-static call in the case of not valid serial number of NFT, thus it would not match the mono response from [call-results-snapshot.json](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -60,7 +61,7 @@ public static Tuple expiryTupleFor(@NonNull final Token token) {
         return Tuple.of(
                 token.expirationSecond(),
                 headlongAddressOf(token.autoRenewAccountIdOrElse(ZERO_ACCOUNT_ID)),
-                token.autoRenewSeconds());
+                (token.autoRenewSeconds() > 0 ? token.autoRenewSeconds() : 0));
```
tinker-michaelj, line 64: This expression would be different from just `token.autoRenewSeconds()` **only** if `token.autoRenewSeconds() < 0`.  

Is that possible? 🤔

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
```diff
@@ -173,6 +173,10 @@ public static long asExactLongValueOrZero(@NonNull final BigInteger value) {
      */
     public static com.esaulpaugh.headlong.abi.Address headlongAddressOf(@NonNull final AccountID accountID) {
         requireNonNull(accountID);
+        if (accountID.account().kind() == AccountID.AccountOneOfType.UNSET) {
```
tinker-michaelj, line 176: Why would this ever be possible? 🤔

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
```diff
@@ -90,11 +93,27 @@ public NftTokenInfoCall(
...
                 NON_FUNGIBLE_TOKEN_INFO
                         .getOutputs()
                         .encodeElements(
-                                status.protoOrdinal(), nftTokenInfoTupleFor(token, nonNullNft, serialNumber, ledgerId)),
+                                status.protoOrdinal(),
+                                nftTokenInfoTupleFor(token, nonNullNft, serialNumber, ledgerId, ownerAccount)),
                 gasRequirement);
     }
+
+    private Account getOwnerAccount(Nft nft, Token token) {
```
tinker-michaelj, line 109: Since we don't have unit tests for this new code, I think we should make the relevant logic in `OwnerOfCall` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/ownerof/OwnerOfCall.java#L57) a `static` helper and re-use that code here instead of duplicating it.

tinker-michaelj (changes requested):
Overall a very nice fix 💯 , but I had just enough questions and nits to request changes. 😇 

Oh, and there is a `@HapiTest` failure [here](https://scans.gradle.com/s/pix2apuvesulo/tests/task/:test-clients:hapiTest/details/com.hedera.services.bdd.suites.contract.precompile.TokenInfoHTSSuite/happyPathGetNonFungibleTokenInfo?top-execution=1) - `happyPathGetNonFungibleTokenInfo()`, we should investigate that.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
ilko-iliev-lime: Yes, unfortunately. The test was failing because autoRenewSeconds is an unsigned integer and it was a negative vaue.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -230,14 +231,15 @@ public static Tuple nftTokenInfoTupleFor(
             @NonNull final Token token,
             @NonNull final Nft nft,
             final long serialNumber,
-            @NonNull final String ledgerId) {
+            @NonNull final String ledgerId,
+            Account ownerAccount) {
```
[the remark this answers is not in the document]
ilko-iliev-lime: Ty, included it :)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
ilko-iliev-lime: When the spenderId of the token is null, we then use the ZERO_ACCOUNT_ID, which has an alias == null and that's where we failed with NPE on line 182. I could move the check one level above, but am not sure there are other tests failing because of that.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCall.java
ilko-iliev-lime: Updated it, thank you :)

PR #9704, opened 2023-11-06
move atomic classes from metrics to concurrent classes

hendrikebbers:
Fix for https://github.com/hashgraph/hedera-services/issues/9705

This PR just moves 2 classes in a different package.


platform-sdk/swirlds-common/src/main/java/module-info.java
```diff
@@ -160,6 +160,7 @@
             com.swirlds.platform.test,
             com.swirlds.platform.gui;
     exports com.swirlds.common.startup;
+    exports com.swirlds.common.threading.atomic;
```
jsync-swirlds, line 163: The rename also requires we remove the old name on line 47:
```
/runner/_work/hedera-services/hedera-services/platform-sdk/swirlds-common/src/main/java/module-info.java:47: error: package is empty or does not exist: com.swirlds.common.metrics.atomic
    exports com.swirlds.common.metrics.atomic
```

cody-littley (dismissed):

jeromy-cannon (dismissed):

platform-sdk/swirlds-common/src/main/java/module-info.java
hendrikebbers: THX, no idea why my local build ignored that. maybe gradle cache

jsync-swirlds (approved):

cody-littley (approved):

imalygin (approved):

PR #9732, opened 2023-11-07
Fix ContractHTSSuite tests

### LINKED ISSUE #9656: Fix ContractHTSSuite tests

thenswan:
**Description**:
Fix ContractHTSSuite tests.

**Related issue(s)**:
Fixes #9656


tinker-michaelj (changes requested):
The `TokenService` and `ScheduleService` both dispatch txns so the `HandleContext` should not always set a `ContractCallResult`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -607,6 +609,10 @@ private void dispatchSyntheticTxn(
             dispatcher.dispatchPureChecks(txBody);
         } catch (PreCheckException e) {
             childRecordBuilder.status(e.responseCode());
+            var output = ReturnTypes.encodedRc(e.responseCode());
```
lukelee-sl, line 612: It seems to me like we are making an assumption here that all PreCheckExceptions result from a contract call. I'm not sure if this is true and perhaps this can lead to unexpected outcomes.

thenswan:
@lukelee-sl @tinker-michaelj updated the PR with another option.

tinker-michaelj (approved):
Ah very nice, LGTM!

lukelee-sl (approved):
LGTM

JeffreyDallas (approved):

jeromy-cannon (approved):

PR #9735, opened 2023-11-07
Add test to confirm preceding child record for HTS transfer

georgi-l95:
**Description**:
Added new test in the AutoCreationSuite which covers this new case and passes in mono codebase as well.
With this change we cover this happy path from issue https://github.com/hashgraph/hedera-services/issues/9424:

- [x] A "classic transfer" sending an NFT to an absent receiver address (outside long-zero range) generates a preceding child with a synthetic CryptoCreate and TransactionRecord identical that created by mono-service when it resolves to SUCCESS.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


tinker-michaelj (dismissed):
LGTM! 🙌

tinker-michaelj (approved):
LGTM!

PR #9739, opened 2023-11-07
Auto-snapshot management

### LINKED ISSUE #9697: Automate `HapiSpec` snapshot matching
### Problem

We don't want to have to manually add `SnapshotModeOp` operations to every spec for local record validation.

### Solution

Introduce a `HapiSpec` property,
```
recordStream.autoSnapshotManagement=false
recordStream.autoSnapshotTarget=MONO_SERVICE
recordStream.autoMatchTarget=HAPI_TEST
```
to automatically insert:
  1. A `snapshotMode(FUZZY_MATCH_AGAINST_...)` when a snapshot already exists under _record-snapshots/_ (based on value of `recordStream.autoSnapshotTarget`).
  2. A `snapshotMode(TAKE_FROM_...)` when no snapshot exists under  _record-snapshots/_ (based on value of `recordStream.autoMatchTarget`).

This will allow us to run locally against mono-service with `recordStream.autoSnapshotManagement=true` to create snapshots; then validate their stability with `recordStream.autoMatchTarget=MONO_SERVICE`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9697 
 - Consolidates per-`HapiSpec` JSON snapshots into a single per-`HapiSuite` snapshot.
 - Adds an `AutoSnapshotModeOp` that delegates to a `SnapshotModeOp` with a mode determined by:
    1. The presence of a JSON snapshot in _record-snapshot/_ (takes a snapshot if it is missing, and fuzzy-matches if it is present).
    2. The values of the properties [here](https://github.com/hashgraph/hedera-services/blob/09697-auto-snapshot-matching/hedera-node/test-clients/src/main/resource/spec-default.properties#L40), which control which type of network's record streams we would use.
- Updates `HapiSpec` to insert an `AutoSnapshotModeOp` at the start of the spec if `recordStream.autoSnapshotManagement=true`.
- Validates behavior of taking and fuzzy-matching against mono-service streams for `CryptoTransferSuite`.
- Makes `HapiCryptoTransfer` transfer list construction deterministic by sorting.


hedera-node/test-clients/record-snapshots/CryptoTransfer-SpecialAccountsBalanceCheck.json
```diff
@@ -0,0 +1 @@
+{"placeholderNum":1008,"encodedItems":[]}
```
iwsimon, line 1: No items here. Do we still need this to be generated or this is an example to handle empty item?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
```diff
@@ -585,12 +586,12 @@ private HapiSpec payerBalanceIsReflectsAllChangesBeforeFeeCharging() {
...
-        // The expected fee for two token transfers to a receiver with no auto-creation;
-        // note it is approximate because the fee will vary slightly with the size of
-        // the sig map, depending on the lengths of the public key prefixes required
+        // The expected (network + service) fee for two token transfers to a receiver
+        // with no auto-creation; note it is approximate because the fee will vary slightly
+        // with the size of the sig map, depending on the lengths of the public key prefixes required
         final long approxTransferFee = 1163019L;
 
-        return defaultHapiSpec("canAutoCreateWithFungibleTokenTransfersToAlias")
+        return onlyDefaultHapiSpec("canAutoCreateWithFungibleTokenTransfersToAlias", NONDETERMINISTIC_TRANSACTION_FEES)
```
iwsimon, line 594: defaultHapiSpec("canAutoCreateWithFungibleTokenTransfersToAlias", NONDETERMINISTIC_TRANSACTION_FEES)?

Neeharika-Sompalli, line 594: Need to change back to defaultHapiSpec

hedera-node/test-clients/record-snapshots/CryptoTransfer-SpecialAccountsBalanceCheck.json
tinker-michaelj: Hi @iwsimon , this spec [doesn't actually submit any transactions](https://github.com/hashgraph/hedera-services/blob/09697-auto-snapshot-matching/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoTransferSuite.java#L1816), haha.

Neeharika-Sompalli (dismissed):
LGTM ! Thanks @tinker-michaelj  🙌

iwsimon (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
tinker-michaelj: Fixed.

tinker-michaelj: Fixed!

Neeharika-Sompalli (approved):
Looks Great 🎉 . Thanks @tinker-michaelj

PR #9747, opened 2023-11-07
09448 Implement intake with new framework

### LINKED ISSUE #9448: Implement event intake via new task framework
```[tasklist]
### Prerequisites
- [ ] https://github.com/hashgraph/hedera-services/issues/9412
- [ ] https://github.com/hashgraph/hedera-services/issues/9285
- [ ] https://github.com/hashgraph/hedera-services/issues/8670
- [ ] https://github.com/hashgraph/hedera-services/issues/9196
- [ ] https://github.com/hashgraph/hedera-services/issues/9449
- [ ] https://github.com/hashgraph/hedera-services/issues/9476
- [ ] https://github.com/hashgraph/hedera-services/issues/9525
- [ ] https://github.com/hashgraph/hedera-services/issues/9574
- [ ] https://github.com/hashgraph/hedera-services/issues/9666
- [ ] https://github.com/hashgraph/hedera-services/issues/9675
- [ ] https://github.com/hashgraph/hedera-services/issues/9674
```

New intake flow:

`gossip --> internal event validation --> hashing --> deduplication --> signature validation --> orphan buffer --> linker —> event intake`

alittley:
- closes #9448


cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
```diff
@@ -97,6 +104,20 @@ private void wire() {
         orphanBufferScheduler.getEventOutput().solderTo(inOrderLinkerScheduler.getEventInput());
         inOrderLinkerScheduler.getEventOutput().solderTo(linkedEventIntakeScheduler.getEventInput());
```
edward-swirldslabs, line 105: This pattern makes it look like schedulers are all adhering to a common interface, but every scheduler defines their own get input wire and get output wire method.  The conformance to a pattern is happenstance.   Should there be a common interface or common base class that explains and enforces the similarity?  (Feel free to say no)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/LinkedEventIntakeScheduler.java
```diff
@@ -17,25 +17,30 @@
...
+import java.util.List;
 
 /**
  * Wiring for the {@link LinkedEventIntakeScheduler}.
  */
 public class LinkedEventIntakeScheduler {
-    private final InputWire<EventImpl, Void> eventInput;
+    private final InputWire<EventImpl, List<ConsensusRound>> eventInput;
+
+    private final OutputWire<ConsensusRound> eventOutput;
```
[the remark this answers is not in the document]
edward-swirldslabs: I'm feeling like the second parameter of the input wire is useless noise if it doesn't have a tight correspondence with an output wire.   For example in 

```
public class EventDeduplicatorScheduler {
    private final TaskScheduler<GossipEvent> taskScheduler;
    private final InputWire<GossipEvent, GossipEvent> eventInput;
    private final InputWire<Long, GossipEvent> minimumGenerationNonAncientInput;
```

The `GossipEvent` parameter of the minGenNonAncient input wire is completely nonsense.   It makes it look like a long goes in and a single GossipEvent is supposed to come out.  

If the schedulers/components are not strongly typed with output wire types, then it is just noise on the input wire's type because it is not forced to correspond to anything, it only corresponds by convention and we're breaking the convention in many places.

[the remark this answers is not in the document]
edward-swirldslabs: Much of my commentary here is on the framework itself now that I see it being applied to a real world context and not toy examples.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
```diff
@@ -105,17 +116,26 @@ public boolean isIllegalDirectSchedulerUsagePresent() {
...
+        inOrderLinkerWiring.eventOutput().solderTo(linkedEventIntakeWiring.eventInput());
+
+        final OutputWire<Long> minimumGenerationNonAncientOutput =
+                linkedEventIntakeWiring.minimumGenerationNonAncientOutput();
+
+        minimumGenerationNonAncientOutput.solderTo(eventDeduplicatorWiring.minimumGenerationNonAncientInput(), INJECT);
+        minimumGenerationNonAncientOutput.solderTo(
+                eventSignatureValidatorWiring.minimumGenerationNonAncientInput(), INJECT);
+        minimumGenerationNonAncientOutput.solderTo(orphanBufferWiring.minimumGenerationNonAncientInput(), INJECT);
+        minimumGenerationNonAncientOutput.solderTo(inOrderLinkerWiring.minimumGenerationNonAncientInput(), INJECT);
```
cody-littley, line 132: Optional suggestion.

As we add more and more things that need to be soldered, this function is going to get large. I wonder if it would make sense to split certain related wiring things into helper functions. For example, we could create a function `solderMinimumGenerationNonAncient()` that encapsulates these lines.

cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/LinkedEventIntakeScheduler.java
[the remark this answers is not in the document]
cody-littley: Note that I intend to make the simplification we discussed with wires soon (i.e. where we hide the second generic argument of an input wire).

cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
alittley: Let me know if the refactor mitigates your concerns - there wasn't any interface added for this, but things have been reorganized, and naming improved.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/EventDeduplicatorWiring.java
```diff
@@ -30,8 +32,8 @@
  * @param flushRunnable                    the runnable to flush the deduplicator
  */
 public record EventDeduplicatorWiring(
-        @NonNull InputWire<GossipEvent, GossipEvent> eventInput,
-        @NonNull InputWire<Long, GossipEvent> minimumGenerationNonAncientInput,
+        @NonNull BindableInputWire<GossipEvent, GossipEvent> eventInput,
```
cody-littley, line 35: Hmm... this is perhaps an unfortunate side effect of using records. The goal of creating an `InputWire<IN>` was so that this interface didn't need to expose the output type. The output type of each input wire is really an implementation detail of the binding.

Two possible ways around this.

1) turn these back into classes, and make the getter methods return `InputWire` instead of `BindableInputWire`.
2) store these as `InputWire<>` types, and cast to the bindable type at binding time. Ugly, but it makes the API cleaner.

Perhaps we can discuss this with Lazar tomorrow.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Instantiate the schedulers for the platform, for the given wiring model
+     *
+     * @param model the wiring model
+     * @return the instantiated platform schedulers
+     */
+    public static PlatformSchedulers create(@NonNull final WiringModel model) {
+        return new PlatformSchedulers(
+                model.schedulerBuilder("internalEventValidator")
+                        .withType(SEQUENTIAL)
```
lpetrovic05, line 56: I think we decided not to use SEQUENTIAL yet. I think that a bunch of these are supposed to run on the same thread, so that the execution profile is the same as current develop

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
```diff
@@ -0,0 +1,98 @@
...
+                        .withUnhandledTaskCapacity(500)
```
lpetrovic05, line 57: I think the queue capacity is set by the config

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
alittley: We should discuss this as a group - I don't recall deciding on that.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
cody-littley: Regardless of how we set these, I think 500 is a reasonable value. We shouldn't use 10k like we currently have configured for the intake queue, since if we have X queues in the new pipeline, the total capacity would be X*10k.

edward-swirldslabs (dismissed):

cody-littley (approved):

lpetrovic05 (dismissed):

cody-littley (approved):

lpetrovic05 (approved):

PR #9762, opened 2023-11-07
Use a lock on StreamFileProducerConcurrent properties to prevent race

### LINKED ISSUE #9755: Race in StreamFileProducerConcurrent
### Description

The documentation for the `StreamFileProducerConcurrent` class says that the public methods are only ever called in the context of the handle thread, but that's not exactly true. In particular, `close` can be called outside the context of the handle thread. `switchBlocks` and `writeRecordStreamItems` both set `currentRecordFileWriter` which is fine on their own since they are only ever called from the handle thread. However, `currentRecordFileWriter` is accessed in `close` so it's possible for it to be:

1. Read in `close()` in thread A.
2. Modified in `switchBlocks` or `writeRecordStreamItems` in thread B.

This could result in data not being written to the steam.

### Steps to reproduce

The `close()` [method](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java#L228) is triggered in Hedera.java [outside the context of the handle thread](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java#L644) and this class is not thread safe.

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

v0.44

### Operating system

None

nickpoorman:
**Description**:

The documentation for the `StreamFileProducerConcurrent` class says that the public methods are only ever called in the context of the handle thread, but that's not exactly true. In particular, `close` can be called outside the context of the handle thread. `switchBlocks` and `writeRecordStreamItems` both set `currentRecordFileWriter` which is fine on their own since they are only ever called from the handle thread. However, `currentRecordFileWriter` is accessed in `close` so it's possible for it to be concurrently:

1. Read in `close()` in thread A.
2. Modified in `switchBlocks` or `writeRecordStreamItems` in thread B.

Given this is called on the handle thread we may not want the ReentrantLock.

Instead of wrapping these methods with a ReentrantLock, which would require each call to acquire a lock and release it, `close()` instead signals that another BlockRecordWriter should not be opened in `switchBlocks` after close() has been called. `close()` then blocks until `switchBlocks` has closed the final block.

However, there is a test showing that the desired behavior is to keep creating writers in the event that there is an exception. Because of this, I have reverted back to using a lock. The lock-free implementation is in: https://github.com/hashgraph/hedera-services/pull/9762/commits/76b4065f67fe4ec00047e49622a06769a7b0e00b 

**Related issue(s)**:

Fixes #9755


nickpoorman:
Hmm, I guess the desired behavior is to keep creating writers even in the event that a close has occurred. Reverting back to the lock.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -106,18 +110,23 @@ public StreamFileProducerConcurrent(
...
-        }
+            if (runningHashes.runningHash() == null) {
+                throw new IllegalArgumentException("The initial running hash cannot be null");
+            }
 
-        lastRecordHashingResult = completedFuture(runningHashes.runningHash());
-        lastRecordHashingResultNMinus1 = completedFuture(runningHashes.nMinus1RunningHash());
-        lastRecordHashingResultNMinus2 = completedFuture(runningHashes.nMinus2RunningHash());
-        lastRecordHashingResultNMinus3 = completedFuture(runningHashes.nMinus3RunningHash());
+            lastRecordHashingResult = completedFuture(runningHashes.runningHash());
```
jsync-swirlds, line 123: If we make some of these volatile, could we avoid any of the locks below (e.g. in `switchBlocks`)?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -143,34 +152,40 @@ public void switchBlocks(
...
-                                final var lastRunningHash = twoResults.b();
-                                closeWriter(writer, lastRunningHash);
-                                return createBlockRecordWriter(
-                                        lastRunningHash, newBlockFirstTransactionConsensusTime, newBlockNumber);
-                            },
-                            executorService);
+            if (currentRecordFileWriter == null) {
+                // We are at the start of a new block and there is no old one to close or wait for. So just create a new
+                // one
+                // which creates a new file and writes initializes it in the background
```
jsync-swirlds, line 170: Could combine these two lines
```suggestion
                // one which creates a new file and writes initializes it in the background
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -143,34 +152,40 @@ public void switchBlocks(
...
+                currentRecordFileWriter = lastRecordHashingResult.thenApply(lastRunningHash -> createBlockRecordWriter(
```
jsync-swirlds, line 171: I know this isn't new for this PR, but if we're avoiding races, then this should probably be `thenApplyAsync` rather than `thenApply`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -143,34 +152,40 @@ public void switchBlocks(
...
+                        lastRunningHash, newBlockFirstTransactionConsensusTime, newBlockNumber));
+            } else {
+                // wait for all background threads to finish, then in new background task finish the current block
+                currentRecordFileWriter = currentRecordFileWriter
```
jsync-swirlds, line 175: Worth noting that this won't run the indicated code within the context of the lock; the lastRecordHashingResult used will be whatever value is current when the asynchronous tasks actually run on some other thread.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -182,64 +197,79 @@ public void switchBlocks(
...
-                                    writer.writeItem(item);
-                                } catch (final Exception e) {
-                                    logger.error("Error writing record item to file", e);
-                                }
-                            });
-                            return writer;
-                        },
-                        executorService);
+        try {
+            assert lastRecordHashingResult != null : "initRunningHash() must be called before writeRecordStreamItems";
```
jsync-swirlds, line 203: Both of these should probably be upgraded to full conditionals rather than test/debug-only assert.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -182,64 +197,79 @@ public void switchBlocks(
...
-                        closeWriter(writer, lastRunningHash);
-                    })
-                    .join();
+        lock.lock(); // Block until the lock is acquired
+
+        try {
+            if (currentRecordFileWriter != null) {
+                CompletableFuture.allOf(currentRecordFileWriter, lastRecordHashingResult)
+                        .thenAccept(aVoid -> {
+                            final var writer = currentRecordFileWriter.join();
```
jsync-swirlds, line 259: These `join` calls should probably be `get` calls to avoid unnecessary thread switching.

netopyr:
We should discuss this with @jasperpotts. I'd hate to introduce a lock on the handle thread.

netopyr (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
nickpoorman: I agree we should do this in a future PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
nickpoorman: I agree that the future won't run in the context of the lock. My concern was with the setting of the future into `currentRecordFileWriter` as well as the reading of variables like `lastRecordHashingResult`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
nickpoorman: Good point. Probably good for a follow-up PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
nickpoorman: Yes, there are future-related concepts in this class that need optimizing but that's outside the scope of what this PR is trying to fix. Good point for a follow up PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -182,64 +197,79 @@ public void switchBlocks(
...
+            lastRecordHashingResultNMinus2 = lastRecordHashingResultNMinus1;
+            lastRecordHashingResultNMinus1 = lastRecordHashingResult;
+            lastRecordHashingResult = lastRecordHashingResult
+                    .thenCombine(futureSerializedRecords, TwoResults::new)
+                    .thenApplyAsync(
+                            twoResults -> format.computeNewRunningHash(twoResults.a(), twoResults.b()),
+                            executorService);
+            // write serialized items to record file in a background thread
+            currentRecordFileWriter = currentRecordFileWriter
+                    .thenCombine(futureSerializedRecords, TwoResults::new)
```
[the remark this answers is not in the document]
nickpoorman: Good point for a follow up PR.

nickpoorman:
Note to future readers: this introduces a ReentrantLock, which under zero-contention should have almost no significant overhead since the only time there is contention on these methods is when the server calls `close` to shut down. I believe at most this introduces a volatile write.

jsync-swirlds (commented):
Resolving conversations to move this forward, but these concerns still need to be incorporated in future work.

nickpoorman:
This thing got all messed up after trying to use Github to apply a change, the DCO didn't apply correctly.

netopyr (approved):

jsync-swirlds (approved):

PR #9765, opened 2023-11-08
Support complex keys in `VerificationStrategy.asSignatureTestIn()` standalone key verifier

### LINKED ISSUE #9756: Fix `VerificationStrategy.asSignatureTestIn()` for complex key structures
### Problem

The standalone signature verifier created [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/VerificationStrategy.java#L61) needs to deal with complex keys.

### Solution

Fix this.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9756 
 - The `ApprovalSwitchHelper` switches debits that don't have authorization from a key's signature into approval-based debits.
    * But in `develop` the `VerificationStrategy.asSignatureTestIn()` does not return a verification test that works with complex key structures.
    * Therefore we see a spec like `rollbackOnFailedMintAfterFungibleTransfer()` incorrectly use approval-based debits.
 - This PR updates the standalone key verifier to work with complex keys.


dikel (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM !

lukelee-sl (approved):
LGTM

PR #9769, opened 2023-11-08
Fix some tests from LazyCreateThroughPrecompileSuite

MiroslavGatsanoga:
**Description**:
Further refines the checks enforced (or not) when a synthetic transaction is dispatched:
  1. Makes the `Predicate<Key>` callback `@Nullable` so e.g. hollow account finalization can dispatch a `CryptoUpdate` targeting an immutable account without getting `INVALID_SIGNATURE`.
  2. Limits the "HAPI-style" payer signature and solvency checks to **only** transactions dispatched by the `ScheduleService` (these are never appropriate for `ContractService` dispatches).
  3. Fixes the dispatch `validate()` method to actually validate the dispatched `TransactionBody` instead of the top-level `TransactionBody` 😱 .
  4. Adds a `FailureCustomizer` interface to `DispatchForResponseCodeHtsCall`, uses it for `TokenUpdate` dispatches to [customize some pre-handle failure codes](https://github.com/hashgraph/hedera-services/blob/68cf767244d1143234c101fe898bcb84ef3f6a79/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java#L52) in the expected way.
  5. Fixes all but 3 `ContractXTest`'s; the remaining three failures are:
      - `CreatesXTest`: Needs status customization for `INVALID_ACCOUNT_ID` -> `INVALID_TREASURY_FOR_TOKEN`
      - `ClassicViewsXTest` - `GET_TOKEN_INFO` wrong (likely precedes payer validation PR merge)
      - `SetApprovalForAllXTest` - Fails with `"given selector does not match: expected: 367605ca, found: a22cb465"` (likely precedes payer validation PR merge)

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/9842 (enables 5 `@HapiTest`)

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
```diff
@@ -163,7 +163,10 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
                 // converted into the same EVM address as the alias
                 final var payerEvmAddress = extractEvmAddress(payerKey);
                 final var accountKeyEvmAddress = extractEvmAddress(op.keyOrThrow());
-                if (!alias.equals(payerEvmAddress) && !alias.equals(accountKeyEvmAddress)) {
+                // payerEvmAddress will be null for cases where we have synthetic crypto-create used to create a
+                // contract as part of a contract call or contract create,
+                // those cases should not be treated as hollow account creation
+                if (payerEvmAddress != null && !alias.equals(payerEvmAddress) && !alias.equals(accountKeyEvmAddress)) {
```
MiroslavGatsanoga, line 169: @tinker-michaelj Not sure if this is ok to be done here, but was not able to think of an alternative. Wdyt? 🤔

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -528,21 +535,22 @@ <T> T dispatchChildTransaction(
      * @throws IllegalArgumentException if the transaction body did not have an id
      */
     @NonNull
-    default <T> T dispatchChildTransaction(
+    default <T> T dispatchScheduledChildTransaction(
```
jsync-swirlds, line 538: Rather than create this new method, it may be cleaner to just have Schedule Service pass in the `SCHEDULED` category.  The same lines changed for the method rename could pass the category parameter instead.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
MiroslavGatsanoga: Actually if the payer key for the top-level contract call/create is ECDSA key then we'll still end up inside the `if` which does not seem right, so probably we need to do something else. 😞

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
tinker-michaelj: True @jsync-swirlds , I think once everything is passing we may want to consolidate to a single `dispatch()` method that receives a `DispatchModes` argument (or the like) to discriminate the desired behavior. For now, will just try to get all needed dispatch behaviors confirmed working. 😝

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -704,80 +697,94 @@ private void dispatchSyntheticTxn(
...
 
+            syntheticPayerKey = payerAccount.keyOrThrow();
+            requireNonNull(keyVerifier, "keyVerifier must not be null when enforcing HAPI-style payer checks");
+            final var payerKeyVerification = keyVerifier.verificationFor(syntheticPayerKey);
+            if (payerKeyVerification.failed()) {
+                throw new PreCheckException(INVALID_SIGNATURE);
+            }
+        }
+
+        // Given the current HTS system contract interface and ScheduleService
```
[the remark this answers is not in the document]
tinker-michaelj: Agreed, makes sense. 👍

tinker-michaelj (approved):
LGTM

MiroslavGatsanoga:
Created a follow up issue for fixing the 3 left `ContractXTest`s.

PR #9778, opened 2023-11-08
09427: TokenAndTypeCheckSuite tests fix.

### LINKED ISSUE #9427: Unconditional, enriched records of read-only HTS system contract calls.

ilko-iliev-lime:
**Description**:
Purpose of the PR is fixing the TokenAndTypeCheckSuite. Problems with the `checkTokenAndTypeNegativeCases` were the following: 

1. In case of a call with invalid token ID to the IsTokenCall class, the status of the result was INVALID_TOKEN_ID, which doesn't match the mono output (success in all cases and false on the isToken field). 
2. In case of a call with invalid token ID to the TokenTypeCall class, the status of the result was REVERTED_SUCCESS, which is a status which overrides the correct one after it is returned, so I had to override the execute function. 

**Related issue(s)**:

#9427 

**Notes for reviewer**:

**Checklist**


tinker-michaelj:
Ah looks like we are missing verified signoffs on commits here.

jeromy-cannon (approved):

jsync-swirlds:
This has a bad merge and is rewriting several commits already present on develop (which is why the reviewer set expanded).
That needs to be fixed before trying to merge to develop.

jsync-swirlds (changes requested):
Please fix the cross-merge issue noted above prior to merge.

PR #9781, opened 2023-11-09
Fix `ContractCreate` child record stream items (both top-level and internal)

### LINKED ISSUE #9775: Suppress dispatched child record for top-level `ContractCreate`; sub synth `CryptoCreate` body with `ContractCreate` for internal
### Description

The top-level `ContractCreate` should not produce a child record for the internal dispatch done [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java#L297).

### Steps to reproduce

Run any `@HapiTest` using `ContractCreate` with record snapshot matching on.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #9775 
 - Adds an SPI type [`ExternalizedRecordCustomizer`](https://github.com/hashgraph/hedera-services/blob/09775-fix-sc-create-children/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java#L22) that handlers can use to customize the records of their dispatched child transactions---either to suppress the records entirely, or transform the externalized `Transaction` (see javadoc for use cases).
 - Implements record suppression in `RecordListBuilder`.
 - Implements externalized transform in `SingleTransactionRecordListBuilderImpl`.
 - ➡️ Then uses the `ExternalizedRecordCustomizer` in `HandleHederaOperations` to:
     1. Suppress the child record of a dispatch used to implement a top-level `ContractCreate`.
     2. Replace the dispatched `CryptoCreateTransactionBody` of an internal contract creation with the equivalent `ContractCreateTransactionBody` before externalization to the record stream.
 - Addresses reviewer comments from https://github.com/hashgraph/hedera-services/pull/9774, primarily making it explicit that `gasPrice` and `gasCost` fields are denominated in tinybar.


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Allows a {@link com.hedera.node.app.spi.workflows.TransactionHandler} that dispatches child transactions
+ * to customize exactly how these records are externalized. Specifically, it allows the handler to,
+ * <ul>
+ *     <li>Completely suppress the record because it contains redundant information (as in the case of
+ *     the child transaction dispatched to implement a top-level HAPI {@code ContractCreate}).</li>
+ *     <li>Transform the dispatched {@link Transaction} immediately before it is externalized (as
+ *     in the case of the child {@link com.hedera.hapi.node.token.CryptoCreateTransactionBody} dispatched
+ *     to implement an internal contract creation, which should be externalized as an equivalent
+ *     {@link com.hedera.hapi.node.contract.ContractCreateTransactionBody}.</li>
```
david-bakin-sl, line 31: Looks like `apply` is never called if `shouldSuppressRecord` returns `true`. (I infer this from the two implementations where `apply` simply throws an exception.). Could this be explicitly mentioned here?

I really _like_ this kind of customization point, BTW.

(Though an alternative would be an `apply` that returned an `Optional`.  (not extending `UnaryOperator` of course.) Java pattern matching in switch statements is almost enough to help but the rest of the type system isn't strong enough to make the whole thing worth it yet.)

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java
```diff
@@ -0,0 +1,46 @@
...
+ * </ul>
+ */
+@FunctionalInterface
+public interface ExternalizedRecordCustomizer extends UnaryOperator<Transaction> {
+    ExternalizedRecordCustomizer NOOP_EXTERNALIZED_RECORD_CUSTOMIZER = tx -> tx;
```
david-bakin-sl, line 36: You know, why bother to define `UnaryOperator::identity` when this obvious idiom `i -> i` is so much shorter and at least as clear to read? 

(Plus I couldn't even figure out how to use it here as a method reference due to language/type system limitations.)

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/records/GasFeeRecordBuilderTest.java
```diff
@@ -17,21 +17,29 @@
...
+            public long transactionFee() {
+                return totalFee;
+            }
+
+            @Override
+            public GasFeeRecordBuilder transactionFee(final long transactionFee) {
+                totalFee = transactionFee;
+                return this;
+            }
+        };
```
david-bakin-sl, line 40: Thank you! I do happen to like this better.  (Looks worse than it should due to canned formatting rules inserting lots of newlines for what ought to be one-liners ... and because you were careful to give a name to `totalFee` instead of just jamming it into L32 ... which is nice.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
```diff
@@ -159,13 +159,23 @@ public class SnapshotModeOp extends UtilOp implements SnapshotOp {
 
     public static void main(String... args) throws IOException {
         // Helper to review the snapshot saved for a particular HapiSuite-HapiSpec combination
-        final var snapshotFileMeta = new SnapshotFileMeta("CryptoTransfer", "okToRepeatSerialNumbersInBurnList");
-        final var snapshot = loadSnapshotFor(PROJECT_ROOT_SNAPSHOT_RESOURCES_LOC, snapshotFileMeta);
+        final var snapshotFileMeta = new SnapshotFileMeta("ContractCall", "MultipleSelfDestructsAreSafe");
```
david-bakin-sl, line 162: I get that this is just a CLI utility for helping diagnose failing BDDs, but I don't get why/when you'd change the `SnapshotFileMeta` parameters?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
```diff
@@ -159,13 +159,23 @@ public class SnapshotModeOp extends UtilOp implements SnapshotOp {
...
+        final var maybeSnapshot = suiteSnapshotsFrom(
+                        resourceLocOf(PROJECT_ROOT_SNAPSHOT_RESOURCES_LOC, snapshotFileMeta.suiteName()))
+                .flatMap(
+                        suiteSnapshots -> Optional.ofNullable(suiteSnapshots.getSnapshot(snapshotFileMeta.specName())));
+        if (maybeSnapshot.isEmpty()) {
+            throw new IllegalStateException("No such snapshot");
+        }
+        final var snapshot = maybeSnapshot.get();
         final var items = snapshot.parsedItems();
+        final var dumpLoc = Files.newBufferedWriter(Paths.get(snapshotFileMeta + ".txt"));
```
david-bakin-sl, line 172: Also given this usage in a utility CLI I honestly don't care but IntelliJ points out that this file should be opened under a try-with-resources.

david-bakin-sl (commented):
Halfway though this really nice PR but going for lunch so submitting current comments now.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java
tinker-michaelj: Added a comment and [convenience implementation](https://github.com/hashgraph/hedera-services/blob/27c8a835a98208da9326948af21f0c7c7dfb08d3/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java#L43) that suppresses the record and throws on receiving an `apply()` call.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
tinker-michaelj: Ah interestingly this _is_ important, as a `BufferedWriter` will not necessarily flush its contents to the file without it!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
tinker-michaelj: Ah interestingly this _is_ important, as a `BufferedWriter` will not necessarily flush its contents to the file without it! 

Fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
david-bakin-sl: I did get _that_ ... what I _didn't_ get was why you'd change the arguments to `SnapshotFileMeta` but that was because I stupidly _didn't look at the constructor signature_ which tells all.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
david-bakin-sl: oops!  now I Iearned something too, thanks!

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/ExternalizedRecordCustomizer.java
david-bakin-sl: Yes! That's what I should have suggested ... well, next time ....

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -192,8 +199,23 @@ public SingleTransactionRecordBuilderImpl(@NonNull final Instant consensusNow) {
...
+     * Creates new transaction record builder with both explicit reversing behavior and
+     * transaction construction finishing.
+     *
+     * @param consensusNow the consensus timestamp for the transaction
+     * @param reversingBehavior the reversing behavior (see {@link RecordListBuilder}
+     */
+    public SingleTransactionRecordBuilderImpl(
+            @NonNull final Instant consensusNow,
+            @NonNull final ReversingBehavior reversingBehavior,
+            @Nullable final ExternalizedRecordCustomizer customizer) {
```
david-bakin-sl, line 215: Out of curiosity, why allow it to be nullable?  Almost as easy to provide one, especially given `NOOP_EXTERNALIZED_RECORD_CUSTOMIZER`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -202,6 +224,9 @@ public SingleTransactionRecordBuilderImpl(
      * @return the transaction record
      */
     public SingleTransactionRecord build() {
+        if (customizer != null) {
+            transaction = customizer.apply(transaction);
+        }
```
david-bakin-sl, line 229: I don't see a unit test for this?  (e.g., with a non-identify `apply`)

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/record/RecordListBuilderTest.java
```diff
@@ -645,6 +646,43 @@ void testAddMultipleRemovableChildren() {
...
+            @Override
+            public Transaction apply(Transaction transaction) {
+                throw new UnsupportedOperationException("The top-level creation record should be suppressed");
+            }
+
+            @Override
+            public boolean shouldSuppressRecord() {
+                return true;
+            }
+        };
```
david-bakin-sl, line 666: You could now use `SUPPRESSING_EXTERNALIZED_RECORD_CUSTOMIZER` here (which of course works because you gave it its own test)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -319,17 +345,30 @@ private void dispatchAndMarkCreation(
...
-                // add dummy transaction, because SingleTransactionRecord require NonNull on build
-                .transaction(Transaction.newBuilder()
-                        .signedTransactionBytes(Bytes.EMPTY)
-                        .build())
-                .contractCreateResult(ContractFunctionResult.newBuilder()
-                        .contractID(contractId)
-                        .evmAddress(evmAddress)
-                        .build());
+    private ExternalizedRecordCustomizer contractBodyCustomizerFor(@NonNull final ContractCreateTransactionBody op) {
+        return transaction -> {
```
david-bakin-sl, line 349: lambda syntax and 2nd- (and higher-) order functions provide such sweet expressions of intent ...

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/scope/HandleHederaOperationsTest.java
```diff
@@ -249,16 +262,51 @@ void createContractWithParentDispatchesAsExpectedThenMarksCreated() {
...
+                .signedTransactionBytes(SignedTransaction.PROTOBUF.toBytes(SignedTransaction.newBuilder()
+                        .bodyBytes(TransactionBody.PROTOBUF.toBytes(cryptoCreateBody))
+                        .build()))
+                .build();
+        final var cryptoCreateOutput = internalFinisher.apply(cryptoCreateInput);
+        final var finishedBody = TransactionBody.PROTOBUF.parseStrict(SignedTransaction.PROTOBUF
+                .parseStrict(cryptoCreateOutput.signedTransactionBytes().toReadableSequentialData())
+                .bodyBytes()
+                .toReadableSequentialData());
+        assertEquals(expectedOp, finishedBody.contractCreateInstanceOrThrow());
```
david-bakin-sl, line 290: Just a comment: This kind of test is a place you could use soft assertions.  You have three assertions here in one custom assert method - but though they're related they're not dependent (I don't think).  And so with soft assertions you could , in one test, get _all_ failures of the finisher, not just the first to hit.  For example.

david-bakin-sl (approved):
LGTM (with the usual kind of comments of course)

lukelee-sl (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
tinker-michaelj: Made `customizer` non-null.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/scope/HandleHederaOperationsTest.java
tinker-michaelj: Ah yes stillll need to learn `AssertJ` 😳 .

Neeharika-Sompalli (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
david-bakin-sl: That isn't what I meant really.  I meant: Only things I see tested are where the customizer is the identity function or the suppressor.  There isn't a direct test with a customizer that does something non-trivial.  (There's a test where the cryptocreate gets changed to a contractcreate - I think I saw that yesterday though can't find it now - but that's rather indirect and I can't find a _direct_ test of this, which should probably be in `SingleTransactionRecordBuilderTest`.)

Issue #9786, opened 2023-11-09
Define goals and best pratices for platform base


hendrikebbers:
We should create a short documentation about the scope of Platform Base

PR #9787, opened 2023-11-09
Documentation of the Platform Base Scope

hendrikebbers:
fixes https://github.com/hashgraph/hedera-services/issues/9786


platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,49 @@
...
+
+The modules that belong to Platform Base must fulfill the following requirements:
+
+- They must provide an easy to understand API
+- They must provide documentation for all public interfaces, classes, methods...
+- They must provide documentation for each package (`package-info.java`)
+- Each module must be a Java module
+- Today all features must be supported on the modulepath and the classpath (might be modulepath only in future).
+- If the project provide test fixtures, they must be provided as a module as well.
+- A goal is to have all unit tests defined as modules, too.
```
jsync-swirlds, line 15: This should not be a goal because then packages cannot be replicated in test code, so code exposure for test purposes is greatly expanded, making encapsulation almost non-existent.
Unit tests not in modules is perfectly reasonable, particularly as unit test code should never be reused or published.

platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,49 @@
+# Platform Base
+
+Platform base is a superset of modules that belong to the base team and are used everywhere within the project.
```
jsync-swirlds, line 3: superset ==> set?

platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,49 @@
...
+- swirlds-logging (`com.swirlds.logging`)
+- swirlds-common (`com.swirlds.common`)
+- swirlds-config-impl (`com.swirlds.config.impl`)
+- swirlds-config-benchmark (`com.swirlds.config.impl`)
+
+Additional modules will follow in future since we plan to split swirlds-common into multiple modules.
+
+## Structure and best practices in the rest of the repository
+
+Since Platform Base defines a lot of pattern and best practices for the module the team sees it as a good idea to follow the same pattern in the rest of the repository.
```
jsync-swirlds, line 35: `Since` is unnecessary

platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,49 @@
...
+Therefore, the Platform Base team tries to help all other colleagues to have a great structure and module definition in the complete repository.
+Here the team creates issues and pull requests to help all the other teams to have a great structure as well.
+Next to that the Platform Base team tries to remove as many external dependencies as possible to make it easier for all other teams to use the modules.
+Here the team again creates issues and pull requests to remove external dependencies and replace them with internal modules in all modules of the repository.
```
jsync-swirlds, line 39: This is both a bit awkward and includes several repetitions. The following _might_ be a better wording.
```suggestion
Platform Base works to ensure that all modules within our scope of responsibility are well designed and serve as examples of best practice.  To this end we define patterns and recommended practices, create well defined issues, and conduct high quality pull request review.
All modules in Platform Base declare the absolute minimum external dependencies so that modules are more widely usable.
Where necessary Platform Base will create issues and submit pull requests to improve design throughout the entire system, and to continually minimize external dependencies.
```

platform-sdk/docs/base/base.md
hendrikebbers: Well but code behaves differently on the module path against the class path. Therefore having unit test per definition on the class path ends in different behavior at test time VS at runtime. Replicating packages is a pattern that allows tests to sneak into classed by accessing package private methods. From my point of view we now have a really good separation between public and private API and a private API does not need to be package private anymore. Just move it to a private package and export it only for the test.

platform-sdk/docs/base/base.md
hendrikebbers, line 3:
```suggestion
Platform base is a set of modules that belong to the base team and are used everywhere within the project.
```

platform-sdk/docs/base/base.md
hendrikebbers:
```suggestion
Platform Base defines a lot of pattern and best practices for the module the team sees it as a good idea to follow the same pattern in the rest of the repository.
```

hendrikebbers:
```suggestion
Platform Base defines a lot of pattern and best practices for the module the team sees it as a good idea to follow the same pattern in the rest of the repository.
```

poulok (approved):

platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,50 @@
+# Platform Base
+
+Platform base is a set of modules that belong to the base team and are used everywhere within the project.
```
imalygin, line 3: belong -> belongs

platform-sdk/docs/base/base.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Goals
+
+The modules that belong to Platform Base must fulfill the following requirements:
+
+- They must provide an easy to understand API
+- They must provide documentation for all public interfaces, classes, methods...
```
imalygin, line 10: triple dot is unnecessary here :)

platform-sdk/docs/base/base.md
jsync-swirlds: Multiple modules belong; a single module belongs.

imalygin (dismissed):

platform-sdk/docs/base/base.md
hendrikebbers, line 10:
```suggestion
- They must provide documentation for all public interfaces, classes, methods...
```

platform-sdk/docs/base/base.md
hendrikebbers: SPI works in a different way for example between module path and class path. Reflection is another example. The definition in the module-info of a project or of dependent project can change how code behaves. That is just a fact.

jsync-swirlds: These are still packaging and integration differences.
Sacrificing object encapsulation to try to find integration issues in unit tests is, in my opinion, a flawed choice.
Integration or cross-cut tests are a much better place to discover such things.  Those tests never need access to object internals or package private elements, absolutely should be in different modules/packages, and can discover these external integration issues more effectively than unit tests.

hendrikebbers: I need to disagree on that one. If I have a unit that internally uses reflection I already have that problem. But that can be solved by having some definitions (for example using Google Auto Service for SPI). Anyway we need a global (platform+services) definition how tests (unit, integration,...) are executed.

hendrikebbers: I removed that point for now to get the PR merged ;)

poulok (approved):

platform-sdk/docs/base/base.md
jsync-swirlds: Is there a response to this comment, or did it get resolved accidentally?

hendrikebbers: Since there were several changes in that section it somehow got missed. I assumed I just copies your change and committed it.

jsync-swirlds (approved):

PR #9795, opened 2023-11-09
Externalize the generation of child records for view calls

### LINKED ISSUE #8920: Implement `calculateFees` in ContractCallHandler
We have a failing test `cannotSendValueToTokenAccount` in `ContractDeleteSuite` that depends on proper fee calculation in order to pass.

### LINKED ISSUE #9565: Fix E2E tests in ContractMintHTSSuite

### LINKED ISSUE #9734: Fix ContractKeysHTSSuite tests

### LINKED ISSUE #9848: Generalize child record externalization on a higher level in HtsSystemContract
Currently child records are populated separately on every view method call. The essence of this enhancement is to generalize this behaviour on a higher level for example in `HtsSystemContract` so that we can remove the redundant code

petreze:
**Description**:
Reduces code duplication and fixes several remaining details involved in externalizing child `ContractCallResult`'s. 

1️⃣ In `hedera-app`,
 - Wires up and uses `ChildRecordFinalizer` to set transfer lists in the records of dispatched transactions.
 - Removes `ExchangeRateSet` from all child record receipts (not just preceding).
 - Adds record builder getters for `newTotalSupply` and minted serial numbers (we need this to encode output from burn/mint HTS system contract calls).

2️⃣ In `contract-service`,
  - Adds `SystemContractOperations#externalizePreemptedDispatch()` method to let `HtsCall` implementations externalize dispatches that were pre-empted for reasons not enforced by handlers (e.g. to prevent transfers to system accounts).
  - Reduces duplicate code in externalized view call results by adding an `isViewCall` flag to `FullResult` and externalizing results in `HtsSystemContract`.
  - Adds a `ContractCallRecordBuilder` to `FullResult` so that `HtsCall` implementations can directly return their used `ContractCallRecordBuilder`.
      * Updates `HtsSystemContract` to add the implied synthetic `ContractCallResult` to such record builders.
  - Makes `FullResult` a top-level record for readability.
  - Adds `HtsCall.allowsStaticFrame()` and reverts in `HtsSystemContract` when the frame is static and this method returns false.
  - Updates `GrantRevokeKycTranslator` to match mono-service behavior of leaving `INVALID_SIGNATURE` response code untouched (instead of converting to `INVALID_FULL_PREFIX_SIGNATURE_FOR_PRECOMPILE`).
  - Fixes burn and mint return types (neither is just an `int64 responseCode`).
  - Fixes `ClassicTransfersCall` to:
      1. Revert or halt on unsupported behavior.
      2. Emit ERC-style log events on `transferFrom()` and `transferFromNFT()`.
      3. Consolidate multiple `tokenTransferList`'s with the same `TokenID`.

3️⃣ In `token-service`,
  - Fixes `FinalizeParentRecordHandler` to deduct child token transfers from the top-level record (unlike in mono-service these need to be explicitly deducted from parent).
  - Reverses order of modifying allowances and token balances in `AdjustFungibleTokenChangesStep` to match mono-service behavior.
  - Adds a `ChildFinalizeContext` as the base interface of `FinalizeContext`, with just what is needed to finalize a single child record.

4️⃣ In `test-clients`,
 - Adds record snapshots for `CryptoTransferHTSSuite`
 - Adds some helpers to make record assertions more flexible if needed.
 - Adds `ACCEPTED_MONO_GAS_CALCULATION_DIFFERENCE` snapshot matching mode to allow mod-service gas calculations to differ in edge cases.
 - Enables 55-ish `@HapiTest`s.

➡️ _Next steps:_
 - Eliminate all duplicate code in externalized call results.
 - Remove type parameter from `DispatchForResponseCodeHtsCall` (can always just use `ContractCallRecordBuilder`).
 - Try to replace all non-transfer `HtsCall` implementations with `DispatchForResponseCodeHtsCall` instances where custom behavior is achieved via `FailureCustomizer` and `OutputFn` instead of standalone `HtsCall` implementations.
 - Remove `blockValues`, `value`, `isStaticCall` from `HtsCallAttempt` (these are now available on the `MessageFrame` passed to `execute()`).

**Related issue(s)**:

- Fixes #8920 
- Fixes #9848 
- Fixes #9734 
- Fixes #9565 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
```diff
@@ -70,6 +76,31 @@ private static FullResult resultOfExecuting(
         final HtsCall.PricedResult pricedResult;
         try {
             pricedResult = call.execute(frame);
+            final var proxyWorldUpdater = FrameUtils.proxyUpdaterFor(frame);
+            if (proxyWorldUpdater != null) {
```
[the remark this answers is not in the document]
petreze: Yes, I would agree to that in a normal scenario but running `HtsSystemContractTest.java` without the check yields 2 failing tests because of `proxyWorldUpdater ` is null in these cases

image

So we should either adapt the tests or leave the check..
Your thoughts?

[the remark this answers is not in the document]
tinker-michaelj: Ah yes we should make those unit tests mock the `MessageFrame` as necessary to have a non-null updater here. 🙏

tinker-michaelj (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/TokenContextImpl.java
```diff
@@ -84,6 +84,11 @@ public <T> T userTransactionRecordBuilder(@NonNull Class<T> recordBuilderClass)
         return castRecordBuilder(recordListBuilder.userTransactionRecordBuilder(), recordBuilderClass);
     }
 
+    @Override
+    public boolean hasChildRecords() {
+        return !recordListBuilder.childRecordBuilders().isEmpty();
```
Neeharika-Sompalli, line 89: Should this include precedingChildRecords also ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -382,13 +383,7 @@ public SingleTransactionRecordBuilderImpl syncBodyIdFromRecordId() {
...
                     .transactionID(newTransactionID)
                     .build();
-            final var newBodyBytes = TransactionBody.PROTOBUF.toBytes(body);
-            final var newSignedTransaction =
-                    SignedTransaction.newBuilder().bodyBytes(newBodyBytes).build();
-            final var signedTransactionBytes = SignedTransaction.PROTOBUF.toBytes(newSignedTransaction);
-            this.transaction = Transaction.newBuilder()
-                    .signedTransactionBytes(signedTransactionBytes)
-                    .build();
+            this.transaction = SingleTransactionRecordBuilder.transactionWith(body);
```
Neeharika-Sompalli, line 386: Cool Thank you 💯

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeParentRecordHandler.java
```diff
@@ -77,52 +83,88 @@ public void finalizeParentRecord(@NonNull final AccountID payer, @NonNull final
...
-        deductChangesFromChildRecords(hbarChanges, context);
+        final var fungibleChanges = fungibleChangesFrom(writableTokenRelStore, tokenStore);
+        final var nftChanges = nftChangesFrom(writableNftStore, tokenStore);
+        if (context.hasChildRecords()) {
+            // All the above changes maps are mutable
+            deductChangesFromChildRecords(context, fungibleChanges, nftChanges, hbarChanges);
+            // In the current system a parent transaction that has child transactions cannot
+            // *itself* cause any net fungible or NFT transfers; fail fast if that happens
+            validateTrue(fungibleChanges.isEmpty(), FAIL_INVALID);
+            validateTrue(nftChanges.isEmpty(), FAIL_INVALID);
```
Neeharika-Sompalli, line 96: You mean all the changes will be in parent record transferList ?

Neeharika-Sompalli (approved):
Just a few questions. Not blocked by the questions. Token Service changes LGTM ! Thanks @tinker-michaelj @petreze !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractNonRevertibleTokenViewCall.java
```diff
@@ -16,77 +16,21 @@
...
 
 /**
  * Implementation support for view calls that require an extant token.
  * Classic view function calls are generally non-revertible.
  */
-public abstract class AbstractNonRevertibleTokenViewCall extends AbstractHtsCall {
-    @Nullable
-    protected final Token token;
-
+public abstract class AbstractNonRevertibleTokenViewCall extends AbstractTokenViewCall {
```
lukelee-sl, line 29: Is this class still necessary?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/FullResult.java
```diff
@@ -0,0 +1,119 @@
...
+import org.hyperledger.besu.evm.precompile.PrecompiledContract;
+
+/**
+ * Represents the result of executing a Hedera system contract.
+ *
+ * @param result the result of the computation
+ * @param gasRequirement the gas requirement of the computation
+ * @param recordBuilder the record builder, if any, generated as a side effect of the computation
+ */
+public record FullResult(
```
lukelee-sl, line 41: it might be nice to have unit test coverage for this.

lukelee-sl (commented):
Much simpler.  Thanks for making the changes.  A few minor comments but definitely non blocking

lukelee-sl (dismissed):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeParentRecordHandler.java
tinker-michaelj: I mean that _if_ the child records include token transfers, then the parent **must** not have any token transfers itself after subtracting all token transfers from child records.

(Since a parent with children is either a contract or schedule operation; neither of which can directly transfer tokens.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
```diff
@@ -16,60 +16,114 @@
...
+                        STATIC_CALL_REVERT_REASON, contractsConfigOf(frame).precompileHtsDefaultGasCost());
+            }
+            log.info("Created call {}", call.getClass().getSimpleName());
         } catch (final RuntimeException e) {
             log.debug("Failed to create HTS call from input {}", input, e);
             return haltResult(ExceptionalHaltReason.INVALID_OPERATION, frame.getRemainingGas());
         }
 
-        return resultOfExecuting(call, input, frame);
+        return resultOfExecuting(attempt, call, input, frame);
```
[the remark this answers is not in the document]
tinker-michaelj: Yes this is all a mess, I am just trying to get all `@HapiTest` passing to make sure all the required information is available everywhere then do a major refactor. 😝

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/AbstractNonRevertibleTokenViewCall.java
tinker-michaelj: Not sure. 😝

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/TokenContextImpl.java
tinker-michaelj: Agreed Neeha, that would be more future-proof.

Neeharika-Sompalli (dismissed):
LGTM !

tinker-michaelj (approved):
LGTM

Issue #9808, opened 2023-11-09
Write PCES files to state directories


cody-littley:
When we write a state file to disk, copy PCES files containing all non-ancient events (relative to the state) into the directly where we write the state. The purpose of this change is to enable us to perform migration tests using production states.

PR #9809, opened 2023-11-09
Copy the PCES into saved state directories.

cody-littley:
Related to #9808 

Needs to be cherry picked to `release/0.44`.

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
@@ -93,7 +93,7 @@ import java.time.Duration;
  * @param replayHashPoolSize                              the number of threads used for hashing events during replay
  * @param copyRecentStreamToStateSnapshots                if true, then copy recent PCES files into the saved state
  *                                                        snapshot directories every time we take a state snapshot. The
- *                                                        files copied are gauranteed to contain all non-ancient events
+ *                                                        files copied are guaranteed to contain all non-ancient events
  *                                                        w.r.t. the state snapshot.
  */
 @ConfigData("event.preconsensus")
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
@@ -77,7 +77,8 @@ import java.time.Duration;
  *                                                        Both. Use this with caution.
  * @param databaseDirectory                               the directory where preconsensus events will be stored,
  *                                                        relative to
- *                                                        {@link com.swirlds.common.config.StateConfig#savedStateDirectory()}.
+ *                                                        {@link
+ *                                                        com.swirlds.common.config.StateConfig#savedStateDirectory()}.
  * @param enableStorage                                   if true, then stream preconsensus events to files on disk. If
  *                                                        this is disabled then a network wide crash (perhaps due to a
  *                                                        bug) can cause transactions that previously reached consensus
@@ -90,6 +91,10 @@ import java.time.Duration;
  * @param replayQueueSize                                 the size of the queue used for holding preconsensus events
  *                                                        that are waiting to be replayed
  * @param replayHashPoolSize                              the number of threads used for hashing events during replay
+ * @param copyRecentStreamToStateSnapshots                if true, then copy recent PCES files into the saved state
+ *                                                        snapshot directories every time we take a state snapshot. The
+ *                                                        files copied are gauranteed to contain all non-ancient events
+ *                                                        w.r.t. the state snapshot.
  */
 @ConfigData("event.preconsensus")
 public record PreconsensusEventStreamConfig(
@@ -106,4 +111,5 @@ public record PreconsensusEventStreamConfig(
         @ConfigProperty(defaultValue = "true") boolean enableStorage,
         @ConfigProperty(defaultValue = "true") boolean enableReplay,
         @ConfigProperty(defaultValue = "1024") int replayQueueSize,
-        @ConfigProperty(defaultValue = "8") int replayHashPoolSize) {}
+        @ConfigProperty(defaultValue = "8") int replayHashPoolSize,
+        @ConfigProperty(defaultValue = "true") boolean copyRecentStreamToStateSnapshots) {}
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
@@ -164,18 +164,21 @@ public final class SignedStateFileWriter {
         writeEmergencyRecoveryFile(directory, signedState);
         writeStateAddressBookFile(directory, signedState.getAddressBook());
         writeSettingsUsed(directory, platformContext.getConfiguration());
-        copyPreconsensusEventStreamFiles(
-                platformContext,
-                selfId,
-                directory,
-                signedState.getState().getPlatformState().getPlatformData().getMinimumGenerationNonAncient());
+
+        if (selfId != null) {
+            copyPreconsensusEventStreamFiles(
+                    platformContext,
+                    selfId,
+                    directory,
+                    signedState.getState().getPlatformState().getPlatformData().getMinimumGenerationNonAncient());
+        }
     }
 
     /**
      * Copy preconsensus event files into the signed state directory. These files are necessary for the platform to use
-     * the state file as a starting point. Note that starting with this only the PCES files in the state file does not
+     * the state file as a starting point. Note: starting a node using the PCES files in the state directory does not
      * guarantee that there is no data loss (i.e. there may be transactions that reach consensus after the state
-     * snapshot that are not replayed), but it does allow a node to start up and participate in gossip.
+     * snapshot), but it does allow a node to start up and participate in gossip.
      *
      * <p>
      * This general strategy is not very elegant is very much a hack. But it will allow us to do migration testing using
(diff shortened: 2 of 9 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+        final List<PreconsensusEventFile> allFiles = new ArrayList<>();
+        final Path preconsensusEventStreamDirectory =
+                PreconsensusEventFileManager.getDatabaseDirectory(platformContext, selfId);
+        try (final Stream<Path> stream = Files.walk(preconsensusEventStreamDirectory)) {
+            stream.filter(Files::isRegularFile).forEach(path -> {
+                try {
+                    allFiles.add(PreconsensusEventFile.of(path));
+                } catch (final IOException e) {
+                    // Ignore, this will get thrown for each file that is not a PCES file
+                }
```
lpetrovic05, line 221: could we avoid this exception by adding a method like `isPcesFile`? I know this was done in a rush, so this is not a requirement

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+                Files.createLink(destination, file.getPath());
+            } catch (final IOException e) {
+                logger.error(
+                        EXCEPTION.getMarker(),
+                        "Exception when hard linking preconsensus event file {} to {}",
+                        file.getPath(),
+                        destination,
+                        e);
+            }
+        }
```
lpetrovic05, line 269: we could also take the max generation of judges as an upper limit to reduce the number of files being copied. or maybe the generation of the keystone event

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+                "Copying {} preconsensus event files to state snapshot directory",
+                filesToCopy.size());
+
+        // Although the last file may be currently in the process of being written, all previous files will
+        // be closed and immutable and so it's safe to hard link them.
+        for (int index = 0; index < filesToCopy.size() - 1; index++) {
+            final PreconsensusEventFile file = filesToCopy.get(index);
+            final Path destination = pcesDestination.resolve(file.getFileName());
+            try {
+                Files.createLink(destination, file.getPath());
```
lpetrovic05, line 260: so if the original files are deleted, these links are broken?

cody-littley: This creates a hard link. Think of a hard link like a reference to an object in java. You aren't copying the file, you just copy a reference to a file. As long as there is at least one reference to the file, then the OS won't actually delete the file contents.

If this were a symbolic link, it would be a different story. Symbolic links are analogous to weak references in java. It doesn't copy the data, but it also doesn't prevent GC on the underlying object/file.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: If we do this, I'd prefer to do it as a stand alone PR. This is the same pattern used internally by the PCES file manager, and so if we change it here we probably want to change it in both places.

Normally I'd consider doing such things in a PR like this, but this is an urgent rush job that needs to be cherry picked ASAP.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
```diff
@@ -90,6 +91,10 @@
  * @param replayQueueSize                                 the size of the queue used for holding preconsensus events
  *                                                        that are waiting to be replayed
  * @param replayHashPoolSize                              the number of threads used for hashing events during replay
+ * @param copyRecentStreamToStateSnapshots                if true, then copy recent PCES files into the saved state
+ *                                                        snapshot directories every time we take a state snapshot. The
+ *                                                        files copied are gauranteed to contain all non-ancient events
```
alittley, line 96: *guaranteed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
         writeStateFile(directory, signedState);
         writeHashInfoFile(directory, signedState.getState());
         writeMetadataFile(selfId, directory, signedState);
         writeEmergencyRecoveryFile(directory, signedState);
         writeStateAddressBookFile(directory, signedState.getAddressBook());
-        writeSettingsUsed(directory, configuration);
+        writeSettingsUsed(directory, platformContext.getConfiguration());
+        copyPreconsensusEventStreamFiles(
+                platformContext,
+                selfId,
```
alittley, line 169: `selfId` may be `null`, but it's being passed into a method where it's annotated `@NonNull`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+                directory,
+                signedState.getState().getPlatformState().getPlatformData().getMinimumGenerationNonAncient());
+    }
+
+    /**
+     * Copy preconsensus event files into the signed state directory. These files are necessary for the platform to use
+     * the state file as a starting point. Note that starting with this only the PCES files in the state file does not
```
alittley, line 176: Confusing grammar "Note that starting with this only the"

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,142 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+
+        // Sort by sequence number
+        Collections.sort(allFiles);
+
+        // Discard all files that either have an incorrect origin or that do not contain non-ancient events
+        final List<PreconsensusEventFile> filesToCopy = new ArrayList<>();
+        final PreconsensusEventFile lastFile = allFiles.get(allFiles.size() - 1);
+        for (final PreconsensusEventFile file : allFiles) {
+            if (file.getOrigin() == lastFile.getOrigin()
+                    && file.getMaximumGeneration() <= minimumGenerationNonAncient) {
```
alittley, line 238: Shouldn't it be > minimumGenerationNonAncient?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: I started work on this, but realized that this information is not currently in the state. Adding it is something we might do eventually, but not something I think is worth it it right away, given the rush.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PreconsensusEventStreamConfig.java
cody-littley: fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: Good catch. We should skip the PCES copy operation if the node ID is null.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley:
```
     * Copy preconsensus event files into the signed state directory. These files are necessary for the platform to use
     * the state file as a starting point. Note: starting a node using the PCES files in the state directory does not
     * guarantee that there is no data loss (i.e. there may be transactions that reach consensus after the state
     * snapshot), but it does allow a node to start up and participate in gossip.
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: Actually, it should be `>=`. This explains the bug I've been chasing based on test results (I was about to add a bunch of logging code to chase down the origin of this bug).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,241 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
-     * @param signedState   the signed state being written to disk
-     * @param configuration the configuration used
+     * @param platformContext the platform context
+     * @param selfId          the id of the platform
+     * @param directory       the directory where all files should be placed
+     * @param signedState     the signed state being written to disk
      */
     public static void writeSignedStateFilesToDirectory(
+            @Nullable final PlatformContext platformContext,
             @Nullable NodeId selfId,
```
alittley, line 153: Ignorable nit: `selfId` may be final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -135,27 +143,241 @@ public static void writeStateFile(final Path directory, final SignedState signed
...
+
+    /**
+     * Copy a list of preconsensus event files into a directory.
+     *
+     * @param filesToCopy     the files to copy
+     * @param pcesDestination the directory where the files should be copied
+     */
+    private static void copyPreconsensusFileList(
+            @NonNull final List<PreconsensusEventFile> filesToCopy, @NonNull final Path pcesDestination)
+            throws IOException {
```
alittley, line 326: This method doesn't throw an `IOException`

alittley (approved):

poulok (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: fixed

alittley (approved):

lpetrovic05 (approved):

PR #9813, opened 2023-11-10
Minor improvements to exception handling and null checks

lukelee-sl:
**Description**:
Handle specific exceptions when applicable and add a null check as appropriate.


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCreateOperation.java
```diff
@@ -50,7 +51,7 @@ protected long cost(@NonNull final MessageFrame frame) {
         final var updater = (ProxyWorldUpdater) frame.getWorldUpdater();
```
david-bakin-sl, line 51: As long as you're adding checks, doesn't the argument `frame` need one?  (If it does, same goes for `CustomCreate2Operation.setupPendingCreation`.)

(I suppose you could argue that if `frame` was ever `null` here in the contract service, at least by the time you got down here, something is so seriously wrong that in fact you could never have got here in the first place! I don't see any other nullity checks for `frame` so that must be what's going on.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCreateOperation.java
```diff
@@ -50,7 +51,7 @@ protected long cost(@NonNull final MessageFrame frame) {
...
 
         final var origin = frame.getRecipientAddress();
-        final var originNonce = updater.getAccount(origin).getNonce();
+        final var originNonce = requireNonNull(updater.getAccount(origin)).getNonce();
         // Decrement nonce by 1 to normalize the effect of transaction execution
         final var address = Address.contractAddress(origin, originNonce - 1);
```
david-bakin-sl, line 56: How come `CustomCreate2Operation.setupPendingCreation` doesn't have to decrement the nonce?  What's the difference?

david-bakin-sl (approved):
LGTM with one actual question.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCreateOperation.java
lukelee-sl: Yes agreed if frame is null then something catastrophic has happened.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCreateOperation.java
lukelee-sl: Because when determining the address where we need to create the new contract, for `Create2` we are given the address and for `Create` we need to calculate the address and part of the address calculation depends on using the correct nonce.

david-bakin-sl (approved):
LGTM

tinker-michaelj (approved):
LGTM!

PR #9814, opened 2023-11-10
Detect delegated sender authorization in `CustomMessageCallProcessor` 

### LINKED ISSUE #8717: Fix `delegateContractIdRequiredForTransferInDelegateCall` test from `ContractCreateSuite`
Spent some time debugging it as part of https://github.com/hashgraph/hedera-services/issues/8637 but it does not look like something straightforward to fix, seems to be failing due [to the initial transfer to the beneficiary](https://github.com/hashgraph/hedera-services/blob/c672967964359c08a8b21c4a17563a1b3e599f51/hedera-node/test-clients/src/main/resource/contract/contracts/SendInternalAndDelegate/SendInternalAndDelegate.sol#L17) in the contract function for some reason and after that not sure if it will pass with the delegate call.

Nodes from Valentin T: I spent some time debugging but I had to hop to a more urgent story.
The `contractCall` that is failing has a comment that is helpful to understand what is going on:
  ```
/* Without delegateContractId permissions, the second send via delegate call will
   * fail, so only half of totalToSend will make it to the beneficiary. (Note the entire
   * call doesn't fail because exceptional halts in "raw calls" don't automatically
   * propagate up the stack like a Solidity revert does.) */
```

Comparing the mono and mod they:
- Both have **EXCEPTIONAL_HALT** status in `AbstractMessageProcessor.process`
- Both have true for these two **ifs** changing the status of the frame to `COMPLETED_FAILED`:
```
if (frame.getState() == MessageFrame.State.COMPLETED_FAILED) {
      if (operationTracer != null && frame.getMessageStackSize() > 1) {
        operationTracer.traceContextExit(frame);
      }
      completedFailed(frame);
}
``` 
- The difference is that after the process the mono initialFrame status is `COMPLETED_SUCCESS` but for mod is `COMPLETED_FAILED`

Since we can't control the AbstractMessageProcessor I think it's something to do with how we initialize the `initialFrame`. Or even maybe something before that.

tinker-michaelj:
**Description**:
 - Closes #8717 
 - On `develop`, the `CustomMessageCallProcessor` incorrectly checks whether the **executing** frame is a `delegatecall` to decide when to activate only `delegatable_contract_id` keys in a `receiverSigRequired` key structure.
     * But this just limits a contract's ability to use `DELEGATECALL` to transfer value to an account it controls, which does not make sense.
     * Instead we want to limit when a contract can use _delegated code_ to transfer value to an account that it controls.
 - This PR fixes the mistake by adding a `FrameUtils.acquiredSenderAuthorizationViaDelegateCall()` method that instead tests if the executing frame acquired its `sender` authorization via `delegatecall` (i.e. whether its parent receiver address was running delegated code when it initiated the call represented by the executing frame).

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -17,8 +17,8 @@
 package com.hedera.node.app.service.contract.impl.exec.processors;
 
 import static com.hedera.node.app.service.contract.impl.exec.failure.CustomExceptionalHaltReason.INVALID_FEE_SUBMITTED;
+import static com.hedera.node.app.service.contract.impl.exec.utils.FrameUtils.acquiredSenderAuthorizationViaDelegateCall;
 import static com.hedera.node.app.service.contract.impl.exec.utils.FrameUtils.alreadyHalted;
-import static com.hedera.node.app.service.contract.impl.exec.utils.FrameUtils.isDelegateCall;
 import static com.hedera.node.app.service.contract.impl.exec.utils.FrameUtils.transfersValue;
 import static org.hyperledger.besu.evm.frame.ExceptionalHaltReason.INSUFFICIENT_GAS;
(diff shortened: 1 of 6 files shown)


Neeharika-Sompalli (approved):
LGTM as per hedera-base changes ! Thanks @tinker-michaelj

Neeharika-Sompalli (approved):
LGTM !

lukelee-sl (approved):
LGTM

PR #9824, opened 2023-11-10
chore: Enable HapiTest debug via environment variable

jsync-swirlds:
* Added the environment variable `HAPI_DEBUG_NODE`.  When set to "true" this enables an in-process node for debugging HapiTest specs.
* Fixed a bug in the `HapiTestEngine` that caused failures if a method with non-empty parameter list was annotated with `@HapiTest`; now it skips that test and logs a note.
* Minor fix to exception handling in `HapiTestEnv`.

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
@@ -25,6 +25,7 @@ import com.hedera.services.bdd.spec.HapiSpec;
 import com.hedera.services.bdd.spec.props.JutilPropertySource;
 import com.hedera.services.bdd.suites.HapiSuite;
 import com.hedera.services.bdd.suites.TargetNetworkType;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URI;
 import java.nio.file.Path;
@@ -34,6 +35,9 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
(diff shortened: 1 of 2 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -67,14 +67,15 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster, final
...
             for (int nodeId = 0; nodeId < numNodes; nodeId++) {
-                final var workingDir =
+                final Path workingDir =
                         Path.of("./build/hapi-test/node" + nodeId).normalize();
                 setupWorkingDirectory(workingDir, configText);
-                final var nodeName = NODE_NAMES[nodeId];
-                final var acct = AccountID.newBuilder().accountNum(3L + nodeId).build();
+                final String nodeName = NODE_NAMES[nodeId];
+                final AccountID acct =
+                        AccountID.newBuilder().accountNum(3L + nodeId).build();
```
rbair23, line 78: Not a fan of `var` here I guess? :-)

rbair23 (approved):

Neeharika-Sompalli (approved):
LGTM !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -90,18 +91,13 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster, final
...
         for (final var node : nodes) {
             node.start();
         }
-
         for (final var node : nodes) {
-            try {
-                node.waitForActive(CAPTIVE_NODE_STARTUP_TIME_LIMIT);
-            } catch (TimeoutException e) {
-                throw new RuntimeException(e);
-            }
```
mhess-swl, line 104: Why the removal of this catch?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -178,9 +182,12 @@ public HapiTestEngineExecutionContext before(HapiTestEngineExecutionContext cont
                 return context;
             }
 
+            // Allow for a simple switch to enable in-process Alice node for debugging
+            final String debugEnv = System.getenv("HAPI_DEBUG_NODE");
```
mhess-swl, line 186: 🤩🤩🤩

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
jsync-swirlds: Not for primitives and common classes that aren't much longer than `var`.  It just makes the code harder to read in that case as the variable type is unclear.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
jsync-swirlds: It isn't necessary and made an _already handled_ exception into an unhandled runtime exception.

iwsimon (approved):
LGTM

kimbor (approved):

PR #9832, opened 2023-11-10
Fix `customFeesOnlyUpdatableWithKey` in `TokenUpdateSpec`

georgi-l95:
**Description**:
This PR fixes `customFeesOnlyUpdatableWithKey` in `TokenUpdateSpec` by adding the necessary check in the preHandle and adding the error in the `spec-default.properties`

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/resource/spec-default.properties
```diff
@@ -110,7 +110,7 @@ num.opFinisher.threads=8
...
...TFORM_NOT_ACTIVE,TRANSACTION_EXPIRED,TRANSACTION_HAS_UNKNOWN_FIELDS,TRANSACTION_ID_FIELD_NOT_ALLOWED,TRANSACTION_OVERSIZE,TRANSFER_ACCOUNT_SAME_AS_DELETE_ACCOUNT,EMPTY_ALLOWANCES,REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT,TOKEN_HAS_NO_FREEZE_KEY,TOKEN_HAS_NO_SUPPLY_KEY,INVALID_TOKEN_INITIAL_SUPPLY,INVALID_TOKEN_DECIMALS,INVALID_TOKEN_MAX_SUPPLY,ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS,TRANSFERS_NOT_ZERO_SUM_FOR_TOKEN,INVALID_ACCOUNT_AMOUNTS,TOKEN_NAME_TOO_LONG,TOKEN_SYMBOL_TOO_LONG,INVALID_TOKEN_NFT_SERIAL_NUMBER,PERMANENT_REMOVAL_REQUIRES_SYSTEM_INITIATION,MISSING_TOKEN_SYMBOL,MISSING_TOKEN_NAME,INVALID_EXPIRATION_TIME,EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS,INVALID_ALLOWANCE_OWNER_ID,FUNGIBLE_TOKEN_IN_NFT_ALLOWANCES,TOKEN_NOT_ASSOCIATED_TO_ACCOUNT,MAX_ALLOWANCES_EXCEEDED,INVALID_ALLOWANCE_SPENDER_ID,AMOUNT_EXCEEDS_TOKEN_MAX_SUPPLY,NFT_IN_FUNGIBLE_TOKEN_ALLOWANCES,NEGATIVE_ALLOWANCE_AMOUNT,DELEGATING_SPENDER_DOES_NOT_HAVE_APPROVE_FOR_ALL,DELEGATING_SPENDER_CANNOT_GRANT_APPROVE_FOR_ALL,INVALID_TOKEN_MINT_AMOUNT,INVALID_TOKEN_BURN_AMOUNT,INVALID_WIPING_AMOUNT,INVALID_NFT_ID,BATCH_SIZE_LIMIT_EXCEEDED,METADATA_TOO_LONG,INVALID_RENEWAL_PERIOD,INVALID_CUSTOM_FEE_SCHEDULE_KEY
```
jsync-swirlds, line 113: What does this property actually do?

georgi-l95: So I've added `INVALID_CUSTOM_FEE_SCHEDULE_KEY` to it and as far as I understand it, those are the ResponseCodes that we check in the HAPI Test Engine. If the response code is not here, it resolves to OK and that's what was causing the issue with this test (and the missing validation).

jsync-swirlds (dismissed):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -110,6 +112,9 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         if (op.hasAdminKey()) {
             context.requireKey(op.adminKeyOrThrow());
         }
+        if (op.hasFeeScheduleKey()) {
+            validateTruePreCheck(isValid(op.feeScheduleKey()), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
```
netopyr, line 116: @Neeharika-Sompalli, @mhess-swl: Is this the right place for such a check? Shouldn't this be checked in handle?

georgi-l95: btw this was my thought as well, but moving it to handle and adjusting the test, breaks it in the mono

netopyr: Actually digging into it some more, it looks like `isValid()` does not rely on state, and therefore this check can be part of the `pureCheck()` method IMO.

georgi-l95: I can move it there, seems logical to me also, let's see what @Neeharika-Sompalli and/or @mhess-swl also think ?

Neeharika-Sompalli: @georgi-l95  yes it doesn't depend on state. It can be moved to pureCheck

Neeharika-Sompalli (approved):

Neeharika-Sompalli (approved):
LGTM ! Thanks @georgi-l95 !

jsync-swirlds (approved):

PR #9840, opened 2023-11-10
9839 Prevented possible NPE caused by a race condition

### LINKED ISSUE #9839: `DataFileCollectionTest#merge` fails non-deterministially
### Description

On some rare occasions `DataFileCollectionTest#merge` may fail due to a race condition between reading and compacting threads. 

Specifically, reading thread enters method `DataFileCollection#readDataItemBytes`, gets an instance of `DataFileReader` using method `DataFileCollection#readerForDataLocation` (this method checks if the underlying file channel is open) and then compacting thread closes this reader. The subsequent attempts to use this reader result in the following exception: 
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.nio.channels.FileChannel.read(java.nio.ByteBuffer, long)" because "fileChannel" is null	
at com.swirlds.merkledb.utilities.MerkleDbFileUtils.completelyRead(MerkleDbFileUtils.java:91)	
at com.swirlds.merkledb.files.DataFileReader.read(DataFileReader.java:383)	
at com.swirlds.merkledb.files.DataFileReader.readDataItemBytes(DataFileReader.java:194)	
at com.swirlds.merkledb.files.DataFileCollection.readDataItem(DataFileCollection.java:491)	
at com.swirlds.merkledb.files.DataFileCollection.retryReadUsingIndex(DataFileCollection.java:513)	
at com.swirlds.merkledb.files.DataFileCollection.readDataItemUsingIndex(DataFileCollection.java:598)	
at com.swirlds.merkledb.files.DataFileCollectionTest.lambda$merge$10(DataFileCollectionTest.java:403)	
```

To prevent that, we need to have a null check in `DataFileReader#read`. Also, callers of this method should be aware that the method can return null. `DataFileCollection#retryReadUsingIndex` in its turns is already capable of handling nulls - it will retry on it, and that will help. 

### Steps to reproduce

1. Run DataFileCollectionTest
2. Hope for the best

### Additional context

_No response_

### Hedera network

other

### Version

v0.45

### Operating system

None

imalygin:
**Description**:

This change prevents possible NPE described in #9839 Instead, the method returns null and other methods pass that null to the point where it's retried on (see `DataFileCollection#retryReadUsingIndex`)

**Related issue(s)**:

Fixes #9839 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReader.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReader.java
@@ -186,6 +186,9 @@ public final class DataFileReader<D> implements AutoCloseable, Comparable<DataFi
         if (dataItemSerializer.isVariableSize()) {
             // read header to get size
             final ByteBuffer serializedHeader = read(byteOffset, dataItemSerializer.getHeaderSize());
+            if (serializedHeader == null) {
+                return null;
+            }
             final DataItemHeader header = dataItemSerializer.deserializeHeader(serializedHeader);
             bytesToRead = header.getSizeBytes();
         } else {
@@ -376,6 +379,11 @@ public final class DataFileReader<D> implements AutoCloseable, Comparable<DataFi
         for (int retries = 3; retries > 0; retries--) {
             final int fcIndex = leaseFileChannel();
             final FileChannel fileChannel = fileChannels.get(fcIndex);
+            if (fileChannel == null) {
+                // On rare occasions, if we have a race condition with compaction, the file channel
+                // may be closed. We need to return null, so that the caller can retry with a new reader
+                return null;
+            }
             try {
                 buffer.position(0);
                 buffer.limit(bytesToRead);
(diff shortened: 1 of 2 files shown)


artemananiev (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReader.java
```diff
@@ -376,6 +379,11 @@ private ByteBuffer read(final long byteOffsetInFile, final int bytesToRead) thro
         for (int retries = 3; retries > 0; retries--) {
             final int fcIndex = leaseFileChannel();
             final FileChannel fileChannel = fileChannels.get(fcIndex);
+            if (fileChannel == null) {
```
jsync-swirlds, line 382: Do we need to check if the file channel is actually open here (e.g. ` || !filChannel.isOpen()`)?

jsync-swirlds (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReader.java
imalygin: I don't think we need this condition. If this code even encounters a closed channel, it processes the exception here:
```
            } catch (final ClosedChannelException e) {
                // This exception may be thrown, if the channel was closed, because a different
                // thread reading from the channel was interrupted. Re-create the file channel
                // and retry
                reopenFileChannel(fcIndex, fileChannel);
            } 
```

PR #9856, opened 2023-11-13
Throw if gsutil not installed

### LINKED ISSUE #9855: Throw error if gsutils not installed when running JTR
Alert user if gsutil is not installed, instead of hanging

alittley:
- closes #9855


lpetrovic05 (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/utils/JtrUtils.java
```diff
@@ -81,6 +81,8 @@ public static JrsReportData scrapeTestData(
                 new VirtualTerminal().setProgressIndicatorEnabled(true).setThrowOnError(true);
         final Instant now = Instant.now();
 
+        terminal.run("gsutil");
```
cody-littley, line 84: Can you add a comment saying why we are running this? Obvious now, may not be obvious later.

cody-littley (approved):

PR #9870, opened 2023-11-13
HIP-796 `HapiSpec`'s

### LINKED ISSUE #9648: Create EETs for all HIP-796 user stories
### Problem

We need (initially failing) EETs for all the user stories in [HIP-796](https://hips.hedera.com/hip/hip-796).

### Solution

Express these EETs these `HapiSpecOperation`'s tailored to the abstractions used in HIP-796.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #9648
 - Adds `HapiSpec`'s for every consensus software user story in [HIP-796: "Lockable Fractional Amounts of Fungible Tokens on Hedera"](https://hips.hedera.com/hip/hip-796).
     * Introduces a family of `HapiSpecOperation` factories in the `Hip796Verbs` file that raise the level of abstraction used in the DSL for the included `HapiSpec`'s.
     * Reduces the line count needed to specify a token by adding a [`TokenFeature` enum](https://github.com/hashgraph/hedera-services/blob/hip796-test-dsl/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip796/operations/TokenFeature.java#L20) and `tokenWithFeatures()` factories that accept multiple features.
- Should be followed by:
   * https://github.com/hashgraph/hedera-services/issues/9885
   * https://github.com/hashgraph/hedera-services/issues/9886
   * https://github.com/hashgraph/hedera-services/issues/9888


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/HapiTxnOp.java
```diff
@@ -577,6 +578,10 @@ public T feeUsd(double price) {
         return self();
     }
 
+    public T signedByPayerAnd(String... keys) {
+        return signedBy(Stream.concat(Stream.of(DEFAULT_PAYER), Stream.of(keys)).toArray(String[]::new));
```
jsync-swirlds, line 582: We're creating three streams and an array just to add one item to an array?
This could be done with a single allocation and a single array copy.
```suggestion
        final String[] copy = new String[keys.length + 1];
        copy[0] = DEFAULT_PAYER;
        System.arrayCopy(keys, 0, copy, 1, keys.length);
        return signedBy(copy);
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip796/Hip796Verbs.java
```diff
@@ -0,0 +1,324 @@
...
+    // --- Lock management verbs ---
+    /**
+     * Locks the given amount of units of the given token in the given account.
+     *
+     * @param account the account to lock the units in
+     * @param token the token to lock the units of
+     * @param amount the amount of units to lock
+     * @return a {@link HapiTxnOp} that can be used to lock the units
+     */
+    public static HapiTxnOp lockUnits(@NonNull final String account, @NonNull final String token, final long amount) {
```
Neeharika-Sompalli, line 145: I like the format of keeping everything under `hip796` folder 💯

Neeharika-Sompalli (approved):
LGTM ! Thanks @tinker-michaelj  🙌 . We can definitely keep improving this as needed while implementation.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/HapiTxnOp.java
tinker-michaelj: Changed as suggested.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip796/Hip796Verbs.java
tinker-michaelj: Thx @Neeharika-Sompalli 💯

PR #9871, opened 2023-11-13
Ensure all non-daemon threads are shutdown even if `exec()` throws 

tinker-michaelj:
**Description**:
 - It is possible that recent "hanging" runs of `recordStreamValidation()` are due to an intermittent failure in a validator that itself runs `HapiSpec`'s (e.g. the `TokenReconciliationValidator`) throwing an exception instead of returning a failed assertion from `HapiSpecOperation.exec()`.
 - This would prevent the finalizer `ExecutorService` from being shut down, and lead to the test hanging forever.


Neeharika-Sompalli (approved):
LGTM !

Issue #9766, opened 2023-11-08
Follow up for Support "isolated" flag on HapiTestSuite


vtronkov:
There were a couple of minor things to fix in https://github.com/hashgraph/hedera-services/pull/8733 but we preferred to merge it and create a follow-up story.

In this story, we will address the following comments:
- [x] https://github.com/hashgraph/hedera-services/pull/8733#discussion_r1334947358
- [x] https://github.com/hashgraph/hedera-services/pull/8733#discussion_r1334966987
- [x] https://github.com/hashgraph/hedera-services/pull/8733#discussion_r1334995163
- [x] https://github.com/hashgraph/hedera-services/pull/8733#discussion_r1386581389

PR #9873, opened 2023-11-14
chore: Follow up for Support "isolated" flag on HapiTestSuite.

vtronkov:
**Description**:
Addressing comments from https://github.com/hashgraph/hedera-services/pull/8733

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9766


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
```diff
@@ -170,9 +170,7 @@ public void waitForActive(long seconds) throws TimeoutException {
     @Override
```
vtronkov, line 170: @tinker-michaelj Regarding https://github.com/hashgraph/hedera-services/pull/8733#discussion_r1334995163 - debugging it looks like the `th.hedera` is null in the first couple of iterations. I'm guessing that there is some time needed for it to be initialized.
Extracting the `th != null && t.hedera != null` out of the loop breaks the `waitForActive` and we are timing out when we try to start the node

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -503,27 +503,14 @@ public void init(@NonNull final Platform platform, @NonNull final NodeId nodeId)
...
-                        if (wasActive) shutdownGrpcServer();
-                    }
-                    case OBSERVING -> logger.info("Hederanode#{} is OBSERVING", nodeId);
-                    case FREEZING -> {
-                        logger.info("Hederanode#{} is FREEZING", nodeId);
+
+                    case FREEZE_COMPLETE, REPLAYING_EVENTS, STARTING_UP, OBSERVING, RECONNECT_COMPLETE -> logger.info(
+                            "Hederanode#{} is {}", nodeId, platformStatus.name());
+
+                    case CATASTROPHIC_FAILURE, CHECKING, FREEZING, BEHIND -> {
```
jsync-swirlds, line 510: I don't think we should be shutting down the server if it's in `CHECKING` state, as that state is recoverable and transient.  Even `BEHIND` can recover (the node will reconnect).
If we do not shut down when `FREEZING` then we can probably re-enable some tests that send a `FREEZE` transaction (followed by cancelling the freeze).

Thoughts?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -66,17 +67,18 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster, final
...
+            for (int nodeId = 0; nodeId < nodesSize; nodeId++) {
                 final Path workingDir =
                         Path.of("./build/hapi-test/node" + nodeId).normalize();
                 setupWorkingDirectory(workingDir, configText);
                 final String nodeName = NODE_NAMES[nodeId];
                 final AccountID acct =
                         AccountID.newBuilder().accountNum(3L + nodeId).build();
-                if (useInProcessAlice && nodeId == 0) {
+                boolean currentNodeAlice = nodeId == 0;
+                if (IN_PROCESS_ALICE.equals(nodesType) && currentNodeAlice) {
```
jsync-swirlds, line 81: Enums should be compared with `==`.
```suggestion
                if (IN_PROCESS_ALICE == nodesType && currentNodeAlice) {
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEnv.java
```diff
@@ -88,6 +90,13 @@ public HapiTestEnv(@NonNull final String testName, final boolean cluster, final
         }
     }
 
+    // Defines the types of nodes(InProcessHapiTestNode or SubProcessHapiTestNode) for the test executing nodes.
+    enum HapiTestNodesType {
+        // Makes the fist starting node(Alice, id=0) to use InProcessHapiTestNode. This gives us the ability to debug
+        IN_PROCESS_ALICE,
+        UNDEFINED
```
jsync-swirlds, line 97: This should probably be `OUT_OF_PROCESS_ALICE` for symmetry.  It certainly isn't `UNDEFINED`; it is defined, just not in-process.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
vtronkov: Makes sense to me but this is way out of my depth. cc: @rbair23

jsync-swirlds (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
netopyr: That is concerning. If `th` and probably the content of `th` is changed on a different thread, we must ensure it is thread-safe. As a minimum, `th` must be volatile. I am unsure if this is sufficient. Can the start() and shutdown() methods be called from different threads?

vtronkov: I think this shouldn't happened but I'm not 100% sure. @rbair23 any thoughts?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
tinker-michaelj: Hmm Neeha at least when `hedera.profiles.active=PROD`, I can't see a reason to keep the gRPC server open on either `FREEZING` or `FREEZE_COMPLETE`. 

Maybe when `hedera.profiles.active=DEV`, there is some reason not to shut down gRPC server as @jsync-swirlds mentioned?

Don't have strong intuition here. 🤔

jsync-swirlds: Currently even dev and test run with the PROD profile.
When `FREEZING` the node probably should still stay active until `FREEZE_COMPLETE`; mostly because it could be cancelled before the freeze is completed (we have existing tests that start a freeze, verify `FREEZING` state, then cancel the freeze), and shutting down the server before then generally breaks the ability to cancel the freeze process.

vtronkov: What is the verdict here, guys? When should we shut down the GRPC server and when should we just log a message?

jsync-swirlds: I believe we should only shut down the gRPC server on FREEZE_COMPLETE or CATASTROPHIC_FAILURE to preserve pre-existing expectations and behaviors.
We should await confirmation from @Neeharika-Sompalli and @tinker-michaelj, however.

Neeharika-Sompalli: yeah I agree with @jsync-swirlds. That seem right to me.

tinker-michaelj: Same, seems reasonable to limit shutdown to just `FREEZE_COMPLETE` and `CATASTROPHIC_FAILURE`.

vtronkov: Updated 👍 Thank you, guys!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/InProcessHapiTestNode.java
jsync-swirlds: Currently, the in-process node runs on the HapiTest junit test runner thread (and only when debugging locally, it is never used in CI); nothing else will try to start or shut down this node.  Unless we start _debugging_ Hapi Test specs in parallel, we're pretty safe with a single-thread assumption here.

tinker-michaelj (approved):
LGTM

PR #9880, opened 2023-11-14
Split HAPI tests and run them concurrently in CI

### LINKED ISSUE #9876: Split HAPI tests and run them concurrently
The HAPI tests take way too long to run in CI. Using the `@Tag`, they should be divided into 3-4 groups. The build script needs to be updated to run these groups concurrently. The goal should be that each of these groups takes about 30 minutes to finish.

netopyr:
This PR splits the HAPI tests into several groups (taking advantage of the `@Tag` annotation). The build scripts were updated to run these groups parallel in CI.

For convenience, there is still a Gradle task to run all tests, though it is not used in CI.

Closes #9876

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
@@ -216,7 +216,6 @@ public class AutoAccountCreationSuite extends HapiSuite {
     }
 
     @HapiTest
-    @Tag("crypto")
     private HapiSpec canAutoCreateWithHbarAndTokenTransfers() {
         final var initialTokenSupply = 1000;
         return defaultHapiSpec("canAutoCreateWithHbarAndTokenTransfers", EXPECT_STREAMLINED_INGEST_RECORDS)
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
@@ -18,7 +18,7 @@ package com.hedera.services.bdd.suites.crypto;
 
 import static com.google.protobuf.ByteString.copyFromUtf8;
 import static com.hedera.node.app.service.evm.utils.EthSigsUtils.recoverAddressFromPubKey;
-import static com.hedera.services.bdd.junit.TestTags.CRYPTO;
+import static com.hedera.services.bdd.junit.TestTags.TOKEN;
 import static com.hedera.services.bdd.spec.HapiPropertySource.asSolidityAddress;
 import static com.hedera.services.bdd.spec.HapiSpec.defaultHapiSpec;
 import static com.hedera.services.bdd.spec.PropertySource.asAccount;
@@ -118,7 +118,7 @@ import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Tag;
 
 @HapiTestSuite
-@Tag(CRYPTO)
+@Tag(TOKEN)
 public class AutoAccountCreationSuite extends HapiSuite {
 
     private static final Logger LOG = LogManager.getLogger(AutoAccountCreationSuite.class);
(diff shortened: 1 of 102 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
```diff
@@ -213,6 +216,7 @@ public List<HapiSpec> getSpecsInSuite() {
     }
 
     @HapiTest
+    @Tag("crypto")
```
vtronkov, line 219: I think we should delete this tag since we've tagged the whole class

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
```diff
@@ -184,7 +184,8 @@ public List<HapiSpec> getSpecsInSuite() {
 
     @HapiTest
     private HapiSpec associateHasExpectedSemantics() {
-        return defaultHapiSpec("AssociateHasExpectedSemantics")
+        return propertyPreservingHapiSpec("AssociateHasExpectedSemantics")
+                .preserving("tokens.maxRelsPerInfoQuery")
```
vtronkov, line 188: Why do we need to change this line?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoGetInfoRegression.java
```diff
@@ -94,7 +98,8 @@ private HapiSpec fetchesOnlyALimitedTokenAssociations() {
         final var token6 = "token6";
         final var token7 = "token7";
         final var token8 = "token8";
-        return defaultHapiSpec("FetchesOnlyALimitedTokenAssociations")
+        return propertyPreservingHapiSpec("FetchesOnlyALimitedTokenAssociations")
+                .preserving("tokens.maxRelsPerInfoQuery")
```
vtronkov, line 102: Why do we need to change this line?

vtronkov (dismissed):
Total duration - **48m 17s**! In contrast with **1h 30m+** that's a pretty impressive optimization. Multiply it by all the builds we are running everyday 🤯 Fantastic job!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
netopyr: Good catch! Leftover from an early version. Will remove it.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
netopyr: This test changes the property `tokens.maxRelsPerInfoQuery`. We need to revert such changes. Otherwise, other tests can fail.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoGetInfoRegression.java
netopyr: Same here. This test "resets" the value of `tokens.maxRelsPerInfoQuery` to 3. We did not notice because `FileUpdateSuite` set the value to 1000 again. Once `FileUpdateSuite` was moved to another group, a third test that required a maximum of 4 failed. 

The clean solution is to use `propertyPreservingHapiSpec` from my understanding, because it really resets the property to its original value.

nathanklick (changes requested):
These changes will create a conflict with the higher priority #10034 pull request. Requesting that these be held to after the other PR goes into the `develop` branch tomorrow. 

Additionally, before this merges we need to look at the current runner utilization and ensure we have sufficient runners available to handle the increased demand due to the concurrent jobs introduced.

nathanklick (approved):

vtronkov (approved):

Neeharika-Sompalli (approved):
LGTM!

PR #9887, opened 2023-11-14
fix: Resolve HelloWorldEthereumSuite errors

### LINKED ISSUE #8675: Resolve `HelloWorldEthereumSuite` errors
### Problem

We have the following tests suite `HelloWorldEthereumSuite` that has problems when it's running against modularization codebase.

### Solution

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
- Enables HapiTests from `HelloWorldethEreumSuite`
- Resolved one issue in the test `badRelayClient` related to a wrong revert status in the created child record, but it is still failing. The problem is described in a separate [issue](https://github.com/hashgraph/hedera-services/issues/9958), so the test still is not enabled at this point.
- `topLevelBurnToZeroAddressReverts` and `internalBurnToZeroAddressReverts` ONLY fail when all smart contract HapiTests are run. When running them individually or the whole suite they are passing. I left them disabled for now since I was not able to debug this issue.

**Related issue(s)**:

Fixes #8675 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
```diff
@@ -77,12 +94,23 @@ public void handle(@NonNull final HandleContext context) throws HandleException
         // Create the transaction-scoped component
         final var component = provider.get().create(context, ETHEREUM_TRANSACTION);
 
+        // Assemble the appropriate top-level record for the result
```
tinker-michaelj, line 97: ⚠️ Hmm none of these changes should be necessary, since the `MessageCallProcessor` should (must!) refuse to lazy-create a long-zero address. 🤔

Is the only problem that you get e.g. `CONTRACT_REVERT_EXECUTED` instead of `INVALID_CONTRACT_ID` when enabling these `@HapiTest`s?

tinker-michaelj (changes requested):
I think we _probably_ want to just update the test client's expectations to maybe accept `CONTRACT_REVERT_EXECUTED` instead of `INVALID_CONTRACT_ID`.

If the changes in `EthereumTransactionHandler` are in fact necessary, they are hiding a catastrophic bug in the `CustomMessageCallProcessor` downstream (which should already refuse to transfer value to a non-existent long-zero address).

edward-swirldslabs (changes requested):
Please remove the commit with `chore!` from your commit history.

Issue #9883, opened 2023-11-14
UncheckedIO exception - no such pces file


alittley:
- Release 44 `FCM-Restart-2.5k-15m`: `FCM-Restart-2.5k-15m` failed: http://35.247.76.217:8095/swirlds-automation/release/0.44/4N/Restart/20231114-102342-GCP-Daily-Restart-4N/FCM-Restart-2.5k-15m/
- develop `FCM-MultiSBReconnect-3R-1k-20m`: http://35.247.76.217:8095/swirlds-automation/develop/10N/MultiSBReconnect/20231114-055744-GCP-Daily-MultiSBReconnect-3R-10N/FCM-MultiSBReconnect-3R-1k-20m/
- develop `FCM-Hotspot-5k-25m-3Maccts`: http://35.247.76.217:8095/swirlds-automation/develop/4N/Misc/20231114-061044-GCP-Daily-Hotspot-4N/FCM-Hotspot-5k-25m-3Maccts/

```
Exception when writing the signed state for round 628 to disk:
java.io.UncheckedIOException: java.nio.file.NoSuchFileException:
data/saved/preconsensus-events/1/2023/11/14/2023-11-14T10+54+59.070612160Z_seq72_ming2289_maxg2460_orgn0.pces
```

Presumably related to https://github.com/hashgraph/hedera-services/pull/9857

PR #9890, opened 2023-11-14
Fix race condition in PCES file copy.

cody-littley:
Related to #9883 

Needs to be cherry picked into 0.44

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
@@ -19,6 +19,7 @@ package com.swirlds.platform.state.signed;
 import static com.swirlds.common.io.utility.FileUtils.executeAndRename;
 import static com.swirlds.common.io.utility.FileUtils.writeAndFlush;
 import static com.swirlds.logging.legacy.LogMarker.EXCEPTION;
+import static com.swirlds.logging.legacy.LogMarker.STARTUP;
 import static com.swirlds.logging.legacy.LogMarker.STATE_TO_DISK;
 import static com.swirlds.platform.config.internal.PlatformConfigUtils.writeSettingsUsed;
 import static com.swirlds.platform.state.signed.SignedStateFileUtils.CURRENT_ADDRESS_BOOK_FILE_NAME;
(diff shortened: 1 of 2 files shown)


edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -166,14 +175,37 @@ public static void writeSignedStateFilesToDirectory(
...
+
+        int triesRemaining = COPY_PCES_MAX_RETRIES;
+        while (triesRemaining-- > 0) {
+            try {
+                copyPreconsensusEventStreamFiles(
+                        platformContext, selfId, destinationDirectory, minimumGenerationNonAncient);
+                return;
+            } catch (final PreconsensusEventFileRenamed e) {
+                logger.error(
+                        EXCEPTION.getMarker(),
```
alittley, line 201: A lot of tests will keep failing with this exception

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -345,15 +377,20 @@ private static void copyPreconsensusFileList(
                 "Copying {} preconsensus event files to state snapshot directory",
                 filesToCopy.size());
 
+        // The last file might be in the process of being written, so we need to do a deep copy of it.
+        // Unlike the other files we are going to copy which have a long lifespan and are expected to
+        // be stable, the last file is actively in flux. It's possible that the last file will be
+        // renamed by the time we get to it, which may cause this copy operation to fail. Attempt
+        // to copy this file first, so that if se wail we can abort and retry without other side
```
alittley, line 384: ~~se wail~~ *we fail*

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: Oops, the intention was to only log this if we run out of tries. I've refactored this so that it will only log when the number of remaining tries is 0.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
cody-littley: fixed

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateFileWriter.java
```diff
@@ -166,14 +175,40 @@ public static void writeSignedStateFilesToDirectory(
         writeSettingsUsed(directory, platformContext.getConfiguration());
 
         if (selfId != null) {
-            copyPreconsensusEventStreamFiles(
+            copyPreconsensusEventStreamFilesRetryOnFailure(
```
lpetrovic05, line 178: It seems that if one file fails, we will retry all of them. It seems better to retry on every file individually. There might be issues if we try to copy the same files again. I will approve, up to you if you think this change should be made.

lpetrovic05 (approved):

PR #9921, opened 2023-11-16
Implement pureChecks for file handler classes

### LINKED ISSUE #9893: implement pureChecks for File handlers

kimbor:
**Description**:
Implement pureChecks for file handler classes.

**Related issue(s)**:

Fixes #9893


hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/utils/FileServiceUtils.java
```diff
@@ -63,23 +63,20 @@ public static void validateContent(@NonNull byte[] content, @NonNull FilesConfig
...
-            @Nullable final FileID fileId,
+            @NonNull final FileID fileId,
             @NonNull final ReadableFileStore fileStore,
-            @NonNull final PreHandleContext context,
-            boolean isDelete)
+            @NonNull final PreHandleContext context)
             throws PreCheckException {
         requireNonNull(context);
-
-        if (fileId == null) {
```
jsync-swirlds, line 80: I think this specific check was more correct as it was before, a NPE makes dispatch less reliable.

jsync-swirlds (dismissed):
Looks good, with one suggestion.

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/utils/FileServiceUtils.java
kimbor: We are now checking that fileId != null in pureChecks, before we ever get here. It should not ever happen that fileId is null here.

jsync-swirlds: Except that some paths do not call pureChecks (e.g. when a contract or scheduled transaction is gathering keys to be validated).

jsync-swirlds: True, but can we guarantee that will be true indefinitely?  It's not a major concern, I just believe that avoiding NPE, when possible, is preferred.

povolev15 (dismissed):
LGTM, maybe just Joseph suggestion but it is up to you

iwsimon (approved):
LGTM

jsync-swirlds (approved):

iwsimon (approved):

PR #9924, opened 2023-11-16
System files reference

netopyr:
Added a simple table in the documentation that shows the content type of each system file and where it is processed. It is tedious to look this up in code each time.


povolev15 (approved):
LGTM

Issue #9926, opened 2023-11-16
Support @Tag annotation on a class level


vtronkov:
We want to support a class-level @Tag annotation. The current implementation(https://github.com/hashgraph/hedera-services/issues/9436) only supports methods

PR #9927, opened 2023-11-16
Support @Tag annotation on a class level

vtronkov:
**Description**:
Add @Tag support on class-level

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9926


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -409,4 +399,39 @@ public Set<TestTag> getTags() {
             return this.testTags;
         }
     }
+
+    private static Set<TestTag> getTagsIfAny(Class<?> testMethod) {
```
netopyr, line 403: Nit: the parameter should be renamed

netopyr (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
vtronkov: Oops, copy-paste error. Fixed 👍

netopyr (approved):
LGTM

Issue #9928, opened 2023-11-16
Fix ContractXTest's from #9769


MiroslavGatsanoga:
The remaining three failures from https://github.com/hashgraph/hedera-services/pull/9769 are:
      - `CreatesXTest`: Needs status customization for `INVALID_ACCOUNT_ID` -> `INVALID_TREASURY_FOR_TOKEN`
      - `ClassicViewsXTest` - `GET_TOKEN_INFO` wrong (likely precedes payer validation PR merge)
      - `SetApprovalForAllXTest` - Fails with `"given selector does not match: expected: 367605ca, found: a22cb465"` (likely precedes payer validation PR merge)

PR #9930, opened 2023-11-16
Fix remaining contract x tests

stoyanov-st:
**Description**:
Fixing two out of three X tests from https://github.com/hashgraph/hedera-services/pull/9769 

**Related issue(s)**:

Fixes #
#9928 
**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/xtest/java/contract/CreatesXTest.java
+++ b/hedera-node/hedera-app/src/xtest/java/contract/CreatesXTest.java
@@ -16,9 +16,9 @@
 
 package contract;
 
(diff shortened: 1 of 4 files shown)


tinker-michaelj (approved):
I think we may want to update the `TokenCreate` call to use `INVALID_TREASURY_ACCOUNT_FOR_TOKEN` via [this pattern](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java#L52). But this is a fine short-term change imo. 💯

PR #9931, opened 2023-11-16
#9427: Revert after INVALID_ACCOUNT_ID static call

### LINKED ISSUE #9427: Unconditional, enriched records of read-only HTS system contract calls.

ilko-iliev-lime:
**Description**:
The PR handles a missed case. When we make a static GetAllowance call and the accountId is invalid, then the transaction must be reverted. 

See snapshot [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/call-results-snapshot.json).

**Related issue(s)**:

Fixes #9427

**Notes for reviewer**:

**Checklist**


tinker-michaelj (approved):
LGTM!

PR #9946, opened 2023-11-16
9944 Fix record cache handling to remediate "RECORD_NOT_FOUND" responses.

### LINKED ISSUE #9944: Fix record cache handling to remediate "RECORD_NOT_FOUND" responses
Most ScheduleRecordSpec tests return RECORD_NOT_FOUND because the dispatch of the triggered transactions is producing child records rather than the expected top level records.  Likewise some of the values in the triggered transaction are not matching Mono Service expectations.
We need to make necessary changes so those records can be found via the get transaction info query.

jsync-swirlds:
* Fixed an arithmetic bug in HandleWorkflow where we added instead of subtracting an offset
 * Adjusted RecordCacheImpl to treat scheduled transactions as top-level istead of children of the sign or create.
 * Adjusted RecordListBuilter to only create a synthetic transactionID if none is provided.
    * Includes modifying AbstractGrantApprovalCall to _not_ send an unnecessary transactionID.
 * Modified the DeduplicationCacheImpl to consider the `scheduled` flag when finding duplicate transaction ID values

Note: There remains a bug in consensus timestamp for child transactions.  That will be fixed in a separate PR as it also enables quite a few HapiTest methods, which makes that PR larger and more complex to review.

Fixes #9944

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/DeduplicationCacheImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/DeduplicationCacheImpl.java
@@ -39,11 +39,16 @@ import javax.inject.Singleton;
 public final class DeduplicationCacheImpl implements DeduplicationCache {
     /**
      * The {@link TransactionID}s that this node has already submitted to the platform, sorted by transaction start
-     * time, such that earlier start times come first. We guard this data structure within a synchronized block.
+     * time, such that earlier start times come first.
+     * <p>
+     * Note that an ID with scheduled set is different from the same ID without scheduled set.
+     * In fact, an ID with scheduled set will always match the ID of the ScheduleCreate transaction that created
+     * the schedule, except scheduled is set.
      */
     private final Set<TransactionID> submittedTxns = new ConcurrentSkipListSet<>(
             (t1, t2) -> Comparator.comparing(TransactionID::transactionValidStartOrThrow, TIMESTAMP_COMPARATOR)
                     .thenComparing(TransactionID::accountID, ACCOUNT_ID_COMPARATOR)
+                    .thenComparing(TransactionID::scheduled)
                     .compare(t1, t2));
 
     /** Used for looking up the max transaction duration window. */
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
@@ -197,7 +197,6 @@ public class RecordCacheImpl implements HederaRecordCache {
         if (history == null) {
             return DuplicateCheckResult.NO_DUPLICATE;
         }
-
         return history.nodeIds().contains(nodeId) ? DuplicateCheckResult.SAME_NODE : DuplicateCheckResult.OTHER_NODE;
     }
 
@@ -239,8 +238,10 @@ public class RecordCacheImpl implements HederaRecordCache {
         history.nodeIds().add(nodeId);
 
         // Either we add this tx to the main records list if it is a user/preceding transaction, or to the child
-        // transactions list of its parent
-        final var listToAddTo = isChildTx ? history.childRecords() : history.records();
+        // transactions list of its parent.  Note that scheduled transactions are always child transactions, but
+        // never produce child *records*; instead, the scheduled transaction record is treated as
+        // a user transaction record.
+        final var listToAddTo = (isChildTx && !txId.scheduled()) ? history.childRecords() : history.records();
         listToAddTo.add(transactionRecord);
 
         // Add to the payer-to-transaction index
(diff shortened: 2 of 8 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -376,25 +377,26 @@ public Result build() {
...
         int nextNonce = count + 1; // Initialize to be 1 more than the number of preceding items
         count = childRecordBuilders == null ? 0 : childRecordBuilders.size();
         for (int i = 0; i < count; i++) {
-            final var recordBuilder = childRecordBuilders.get(i);
-            records.add(recordBuilder
-                    .transactionID(idBuilder.nonce(nextNonce++).build())
-                    .syncBodyIdFromRecordId()
-                    .build());
+            final SingleTransactionRecordBuilderImpl recordBuilder = childRecordBuilders.get(i);
+            // Only create a new transaction ID for child records is one is not provided
```
mhess-swl, line 392: Not a blocker
```suggestion
            // Only create a new transaction ID for child records if one is not provided
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -257,7 +257,7 @@ private void handlePlatformTransaction(
             @NonNull final ConsensusTransaction platformTxn) {
         // Get the consensus timestamp. FUTURE We want this to exactly match the consensus timestamp from the hashgraph,
         // but for compatibility with the current implementation, we adjust it as follows.
-        final Instant consensusNow = platformTxn.getConsensusTimestamp().minusNanos(1000 + 3L);
+        final Instant consensusNow = platformTxn.getConsensusTimestamp().minusNanos(1000 - 3L);
```
mhess-swl, line 260: Are we sure this is correct? This puts the transaction time OUTSIDE of the consTime +/- 1000 nanos range. Why would we want that?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/DeduplicationCacheImpl.java
```diff
@@ -39,11 +39,14 @@
 public final class DeduplicationCacheImpl implements DeduplicationCache {
     /**
      * The {@link TransactionID}s that this node has already submitted to the platform, sorted by transaction start
-     * time, such that earlier start times come first. We guard this data structure within a synchronized block.
+     * time, such that earlier start times come first. An ID with scheduled set is different from the same ID
+     * without scheduled set (and, in fact, an ID with scheduled set will always match the ID of the ScheduleCreate
+     * transaction that created the schedule, except scheduled is set).
```
mhess-swl, line 44: Just a thought
```suggestion
     * time, such that earlier start times come first. 
     *
     * <p>Note that an ID with scheduled set is different from the same ID
     * without scheduled set (and, in fact, an ID with scheduled set will always match the ID of the ScheduleCreate
     * transaction that created the schedule, except scheduled is set).
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -239,8 +238,9 @@ private void addToInMemoryCache(
         history.nodeIds().add(nodeId);
 
         // Either we add this tx to the main records list if it is a user/preceding transaction, or to the child
-        // transactions list of its parent
-        final var listToAddTo = isChildTx ? history.childRecords() : history.records();
+        // transactions list of its parent.  Note that scheduled transactions are always child transactions, but
+        // never produce child *records*, the scheduled transaction record is treated as a user transaction record.
```
mhess-swl, line 242: ...therefore?
```suggestion
        // never produce child *records*, and therefore the scheduled transaction record is treated as a user transaction record.
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
jsync-swirlds: Actually, this fixes to put the time back in.
Subtraction (1000 + 3) means subtract 1003, which is outside.  This fixes the code to subtract 997, as originally intended.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: it is more correct to say `instead` in this case.  Cause and effect are inverse from what therefore implies.
I'll make the corresponding change.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
mhess-swl: Ahhh I didn't see the double negative, got it 👍

jsync-swirlds: It's confusing, which is probably why it was incorrect in the first place.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
mhess-swl: That's fine, I wasn't sure what the connecting phrase should be. Just that it wasn't there and my brain short-circuited

mhess-swl (approved):

tinker-michaelj (approved):
LGTM!

PR #9947, opened 2023-11-16
Fixed HollowAccountCompletionFuzzing

### LINKED ISSUE #9505: Fix HollowAccountCompletionFuzzing
- [x] `hollowAccountCompletionFuzzing()`

Hollow accounts are not correctly implemented yet. Probably makes sense to wait until [HollowAccountFinalizationSuite](https://github.com/hashgraph/hedera-services/issues/9362) passes.

iwsimon:
**Description**:

**Related issue(s)**:

Fixes #9505 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/SolvencyPreCheck.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/SolvencyPreCheck.java
@@ -83,10 +83,14 @@ public class SolvencyPreCheck {
      *
      * @param storeFactory the {@link ReadableStoreFactory} used to access readable state
      * @param accountID the {@link AccountID} of the payer
+     * @param ingestCheck if true, the check is being performed during an ingest workflow.
      * @throws PreCheckException if the payer account is invalid
      */
     @NonNull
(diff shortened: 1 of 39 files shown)


tinker-michaelj (changes requested):
I confirmed that if we just add `PAYER_ACCOUNT_DELETED` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/providers/ops/hollow/RandomOperationSignedBy.java#L44) a 120s run of this `@HapiTest` was stable.

I would prefer this instead of introducing extra cases to the node software here. 😞

iwsimon:
> I confirmed that if we just add `PAYER_ACCOUNT_DELETED` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/providers/ops/hollow/RandomOperationSignedBy.java#L44) a 120s run of this `@HapiTest` was stable (with no other changes needed anywhere).
> 
> I would prefer this instead of introducing extra cases to the node software here. 😞

It makes sense to modify the test-clients code to include more precheck scenarios than node code to check where the code is executed throw different PreCheckException. With this change though, we still had FAIL_INVALID errors. It's fixed by the change in HandleWorkflow.

PR #9950, opened 2023-11-17
fix(evm-equiv): Overflow when computing gas cost didn't return correct response code

### LINKED ISSUE #9917: Evm equivalence: In the case of numeric overflow when computing `gasPrice*gasLimit` we don't respond the way Ethereum does
Ethereum rejects the transaction.  We should do the same and also either match Ethereum's behavior w.r.t. to charging for gas or explain our choice.

Characterization on Ethereum also needed. (Confirm that they fail transaction with "no funds".)

david-bakin-sl:
**Description**:

Numeric overflow when computing gas cost (`limit * price`) didn't return correct response code, per EVM equivalence.
* Should return some form of `INSUFFICIENT_{FEE,PAYER_BALANCE,GAS}` but instead transaction simply failed (as the overflow wasn't handled) with a nonspecific code.

Now returns `INSUFFICIENT_PAYER_BALANCE`.

**Related issue(s)**:

Fixes #9917

**Notes for reviewer**:

1. Currently in DRAFT because only has unit tests, no Hapi tests.
1. This fix needed for mono only, modular already handles this overflow (but addition of a HAPI test to check that is important).

**Checklist**

- [½] Tested (unit only, integration test coming)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/EvmTxProcessor.java
```diff
@@ -112,7 +112,13 @@ protected TransactionProcessingResult execute(
             final BigInteger userOfferedGasPrice,
             final long maxGasAllowanceInTinybars,
             final Account relayer) {
-        final Wei gasCost = Wei.of(Math.multiplyExact(gasLimit, gasPrice));
+
+        Wei gasCost;
+        {
+            final Wei wGasLimit = Wei.of(gasLimit);
+            final Wei wGasPrice = Wei.of(gasPrice);
```
shemnon, line 119: Wei has a long multiply, so only one needs to be converted to Wei.

hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/contracts/execution/CallLocalEvmTxProcessorTest.java
```diff
@@ -71,6 +73,9 @@
 @ExtendWith(MockitoExtension.class)
 class CallLocalEvmTxProcessorTest {
     private static final int MAX_STACK_SIZE = 1024;
+    public static final long ONE_HBAR = 100_000_000L;
```
lukelee-sl, line 76: nit: why do you need these constants?  will they be utilized later?

david-bakin-sl: TY. That was a copy-paste thing, needed for the mocks over at `CallEvmTxProcessorTest`, then some of those mocks weren't needed here because gas isn't even computed, and I forgot to clean up.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/EvmTxProcessor.java
david-bakin-sl: fixed.

lukelee-sl (dismissed):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
```diff
@@ -905,6 +907,46 @@ HapiSpec transferTokensViaEip2930TxSuccessfully() {
                                                         .withErcFungibleTransferStatus(true)))))));
     }
 
+    @HapiTest
+    HapiSpec overflowFromGasPriceTimesGasLimitFailsAppropriately() {
+        return defaultHapiSpec("overflowFromGasPriceTimesGasLimitFailsAppropriately")
+                .given(
+                        withOpContext((spec, ctxLog) -> spec.registry().saveContractId("invalid", asContract("1.1.1"))),
```
stoqnkpL, line 914: Do we need to be calling an invalid contract for this scenario? 
It's no big deal, but it's a bit confusing IMO because it mixes a couple different test scenarios. It would be more straightforward if we just attempt to deploy a small/empty contract or something similar, any operation that would be otherwise successful, except for the gas price * gas cost overflow.

stoqnkpL (commented):
Hmm, the PR description states:
> didn't return correct response code, per EVM equivalence

but then the EET asserts that we return status `INSUFFICIENT_ACCOUNT_BALANCE` which is a Hedera-specific status and not really the root-cause of the failure, right? The root cause is the overflow?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
david-bakin-sl: It seemed easier to do it this way since - with knowledge of what's going on inside - this check happens _before_ any contract is called (since it is validating fees are available).

But you're right, it is confusing, and I'll fix it.

david-bakin-sl:
> Hmm, the PR description states:
> 
> > didn't return correct response code, per EVM equivalence
> 
> but then the EET asserts that we return status `INSUFFICIENT_ACCOUNT_BALANCE` which is a Hedera-specific status and not really the root-cause of the failure, right? The root cause is the overflow?

Yes the root cause is an overflow.  The error returned in the Ethereum test suites is "no funds" - [see here for example](https://github.com/shemnon/besu/blob/532797fc1725d3264ca51d083cb2b8a89123f034/ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/HighGasPrice.json#L15) (the overflow example is at the bottom of this file, L167 & L165).  So this is the "expected" response (for some interpretation of "expected").  At any rate it isn't an "undefined" kind of "transaction failed" which is what was happening when the arithmetic overflow was eventually caught elsewhere higher up.

I'll fix up the PR comment (which will become the merge comment) to explain this better.

david-bakin-sl:
Mono is dead.  Need to reevaluate the issue and see if it is still happening in modular services.

PR #9955, opened 2023-11-17
Fix `whitelistpositivecase` test in LeakyContractTestsSuite

### LINKED ISSUE #9407: Fix whitelistPositiveCase test
We get 0 instead of expected 1 for tokenAllowancesCount for PretendPair contract

MrValioBg:
**Description**:
Fix `whitelistpositivecase` test in LeakyContractTestsSuite.
Add support for besu Address in hedera-config in order to enable permittedDelegateCallers and add check in FrameUtils for qualified delegate callers.

**Related issue(s)**:

Fixes #9407 

**Checklist**


hedera-node/hedera-config/src/main/java/module-info.java
```diff
@@ -12,7 +12,9 @@
     requires transitive com.hedera.pbj.runtime;
     requires transitive com.swirlds.config.api;
     requires transitive com.swirlds.config.extensions;
+    requires transitive org.hyperledger.besu.datatypes;
```
netopyr, line 15: I think we should try to avoid these dependencies in our config module.

MrValioBg: Hey, thanks for the review! Can you elaborate on why we should avoid this dependency, please?
 This is needed to support the EVM Address type in the `permittedDelegateCallers`.

netopyr: The bundling between our App and Besu needs to be loose. One never knows what the future brings, and we may be forced to create a version that uses a different VM or includes multiple VMs. To make this possible, we must ensure that as little code as possible depends on Besu and that it is only in a few well-defined areas. I believe so far, only the smart contract module depends on Besu.

I believe in this case, we have a few options:

1. We add the dependency. I would hate that, but this may be the best option if everything else is too much work right now.
2. We move `AddressTypeConverter` and `ContractsConfig` to the smart contracts module. Initially, the plan was to move all `Config`-classes to their specific service module anyway. But we were unsure because having all the configs in one place is also nice. But moving the configs to the code they configure seems cleaner and would allow the definition of specific types. Maybe this is a good time to start moving the `Config`-classes. 🤔
3. We create our own class that stores the value and add a converter to `Address` in the smart contract module.

I had a fourth idea that I forgot. 🙂 Maybe it comes back, and I will add it here.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameUtils.java
```diff
@@ -136,11 +136,15 @@ public static boolean unqualifiedDelegateDetected(final MessageFrame frame) {
...
-        // but we accept delegates if the token redirect contract calls us,
-        // so if they are not a token, then we are a delegate and we are done.
-        if (isToken(frame, recipient)) {
+        final var contractsConfig = configOf(frame).getConfigData(ContractsConfig.class);
+
+        // Evaluate whether the recipient is either a token or on the permitted callers list.
+        // This determines if we should treat this as a delegate call.
+        // We accept delegates if the token redirect contract calls us.
+        if (isToken(frame, recipient)
+                || contractsConfig.permittedDelegateCallers().contains(recipient)) {
```
tinker-michaelj, line 147: Per @netopyr's point, because the current mono-service value of this property here is **never** going to change, I don't see an issue with making `contractsConfig.permittedDelegateCallers()` return a `Set<Long>` and then doing,
```
if (isToken(frame, recipient) || (ConversionUtils.isLongZero(recipient) && 
    contractsConfig.permittedDelegateCallers().contains(
        ConversionUtils.numberOfLongZero(recipient))) {
```

hedera-node/hedera-config/src/main/java/module-info.java
MrValioBg: Thanks for the explanation. 🙂 I'll keep that in mind for future work regarding Besu and similar dependencies. I agree with @tinker-michaelj that implementing this using Set<Long> might be the easiest approach, so I went ahead and made the changes

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameUtils.java
MrValioBg: Yep, I got it, thank you! I did the implementation with `Set<Long> `

tinker-michaelj (approved):
LGTM, very nice work @MrValioBg !

PR #9821, opened 2023-11-10
signed state reserver

lpetrovic05:
closes #9819

PR #9959, opened 2023-11-17
signed state nexus

### LINKED ISSUE #9954: Create signed state nexus
Some signed states need to be kept in thread safe containers to be requested at any point in time. 
Create a thread-safe container that will keep a signed state and manage its reservations.

lpetrovic05:
closes #9954
must be merged after #9821

@cody-littley I hacked the signed state reservation a bit, we should discuss this


cody-littley (commented):
My concern with the approach taken by this PR is that it's basically a more complicated implementation of `SignedStateReference.java`. I get that you are trying to be lockless. But in this scenario, I'm not sure the added complexity is justified.

lpetrovic05:
> My concern with the approach taken by this PR is that it's basically a more complicated implementation of `SignedStateReference.java`. I get that you are trying to be lockless. But in this scenario, I'm not sure the added complexity is justified.

As mentioned in the call, I will need this for prehandle. This is the current code that handles reservations for prehandle:
```
        State immutableState = latestImmutableState.get();
        while (!immutableState.tryReserve()) {
            immutableState = latestImmutableState.get();
        }
        transactionHandler.preHandle(event, immutableState.getSwirldState());
```

This is very dangerous, because if the `immutableState` is ever closed (for whatever reason), this while loop will never exit. So think if this new code as a replacement for bad code, instead of new code.

PR #9965, opened 2023-11-17
Fix CongestionPricingSuite

### LINKED ISSUE #9178: Fix CongestionPricingSuite tests

JivkoKelchev:
**Description**:

Take congestion multiplier into account when fees and gas are calculated.
* Inject `CongestionMultipliers` instance to `FeeManager`, in order to be used when `FeeCalculator` is created
* Change writable stores to readable ones in `EntityUtilizationMultiplier`
* Replace `isProduction` boolean property with a config check when an unchecked submission is detected 
* Add second test in to the suite to validate fee calculations (for CryptoTransfer)

**Related issue(s)**:

Fixes #9178

**Notes for reviewer**:

* Change writable stores to readable ones in `EntityUtilizationMultiplier`
I had problems with the state mutability, while trying to inject `CongestionMultipliers` and needed state into `FeeManager`. So I decided to change all stores in `EntityUtilizationMultiplier`. 
However, I'm not sure if this [change](https://github.com/hashgraph/hedera-services/pull/9965/files#diff-cf84ada12eaa3d255f88e226addb9d1fdcb930af768ff44cdf84e758e8e94cdcL118-R176) will affect the calculations.

* Replace `isProduction` boolean property with a config check when an unchecked submission is detected 
I had trouble running Hapi tests with "DEV" active profile. When we start nodes from HapiTestEngine, ConfigProviders can't resolve relative paths to properties files, because working dir is `hedera-node/test-clients`. 
I fixed this locally by setting an environment variable `HEDERA_APP_PROPERTIES_PATH=/Users/zhivkokelchev/Projects/hedera-services/hedera-node/configuration/dev/application.properties`. But unfortunately in CI we still have the same problem. 
So I tried to override the active profile property and it didn't work, because `isProduction` was set in `SubmissionManager` constructor. I think that this [check](https://github.com/hashgraph/hedera-services/pull/9965/files#diff-084cf8913fc50efa341112c426f21f62129ab9ecf5810e0bdaa73ae9b6cfb578R134) won't hurt performance much, because it is applied only for unchecked submissions.

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/FeeManager.java
```diff
@@ -193,7 +206,12 @@ public FeeCalculator createFeeCalculator(
         final var feeData = getFeeData(functionality, consensusTime, SubType.DEFAULT);
 
         // Create the fee calculator
-        return new FeeCalculatorImpl(feeData, exchangeRateManager.activeRate(consensusTime));
+        return new FeeCalculatorImpl(
+                feeData,
+                exchangeRateManager.activeRate(consensusTime),
+                congestionMultipliers,
+                this.state,
```
netopyr, line 213: I am pretty sure this is not correct. If the state is required, we probably need to add it as a parameter to `createFeeCalculator()`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/SubmissionManager.java
```diff
@@ -130,7 +129,9 @@ public void submit(@NonNull final TransactionBody txBody, @NonNull final Bytes t
         // FUTURE This should be deprecated and removed. We do not want this in our production system.
         if (txBody.hasUncheckedSubmit()) {
             // We do NOT allow this call in production!
-            if (isProduction) {
+            // check profile dynamically, this way we allow profile overriding in Hapi tests
+            final var hederaConfig = configProvider.getConfiguration().getConfigData(HederaConfig.class);
+            if (Profile.valueOf(hederaConfig.activeProfile()) == Profile.PROD) {
```
netopyr, line 134: Interesting. I think `HederaConfig.activeProfile()` should return a `Profile` directly. Wonder why it returns a `String`. 🤔

netopyr (changes requested):
The handling of `state` looks suspicious and needs to be double-checked.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/FeeManager.java
JivkoKelchev: The sate is removed from `FeeManger`. As you suggested, now it is passed to `createFeeCalculatore()` as `ReadableStoreFactory`. Also `CongestionMultipliers` and `EntityUtilizationMultiplier` were refactored. Now they use `ReadableStoreFactory` instead of state.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/SubmissionManager.java
JivkoKelchev: I'm not sure why it was done like this. 
However I made it to work with `Profile`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/ChildFeeContextImpl.java
```diff
@@ -72,8 +74,9 @@ public Instant currentTime() {
     @Override
     public @NonNull FeeCalculator feeCalculator(@NonNull final SubType subType) {
         try {
+            var storeFactory = new ReadableStoreFactory((SavepointStackImpl) context.savepointStack());
```
netopyr, line 77: Do you think we can avoid casting to an implementation class here? How about passing the whole `context` to `createFeeCalculator()`. As an additional benefit, it would help to shorten the parameter list.

JivkoKelchev: Yes, I changed the `SavePointStackImpl` to `HederaState` interface.

In this particular case, we can replace consensus time and state with the whole context (HandleContext)
But unfortunately, this method is called also in `FeeContextImpl.feeCalculator()`, here we don’t have access to `HandleContext`. 
I also checked the classes that initialize `FeeContextImpl` - `IngestChecker` and `QueryChecker` and I don’t see a way to pass handle context from here either.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/config/ConfigProviderImpl.java
```diff
@@ -22,25 +22,7 @@
 import com.hedera.node.config.ConfigProvider;
 import com.hedera.node.config.VersionedConfigImpl;
 import com.hedera.node.config.VersionedConfiguration;
-import com.hedera.node.config.converter.AccountIDConverter;
```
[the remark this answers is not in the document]
JivkoKelchev: It was IntelliJ, I just disabled the wildcard imports.

tinker-michaelj (approved):
LGTM, ty @JivkoKelchev !

netopyr (approved):

PR #9969, opened 2023-11-17
feat: Remove address validity checks for modularized code base

### LINKED ISSUE #9459: Removal of address validity checks for top-level and internal EVM calls

lukelee-sl:
**Description**:

- Create a new 0.46 EVM module
- Add new feature flag to enable calling non-existing contract addresses without failing with `INVALID_SOLIDITY_ADDRESS` error
- Create additional code paths that do not error out on calls to non-existing contract addresses in TransactionProcessor
- Fix custom operations and handlers that are affected
- Modify unit tests

**Related issue(s)**:

Fixes #9459
See Also #9628 - pr for addressing this issue in mono service

**Checklist**


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
```diff
@@ -230,9 +230,13 @@ private ResponseCodeEnum validateSemantics(final TransactionBody transactionBody
         if (possiblySanityCheckOp(op, target)) {
             try {
                 final var receiver = accountStore.loadContract(target.toId());
-                validateTrue(receiver != null && receiver.isSmartContract(), INVALID_CONTRACT_ID);
+                if (!receiver.isSmartContract()) {
+                    return INVALID_CONTRACT_ID;
+                }
             } catch (InvalidTransactionException e) {
-                return e.getResponseCode();
+                // Continue silently as we wish to allow calls to deleted and non-existent contracts
```
tinker-michaelj, line 237: Is this meant to be in this mod-service PR?

lukelee-sl: No this should be removed and any modifications to this class should be with the parent.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FeatureFlags.java
```diff
@@ -65,4 +66,14 @@ default boolean isImplicitCreationEnabled(@NonNull MessageFrame frame) {
...
     boolean isImplicitCreationEnabled(@NonNull Configuration config);
+
+    /**
+     * If true calls to non-existing contract addresses will result in a successful NOOP.  If false,
+     * calls such calls will result in a revert with status {@code INVALID_SOLIDITY_ADDRESS}.
+     * @param config the {@link Configuration}
+     * @#param possiblyGrandFathered the account number to check for grandfathering
+     * @return true if calls to non-existing contract addresses will result in a successful NOOP.
+     */
+    boolean isAllowCallsToNonContractAccountsEnabled(
```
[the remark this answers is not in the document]
lukelee-sl: Yes, makes sense.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java
```diff
@@ -259,6 +252,68 @@ private InvolvedParties computeInvolvedParties(
         return parties;
     }
 
+    private boolean contractNotRequired(@NonNull final HederaEvmAccount to, @NonNull final Configuration config) {
```
tinker-michaelj, line 255: `to` should be annotated `@Nullable`, right?

Not sure if there are cases where it would matter, but I think we should also get the number of the account via,
```
final var maybeGrandfatheredNumber = (to == null) ? null : to.hederaId().accountNumOrThrow();

return featureFlags.isAllowCallsToNonContractAccountsEnabled(config, maybeGrandfatheredNumber);
```
since this would work even if the account had an EVM address outside the long-zero subspace.

lukelee-sl: This seems to fail in the case that the `to` address is a token address.

tinker-michaelj (approved):
LGTM, tyvm @lukelee-sl !

PR #9972, opened 2023-11-17
09971 simple timer api

### LINKED ISSUE #9971: Wiring framework: simpler API for timer input
Simplify the API for wiring a component to a timer. Instead of providing an output wire that provides a timer, instead build a bindable object and automatically do the wiring.

cody-littley:
Closes #9971


lpetrovic05 (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/StandardWiringModel.java
```diff
@@ -108,18 +108,30 @@ public final TaskSchedulerMetricsBuilder metricsBuilder() {
...
-     * {@inheritDoc}
+     * Build a wire that produces an instant (reflecting current time) at the specified rate. Note that the exact rate
+     * of heartbeats may vary. This is a best effort algorithm, and actual rates may vary depending on a variety of
+     * factors.
+     *
+     * @param period the period of the heartbeat. For example, setting a period of 100ms will cause the heartbeat to be
+     *               sent at 10 hertz. Note that time is measured at millisecond precision, and so periods less than 1ms
+     *               are not supported.
+     * @return the output wire
+     * @throws IllegalStateException if the heartbeat has already started
```
alittley, line 119: Will this actually throw if a heartbeat was already started? What if a given component wants multiple heartbeats of different frequency to do separate tasks, is that not supported?

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/StandardWiringModel.java
cody-littley: You are permitted to have multiple heartbeats (although I think this pattern is odd, and would probably be a code smell in most cases). What this should say is that this will throw if the model has already been started.

Javadoc is updated.

alittley (approved):

PR #9979, opened 2023-11-17
Fix inital set of ScheduleExecutionSpecs HapiTest methods.

### LINKED ISSUE #9048: ScheduleExecutionSpecs ( 0 passing)

jsync-swirlds:
* Modified the workflow to skip the correct configured number of nanoseconds before the first child transaction consensus time.
* Marked all tests in ScheduleExecutionSpecs with `@HapiTest` or `@todo('issue#')` and file issues for groups with similar failures
* Small code cleanups to move shared methods to ScheduleUtils instead of scattered across various test suites.

Fixes #9048


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -253,19 +253,21 @@ private SingleTransactionRecordBuilderImpl doAddChild(
...
         }
 
-        // The consensus timestamp of the first item in the child list is T+1, where T is the time of the user tx
-        final var parentConsensusTimestamp = userTxnRecordBuilder.consensusNow();
-        final var prevConsensusNow = childRecordBuilders.isEmpty()
-                ? userTxnRecordBuilder.consensusNow()
+        // The consensus timestamp of the first item in the child list is T+K (in nanoseconds),
+        // where T is the time of the user tx and K is the maximum number of "preceding" records
+        // defined for the current configuration.
+        final long maxPrecedingRecords = consensusConfig.handleMaxPrecedingRecords();
```
netopyr, line 265: Nit: we can use this variable already in the condition above

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -253,19 +253,21 @@ private SingleTransactionRecordBuilderImpl doAddChild(
...
+        final ConsensusConfig consensusConfig = configuration.getConfigData(ConsensusConfig.class);
         if (childCount >= consensusConfig.handleMaxFollowingRecords()) {
             throw new HandleException(ResponseCodeEnum.MAX_CHILD_RECORDS_EXCEEDED);
         }
 
-        // The consensus timestamp of the first item in the child list is T+1, where T is the time of the user tx
-        final var parentConsensusTimestamp = userTxnRecordBuilder.consensusNow();
-        final var prevConsensusNow = childRecordBuilders.isEmpty()
-                ? userTxnRecordBuilder.consensusNow()
+        // The consensus timestamp of the first item in the child list is T+K (in nanoseconds),
```
netopyr, line 262: This is not what was explained to me when we talked about timestamps and I do not comfortable approving it. I think @tinker-michaelj needs to review the code.

jsync-swirlds: Understood.  The tests require this offset, so mono clearly _does this_, but the change should still be reviewed most carefully.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
jsync-swirlds: If you mean line 258, that's a different value (following versus preceding).
Otherwise I'm a bit confused what condition is being referenced.

netopyr: Oh, you are right. Sorry, false alarm.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
jsync-swirlds: We discussed this.
It turns out that Scheduled child transactions are very much treated like new top level transactions, instead of like children, in most situations, including this one.  The Scheduled transaction is offset an extra [max preceding transactions] because the scheduled transaction may, itself, have additional preceding transactions, which would (in mono) end up with the same nanosecond as other transactions if the scheduled (triggered) transaction doesn't offset this extra amount.

We might not handle the "scheduled transaction triggers a preceding transaction" situation in modular code in the preceding transaction dispatch ideally at the moment, but we don't think it's necessary to support in the first modular release (it just requires yet more special case handling in the preceding transaction logic if we do decide to support that edge case, however).

kimbor (approved):
lgtm!

jsync-swirlds:
> lgtm!

Had to fix a unit test.  Please re-review.

tinker-michaelj (approved):
LGTM!

kimbor (approved):

PR #9980, opened 2023-11-19
Extend `FullResult` with dispatched `RecordBuilder` 

tinker-michaelj:
**Description**:
 - _TBD_


tinker-michaelj:
Will merge via https://github.com/hashgraph/hedera-services/pull/9795

PR #9997, opened 2023-11-20
Fix tests from LeakyContractTestsSuite

### LINKED ISSUE #9402: Fix autoAssociationSlotsAppearsInInfo test

MrValioBg:
Fix tests from LeakyContractTestsSuite

**Description**:

**Related issue(s)**:

Fixes #9402 #9408 

**Checklist**


tinker-michaelj (changes requested):
The `nonceInfo` feature flag is clean enough to implement that it seems totally reasonable.

But the token associations limit feature flag we will never re-enable. 😞 So continuing to enforce it and adding noticeable code to do so doesn't really make sense. 😞 

I think we should remove those changes and replace existing spec `autoAssociationSlotsAppearsInInfo()` with,
```
    @HapiTest
    private HapiSpec autoAssociationSlotsAppearsInInfo() {
        final int maxAutoAssociations = 100;
        final String CONTRACT = "Multipurpose";

        return propertyPreservingHapiSpec("autoAssociationSlotsAppearsInInfo")
                .preserving(CONTRACT_ALLOW_ASSOCIATIONS_PROPERTY)
                .given(overriding(
                        CONTRACT_ALLOW_ASSOCIATIONS_PROPERTY,
                        "true"))
                .when()
                .then(
                        newKeyNamed(ADMIN_KEY),
                        uploadInitCode(CONTRACT),
                        contractCreate(CONTRACT).adminKey(ADMIN_KEY).maxAutomaticTokenAssociations(maxAutoAssociations),
                        getContractInfo(CONTRACT)
                                .has(ContractInfoAsserts.contractWith().maxAutoAssociations(maxAutoAssociations))
                                .logged());
    }

```

MrValioBg:
> The `nonceInfo` feature flag is recent enough, and your implementation is very clean, so keeping that in mod-service seems totally reasonable.
> 
> But the token associations limit feature flag we will never re-enable. So continuing to enforce it and adding noticeable code to do so doesn't really make sense. 😞
> 
> I think we should remove those changes and replace existing spec `autoAssociationSlotsAppearsInInfo()` with,
> 
> ```
>     @HapiTest
>     private HapiSpec autoAssociationSlotsAppearsInInfo() {
>         final int maxAutoAssociations = 100;
>         final String CONTRACT = "Multipurpose";
> 
>         return propertyPreservingHapiSpec("autoAssociationSlotsAppearsInInfo")
>                 .preserving(CONTRACT_ALLOW_ASSOCIATIONS_PROPERTY)
>                 .given(overriding(
>                         CONTRACT_ALLOW_ASSOCIATIONS_PROPERTY,
>                         "true"))
>                 .when()
>                 .then(
>                         newKeyNamed(ADMIN_KEY),
>                         uploadInitCode(CONTRACT),
>                         contractCreate(CONTRACT).adminKey(ADMIN_KEY).maxAutomaticTokenAssociations(maxAutoAssociations),
>                         getContractInfo(CONTRACT)
>                                 .has(ContractInfoAsserts.contractWith().maxAutoAssociations(maxAutoAssociations))
>                                 .logged());
>     }
> ```

Understood! I Implemented the changes :)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/RootProxyWorldUpdater.java
```diff
@@ -46,7 +46,7 @@ public class RootProxyWorldUpdater extends ProxyWorldUpdater {
 
     private boolean committed = false;
     private List<ContractID> createdContractIds;
-    private List<ContractNonceInfo> updatedContractNonces;
+    private List<ContractNonceInfo> updatedContractNonces = new ArrayList<>();
```
tinker-michaelj, line 49: Could initialize to the `Collections.emptyList()` singleton here to avoid an extra object creation.

tinker-michaelj (approved):
LGTM, ty @mustafauzunn !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/RootProxyWorldUpdater.java
MrValioBg: Yes, thats true. Done. :)

tinker-michaelj (approved):
LGTM, tyvm @MrValioBg !

PR #10004, opened 2023-11-20
cherry picked nexus

### LINKED ISSUE #9954: Create signed state nexus

lpetrovic05:
closes #9954

original PR #9959
comment from original PR: https://github.com/hashgraph/hedera-services/pull/9959#issuecomment-1818485177

@cody-littley I hacked the signed state reservation a bit, we should discuss this

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
@@ -117,6 +117,27 @@ public class ReservedSignedState implements AutoCloseableNonThrowing {
     }
 
     /**
+     * Try to get another reservation on the signed state. If the signed state is not closed, then a new reservation
+     * will be returned. If the signed state is closed, then null will be returned.
+     *
+     * @param reason a short description of why this SignedState is being reserved. Each location where a SignedState is
+     *               reserved should attempt to use a unique reason, as this makes debugging reservation bugs easier.
+     * @return a new wrapper around the state that holds a new reservation, or null if the signed state is closed
+     */
+    public @Nullable ReservedSignedState tryGetAndReserve(@NonNull final String reason) {
+        if (signedState == null) {
+            return new ReservedSignedState();
+        }
+        if (!signedState.tryIncrementReservationCount(reason, reservationId)) {
+            return null;
+        }
+        final ReservedSignedState reservedSignedState = new ReservedSignedState(signedState, reason);
+        // the constructor will increment the reservation count again, so we need to decrement it here
+        signedState.decrementReservationCount(reason, reservationId);
+        return reservedSignedState;
+    }
+
+    /**
      * Get the signed state. Does not take any reservations on the signed state, this is the responsibility of the
      * caller. Do not keep a reference to this signed state outside the scope of this wrapper object without properly
      * reserving it.
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
@@ -86,7 +86,8 @@ public class ReservedSignedState implements AutoCloseableNonThrowing {
      *                    SignedState is reserved should attempt to use a unique reason, as this makes debugging
      *                    reservation bugs easier.
      */
-    static @NonNull ReservedSignedState createAndReserve(@NonNull final SignedState signedState, @NonNull final String reason){
+    static @NonNull ReservedSignedState createAndReserve(
+            @NonNull final SignedState signedState, @NonNull final String reason) {
         final ReservedSignedState reservedSignedState = new ReservedSignedState(signedState, reason);
         signedState.incrementReservationCount(reason, reservedSignedState.getReservationId());
         return reservedSignedState;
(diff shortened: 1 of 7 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
```diff
@@ -116,6 +116,27 @@ public boolean isNotNull() {
...
+    public @Nullable ReservedSignedState tryGetAndReserve(@NonNull final String reason) {
+        if (signedState == null) {
+            return new ReservedSignedState();
+        }
+        if (!signedState.tryIncrementReservationCount(reason, reservationId)) {
+            return null;
+        }
+        final ReservedSignedState reservedSignedState = new ReservedSignedState(signedState, reason);
+        // the constructor will increment the reservation count again, so we need to decrement it here
+        signedState.decrementReservationCount(reason, reservationId);
```
cody-littley, line 136: Is it possible to do this without an extra increment+decrement? It seems like we lose a lot of benefit from being lockless if we do two extra atomic operations.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateNexus.java
```diff
@@ -0,0 +1,104 @@
...
+            final ReservedSignedState newReservation = state.tryGetAndReserve(reason);
+            if (newReservation != null) {
+                return newReservation;
+            }
+            // if tryGetAndReserve returned null, then we should check if set() was called in the meantime
+            // if yes, then we should try again and reserve the new state
+        } while (state != currentState.get());
+        // this means we cannot reserve the state we are holding, this is probably an error, since we should hold a
+        // reservation on it
+        return null;
```
cody-littley, line 60: Would it be possible to refactor this so that we don't call `currentState.get()` twice each cycle around the loop? It should be possible to maintain this current sanity check logic with a single `get()` per cycle.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateNexus.java
```diff
@@ -0,0 +1,104 @@
...
+        setState(null);
+    }
+
+    /**
+     * Same as {@link #setState(ReservedSignedState)}
+     */
+    @Override
+    public void accept(@Nullable final ReservedSignedState reservedSignedState) {
+        setState(reservedSignedState);
+    }
```
cody-littley, line 103: Why do we need to implement `Consumer`?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
```diff
@@ -116,6 +116,27 @@ public boolean isNotNull() {
...
 
+    /**
+     * Try to get another reservation on the signed state. If the signed state is not closed, then a new reservation
+     * will be returned. If the signed state is closed, then null will be returned.
+     *
+     * @param reason a short description of why this SignedState is being reserved. Each location where a SignedState is
+     *               reserved should attempt to use a unique reason, as this makes debugging reservation bugs easier.
+     * @return a new wrapper around the state that holds a new reservation, or null if the signed state is closed
+     */
+    public @Nullable ReservedSignedState tryGetAndReserve(@NonNull final String reason) {
```
cody-littley, line 127: I'd suggest adding a short unit test to verify this function's behavior. (OK if it doesn't test race conditions, just that the reservation count is mutated as expected.)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedState.java
```diff
@@ -309,6 +309,19 @@ void incrementReservationCount(@NonNull final String reason, final long reservat
         reservations.reserve();
     }
 
+    /**
+     * Try to increment the reservation count.
+     */
+    boolean tryIncrementReservationCount(@NonNull final String reason, final long reservationId) {
```
cody-littley, line 315: I'd suggest adding a short unit test to verify this function's behavior. (OK if it doesn't test race conditions, just that the reservation count is mutated as expected.)

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateNexus.java
lpetrovic05: we don't, I just thought it would be convenient when wiring it up

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedStateNexus.java
lpetrovic05: It is possible, but it would make it less readable I think. And I think its not worth it for a very rare scenario where:
- the underlying state has changed while we were trying to reserve it
- AND the old state has been destroyed as well

The likelihood of that happening is very low I think. Your call, if you still think I should change it, I will.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
lpetrovic05: fixed. I mentioned this in the PR comment then forgot about it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/ReservedSignedState.java
lpetrovic05: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedState.java
lpetrovic05: all the tests that exist today verify this behavior through the ReservedSignedStateTests. I added a test there which tests this method as well.

alittley (approved):

PR #10038, opened 2023-11-20
fix: ContractCallLocalSuite Fix

### LINKED ISSUE #10309: `ContractCallLocalSuite.htsOwnershipCheckWorksWithAliasAddress()`

ilko-iliev-lime:
Description:
 - Closes #10309

Purpose of the PR is fixing the htsOwnershipCheckWorksWithAliasAddress in the ContractCallLocalSuite. It fixes some problems in the NftTokenInfoCall and covers some unhandled cases.

Related issue(s):

Fixes https://github.com/hashgraph/hedera-services/issues/8813
https://github.com/hashgraph/hedera-services/pull/9617

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/nfttokeninfo/NftTokenInfoCallTest.java
```diff
@@ -58,15 +45,17 @@ class NftTokenInfoCallTest extends HtsCallTestBase {
 
     @Test
     void returnsNftTokenInfoStatusForPresentToken() {
+        com.hedera.pbj.runtime.io.buffer.Bytes ledgerId = com.hedera.pbj.runtime.io.buffer.Bytes.fromHex(LEDGER_ID);
```
MrValioBg, line 48:
```suggestion
        final var ledgerId = com.hedera.pbj.runtime.io.buffer.Bytes.fromHex(LEDGER_ID);
```

MrValioBg (dismissed):
LGTM! 💯

MrValioBg (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -60,7 +62,7 @@ public static Tuple expiryTupleFor(@NonNull final Token token) {
         return Tuple.of(
                 token.expirationSecond(),
                 headlongAddressOf(token.autoRenewAccountIdOrElse(ZERO_ACCOUNT_ID)),
-                token.autoRenewSeconds());
+                (token.autoRenewSeconds() > 0 ? token.autoRenewSeconds() : 0));
```
tinker-michaelj, line 65: Is it possible for `token.autoRenewSeconds()` to be `< 0`?

Seems this ternary expression is redundant?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
```diff
@@ -132,6 +132,10 @@ public static long asExactLongValueOrZero(@NonNull final BigInteger value) {
      */
     public static com.esaulpaugh.headlong.abi.Address headlongAddressOf(@NonNull final AccountID accountID) {
         requireNonNull(accountID);
+        if (accountID.account().kind() == AccountID.AccountOneOfType.UNSET) {
```
tinker-michaelj, line 135: ⚠️ I'm very concerned if it was possible to reach here with an `AccountID` of `UNSET` type. 😱 

Was adding this necessary?

tinker-michaelj (changes requested):
Overall looks great, but there is a very scary workaround for an `UNSET` account id that we should either fix the cause of; or remove that workaround.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
ilko-iliev-lime: The test failed, because of a negative value. That's why I added the expression.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
ilko-iliev-lime: @tinker-michaelj When the spenderId of the token is null, we then use the ZERO_ACCOUNT_ID, which has an alias == null and that's where we failed with NPE on line 182. I could move the check one level above, but am not sure there are other tests failing because of that.

tinker-michaelj (dismissed):
LGTM, ty @ilko-iliev-lime !

tinker-michaelj (approved):
LGTM, ty @ilko-iliev-lime !

PR #10051, opened 2023-11-21
chore: simplify Gradle signing setup

jjohannes:
**Description**:
This allows running maven central publishing with configuration cache enabled, so we do not need to call Gradle in a special way here on CI.

Revert of / follow up to: #10027 

**Checklist**

### DIFF
--- a/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.maven-publish.gradle.kts
+++ b/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.maven-publish.gradle.kts
@@ -25,54 +25,50 @@ java {
     withSourcesJar()
 }
 
-publishing {
-    publications {
-        create<MavenPublication>("maven") {
-            from(components.getByName("java"))
-            versionMapping {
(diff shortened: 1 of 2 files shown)


build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.maven-publish.gradle.kts
```diff
@@ -25,54 +25,50 @@ java {
...
 
-signing {
-    useGpgCmd()
-    sign(publishing.publications.getByName("maven"))
-}
+val publishSigningEnabled =
+    providers.gradleProperty("publishSigningEnabled").getOrElse("false").toBoolean()
 
-tasks.withType<Sign>().configureEach {
-    onlyIf { providers.gradleProperty("publishSigningEnabled").getOrElse("false").toBoolean() }
```
jjohannes, line 77: Accessing `gradleProperty` inside `onlyIf` is not allowed (`onlyIf` runs "too close" to execution time).
Instead, we make the whole "sigining {}" configuration conditional (see below).

PR #10056, opened 2023-11-21
build: patch everything we use to be a real Java Module

jjohannes:
**Description**:
- Following up to #8561 – all Jars that are only used for testing, jmh, and annotation processing are now also patched to get a `module-info`.
- The new `failOnAutomaticModules.set(true)` feature of `extra-module-info` ensures that no non-module Jars can be added (without corresponding patching) in the future. 
- Updates in the `java-module-dependencies` plugin allow us to remove some tweaks we had to do

Again, there are probably missing `requires` to core Java Modules in some of the patched Jars. This will be seen once we start using the Module Path at (test) runtime. This will be follow up work.

**Notes for reviewer**:

**Checklist**


platform-sdk/platform-apps/tests/StressTestingTool/settings.gradle.kts
```diff
@@ -1,21 +0,0 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-pluginManagement { includeBuild("../../../../build-logic") }
-
-includeBuild("../../../../hedera-dependency-versions")
-
-includeBuild("../../../../platform-sdk")
```
jjohannes, line 21: This file is unused code - leftover that we missed during the Gradle migration steps

Issue #9767, opened 2023-11-08
`PrngSystemContract` should externalize `InvalidTransactionException#getStatusCode()` as `errorMessage`


tinker-michaelj:
### Description

`PrngSystemContract` is [alway externalizing `FAIL_INVALID`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java#L89) as its error message, even when the cause is not an internal failure.

Instead it should externalize an `errorMsg` of `e.getResponseCode()`.

### Steps to reproduce

Call the `Prng` system contract with insufficent gas. Error message in contract call result will be `FAIL_INVALID` instead of `INSUFFICIENT_GAS`.

### Hedera network

other

### Version

`develop`

### Operating system

None

Issue #9425, opened 2023-10-23
Unconditional, enriched, traced records of `UtilPrng` system contract calls.


tinker-michaelj:
Validate the below happy path scenarios:
 - [x] The record produced by a successful call to the `PrngSystemContract` matches that produced by mono-service when the top-level transaction resolves to `SUCCESS`.
 - [x] A call to the `PrngSystemContract` can still be made within a `ContractCallLocal`.

And valid the below failure scenarios:
 - [x] The record produced by a successful call to the `PrngSystemContract` matches that produced by mono-service when the top-level transaction resolves to `CONTRACT_REVERT_EXECUTED`.
  - [x] A call to the `PrngSystemContract` with `INSUFFICIENT_GAS` matches that produced by mono-service.

PR #10074, opened 2023-11-22
feat: Unconditional, enriched, traced records of UtilPrng system contract calls

vtronkov:
**Description**:
* Syncing the child records in mod with the child records in mono 

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/9425
Fixes https://github.com/hashgraph/hedera-services/issues/9767
Fixes https://github.com/hashgraph/hedera-services/issues/10330

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/SystemContractUtils.java
```diff
@@ -18,16 +18,35 @@
...
+    PrngSystemPrecompiledContract.computePrecompile > createSuccessfulChildRecord >
+    addContractCallResultToRecord > PrecompileUtils.addContractCallResultToRecord.
+    For the contractId in EvmFnResult is passed HTS_PRECOMPILE_MIRROR_ENTITY_ID - which is using the
+    HTC contract address(0x167). This seems like a bug to me.
+    Is this how it's suppose to work? Should I fix it in mono or should I just mimic it here and
+    create a story to later fix it?
+     */
+    public static final String HTS_PRECOMPILED_CONTRACT_ADDRESS = "0x167";
+    public static final ContractID HTS_PRECOMPILE_MIRROR_ID = contractIdFromEvmAddress(
+            Address.fromHexString(HTS_PRECOMPILED_CONTRACT_ADDRESS).toArrayUnsafe());
```
vtronkov, line 48: Thoughts?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/PrngPrecompileSuite.java
```diff
@@ -261,6 +293,24 @@ GET_SEED, prng, isRandomResult(new Object[] {new byte[32]}))))
                         .logged());
     }
 
+    @HapiTest
+    private HapiSpec prngPrecompileInsufficientGas() {
+        final var prng = THE_PRNG_CONTRACT;
+        final var randomBits = "randomBits";
+        return defaultHapiSpec("prngPrecompileInsufficientGas")
+                .given(
+                        // snapshotMode(FUZZY_MATCH_AGAINST_HAPI_TEST_STREAMS),
```
vtronkov, line 302: Will be enabled in https://github.com/hashgraph/hedera-services/issues/10166

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -108,20 +115,35 @@ void createSuccessfulRecord(
...
 
     void createFailedRecord(
-            @NonNull MessageFrame frame, @NonNull final String errorMsg, @NonNull final ContractID contractID) {
+            @NonNull MessageFrame frame,
+            @NonNull final ResponseCodeEnum responseCode,
+            @NonNull final ContractID contractID) {
         if (!frame.isStatic()) {
             requireNonNull(frame);
             requireNonNull(contractID);
-            contractFunctionResultFailedFor(gasRequirement, errorMsg, contractID);
```
vtronkov, line 121: This was called twice and the result from the method call was not used

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -108,20 +115,35 @@ void createSuccessfulRecord(
...
             var updater = (ProxyWorldUpdater) frame.getWorldUpdater();
-            updater.externalizeSystemContractResults(
-                    contractFunctionResultFailedFor(gasRequirement, errorMsg, contractID), FAIL_INVALID);
+
+            final var senderId = ((ProxyEvmAccount) updater.getAccount(frame.getSenderAddress())).hederaId();
+            var contractResult = contractFunctionResultFailedFor(gasRequirement, responseCode.toString(), contractID);
+            contractResult = contractResult
+                    .copyBuilder()
+                    .functionParameters(tuweniToPbjBytes(frame.getInputData()))
+                    .errorMessage(null)
```
vtronkov, line 140: The error message is not expected for prng contract

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleSystemContractOperations.java
```diff
@@ -94,11 +101,33 @@ public ContractCallRecordBuilder externalizePreemptedDispatch(
     public void externalizeResult(
             @NonNull final ContractFunctionResult result, @NonNull final ResponseCodeEnum responseStatus) {
         final var childRecordBuilder = context.addChildRecordBuilder(ContractCallRecordBuilder.class);
-        childRecordBuilder
-                .transaction(Transaction.DEFAULT)
-                .contractID(result.contractID())
-                .status(responseStatus)
-                .contractCallResult(result);
+
+        final var body = TransactionBody.newBuilder()
```
tinker-michaelj, line 105: Did we consider dispatching a `UtilPrngTransactionBody` instead of using this one-off pattern?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
```diff
@@ -106,11 +108,18 @@ private static FullResult resultOfExecuting(
...
                     final var output = pricedResult.fullResult().result().getOutput();
+                    var updater = (ProxyWorldUpdater) frame.getWorldUpdater();
+                    final var senderId = ((ProxyEvmAccount) updater.getAccount(frame.getSenderAddress())).hederaId();
+
                     enhancement
                             .systemOperations()
                             .externalizeResult(
                                     contractFunctionResultSuccessFor(
-                                            pricedResult.fullResult().gasRequirement(), output, HTS_CONTRACT_ID),
+                                            pricedResult.fullResult().gasRequirement(),
```
tinker-michaelj, line 118: Would this be externalizing a `UtilPrngTransactionBody` for all view calls?

tinker-michaelj (changes requested):
Now that this PR shows the work we need to get an appropriate record for `PrngSystemContract`, it looks very much like our `PrngSystemContract`  should just dispatch a `UtilPrngTransactionBody` (as almost all HTS system contract calls do).

What do you think @lukelee-sl ?

lukelee-sl:
> Now that this PR shows the work we need to get an appropriate record for `PrngSystemContract`, it looks very much like our `PrngSystemContract` implementation consist of just dispatching a `UtilPrngTransactionBody` (as almost all HTS system contract calls do).
> 
> What do you think @lukelee-sl ?

That makes a lot of sense to me.  Seeing how things work for the HTS system contract it would be nice to conform to that standard.

vtronkov:
Note: the `PrngPrecompileSuite` might fail now but tomorrow I will re-generate the json files and it should pass.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleSystemContractOperations.java
vtronkov: Okay, now this logic is only for the failed scenario. I'm using this pattern here instead of using dispatch because as far as I can see it can't work with the dispatch. If I dispatch a UTIL_PRGN it will always generate prng number but in the case when it's failing we don't want that. We only want to have an empty field.

The other option I can think of is If I dispatch `UtilPrngTransactionBody` with some negative `range` and then change the `UtilPrngHandler` to check if the range is negative then just create an empty prng number field or something like that. But this doesn't sit well with me 🤷 I'm open to discussions here.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
vtronkov: With the new changes I made - it will not 😄

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -108,20 +119,53 @@ void createSuccessfulRecord(
...
+
+            var data = contractFunctionResultSuccessFor(
+                    gasRequirement, randomNum, frame.getRemainingGas(), frame.getInputData(), senderId);
+
+            updater.enhancement()
+                    .systemOperations()
+                    .dispatch(
+                            synthBody(),
+                            new ActiveContractVerificationStrategy(
+                                    senderId.accountNum(), contractID.evmAddress(), false, UseTopLevelSigs.NO),
```
vtronkov, line 132: Not sure about the parameters here

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleSystemContractOperations.java
tinker-michaelj: Ah I see @vtronkov...what about using `systemContractOperations().externalizePreemptedDispatch()` as [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersCall.java#L118)?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/SystemContractUtils.java
tinker-michaelj: Ah we can fix later imo, since we are still aiming to do "differential testing" between mono- and mod-service record streams.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleSystemContractOperations.java
vtronkov: Good point, I've missed that method. Fixed 👍

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
```diff
@@ -108,11 +110,18 @@ private static FullResult resultOfExecuting(
 
                 if (responseCode == SUCCESS) {
                     final var output = pricedResult.fullResult().result().getOutput();
+                    var updater = (ProxyWorldUpdater) frame.getWorldUpdater();
+                    final var senderId = ((ProxyEvmAccount) updater.getAccount(frame.getSenderAddress())).hederaId();
```
tinker-michaelj, line 114: We can get the `senderId` directly from `attempt.senderId()` here instead of going through the `updater` again.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -74,6 +84,7 @@ public FullResult computeFully(@NonNull final Bytes input, @NonNull final Messag
         final ContractID contractID = asEvmContractId(Address.fromHexString(PRNG_PRECOMPILE_ADDRESS));
 
         try {
+            validateTrue(frame.getRemainingGas() >= gasRequirement, INSUFFICIENT_GAS);
```
tinker-michaelj, line 87: We shouldn't need this, should we?

I'd expect [this code path](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java#L181) to already take care of that.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -108,38 +119,72 @@ void createSuccessfulRecord(
...
+            final var senderId = ((ProxyEvmAccount) updater.getAccount(frame.getSenderAddress())).hederaId();
+
+            var data = contractFunctionResultSuccessFor(
+                    gasRequirement, randomNum, frame.getRemainingGas(), frame.getInputData(), senderId);
+
+            updater.enhancement()
+                    .systemOperations()
+                    .dispatch(
+                            synthBody(),
+                            new ActiveContractVerificationStrategy(
```
tinker-michaelj, line 131: This `UtilPrngTransactionBody` has no non-payer signatures required.

So really we could pass in a `VerificationStrategy` that just always returns `Decision.INVALID`, since it will never be used.

Just a nice-to-have, not blocking.

tinker-michaelj (approved):
LGTM, very nice work @vtronkov 💯

PR #10077, opened 2023-11-22
Provide more Mint X tests for Non-payer key verification

### LINKED ISSUE #10041: mintToken (MintTokenPrecompile) - X tests key verification

dikel:
**Description**:
Adds additional negative mint X test scenario for non-payer keys and fixes the old tests

**Related issue(s)**:

Fixes #10041 

**Checklist**


petreze (approved):

agadzhalov (approved):
LGTM

stoyanov-st (commented):
Nice work! Make sure to add negative tests for all the selectors and token types.

agadzhalov (dismissed):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -70,18 +69,12 @@ protected long mintFungible(
...
-            final boolean isMintOnTokenCreation,
             @NonNull final WritableAccountStore accountStore,
             @NonNull final WritableTokenStore tokenStore,
             @NonNull final WritableTokenRelationStore tokenRelationStore) {
         requireNonNull(token);
         requireNonNull(treasuryRel);
-        // validate token supply key exists for mint or burn.
-        // But this flag is not set when mint is called on token creation with initial supply.
-        // We don't need to check the supply key ONLY in that case
-        if (!isMintOnTokenCreation) {
```
tinker-michaelj, line 82: ⚠️ I don't understand how we can safely delete this...won't it break any `@HapiTest` that creates a token without a supply key?

tinker-michaelj (changes requested):
Tests look great, but I don't understand how we can safely remove the `isMintOnTokenCreation` flag. (See comment.) 🤔

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
dikel: Hey @tinker-michaelj, the `BaseTokenHandler::mintFungible` is only called from 2 places: `TokenCreateHandler` where the `isMintOnTokenCreation` flag is hard-coded to true and from `TokenMintHandler`, where the flag is false. So the supply key check is only executed on mints. Thats why I removed the logic from the `BaseTokenHandler` and moved it in the `TokenMintHandler`

tinker-michaelj: Ah makes sense. 🙏

tinker-michaelj (approved):
LGTM, ty @dikel !

PR #10078, opened 2023-11-22
Transition to RECONNECT_COMPLETE earlier in the process

### LINKED ISSUE #9164: Reconnect state saving too quickly causes status to remain in `RECONNECT_COMPLETE`
- If a state received via reconnect saves prior to the status state machine transitioning from `BEHIND` to `RECONNECT_COMPLETE`, then the platform will wait until the next periodic snapshot to transition out of `RECONNECT_COMPLETE`

alittley:
- closes #9164

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
@@ -69,7 +69,6 @@ import com.swirlds.common.system.events.EventDescriptor;
 import com.swirlds.common.system.status.PlatformStatus;
 import com.swirlds.common.system.status.PlatformStatusManager;
 import com.swirlds.common.system.status.actions.DoneReplayingEventsAction;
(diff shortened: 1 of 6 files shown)


cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
```diff
@@ -1043,7 +1042,6 @@ private void loadReconnectState(final SignedState signedState) {
 
         gossip.resetFallenBehind();
         eventCreator.resumeEventCreation();
-        platformStatusManager.submitStatusAction(new ReconnectCompleteAction(signedState.getRound()));
```
lpetrovic05, line 1046: I think it might have been better to just move this line up before the state saving is kicked off. That is a smaller change, and we also don't add a new dependency to reconnet

alittley: Fixed

edward-swirldslabs (approved):

PR #10086, opened 2023-11-22
9483 Added metrics for total space consumption by compaction level

### LINKED ISSUE #9483: MerkleDb metrics for total file size at each compaction level
This is a follow-up for #7501. With that feature, MerkleDb now has different compaction levels, and data is migrating from lower levels to higher levels over time. There is a couple metrics for each level: time spent in ms, and disk space saved in Mb. It would b great to have one more metric per level: total file size.

imalygin:
**Description**:

This PR adds an additional set of metrics to estimate total space consumption in file storages per compaction level

**Related issue(s)**:

Fixes #9483 

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
@@ -94,14 +94,20 @@ public class MerkleDbStatistics {
     private final List<LongAccumulator> hashesStoreCompactionTimeMsList;
     /** Hashes store compactions - saved space in Mb */
     private final List<DoubleAccumulator> hashesStoreCompactionSavedSpaceMbList;
-    /** Leaves store small compactions - time in ms */
+    /** Hashes store - total space in Mb */
+    private final List<DoubleAccumulator> hashesStoreTotalSpaceMbList;
+    /** Leaves store compactions - time in ms */
     private final List<LongAccumulator> leavesStoreCompactionTimeMsList;
-    /** Leaves store small compactions - saved space in Mb */
+    /** Leaves store compactions - saved space in Mb */
     private final List<DoubleAccumulator> leavesStoreCompactionSavedSpaceMbList;
-    /** Leaf keys store small compactions - time in ms */
+
+    private final List<DoubleAccumulator> leavesStoreTotalSpaceMbList;
+    /** Leaf keys store compactions - time in ms */
     private final List<LongAccumulator> leafKeysStoreCompactionTimeMsList;
-    /** Leaf keys store small compactions - saved space in Mb */
+    /** Leaf keys store compactions - saved space in Mb */
     private final List<DoubleAccumulator> leafKeysStoreCompactionSavedSpaceMbList;
+
+    private final List<DoubleAccumulator> leafKeysStoreTotalSpaceMbList;
     /** Off-heap usage in MB of hashes store index */
     private IntegerGauge offHeapHashesIndexMb;
     /** Off-heap usage in MB of leaves store index */
@@ -125,10 +131,13 @@ public class MerkleDbStatistics {
         this.label = CommonUtils.throwArgNull(label, "label");
         hashesStoreCompactionTimeMsList = new ArrayList<>();
         hashesStoreCompactionSavedSpaceMbList = new ArrayList<>();
+        hashesStoreTotalSpaceMbList = new ArrayList<>();
         leavesStoreCompactionTimeMsList = new ArrayList<>();
         leavesStoreCompactionSavedSpaceMbList = new ArrayList<>();
+        leavesStoreTotalSpaceMbList = new ArrayList<>();
         leafKeysStoreCompactionTimeMsList = new ArrayList<>();
         leafKeysStoreCompactionSavedSpaceMbList = new ArrayList<>();
+        leafKeysStoreTotalSpaceMbList = new ArrayList<>();
     }
 
     private static IntegerGauge buildIntegerGauge(final Metrics metrics, final String name, final String description) {
@@ -240,6 +249,10 @@ public class MerkleDbStatistics {
                     metrics,
                     DS_PREFIX + COMPACTIONS_PREFIX + LEVEL_PREFIX + level + "_hashesSavedSpaceMb_" + label,
                     "Saved space during compactions of level %s, hashes store, %s, Mb".formatted(level, label)));
+            hashesStoreTotalSpaceMbList.add(buildDoubleAccumulator(
+                    metrics,
+                    DS_PREFIX + COMPACTIONS_PREFIX + LEVEL_PREFIX + level + "_hashesTotalSpaceMb_" + label,
+                    "Total space taken by files of level %s, hashes store, %s, Mb".formatted(level, label)));
 
             // Leaves store
             leavesStoreCompactionTimeMsList.add(buildLongAccumulator(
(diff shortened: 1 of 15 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCompactor.java
```diff
@@ -438,6 +442,18 @@ public boolean compact() throws IOException, InterruptedException {
                     (filesToCompactSize - compactedFilesSize) * UnitConstants.BYTES_TO_MEBIBYTES);
         }
 
+        if (reportTotalSpaceMetricFunction != null) {
```
artemananiev, line 445: Is this the only place where to update "total space" metrics? What about MerkleDbDataSource.saveRecords(), which creates files on compaction level 0?

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
```diff
@@ -240,6 +249,10 @@ public void registerMetrics(final Metrics metrics) {
                     metrics,
                     DS_PREFIX + COMPACTIONS_PREFIX + LEVEL_PREFIX + level + "_hashesSavedSpaceMb_" + label,
                     "Saved space during compactions of level %s, hashes store, %s, Mb".formatted(level, label)));
+            hashesStoreTotalSpaceMbList.add(buildDoubleAccumulator(
+                    metrics,
+                    DS_PREFIX + COMPACTIONS_PREFIX + LEVEL_PREFIX + level + "_hashesTotalSpaceMb_" + label,
```
artemananiev, line 254: "total space size in Mb" is probably not the best name suffix for the new metrics. We already have metrics named "files size in Mb" (e.g. `hashesStoreFileSizeMb`). These two names look too similar. What about "file size by level in Mb" (e.g. `hashesFileSizeByLevelMb`)?

artemananiev, line 254: Since the new metrics are not specific to compactions, I'd add them to FILES category rather than to COMPACTIONS.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCompactor.java
```diff
@@ -438,6 +442,18 @@ public boolean compact() throws IOException, InterruptedException {
...
+            Map<Integer, List<DataFileReader<?>>> readersByLevel =
+                    getReadersByLevel(dataFileCollection.getAllCompletedFiles());
+            for (int i = 0; i < MAX_COMPACTION_LEVEL; i++) {
```
jsync-swirlds, line 448: Why not use the actual number of entries in the map instead of the maximum compaction level?

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
```diff
@@ -229,37 +238,48 @@ public void registerMetrics(final Metrics metrics) {
 
         // Compaction
 
-        for (int i = 0; i < config.maxCompactionLevel(); i++) {
-            int level = i + 1;
+        for (int i = 0; i <= config.maxCompactionLevel(); i++) {
```
jsync-swirlds, line 241: nit: The bulk of the loop might read a little better if we renamed `i` to `level`.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
```diff
@@ -458,12 +478,27 @@ public void setHashesStoreCompactionTimeMs(final Integer compactionLevel, final
      * @param value the value to set
      */
     public void setHashesStoreCompactionSavedSpaceMb(final int compactionLevel, final double value) {
-        assert compactionLevel >= 1 && compactionLevel <= config.maxCompactionLevel();
+        assert compactionLevel >=0 && compactionLevel <= config.maxCompactionLevel();
```
jsync-swirlds, line 481: nit: missing a space after `>=`

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Sure, I'll update the name

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCompactor.java
imalygin: For no particular reason. You're right, map size as a right boundary is more appropriate.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Fair enough. I'll rename it

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Agreed

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCompactor.java
imalygin: The question of how often we want these metrics to be reported. Right now it's implemented so that it's reported every compaction. In this case we'll get the best case scenario of the space usage because most of the default compaction level files should be eliminated. If we add it to saveRecords(), we'll see some dynamics of the default compaction level.

imalygin: It makes sense, I'll add it.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Done

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Done

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
imalygin: Fixed

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileCompactor.java
imalygin: I ended up with adding it into `DataFileCompactor#compact` at the very beginning. Here is the logic:
1. We want to report this metric every time we save records to the disk
2. Method `DataFileCompactor#compact` is invoked every time we save records to the disk but we don't always run the compaction because there may be not enough files to compact. Therefore, it a good place to access the size of the files prior to compaction.

artemananiev (approved):

jsync-swirlds (approved):

artemananiev (approved):

PR #10097, opened 2023-11-23
fix: disassociate deleted nft does not commit a transfer list

georgi-l95:
**Description**:
This PR move `dissociationFromExpiredTokensAsExpected()` test to `TokenUnhappyAccountsSuite` as it's using expiry logic, which is not implemented.
Fixes dissaciation of deleted tokens and especially nfts, by checking for changes in the token reletions to add them in the transfers list while finalizing the record.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
@@ -83,13 +83,15 @@ public class RecordFinalizerBase {
      *
      * @param writableTokenRelStore the {@link WritableTokenRelationStore} to get the token relation balances from
      * @param tokenStore the {@link ReadableTokenStore} to get the token from
+     * @param nftChanges the map of nft changes from previous step
      * @return a {@link Map} of {@link EntityIDPair} to {@link Long} representing the token relation balances for all
      * modified token relations
      */
     @NonNull
     protected Map<EntityIDPair, Long> fungibleChangesFrom(
             @NonNull final WritableTokenRelationStore writableTokenRelStore,
-            @NonNull final ReadableTokenStore tokenStore) {
+            @NonNull final ReadableTokenStore tokenStore,
+            final Map<TokenID, List<NftTransfer>> nftChanges) {
         final var fungibleChanges = new HashMap<EntityIDPair, Long>();
         for (final EntityIDPair modifiedRel : writableTokenRelStore.modifiedTokens()) {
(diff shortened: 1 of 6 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
```diff
@@ -120,6 +121,45 @@ protected Map<EntityIDPair, Long> fungibleChangesFrom(
         return fungibleChanges;
     }
 
+    /**
+     * Gets all non-fungible tokenRelation balances for all modified token relations from the given {@link WritableTokenRelationStore}.
+     *
+     * @param writableTokenRelStore the {@link WritableTokenRelationStore} to get the token relation balances from
+     * @param tokenStore the {@link ReadableTokenStore} to get the token from
+     */
+    protected void nonFungibleChangesFrom(
```
tinker-michaelj, line 130: Hmm...
 - This is just a copy+paste of the `fungibleChangesFrom()` method with a different name, right?
 - And the goal is to make the record fungible transfer list include the #-of-NFT-owned "balances" that change **specifically** when a deleted or expired non-fungible token is dissociated from an account, correct?

Assuming I understand the goal here, imo instead we should:
 1. Rename `fungibleChangesFrom()` to `tokenBalanceChangesFrom()`; and,
 2. Add a third parameter to specify the `TokenType` that should be filtered for [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java#L98).

(Also, since this code will run **every** `handleTransaction`, we might consider adding a `boolean` return value to the `tokenBalanceChangesFrom()` method that lets us detect whether any token types were skipped and only do this in the specific case above.)

But this amount of duplication is undesirable imo. 😞

tinker-michaelj (changes requested):
I believe the _intent_ of the code is correct, but imo the amount of duplication makes that intent very hard to discern. Suggested a solution in the comment.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
georgi-l95: Yes, you're right. I've refactor the change. Can you take a look again? Thank you 🙏

tinker-michaelj (approved):
LGTM, tyvm @georgi-l95 !

PR #10108, opened 2023-11-27
fix: 2098 metadata clone

### LINKED ISSUE #2098: Return `.clone()` of NFT `metadata` to query code
### Problem

The `MerkleUniqueToken` returns the underlying `metadata` array to query code. Although this code certainly does not mutate the `metadata`, it is a best practice to return a defensive copy in this case.

### Solution

Return `metadata.clone()` instead of `metadata`.

### Alternatives

_No response_

mmalik-al:
**Description**:
This PR changes the following:
* Return `metadata.clone()` instead of metadata in `MerkleUniqueToken`
* Fixes `MerkleUniqueTokenTest` to verify array contents
* Fixes `NftPropertyTest` to verify array contents

**Related issue(s)**:
Closes #2098


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleUniqueToken.java
```diff
@@ -247,7 +251,9 @@ public void setNext(final NftNumPair next) {
     }
 
     public byte[] getMetadata() {
-        return metadata;
+        return Optional.ofNullable(metadata)
+                .map(byte[]::clone)
+                .orElse(null);
```
nathanklick, line 256: Using an `Optional` with the use of `.map().orElse()` here creates unnecessary memory allocations and potentially creates a performance issue here by putting unnecessary strain on the garbage collector.

```suggestion
        return (metadata != null) ? metadata.clone() : null;
```

mmalik-al:
@jsync-swirlds @nathanklick I've pushed changes that should solve issues mentioned in the comments

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleUniqueToken.java
```diff
@@ -52,12 +52,15 @@
 import com.swirlds.common.merkle.MerkleLeaf;
 import com.swirlds.common.merkle.impl.PartialMerkleLeaf;
 import com.swirlds.common.merkle.utility.Keyed;
+
```
jsync-swirlds, line 55: You will notice that the spotless (code style enforcement) checks failed on this line.  The code style does not allow for separating groups of imports with blank lines like this one.
The gradle command `./gradlew spotlessApply` can be run on the code locally to fix this and any other style issues.

jsync-swirlds:
> @jsync-swirlds @nathanklick I've pushed changes that should solve issues mentioned in the comments

I've enabled the workflows to automatically test the code with this change.  A couple style enforcement checks are failing; those should be easy to fix.
The `PR Formatting/Title Check` is wanting the _title_ of the PR to conform to "Conventional Commit" syntax, which you can find information about [here](https://www.conventionalcommits.org/en/v1.0.0/).  Only the title of the PR is checked and needs to match that convention.

mmalik-al:
@jsync-swirlds Thanks, I've merged latest develop branch into this PR. Hopefully it will be ok this time.

tinker-michaelj (approved):
LGTM!

jsync-swirlds (approved):

jsync-swirlds:
Accepted.  Thank you for your contribution, @mmalik-al.

PR #10137, opened 2023-11-28
fix: add record list checkpoint, so following & preceding child records are reverted correctly

### LINKED ISSUE #9403: Fix propagatesNestedCreations, evmLazyCreateViaSolidityCall tests
We get more than expected number of child records, we might not be handling calls to revert properly e.g. the  grandchild.propagate_abortively() call from NestedCreations.sol or the nestedLazyCreateThenRevert() call from NestedLazyCreateContract

### LINKED ISSUE #10352: `Create2OperationSuite.canUseAliasesInPrecompilesAndContractKeys()`
`ContractID`/`delegatable_contract_id` keys with 20-byte EVM addresses should be considered non-empty keys for token roles.

### LINKED ISSUE #10419: (mod-service) Should be able to revert suffix of preceding records
### Description

Since the message call represented by a given `MessageFrame` can dispatch revertable children both following (internal `CONTRACT_CREATION`) **and** preceding (lazy-creation), we need to be able to specify reverting from the state of the `RecordListBuilder` at the beginning of an EVM message call.

### Steps to reproduce

Run `propagatesNestedCreations()`.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

MrValioBg:
**Description**:

Add support for reverting call from specific "checkpoint". Every time a new ProxyWorldUpdater is created we make a checkpoint. When revert is called we revert both preceding & following child records from this checkpoint.

This will fix more tests in LeakyContractTestSuites and in Create2OperationSuite.

**Related issue(s)**:

Fixes #9403 
Fixes #10419 
Fixes #10352 

**Notes for reviewer**:

**Todo**

- [x] Verify that javadoc is ok (Code comments, README, etc.)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -385,6 +385,25 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
         }
     }
 
+    /**
+     * Reverts or removes the last child transaction on which the revert() method was called.
+     * <p> If the child record is REVERSIBLE, then the status is set to {@link ResponseCodeEnum#REVERTED_SUCCESS}.
+     * If the child record is REMOVABLE, then the record is removed from the list.
+     */
+    public void revertLastChildOnly() {
```
MrValioBg, line 393: Hey @Neeharika-Sompalli and @tinker-michaelj, I need your opinion on this. Is it okay to revert only the last child record in this case, considering the explanation in the other comment? If we anticipate that the child calling revert might not always be the last one, perhaps we can keep track of the children or the context?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -395,6 +396,38 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
...
+            return;
+        }
+
+        // We get to here when the revert was executed on a child transaction
+        // We need to revert all children that were added after the child transaction that was reverted
+        revertChildrenOf((SingleTransactionRecordBuilderImpl) checkPoint.lastFollowingRecord());
+
+        // We also need to revert all preceding transactions that were added after the first preceding transaction
+        if (checkPoint.firstPrecedingRecord() != null) {
+            var indexOf = precedingTxnRecordBuilders.indexOf(checkPoint.firstPrecedingRecord());
```
tinker-michaelj, line 416: As in your comment, don't we want to revert only preceding records added **after** `checkPoint.firstPrecedingRecord()`?

I.e., shouldn't we use `indexOf + 1` here?

MrValioBg: Yes, thats true. Great catch! 🙏🏻 
Just commited the changes :)

tinker-michaelj (approved):
LGTM, nice work @MrValioBg !

PR #10141, opened 2023-11-28
fix: canMergeCreate2MultipleCreatesWithHollowAccount

### LINKED ISSUE #9922: Fix `canMergeCreate2MultipleCreatesWithHollowAccount` e2e test

agadzhalov:
**Description**:
The test was failing due to two reasons:

1. Hardcoded [INITIAL_CONTRACT_NONCE](https://github.com/hashgraph/hedera-services/pull/10141/files#diff-e14f53f3715b9e33a6663bf27ac3a2634eacee2bcedbfcd154597a19341521d5L125) which is always `1`.
    - To fix it `finalizeHollowAccountAsContract()`  should accept the contract nonce value and [set](https://github.com/hashgraph/hedera-services/blob/268db0f4049c1a71ddf03249d8e96c7973e88e03/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java#L124) it to the resulting account after merging with the hollow account.
    - Logic in [mono](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaCreateOperationExternalizer.java#L159)
2. Wrong `contractId` externalized in [finalizeHollowAccount()](https://github.com/hashgraph/hedera-services/pull/10141/files#diff-d0ef7ec62e8802a84c068417e0c720f2761cf243b82d4cec3cf8f80682a51c4aR291)
    - `getHederaContractId()` retrieves wrong [pendingCreation](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java#L162) number 
    - To fix it we have to return the proper `hollowAccountId` where the finalization happens - [finalizeHollowAccountAsContract()](https://github.com/hashgraph/hedera-services/blob/268db0f4049c1a71ddf03249d8e96c7973e88e03/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java#L127) and pass it to the [extarnalization](https://github.com/hashgraph/hedera-services/blob/268db0f4049c1a71ddf03249d8e96c7973e88e03/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java#L289) logic.
    - The use of `getHederaContractId()` in this context seems to be redundant 

**Related issue(s)**:

Fixes #9922

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
@@ -19,7 +19,6 @@ package com.hedera.node.app.service.contract.impl.exec.scope;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_SIGNATURE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.OK;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.SUCCESS;
-import static com.hedera.node.app.service.contract.impl.exec.processors.ProcessorModule.INITIAL_CONTRACT_NONCE;
 import static com.hedera.node.app.service.contract.impl.utils.SynthTxnUtils.synthHollowAccountCreation;
 import static com.hedera.node.app.spi.workflows.HandleContext.TransactionCategory.CHILD;
 import static java.util.Objects.requireNonNull;
@@ -122,7 +121,7 @@ public class HandleHederaNativeOperations implements HederaNativeOperations {
         final var accountStore = context.readableStore(ReadableAccountStore.class);
         final var hollowAccountId = requireNonNull(accountStore.getAccountIDByAlias(evmAddress));
         final var tokenServiceApi = context.serviceApi(TokenServiceApi.class);
-        tokenServiceApi.finalizeHollowAccountAsContract(hollowAccountId, INITIAL_CONTRACT_NONCE);
+        tokenServiceApi.finalizeHollowAccountAsContract(hollowAccountId);
         // FUTURE: For temporary backward-compatibility with mono-service, consume an entity id
         context.newEntityNum();
     }
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
@@ -19,7 +19,6 @@ package com.hedera.node.app.service.contract.impl.exec.scope;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_SIGNATURE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.OK;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.SUCCESS;
-import static com.hedera.node.app.service.contract.impl.exec.processors.ProcessorModule.INITIAL_CONTRACT_NONCE;
 import static com.hedera.node.app.service.contract.impl.utils.SynthTxnUtils.synthHollowAccountCreation;
 import static com.hedera.node.app.spi.workflows.HandleContext.TransactionCategory.CHILD;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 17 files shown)


petreze (dismissed):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java
```diff
@@ -282,10 +283,10 @@ public void setupInternalAliasedCreate(@NonNull final Address origin, @NonNull f
      * {@inheritDoc}
      */
     @Override
-    public void finalizeHollowAccount(@NonNull final Address alias) {
-        evmFrameState.finalizeHollowAccount(alias);
+    public void finalizeHollowAccount(@NonNull final Address alias, final long contractNonce) {
+        AccountID hollowAccountId = evmFrameState.finalizeHollowAccount(alias, contractNonce);
         // add child record on merge
-        var contractId = getHederaContractId(alias);
+        var contractId = ConversionUtils.asNumericContractId(hollowAccountId);
```
MiroslavGatsanoga, line 289: **question:** Should we reset the `pendingCreation` to null for this case in which we merge the CREATE2 contract creation into the existing hollow account i.e. if we leave set as it is now could that lead to some unexpected side effect?

tinker-michaelj: I think @MiroslavGatsanoga has a point here...In fact, if we set `pendingCreation = null` **before** calling `getHederaContractId(alias)`, everything will work as expected without needing to return the `hollowAccountId` from `finalizeHollowAccount()`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
```diff
@@ -117,14 +116,15 @@ public void setNonce(final long contractNumber, final long nonce) {
...
      */
     @Override
-    public void finalizeHollowAccountAsContract(@NonNull final Bytes evmAddress) {
+    public AccountID finalizeHollowAccountAsContract(@NonNull final Bytes evmAddress, final long contractNonce) {
         requireNonNull(evmAddress);
         final var accountStore = context.readableStore(ReadableAccountStore.class);
         final var hollowAccountId = requireNonNull(accountStore.getAccountIDByAlias(evmAddress));
         final var tokenServiceApi = context.serviceApi(TokenServiceApi.class);
-        tokenServiceApi.finalizeHollowAccountAsContract(hollowAccountId, INITIAL_CONTRACT_NONCE);
+        tokenServiceApi.finalizeHollowAccountAsContract(hollowAccountId, contractNonce);
```
tinker-michaelj, line 124: Just like the redundancy in `mono-service`, afaict this is a no-op, simply setting the existing hollow account's `nonce` to be what it already is.

To confirm, I just changed two lines from `develop`:
  1. Added `pendingCreation = null` before L#288 [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java#L288).
  2. Commented out [this line](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/api/TokenServiceApiImpl.java#L175) that was always resetting the hollow account's nonce to `INITIAL_CONTRACT_NONCE`.
  
...and `canMergeCreate2MultipleCreatesWithHollowAccount()` passed with no other changes. 😁 

So I think we can just do that `pendingCreation = null` trick and remove the `contractNonce` parameter from ` TokenServiceApi.finalizeHollowAccountAsContract()` and this issue will be fixed.

tinker-michaelj (changes requested):
Very nice work @agadzhalov, I think mono-service had redundant code here though 😝. (See inline comments). 

As noted in [this comment](https://github.com/hashgraph/hedera-services/pull/10141/files#r1411461078) I think we can actually make this a zero-line fix by adding one line in `ProxyWorldUpdater` and subtracting one line from `TokenServiceApi`. 

Lmk what you think...afaict it works. 🤷

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
agadzhalov: addressed. @MiroslavGatsanoga thank you for the guiding questing. @tinker-michaelj thank you for the suggestion. This way the solution is way more elegant. 🙏

tinker-michaelj (approved):
Tyvm @agadzhalov , LGTM!

PR #10176, opened 2023-11-29
feat: Implement synthetic records for immediate genesis reconnect scenario

### LINKED ISSUE #9757: Handle event stream recovery failure via synthetic records
[This PR](https://github.com/hashgraph/hedera-services/pull/7987) fixes an issue in the mono service code where an ISS is possible when an event stream recovery is initiated during services startup. The fix essentially generates synthetic records in this case, but without actually creating account objects in state. 

We need to implement an equivalent mechanism in the modular startup code.

mhess-swl:
The basic idea behind this PR is that the `TokenSchema` class has been split out into two parts: the synthetic records that need to be created, and the actual creation of the accounts. The `Hedera` class now has logic to determine if any synthetic records need to be generated, and if so, the `SyntheticRecordsGenerator` class will generate _only_ the records. If applicable, these synthetic records will also be passed to the token schema class to actually create the entities in state. 

The modular behavior is now similar to the mono service in how it handles this scenario, but unfortunately we haven't figured out a way to test this. 

Closes #9757


hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -241,7 +243,9 @@ public Hedera(@NonNull final ConstructableRegistry constructableRegistry) {
                 () -> HapiUtils.toString(version.getServicesVersion()),
                 () -> HapiUtils.toString(version.getHapiVersion()));
 
-        // Create a record builder for any genesis records that need to be created
+        // Create a records generator for any synthetic records that need to be CREATED
+        this.recordsGenerator = new SyntheticRecordsGenerator();
+        // Create a records builder for any genesis records that need to be RECORDED
         this.genesisRecordsBuilder = new GenesisRecordsConsensusHook();
```
tinker-michaelj, line 249: Should `GenesisRecordsConsensusHook` be renamed to `GenesisRecordsBuilder`? Imo as written, this is mildly confusing.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -354,8 +363,39 @@ private void onStateInitialized(
...
             @NonNull final InitTrigger trigger,
             @Nullable final SoftwareVersion previousVersion) {
+        // Initialize the configuration from disk. We must do this BEFORE we run migration, because the various
+        // migration methods may depend on configuration to do their work. For example, the token service migration code
+        // needs to know the token treasury account, which has an account ID specified in config. The initial config
+        // file in state, created by the file service migration, will match what we have here, so we don't have to worry
+        // about re-loading config after migration.
+        logger.info("Initializing configuration with trigger {}", trigger);
+        switch (trigger) {
+            case EVENT_STREAM_RECOVERY -> configProvider =
```
tinker-michaelj, line 373: Why is this case different from `RESTART, RECONNECT`?

tinker-michaelj (approved):
One question and a couple nits but LGTM, tyvm @mhess-swl !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
mhess-swl: I'm fine with that. This class acts as both a record builder and a consensus hook, but that's definitely not clear here

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
mhess-swl: Good question. The code initializing this config provider was previously instantiated with `new ConfigProviderImpl(false)`, so maybe that's already sufficient and the extra config constructor isn't needed. I'll ask around.

iwsimon (approved):
LGTM, small comment.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/SyntheticRecordsGenerator.java
```diff
@@ -0,0 +1,264 @@
...
+    public static long[] nonContractSystemNums(final long numReservedSystemEntities) {
+        return LongStream.rangeClosed(FIRST_POST_SYSTEM_FILE_ENTITY, numReservedSystemEntities)
+                .filter(i -> i < FIRST_RESERVED_SYSTEM_CONTRACT || i > LAST_RESERVED_SYSTEM_CONTRACT)
+                .toArray();
+    }
+
+    @NonNull
+    private Key superUserKey(@NonNull final BootstrapConfig bootstrapConfig) {
+        final var superUserKeyBytes = bootstrapConfig.genesisPublicKey();
+        if (superUserKeyBytes.length() != 32) {
```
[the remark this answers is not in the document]
mhess-swl: I would if I knew why it needs to be 32 bytes 😅

iwsimon (approved):
LGTM

tinker-michaelj (approved):
LGTM, tyvm @mhess-swl !

Issue #10181, opened 2023-11-29
ISS detection is broken


cody-littley:
As of 0.44, ISS detection is broken.

PR #10182, opened 2023-11-30
Fix ISS detection.

cody-littley:
Related to #10181

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
@@ -108,29 +108,9 @@ public class BasicSoftwareVersion implements SoftwareVersion {
         }
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-
-        if (!(o instanceof BasicSoftwareVersion that)) {
-            return false;
-        }
-
-        return softwareVersion == that.softwareVersion;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public int hashCode() {
-        return Long.hashCode(softwareVersion);
-    }
+    // Intentionally do not implement equals() or hashCode(). Although it is legal to do so, it is not required,
+    // and all platform operations should be functional without it. Since this class is used for platform testing,
+    // we should make sure this crutch is not available.
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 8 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
```diff
@@ -108,29 +108,9 @@ public int compareTo(final SoftwareVersion that) {
...
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-
-        if (!(o instanceof BasicSoftwareVersion that)) {
-            return false;
-        }
-
-        return softwareVersion == that.softwareVersion;
```
edward-swirldslabs, line 124: Would it be better to implement it but throw an exception if you want to force us to not use it anywhere?

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
```diff
@@ -108,29 +108,9 @@ public int compareTo(final SoftwareVersion that) {
...
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public int hashCode() {
-        return Long.hashCode(softwareVersion);
-    }
+    // Intentionally do not implement equals() or hashCode(). Although it is legal to do so, it is not required,
+    // and all platform operations should be functional without it. Since this class is used for platform testing,
+    // we should make sure this crutch is not available.
```
lpetrovic05, line 114: I don't understand the benefit of removing it

lpetrovic05, line 114: as I understand it, the Hedera software doesn't implement this, so the test apps shouldn't either. Can you please make that more clear in this comment

cody-littley: I feel like the current comment already says this directly.

> Since this class is used for platform testing,
    // we should make sure this crutch is not available.

Can you suggest a better wording?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
cody-littley: `BasicSoftwareVersion` was created with the idea that some applications may choose to use it directly as opposed to implementing their own software version. Even though this is used for testing, it's not considered test code, and so adding logic designed for tests seems like the wrong thing to do.

alittley (approved):

PR #10183, opened 2023-11-30
Fix ISS detection.

cody-littley:
Related to #10181

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
@@ -108,29 +108,9 @@ public class BasicSoftwareVersion implements SoftwareVersion {
         }
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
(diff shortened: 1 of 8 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
```diff
@@ -108,29 +108,9 @@ public int compareTo(final SoftwareVersion that) {
...
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-
-        if (!(o instanceof BasicSoftwareVersion that)) {
```
edward-swirldslabs, line 120: Would it be better to implement it and throw an exception?

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/system/BasicSoftwareVersion.java
cody-littley: This is intended to be used by non-hedera applications, so I think throwing is the wrong thing to do. It would make sense to do this if this class was not accessible outside of test code.

PR #10185, opened 2023-11-30
fix: Fuzzy matching for CreateOperationSuite and Create2OperationSuite 09431

JivkoKelchev:
**Description**:

This PR modifies contract creation records to match records from mono. It cover all CreateOperationSuite and Create2OperationSuite tests with fuzzy match against mono.

**Related issue(s)**:

Fixes part of #9431

**Notes for reviewer**:

1. There are four test with `FULLY_NONDETERMINISTIC` mode, because in case of nested creates, the order of records is inverted. 
For example: we have Contract `Grandparent` has a constructor that creates an instance of `Parent`, whose constructor itself creates an instance of `Child`.
- In mono the order of records would be: record of `Grandparent`, record of `Parent`, record of `Child`
- In mod the order is: record of `Child`, record of `Parent`, record of `Grandparent`

2. There are two test without fuzzy match enabled
- `Create2OperationSuite#canCallFinalizedContractViaHapi`. There is a [PR](https://github.com/hashgraph/hedera-services/pull/10539) that will fix the problems and fuzzy match can be enabled, once the PR is merged.
- `Create2OperationSuite#canUseAliasesInPrecompilesAndContractKeys`. Here I tried to implement the same behavior as mono by introducing `STANDIN_CONTRACT_ID_KEY` concept, but after that I had troubles with transfer allowance. So I decide to leave it to someone who is more familiar with this topic.

**Checklist**


tinker-michaelj (approved):
LGTM!

PR #10200, opened 2023-11-30
fix: Handle ArithmeticException in safeFractionMultiply()

### LINKED ISSUE #9690: Handle ArithmeticException in safeFractionMultiply()
The method `AdjustmentUtils.safeFractionMultiply()` throws an `ArithmeticException` if the `denominator` is `0`. 

We should add this to the javadocs of the method. Also, we need to look at all the places where this method is called and either check the denominator before calling the method or catching the `ArithmeticException`.

iwsimon:
**Description**:

**Related issue(s)**:

Fixes #9690 

**Notes for reviewer**:

**Checklist**


Neeharika-Sompalli (approved):
LGTM ! Seems there are HAPI test failures

PR #10212, opened 2023-12-01
fix: use mono failure code for immutable account debit in HTS transfer

### LINKED ISSUE #10177: `AtomicCryptoTransferHTSSuite.cryptoTransferSpecialAccounts()`
mono-service uses `INVALID_FULL_SIGNATURE_PREFIX_FOR_PRECOMPILE` when an HTS transfer system contract tries to debit an immutable account; mod-service dispatch returns `INVALID_ACCOUNT_ID`.

### LINKED ISSUE #10220: fix: enable `ContractCallSuite.invalidContract()`
We should validate the called `ContractID` exists and abort before EVM transaction if not.

tinker-michaelj:
**Description**:
  - Closes #10177 
  - Closes #10220 
  - Translates `INVALID_ACCOUNT_ID` to `INVALID_FULL_PREFIX_SIGNATURE_FOR_PRECOMPILE` if the dispatched transfer tried to debit an immutable system account.
      * Purely done for mono-service compatibility and differential testing.
      * Avoids duplication with `SystemAccountCreditScreen` by extracting a `TransferAuditTest` type.


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -108,6 +108,7 @@ public CallOutcome call() {
                     result.recipientId(),
                     result.gasPrice());
         } catch (HandleException abort) {
+            abort.printStackTrace();
```
lukelee-sl, line 111: is this appropriate?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HederaOperations.java
```diff
@@ -242,4 +247,29 @@ public interface HederaOperations {
...
+
+    /**
+     * Given a {@link ContractID} and the current Hedera config, returns the given id if its shard and realm
+     * match the config; otherwise, returns {@link #MISSING_CONTRACT_ID}.
+     *
+     * @param contractId the contract ID to validate
+     * @param hederaConfig the current Hedera config
+     * @return the validated contract ID
+     */
+    default ContractID configValidated(@NonNull final ContractID contractId, @NonNull final HederaConfig hederaConfig) {
```
lukelee-sl, line 268: not sure how this will affect the stuff I'm working on with allowing calls to non-existing contracts.  will test after this merges.

lukelee-sl (approved):
LGTM.  Thanks for teasing out that ugly debit logic.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HederaOperations.java
tinker-michaelj: Hmm good question...but I assume in any case we do not want to allow calls to contracts in a different shard?

Issue #9396, opened 2023-10-23
Fix transferFailsWithIncorrectAmounts test


yosiftouma:
We get an unexpected INSUFFICIENT_TOKEN_BALANCE child record instead of no child records at all

PR #10214, opened 2023-12-01
test: Fix transferFailsWithIncorrectAmounts test

stoyanov-st:
**Description**:
Added a missing check for the amounts to be transferred while executing the classic token transfer function.

**Related issue(s)**:

Fixes #
#9396 
**Notes for reviewer**:

**Checklist**


MrValioBg (dismissed):
LGTM!

netopyr (dismissed):

tinker-michaelj (changes requested):
We don't want to throw exceptions out of `HtsCall` implementations; we should instead simply return an appropriate `haltResult()` or `revertResult()` when a negative number is passed here.

tinker-michaelj (approved):
I think it's more general and readable to add a `@Nullable ResponseCodeEnum preemptingFailureStatus` to the `ClassicTransfersCall` constructor; and set it based on the 3rd argument being negative.

But this works for now, tyvm @stoyanov-st !

tinker-michaelj (changes requested):
Hmm actually this change has unintended consequences and breaks [other specs](https://scans.gradle.com/s/latgf2qeba5yk).

I'd suggest adding the `@Nullable ResponseCodeEnum preemptingFailureStatus`.

Screen Shot 2023-12-14 at 9 56 34 AM

tinker-michaelj (approved):
Looks great @stoyanov-st , tyvm! 💯

PR #10230, opened 2023-12-01
fix: Resolve EthereumSuite Errors

### LINKED ISSUE #8674: Resolve `EthereumSuite` errors
### Problem

We have the following tests suite `EthereumSuite` that has problems when it's running against modularization codebase.

### Solution

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Enables all but 1 Hapi tests in `EthereumSuite`

**Related issue(s)**:

Fixes #8674

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
```diff
@@ -296,6 +295,66 @@ List<HapiSpec> feePaymentMatrix() {
                 .toList();
     }
 
+    @HapiTest
+    HapiSpec matrixedPayerRelayerTest1() {
```
Ivo-Yankov, line 299: I did not know how to add the `@HapiTest` annotation to the tests from `feePaymentMatrix` so I created separate functions. I am open to suggestions on how to do this in a more better way.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java
```diff
@@ -205,10 +208,34 @@ private InvolvedParties computeInvolvedParties(
         if (transaction.isCreate()) {
             final Address to;
             if (transaction.isEthereumTransaction()) {
+                final ReadableAccountStore accountStore =
+                        updater.enhancement().nativeOperations().readableAccountStore();
+                var createBody = requireNonNull(transaction.hapiCreation());
```
tinker-michaelj, line 213: Imo it would be much cleaner to include this code directly where the synthetic `ContractCreateTransactionBody` is first built [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/SynthTxnUtils.java#L149). (The `senderId` and `accountStore` should both be available there.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
```diff
@@ -64,41 +72,55 @@ public EthereumTransactionHandler(
...
                 requireNonNull(requireNonNull(component.hydratedEthTxData()).ethTxData());
+
+        // Run its in-scope transaction and get the outcome
+        final var outcome = component.contextTransactionProcessor().call();
+
         final var recordBuilder = context.recordBuilder(EthereumTransactionRecordBuilder.class)
                 .ethereumHash(Bytes.wrap(ethTxData.getEthereumHash()))
-                .status(outcome.status());
+                .status(outcome.status())
+                .feeChargedToPayer(outcome.tinybarGasCost());
```
tinker-michaelj, line 112: I think the two changes from [this commit](https://github.com/hashgraph/hedera-services/commit/1313d2ead9c91d5d79cbb6677eca632a2ffe16d7) that re-use the existing `recordBuilder.withTinybarGasFee()` method is preferable. 

Very nice catch though. 🙌

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
```diff
@@ -64,41 +72,55 @@ public EthereumTransactionHandler(
...
     }
 
     @Override
     public void handle(@NonNull final HandleContext context) throws HandleException {
         // Create the transaction-scoped component
         final var component = provider.get().create(context, ETHEREUM_TRANSACTION);
 
-        // Run its in-scope transaction and get the outcome
-        final var outcome = component.contextTransactionProcessor().call();
+        final var hevmTransactionFactory = component.contextTransactionProcessor().hevmTransactionFactory;
```
tinker-michaelj, line 96: Hmm...we don't need to do all this to validate the sender nonce. We can just add the validation [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java#L235) like,
```
        if (transaction.isEthereumTransaction()) {
            validateTrue(transaction.nonce() == parties.sender().getNonce(), WRONG_NONCE);
            parties.sender().incrementNonce();
        }
```
and then add translation for `WRONG_NONCE` as one more `else if` branch [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java#L142),
```
    private static final Bytes WRONG_NONCE_REASON = Bytes.wrap(WRONG_NONCE.name());
    ...
                } else if (revertReason.equals(WRONG_NONCE_REASON)) {
                return WRONG_NONCE;

```
etc.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
tinker-michaelj: This looks good to me!

tinker-michaelj (approved):
Really nice work @Ivo-Yankov, I added three comments and assigned you to a [new issue](https://github.com/hashgraph/hedera-services/issues/10599) to resolve them since checks are passing here!

PR #10278, opened 2023-12-05
build: perform git clone/update of 'hedera-protobufs' in a task

jjohannes:
**Description**:

After #9651, we now fully control the `git` commands that clone and update `hedera-protobufs` and can now:
- do the clone in a task instead of (always) during Gradle configuration
- clone into a subfolder of `build`, so `gradle clean` will clean it up as well

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1062,10 +1062,6 @@ saved.txt
 *.pces
 *.pcesD
 
-### Hedera Node Files
-hedera-node/hapi/hedera-protobufs
-hedera-node/hapi/checkouts.bin
-
 # Do not ignore the files required by the Platform Core tests
 !platform-sdk/**/src/test/resources/**/*.jar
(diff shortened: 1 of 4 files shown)


nickpoorman:
@jjohannes When I change the origin and then switch back to the `develop` branch, it assumes the origin is correct. We probably need a step in there that verifies it has the correct origin. Otherwise, `git fetch` doesn't fetch the branches from the correct origin and the `git checkout` fails.

jjohannes:
@nickpoorman I think I don't fully understand the workflow yet, when you work with _hedera-protobufs_. Do you mean while working with the fork, this solution is fine, but the moment you are done, it is broken? Do you want a check that checks the `origin` (of the other repo) only if we are on `develop` (in this repo)?

Can we integrate this PR first and do that as follow-up?

nathanklick (approved):

PR #10282, opened 2023-12-05
fix: fuzzy match records with mono-service for ContractMintHTSSuite

### LINKED ISSUE #10218: Fuzzy match records with mono-service for ContractMintHTSSuite

thenswan:
**Description**:
Fuzzy match records with mono-service for ContractMintHTSSuite

**Related issue(s)**:
Fixes #10218


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -367,6 +367,7 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
                 childRecordBuilders.set(i, null);
             } else {
                 if (child.reversingBehavior() == ReversingBehavior.REVERSIBLE && SUCCESSES.contains(child.status())) {
+                    child.tokenTransferLists().clear();
```
tinker-michaelj, line 370: Awesome catch. 🙌 I created [this issue](https://github.com/hashgraph/hedera-services/issues/10308) to give `SingleTransactionRecordBuilderImpl` a method to null out all its "side-effect" fields like `tokenTransferLists`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
```diff
@@ -316,9 +318,14 @@ private TokenTransferList sendingUnitsFromTo(
             @NonNull final AccountID to,
             final long amount,
             final IsApproval isApproval) {
+        final var accountAmounts = new ArrayList<AccountAmount>();
```
tinker-michaelj, line 321: Another nice catch. 💯

Issue #10217, opened 2023-12-01
Rename test classes to follow one consistent scheme


mmalik-al:
### Problem

Currently we have many test classes that are named either `*Test.java` or `*Tests.java`. This introduces unnecessary inconsistencies across entire project. 

### Solution

* Rename all test classes to the same suffix. `*Test.java` OR `*Tests.java`

poulok:
Closing as this is not important to fix and will cause lots of conflicts with priority PRs.

PR #10286, opened 2023-12-05
chore: rename test classes to follow one consistent scheme

mwb-al:
**Description:**
This PR changes the following:

- Rename all test classes to the same suffix *Tests.java

**Related issue(s):**
Closes [#10217](https://github.com/hashgraph/hedera-services/issues/10217)


hendrikebbers:
Hi @mwb-al 
thank you very much for your contribution! Sorry for the delay, at the end of the year we missed to look at community contributions based on holidays.
This Change is way to big for one PR since it will mostly daly end in merge conflicts. Would it be possible to do the editing in smaller PRs? Like having 1 PR for each module of the projects. Best would be to start with a small module and once the PR is merged doing all the rest. What do you think.

Issue #9399, opened 2023-10-23
Fix createTokenWithInvalidRoyaltyFee test


yosiftouma:
- Overriding CONTRACTS_MAX_NUM_WITH_HAPI_SIGS_ACCESS to “10_000_000” that contains underscores seems to throw an exception
- We don’t get the CUSTOM_FEE_MUST_BE_POSITIVE in the contractCallResult

PR #10287, opened 2023-12-05
fix: LeakyContractTestsSuite HAPI tests 

### LINKED ISSUE #9398: Fix createTokenWithInvalidFeeCollector, createTokenWithInvalidFixedFeeWithERC721Denomination tests

MrValioBg:
**Description**:

- Use ContracCallRecordBuilder instead of CryptoCreateRecordBuilder
- Also adds support for multiple priority statuses (implementation provided by a cherry-picked commit authored by @tinker-michaelj).
 - Don't activate tests that seem to require legacy key verification - comment above them about the behavior.

These changes aim to provide correct logic and fix tests in the LeakyContractTestsSuite.

**Related issue(s)**:

Fixes #9398 
#9399 
#9397 

**Notes for reviewer**:
Please check the github comments I left in the code.

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateTranslator.java
```diff
@@ -152,7 +152,7 @@ public ClassicCreatesCall callFrom(@NonNull HtsCallAttempt attempt) {
                 attempt.systemContractGasCalculator(),
                 attempt.enhancement(),
                 nominalBodyFor(attempt),
-                attempt.defaultVerificationStrategy(),
+                attempt.contractVerificationStrategy(),
```
MrValioBg, line 155: Is this behaviour expected? Should we allow top level signatures for crypto create ?  

I am seeking opinions on this, since couldn't find how its exactly done it mono, because of the different flows.

tinker-michaelj (changes requested):
We should just add a comment to any `HapiSpec` that expects to be able to use top-level signatures like,
```
// Requires legacy security model, cannot be enabled as @HapiTest without refactoring to use contract keys
```

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateTranslator.java
MrValioBg: But wouldn't this result in the test failing in mono? 🤔

tinker-michaelj (dismissed):
LGTM, ty @MrValioBg !

tinker-michaelj (approved):
LGTM!

PR #10312, opened 2023-12-06
fix: correct owner/spender priority addresses

### LINKED ISSUE #10313: `ContractMusicalChairsSuite.playGame()`

### LINKED ISSUE #10326: Use only `null` for unset NFT spender / treasury owner in mod `TokenService`, never `AccountID.DEFAULT`
### Problem

Currently both `AccountID.DEFAULT` and `null` are used for an NFT's `spenderId` and `ownerId` fields (the former when it is unset, the latter when the NFT is treasury-owned).

### Solution

Only use `null` for these cases so a `nft.hasSpenderId()` or `nft.hasOwnerId()` check allows us to apply conditional logic targeting these cases.

### Alternatives

_No response_

### LINKED ISSUE #10331: `ContractUpdateSuite.fridayThe13thSpec()`
Should use `MODIFYING_IMMUTABLE_CONTRACT` instead of `INVALID_SIGNATURE` when update targets an immutable contract.

### LINKED ISSUE #10333: `ContractUpdateSuite.canMakeContractImmutableWithEmptyKeyList()`

### LINKED ISSUE #10350: `Create2OperationSuite.canDeleteViaAlias()`
Need to enforce `OBTAINER_SAME_CONTRACT_ID`; and clear the alias mapping for a contract deleted via HAPI.

### LINKED ISSUE #10351: `Create2OperationSuite.priorityAddressIsCreate2ForStaticHapiCalls()`
For mono-service fidelity, we should permit `0.0.X` ids for HAPI calls to contracts with priority EVM addresses.

tinker-michaelj:
**Description**:
 - Closes #10326 
 - Closes #10331 
 - Closes #10333
 - Closes #10313 
 - Closes #10350 
 - Closes #10351 
 - Updates `TokenService` components to consistently use `null` to reflect (1) treasury NFT ownership and (2) missing NFT `spender` ids.
 - Fixes `nftTokenInfoTupleFor()` to return priority addresses for NFT owners and spenders.


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/NFTOwnersChangeStep.java
```diff
@@ -87,18 +87,12 @@ public void doIn(final TransferContext transferContext) {
                     validateSpenderHasAllowance(senderAccount, topLevelPayer, tokenId, nft);
                 }
 
-                final var copyNft = nft.copyBuilder();
-                // If the nft owner is not set then set it to the treasury account
-                if (!nft.hasOwnerId() || nft.ownerId().equals(AccountID.DEFAULT)) {
-                    copyNft.ownerId(treasury);
-                }
```
Neeharika-Sompalli, line 94: Should we delete this ? Or just change the condition in `if` to see if owner is null ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/api/TokenServiceApiImpl.java
```diff
@@ -489,7 +489,12 @@ public void deleteAndTransfer(
         // get the account from account store that has all balance changes
         // commit the account with deleted flag set to true
         final var updatedDeleteAccount = requireNonNull(accountStore.getForModify(deletedId));
-        accountStore.put(updatedDeleteAccount.copyBuilder().deleted(true).build());
+        accountStore.removeAlias(updatedDeleteAccount.alias());
+        accountStore.put(updatedDeleteAccount
+                .copyBuilder()
+                .alias(Bytes.EMPTY)
```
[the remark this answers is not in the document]
tinker-michaelj: For mono-service fidelity with `GetContractInfo` query, we need to.

Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj 🙂

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -233,17 +236,28 @@ public static Tuple nftTokenInfoTupleFor(
...
+        requireNonNull(token);
+        requireNonNull(ledgerId);
+        requireNonNull(nativeOperations);
 
+        final var nftMetaData = nft.metadata() != null ? nft.metadata().toByteArray() : Bytes.EMPTY.toByteArray();
         return Tuple.of(
                 tokenInfoTupleFor(token, ledgerId),
                 serialNumber,
-                headlongAddressOf(ownerAccount),
+                priorityAddressOf(nft.ownerIdOrElse(token.treasuryAccountIdOrElse(ZERO_ACCOUNT_ID)), nativeOperations),
```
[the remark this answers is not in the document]
tinker-michaelj: Oh actually `Token.DEFAULT` can be passed into that method, so we do need the `OrElse(ZERO_ACCOUNT_ID)` construct. 😝

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -233,17 +236,30 @@ public static Tuple nftTokenInfoTupleFor(
...
+                // The odd construct allowing a token to not have a treasury account set is to accommodate
+                // Token.DEFAULT being passed into this method, which a few HtsCall implementations do
+                priorityAddressOf(nft.ownerIdOrElse(token.treasuryAccountIdOrElse(ZERO_ACCOUNT_ID)), nativeOperations),
                 nft.mintTimeOrElse(new Timestamp(0, 0)).seconds(),
                 nftMetaData,
-                headlongAddressOf(nft.spenderIdOrElse(ZERO_ACCOUNT_ID)));
+                priorityAddressOf(nft.spenderIdOrElse(ZERO_ACCOUNT_ID), nativeOperations));
+    }
+
+    private static Address priorityAddressOf(
```
david-bakin-sl, line 257: Haven't seen the term "priority address" before.  It's in the title of this PR but otherwise not.  What does it mean?

I do see a comment way over at `EvmFrameState` which is not an obvious spot to look. Maybe the comment here?  (And based on that comment I'm still not sure why it is a "priority" address.) (We really do need a glossary of addresses ...)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -62,7 +65,7 @@ public static Tuple expiryTupleFor(@NonNull final Token token) {
         return Tuple.of(
                 token.expirationSecond(),
                 headlongAddressOf(token.autoRenewAccountIdOrElse(ZERO_ACCOUNT_ID)),
-                (token.autoRenewSeconds() > 0 ? token.autoRenewSeconds() : 0));
```
david-bakin-sl, line 65: I don't see anywhere in the code where non-negativity of autoRenewSeconds is enforced, if not here. Maybe it doesn't belong here but instead over at Token.Builder.autoRenewSeconds (which unfortunately is generated by PBJ).

Anyway, I may have missed it.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmStaticTransactionFactory.java
```diff
@@ -65,18 +70,15 @@ public HederaEvmTransaction fromHapiQuery(@NonNull final Query query) {
         final var op = query.contractCallLocalOrThrow();
         assertValidCall(op);
         final var senderId = op.hasSenderId() ? op.senderIdOrThrow() : payerId;
+        var targetId = op.contractIDOrThrow();
+        // For mono-service fidelity, allow calls using 0.0.X id even to contracts with a priority EVM address
+        final var maybeContract =
+                context.createStore(ReadableAccountStore.class).getContractById(targetId);
+        if (maybeContract != null && maybeContract.alias().length() == EVM_ADDRESS_LENGTH_AS_LONG) {
```
david-bakin-sl, line 77: I think `alias()` can return `null` (in general, perhaps not in this case?)

david-bakin-sl (commented):
Got part way though this and realized I didn't have sufficient knowledge to complete it.  Pinged Luke to do this properly.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -233,17 +236,30 @@ public static Tuple nftTokenInfoTupleFor(
...
                 nft.mintTimeOrElse(new Timestamp(0, 0)).seconds(),
                 nftMetaData,
-                headlongAddressOf(nft.spenderIdOrElse(ZERO_ACCOUNT_ID)));
+                priorityAddressOf(nft.spenderIdOrElse(ZERO_ACCOUNT_ID), nativeOperations));
+    }
+
+    private static Address priorityAddressOf(
+            @NonNull final AccountID accountId, @NonNull final HederaNativeOperations nativeOperations) {
+        requireNonNull(accountId);
+        return (ZERO_ACCOUNT_ID == accountId)
```
lukelee-sl, line 260: Should we use `equals`?  I realize that `ZERO_ACCOUNT_ID` is static and probably what was used but what if someone creates a constant somewhere else?  Maybe I worry too much.

lukelee-sl:
Are we really using `ZERO_ADDRESS` instead of null to mark unset spender, treasury account?

lukelee-sl (approved):
LGTM, I wonder if there are other places where there needs to be similar changes.  Thank you

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
tinker-michaelj: Any time a contract has an `0xab...cd` EVM address (outside the long-zero subspace), it _must_ be referenced by that address anywhere inside the EVM. (This is to avoid creating issues with contracts that use addresses as e.g. keys in a ERC-20 `balances` map; or store a particular address as a privileged account.) 

I.e. `0xab...cd` addresses take "priority" over long-zero addresses.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
tinker-michaelj: `TokenCreateHandler` should not be creating `Token`'s with negative `autoRenewSeconds`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
tinker-michaelj: This use of `ZERO_ACCOUNT_ID` is internal to `TokenTupleUtils`...could also make the first argument here `@Nullable` and then `return (accountId == null)`...semantics should be the same.

Will ponder doing elsewhere...but...passing checks 😳

tinker-michaelj:
> Are we really using `ZERO_ADDRESS` instead of null to mark unset spender, treasury account? Maybe the issue should be corrected if this is the case.

Apparently for info of a missing token we return all zeros in its `Token` struct (including e.g. its `treasury` address)? I'm not entirely sure...didn't dig deep in that direction. 🤔

PR #10322, opened 2023-12-06
feat: Hook midnight rate updates into handle workflow

netopyr:
This PR ensures the midnight rates are updated when a new staking period begins.


tinker-michaelj (approved):
Makes sense to me!

PR #10340, opened 2023-12-06
fix: Validate records of Consensus Service handlers using snapshotMod()

### LINKED ISSUE #10234: Validate records of Consensus Service handlers using snapshotMod()
Validate records of Consensus Service handlers by comparing the records generated from mono-service to the records generated in modular service using snapshotMode()
- [x] Fuzzy match records with mono-service for TopicCreateSuite
- [x] Fuzzy match records with mono-service for TopicDeleteSuite
- [x] Fuzzy match records with mono-service for TopicUpdateSuite
- [x] Fuzzy match records with mono-service for SubmitMessageSuite

iwsimon:
**Description**: Also added a couple of missed error checkings in ConsensusCreateTopicHandler.

**Related issue(s)**:

Fixes #10234 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusCreateTopicHandler.java
```diff
@@ -106,6 +111,13 @@ public void handle(@NonNull final HandleContext handleContext) {
             builder.submitKey(op.submitKey());
         }
 
+        if (op.hasAutoRenewAccount()) {
```
tinker-michaelj, line 114: It would be strongly preferable to re-use [`ExpiryValidatorImpl.resolveCreateAttempt()`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/ExpiryValidatorImpl.java#L66) here.

But if it's overly tricky for some reason, this is ok too. 😁

tinker-michaelj (approved):
LGTM, ty @iwsimon !

Issue #9405, opened 2023-10-23
Fix evmLazyCreateViaSolidityCallTooManyCreatesFails test


yosiftouma:
We should add a check for exceeding max child records and return MAX_CHILD_RECORDS_EXCEEDED instead for the ContractCall of createTooManyHollowAccounts function in NestedLazyCreateContract.sol

PR #10341, opened 2023-12-06
chore: Enable `evmLazyCreateViaSolidityCallTooManyCreatesFails` HapiTest

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/9405

- Change to use `dispatchPrecedingChildRecord` for dispatching lazy creations instead of `dispatchChildTransaction`


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -542,11 +542,13 @@ public <T> T doDispatchPrecedingTransaction(
...
             throw new IllegalArgumentException("Only user- or child-transactions can dispatch preceding transactions");
         }
+        // This condition fails, because for lazy-account creation we charge fees, before dispatching the transaction,
+        // and the state will be modified.
 
-        if (stack.depth() > 1) {
-            throw new IllegalStateException(
-                    "Cannot dispatch a preceding transaction when a savepoint has been created");
-        }
+        //        if (stack.depth() > 1) {
```
lukelee-sl, line 548: Is there a way to retain this check and the one below and only skip for lazy/auto create transactions?

lukelee-sl (dismissed):
LGTM.  Just a small question

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
Neeharika-Sompalli: It will be triggered for any preceding ChildTransactions. Currently it is just lazy and auto-creations. 🙂

tinker-michaelj (approved):
LGTM, ty @Neeharika-Sompalli --- though still feels a bit 🪄 haha...why does it work!?

tinker-michaelj:
Ok, I at least understand the latest [failure mode](https://scans.gradle.com/s/irrxv2xo3y6ua/tests/task/:test-clients:hapiTestMisc/details/com.hedera.services.bdd.suites.leaky.LeakyContractTestsSuite/evmLazyCreateViaSolidityCallTooManyCreatesFails?top-execution=1) (expected `INVALID_ACCOUNT_ID`, was `OK`) for the `evmLazyCreateViaSolidityCallTooManyCreatesFails()` spec; a bit long to type here.

_But_...we also managed to surface a catastrophic timing bug in the latest run of the `deletedAccountCannotBePayer()` spec. From logs, two nodes got
```
2023-12-08 03:09:36.763 ERROR 502  HandleWorkflow - An unexpected exception was thrown during handle
java.lang.NullPointerException: keyVerifications must not be null
	at java.base/java.util.Objects.requireNonNull(Objects.java:233) ~[?:?]
	at com.hedera.node.app.signature.DefaultKeyVerifier.<init>(DefaultKeyVerifier.java:66) ~[app-0.45.0-SNAPSHOT.jar:?]
	at com.hedera.node.app.workflows.handle.HandleWorkflow.handleUserTransaction(HandleWorkflow.java:357) ~[app-0.45.0-SNAPSHOT.jar:?]
	at com.hedera.node.app.workflows.handle.HandleWorkflow.handlePlatformTransaction(HandleWorkflow.java:267) ~[app-0.45.0-SNAPSHOT.jar:?]
	at com.hedera.node.app.workflows.handle.HandleWorkflow.handleRound(HandleWorkflow.java:236) ~[app-0.45.0-SNAPSHOT.jar:?]
```
...and the other two didn't. At first glance, I think there are two problems:
 1. A `NODE_DILIGENCE_FAILURE` pre-handle result has null `keyVerifications`, and this blows up as above in `HandleWorfklow`; plus,
 2. A `SO_FAR_SO_GOOD` pre-handle result that hits a `PreCheckException` in `addMissingSignatures()` simply returns the original result [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L849)---meaning that the two nodes that detected the deleted payer account in `preHandle` got a NPE, while the two nodes that _didn't_ detect it avoided the NPE. 😱

Neeharika-Sompalli:
> Ok, I at least understand the latest [failure mode](https://scans.gradle.com/s/irrxv2xo3y6ua/tests/task/:test-clients:hapiTestMisc/details/com.hedera.services.bdd.suites.leaky.LeakyContractTestsSuite/evmLazyCreateViaSolidityCallTooManyCreatesFails?top-execution=1) (expected `INVALID_ACCOUNT_ID`, was `OK`) for the `evmLazyCreateViaSolidityCallTooManyCreatesFails()` spec; a bit long to type here.
> 
> _But_...we also managed to surface a catastrophic timing bug in the latest run of the `deletedAccountCannotBePayer()` spec. From logs, two nodes got
> 
> ```
> 2023-12-08 03:09:36.763 ERROR 502  HandleWorkflow - An unexpected exception was thrown during handle
> java.lang.NullPointerException: keyVerifications must not be null
> 	at java.base/java.util.Objects.requireNonNull(Objects.java:233) ~[?:?]
> 	at com.hedera.node.app.signature.DefaultKeyVerifier.<init>(DefaultKeyVerifier.java:66) ~[app-0.45.0-SNAPSHOT.jar:?]
> 	at com.hedera.node.app.workflows.handle.HandleWorkflow.handleUserTransaction(HandleWorkflow.java:357) ~[app-0.45.0-SNAPSHOT.jar:?]
> 	at com.hedera.node.app.workflows.handle.HandleWorkflow.handlePlatformTransaction(HandleWorkflow.java:267) ~[app-0.45.0-SNAPSHOT.jar:?]
> 	at com.hedera.node.app.workflows.handle.HandleWorkflow.handleRound(HandleWorkflow.java:236) ~[app-0.45.0-SNAPSHOT.jar:?]
> ```
> 
> ...and the other two didn't. At first glance, I think there are two problems:
> 
> 1. A `NODE_DILIGENCE_FAILURE` pre-handle result has null `keyVerifications`, and this blows up as above in `HandleWorfklow`; plus,
> 2. A `SO_FAR_SO_GOOD` pre-handle result that hits a `PreCheckException` in `addMissingSignatures()` simply returns the original result [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L849)---meaning that the two nodes that detected the deleted payer account in `preHandle` got a NPE, while the two nodes that _didn't_ detect it avoided the NPE. 😱

😱 Wow. Thanks for the detailed explanation @tinker-michaelj . Will look at the logs now

tinker-michaelj:
> I also bumped into the first issue today. What do you think about using empty Sets and Maps in `PreHandleResult` instead of `null` as a short-term solution?
> 
> In the long run, we should revisit `PreHandleResult` and how it is used. The current design is very fragile.

Makes sense...imo even if `PreHandleResult` keeps its current design, we should isolate its construction in a code path that is 100% reusable between `preHandle` and `handle` so we can sleep at night. 😁

tinker-michaelj (approved):
LGTM!

PR #10348, opened 2023-12-06
feat: Set recipient to null as appropriate during contract tracing

### LINKED ISSUE #10346: Set recipient to null as appropriate during contract tracing
### Problem

While creating the trace logs for contract execution, if the recipient address is not in the long-zero address space it cannot be converted into a known entity id.

### Solution

In the case described above, set the recipient to null.

### Alternatives

_No response_

lukelee-sl:
**Description**:
If the recipient address cannot be converted into an Entity Id then set its value to null.  We search for the recipient address in both the ledger and in the alias map prior to setting the value to null.

**Related issue(s)**:

Fixes #10346

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
@@ -174,7 +174,7 @@ public class HederaTracer implements HederaOperationTracer {
                 && messageFrame.getWorldUpdater().getAccount(contractAddress) == null) {
             action.setTargetedAddress(contractAddress.toArray());
         } else {
-            final var recipient = getEntityIdOrNullByAddressAndMessageFrame(contractAddress, messageFrame);
+            final var recipient = getEntityIdOrNullByAddressAndMessageFrame(contractAddress, messageFrame, action);
             if (CodeV0.EMPTY_CODE.equals(messageFrame.getCode())) {
                 // code can be empty when calling precompiles too, but we handle
                 // that in tracePrecompileCall, after precompile execution is completed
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
```diff
@@ -203,8 +203,10 @@ private void finalizeActionFor(final SolidityAction action, final MessageFrame f
                     action.setOutput(frame.getOutputData().toArrayUnsafe());
                     if (action.getInvalidSolidityAddress() != null) {
                         // we had a successful lazy create, replace targeted address with its new Hedera id
```
tinker-michaelj, line 205: Is there a typo in this comment? (At least superficially, it seems if lazy creation was successful, we wouldn't have a non-null `invalidSolidityAddress`?)

lukelee-sl: It looks like the poorly named `invalidSolidityAddress` field is being overloaded to also hold the address in case of a lazy create.  
https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java#L175
https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/SolidityAction.java#L160
I'm not sure how dangerous it'll be to introduce a new proper field.

tinker-michaelj (dismissed):
LGTM!

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/traceability/HederaTracer.java
```diff
@@ -363,4 +365,14 @@ private static <E, I> String get(
             final E subject, final Function<E, I> getter, final Function<I, String> processor) {
         return null != subject ? processor.compose(getter).apply(subject) : "null";
     }
+
+    private EntityId getEntityIdOrNullByAddressAndMessageFrame(
+            Address address, MessageFrame frame, SolidityAction action) {
+        try {
+            return EntityId.fromAddress(asMirrorAddress(address, frame));
+        } catch (IllegalArgumentException e) {
+            action.setTargetedAddress(address.toArray());
```
lukelee-sl, line 374: Steven suggested we add this line so that the user will get some idea of who the recipient was if the lookup fails.

tinker-michaelj (approved):
LGTM!

PR #10356, opened 2023-12-07
fix: finalize actions from stack for ALL failed creation attempts

### LINKED ISSUE #8533: Port sidecar validations to mod `ContractService`
### Problem

Need to port https://github.com/hashgraph/hedera-services/pull/8404 to mod code.

### Solution

Do so.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
- Closes #8533 
- Fix incorrect behavior which, upon a creation failing in a `AbstractMessageProcessor.start()` implementation, would leave the pending action for the originating `CALL`/`CREATE` _still on the tracer's action stack_.
    * This removes the need for the `Source` enum's `POPPED_FROM_STACK`/`READ_FROM_LIST_END` distinction.
- Ensure a failed creation is traced even in the case of a failed value transfer to the new contract address.
    * This removes the need for the `HaltShouldTraceAccountCreation `'s enum's `YES`/`NO` distinction.

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/EvmActionTracer.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/EvmActionTracer.java
@@ -109,10 +109,10 @@ public class EvmActionTracer implements ActionSidecarContentTracer {
             @NonNull final MessageFrame frame, @NonNull final Optional<ExceptionalHaltReason> haltReason) {
         requireNonNull(frame);
         requireNonNull(haltReason);
-        // It is important NOT to finalize the last action on the stack unless the creation
-        // failed, as otherwise the action could accidentally be finalized twice depending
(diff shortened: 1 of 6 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/EvmActionTracer.java
```diff
@@ -111,8 +109,16 @@ public void traceAccountCreationResult(
             @NonNull final MessageFrame frame, @NonNull final Optional<ExceptionalHaltReason> haltReason) {
         requireNonNull(frame);
         requireNonNull(haltReason);
-        if (hasActionSidecarsEnabled(frame)) {
-            actionStack.finalizeLastAction(READ_FROM_LIST_END, frame, stackValidationChoice(frame));
+        // It is important NOT to finalize the last action on the stack unless a halt reason
```
lukelee-sl, line 112: thanks for this comment.  I think it clarifies what the issue is here.

lukelee-sl (approved):
LGTM

PR #10361, opened 2023-12-07
chore: Fix `HashMap` in token service `CryptoTransfer` to `LinkedHashMap`

Neeharika-Sompalli:
Fix `HashMap` in token service `CryptoTransfer` to `LinkedHashMap`


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/leaky/LeakyCryptoTestsSuite.java
```diff
@@ -303,6 +303,10 @@ private HapiSpec getsInsufficientPayerBalanceIfSendingAccountCanPayEverythingBut
                                 .gas(gasToOffer)
                                 .payingWith(civilian)
                                 .balance(maxSendable.get())
+                                // because this fails depending on the previous operation reaching
+                                // consensus before the current operation or after, since we have added
+                                // deferStatusResolution
+                                .hasPrecheckFrom(OK, INSUFFICIENT_PAYER_BALANCE)
```
Neeharika-Sompalli, line 309: This is to fix a flaky integration test reported

PR #10366, opened 2023-12-07
fix: ensure failed `CREATE2` action is finalized w/ proper frame

### LINKED ISSUE #10360: Failed `CREATE2` needs `traceAccountCreationResult()` tracer hook implemented in mono-service
### Description

If a `CREATE2` fails because it collides with an existing contract (i.e. if the existing contract has not been `selfdestruct`ed), then the `ContractAction` sidecar is wrong because it is finalized against the wrong `MessageFrame`.

### Steps to reproduce

Use `CREATE2` to deploy a contract at the same deterministic address twice.

### Additional context

_No response_

### Hedera network

mainnet, other

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #10360 
 - Since `tracePostExecution()` is never called upon a `CREATE2` address collision, re-implements `traceAccountCreationResult()` in mono-service `HederaTracer` and updates the `MessageCallProcessors` not to make duplicate calls to `tracePostExecution()`.

### DIFF
--- a/hedera-node/hedera-app/build.gradle.kts
+++ b/hedera-node/hedera-app/build.gradle.kts
@@ -141,6 +141,32 @@ tasks.jar {
     }
 }
 
+// Define build directory
+val buildDir = layout.projectDirectory.dir("./build/node")
+
+// Copy everything from hedera-node/data
+val copyNodeData =
+    tasks.register<Sync>("copyNodeData") {
+        dependsOn(copyLib)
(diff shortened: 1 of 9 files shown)


iwsimon (dismissed):
LGTM

Nana-EC (dismissed):
LGTM

lukelee-sl (approved):
Makes sense.  LGTM.

PR #10379, opened 2023-12-08
Consider input bytes when calculating gas cost - mono service fix

### LINKED ISSUE #9988: Consider the input bytes when calculating Intrinsic Gas costs in the EVM
We’re ignoring the input data when calculating the intrinsic gas cost and we always forward Bytes.EMPTY to the gas calculator [here](https://github.com/hashgraph/hedera-services/blob/95ff4d5fde1c769c47a988dc75678120561e23fc/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java#L183).
Consider updating the logic to actually forward the input bytes from the transaction object. This will bring us a step closer to Ethereum-equivalency.

stoqnkpL:
**Description**:
Take into consideration the input bytes' size from the evm transaction when calculating the intrinsic gas cost. 

**Related issue(s)**:

Fixes #9988

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
+++ b/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
@@ -179,8 +179,8 @@ public abstract class HederaEvmTxProcessor {
         }
     }
 
-    public void setupFields(final boolean contractCreation) {
-        this.intrinsicGas = gasCalculator.transactionIntrinsicGasCost(Bytes.EMPTY, contractCreation);
+    public void setupFields(final Bytes payload, final boolean contractCreation) {
+        this.intrinsicGas = gasCalculator.transactionIntrinsicGasCost(payload, contractCreation);
         this.updater = worldState.updater();
         this.coinbase = dynamicProperties.fundingAccountAddress();
     }
--- a/hedera-node/hedera-evm/src/test/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessorTest.java
+++ b/hedera-node/hedera-evm/src/test/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessorTest.java
@@ -152,7 +152,7 @@ class HederaEvmTxProcessorTest {
         givenValidMock(0L);
         given(globalDynamicProperties.fundingAccountAddress()).willReturn(fundingAccount);
 
-        evmTxProcessor.setupFields(false);
+        evmTxProcessor.setupFields(Bytes.EMPTY, false);
         final var result =
                 evmTxProcessor.execute(sender, receiver, 33_333L, 1234L, 1L, Bytes.EMPTY, true, mirrorReceiver);
         assertTrue(result.isSuccessful());
@@ -184,7 +184,7 @@ class HederaEvmTxProcessorTest {
 
         given(globalDynamicProperties.maxGasRefundPercentage()).willReturn(MAX_REFUND_PERCENT);
         given(globalDynamicProperties.fundingAccountAddress()).willReturn(fundingAccount);
-        evmTxProcessor.setupFields(false);
+        evmTxProcessor.setupFields(Bytes.EMPTY,false);
         final var result =
                 evmTxProcessor.execute(sender, receiver, 0L, GAS_LIMIT, 1234L, Bytes.EMPTY, false, mirrorReceiver);
         assertTrue(result.isSuccessful());
@@ -200,7 +200,7 @@ class HederaEvmTxProcessorTest {
         given(gasCalculator.transactionIntrinsicGasCost(Bytes.EMPTY, false)).willReturn(intrinsicGasCost);
         given(globalDynamicProperties.fundingAccountAddress()).willReturn(fundingAccount);
 
-        evmTxProcessor.setupFields(false);
+        evmTxProcessor.setupFields(Bytes.EMPTY,false);
         final var result =
                 evmTxProcessor.execute(sender, receiver, 0L, GAS_LIMIT, 1234L, Bytes.EMPTY, false, mirrorReceiver);
         assertTrue(result.isSuccessful());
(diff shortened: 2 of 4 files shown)


hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -179,8 +180,8 @@ public HederaEvmTransactionProcessingResult execute(
         }
     }
 
-    public void setupFields(final boolean contractCreation) {
-        this.intrinsicGas = gasCalculator.transactionIntrinsicGasCost(Bytes.EMPTY, contractCreation);
+    public void setupFields(final Bytes payload, final boolean contractCreation) {
+        this.intrinsicGas = gasCalculator.transactionIntrinsicGasCost(payload, contractCreation);
```
Nana-EC, line 184: Where is `pay-load` defined?

Nana-EC (commented):
LG
One question

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
stoqnkpL: It's extracted from the transaction object and forwarded by EvmTxProcessor.

Nana-EC: Ignore, I was asking why it was `pay-load` vs `payload`.
I couldn't find the variable `pay-load`, turns out the hyphen was just GitHub line wrapping.

Nana-EC (approved):
LG

Is there a way for us to describe the gas calculation so test readers can understand why the new numbers are correct e.g. 21k + x + y etc

Maybe not in this PR if it's a lot of overhead but something to consider to improve.
Maybe not as part of this PR

IvanKavaldzhiev (approved):
LGTM

lukelee-sl:
Should we consider doing the same thing here for modularized smart contract service?  Or in a separate pr?
https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java#L117

tinker-michaelj (approved):
LGTM, tyvm @stoqnkpL !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AssociatePrecompileSuite.java
```diff
@@ -316,7 +316,7 @@ private HapiSpec invalidSingleAbiCallConsumesAllProvidedGas() {
                             .getTransactionRecord(INVALID_SINGLE_ABI_CALL_TXN)
                             .getContractCallResult()
                             .getGasUsed();
-                    assertEquals(99011, gasUsed);
+                    assertEquals(99014, gasUsed);
```
david-bakin-sl, line 319: This number was found empirically, I assume?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractMintHTSSuite.java
```diff
@@ -191,7 +191,7 @@ private HapiSpec transferNftAfterNestedMint() {
                                                                     .withStatus(SUCCESS)
                                                                     .withTotalSupply(1L)
                                                                     .withSerialNumbers(1L))
-                                                            .gas(3_837_920L)
+                                                            .gas(3_836_587L)
                                                             .amount(0L)
```
david-bakin-sl, line 195: And again here?  I have no real objection to finding these numbers by running the test and seeing what works and what fails, it's just unfortunate that that's the way it has to work.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
```diff
@@ -299,7 +300,7 @@ private HapiSpec traceabilityE2EScenario1() {
                                                 .setCallType(CREATE)
                                                 .setCallOperationType(CallOperationType.OP_CREATE)
                                                 .setCallingAccount(TxnUtils.asId(GENESIS, spec))
-                                                .setGas(197000)
+                                                .setGas(48248)
```
david-bakin-sl, line 303: Some of these gas numbers are _smaller_ than before - so that's not the effect of accounting for input bytes.  What is it?

david-bakin-sl (approved):
LGTM.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AssociatePrecompileSuite.java
stoqnkpL: yes, the formula for calculating the gas cost is:
start with the base cost of 21k gas
add 4 gas for each zero in the input bytes
add 16 gas for each non-zero in the input bytes
add 32k gas if this is a creation
It’s all calculated in the [CustomGasCalculator](https://github.com/hashgraph/hedera-services/blob/d0e5885c155e92e69d3e49fb9ef4fe4fe76baa50/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCalculator.java#L50).
I think it's impractical to calculate it per test, but it's useful to have these assertions and checks to help us validate the values only change when it makes sense.

stoqnkpL:
> Should we consider doing the same thing here for modularized smart contract service? Or in a separate pr? https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java#L117

it's already resolved as part of this PR. Or do you think there's more logic that should be updated?

lukelee-sl (commented):
LGTM

PR #10385, opened 2023-12-08
fix: clear "side-effect fields" after switch a child record from SUCCESS to REVERTED_SUCCESS

### LINKED ISSUE #10308: `mod-service` needs equivalent to `ExpirableTxnRecord#Builder.revert()`
### Description

As surfaced [here](https://github.com/hashgraph/hedera-services/pull/10282/files#diff-470dea0a4a08c78bdea7b6a877e6466dbdf1a22598a28c7f012ec4733b3a9860R370) in a fuzzy-matching PR, when we switch a child record from `SUCCESS` to `REVERTED_SUCCESS`, we are not clearing its "side-effect fields" (`transferList`, `tokenTransferLists`, created ids in the `transactionReceipt`, etc.) 

### Steps to reproduce

Look at any child record of e.g. a HTS transfer system contract call with `REVERTED_SUCCESS`. It will still have `tokenTransferLists` set despite them actually being reverted.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

thenswan:
**Description**:
Clear "side-effect fields" after switch a child record from `SUCCESS` to `REVERTED_SUCCESS`.

**Related issue(s)**:
Fixes #10308

**Notes for reviewer**:
At first, I followed this strategy to remove side-effect fields:
- Find specs with system contract call with `REVERTED_SUCCESS`
- Fuzzy match them mod vs mono (individual specs)
- Clear side-effect fields based on results of fuzzy matching

What did I check:
- `ContractMintHTSSuite.rollbackOnFailedMintAfterFungibleTransfer()` — side-effect fields removed in prev PR
- `ContractDeleteSuite.cannotUseMoreThanChildContractLimit()` — checked records and didn’t find any side-effect fields under `REVERTED_SUCCESS`
- `LazyCreateThroughPrecompileSuite.erc20TransferLazyCreate()`  — checked and removed `automatic_token_associations` under `REVERTED_SUCCESS`
- `LazyCreateThroughPrecompileSuite.erc721TransferFromLazyCreate()` — checked records and didn’t find any side-effect fields under `REVERTED_SUCCESS`
- `LazyCreateThroughPrecompileSuite.erc20TransferFromLazyCreate()` is still WIP and I can’t check it
- `Create2OperationSuite.canUseAliasesInPrecompilesAndContractKeys()` is still WIP and I can’t check it

Other side-effect fields were mentioned in the issue (ones I that didnt encounter while did relevant spec fuzzy matching, like: transferList, created ids in the transactionReceipt, etc.) I added anyway from from mono `TxnReceipt#Builder.revert()` and `ExpirableTxnRecord#Builder.nullOutSideEffectFields()`.

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
@@ -32,6 +32,7 @@ import com.hedera.node.app.state.SingleTransactionRecord;
 import com.hedera.node.app.workflows.handle.HandleContextImpl;
 import com.hedera.node.app.workflows.handle.record.SingleTransactionRecordBuilderImpl.ReversingBehavior;
 import com.hedera.node.config.data.ConsensusConfig;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.config.api.Configuration;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.time.Instant;
@@ -376,7 +377,31 @@ public final class RecordListBuilder {
                 followingChildRemoved = true;
             } else {
                 if (child.reversingBehavior() == ReversingBehavior.REVERSIBLE && SUCCESSES.contains(child.status())) {
+                    // clear "side effect" fields
+                    child.accountID(null);
+                    child.contractID(null);
+                    child.fileID(null);
+                    child.tokenID(null);
+                    child.scheduleID(null);
+                    child.scheduledTransactionID(null);
+                    child.serialNumbers().clear();
+                    child.topicRunningHash(Bytes.EMPTY);
+                    child.newTotalSupply(-1L);
+                    child.topicRunningHashVersion(0L);
+                    child.topicSequenceNumber(0L);
                     child.tokenTransferLists().clear();
+                    child.automaticTokenAssociations().clear();
+                    if (child.transferList().hasAccountAmounts()) {
+                        child.transferList().accountAmounts().clear();
+                    }
+                    child.paidStakingRewards().clear();
+                    child.contractCreateResult(null);
+                    child.scheduleRef(null);
+                    child.assessedCustomFees().clear();
+                    child.alias(Bytes.EMPTY);
+                    child.ethereumHash(Bytes.EMPTY);
+                    child.evmAddress(Bytes.EMPTY);
+
                     child.status(ResponseCodeEnum.REVERTED_SUCCESS);
                 }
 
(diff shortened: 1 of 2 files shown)


thenswan:
@tinker-michaelj can you please take a look?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -376,7 +383,31 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
                 followingChildRemoved = true;
             } else {
                 if (child.reversingBehavior() == ReversingBehavior.REVERSIBLE && SUCCESSES.contains(child.status())) {
+                    // clear "side effect" fields
+                    child.accountID(AccountID.DEFAULT);
```
tinker-michaelj, line 387: We want to use `null` for missing fields here, since the serialized protobuf will be different than just setting `AccountID.DEFAULT`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
```diff
@@ -376,7 +383,31 @@ public void revertChildrenOf(@NonNull final SingleTransactionRecordBuilderImpl r
...
+                    child.contractID(ContractID.DEFAULT);
+                    child.fileID(FileID.DEFAULT);
+                    child.tokenID(TokenID.DEFAULT);
+                    child.scheduleID(ScheduleID.DEFAULT);
+                    child.scheduledTransactionID(TransactionID.DEFAULT);
+                    child.serialNumbers().clear();
+                    child.topicRunningHash(Bytes.EMPTY);
+                    child.newTotalSupply(-1L);
```
tinker-michaelj, line 395: This should be `0`, right? Otherwise we would externalize a new total supply of `-1`, which doesn't seem correct.

tinker-michaelj (changes requested):
For fields other than `Bytes`, we want to reset values to `null` instead of `DEFAULT` since only `null` will omit the field from the serialized message.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/RecordListBuilder.java
thenswan: @tinker-michaelj got it, updated this one.

tinker-michaelj (approved):
LGTM, ty @thenswan !

PR #10404, opened 2023-12-09
fix: keep first cons time of current block in state

### LINKED ISSUE #10400: Add current consensus time of first block to `BlockInfo`
### Problem

Reconnecting nodes will ISS due to non-deterministic block start times unless the provisional consensus time of current block is in state.

### Solution

Keep provisional consensus time of current block in state.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #10400 by using `BlockInfo` to keep the provisional first consensus time of the current block in state.
 - Fixes genesis block number to be `0` not `1`.
 - Keeps existing unit tests passing despite their erroneous use of an ephemeral first consensus time tracker.


Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj 🙌

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/BlockRecordService.java
```diff
@@ -46,6 +47,8 @@ public final class BlockRecordService implements Service {
     /** The original hash, only used at genesis */
     private static final Bytes GENESIS_HASH = Bytes.wrap(new byte[48]);
 
+    public static final Timestamp EPOCH = new Timestamp(0, 0);
```
Neeharika-Sompalli, line 50: We could probably add some description here for this saying we use it only on GENESIS ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
```diff
@@ -151,60 +152,84 @@ public void close() {
...
-    /** {@inheritDoc} */
+    /**
+     * We need this to preserve unit test expectations written that assumed a bug in the original implementation,
+     * in which the first consensus time of the current block was not in state.
+     *
+     * @param consensusTime the consensus time at which to switch to the current block
+     */
+    @VisibleForTesting
+    public void switchBlockAt(@NonNull final Instant consensusTime) {
+        streamFileProducer.switchBlocks(lastBlockInfo.lastBlockNumber(), provisionalCurrentBlockNumber, consensusTime);
```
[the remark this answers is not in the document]
tinker-michaelj: Removed this field.

nickpoorman (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/BlockRecordService.java
tinker-michaelj: Ah oops missed this. 😞 Adding to a following PR now.

Neeharika-Sompalli (approved):
LGTM !

nathanklick (approved):

PR #10436, opened 2023-12-12
build: fix Jar artifacts determinism

jjohannes:
1. Do not use `git rev-parse --short` – it is not deterministic. Instead always use a 7 characters long short hash, which is the same behavior that we had in earlier versions of the build where we used JGit under the hood for this.
2. Track the full Java version as input to JavaCompile – otherwise class files compiled with e.g. 17.0.3 can be taken from the cache in a 17.0.9 run – although the `module-info.class` files differ.
3. Normalize line headings after the annotation processors wrote new files during compilation with platform specific ending. Better solution would be having these fixed:
    - https://github.com/google/auto/issues/1656
    - https://github.com/gradle/gradle/issues/27385  
5. No longer use Gradle's file rewrite functionality of `processResources` as it also uses system dependent line endings – https://github.com/gradle/gradle/issues/1151
6. Configure Git on Windows with `core.autocrlf false` – otherwise some `csv` and `json` files that look correct in the repo ended up having CRLF line endings in the Windows run


nathanklick (approved):

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (approved):
LGTM

david-bakin-sl:
Eeek.  #2 in the PR comment is kind of nasty!

iwsimon (commented):
LGTM

PR #10446, opened 2023-12-12
chore: make hapi tests non private

### LINKED ISSUE #10443: Make hapi test methods non-private
Hapi tests are configured similarly to JUnit tests in IntelliJ. With the latest version of IntelliJ, JUnit tests with `private` method modifiers can no longer be run directly from the IDE. This means that hapi tests can now only be run for the entire class, not one at a time, which makes development more time consuming.

Solution is to remove the `private` keyword from all hapi test methods. `private` can be changed to `final` because these methods should not be called from subclasses.

kimbor:
**Description**:
Change `private` method modifier to `final` for all hapi test methods.

**Related issue(s)**:

Fixes #10443


jsync-swirlds (approved):

povolev15:
you have some conflicts

povolev15 (dismissed):
LGTM

povolev15 (approved):

PR #10449, opened 2023-12-12
fix: Add check for empty inline initcode for contract creation

lukelee-sl:
**Description**:
Add a check to ensure that contracts with empty initcode cannot be created.  There was a check for checking empty smart contract byte code files but not for inline transactions.

**Checklist**

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
@@ -25,6 +25,7 @@ import static com.hedera.node.app.service.mono.state.EntityCreator.NO_CUSTOM_FEE
 import static com.hedera.node.app.service.mono.utils.EntityIdUtils.contractIdFromEvmAddress;
 import static com.hederahashgraph.api.proto.java.ContractCreateTransactionBody.InitcodeSourceCase.INITCODE;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.AUTORENEW_DURATION_NOT_IN_RANGE;
+import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.CONTRACT_BYTECODE_EMPTY;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.CONTRACT_FILE_EMPTY;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.CONTRACT_NEGATIVE_GAS;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.CONTRACT_NEGATIVE_VALUE;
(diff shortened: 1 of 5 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
```diff
@@ -349,6 +350,7 @@ public ResponseCodeEnum validate(final TransactionBody contractCreateTxn) {
 
     Bytes prepareCodeWithConstructorArguments(final ContractCreateTransactionBody op) {
         if (op.getInitcodeSourceCase() == INITCODE) {
+            validateFalse(op.getInitcode().isEmpty(), CONTRACT_BYTECODE_EMPTY);
```
david-bakin-sl, line 353: Can't decide whether it's a good idea or not a good idea that the error returned here isn't the same as the error returned at L366.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
```diff
@@ -295,6 +296,7 @@ private void assertValidCreation(@NonNull final ContractCreateTransactionBody bo
 
     private Bytes initcodeFor(@NonNull final ContractCreateTransactionBody body) {
         if (body.hasInitcode()) {
+            validateTrue(body.initcode().length() > 0, CONTRACT_BYTECODE_EMPTY);
```
david-bakin-sl, line 299: I actually like this `validateTrue` better than the previously seen `ContractCreateTransitionLogic` L353 `validateFalse`.  `validateFalse` just doesn't work, generally, with my brain: It is easier (for me) to immediately grok validate TRUE **not** empty than validate FALSE **is** empty.

I wasn't going to mention it up there, but since you haven't been _consistent_ I consider it fair game now.

david-bakin-sl (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
lukelee-sl: This bothered me too but if you noticed, I was being consistent with the check that returns `CONTRACT_FILE_EMPTY` within in the same file while not being consistent across mono vs modular code base.  I think this is the less of two evils as the mono code will eventually be deprecated?

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
Nana-EC: I think it's fine as this scenario is for the non file case right?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
Nana-EC: May post modularization we make them all validateTrue 🤔

Nana-EC (approved):
LG

Neeharika-Sompalli (approved):
Makes sense to me. Thanks @lukelee-sl !

PR #10453, opened 2023-12-12
feat: use address book to determine node id

### LINKED ISSUE #10450: use address book to determine node id at startup
Currently ServicesMain requires a node id as command-line argument and then it starts up a node with that id. Instead, node id should be optional. 

If node id is supplied, use the address book in config.txt to determine the associated IP address for the node and then compare that IP to the NIC card to make sure that it's local (aka this machine). If it is a local IP, start up a node with the supplied node id. If the IP is not local then hard exit.

If node id is not supplied, check every address in the address book. If there is exactly one address with an IP that is local to this machine, start up a node using the node id for that address. If there are zero or more than one addresses with local IPs then hard exit.

kimbor:
**Description**:
Change the way that node id is determined at startup.

**Related issue(s)**:

Fixes #10450


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
```diff
@@ -433,6 +433,12 @@ public static void checkNodesToRun(@NonNull final Collection<NodeId> nodesToRun)
                     new NodeAddressMismatchPayload(Network.getInternalIPAddress(), externalIpAddress));
             exitSystem(NODE_ADDRESS_MISMATCH);
         }
+        if (!allowMultipleNodes && nodesToRun.size() > 1) {
+            logger.error(
+                    EXCEPTION.getMarker(),
+                    "Multiple nodes are configured to run. Only one node can be started per java process.");
+            exitSystem(NODE_ADDRESS_MISMATCH);
+        }
```
nathanklick, line 441: This check needs to move up to the `ServicesMain` class after we determine if there is a node ID passed in the command line args.

jeromy-cannon (dismissed):

nathanklick (dismissed):

tinker-michaelj (changes requested):
This scans to me; but imo is too critical a change to commit without unit test coverage.

Neeharika-Sompalli:
> This scans to me; but imo is too critical a change to commit without unit test coverage.


➕  Agreed.

nathanklick:
> > This scans to me; but imo is too critical a change to commit without unit test coverage.
> 
> ➕ Agreed.

There's a limit to how much we can unit test the `main()` method, but we can cover the supporting methods. However, each and every HapiTest execution does exert testing of these methods (for at least the happy path).

netopyr (dismissed):
LGTM

jeromy-cannon (approved):

tinker-michaelj (approved):
LGTM, tyvm @kimbor !

PR #10490, opened 2023-12-13
fix: use the marker EXCEPTION for error level messages

### LINKED ISSUE #10488: Always log errors with the exception marker
In the platform test validators, error level log messages don't always fail the test, depending on which log marker is used. In order to be safe, we should use `EXCEPTION.getMarker()` for ALL platform level error messages that we want to fail a test. If we don't want a log message to fail a test, we should either configure the validator to ignore the log based on some other factor, or we should downgrade the log message to a warning or info level log.

cody-littley:
Closes #10488

### DIFF
--- a/platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java
+++ b/platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java
@@ -17,6 +17,7 @@
 package com.swirlds.demo.merkle.map.internal;
 
 import static com.swirlds.common.utility.CommonUtils.hex;
+import static com.swirlds.logging.legacy.LogMarker.EXCEPTION;
 import static com.swirlds.merkle.map.test.lifecycle.EntityType.Crypto;
 import static com.swirlds.merkle.map.test.lifecycle.EntityType.FCQ;
 import static com.swirlds.merkle.map.test.lifecycle.EntityType.NFT;
@@ -323,7 +324,7 @@ public final class ExpectedFCMFamilyImpl implements ExpectedFCMFamily {
                         entityType,
                         type.name());
             } catch (InterruptedException e) {
-                logger.error(ERROR, "", e);
+                logger.error(EXCEPTION.getMarker(), "", e);
             }
         }
     }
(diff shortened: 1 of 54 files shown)


edward-swirldslabs (commented):
How do we know this doesn't break any validators that were expecting the marker types that were replaced?

cody-littley:
@edward-swirldslabs 

> How do we know this doesn't break any validators that were expecting the marker types that were replaced?

Well, I will know if it breaks the abbreviated panel. I intend to monitor test results after this merges, and follow up with any fixes as needed.

platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java
```diff
@@ -761,7 +762,7 @@ private boolean creatingOnExistingEntities(
                         expectedValue.getEntityType());
             } else {
                 logger.error(
-                        PERFORM_ON_CREATE_DELETE,
```
anthony-swirldslabs, line 764: It seems that this marker is used here because this condition is acceptable under certain conditions, as documented at:

https://github.com/hashgraph/hedera-services/blob/4db779426c45f1d84eae528a147351a53d4915f3/platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java#L94-L99

I'm wondering if your change would break those cases or not. In an ideal world, the code would probably branch to detect those conditions and emit a warning instead of an error, I guess.

alittley (approved):

platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java
anthony-swirldslabs: @cody-littley : In an above comment, I see you're going to try and fix any potential failures after the change is deployed, which is fine.

In that case, if we don't want anyone to use non-standard markers, such as the one quoted above, why don't we delete them from the code-base to discourage their usage?

artemananiev (approved):

nathanklick (approved):

lpetrovic05 (approved):

platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/merkle/map/internal/ExpectedFCMFamilyImpl.java
cody-littley: I responded to this in the `#logging` slack channel. Let's discuss there.

PR #10495, opened 2023-12-13
feat: Make Consensus Services WritableTopicStore WritableKVState Key to be comparable. 

### LINKED ISSUE #10396: Make Consensus Services WritableTopicStore WritableKVState Key to be comparable.
Added TopicIdComparator to impose a deterministic ordering on collections of TopicID objects in WritableTopicStore.

Changed WritableKVStateBase constructor to take Comparator as a parameter to create a TreeMap for modifications to enforce the order of the Key. Updated all the related Writable/Readable classes for the related changes.

Added unit tests and updated all the unit tests.

Created issue #10531, #10532 when all services have made their WritableStore WritableKVState Key comparable 

Discussed this at Jan 9th modularization call, don't pass in comparator, use PBJ Comparable ID as SortedMap key for WritableKVStateBase.modifications. This has to wait for PBJ Key to implement Comparable.

iwsimon:
**Description**:

**Related issue(s)**:

Fixes #10396 

**Notes for reviewer**:

Discussed this at Jan 9th modularization call, don't pass in Comparator, use PBJ Comparable ID as SortedMap key for WritableKVStateBase.modifications. This has to wait for PBJ Key to implement Comparable.

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
```diff
@@ -170,6 +174,17 @@ public final Set<K> modifiedKeys() {
         return modifications.keySet();
     }
 
+    /** {@inheritDoc}
+     * To add the comparator to the underlying data source, we create a new TreeMap with the Comparator
+     * This should be called on each writable store constructor until all other types of Key Comparator are implemented.
+     * */
+    @Override
+    public final void updateComparator(@NonNull final Comparator<K> comparator) {
```
netopyr, line 182: I think the `Comparator` should be set during construction and never changed again.

iwsimon: Yes. Each store is responsible to update their comparator.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
```diff
@@ -170,6 +174,17 @@ public final Set<K> modifiedKeys() {
...
+        final var tmp = new TreeMap<K, V>(comparator);
+        tmp.putAll(modifications);
+        this.modifications = tmp;
```
netopyr, line 185: I am afraid I am missing something very essential here. If we update the `Comparator`, we are not changing the order of the elements in the data structure but the order of modifications, right? Why is this necessary? I assume that if any code is executed deterministically and single-threaded, the modifications should always be in the same order.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
netopyr: Sorry for not being clear. I would have expected that the `Comparator` is immutable. There is probably no valid use case where we want to change the `Comparator` later, or is there? If the `Comparator` is immutable, the code should reflect that.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
iwsimon: You have a linkedhashmap with put, remove options. If not sorted, how can you guarantee a deterministic order? We care about the writableKVState to be deterministic.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
```diff
@@ -28,15 +28,22 @@
...
      * Create a new StateBase.
      *
      * @param stateKey The state key. Cannot be null.
      */
-    protected WritableKVStateBase(@NonNull final String stateKey) {
+    // @todo("10153") make comparator nonnull and make modifications a SortedMap
+    protected WritableKVStateBase(@NonNull final String stateKey, @Nullable final Comparator<K> comparator) {
         super(stateKey);
+        if (comparator != null) modifications = new TreeMap<>(comparator);
+        else modifications = new LinkedHashMap<>();
```
jsync-swirlds, line 42: I think we might have missed on additional case here, but it isn't clear how to add it in (darn Erasure behaviors!).
Basically, if `K` is a `Comparable`, then we should use the TreeMap in that case as well, without an explicit comparator.  We don't have an instance of `K` to test here, however, so I'm unsure how to test the type (or if we even could).
Probably not worth worrying about at this time, but something perhaps to note as a future issue.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
iwsimon: This method has been removed. WritableTopicStore uses a private static final Comparator.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
iwsimon: A TreeMap is created in WritableKVStateBase constructor to enforce the Key order of modification when every service calls its' WritableXXXStore.put().

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/state/WritableKVStateBase.java
iwsimon: Agree we should use TreeMap when K is Comparable. Generics Type Erasure is hard to check here. We will make comparator @NonNull when all services have implemented their comparator key. That will cover all cases. We should stick to either Comparator or Comparable, not mix.

jsync-swirlds:
A PR merged to develop in PBJ that now allows Messages to be marked as "Comparable" (with some limitations).  This PR will need to be updated once the proto files are updated and the latest PBJ version is incorporated.

Neeharika-Sompalli:
Seems this is no more valid. Can we close this @iwsimon ?

iwsimon:
Closed as it's not needed any more.

PR #10499, opened 2023-12-13
fix: always set intrinsic gas even if no gas charges

### LINKED ISSUE #10498: (mod-service) always set intrinsic gas for zero-charge `GasCharges`
### Description

A `ContractCallLocal` is not charging intrinsic gas as expected.

### Steps to reproduce

Run `FileUpdateSuite.allUnusedGasIsRefundedIfSoConfigured()`

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #10498


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java
```diff
@@ -96,6 +95,8 @@ public void maybeRefundGiven(
      * Tries to charge gas for the given transaction based on the pre-fetched sender and relayer accounts,
      * within the given context and world updater.
      *
+     * <p>Even if there are gas charges, still returns the intrinsic gas cost of the transaction.
```
lukelee-sl, line 98: Is this comment accurate?  It implies to me that we will always return intrinsic gas

lukelee-sl (approved):
LGTM.  Just a question regarding a new comment

Neeharika-Sompalli (approved):
LGTM !

vtronkov (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java
tinker-michaelj: Ugh typo, thanks! Fixing on different branch.

PR #10523, opened 2023-12-14
fix: recordCache to commit added entries and implemented correctly the remove elements from the queue 

### LINKED ISSUE #10170: Memory leak in RecordCacheImpl
### Description

`RecordCacheImpl.histories` accumulates transaction histories which are not removed when transactions expire (see `removeExpiredTransactions()`). 

### Steps to reproduce

The bug was discovered when running the `NftTransferLoadTest` benchmark. The node would start seeing ZGC allocation stalls and stop being `ACTIVE`. 

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

Linux

### LINKED ISSUE #10403: RecordCacheImpl fails to remove expired transactions
### Description

The queue iterator in `RecordCacheImpl.removeExpiredTransactions()` is always empty.

### Steps to reproduce

Any test that submits transactions should do.

### Additional context

See also #10170 

### Hedera network

other

### Version

develop

### Operating system

Linux

povolev15:
[fix recordCache to commit added entries](fix: recordCache to commit added entries and implemented correctly the remove elements from the queue)

**Related issue(s)**:

Fixes #10403 
Fixes #10170


jsync-swirlds (approved):
One possible improvement, but not required.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -258,32 +267,33 @@ private void removeExpiredTransactions(
...
+                final var rec = entry.transactionRecordOrThrow();
+                final var txId = rec.transactionIDOrThrow();
+                // If the timestamp is before the current time, then it has expired
+                if (isBefore(txId.transactionValidStartOrThrow(), earliestValidState)) {
+                    // Remove from the histories
+                    queue.removeIf(TruePredicate.INSTANCE);
+                    // Remove from the payer to transaction index
+                    final var payerAccountId = txId.accountIDOrThrow(); // NOTE: Not accurate if the payer was the node
+                    final var transactionIDs =
+                            payerToTransactionIndex.computeIfAbsent(payerAccountId, ignored -> new HashSet<>());
```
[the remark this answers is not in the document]
povolev15: sure lets see if it works for Oleg and can be done on the second iteration.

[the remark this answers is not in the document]
jsync-swirlds: Nevermind, `computeIfAbsent` requires a function that takes a parameter, which won't fit the hash set `new` method handles.  `putIfAbsent` would work, but then we're creating objects we don't need if the entry does exist.  The most performant solution requires a bit more code and isn't as "clean".  If performance is important we can switch to that later.

jsync-swirlds (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -188,6 +193,10 @@ public void add(
             addToInMemoryCache(nodeId, payerAccountId, rec);
             queue.add(new TransactionRecordEntry(nodeId, payerAccountId, rec));
         }
+
+        if (states instanceof CommittableWritableStates committable) {
+            committable.commit();
+        }
```
rbair23, line 199: Instead of this, can we not let CommittableWritableStates extend from WritableStates and update the HederaState interface to return a ComittableWritableStates? Then we don't need the cast.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -257,33 +266,39 @@ private void removeExpiredTransactions(
...
+                final var txId = rec.transactionIDOrThrow();
+                // If the valid start time is before the earliest valid start, then it has expired
+                if (isBefore(txId.transactionValidStartOrThrow(), earliestValidStart)) {
+                    // Remove from the histories.  Note that all transactions are added to this map
+                    // keyed to the "user transaction" ID, so removing the entry here removes both
+                    // "parent" and "child" transaction records associated with that ID.
+                    histories.remove(txId);
+                    // remove from queue as well.  The queue only permits removing the current "HEAD",
+                    // but that should always be correct here.
+                    queue.removeIf(TruePredicate.INSTANCE);
```
rbair23, line 285: Will this empty everything out of the queue?

Was there something functionally incorrect in the previous code, or was it just wasteful creating the iterator? What is the purpose to this change?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: We tried having CommitableWritableStates extend WritableStates.  It caused another unrelated part of the codebase, which relies on the fact that they are _not_ cross-castable, to fail.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: There is no iterator for FCQueue, it can only be removed from the head, and nowhere else.

hedera-node/hedera-network-admin-service-impl/src/test/java/com/hedera/node/app/service/networkadmin/impl/test/handlers/NetworkAdminHandlerTestBase.java
```diff
@@ -207,7 +207,7 @@ protected void refreshRecordCache() {
         lenient().when(wsa.getHederaState()).thenReturn(state);
         lenient().when(props.getConfiguration()).thenReturn(versionedConfig);
         lenient().when(versionedConfig.getConfigData(HederaConfig.class)).thenReturn(hederaConfig);
-        lenient().when(hederaConfig.transactionMaxValidDuration()).thenReturn(180L);
+        lenient().when(hederaConfig.transactionMaxValidDuration()).thenReturn(123456789999L);
```
mhess-swl, line 210: The typical configured value is 180 (3 min). Why does this value have to be increased so dramatically in order to pass?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
mhess-swl: I ran into this yesterday, that the `remove()` operation wasn't implemented. I'm really curious if purging transactions has been working at all.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -178,8 +182,9 @@ public void add(
 
         // To avoid having a background thread cleaning out this queue, we spend a little time when adding to the queue
         // to also remove from the queue any transactions that have expired.
-        final var queue = getQueue();
-        final var firstRecord = transactionRecords.get(0);
+        final WritableStates states = getWritableState();
+        final WritableQueueState<TransactionRecordEntry> queue = states.getQueue(TXN_RECORD_QUEUE);
+        final SingleTransactionRecord firstRecord = transactionRecords.get(0);
```
mhess-swl, line 187: Nit: is the queue guaranteed to always have at least one record?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -257,33 +266,39 @@ private void removeExpiredTransactions(
...
+                    // but that should always be correct here.
+                    queue.removeIf(TruePredicate.INSTANCE);
+                    // Remove from the payer to transaction index
+                    final var payerAccountId = txId.accountIDOrThrow(); // NOTE: Not accurate if the payer was the node
+                    final var transactionIDs =
+                            payerToTransactionIndex.computeIfAbsent(payerAccountId, ignored -> new HashSet<>());
+                    transactionIDs.remove(txId);
+                    if (transactionIDs.isEmpty()) {
+                        payerToTransactionIndex.remove(payerAccountId);
+                    }
```
mhess-swl, line 293: The loop updates `histories`, the queue itself, and `payerToTransactionIndex`, but does nothing with the deduplication cache. Should the transaction ID also be removed from there?

hedera-node/hedera-network-admin-service-impl/src/test/java/com/hedera/node/app/service/networkadmin/impl/test/handlers/NetworkAdminHandlerTestBase.java
jsync-swirlds: It didn't.  Lev just chose a value that could never be exceeded as the intention for this particular test value is to not expire events.
It does have to be fairly excessive as there is a bit of "current time" impact that needed to be accommodated, but setting the value the same as the mock consensus time is probably fine.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: If it was used, probably.  At the moment the _use_ of the deduplication cache is not implemented (and it's not certain it needs to be; it might be removed in a future PR).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: I believe that at this point the queue state will always return something from get(0), as it was tested non-empty before getting here (line 178).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
jsync-swirlds: Transactions were never removed from anything before this PR. The queue was always empty, so we never entered the remove loop, because we weren't committing the state.

The two bugs that this PR fixes are, paraphrased, "queue in state is always empty" and "histories leaks memory because nothing is ever removed from that map".

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
mhess-swl: Missed that, thanks

mhess-swl (approved):

tinker-michaelj (approved):
LGTM!

Issue #10084, opened 2023-11-22
Implement migration of state from mono to modular service


mhess-swl:
The format of bytes stored in state has changed in the new modular service. Therefore a migration from the mono format to the modular format will need to execute when starting up the modular app from a mono state. 

This migration is critical because we will not be able to do any differential testing, as that testing relies on replaying events from a common starting point.

mhess-swl:
Work for this ticket will generally start with the `migration-experiment` branch. 

There's an initial stab at the contract store migration in the `migration-experiment.dsb` branch.

mhess-swl:
PR for contract migrations is #10159

mhess-swl:
Quick update on this ticket:

With a couple of clever tricks (probably not acceptable in prod environments, we’ll need to revisit), the modular node will now load the mono ServicesState as a modular MerkleHederaState. We now have a working prototype that can read and write to a 'hybrid' state containing _both_ sets of state keys, with a few pieces of state migrating successfully (accounts, contracts, and consensus topics). 

Our tentative plan is as follows:
* Implement the remaining migration pieces
* Work with @david-bakin-sl to successfully run a manual differential test (further action items will definitely be needed after this step is complete)
* Verify that a migration can be run, the state file written to disk, the node shutdown, and then started again with the hybrid state (or the mono state). In other words, make sure the hybrid state file isn’t corrupted somehow
* Attempt deletion of the old state keys
* At some point, figure out how to load ServicesState without spoofing the class ID (the aforementioned clever tricks)

mhess-swl:
I've pushed up a change that allows the node to start correctly (to the same branch, `migration-experiment_bbm-v1`). 

The only thing left to do for this proof-of-concept is to handle the `ExpirableTxnRecord`s in the FCQueue. In our discussion today, because it's possible that the migration will take less than the configured expiry time of these records (3 minutes), we decided that we want to take all the transaction IDs of all the records in the mono state and dump them into the modular code's deduplication cache. That way we can avoid a node submitting a duplicate transaction

mhess-swl:
I've pushed up a rough migration for the expiring records to the migration branch. It's possible we'll need to tweak the times so the node doesn't immediately purge the queued transactions from the mono state (see issue #10575).

mhess-swl:
For reference, this ticket is closely related to #10571

mhess-swl:
[...]

PR #10526, opened 2023-12-14
fix: Eliminate extra service registration

mhess-swl:
There's an unnecessary round of registration happening for the `EntityIdService` in the `OrderedServiceMigrator`. Since the registration is already invoked in the `Hedera` class (along with all other services), there's no need to do so here. 

Discovered while implementing #10084


tinker-michaelj (approved):
Makes sense to me, thanks @mhess-swl !

Neeharika-Sompalli (approved):

PR #10534, opened 2023-12-15
fix: use `haltResult()` instead of throwing `HandleException` in `HtsSystemContract`

### LINKED ISSUE #9962: Use exceptional halt instead of re-thrown exception on `HtsSystemContract` resource exhaustion
### Problem

It will be unsafe and perhaps highly problematic to propagate a `HandleException` from [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java#L73).

### Solution

Switch this to an exceptional halt.

### Alternatives

_No response_

petreze:
**Description**:
This PR changes the approach in [HtsSystemContract](https://github.com/hashgraph/hedera-services/blob/09962-fix-handleexception-htssystemcontract/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java#L131) when `HandleException` is caught and instead of throwing the exception, we halt the result via `haltResult()`

**Related issue(s)**:

Fixes #9962 

**Notes for reviewer**:
Please check if this is the most suitable approach handling it

The following is tested while executing suite test `cannotUseMoreThanChildContractLimit()` in `ContractDeleteSuite` and the test should pass after we remove the throwing of `HandleException`. A successful execution was achieved also by replacing the throwing with `return revertResult(handleException.getStatus(), frame.getRemainingGas());` but I'm not sure if this would be a proper approach as inside the issue it is stated that this should be done by using exceptional halt

**Checklist**


tinker-michaelj (approved):
Makes sense to me, tyvm @petreze !

agadzhalov (approved):
LGTM

PR #10551, opened 2023-12-15
fix: Fix and enable all Schedule HapiTests

### LINKED ISSUE #9970: Determine solution for Monoservice failure-resistant key gathering
Monoservice gathers keys required by a transaction even when the transaction body is invalid or corrupted.  This causes scheduled transactions to proceed until the child is executed, at which time it fails with an invalid transaction body.  There are existing end to end tests (e.g. `ScheduleExecutionSpecs.scheduledBurnFailsWithInvalidTxBody()`) that require this behavior.  These include at least:
- [x] scheduledBurnFailsWithInvalidTxBody
- [x] scheduledMintFailsWithInvalidTxBody
- [x] scheduledBurnForUniqueFailsWithInvalidAmount
- [x] scheduledMintFailsWithInvalidAmount
- [x] scheduledXferFailingWithNonNetZeroTokenTransferPaysServiceFeeButNoImpact
- [x] scheduledXferFailingWithRepeatedTokenIdPaysServiceFeeButNoImpact
- [x] scheduledXferFailingWithEmptyTokenTransferAccountAmountsPaysServiceFeeButNoImpact
- [x] executionWithInvalidAccountAmountsFails

Modular services do not gather keys for invalid inputs; if the transaction is not valid in pureChecks, the gathering of keys fails.  This makes passing the existing end to end test impossible, and may prevent matching monoservice behavior in differential testing impossible as well, because we will fail a create transaction that would succeed for monoservice (and later fail _only_ the execution record, even the ScheduleSign transaction will succeed; only the actual execution of the scheduled transaction fails).

We need to determine how this will be handled in modular services for differential testing, and whether we can/should have an alternate mechanism to match monoservice behavior for key gathering.

### LINKED ISSUE #9976: Fix ScheduleExecutionSpecs methods that fail with invalid transfer list
Some specs in ScheduleExecutionSpecs still fail with an invalid transfer list.  We need to find out what causes this and fix it.

- executionWithDefaultPayerWorks
- executionWithCustomPayerWorks
- executionWithCustomPayerAndAdminKeyWorks
- executionWithCustomPayerWhoSignsAtCreationAsPayerWorks

### LINKED ISSUE #9977: Fix ScheduleExecutionSpecs methods that fail with wrong balance
Some specs in ScheduleExecutionSpecs still fail with the wrong balance.  We need to find out what causes this and fix it.  This _appears_ to be related to fee charging not causing child transactions to fail (the parent payer is charged instead?).

- executionWithDefaultPayerButNoFundsFails
- executionWithCustomPayerButNoFundsFails
- executionWithDefaultPayerButAccountDeletedFails
- executionWithCustomPayerButAccountDeletedFails

povolev15:
fix: schedule test, fee children calculation and passedVerification)

**Related issue(s)**:

- Fixes #9976 
- Fixes #9977 
- Fixes #9970

**Notes for reviewer**:

**Checklist**


tinker-michaelj (approved):
LGTM!

PR #10603, opened 2023-12-21
chore: Disable reconnect test in CI 

Neeharika-Sompalli:
Until a new tag is added for reconnect test to be run in a different environment, disable it running under Misc tag in CI. This could probably be causing the hanging for Misc tests


tinker-michaelj (approved):
Soo glad you spotted this @Neeharika-Sompalli , tyvm !!

PR #10613, opened 2023-12-21
chore: add dependency between test jobs and spotless

### LINKED ISSUE #10608: Update PR workflow to only trigger test job after build and spot less passed
### Problem

Every time a minor change to PR due to spotlessApply, all the PR tests have to be restarted or interrupted, kind of wasteful.

### Solution

updated the workflow to only start test jobs after build and spotless scan are successful.

### Alternatives

_No response_

JeffreyDallas:
**Description**:

Closes #10608


JeffreyDallas:
no longer needed

PR #10614, opened 2023-12-21
chore: 10561 chage withConverter signature

### LINKED ISSUE #10561: Extend ConfigurationBuilder to Include converter type in method signature when adding converters
With the introduction of default support for Enum conversion in #10535, the need to force each module to add a converter specifically for the Enums has been eliminated. However, to increase flexibility and clarity of the ConfigurationBuilder, we should extend the withConverter method to include the type of the converter in its signature.

**Current Signature**
```
ConfigurationBuilder withConverter(ConfigConverter<?> converter)
```
**Proposed Signature**
```
<T> ConfigurationBuilder withConverter(Class<T> type, ConfigConverter<T> converter)
```
**Impact**
by allowing us to specify the type of converter explicitly, we can remove the mechanism in `ConverterService` that aimed to discover the type of the converter using reflection, providing more flexibility when configuring converters.

mxtartaglia-sl:
**Description**:
To increase flexibility and clarity of the ConfigurationBuilder, we should extend the `withConverter` method to include the type of the converting class in its signature.
By allowing us to specify the type of converter explicitly, we can remove the mechanism in `ConverterService` that aimed to discover the type of the converter using reflection, providing more flexibility when configuring converters.

**Related issue(s)**:

Fixes #10561


cody-littley (approved):

Neeharika-Sompalli (approved):
LGTM !

nathanklick (dismissed):

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
```diff
@@ -62,23 +62,15 @@ public interface ConfigurationBuilder {
...
-    /**
-     * Adds converters (see {@link ConfigConverter}). If this method is called after the config has been created (see
-     * {@link #build()}) a {@link IllegalStateException} will be thrown.
-     *
-     * @param converters the converters that should be used for the configuration
-     * @return the builder instance (useful for fluent API)
-     * @throws IllegalStateException if this method is called after the config has been created
-     */
-    @NonNull
-    ConfigurationBuilder withConverters(@NonNull final ConfigConverter<?>... converters) throws IllegalStateException;
```
hendrikebbers, line 81: Why have you removed methods? adding the new one makes sense but I do not get why you remove methods.

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConverterService.java
```diff
@@ -98,24 +96,6 @@ class ConverterService implements ConfigLifecycle {
         this.converters = new ConcurrentHashMap<>();
     }
 
-    @NonNull
-    private <T, C extends ConfigConverter<T>> Class<T> getConverterType(@NonNull final Class<C> converterClass) {
```
hendrikebbers, line 102: I would have expected that the given functionality will stay in place for now and the new one is added in addition.

mxtartaglia-sl: Yes. The impact would have been more contained if I added another method to the existing alternatives. Still, there would be a slight difference in the behavior of the two methods. While the new one accepted lambda objects (maybe not important for us), the existing one would fail on runtime for the same case, which seemed inconsistent (can be documented). Also, it seemed more straightforward to use if you don't have to reflect on which signature is better, and although less flexible, by only providing the option of having the types explicitly declared in the signature seemed less prone to errors.
 Happy to add the other signatures back.

mxtartaglia-sl: As discussed with @hendrikebbers , I will recover the removed methods add the `@deprecated` annotation and create a second ticket for its future removal and documentation changes. The reasoning behind this:  we want to make sure to have a workflow compatible with cases where we have external clients for this functionality, and an all-or-nothing approach is not suitable for the future

Neeharika-Sompalli (approved):

cody-littley (dismissed):

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
mxtartaglia-sl: added them back and marked as deprecated

nathanklick (dismissed):

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
```diff
@@ -76,10 +79,27 @@ public interface ConfigurationBuilder {
      * @param converters the converters that should be used for the configuration
      * @return the builder instance (useful for fluent API)
      * @throws IllegalStateException if this method is called after the config has been created
+     *
+     * @deprecated Use {@link ConfigurationBuilder#withConverter(Class, ConfigConverter)}
      */
     @NonNull
+    @Deprecated
```
hendrikebbers, line 86:
```suggestion
    @Deprecated(forRemoval=true)
```

platform-sdk/swirlds-config-api/src/main/java/com/swirlds/config/api/ConfigurationBuilder.java
```diff
@@ -65,8 +65,11 @@ public interface ConfigurationBuilder {
      * @param converter the converter that should be used for the configuration
      * @return the builder instance (useful for fluent API)
      * @throws IllegalStateException if this method is called after the config has been created
+     *
+     * @deprecated Use {@link ConfigurationBuilder#withConverter(Class, ConfigConverter)}
      */
     @NonNull
+    @Deprecated
```
hendrikebbers, line 72:
```suggestion
    @Deprecated(forRemoval=true)
```

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -132,26 +133,45 @@ private void addConfigSource(@NonNull final ConfigSource configSource) {
...
     }
 
+    /**
+     * @param converter the converter that should be used for the configuration
+     * @deprecated Use {@link ConfigurationBuilderImpl#withConverter(Class, ConfigConverter)}
+     */
     @NonNull
     @Override
+    @SuppressWarnings({"unchecked"})
+    @Deprecated
```
hendrikebbers, line 143:
```suggestion
    @Deprecated(forRemoval=true)
```

platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderImpl.java
```diff
@@ -132,26 +133,45 @@ private void addConfigSource(@NonNull final ConfigSource configSource) {
...
         return this;
     }
 
+    /**
+     * @param converters the converters that should be used for the configuration
+     * @deprecated Use {@link ConfigurationBuilderImpl#withConverter(Class, ConfigConverter)}
+     */
     @NonNull
     @Override
+    @Deprecated
```
hendrikebbers, line 155:
```suggestion
    @Deprecated(forRemoval=true)
```

Neeharika-Sompalli (approved):
LGTM !

cody-littley (approved):

nathanklick (approved):

hendrikebbers (approved):

Issue #9423, opened 2023-10-23
Conditional records of hollow account creation via internal value transfer.


tinker-michaelj:
Validate the below happy path scenarios:
 - [x] An `EthereumTransaction` sending value to an absent `to` address (outside long-zero range) generates a preceding child with a synthetic `CryptoCreate` and `TransactionRecord` identical that created by mono-service when it resolves to `SUCCESS`: https://github.com/hashgraph/hedera-services/pull/10539
 - [x] A `ContractCreate` with an internal message call to a `MessageFrame` with absent `receiver` address (outside long-zero range) with sufficient gas generates a preceding child with a synthetic `CryptoCreate` and `TransactionRecord` identical that created by mono-service: https://github.com/hashgraph/hedera-services/pull/11163

And validate the below failure scenarios:
 - [x] An `EthereumTransaction` sending value to an absent `to` address (outside long-zero range) generates no preceding child when it resolves to `INSUFFICIENT_GAS`: https://github.com/hashgraph/hedera-services/pull/10539
 - [x] A `ContractCall` with 4 internal message calls to `MessageFrame`s with absent `receiver` addresses (outside long-zero range) with sufficient gas generates no preceding children because the fourth message call reverts with a `MAX_CHILD_RECORDS_EXCEEDED` reason: https://github.com/hashgraph/hedera-services/pull/10631
 - [x] A `ContractCall` that performs a valid lazy creation in a frame at stack depth 2, and then reverts that frame, generates no preceding children even when the frame at stack depth 1 deals with its child's reversion via a `try`-`catch`: https://github.com/hashgraph/hedera-services/pull/11131

vtronkov:
All happy path scenarios are done

PR #10631, opened 2023-12-22
fix: conditional records hollow account creation via internal transfer with max child records exceeded

MiroslavGatsanoga:
**Description**:

Verify that `evmLazyCreateViaSolidityCallTooManyCreatesFails` spec from `LeakyContractTestsSuite` run against modularized codebase produces the same records as when being run against mono.

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/9423
Fixes https://github.com/hashgraph/hedera-services/issues/10316

**Notes for reviewer**:

- Fuzzy matching only `evmLazyCreateViaSolidityCallTooManyCreatesFails` provided to be mor challenging than initially estimated so I decided to enable the fuzzy matching for the whole `LeakyContractTestsSuite`.
- Additionally I added order annotations for the specs in `LeakyContractTestsSuite` & `LeakyCryptoTestsSuite` so that when run as HAPI tests (i.e. against modularized) they will run in the same order as against mono.

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
```diff
@@ -433,6 +434,18 @@ private void fuzzyMatch(
...
 
+            final boolean isFileUpdate = FileUpdateTransactionBody.class.isAssignableFrom(expectedType);
+            final boolean isForFile121 =
+                    FileID.newBuilder().setFileNum(121).build().equals(expectedField.getValue());
+            final boolean isForFile111 =
+                    FileID.newBuilder().setFileNum(111).build().equals(expectedField.getValue());
+            // skip matching FileUpdate for properties file 121 or fee schedules file 111 since depending on if leaky
+            // tests are run before or not,
+            // the FileUpdate contents can be different
+            if (isFileUpdate && (isForFile121 || isForFile111)) {
```
tinker-michaelj, line 445: Can't we just updated `ParsedItem.isPropertyOverride()` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/domain/ParsedItem.java#L55) to also ignore fee schedule updates?

Actually, should we generalize that method to `isSpecialFileChange()` and do something like,
```
        return itemRecord.getReceipt().getStatus() == SUCCESS
                && (itemBody.hasFileUpdate() || itemBody.hasFileAppend())
                && PROPERTIES_FILE_ID.equals(itemBody.getFileUpdate().getFileID()) || FEE_SCHEDULE_FILE_ID.equals(itemBody.getFileUpdate().getFileID());

```

MiroslavGatsanoga: Thanks Michael, I added it [as such](https://github.com/hashgraph/hedera-services/pull/10631/commits/cec1288baaabebe8c06520e55eace57c9dd3ae22#diff-fa1ee390421aaf33ca4cb842c66f37462f4f660326f6c3c8ed7ef575f1410101R55-R70).

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -418,7 +418,8 @@ private void dispatchAndMarkCreation(
                 .contractCreateResult(ContractFunctionResult.newBuilder()
                         .contractID(contractId)
                         .evmAddress(evmAddress)
-                        .build());
+                        .build())
+                .accountID(null); // in mono, we don't have accountID set for contract creations
```
tinker-michaelj, line 422: Don't we already do this [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java#L808)?

tinker-michaelj (approved):
LGTM, ty @MiroslavGatsanoga  and @vtronkov !

PR #10634, opened 2023-12-22
chore: Misc contract cleanup and fixes

### LINKED ISSUE #10599: Cleanup code as suggested in review of #10230
### Problem

The fixes in #10230 used extra lines of code and duplicated a little bit of logic. We would like to,
 - [x] Reuse `withTinybarGasFee()` instead of adding a new gas fee builder method.
 - [ ] Customize the `ContractCreateTransactionBody` implied by an `EthereumTransaction` when first creating the synthetic body.
 - [x] Validate the sender nonce at the first point-of-contract in `TransactionProcessor`.

### Solution

Do this as suggested in comments,
 - https://github.com/hashgraph/hedera-services/pull/10230/files#r1433084701
 - https://github.com/hashgraph/hedera-services/pull/10230/files#r1433141865
 - https://github.com/hashgraph/hedera-services/pull/10230/files#r1433107791

### Alternatives

_No response_

Neeharika-Sompalli:
**Description**:
 - Closes #10599 by refactoring as suggested.
 - Adds `AbortException` to eliminate duplication of computing the effective `senderId`; now the exception thrown from an aborted `processTransaction()` call includes the `senderId` from `computeInvolvedParties()`.
 - Since we no longer need to use the `feesOnlyUpdater` after an abort, this also lets us easily preserve the HAPI fees charged by the workflow by just committing the root updater after an abort.
    * Avoids partial gas charges by refactoring `CustomGasCharging` to only charge _any_ fee if _all_ fees are payable.
 - Removes unreachable `try`/`catch (ResourceExhaustedException)` around `runToCompletion()`.
 - For mono-service fidelity, allows long-zero addresses to be used in top-level HAPI operations even when the referenced contract has an EVM address outside the long-zero subspace (accomplished via `asPriorityId()` calls in `HevmTransactionFactory`).
 - Enables last two specs in `EthereumSuite`.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
@@ -56,7 +56,6 @@ import com.hedera.node.app.hapi.utils.ethereum.EthTxData;
 import com.hedera.node.app.records.BlockRecordManager;
 import com.hedera.node.app.service.token.ReadableAccountStore;
 import com.hedera.node.app.service.token.api.TokenServiceApi;
-import com.hedera.node.app.service.token.impl.WritableAccountStore;
 import com.hedera.node.app.service.token.records.ChildRecordFinalizer;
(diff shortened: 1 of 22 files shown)


tinker-michaelj (approved):
LGTM

PR #10636, opened 2023-12-22
chore: cleanup while itemizing remaining disabled specs

### LINKED ISSUE #10635: Review and itemize all `HapiSpec` not yet enabled as `@HapiTest`
### Problem

We need a single-pane-of-glass view of all not-yet-enabled `HapiSpec`'s.

### Solution

Remove all `HapiSpec`s that shouldn't be enabled from `@HapiTestSuite`'s; create an issue with a single tasklist for all remaining specs.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #10635 
 - Besides the 86 specs now itemized in https://github.com/hashgraph/hedera-services/issues/10637,
    * Uses `CONTRACT_EXECUTION_EXCEPTION` as failures status for top-level lazy creation of a mirror address.
    * Fixes and enables `failsForInsufficientPayment()` by using a civilian payer.
    * Cleans up logging for `NftTransferSuite` but does not fix it.
    * Fixes `receiptUnavailableAfterCacheTtl()` by adding a transaction to trigger receipt expiration; but does not fix it.
    * Removes obsolete spec `recordAvailableInPayerState()` (assumed threshold records exist).
    * Fixes and enables `deletedAccountRecordsUnavailableAfterTtl()` to have correct expectations as `recordsStillQueryableWithDeletedPayerId()`.
    * Fixes and enables `recordUnavailableIfRejectedInPrecheck()` by reflecting that record existence is only checked for `ANSWER_ONLY` in mod-service.
    * Adds `@BddMethodIsNotATest` to several `HapiSpec` "factory methods" that are not directly run against mod-service.
    * Enables remaining `HelloWorldEthereumSuite` specs as they pass now.
    * Fixes and enables `optimisticSpecialFileUpdate()` by only checking the `0.0.150` hash and not its contents.
    * Tries to stabilize `congestionMultipliersRefreshOnPropertyUpdate()` by using a longer congestion period and just checking if _any_ submitted transaction was charged the expected multiplier.


tinker-michaelj:
Will be merged via #10640

PR #10149, opened 2023-11-28
feat: Stop returning token associations and token balances data in queries as per HIP-367

Neeharika-Sompalli:
Fixes #10089 

- Stop returning token associations and token balances data in queries as per HIP-367
The queries modified are 
     - GetAccountBalance
     - GetAccountInfo
     - GetContractInfo
- Changes are done in both mono and modular code
- Changed test-clients to do a `getAccountDetails` query for testing if there are any assertions related to token relationships or token balances

PR #10638, opened 2023-12-22
fix: Add a feature flag for token balances and token relationships and enable them

Neeharika-Sompalli:
Reverts this PR https://github.com/hashgraph/hedera-services/pull/10149

Adds a feature flag `tokens.balancesInQueries.enabled` to enable/disable the token balances and token relationships information to be returned in queries

### DIFF
--- a/hedera-node/configuration/dev/bootstrap.properties
+++ b/hedera-node/configuration/dev/bootstrap.properties
@@ -3,6 +3,7 @@ accounts.storeOnDisk=true
 tokens.storeRelsOnDisk=true
 tokens.nfts.useVirtualMerkle=true
 accounts.blocklist.enabled=false
+tokens.balancesInQueries.enabled=true
 accounts.blocklist.path=
 
 contracts.knownBlockHash=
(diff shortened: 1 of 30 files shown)


tinker-michaelj (approved):

tinker-michaelj:
Node software changes LGTM 💯 , but [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/queries/contract/HapiGetContractInfo.java#L161) we will end up with a _repeated_ list of `tokenRelationships` when the feature flag is on and the `GetContractInfo` query is returning `tokenRelationships` again. 

Instead we need,
```
            actualInfo = actualInfo.toBuilder()
                    .clearTokenRelationships()
                    .addAllTokenRelationships(actualTokenRels)
                    .build();
```
to ensure our expectations use only the `getAccountDetails` response no matter the feature flag.

Neeharika-Sompalli:
> actualInfo = actualInfo.toBuilder()
>                     .clearTokenRelationships()
>                     .addAllTokenRelationships(actualTokenRels)
>                     .build();

Ah yes. Thank you very much @tinker-michaelj 🙏

tinker-michaelj (approved):
LGTM!

rbair23 (approved):

PR #10660, opened 2023-12-28
feat: add setting for birth round ancient threshold

### LINKED ISSUE #10659: Add setting for birth round ancient threshold
Add a setting that will control the migration to birth round based ancient thresholds.

cody-littley:
Closes #10659


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/eventhandling/EventConfig.java
```diff
@@ -73,4 +76,5 @@ public record EventConfig(
         @ConfigProperty(defaultValue = "/opt/hgcapp/eventsStreams") String eventsLogDir,
         @ConfigProperty(defaultValue = "true") boolean enableEventStreaming,
         @ConfigProperty(defaultValue = "8") int prehandlePoolSize,
-        @ConfigProperty(defaultValue = "false") boolean useLegacyIntake) {}
+        @ConfigProperty(defaultValue = "false") boolean useLegacyIntake,
+        @ConfigProperty(defaultValue = "false") boolean useBirthRoundAncientThreshold) {}
```
edward-swirldslabs, line 80: I'm wondering if this flag will be necessary.  

Argument for Not Necessary: 
* Everywhere except for things based on Sequence data structures with `shiftWindow` method will use the `NonAncientEventWindow.isAncient()` methods.   
* To flip the switch in 1 commit, modify the `isAncient()` implementation and modify the sequence data structure implementations to use `minRoundNonAncient` instead of `minGenNonAncient`. 
* I thought we were trying to avoid simultaneous support in code for both definitions of ancient.    If you use this flag then does our code need to be able to support both at the same time? Or are you going to do a complete re-implementation of PCES and the sequence based data structures and swap based on this flag?

edward-swirldslabs (approved):

PR #10663, opened 2023-12-28
chore: reduce static config use

### LINKED ISSUE #10661: Reduce use of static configuration: iteration 1
It turns out there is still a lot of use of static configuration. Reduce this usage, but limit the number of changes to an amount that is small enough for review.

cody-littley:
Closes #10661


hedera-node/hedera-config/src/testFixtures/java/com/hedera/node/config/testfixtures/HederaTestConfigBuilder.java
```diff
@@ -125,7 +124,6 @@ public static TestConfigBuilder create() {
                 .withConfigDataType(MetricsConfig.class)
                 .withConfigDataType(PrometheusConfig.class)
                 .withConfigDataType(PlatformStatusConfig.class)
-                .withConfigDataType(FCHashMapConfig.class)
```
hendrikebbers, line 128: Why do you remove that class from services?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/singleton/ConfigurationHolder.java
```diff
@@ -34,14 +33,14 @@
...
- * done once in the browser or for unit tests and nowhere else!
+ * config sources. Therefore the default values will be used for all values of config data records (see
+ * {@link ConfigData}). A new config can be set by calling the {@link #setConfiguration(Configuration)} method. That
+ * should be done once in the browser or for unit tests and nowhere else!
  *
  * @deprecated Will be removed once we have the Platform Context in place or all mentioned problems have been refactored
  */
 @Deprecated(forRemoval = true)
-public final class ConfigurationHolder implements Supplier<Configuration> {
+public final class ConfigurationHolder {
```
hendrikebbers, line 43: No idea why you did the changes in this class. From my point of view that creates no benefit but makes the PR bigger.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
```diff
@@ -26,6 +32,19 @@ public interface Cryptography {
...
 
+    /**
+     * Creates a new {@link Cryptography} instance using the given configuration.
+     *
+     * @param configuration
+     * 		the configuration to use
+     * @return a new {@link Cryptography} instance
+     */
+    @NonNull
+    static Cryptography create(@NonNull final Configuration configuration) {
```
hendrikebbers, line 43: Can we create a CryptographyFactory class for that method and directly annotate it with deprecated?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/CryptographyHolder.java
```diff
@@ -48,30 +54,44 @@ private CryptographyHolder() {
...
         try (final Locked ignored = lock.lock()) {
-            cryptography = null;
+            CryptographyHolder.cryptography = cryptography;
         }
     }
 
+    /**
+     * Setup cryptography. Intended only to be invoked by unit tests.
+     */
+    private static void defaultTestSetup() {
```
hendrikebbers, line 70: The method is called by the get() method. Therefore it is not called by unit test but by logic that is called everywhere. Let's name it init/reset and add info to the javadoc that it is only needed to make it useable in tests. 
Question: How long will the holder stay? Maybe we introduce a Unit extension (see https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/base/test-support/test-support.md) that resets Config Holder & Crypto Holder.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/CryptographyHolder.java
```diff
@@ -48,30 +54,44 @@ private CryptographyHolder() {
...
+    @NonNull
     public static Cryptography get() {
-        if (cryptography == null) {
-            try (final Locked ignored = lock.lock()) {
-                if (cryptography == null) {
-                    cryptography = new CryptoEngine(
-                            getStaticThreadManager(),
-                            ConfigurationHolder.getInstance().get().getConfigData(CryptoConfig.class));
-                }
+        try (final Locked ignored = lock.lock()) {
```
hendrikebbers, line 90: you do not need to lock if cryptography != null

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/MerkleCryptoFactory.java
```diff
@@ -46,22 +51,43 @@ public class MerkleCryptoFactory {
 
     private MerkleCryptoFactory() {}
 
+    /**
+     * Setup cryptography. Intended only to be invoked by unit tests.
+     */
+    private static void defaultTestSetup() {
```
hendrikebbers, line 57: Again I think it is bad practice to define runtime code as "This is test code". We should move that to test fixtures and create a JUnit extension. While that can not all be part of one PR we should create issues for that but not name the methods like that.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/MerkleCryptography.java
```diff
@@ -37,14 +42,26 @@
...
-     * Computes a cryptographic hash for the {@link MerkleInternal} instance. The hash is passed to the object
-     * by calling {@link Hashable#setHash(Hash)}. Convenience method that defaults to {@link DigestType#SHA_384} message
+     * Create a new merkle crypto engine.
+     *
+     * @param configuration the configuration
+     * @param cryptography  the cryptography
+     * @return the new merkle crypto engine
+     */
+    @NonNull
+    static MerkleCryptography create(
```
hendrikebbers, line 52: Let's create a factory class

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/TeachingSynchronizer.java
```diff
@@ -91,24 +92,18 @@ public class TeachingSynchronizer {
...
+     * @param in              the input stream
+     * @param out             the output stream
+     * @param root            the root of the tree
+     * @param breakConnection a method that breaks the connection. Used iff an exception is encountered. Prevents
+     *                        deadlock if there is a thread stuck on a blocking IO operation that will never finish due
+     *                        to a failure.
+     * @param reconnectConfig reconnect configuration from platform
      */
     public TeachingSynchronizer(
+            @NonNull final PlatformContext platformContext,
```
hendrikebbers, line 106: Let's try to avoid using PlatformContext in common

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/internal/TeacherSubtree.java
```diff
@@ -34,8 +36,8 @@ public final class TeacherSubtree implements AutoCloseable {
      * @param root
      * 		the root of the subtree
      */
-    public TeacherSubtree(final MerkleNode root) {
-        this(root, new StandardTeacherTreeView(root));
+    public TeacherSubtree(@NonNull final PlatformContext platformContext, final MerkleNode root) {
```
hendrikebbers, line 39: Try to avoid platform context in common

platform-sdk/swirlds-fchashmap/src/main/java/com/swirlds/fchashmap/config/FCHashMapConfig.java
```diff
@@ -1,34 +0,0 @@
...
- * Configuration for {@link com.swirlds.fchashmap.FCHashMap}.
- *
- * @param rebuildSplitFactor
- *      split the binary tree at this depth relative to the root of the binary tree. The tree will be split into 2^split-factor subtrees, and each subtree will be eligible to be handled on a separate thread.
- * @param rebuildThreadCount
- *      rebuilding the FCHashMap in a MerkleMap, use this many threads to rebuild the tree.
- *
- */
-@ConfigData("fcHashMap")
-public record FCHashMapConfig(
```
hendrikebbers, line 32: Is that config not used at all????

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ApplicationDefinitionLoader.java
```diff
@@ -93,9 +93,11 @@ private ApplicationDefinitionLoader() {}
         // the line is: app, jarFilename, optionalParameters
         final String appJarFilename = appConfig.jarName();
         // this is a real .jar file, so load from data/apps/
-        Path appJarPath = ConfigurationHolder.getConfigData(PathsConfig.class)
-                .getAppsDirPath()
-                .resolve(appJarFilename);
+        final PathsConfig defaultPathsConfig = ConfigurationBuilder.create()
```
hendrikebbers, line 96: Can we pass the config as a param to the method. Since this is not a main method it can end in strange behavior if somebody assumes that settings.txt is used.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -114,8 +112,7 @@ public static void launch(@NonNull final CommandLineArgs commandLineArgs, final
             return;
         }
 
-        final Path log4jPath =
-                ConfigurationHolder.getConfigData(PathsConfig.class).getLogPath();
+        final Path log4jPath = getAbsolutePath("log4j2.xml");
```
hendrikebbers, line 115: Fine with that but please define a constant for the path

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/StaticPlatformBuilder.java
```diff
@@ -78,11 +78,10 @@ static boolean doStaticSetup(@NonNull final Configuration configuration, @Nullab
...
         staticSetupCompleted = true;
 
-        ConfigurationHolder.getInstance().setConfiguration(configuration);
+        ConfigurationHolder.setConfiguration(configuration);
 
-        // Setup logging, using the configuration to tell us the location of the log4j2.xml
-        final PathsConfig pathsConfig = configuration.getConfigData(PathsConfig.class);
-        final Path log4jPath = pathsConfig.getLogPath();
+        // Setup logging
+        final Path log4jPath = getAbsolutePath("log4j2.xml");
```
hendrikebbers, line 84: constant

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
```diff
@@ -87,11 +86,9 @@ public static Configuration buildBasicConfiguration(
             @NonNull final Path settingsPath, @NonNull final List<Path> configurationPaths) throws IOException {
         final ConfigSource settingsConfigSource = LegacyFileConfigSource.ofSettingsFile(settingsPath);
         final ConfigSource mappedSettingsConfigSource = ConfigMappings.addConfigMapping(settingsConfigSource);
-        final ConfigSource threadCountPropertyConfigSource = new ThreadCountPropertyConfigSource();
 
-        final ConfigurationBuilder configurationBuilder = ConfigurationBuilder.create()
-                .withSource(mappedSettingsConfigSource)
-                .withSource(threadCountPropertyConfigSource);
```
hendrikebbers, line 94: Not needed anymore? I remember there was a config that default value depends on the cpu count.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/SystemTransactionType.java
```diff
@@ -22,11 +22,5 @@
 public enum SystemTransactionType {
 
     /** first byte of a system transaction giving a signed state's (round number, signature) */
-    SYS_TRANS_STATE_SIG,
-
-    /** first byte of a system transaction giving all avgPingMilliseconds stats (sent as ping time in microseconds) */
-    SYS_TRANS_PING_MICROSECONDS,
```
hendrikebbers, line 28: You create as horrible big PRs as I often do :D

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/SignedStateFileReadWriteTest.java
```diff
@@ -160,100 +156,6 @@ void writeSavedStateToDiskTest() throws IOException {
...
-        changeConfigAndConfigHolder("foo/bar/baz");
-
-        assertEquals(getAbsolutePath("./foo/bar/baz"), getSignedStatesBaseDirectory(), "unexpected saved state file");
-
-        changeConfigAndConfigHolder("data/saved");
-    }
-
-    @Test
-    @DisplayName("getSignedStatesDirectoryForApp() Test")
-    void getSignedStatesDirectoryForAppTest() {
```
hendrikebbers, line 179: Why do you remove all that tests?

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/system/transaction/internal/TransactionSerializationTest.java
```diff
@@ -88,69 +83,6 @@ void SignatureSerializeDeserializeTest(final int sigSize) throws IOException {
...
-        TestExpectedSerializationLength(systemTransactionBitsPerSecond, true);
-        TestExpectedSerializationLength(deserialized, true);
-
-        TestExpectedSerializationLength(systemTransactionBitsPerSecond, false);
-        TestExpectedSerializationLength(deserialized, false);
-    }
-
-    @ParameterizedTest
-    @ValueSource(ints = {0, 1, 2, 10, 64, 100})
-    void PingSerializeDeserializeTest(final int nodeAmount) throws IOException {
```
hendrikebbers, line 126: Why do you remove that tests?

alittley (dismissed):

hedera-node/hedera-config/src/testFixtures/java/com/hedera/node/config/testfixtures/HederaTestConfigBuilder.java
cody-littley: I removed it from everywhere. Settings for `FCHashMap` became static. I did this for a few reasons:

- I added these settings a few years ago and we have never used anything except for the defaults
- `FCHashMap` is no longer used in production (replaced by `VirtualMap`)
- if this configuration remained it would have to be static config, which I am on a crusade to eliminate :)

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/config/singleton/ConfigurationHolder.java
cody-littley: I've reverted changes to this class.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/CryptographyHolder.java
cody-littley: It's necessary if we want to be thread safe. If thread A sets the variable and thread B reads it without establishing a happens-before relationship with thread A, then the variable it reads may be partially initialized.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/MerkleCryptoFactory.java
cody-littley: Let's discuss this when we are both next online. I'm not sure how to write a test fixture that can be forced to run before all unit tests in the repo.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/TeachingSynchronizer.java
cody-littley: The reconnect classes don't really belong in common. But since I don't want to add 5000 lines to this PR, simpler just to do as you ask and remove the context.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/internal/TeacherSubtree.java
cody-littley: Does not belong in common, but removed since I don't want to move the reconnect classes right now.

platform-sdk/swirlds-fchashmap/src/main/java/com/swirlds/fchashmap/config/FCHashMapConfig.java
cody-littley: These are now just static variables. `FCHashMap` isn't used in production any more, and we never varied configuration when it was.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ApplicationDefinitionLoader.java
cody-littley: Done.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
cody-littley: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/StaticPlatformBuilder.java
cody-littley: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/DefaultConfiguration.java
cody-littley: Only needed by FCHashMap, which is now all static configuration (and no longer production code).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/SystemTransactionType.java
cody-littley: 😜

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/SignedStateFileReadWriteTest.java
cody-littley: These tests tested methods that were only called by these tests. I removed both the platform code and the tests.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/system/transaction/internal/TransactionSerializationTest.java
cody-littley: They are testing serialization for transaction types we don't use anywhere, and have no plans on using anywhere. I deleted both the transactions and the tests on the transactions. They were using static configuration, and my goal was to eliminate static config use by any means possible.

platform-sdk/sdk/settings.txt
```diff
@@ -46,3 +46,5 @@ prometheus.endpointEnabled,               false
                                           #############################
 
 loadKeysFromPfxFiles,                     false
+crypto.cpuVerifierThreadRatio,            0.1
```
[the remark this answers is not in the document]
cody-littley: Yes.

As a part of this PR, when running a multiple nodes in the same JVM, the platform context for each node now has a different cryptography instance. Previously it would always share the same cyrptography instance between all nodes in the same JVM. 

Since laptop resources are in high demand, I figured if we are having 4x the number of cryptography instances, it makes sense to reduce the size of the thread pool for cryptography.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
cody-littley: Done.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/CryptographyHolder.java
cody-littley: >  Let's name it init/reset and add info to the javadoc that it is only needed to make it useable in tests.

done

> Question: How long will the holder stay? 

Hopefully not long. I don't like it. Static access to anything in the context == 🤮 

> Maybe we introduce a Unit extension (see https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/base/test-support/test-support.md) that resets Config Holder & Crypto Holder.

Let's discuss this the next time we are both online. Sounds like a good idea, but I'm not sure exactly what needs to go into making this happen.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/MerkleCryptoFactory.java
cody-littley: Renamed to `init()` like suggested for the config holder. Let's discuss next steps together the next time we are both online.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/MerkleCryptography.java
cody-littley: Done.

edward-swirldslabs (dismissed):

alittley (approved):

edward-swirldslabs (approved):

alittley (approved):

JeffreyDallas (approved):

imalygin (approved):

hendrikebbers (approved):

edward-swirldslabs (approved):

PR #10680, opened 2023-12-29
chore: 7163 use instant source instead of time

### LINKED ISSUE #7163: Evaluate usage of InstantSource in platform
In the services team, we decided to use `java.time.InstantSource` instead of `com.swirlds.base.time.Time`. I think it would make sense to use `InstantSource` in the platform, too. (Sorry, I was unaware of this interface when I introduced `Time` in the platform repository.)

mxtartaglia-sl:
**Description**:
PR for 3erd alternative described in https://github.com/hashgraph/hedera-services/issues/7163

**Related issue(s)**:

Fixes #7163


platform-sdk/swirlds-base/src/main/java/com/swirlds/base/time/Time.java
```diff
@@ -17,46 +17,16 @@
...
-import java.time.Instant;
+import java.time.InstantSource;
 
 /**
- * An API for getting the time. All platform code should utilize this API instead of the raw standard
- * java time APIs. This makes it much easier to simulate time in test environments.
+ * @deprecated users of this class should migrate to {@link TimeSource} for measuring elapsed time or {@link InstantSource} for wall-clock.
  */
-public interface Time {
+@Deprecated
```
poulok, line 25: I am not sure we want to deprecate this interface. Keeping `Time` as a single source or all time information is useful, at least in the platform.

cody-littley: I agree. We should only deprecate this class if we intend to remove it. In the past we have been over eager to deprecate classes without the corresponding drive to actually remove them. The result of this is that we end up with a lot of deprecated, which is ugly.

mxtartaglia-sl: That is very true; my aim here was to start the discussion. It depends on our goals... Having time to implement the two interfaces makes solving the drawbacks mentioned in the ticket easier, but we wouldn't be replacing anything, and it invites one to think if the whole change makes sense. The other option is to remove time, and to modify the mentioned classes that use both TimeSource and InstantSource to receive both parameters/fields, but that also seems ugly and with the latest change @cody-littley  introduced, it became a broader change.

mxtartaglia-sl:
Closing this pull request after discussion with @poulok @hendrikebbers @cody-littley @tinker-michaelj. We agreed that the gain from this pr is minor in comparison with the scope of the change.

PR #10681, opened 2023-12-29
chore: set event birthRound to pendingConsensusRound

### LINKED ISSUE #10598: EventCreator adds correct `birthRound` to events.

edward-swirldslabs:
Fixes #10598


cody-littley (changes requested):
Nothing wrong with the changes in this PR, but after recent discussions we may want to wait to do this.

We need to figure out how we will do the migration before we enable this. One strategy we may use for migration is to intentionally set an old birth round for all events pre-migration, and force all pre-migration events to become ancient at the moment of migration.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/NonAncientEventWindow.java
```diff
@@ -35,6 +36,12 @@ public class NonAncientEventWindow {
     public static NonAncientEventWindow INITIAL_EVENT_WINDOW = new NonAncientEventWindow(
             ConsensusConstants.ROUND_FIRST, ConsensusConstants.ROUND_FIRST, EventConstants.FIRST_GENERATION);
 
+    /**
+     * This static variable is temporary until we throw the switch permanently from minGenNonAncient to
+     * minRoundNonAncient
+     */
+    private static AtomicBoolean useBirthRoundForAncient = new AtomicBoolean(false);
```
cody-littley, line 43: Instead of using static state, would it be possible to pass in another parameter to the constructor of this class? My concern is that using the this static object might make unit tests easy to mess up, e.g. tests that only pass when run in a specific order.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
```diff
@@ -386,6 +386,11 @@
         this.emergencyRecoveryManager = Objects.requireNonNull(emergencyRecoveryManager, "emergencyRecoveryManager");
         final Time time = Time.getCurrent();
 
+        // When the setting goes away, this method call should be deleted
```
alittley, line 389: Consider annotation with `Future work:`

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/NonAncientEventWindow.java
edward-swirldslabs: just hit that, unit tests are messing up because of the static state.  I will pass in a constructor argument.  I think that will still work out since the initial event window should alway say every event is non-ancient whether compared by round or by min generation.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
edward-swirldslabs: This code was removed since the static initialization is no longer happening.

cody-littley (dismissed):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/tipset/TipsetEventCreatorTests.java
```diff
@@ -904,10 +906,11 @@ void noStaleEventsAtCreationTimeTest() {
      * FUTURE WORK: Update this test to use RosterDiff instead of NonAncientEventWindow
      */
     @ParameterizedTest
-    @ValueSource(booleans = {false, true})
+    // @ValueSource(booleans = {false, true})
```
alittley, line 909: Commented out code

alittley (approved):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/tipset/TipsetEventCreatorTests.java
edward-swirldslabs: removed

cody-littley (approved):

alittley (approved):

hendrikebbers (approved):

PR #10699, opened 2024-01-02
fix: Enable fuzzy matching in PrngSeedOperationSuite

### LINKED ISSUE #10629: Enable fuzzy record matching for PrngSeedOperationSuite (3 tests)

stoyanov-st:
**Description**:

**Related issue(s)**:

Fixes #10629 

**Notes for reviewer**:

**Checklist**


tinker-michaelj (changes requested):
We want to move required `SnapshotMatchMode`'s into the `defaultHapiSpec()` factory like [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/opcodes/ExtCodeSizeOperationSuite.java#L80), and then set `@HapiApiSuite(fuzzyMatch = true)`.

tinker-michaelj (approved):
LGTM, ty @stoyanov-st !

PR #10701, opened 2024-01-02
feat: unlink alias on account deletion

### LINKED ISSUE #10698: Unlink alias on account deletion
### Abstract / Motivation

When an account is created via the auto-creation flow, it has a Hedera entity id such as 0.0.x _and_ an alias assigned to it such as an EVM address type alias. If the account is then deleted, the alias remains assigned to it, and thus it becomes un-useable - further transfers to this alias result in status `ACCOUNT_DELETED`.

We should instead create a _new_ Hedera entity and assign the alias to the new account. 

[This EET](https://github.com/hashgraph/hedera-services/blob/f26302ce24acdf83197a4f8ea7073825a6d32cea/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java#L1431) can be executed to reproduce the scenario.

### Specification

Implement the logic needed to un-link the alias from the account when the account is deleted.
Add a feature flag to control whether to release the alias or not.
Implement the change in both mono-service and modularised services codebase.

### Related issues
https://github.com/hashgraph/hedera-services/issues/9728 Address account nonce discrepancies

natanasow:
**Description**:

When an account is created via the auto-creation flow, it has a Hedera entity id such as 0.0.x _and_ an alias assigned to it such as an EVM address type alias. If the account is then deleted, the alias remains assigned to it, and thus it becomes un-useable - further transfers to this alias result in status `ACCOUNT_DELETED`.

We should instead create a _new_ Hedera entity and assign the alias to the new account. 

[This EET](https://github.com/hashgraph/hedera-services/blob/f26302ce24acdf83197a4f8ea7073825a6d32cea/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java#L1431) can be executed to reproduce the scenario.

**Specification**:

Implement the logic needed to un-link the alias from the account when the account is deleted.
Add a feature flag to control whether to release the alias or not.
Implement the change in both mono-service and modularised services codebase.

**Related issue(s)**:

Fixes #10698

**Notes for reviewer**:

**Checklist**


stoqnkpL (dismissed):
LGTM!

lukelee-sl (dismissed):
LGTM. @Neeharika-Sompalli adding you to take a look as this is token service :pray:

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (dismissed):
LGTM, ty @natanasow !

hedera-node/hedera-mono-service/src/main/resources/bootstrap.properties
```diff
@@ -20,6 +20,7 @@ accounts.lastThrottleExempt=100
 accounts.nodeRewardAccount=801
 accounts.stakingRewardAccount=800
 accounts.storeOnDisk=true
+accounts.releaseAliasAfterDeletion=true
```
poulok, line 23: Does this setting also need to be in `application.properties`?

natanasow: I'm not sure :confused:. 

In exactly which `application.properties` file the flag should be added as well because there are 11 of them? Every new feature flag I've been added till now has been defined in `bootstrap.properties` only. Some of the `application.properties` files are completely empty or have a few definitions. So I'm curious what's the main idea behind `application.properties` files, how they differ from the `bootstrap.properties`, and when they should be updated?

stoqnkpL (approved):

hedera-node/hedera-mono-service/src/main/resources/bootstrap.properties
poulok: application.properties is used by the modular code, while bootstrap.properties is used by the mono code. There should be a matching set of files for each location. @netopyr should be able to provide more guidance.

Nana-EC: Kelly is correct.
The point here being due to the complexity of changes we sometime opt for separate mono and modularized PRs
So far as the feature set and configs in this case match up within the milestone we're good and there won't be divergence.
So this PR can update `bootstrap.properties` only and I believe @lukelee-sl will handle the modular update that updates `application.properties`

Nana-EC (approved):
LGTM

hedera-node/hedera-mono-service/src/main/resources/bootstrap.properties
stoqnkpL: This PR addresses both mono and mod codebases so if we need to add the properties to `application.properties` we should do it. 
My understanding was that we don't need to add to the `application.properties` files because we've added them in the hedera-config module [here](https://github.com/hashgraph/hedera-services/pull/10701/files#diff-abd78ef65c6c7e4f1277e5d485f804d88173af0db888f7bcd13e2be51fa0d328R37). 
If we need to add them can you point us to the file that has to be updated please @poulok @netopyr @lukelee-sl

Neeharika-Sompalli (approved):
LGTM ! Thanks @natanasow

Issue #10702, opened 2024-01-02
Enhance RecordStreamEntry to have a getter for the TransactionRecord


stefan-stefanooov:
### Problem

`TransactionRecord` cannot be accessed from the `RecordStreamEntry` record entity.
It is needed for the Debug Mode tooling.### 

### Solution

Adding getter for the `TransactionRecord`

PR #10703, opened 2024-01-02
feat: adding getter to RecordStreamEntry for TransactionRecord

stefan-stefanooov:
**Description**:
 Adding getter to `RecordStreamEntry` for `TransactionRecord`.
We are trying to enhance a new feature that the local node has  -> debug command. It takes consensus timestamp and outputs the corresponding record file to the command line.
We are trying to filter all the unnecessary records form the output but we need some more data to be accessible in the `RecordStreamEntry` entity.

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/10702


tinker-michaelj (approved):
LGTM!

PR #10710, opened 2024-01-02
chore: Clean up schedule HapiSpec suites

jsync-swirlds:
* Removed leftover todo comments
* Consolidated repeated/redundant string constants to ScheduleUtils class
* Consolidated shared utility functions to ScheduleUtils class


povolev15 (approved):
LGTM

PR #10716, opened 2024-01-03
chore: fix mutability exception in pre-handle; stabilize CI

tinker-michaelj:
**Description**:
 - Fixes unintentional use of `WritableStates` in `RecordCacheImpl#getReadableQueue()`.
 - Stabilizes `canAutoCreateWithFungibleTokenTransfersToAlias()` in presence of end-of-day staking records by using `getFirstNonStakingChildRecord()`.
 - Creates a utility method `assumingNoStakingChildRecordCausesMaxChildRecordsExceeded()` that allows transactions requiring all preceding child records to be used in CI; uses it to stabilize two specs.

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2021-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


Neeharika-Sompalli (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationSuite.java
```diff
@@ -1331,43 +1331,28 @@ final HapiSpec multipleAutoAccountCreations() {
...
+                        cryptoTransfer(
+                                        tinyBarsFromToWithAlias(PAYER, "alias4", 7 * ONE_HUNDRED_HBARS),
+                                        tinyBarsFromToWithAlias(PAYER, "alias5", 100))
+                                .via("failedAutoCreate")
+                                .hasKnownStatus(INSUFFICIENT_ACCOUNT_BALANCE),
+                        getTxnRecord("failedAutoCreate")
+                                .hasNonStakingChildRecordCount(0)
+                                .logged(),
+                        getAccountInfo(PAYER)
+                                .has(accountWith().balance((INITIAL_BALANCE * ONE_HBAR) - 3 * ONE_HUNDRED_HBARS))));
```
mhess-swl, line 1355: The `allRunFor` seems to have disappeared, was that intentional?

tinker-michaelj: Yes, these operations are now executed directly by the  `assumingNoStakingChildRecordCausesMaxChildRecordsExceeded()` wrapper operation; it is equivalent.

PR #10735, opened 2024-01-03
test: add additional test for conditional records

georgi-l95:
**Description**:
Add suggested test from https://github.com/hashgraph/hedera-services/issues/9424 :
An ERC-20 transferFrom() sending fungible value to an absent receiver address (outside long-zero range) generates a preceding child with a synthetic CryptoCreate and TransactionRecord identical that created by mono-service when it resolves to SUCCESS.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


tinker-michaelj (commented):
Since this is modifying an existing spec without updating record snapshots, won't it break https://github.com/hashgraph/hedera-services/pull/10768?

georgi-l95:
@tinker-michaelj Yep, I've created this PR, before the PR you've mentioned, but will rebase later when #10768 is merged, if it's the correct scenario that you mentioned in issue #9424 . Or we can close this one, if you think that it's not needed ?

tinker-michaelj (approved):
LGTM @georgi-l95 , we may run into fuzzy-matching problems down the road but this is a good pure HAPI test too. 💯

PR #10750, opened 2024-01-03
chore: clean up contract call sanity precheck logic

lukelee-sl:
**Description**:
Refactor the logic for conditions to precheck for possible invalid contract call transactions
 
**Checklist**

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
```diff
@@ -255,6 +251,25 @@ private ResponseCodeEnum validateSemantics(final TransactionBody transactionBody
...
+
+        // Tokens are valid targets
+        if (entityAccess.isTokenAccount(target.toId().asEvmAddress())) {
+            return false;
+        }
+
+        // Check for possible lazy create
+        if (target.equals(EntityNum.MISSING_NUM)
+                && isOfEvmAddressSize(op.getContractID().getEvmAddress())
+                && op.getAmount() > 0) {
```
Neeharika-Sompalli, line 266: Should this be ?  
```suggestion
              (target.equals(EntityNum.MISSING_NUM)
                && isOfEvmAddressSize(op.getContractID().getEvmAddress())
                ||op.getAmount() > 0
```

lukelee-sl: I suppose it depends on how broad or narrow this precheck should be.  I don't see any not to narrow the scope as you suggest.  Will change as suggested.

Neeharika-Sompalli: My comment was just based on the check that existed in mono-service. Not sure if it needs to be the same also.

Neeharika-Sompalli (approved):
LGTM ! Thanks @lukelee-sl

lukelee-sl:
> LGTM ! Thanks @lukelee-sl

Thank you for taking a look @Neeharika-Sompalli !

Neeharika-Sompalli (approved):
LGTM !

PR #10756, opened 2024-01-03
fix: refactor the code to use explicit Schema and not anonymous Schema and one flaky test

### LINKED ISSUE #10749: Use top-level classes for mod-service `Schema`s in preparation for first release
### Problem

For many `Service`s we are using anonymous `Schema`s that adopt whatever the current version of the software is. 

This is reasonable when we don't know the first release version of mod-service.

It might also be reasonable for services whose `Schema` will probably never change; and just want to make sure they register a schema that will be always be detected as "applicable" by the `MerkleSchemaRegistry` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java#L168).

But it is not ideal for services that want to provide an explicit sequence of release-specific migrations that evolve over time.

### Solution

Extract all anonymous `Schema`'s to top-level classes with names such as `InitialModServiceTokenSchema`.

Add to javadoc a comment like,
```
     * (FUTURE) When mod-service release is finalized, rename this class to e.g.
     * {@code Release47TokenSchema} as it will no longer be appropriate to assume
     * this schema is always correct for the current version of the software.
```

### Alternatives

_No response_

povolev15:
refactor the code to use explicit Schema and not anonymous Schema, it will allow providing an explicit sequence of specific release

It also fixes one flaky test related to record cache

**Related issue(s)**:

Fixes #10749

### DIFF
--- a/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
+++ b/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
@@ -30,11 +30,6 @@ import java.util.Set;
  * this schema is always correct for the current version of the software.
  */
 public class InitialModServiceNetworkSchema extends Schema {
-    /**
-     * (FUTURE) When mod-service release is finalized, rename this class to e.g.
-     * {@code Release47TokenSchema} as it will no longer be appropriate to assume
-     * this schema is always correct for the current version of the software.
-     */
     public InitialModServiceNetworkSchema(SemanticVersion version) {
         super(version);
     }
--- a/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
+++ b/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
+++ b/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.service.networkadmin.impl.schemas;
+
+import com.hedera.hapi.node.base.SemanticVersion;
+import com.hedera.node.app.spi.state.MigrationContext;
+import com.hedera.node.app.spi.state.Schema;
+import com.hedera.node.app.spi.state.StateDefinition;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Set;
+
+/**
+ * General schema for the network service
+ * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+ * {@code Release47NetworkSchema} as it will no longer be appropriate to assume
+ * this schema is always correct for the current version of the software.
+ */
+public class InitialModServiceNetworkSchema extends Schema {
+    /**
+     * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+     * {@code Release47TokenSchema} as it will no longer be appropriate to assume
+     * this schema is always correct for the current version of the software.
+     */
+    public InitialModServiceNetworkSchema(SemanticVersion version) {
+        super(version);
+    }
+
+    @NonNull
+    @Override
+    public Set<StateDefinition> statesToCreate() {
+        return Set.of();
+    }
+
+    @Override
+    public void migrate(@NonNull MigrationContext ctx) {}
+}
--- a/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
+++ b/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.service.schedule.impl.schemas;
+
+import com.hedera.hapi.node.base.ScheduleID;
+import com.hedera.hapi.node.base.SemanticVersion;
+import com.hedera.hapi.node.state.primitives.ProtoLong;
+import com.hedera.hapi.node.state.primitives.ProtoString;
+import com.hedera.hapi.node.state.schedule.Schedule;
+import com.hedera.hapi.node.state.schedule.ScheduleList;
+import com.hedera.node.app.service.schedule.impl.ScheduleServiceImpl;
+import com.hedera.node.app.spi.state.Schema;
+import com.hedera.node.app.spi.state.StateDefinition;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Set;
+
+/**
+ * General schema for the schedule service
+ * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+ * {@code Release47ScheduleSchema} as it will no longer be appropriate to assume
+ * this schema is always correct for the current version of the software.
+ */
+public final class InitialModServiceScheduleSchema extends Schema {
+    public InitialModServiceScheduleSchema(final SemanticVersion version) {
+        super(version);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @NonNull
+    @Override
+    public Set<StateDefinition> statesToCreate() {
+        return Set.of(schedulesByIdDef(), schedulesByExpirySec(), schedulesByEquality());
+    }
+
+    private static StateDefinition<ScheduleID, Schedule> schedulesByIdDef() {
+        return StateDefinition.inMemory(
+                ScheduleServiceImpl.SCHEDULES_BY_ID_KEY, ScheduleID.PROTOBUF, Schedule.PROTOBUF);
+    }
+
+    private static StateDefinition<ProtoLong, ScheduleList> schedulesByExpirySec() {
+        return StateDefinition.inMemory(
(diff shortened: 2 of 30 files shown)


hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
```diff
@@ -0,0 +1,63 @@
...
+import com.hedera.node.app.service.schedule.impl.ScheduleServiceImpl;
+import com.hedera.node.app.spi.state.Schema;
+import com.hedera.node.app.spi.state.StateDefinition;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Set;
+
+/**
+ * General schema for the schedule service
+ * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+ * {@code Release47ScheduleSchema} as it will no longer be appropriate to assume
```
jsync-swirlds, line 34: Rather than renaming classes for each release (which seems like a lot of wasted work), would it not make more sense to work more like Serialized classes and leave the name basic (e.g. `ScheduleServiceSchema`), and have a public constant indicating the "earliest valid release" for each schema?
Then we only change schema classes when needed and update that constant to represent the release when the schema changes.
Example:
* Add abstract method to `Schema`
   * `protected abstract SemanticVersion getEarliestValidVersion();`
   * `protected SemanticVersion getEarliestValidVersion() {return EARLIEST_VALID_RELEASE;}`
* Initial change
   * `public static final SemanticVersion EARLIEST_VALID_RELEASE = new SemanticVersion(0, 0, 0, "Initial-Modularization", "");`
* First main net release
   * `public static final SemanticVersion EARLIEST_VALID_RELEASE = new 
SemanticVersion(0, 47, 0, "", "");`
* Later main net release
   * `public static final SemanticVersion EARLIEST_VALID_RELEASE = new SemanticVersion(0, 49, 2, "", "");`

Then we can check the version in the `Schema` constructor:
```java
    protected Schema(@NonNull SemanticVersion version) {
        this.version = Objects.requireNonNull(version);
        if(!sameOrLaterVersion(version, getEarliestValidVersion())) {
            throw new InvalidSchemaVersionException(version, getEarliestValidVersion());
        }
    }
```

This approach might require fewer unnecessary code changes, and also be, at least somewhat, self-enforcing.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ReadableContractStateStore.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
```
jsync-swirlds, line 2: Why are you back-dating the code?  That seems like it might be very much incorrect.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/InitialModServiceContractSchema.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2020-2024 Hedera Hashgraph, LLC
```
jsync-swirlds, line 2: Why are you back-dating the code?  That seems like it might be very much incorrect.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/records/RecordCreationSuite.java
```diff
@@ -270,8 +271,9 @@ final HapiSpec payerRecordCreationSanityChecks() {
                     final var payerId = spec.registry().getAccountID(PAYER);
                     final var subOp = getAccountRecords(PAYER).logged();
                     allRunFor(spec, subOp);
-                    final var records =
-                            subOp.getResponse().getCryptoGetAccountRecords().getRecordsList();
+                    final var records = subOp.getResponse().getCryptoGetAccountRecords().getRecordsList().stream()
```
jsync-swirlds, line 274: This change seems unrelated.  Was this supposed to be part of a separate PR?

povolev15: It can be unrelated but we fix the flaky test here as well. I can change the description

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/InitialModServiceContractSchema.java
povolev15: I will change it back to 2023. I thought we want to be aligned with all the files

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ReadableContractStateStore.java
povolev15: I will change it back to 2023. I thought we want to be aligned with all the files

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
povolev15: @tinker-michaelj what do you think because I did whatever you asked but Joseph approach is also valid

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ReadableContractStateStore.java
jsync-swirlds: I think we need to be "correct" on these, as it's a formal notice, not code.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
mhess-swl: The idea is to tie each schema to a single release. Before releasing modularization we're naming these schemas in a somewhat arbitrary way, but eventually – when these go to production – we will tie each schema to a single version.  The value proposition of this PR is more to extract the anonymous classes than to focus on the interim names.

povolev15: But I  think Joseph proposal is that the class name will always be the same and we will not change the class name each time.

mhess-swl: We'll only need to change the class name once after this: when we finally push the modularized code to production.  We _want_ each class name to have the version number, e.g. `TokenSchema47`. We'll assign the release name when we're about to release.

jsync-swirlds: If we put the version in the class name, then, as I understand the plan expressed, we're signing up to basically copy/rename every schema every release, not just "once".  Otherwise what's the point of version-named classes?  Are we just wanting to have an historical version name in every class that will never change?

Not sure the entire plan here, but this name-for-release approach just seems like it's asking for a ton of work to manually manage it all when the software could be automating things (and we're possibly rushing something that could use more thought).

It seems like we have a more general problem we need to solve; how to effectively and efficiently handle long-term schema evolution.  I suspect we need something more thorough than naming classes with a software version; most likely we need something like a _versioned_ schema registry that stores all versions of each schema and can manage those versions for migration and other purposes as needed.

I am mostly just asking questions here, because this seems off.
Are we expecting to have release `2035.4.12` referencing schema named `Release47ScheduleSchema`, and are we expecting to see that mixed in with `Release2035_4_12_InterplanetaryConsensusService` in the same code?  Alternatively are we expecting to see an hundred services (again, in a distant future release) all forever referencing `Release47SomethingService` as schema name?

What is the core value of putting version in the _class name_ rather than having the version in a _static constant_ that can be more easily used and compared in code without manually rewriting schema and migration implementations for every change?

mhess-swl: > we're signing up to basically copy/rename every schema every release
We don't have to copy the schema every release. We only have to create a new schema object when there's a change to the schema. If we start with `Release45Schema` and there's no change until release 0.52, we only have two schema classes: `Release45Schema` and `Release52Schema`. 

> ...it's asking for a ton of work to manually manage it all when the software could be automating things...
There's already a mechanism for managing the correct application of schemas based on version. We _want_ to manually manage the changes that each schema makes because there are serious consequences if the schema is ever non-deterministically applied. IMO making it intentional to require another schema is a plus so it's harder to do by accident.

> ...most likely we need something like a versioned schema registry that stores all versions of each schema and can manage those versions for migration and other purposes as needed.
If I understand this correctly, we already have exactly that – `MerkleSchemaRegistry` manages the schema versions and which should be applied during migrations. We register _all_ versions of a schema during node startup (all the way from the beginning of time), and the node decides which schemas (plural) are executed based on the node's current running version. Let's use our example above where there's a `Release45Schema` and a `Release52Schema`; if we start the node on version 0.44, it registers both schema classes, but never executes them because there's no need for a migration. If we start the node on 0.46, `MerkleSchemaRegistry` computes that the migration for `Release45Schema` needs to run and runs it. If we start the node on 0.53, _both_ schema migrations are run _in sequence_. See `MerkleSchemaRegistryTest#migrate()` (parameterized test) for a thorough example. Richard's schema migration implementation is based on the DB version control model, e.g. Flyway (see [here](https://documentation.red-gate.com/flyway/flyway-cli-and-api/concepts/migrations)).

> Are we expecting to have release 2035.4.12 referencing schema named Release47ScheduleSchema...
Not sure what you mean. _What_ in the release would reference `Release47ScheduleSchema`? The only thing that should be referencing the schema object at all is each service implementation, the idea being that each service manages its own schemas independently of other services. 

> ...are we expecting to see that mixed in with Release2035_4_12_InterplanetaryConsensusService in the same code? 
Yes. See `MerkleAccountState#deserialize(...)` in the mono code for an example; it references many releases. The modular code won't work exactly the same way, of course, but it will always be necessary to know precisely which version was in play when de/serializing (at the very least). 

> Alternatively are we expecting to see an hundred services (again, in a distant future release) all forever referencing Release47SomethingService as schema name?
Yes, this pinpoints (by version) precisely when a change was made. Every service for the rest of time will need to run all migrations less than or equal to the running node's current version; the DB change model is built on the central concept of referencing that version.

> What is the core value of putting version in the class name rather than having the version in a static constant that can be more easily used and compared in code...
Every schema instance will already _have_ a version object for comparisons in code. `currentVersion` is only passed in to schemas right now until we release. In reality we can name these classes whatever we want, but because the version is so integral, I find it very convenient to have the version in the class name. 

> ...without manually rewriting schema and migration implementations for every change?
What rewrites are we talking about here? Each migration depends on the migrations before it, so each migration need only handle the _changes_ from version X to version X + 1. 

My apologies if I've misunderstood. There may well be some nuance here, differing from the traditional DB change model, that I haven't picked up on.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceNetworkSchema.java
```diff
@@ -0,0 +1,50 @@
...
+ * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+ * {@code Release47NetworkSchema} as it will no longer be appropriate to assume
+ * this schema is always correct for the current version of the software.
+ */
+public class InitialModServiceNetworkSchema extends Schema {
+    /**
+     * (FUTURE) When mod-service release is finalized, rename this class to e.g.
+     * {@code Release47TokenSchema} as it will no longer be appropriate to assume
+     * this schema is always correct for the current version of the software.
+     */
```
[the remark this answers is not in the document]
povolev15: by mistake removed it thanks

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/InitialModServiceScheduleSchema.java
jsync-swirlds: Thank you for the detailed explanation.
It roughly matches my understanding.  I believe this approach, while it will work, is still signing up for a less-than-ideal future state by replicating schema code and retaining it forever.  That said, it's definitely a larger discussion than would be appropriate within a single PR.

jsync-swirlds (approved):

iwsimon (dismissed):
LGTM

mhess-swl (approved):

kimbor (approved):
approved

tinker-michaelj (approved):
LGTM, ty @povolev15 !

Neeharika-Sompalli (approved):
LGTM !

PR #10757, opened 2024-01-03
chore: remove the unused encryption key pair and silently ignore if present

### LINKED ISSUE #10665: mTLS Keys: Stop loading and no longer require the unused encryption keypair

nathanklick:
## Description

- Removes the encryption key loading from the `CryptoStatic`, `PublicKeys` and `KeysAndCerts` classes. 
- Removes the encryption key from the `equals()` comparison in the `Address` class.
- Silently ignores the encryption key if it is present in the `public.pfx` or `private-nodeN.pfx` files.
- Continues to load and store existing encryption keys if present in the address book from state.
- Minor code formatting issues and other cleanup.

### Related Issues

- Closes #10665


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -44,27 +42,10 @@
...
-import java.security.KeyStoreException;
-import java.security.NoSuchAlgorithmException;
-import java.security.NoSuchProviderException;
-import java.security.PublicKey;
-import java.security.SecureRandom;
-import java.security.Security;
-import java.security.Signature;
-import java.security.SignatureException;
-import java.security.UnrecoverableKeyException;
+import java.security.*;
```
edward-swirldslabs, line 45: Our coding style guide is an extension of Google's and we do not allow imports to have wildcards (3.3.1) : https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

I think there is an IDE setting which can be made to not use wildcards.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -326,23 +292,18 @@ private static String getPrivateKeysFileName(final String memberName) {
...
+     * @return map of key stores
+     * @throws KeyStoreException   if there is no provider that supports {@link CryptoConstants#KEYSTORE_TYPE}
+     * @throws KeyLoadingException in an issue occurs while loading keys and certificates
      */
     @NonNull
     static Map<NodeId, KeysAndCerts> loadKeysAndCerts(
             @NonNull final AddressBook addressBook, @NonNull final Path keysDirPath, @NonNull final char[] password)
             throws KeyStoreException, KeyLoadingException, UnrecoverableKeyException, NoSuchAlgorithmException {
-        Objects.requireNonNull(addressBook, "addressBook must not be null");
+        Objects.requireNonNull(addressBook, ADDRESS_BOOK_MUST_NOT_BE_NULL);
```
edward-swirldslabs, line 306: I'm wondering about the purpose of making this particular string a string-constant but not the other strings in the `Objects.requireNonNull`.  At first I was putting these strings in as the second parameter, but there is so little value in the strings that I have been leaving them off.   But as is, there is an asymmetry of treatment of the strings in this file.

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
nathanklick: @edward-swirldslabs This was the only constant value which was repeated in multiple places.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
nathanklick: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
nathanklick: IMHO, The second parameter is useful for at a glance determining which parameter was `null` without having to pull up the code and match line numbers to the exception stack trace.

edward-swirldslabs (approved):

cody-littley (approved):

PR #10792, opened 2024-01-05
feat: Enable fuzzy matching in `ContractHTSSuite`

### LINKED ISSUE #10789: Enable fuzzy record matching for ContractHTSSuite (1 test)

georgi-l95:
**Description**:
Enable fuzzy matching in `ContractHTSSuite`

**Related issue(s)**:

Fixes #10789 

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
```diff
@@ -98,6 +101,11 @@ final HapiSpec nonZeroTransfersFail() {
         final var theSecondReceiver = "somebody2";
         return defaultHapiSpec("NonZeroTransfersFail")
                 .given(
+                        snapshotMode(
+                                SnapshotMode.FUZZY_MATCH_AGAINST_HAPI_TEST_STREAMS,
+                                SnapshotMatchMode.NONDETERMINISTIC_CONSTRUCTOR_PARAMETERS,
+                                SnapshotMatchMode.NONDETERMINISTIC_FUNCTION_PARAMETERS,
+                                SnapshotMatchMode.NONDETERMINISTIC_TRANSACTION_FEES),
```
Neeharika-Sompalli, line 108: We can just use `fuzzyMatch=true` on the `@HapiTestSpec` instead

georgi-l95: Yes, I tried, but sometimes it fails because transaction fees are not the same.

tinker-michaelj: @georgi-l95 In that case we should just pass the needed snapshot modes into the `defaultHapiSpec()` factory (e.g. [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/opcodes/ExtCodeSizeOperationSuite.java#L80)) and then set `@HapiTestSuite(fuzzyMatch = true)`.

tinker-michaelj (changes requested):
We want to avoid explicit `snapshotMode()` operations so that we both,
  1. Keep spec definitions shorter; and,
  2. Have a single point-of-control if we want to toggle fuzzy-matching off globally due to test instability.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
georgi-l95: Ah, gotcha. Thank you, I've applied your suggestion, can you take a look again ?

tinker-michaelj (approved):
LGTM, thanks @georgi-l95 !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
tinker-michaelj: Done! 😁

PR #10797, opened 2024-01-05
chore: fix mod-service congestion start time mgmt

### LINKED ISSUE #10794: (mod-service) In-state representation of congestion level starts is inconsistent with multiplier business logic
### Description

In state, congestion level starts are [stored](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/CongestionThrottleService.java#L60) in a singleton of type [`CongestionLevelStarts`](https://github.com/hashgraph/hedera-protobufs/blob/add-pbj-types-for-state/services/state/congestion/congestion_level_starts.proto#L31) whose PBJ generation is,
```
public record CongestionLevelStarts(
    @Nullable List<Timestamp> genericLevelStarts,
    @Nullable List<Timestamp> gasLevelStarts
){}
```

In particular, when we [save congestion level starts to state](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/impl/NetworkUtilizationManagerImpl.java#L167), we will be saving a record with two lists whose length is equal to the **number of congestion level starts that are in effect**. (E.g., if there is no "generic" congestion, the first list will be empty; and if there is gas congestion at multiplier level 2, the second list will be size 2.)

But during a reconnect, when we reset the `ThrottleMultiplier`'s congestion level starts from state [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/impl/NetworkUtilizationManagerImpl.java#L135), this will lead to using an `Instant[]` of congestion level starts whose length is **not equal** to the number of multiplier levels in configuration [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/FeesConfig.java#L28).

So after a node reconnects, it will immediately begin to throw 
```
2024-01-05 05:30:55.438 ERROR 546  HandleWorkflow - An unexpected exception was thrown during handle
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at com.hedera.node.app.fees.congestion.ThrottleMultiplier.updateCongestionLevelStartsWith(ThrottleMultiplier.java:215) ~[HederaNode.jar:?]
	at com.hedera.node.app.fees.congestion.ThrottleMultiplier.updateMultiplier(ThrottleMultiplier.java:82) ~[HederaNode.jar:?]
	at com.hedera.node.app.fees.congestion.CongestionMultipliers.updateMultiplier(CongestionMultipliers.java:50) ~[HederaNode.jar:?]
	at com.hedera.node.app.throttle.impl.NetworkUtilizationManagerImpl.trackTxn(NetworkUtilizationManagerImpl.java:76) ~[HederaNode.jar:?]
```
whenever it reconnected from a state without maximum congestion multipliers in effect. 

### Steps to reproduce

Reconnect a node in a mod-service network and submit transactions using non-superuser payer accounts.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #10794 by keeping the size of congestion level start lists in state identical to the number of multiplier levels from config.


Neeharika-Sompalli (approved):
LGTM !

PR #10817, opened 2024-01-08
chore: allow starting from genesis state with null address book.

### LINKED ISSUE #10448: PCLI Genesis State needs to modify address books.
https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/GenesisPlatformStateCommand.java

Does not modify the current or previous address book when transforming a state into a genesis state.  If the state is being migrated to a different network or a different address book, this is going to cause the AddressBookInitializer to throw an exception. 

https://github.com/hashgraph/hedera-services/blob/644a8f4a39f7555deaeea9de163666847f076454/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/address/AddressBookInitializer.java#L195-L204

Options
1. the AddressBookInitializer needs to be modified to accept genesis states with different address books,
2. the PCLI Genesis State Command needs to nullify the address books. 
3. the PCLI Genesis State Command needs to initialize the state with a given address book.

edward-swirldslabs:
**Description**:

Please review carefully.  The following changes have been made. 
* the `pcli state genesis` command now sets the current and previous address book to null. 
* code paths which threw exception during testing due to null address book in the state have been modified to check for genesis or tolerate the presence of null. 
   * NEEDS Attentive Review
     * the signing weight of the genesis state is set to 0
* The startup sequence will adopt the config.txt when given a genesis state. 
* A bug was fixed in the AddressBookUtils class and parsing an address book can now tolerate config.txt. 
* a boolean flag is set to indicate if the address book was changed and the initial state needs to be updated with the new address book. 
* NOTE: after the AddressBookInitializer runs, the initialState should no longer have a null address book. 

**Related issue(s)**:

Fixes #10448 

**Notes for reviewer**:

**This Manual Testing Was Performed:** 
1. run HashgraphDemo with default settings. 
2. Stop after a saved state has been generated. 
3. generate a genesis state from the saved state using `pcli state genesis <source.swh> <outputDir> <--load path/HashgraphDemo.jar>`
4. copy genesis state to round 0 directory for each node. 
5. delete preconsensus event stream files. 
6. delete other node states. 
7. modify config.txt to have added (or removed) nodes. 
8. Verify that the application starts from genesis.

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/PlatformBuilder.java
@@ -263,6 +263,11 @@ public final class PlatformBuilder {
                                                 .copy());
             }
 
+            // At this point the initial state must have the current address book set.  If not, something is wrong.
+            if (initialState.get().getState().getPlatformState().getAddressBook() == null) {
+                throw new IllegalStateException("The current address book of the initial state is null.");
+            }
+
             final SwirldsPlatform platform = new SwirldsPlatform(
                     platformContext,
                     keysAndCerts.get(selfId),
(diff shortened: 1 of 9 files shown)


alittley (approved):
LGTM

cody-littley (approved):

cody-littley (approved):

alittley (approved):

PR #10821, opened 2024-01-08
chore: Component Documentation

### LINKED ISSUE #9580: Wiring framework documentation
Create documentation for the wiring framework.

poulok:
**Description**:
Creates documentation for the new component framework.

**Notes to Reviewers**
Rendered markdown document: https://github.com/hashgraph/hedera-services/blob/09580-wiring-documentation/platform-sdk/docs/components/componentFramework.md

**Related issue(s)**:
Fixes #9580

**Checklist**


platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+Wire add ons (or transformers) are simple units of logic that transform the data type on a wire to a different type. For
+example, a `WireListSplitter` takes data from a wire with a `List` type and streams the contents of the list, one by
+one, onto another wire. Wire transformers operate on the thread defined in the task scheduler and do not come with their
+own concurrency logic. Some types of transformers are described below:
+
+**Wire Filter:** A predicate that operates on the data type of the output wire. Data is only passed to the filter output wire if the
+predicate passes.
+
+**Wire Transformer:** A function that operates on the data type of the output wire. Data produced by the wire transformer and forwarded on its
+output wire may be a different type that the input.
```
alittley, line 173: "that the input" -> "than the input"

alittley (approved):

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+always pass data using the backpressure methods. Secondary data flows must use `inject()` to avoid deadlock**. Circular
+data flows that use backpressure are identified by the wiring model, which logs a warning if cyclic backpressure is
+detected.
+
+### Secondary Output Wires
+
+Each task schedulers comes with a built in output wire. If additional output wires are necessary, they can be created
+using the task scheduler, but they must be passed to the business logic in the constructor so that they can be invoked
+directly by the business logic. It is a violation of the framework principles to invoke a secondary output wire from
+anywhere outside the business logic.
```
anthony-swirldslabs, line 194: This paragraph here says that it would be a violation to invoke output wires from outside of the business logic. In other words, it basically says that "business logic invokes some output wires."

But somewhere above the document states:
```
Each unit of business logic does not know where it's input comes from or who consumes its output. The threading is handled by its task scheduler, and the connection of inputs and outputs is handled externally. Changes to concurrency and wiring have no impact on the business logic.
```

which basically reads as "business logic has no idea about the specifics of its input and output wires." I.e. the business logic doesn't invoke its output wires directly.

These statements seem to contradict each other.

I suggest to remove mentions of the business logic from this paragraph here because it seems to be talking about the capabilities of the task schedulers only.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+metrics on tasks.
+
+Data coming into the task scheduler goes through a few phases:
+
+1. Unscheduled - the data has been received by the task scheduler and a task has been created, but it has not yet been
+   scheduled for execution.
+2. Scheduled or Unprocessed - the task has been scheduled for execution and may have started execution, but has not
+   yet completed execution.
+3. Processed - the task has been executed (the method bound to the task has been called and the executing thread has
+   returned).
```
anthony-swirldslabs, line 59: What about error handling? Any DLQs, or the pipeline just stalls forever until a unit of data is processed?

In general, does the framework support at-most-once, or at-least-once, or exactly-once input data processing? Is this configurable per a component (or a wire) or fixed for the entire framework?

This would affect the overall performance of the framework. This may also affect how the business logic is written. So I suggest to expand on this topic in this document.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Wiring Model
+
+The wiring model tracks all task schedulers and wires that connect them and is capable of doing useful analysis of the
+resulting graph. Most notably, it can produce mermaid style diagrams that show the flow of data through wires and
+components, detect cyclic backpressure that could result in a deadlock, and verify proper wiring of various task
+schedulers according to their restrictions (for example, a concurrent task scheduler is not permitted to be wired to a
+direct scheduler).
+
+### Wire Add Ons
```
anthony-swirldslabs, line 162: BTW, not directly related to this PR, but if feels like each of the add ons listed below could just be a component with a business logic doing its thing and using a Direct task scheduler to avoid any scheduling overhead. Are there any advantages of introducing a new concept of an add-on? Would it be simpler to only have components in the framework?

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+scheduled for execution according to the task scheduler type and implementation. Task schedulers are described in more
+detail [below](#TaskSchedulers).
+
+Here is a picture of a simple component with one input wire of type `Integer`, and one output wire of type `String`:
+
+![SimpleComponent](step4.drawio.svg)
+
+### Business Logic
+
+Business logic is defined in one or more classes that work together and accept optional data, perform an
```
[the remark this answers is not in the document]
cody-littley:
```suggestion
Business logic is defined in one or more classes that work together and accept data (if configured to have input wires), perform an
```

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+operation, and produce optional output. This framework allows business logic to be defined completely independently
+of the threading model and other business logic. Each unit of business logic does not know where it's input comes from
+or who consumes its output. The threading is handled by its task scheduler, and the connection of inputs and outputs is
+handled externally. Changes to concurrency and wiring have no impact on the business logic.
+
+### Wires
+
+Wires are the construct on which data flows between components. Each component can have zero or more input wires
```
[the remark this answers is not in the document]
cody-littley: In theory it's plausible for a component to have zero input wires, although in practice it's probably going to be unusual. An example of this is the heartbeat timer, whose only job is to push periodic Instants to other components. Another example might be something that has only been partially migrated to the wiring framework, and does not yet take input through wires.

platform-sdk/docs/components/componentFramework.md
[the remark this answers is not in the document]
alittley: I guess that's correct - though I'd argue that such a component is operating outside the framework as it's intended to be used. Something operating on a timer should probably use the heartbeat, which is an input wire with an `Instant`

platform-sdk/docs/components/componentFramework.md
cody-littley: > What about error handling? 

An uncaught exception handler can be provided, and all exceptions that occur during data processing are passed to it. If none is provided, then a default exception handler will log any exception that happens during data processing.

When there is an exception, the task scheduler does not stop or crash, it will just continue to the next item to be processed.

cody-littley: > In general, does the framework support at-most-once, or at-least-once, or exactly-once input data processing? Is this configurable per a component (or a wire) or fixed for the entire framework?

Currently each task provided to a scheduler is executed exactly once.

platform-sdk/docs/components/componentFramework.md
cody-littley: The goal of add ons was to take a pattern that we expect to see repeated in many places, and to write boilerplate code to do it make life easier for an end user. Each of these add ons could in theory be emulated entirely by non-framework code, but do do so would result in lots of repeated code.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+order to prevent circular data flows that could result in a deadlock of backpressure, task schedulers provide a way to
+bypass the backpressure mechanism. The method `inject()` ignores the number of unprocessed tasks and the maximum
+configured value in order to prevent such deadlocks. **If backpressure is desired, the primary path of data flow must
+always pass data using the backpressure methods. Secondary data flows must use `inject()` to avoid deadlock**. Circular
+data flows that use backpressure are identified by the wiring model, which logs a warning if cyclic backpressure is
+detected.
+
+### Secondary Output Wires
+
+Each task schedulers comes with a built in output wire. If additional output wires are necessary, they can be created
```
cody-littley, line 191:
```suggestion
Each task schedulers comes with a built in output wire (sometimes referred to as the "primary output wire"). If additional output wires are necessary, they can be created
```

platform-sdk/docs/components/componentFramework.md
cody-littley: > Each unit of business logic does not know ... who consumes its output

In this scenario, the business logic still doesn't know who is consuming the data, only that the data needs to be passed to a particular lambda function.

In general, it's acceptable for the business logic to know that it is sending data out over a wire. But we should avoid scenarios where the business logic must be aware of who is listening for that data on the other end of the wire.

In order to clarify this statement, how about if we rewrote this sentence

> It is a violation of the framework principles to invoke a secondary output wire from
anywhere outside the business logic.

into the following:

```
A component is the sole owner of its output wires. Only the business logic of a component should be allowed to push data into its output wires. It is a violation of the framework principles to share ownership of a secondary output wire by allowing anything external to the component to push data onto the output wire.
```

cody-littley:
Your documentation broke unit tests and integration tests. Kelly, you really should have tested this markdown harder before pushing.

cody-littley:
Looked into your test flakes. The unit test is a logging test that is currently flaky. Reported the integration test failure to the slack channel for investigation by services.

platform-sdk/docs/components/componentFramework.md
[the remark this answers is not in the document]
poulok: I will update as @alittley suggests. @anthony-swirldslabs makes a good point about an input wire not _technically_ being required, but I think it is true that the intent of the framework is for each component to accept some kind of data.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
+# Component Framework
+
+## Goals
+
+The goals of this framework is to provide a structure for defining software components in which:
```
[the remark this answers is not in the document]
poulok: i changed "goals" to "goal" because the singular goal is to provide a structure, and that structure must do the things below.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. Logical units of business logic can be defined in isolation from other logical units of business logic.
```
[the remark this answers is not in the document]
poulok: updated

platform-sdk/docs/components/componentFramework.md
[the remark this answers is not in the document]
poulok: updated

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Building a Pipeline
+
+The steps below describe how simple components can be created and wired together. This is one example of the order of
+steps.
+
+### Step 1 - Create the Task Scheduler
+
+Create the `TaskScheduler` parameterized with the type of data the component will produce. The `TaskScheduler` comes
+with an `OutputWire` build in.
```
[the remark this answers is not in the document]
poulok: updated

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,200 @@
...
+operation, and produce optional output. This framework allows business logic to be defined completely independently
+of the threading model and other business logic. Each unit of business logic does not know where it's input comes from
+or who consumes its output. The threading is handled by its task scheduler, and the connection of inputs and outputs is
+handled externally. Changes to concurrency and wiring have no impact on the business logic.
+
+### Wires
+
+Wires are the construct on which data flows between components. Each component can have zero or more input wires 
+(restrictions apply based on the concurrency of the task scheduler) and one or more output wires. Components that do not
+produce any data have an output void of type `Void`. Wires are strongly typed to the data they transport. Wires of the
```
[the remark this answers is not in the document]
poulok: updated

platform-sdk/docs/components/componentFramework.md
poulok: updated

platform-sdk/docs/components/componentFramework.md
anthony-swirldslabs: Thanks for the clarifications. I think it's worth adding this information to this document as it describes the contract provided by the framework.

platform-sdk/docs/components/componentFramework.md
poulok: There is a distinction between the primary output wire that comes built into the task scheduler and that the task scheduler sends data on, and secondary output wires which must be invoked directly by the business logic and the task schedulers know nothing about. But in either case, the statement quoted says that the business logic does not know WHO/WHAT provides the input or consumes the output. It is only a violation to invoke SECONDARY output wires from outside of the business logic.

I will try expanding this paragraph to make the distinction more clear.

platform-sdk/docs/components/componentFramework.md
anthony-swirldslabs: I'm not sure about the "lots of repeated code" though. A generic component implementing, say, the filter could be parametrized with a data type-specific Function<>. This same generic component implementation could also expose a public static method to simplify adding this component to the data pipeline.

This way the data pipeline configuration code would only need to implement that single method in the Function<> (which, I presume, it has to do today anyway - and if there are repeated use-cases, then the implementation could obviously be extracted into some common package), and then add a single call to that static method to actually add this "add-on" component.

This way no code would ever be repeated. Yet the framework would be simplified because of the less number of concepts/entities in its architecture.

platform-sdk/docs/components/componentFramework.md
poulok: Excellent point, I will add this to the documentation

platform-sdk/docs/components/componentFramework.md
poulok: I see the merits of both approaches. An add on is nothing more than a direct scheduler that happens to have very simple business logic, so why not make it just that? But I think I lean toward keeping them as is because conceptually, a component as an essential part of the system that performs critical business logic. Something that has a business reason for existing. In my opinion, the logic of add ons does not rise to the level. Their operations are trivial and more of a wire compatibility issue than core business logic, therefore they should not be components. That being said, I am not adamant one way or another.

platform-sdk/docs/components/componentFramework.md
poulok: Added

platform-sdk/docs/components/componentFramework.md
```diff
@@ -188,10 +191,24 @@ detected.
...
-directly by the business logic. It is a violation of the framework principles to invoke a secondary output wire from
-anywhere outside the business logic.
+Each task scheduler comes with a built-in output wire, also referred to as the "primary output wire". The task
+scheduler takes the data produced by each task execution and sends it on the primary output wire. Additional output
+wires may be created, but they do not follow the same paradigm as primary output wires. These are referred to as
+"secondary output wires". Secondary output wires must be invoked by the business logic directly. Task schedulers are
+unaware of such secondary output wires. The business logic still does not know who consumes the data of these secondary
+output wires, but must know that the secondary output wire exists in order to send data on it. Therefore, secondary
+output wires must be provided to the business logic, ideally in the constructor. A component is the sole owner of its
+output wires. Only the business logic of a component should be allowed to push data into its output wires. It is a
```
anthony-swirldslabs, line 201:
```suggestion
secondary output wires. Only the business logic of a component should be allowed to push data into its secondary output wires. It is a
```

anthony-swirldslabs (approved):

platform-sdk/docs/components/componentFramework.md
poulok: modified slightly

anthony-swirldslabs (dismissed):

poulok:
> Looked into your test flakes. The unit test is a logging test that is currently flaky. Reported the integration test failure to the slack channel for investigation by services.

Thank you @cody-littley!

platform-sdk/docs/components/componentFramework.md
poulok: done

jsync-swirlds:
> > Looked into your test flakes. The unit test is a logging test that is currently flaky. Reported the integration test failure to the slack channel for investigation by services.
> 
> Thank you @cody-littley!

Note that at least the Unit Test flake is fixed on develop.

alittley (dismissed):

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Backpressure
+
+Backpressure can be configured for some task schedulers. If available and configured, threads are blocked from
+adding data to a component that has reached the limit of unprocessed tasks. This feature prevents too many tasks from
+building up in a given component and signals the components feeding into it to back off so that it can catch up. In
+order to prevent circular data flows that could result in a deadlock of backpressure, task schedulers provide a way to
+bypass the backpressure mechanism. The method `inject()` ignores the number of unprocessed tasks and the maximum
+configured value in order to prevent such deadlocks. **If backpressure is desired, the primary path of data flow must
+always pass data using the backpressure methods. Secondary data flows must use `inject()` to avoid deadlock**. Circular
```
mxtartaglia-sl, line 188: > **If backpressure is desired, the primary path of data flow must
always pass data using the backpressure methods. Secondary data flows must use `inject()` to avoid deadlock**. 

In the Wire section, the document explains that "Wires are the construct on which data flows between components" 
Now, it introduces the "path of data flows" concept and contrasts a primary data flow with a secondary data flow.
I cannot help wondering how a path of data flow is formed. Let's say I have a multiple input wire/output wire component, What defines a primary path from a secondary one? Are secondary Output wires involved?

mxtartaglia-sl (dismissed):
The document is straightforward and can be understood in the first read.
Depending on the answer for "paths of data flows" if is related to secondary output wires, I would only suggest swapping the sections Backpressure and secondary wires.
The other questions are for my own benefit and, if added they might pollute the explanation with unnecessary implementation details.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,218 @@
...
+detail [below](#TaskSchedulers).
+
+Here is a picture of a simple component with one input wire of type `Integer`, and one output wire of type `String`:
+
+![SimpleComponent](step4.drawio.svg)
+
+### Business Logic
+
+Business logic is defined in one or more classes that work together and accept data, perform an
+operation, and produce optional output. This framework allows business logic to be defined completely independently
```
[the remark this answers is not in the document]
poulok: It would be helpful to look at `BindableInputWire.bind()`. The business logic method is bound to the input wire in a lambda function. The lambda function checks the return of the business logic function and only forwards data to the task scheduler to be put on the output wire if the return is non-null.

The details of the implementation are slightly different than the diagram. I hoped to capture the conceptual flow to give readers enough context to aid them in digging into the details further via the code. Do you think it is worth explaining the nuance of how this works in the document?

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,218 @@
...
+of the threading model and other business logic. Each unit of business logic does not know where its input comes from
+or who consumes its output. The threading is handled by its task scheduler, and the connection of inputs and outputs is
+handled externally. Changes to concurrency and wiring have no impact on the business logic.
+
+### Wires
+
+Wires are the construct on which data flows between components. Each component can have one or more input wires
+(restrictions apply based on the concurrency of the task scheduler) and one or more output wires. Components that do not
```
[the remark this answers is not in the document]
poulok: >both component diagrams present a 1 to 1 relationship between Inputwires and business logic methods. Is my understanding correct? Or can different input wire data be combined before being processed by a business logic method?

Interesting question. If I understand correctly, your question is: Can we have two components that send the same data type to a single component and have both of those input wires bound to the same method in the business logic? I believe this is possible. Each input wire can only be bound to a single method, but I don't see anything preventing the creation of multiple input wires of the same type, each of which are bound to the same method.

>Is the purpose of a multiple input wires configuration to deal with different input data types? or is it to break the business logic into smaller parts? If so, what criteria should be applied to assign business logic methods to the same or different components?

Yes, the purpose is to allow a component to accept different types of data. Many components in the platform require multiple types of input data to perform their job. For example, the orphan buffer must know about events coming into the system as well as the threshold for when an event becomes ancient and should be discarded.

The decision on what to make one component or many components is a judgement call. Each component should have a single purpose, and the business logic required to fulfill that purpose would ideally be a single class, but could be multiple classes if that operation is complex enough to support that.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,218 @@
...
+produce any data have an output of type `Void`. Wires are strongly typed to the data they transport. Wires of the
+same type can be "soldered" or connected to determine the flow of data between components.
+
+### Task Schedulers
+
+Task schedulers are responsible for several things, but the most important job is taking data received on the input
+wire, combining it with the business logic operation to be performed on that data to create a task, and scheduling that
```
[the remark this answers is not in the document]
poulok: >Checking my understanding, the logical queue in the diagram below stores the "tasks," which have been added (scheduled) by the previous component in the chain.

Yes, though some implementations do not store the task at all. Instead, they keep a counter of the number of tasks that have not completed execution and block until the counter is below the threshold. This is why it is a "logical" queue.

>...as a final activity for the task, it should schedule the next component task

Not quite. If the thread trying to schedule a task is blocked because there is backpressure, it will remain blocked until there is room to schedule the task. The thread that completes execution of a task that results in there being enough room for the next task to be scheduled does not schedule the next task. That is done by the thread that was blocked. This is important, because in order for backpressure to work, the thread that is scheduling tasks (and executing tasks earlier in the pipeline) must be stopped from creating even more tasks. The purpose is to stop even more tasks from piling up.

platform-sdk/docs/components/componentFramework.md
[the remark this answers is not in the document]
mxtartaglia-sl: Thanks! It makes sense.
I wanted to know if there was some other filtering responsibility to be added between components or if the next component in the chain needed to handle the optionality of the previous one... That would be worth noting, but
as the diagram is, it captures how it works...

platform-sdk/docs/components/componentFramework.md
```diff
@@ -0,0 +1,218 @@
...
+
+Component 1 and 2 are Sequential Thread components, meaning their tasks are executed on a dedicated thread in the order
+they are received. Thread A sends data to Component 1. The creation and scheduling of the task is performed on Thread A,
+while the execution of the task is performed on Thread B, the component's dedicated thread. The data resulting from the
+execution of this task in Component 1 is combined with the business logic of Component 2 - creating a task - which is
+scheduled for Component 2 by Thread B. Component 2 executes its task and sends the resulting data to both Component 3
+and 5 on Thread C, creating and scheduling their respective tasks.
+
+Components 3 and 4 are Direct task schedulers and execute their tasks on the scheduling thread. Their tasks are both
+executed on Thread C.
```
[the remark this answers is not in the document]
poulok: Not at this time. I wrote [this ticket](https://github.com/hashgraph/hedera-services/issues/10870) to document the work to be done.

platform-sdk/docs/components/componentFramework.md
poulok: I can understand now how the double use of "primary" and "secondary" could be confusing. The primary data flow path mentioned here in the context of back pressure is conceptual and not directly related to primary or secondary output wires. It is the flow of data on which back pressure is required to slow down input to the system. I think it makes sense to use the primary output wires for this flow of data, but it is not required by the framework.

platform-sdk/docs/components/componentFramework.md
```diff
@@ -166,11 +167,13 @@ example, a `WireListSplitter` takes data from a wire with a `List` type and stre
...
-**Wire Filter:** A predicate that operates on the data type of the output wire. Data is only passed to the filter output wire if the
+**Wire Filter:** A predicate that operates on the data type of the output wire. Data is only passed to the filter output
+wire if the
 predicate passes.
 
-**Wire Transformer:** A function that operates on the data type of the output wire. Data produced by the wire transformer and forwarded on its
-output wire may be a different type that the input.
+**Wire Transformer:** A function that operates on the data type of the output wire. Data produced by the wire
+transformer and forwarded on its
+output wire may be a different type than the input.
```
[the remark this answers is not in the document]
poulok: done

platform-sdk/docs/components/componentFramework.md
```diff
@@ -166,11 +167,13 @@ example, a `WireListSplitter` takes data from a wire with a `List` type and stre
 one, onto another wire. Wire transformers operate on the thread defined in the task scheduler and do not come with their
 own concurrency logic. Some types of transformers are described below:
 
-**Wire Filter:** A predicate that operates on the data type of the output wire. Data is only passed to the filter output wire if the
+**Wire Filter:** A predicate that operates on the data type of the output wire. Data is only passed to the filter output
+wire if the
```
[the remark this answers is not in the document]
poulok: done

platform-sdk/docs/components/componentFramework.md
poulok: updated to:

```
**If backpressure is desired, only a single non-cyclic data flow
should use backpressure. Any data flow that forms a cycle should use `inject()` to avoid deadlock.**
```

anthony-swirldslabs (approved):

alittley (approved):

mxtartaglia-sl (approved):

Issue #10820, opened 2024-01-08
Need an explicit TokenType on SingleTransactionRecord


nickpoorman:
### Problem

While serializing the token transaction into specific outputs, it has become clear that we do not have a definitive way to determine which type of token the transaction was performed on.

One needs to first lookup the TokenID to determine from state what the toke type is for certain transaction. This happens in the various token service handlers but that type is not forwarded along on the `TransactionRecord`, `TransactionReciept`, or `SingleTransactionRecord` for later use.

For example, there is a comment on the protobuf for `TokenMintTransactionBody` 
>  If both amount and metadata list get filled, a INVALID_TRANSACTION_BODY response code will be returned.

This implies that the token type can be determined based on if the amount of metadata is set. However, we have a definitive enum in state `com.hedera.hapi.node.base.TokenType` that TokenService uses.

This hasn't been a major issue yet, and one could continue to use the same logic downstream to determine the token type. However, we will need to expose the type once further token types are implemented because determining the type becomes a rather fuzzy operation using the current method.

### Solution

1. Expose TokenType to `SingleTransactionRecordBuilderImpl` as `private TokenType tokenType;`
2. Create a `TokenBaseRecordBuilder` that the  various`Token(Mint|Burn|Wipe|etc...)RecordBuilder` interfaces inherit from.
3. Add a setter for `tokenType` to the `TokenBaseRecordBuilder` interface.
```
public interface TokenBaseRecordBuilder extends SingleTransactionRecordBuilder {
    /**
     * Sets the token type of the token involved in the transaction.
     */
    @NonNull
    TokenBaseRecordBuilder tokenType(final @NonNull TokenType tokenType);
}
```
4. Implement the `tokenType` setter on `SingleTransactionRecordBuilderImpl`.
4. Update the `handle` calls for `Token(Mint|Burn|Wipe|etc...)Handler` to set the token type via the new `tokenType` setter.

PR #10827, opened 2024-01-09
feat: Add explicit `TokenType` to `SingleTransactionRecord`.

jsync-swirlds:
* Added `TokenBaseRecordBuilder`.
* Adjusted token record builders to extend `TokenBaseRecordBuilder`.
* Implemented new method in `SingleTransactionRecordBuilderImpl`.
* Added `TransactionOutputs` to `SingleTransactionRecord`, and filled in `SingleTransactionRecordBuilderImpl.build()`.
* Fixed unit tests where needed.

Resolves: #10820

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionRecord.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionRecord.java
@@ -18,10 +18,12 @@ package com.hedera.node.app.state;
 
 import static java.util.Objects.requireNonNull;
 
+import com.hedera.hapi.node.base.TokenType;
 import com.hedera.hapi.node.base.Transaction;
 import com.hedera.hapi.node.transaction.TransactionRecord;
 import com.hedera.hapi.streams.TransactionSidecarRecord;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.List;
 
 /**
@@ -30,10 +32,15 @@ import java.util.List;
 public record SingleTransactionRecord(
         @NonNull Transaction transaction,
         @NonNull TransactionRecord transactionRecord,
-        @NonNull List<TransactionSidecarRecord> transactionSidecarRecords) {
+        @NonNull List<TransactionSidecarRecord> transactionSidecarRecords,
+        @NonNull TransactionOutputs transactionOutputs) {
     public SingleTransactionRecord {
         requireNonNull(transaction, "transaction must not be null");
         requireNonNull(transactionRecord, "record must not be null");
         requireNonNull(transactionSidecarRecords, "transactionSidecarRecords must not be null");
+        requireNonNull(transactionOutputs, "transactionOutputs must not be null");
     }
+
+    // This is used by BlockStream, and is not serialized.
+    public record TransactionOutputs(@Nullable TokenType tokenType) {}
 }
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
@@ -30,6 +30,7 @@ import com.hedera.hapi.node.base.Timestamp;
 import com.hedera.hapi.node.base.TokenAssociation;
 import com.hedera.hapi.node.base.TokenID;
 import com.hedera.hapi.node.base.TokenTransferList;
+import com.hedera.hapi.node.base.TokenType;
 import com.hedera.hapi.node.base.TopicID;
 import com.hedera.hapi.node.base.Transaction;
 import com.hedera.hapi.node.base.TransactionID;
@@ -73,6 +74,7 @@ import com.hedera.node.app.spi.HapiUtils;
 import com.hedera.node.app.spi.workflows.record.ExternalizedRecordCustomizer;
 import com.hedera.node.app.spi.workflows.record.SingleTransactionRecordBuilder;
 import com.hedera.node.app.state.SingleTransactionRecord;
+import com.hedera.node.app.state.SingleTransactionRecord.TransactionOutputs;
 import com.hedera.pbj.runtime.OneOf;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.common.crypto.DigestType;
@@ -177,6 +179,7 @@ public class SingleTransactionRecordBuilderImpl
     private final ExternalizedRecordCustomizer customizer;
 
     private TokenID tokenID;
+    private TokenType tokenType;
 
     /**
      * Possible behavior of a {@link SingleTransactionRecord} when a parent transaction fails,
@@ -309,7 +312,8 @@ public class SingleTransactionRecordBuilderImpl
         // Log end of user transaction to transaction state log
         logEndTransactionRecord(transactionID, transactionRecord);
 
-        return new SingleTransactionRecord(transaction, transactionRecord, transactionSidecarRecords);
+        return new SingleTransactionRecord(
+                transaction, transactionRecord, transactionSidecarRecords, new TransactionOutputs(tokenType));
     }
 
     public void nullOutSideEffectFields() {
@@ -559,6 +563,13 @@ public class SingleTransactionRecordBuilderImpl
     public SingleTransactionRecordBuilderImpl addTokenTransferList(@NonNull final TokenTransferList tokenTransferList) {
         requireNonNull(tokenTransferList, "tokenTransferList must not be null");
         tokenTransferLists.add(tokenTransferList);
+        return this;
+    }
+
+    @Override
+    @NonNull
+    public SingleTransactionRecordBuilderImpl tokenType(final @NonNull TokenType tokenType) {
+        this.tokenType = requireNonNull(tokenType);
         return this;
     }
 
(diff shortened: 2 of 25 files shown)


hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenCreateRecordBuilder.java
```diff
@@ -46,5 +45,5 @@ public interface TokenCreateRecordBuilder extends SingleTransactionRecordBuilder
      * @param tokenAssociation the token association that is created by auto association
      * @return the builder
      */
-    TokenUpdateRecordBuilder addAutomaticTokenAssociation(@NonNull final TokenAssociation tokenAssociation);
```
nickpoorman, line 49: Why was this returning `TokenUpdateRecordBuilder`? Does the new version also need to be returning `TokenUpdateRecordBuilder` instead of `TokenCreateRecordBuilder`?

nickpoorman: cc @Neeharika-Sompalli @iwsimon

jsync-swirlds: I believe it was a copy/paste error.  The method is implemented in two interfaces, and appeared to return the other interface here unintentionally.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -283,6 +285,8 @@ public SingleTransactionRecord build() {
                 .assessedCustomFees(assessedCustomFees)
                 .automaticTokenAssociations(newAutomaticTokenAssociations)
                 .paidStakingRewards(paidStakingRewards)
+                // @todo(nickpoorman) Add token type to RecordBuilder for BlockStream use.
+                // .tokenType(tokenType);
```
Neeharika-Sompalli, line 289: This needs a change to protobufs right ? Is that already done as part of a HIP ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
```diff
@@ -168,6 +168,7 @@ public void handle(@NonNull final HandleContext context) {
 
         // Update record with newly created token id
         recordBuilder.tokenID(newTokenId);
+        recordBuilder.tokenType(newToken.tokenType());
```
Neeharika-Sompalli, line 171: This is implied from the transaction body. Do we need this in `TokenCreateRecord` ?

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenCreateRecordBuilder.java
Neeharika-Sompalli: Yes agree with @jsync-swirlds . Must be a copy/paste error

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
jsync-swirlds: May need to be done in future work, or may just be done in the messages that replace SingleTransactionRecord for BlockStream; hence the TODO for Nick.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
jsync-swirlds: The request was to add the explicit type in the record builder wherever reasonable so that BlockStream doesn't need to determine the type (and possibly become out of sync with the service in doing so).

Worth noting that the value will not be written to the Record; it will be used in the Builder to determine the correct type of "Output" object to use in Block Stream.

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenCreateRecordBuilder.java
iwsimon: Yes, that was copy/paste error. @jsync-swirlds could you please correct it in your branch?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
Neeharika-Sompalli: I don't understand. I thought that it will be added to `TransactionRecord` object.  So, you mean we don't need to add it to the `build()` here ?

jsync-swirlds: Not necessarily, no.  BlockStream is replacing SingleTransactionRecord and Receipts with other objects (as I understand it; @nickpoorman can clarify better than I can), so the comment here is really just a note to help guide that replacement.

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenCreateRecordBuilder.java
jsync-swirlds: It is corrected.

Neeharika-Sompalli (dismissed):
LGTM ! Thanks @jsync-swirlds. I am just not clear if this will be a field in `TransactionRecord` or just information needed internally, so just in builder. But since, that code is still commented in `SingleTransactionRecordBuilderImpl` not a blocker for this PR.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
Neeharika-Sompalli: @nickpoorman  Will this be a field in `TranscationRecord.proto` or just needed internally ?

jsync-swirlds:
> LGTM ! Thanks @jsync-swirlds. I am just not clear if this will be a field in `TransactionRecord` or just information needed internally, so just in builder. But since, that code is still commented in `SingleTransactionRecordBuilderImpl` not a blocker for this PR.

This is a question more for @nickpoorman , but my understanding is that it should be just used in the builder, with the possible exception of a brief "transition" period as we're migrating from the RecordStream to the BlockStream.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
Neeharika-Sompalli: Understood, so it will be serialized in these protobufs files `TransactionResult` and `TransactionOutputs` for BlockStreams. Also, I assume it will be documented in HIP.

Agreed `TokenType` in record makes sense. I am just not sure if its needed in `TokenCreateRecord` since it can be inferred from the transactionBody by mirror node.

iwsimon (dismissed):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
jsync-swirlds: Per external discussion, `TokenType` will be added to the `SingleTransactionRecod`, rather than `TransactionRecord`.  The comments are not quite in the correct place in the code as a result.

iwsimon (dismissed):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCreateHandler.java
jsync-swirlds: Fixed.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionRecord.java
```diff
@@ -30,10 +32,14 @@
...
+        @NonNull List<TransactionSidecarRecord> transactionSidecarRecords,
+        @NonNull TransactionOutputs transactionOutputs) {
     public SingleTransactionRecord {
         requireNonNull(transaction, "transaction must not be null");
         requireNonNull(transactionRecord, "record must not be null");
         requireNonNull(transactionSidecarRecords, "transactionSidecarRecords must not be null");
+        requireNonNull(transactionOutputs, "transactionOutputs must not be null");
     }
+
+    public record TransactionOutputs(@Nullable TokenType tokenType) {}
```
Neeharika-Sompalli, line 44: We could probably add a comment this is used only by BlockStreams

jsync-swirlds: Done.

Neeharika-Sompalli (approved):
LGTM ! Thanks @jsync-swirlds

PR #10842, opened 2024-01-09
chore: Test fuzzy-matching for token specs

### LINKED ISSUE #10829: Fuzzy match records with mono-service for TokenCreateSpecs

### LINKED ISSUE #10830: Fuzzy match records with mono-service for TokenDeleteSpecs

### LINKED ISSUE #10831: Fuzzy match records with mono-service for TokenManagementSpecs

Neeharika-Sompalli:
Fixes #10829 
Fixes #10830 
Fixes #10831 
Test fuzzy-matching for token specs


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
```diff
@@ -590,6 +591,8 @@ private void matchSingleValues(
                 Assertions.assertTrue(
                         (long) expected - (long) actual >= 0,
                         "AccountNum '" + expected + "' was not greater than '" + actual + mismatchContext.get());
+            } else if ("name".equals(fieldName) && matchModes.contains(NONDETERMINISTIC_TOKEN_NAMES)) {
+                Assertions.assertTrue(expected != null && actual != null, "Token name is null");
```
tinker-michaelj, line 595: What about,
```
Assertions.assertEquals(((String) expected).size(), ((String) actual).size())
```
?

tinker-michaelj (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/records/SnapshotModeOp.java
Neeharika-Sompalli: Yes that makes sense. But it uses a randomNumber, in `salted` method

PR #10845, opened 2024-01-09
chore: don't externalize zero staking rewards

tinker-michaelj:
**Description**:
- Updates to only set the `paid_staking_rewards` field in a record if there are non-zero rewards paid; and only include non-zero rewards in the list (c.f. mono-service behavior [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/context/SideEffectsTracker.java#L295)).
- Mark `propagatesNestedCreations()` as `FULLY_NONDETERMINISTIC` because we are (currently, at least) accepting that mod-service externalizes internal `CONTRACT_CREATION`s in the order they are _attempted_; while mono-service externalizes them in the order they _succeed_.


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

vtronkov (approved):

PR #10884, opened 2024-01-10
fix: change EnityNumber to Contract ID in state defintion

### LINKED ISSUE #10747: Use `ContractID` as the bytecode map key
### Problem

The current `ContractService` schema uses an `EntityNumber` as the bytecode key [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ContractSchema.java#L60).

### Solution

Replace with `ContractID`.

### Alternatives

_No response_

povolev15:
change EnityNumber to Contract ID in state defintion

**Related issue(s)**:

Fixes #10747


kimbor:
Fixed in PR #10958

PR #10885, opened 2024-01-10
chore: use metadata-derived class id for `QueueNode` when available

tinker-michaelj:
**Description**:
 - When a `QueueNode` is serialized, write the class id for its `StateMetadata` instead of the deprecated fixed `CLASS_ID`.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/StateMetadata.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/StateMetadata.java
@@ -41,6 +41,7 @@ public final class StateMetadata<K, V> {
     private static final String ON_DISK_VALUE_SERIALIZER_CLASS_ID_SUFFIX = "OnDiskValueSerializer";
     private static final String IN_MEMORY_VALUE_CLASS_ID_SUFFIX = "InMemoryValue";
     private static final String SINGLETON_CLASS_ID_SUFFIX = "SingletonLeaf";
+    private static final String QUEUE_NODE_CLASS_ID_SUFFIX = "QueueNode";
 
     private final String serviceName;
     private final Schema schema;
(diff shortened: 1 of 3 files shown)


Neeharika-Sompalli (approved):
LGTM ! ty @tinker-michaelj

Neeharika-Sompalli (dismissed):
LGTM ! ty @tinker-michaelj

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeParentRecordHandler.java
```diff
@@ -85,7 +92,18 @@ public void finalizeParentRecord(@NonNull final AccountID payer, @NonNull final
...
+            hbarChanges = hbarChangesFrom(writableAccountStore);
+        } catch (HandleException e) {
+            if (e.getStatus() == FAIL_INVALID) {
+                logHbarFinalizationFailInvalid(
+                        payer,
+                        context.userTransactionRecordBuilder(SingleTransactionRecordBuilder.class),
+                        writableAccountStore);
+            }
+            throw e;
+        }
```
Neeharika-Sompalli, line 106: Since we are adding logs we could also add to `modifiedTokenRelBalance != 0` also

Neeharika-Sompalli (approved):
LGTM !

lukelee-sl (approved):
LGTM.  Only looked classes that are smart contracts related.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeParentRecordHandler.java
tinker-michaelj: Hmm but it's possible to have non-zero net changes to token balances...e.g. on a `TokenBurn`, right?

PR #10897, opened 2024-01-11
fix: fuzzy matching ERCPrecompileSuite #10315

### LINKED ISSUE #10315: Enable fuzzy record matching for ERCPrecompileSuite
Fixed by:

- https://github.com/hashgraph/hedera-services/pull/11164
- https://github.com/hashgraph/hedera-services/pull/11178
- https://github.com/hashgraph/hedera-services/pull/11179
- https://github.com/hashgraph/hedera-services/pull/11183
- https://github.com/hashgraph/hedera-services/pull/11185

JivkoKelchev:
**Description**:

Enable fuzzy matching on ERCPrecompileSuite

**Related issue(s)**:

Fixes #10315

**Notes for reviewer**:

In some cases, we have a big difference in gas between mono and mod. That is because in mono we halt instead of revert.
I tried to equalize this [here](https://github.com/hashgraph/hedera-services/pull/10897/files#diff-79e1f4dd9f0425a611aa1116adf155bfbb3530aa08345d1783c9abbaf4d475d0L53-L55) and [here](https://github.com/hashgraph/hedera-services/pull/10897/files#diff-31c1b058a76fd1ce9b241b55d7627ce1ba6699b16c12ad2eae8363bb58eb3396L49-L52) and here.
Is it OK, or we should just use `HIGHLY_NON_DETERMINISTIC_FEES`?

In mono we return boolean result for NFT approve function.
I'm not sure if [this](https://github.com/hashgraph/hedera-services/pull/10897/files#diff-ca72327aa28c47ca4fb0e57b2ed2a4c3b52fd095f2ba0331965e51d0a5bb3165L87-R111) is correct?

I added `ALLOW_EMPTY_ERROR_MSG` mode, because in mono we don't return msg errors on revert.
Do we need to change it in mod too or just skip matching like [this](https://github.com/hashgraph/hedera-services/pull/10897/files#diff-68b01ef05569f1e768e8636a634bc4b7f544331f181185420cacc8338b899d7aR607-R611)?

We had a flaky TransactionID nonce problem on different tests (when node stake update transaction occur).
So instead of adding NONDETERMINISTIC_NONCE to all tests, I did [this](https://github.com/hashgraph/hedera-services/pull/10897/files#diff-68b01ef05569f1e768e8636a634bc4b7f544331f181185420cacc8338b899d7aR670-R674) normalization of the nonce, when we encountered a node stake transaction.
**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/SystemContractUtils.java
```diff
@@ -73,6 +73,17 @@ public static ContractFunctionResult contractFunctionResultSuccessFor(
...
+    @NonNull
+    public static ContractFunctionResult contractFunctionResultSuccessFor(
+            final long gasUsed, final Bytes result, AccountID senderId) {
+        return ContractFunctionResult.newBuilder()
+                .gasUsed(gasUsed)
+                .contractCallResult(tuweniToPbjBytes(result))
+                .senderId(senderId)
+                .contractID(HTS_PRECOMPILE_MIRROR_ID)
+                .build();
+    }
```
vtronkov, line 85: Just FYI @natanasow @JivkoKelchev - you can have some merge conflicts here https://github.com/hashgraph/hedera-services/pull/10977/files#diff-50bf7a98ab68915e3012c2d296c58315119bc2aca16f6054d74cea7c84db2dd3R85

JivkoKelchev:
All changes are split in several smaller PRs.
https://github.com/hashgraph/hedera-services/pull/11164
https://github.com/hashgraph/hedera-services/pull/11178
https://github.com/hashgraph/hedera-services/pull/11179
https://github.com/hashgraph/hedera-services/pull/11183
https://github.com/hashgraph/hedera-services/pull/11185

PR #10938, opened 2024-01-12
chore: Cherry pick 9628

### LINKED ISSUE #9459: Removal of address validity checks for top-level and internal EVM calls

lukelee-sl:
**Description**:
Cherry pick #9628

**Related issue(s)**:

Fixes #9459

**Checklist**


stoqnkpL (approved):
LGTM! tyvm @lukelee-sl

Nana-EC (approved):

PR #10940, opened 2024-01-12
fix: Reclaim throttle usage for auto account creation

### LINKED ISSUE #10216: Implement reclaiming of throttle usage for failed auto-creations
In mono's `TransferLogic.doZeroSum()` we have [some logic](https://github.com/hashgraph/hedera-services/blob/8edfb51c44b55c59369fc3031307e11f62f71a3e/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java#L190) for reclaiming of throttle usage for failed auto-creations that is setup to leak the usage as a number of CryptoCreates [here](https://github.com/hashgraph/hedera-services/blob/8edfb51c44b55c59369fc3031307e11f62f71a3e/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/StoresModule.java#L183). 

We need to implement this logic in modularized codebase as well.

petreze:
**Description**:
This PR adds the functionality to reclaim the previously reserved throttle usage for failed auto account creations.
The [following](https://github.com/hashgraph/hedera-services/pull/9388) is the implementation in mono project

**Related issue(s)**:

Fixes #10216 

**Notes for reviewer**:
One crucial change was made in `HandleContext`. The class was extended to contain the `NetworkUtilizationManager` in order to be able to tweak the throttle capacity for `CRYPTO_CREATE` from within `CryptoTransferHandler`

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -883,6 +890,11 @@ public RecordListCheckPoint createRecordListCheckPoint() {
         return new RecordListCheckPoint(firstPreceding, lastFollowing);
     }
 
+    @Override
+    public void reclaimPreviouslyReservedThrottle(int n, HederaFunctionality function) {
+        networkUtilizationManager.leakUnusedThrottlePreviouslyReserved(n, function);
```
tinker-michaelj, line 895: We need this method to be a little more involved. 
   1. It should only leak capacity from the consensus throttle when the top-level transaction is a `CryptoTransfer`; because a top-level contract operation---`ContractCall`, `ContractCreate`, `EthereumTransaction`---will not have **reserved** capacity from the `CRYPTO_CREATE` throttle bucket. (I.e., if we always leak capacity here, then a `ContractCall` failing an auto-creation will effectively **reduce** the measured network utilization! Not what we want. 😁)
   2. We need to also leak capacity from the **frontend** throttle for **only the node that submitted this transaction**; c.f. mono-service [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/TransferLogic.java#L189). Reclaiming capacity from the frontend throttle is the main point of this PR, as otherwise client's will still get `BUSY` when submitting `CRYPTO_CREATE`s; even though we have detected that some preceding `CRYPTO_CREATE`'s actually failed, and reclaimed their network utilization from the consensus throttle.

petreze: Isn't the implementation of `HandleContext.reclaimPreviouslyReservedThrottle()` okay being a bit generalized so that you can use it for different `HederaFunctionality` types and to add some restrictions on the current invocation of it in [TransferContextimpl](https://github.com/hashgraph/hedera-services/blob/10216-throttle-usage-auto-creation/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/TransferContextImpl.java#L102), based on your comments

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/TransferContextImpl.java
```diff
@@ -93,7 +96,21 @@ public void createFromAlias(final Bytes alias, final int reqMaxAutoAssociations)
             validateTrue(tokensConfig.autoCreationsIsEnabled(), NOT_SUPPORTED);
         }
         // Keep the created account in the resolutions map
-        final var createdAccount = autoAccountCreator.create(alias, reqMaxAutoAssociations);
+        AccountID createdAccount;
+        try {
+            createdAccount = autoAccountCreator.create(alias, reqMaxAutoAssociations);
+        } catch (HandleException e) {
+            if (getHandleContext().getHederaFunctionality() == CRYPTO_TRANSFER
```
tinker-michaelj, line 103: ⚠️ Hmm...the `HandleContext` given to the `CryptoTransferHandler` is _always_ going to have a functionality of `CRYPTO_TRANSFER`.

And its category will never be `PRECEDING` (only synthetic `CRYPTO_CREATE` records are ever preceding).

_However_, your implementation of `isSelfSubmitted()` will only be true in the exact conditions when we want to reclaim throttle capacity, since synthetic `TransactionBody`'s never have `nodeAccountID` set.

So we can use just,
```
if (getHandleContext().isSelfSubmitted()) {
    final int autoCreationsNumber = numOfAutoCreations() + numOfLazyCreations();
    getHandleContext().reclaimPreviouslyReservedThrottle(autoCreationsNumber, CRYPTO_CREATE);
}
```
and that should be what we want! 🙌

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -669,6 +677,27 @@ default <T> T dispatchRemovableChildTransaction(
     @NonNull
     RecordListCheckPoint createRecordListCheckPoint();
 
+    /**
+     * Returns the current {@link HederaFunctionality}
+     *
+     * @return the Hedera functionality in the context
+     */
+    HederaFunctionality getHederaFunctionality();
```
tinker-michaelj, line 685: This is no longer needed (see comment below).

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -669,6 +677,27 @@ default <T> T dispatchRemovableChildTransaction(
...
+
+    /**
+     * Returns the current {@link TransactionCategory} of the transaction
+     *
+     * @return the current transaction category
+     */
+    TransactionCategory getTransactionCategory();
```
tinker-michaelj, line 692: No longer needed (see below).

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/TransferContextImpl.java
petreze: Ahh got it!
Thanks for clarifying all of that Michael 🙌

tinker-michaelj (approved):
LGTM, ty @petreze !

PR #10941, opened 2024-01-12
chore: add warmup hook

### LINKED ISSUE #10902: Hook preload mechanism into app
To allow preloading entities into memory, these two tasks need to be accomplished:

- [ ] Add a method (e.g. `preload()`) to the `TransactionHandler`. It should have an empty default implementation.
- [ ] Call this method in `PreHandleWorkflowImpl`, probably right after all checks have passed. 

We could also call this method in `HandleWorkflow` before we check the signatures in validate(). If we missed any entities in pre-handle, we would still use the time spent validating signatures. But we are probably not gaining much and should leave this as a possible optimization for later.

netopyr:
Adds a `warmup()` method in `TransactionHandler` with an empty default implementation and calls it from the pre-handle workflow.

Closes #10902


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleWorkflowImpl.java
```diff
@@ -291,10 +293,11 @@ private PreHandleResult expandAndVerifySignatures(
         // 2b. Call Pre-Transaction Handlers
         try {
             // First, perform semantic checks on the transaction
-            dispatcher.dispatchPureChecks(txInfo.txBody());
+            dispatcher.dispatchPureChecks(txBody);
             // Then gather the signatures from the transaction handler
             dispatcher.dispatchPreHandle(context);
-            // FUTURE: Finally, let the transaction handler do warm up of other state it may want to use later (TBD)
+            // Finally, let the transaction handler do warm up of other state it may want to use later
+            dispatcher.dispatchWarmup(context);
```
Neeharika-Sompalli, line 300: So, we need `warmCache` on each store now ? So we can call underlying `VirtualMap.warm`?

Neeharika-Sompalli (approved):
These changes LGTM! But do we need to add `warm` to `ReadableStates` and all Stores ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/prehandle/PreHandleWorkflowImpl.java
netopyr: Yes, at least for the entities we want to warm up.

PR #10944, opened 2024-01-12
feat: Integrate Validations in HAPI tests

### LINKED ISSUE #8949: Integrate Validations in HAPI tests
There are four classes that run validations after all suites have been run. Right now, we do not use them during our HAPI tests, but we need to find a way how we can integrate them.

- BalanceValidation
- TokenBalanceValidation
- TransactionBodyValidation
- RecordStreamValidation

thenswan:
Closes #8949 

This PR adds Validations in HAPI tests.

**Notes for reviewer**:
- implemented `RecordStreamValidationTestDescriptor` in `HapiTestEngine` to run record stream validations;
- reused logic from `SnapshotModeOp#finishLifecycle()` to read record stream data;
- reused logic from `TestBase#recordStreamValidation()` and `TestBase#assertValidatorsPass()` to run record stream validations;
- introduced `RECORD_STREAM_VALIDATION` tag to run record stream validations after every other suite category;
- added `RecordStreamValidator#validateRecordsAndSidecarsHapi()` to run validation specs in the Hapi test way.

 List of Validators and description of fixes that made them pass:
- `BalanceReconciliationValidator`
    - @mhess-swl fixed synthetic record streaming in [this commit](https://github.com/hashgraph/hedera-services/pull/10944/commits/c0dc7d13da2df0bf2e2c9441472acac6921b9981), and I updated `GenesisRecordsConsensusHook#createAccountRecordBuilders()`, so it is passing
- `BlockNoValidator`
    - @Neeharika-Sompalli  fixed record closing issue on restart in [this commit](https://github.com/hashgraph/hedera-services/pull/10944/commits/d63b574fa04305f880fd01bf2e18be74942fd612), so it is passing after `MixedOpsRestartTest` 
- `ExpiryRecordsValidator`
- `TokenReconciliationValidator`
- `TransactionBodyValidator`
    - updated `SingleTransactionRecordBuilderImpl#syncBodyIdFromRecordId()`, so it is passing

Some concerns:
- not sure that's OK if validators are being run after running some suite or spec via Intellij;
- in case record stream validation failed, it might be unclear which suite caused record stream to contain issue-related items, maybe adding record stream validation after each suite will make sense as well;
- if some suite(s) will be set to run in isolated mode - it will break record stream validations.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
```diff
@@ -68,7 +72,10 @@ public void process(@NonNull final TokenContext context) {
         final var blockStore = context.readableStore(ReadableBlockRecordStore.class);
 
         // This process should only run ONCE, when a node receives its first transaction after startup
-        if (blockStore.getLastBlockInfo().consTimeOfLastHandledTxn() != null) return;
+        //        if (blockStore.getLastBlockInfo().consTimeOfLastHandledTxn() != null) return;
```
mhess-swl, line 75: This absolutely has to be part of the code. The only time we can publish synthetic records is on the first consensus timestamp we get after starting up the node. There should be no need to publish more synthetic records later.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
```diff
@@ -68,7 +72,10 @@ public void process(@NonNull final TokenContext context) {
...
+
+        // Not sure about this condition
+        if (blockStore.getLastBlockInfo().migrationRecordsStreamed()) return;
```
mhess-swl, line 78: I _think_ this line shouldn't be necessary. If we ever get past the check for `consTimeOfLastHandledTxn() != null)` then I believe this flag should always be false.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
```diff
@@ -160,6 +167,19 @@ private void createAccountRecordBuilders(
...
+            if (balance != 0) {
+                var accountID = AccountID.newBuilder()
+                        .accountNum(account.accountId().accountNumOrElse(0L))
+                        .shardNum(account.accountId().shardNum())
+                        .realmNum(account.accountId().realmNum())
+                        .build();
+
+                recordBuilder.transferList(TransferList.newBuilder()
+                        .accountAmounts(asAccountAmounts(Map.of(accountID, balance)))
+                        .build());
```
mhess-swl, line 180: Is this for funding the treasury? (If not, what is it for?)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
thenswan: @mhess-swl 
- we never get past the check for `consTimeOfLastHandledTxn() != null`
- a condition that I added works as expected -- synthetic records being published only once after starting the node (but I think there should be a better way)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
thenswan: @mhess-swl 
- we never get past the check for `consTimeOfLastHandledTxn() != null`
- a condition that I added works as expected -- synthetic records being published only once after starting the node (but I think there should be a better way)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
thenswan: @mhess-swl it was added, so`transferList` in the item below from the record stream generated by running a HapiTest network corresponds to the same item from the record stream generated by a standalone mono-service node (hence the balance validation passes).

```
record_stream_items {
  transaction {
    signedTransactionBytes: "\nN\n\021\n\t\b\250\353\231\255\006\020\360\003\022\002\030\002 \002Z9\n\"\022 \n\250\342\020d\306\036\253\206\342\251\301dV[Nz\232AF\020n\nl\320:\2149Z\021\016\222\020\200\200\320\247\244\260\344\261EJ\006\b\366\317\252\340\006\210\001\001"
  }
  record {
    receipt {
      accountID {
        accountNum: 2
      }
    }
    transactionHash: "8\260`\247Q\254\2268L\3312~\261\261\343j!\375\267\021\024\276\aCL\f\307\277c\366\341\332\'N\336\277\347oe\373\325\032\322\361H\230\271["
    consensusTimestamp {
      seconds: 1705407972
      nanos: 258727001
    }
    transactionID {
      transactionValidStart {
        seconds: 1705407912
        nanos: 496
      }
      accountID {
        accountNum: 2
      }
      nonce: 2
    }
    memo: "Synthetic system creation"
    transferList {
      accountAmounts {
        accountID {
          accountNum: 2
        }
        amount: 5000000000000000000
      }
    }
  }
}
```

mhess-swl (dismissed):

Neeharika-Sompalli (approved):
LGTM !

mhess-swl (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/BalanceReconciliationValidator.java
```diff
@@ -54,6 +59,29 @@ public void validateRecordsAndSidecars(final List<RecordWithSidecars> recordsWit
         concurrentExecutionOf(validationSpecs);
     }
 
+    @Override
+    public void validateRecordsAndSidecarsHapi(
+            final HapiTestEnv env, final List<RecordWithSidecars> recordsWithSidecars)
+            throws InvocationTargetException, IllegalAccessException {
+        getExpectedBalanceFrom(recordsWithSidecars);
+        System.out.println("Expected balances: " + expectedBalances);
```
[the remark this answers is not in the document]
thenswan: It was originally there and I didn't change it 😄

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
```diff
@@ -160,6 +178,19 @@ private void createAccountRecordBuilders(
...
```
Neeharika-Sompalli, line 191: We should not be building transferList here

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -286,6 +286,12 @@ private void handleUserTransaction(
             @NonNull final ConsensusEvent platformEvent,
             @NonNull final NodeInfo creator,
             @NonNull final ConsensusTransaction platformTxn) {
+        // Determine if this is the first transaction after startup. This needs to be determined BEFORE starting the
+        // user transaction
+        final var firstConsTimeOfLastBlock = blockRecordManager.firstConsTimeOfLastBlock();
+        final var isFirstTransaction =
+                firstConsTimeOfLastBlock != null && !firstConsTimeOfLastBlock.isAfter(Instant.EPOCH);
+
```
Neeharika-Sompalli, line 294: How can we decide its startup based on `firstConsTimeOfLastBlock()` ? 🤔 @tinker-michaelj  thoughts ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -417,9 +417,16 @@ public SingleTransactionRecordBuilderImpl transactionID(@NonNull final Transacti
...
-                inProgressBody().copyBuilder().transactionID(newTransactionID).build();
-        this.transaction = SingleTransactionRecordBuilder.transactionWith(body);
+        var body = inProgressBody().copyBuilder().transactionID(newTransactionID);
+
+        // for preserving correct data cases
+        if (transaction.body() != null && transaction.body().hasNodeStakeUpdate()) {
+            body.nodeStakeUpdate(transaction.body().nodeStakeUpdate());
+        } else if (transaction.body() != null && transaction.body().hasCryptoCreateAccount()) {
+            body.cryptoCreateAccount(transaction.body().cryptoCreateAccount());
+        }
```
Neeharika-Sompalli, line 427: Why do we need this ?

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/GenesisAccountRecordBuilder.java
```diff
@@ -43,4 +44,14 @@ public interface GenesisAccountRecordBuilder {
...
+
+    /**
+     * Tracks the <b>net</b> hbar transfers that need to be applied to the associated accounts
+     * (accounts are specified in the {@code TransferList} input param)
+     *
+     * @param hbarTransfers the net list of adjustments to make to account balances
+     * @return this builder
+     */
+    @NonNull
+    CryptoTransferRecordBuilder transferList(@NonNull TransferList hbarTransfers);
```
Neeharika-Sompalli, line 56: This feels wrong. We should probably just dispatch the `createAccount` right ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
mhess-swl: @Neeharika-Sompalli is initial funding of the treasury not considered a transfer?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
mhess-swl: I wrote this, but can't remember why I didn't use the last handled transaction timestamp. I'll run some tests to see if I can remember.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
mhess-swl: Ahh yes, @Neeharika-Sompalli should we build this record ☝️  somewhere else? It seemed appropriate to build in the consensus hook to me

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/BalanceReconciliationValidator.java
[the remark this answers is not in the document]
mhess-swl: It's in test-clients anyway, so it's fine :)

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/GenesisAccountRecordBuilder.java
mhess-swl: I thought the circumstances that made these synthetic accounts necessary was a situation where the accounts already _were_ created...?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
mhess-swl: Just a mistake on my part. I pushed a change that uses the last handled txn timestamp instead. Good catch!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
thenswan: @Neeharika-Sompalli for preserving correct data case, without this `TransactionBodyValidator` will fail with 
"Invalid TransactionBody type HederaFunctionality.NONE with record:..."

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
Neeharika-Sompalli: @mhess-swl  we do like this in mono-service. In mod-service I was thinking we don't hard code transferList anywhere. May be we should invoke FinalizeRecordHandler? @tinker-michaelj  thoughts ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
Neeharika-Sompalli: Which test case helps us demonstrate this failure. Adding this case only for these two doesnt seem correct. Is that because they are synthetic transactions that dont have special Record interfaces that implement `SingleTransactionRecord` ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/GenesisRecordsConsensusHook.java
Neeharika-Sompalli: May be this is the only way we can do now.

Neeharika-Sompalli (commented):
Most makes sense to me. but the changes in `SingeTransactionRecordBuilderImpl` I cannot understand the cases why it fails.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
```diff
@@ -295,6 +301,18 @@ public Instant firstConsTimeOfLastBlock() {
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public Instant consTimeOfLastHandledTxn() {
+        final var lastHandledTxn = lastBlockInfo.consTimeOfLastHandledTxn();
+        return lastHandledTxn != null
```
Neeharika-Sompalli, line 311: Do we set this to null ? I thought in migration we set as `Instant.EPOCH`

mhess-swl: Yes you're right, we should also check for `Instant.EPOCH`. I'll fix this

Neeharika-Sompalli (approved):
LGTM ! Great work 🙌  Thanks @thenswan

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
```diff
@@ -346,6 +364,26 @@ public void advanceConsensusClock(@NonNull final Instant consensusTime, @NonNull
...
+     * used to determine if migration records should be streamed
+     *
+     * @param blockInfo the block info object to test
+     * @return true if the given block info has a last handled transaction time that is considered a
+     * 'default' or 'unset' value, false otherwise.
+     */
+    public static boolean isDefaultConsTimeOfLastHandledTxn(@Nullable final BlockInfo blockInfo) {
+        if (blockInfo == null || blockInfo.consTimeOfLastHandledTxn() == null) {
+            return true;
+        }
```
Neeharika-Sompalli, line 378: same here I think @mhess-swl

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
mhess-swl: Ohhh, this is only a defensive check. `consTimeOfLastHandledTxn` shouldn't ever be null, but if it is, we'll still return `Instant.EPOCH`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
mhess-swl: Again, only a defensive check. This method will still return the correct value in the `EPOCH` case, and there's a corresponding test case to validate this.

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/GenesisAccountRecordBuilder.java
Neeharika-Sompalli: I guess we can't dispatch here. 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
Neeharika-Sompalli: Agreed, but if it is null, its a bug we should fix. Just confuses about the possibility of being null.

mhess-swl: I could add a comment if that's preferable. `null` is one of those things that is always a possibility, whether there's code handling it or not 🤷‍♂️ I'd hate for it to fail in production because of an edge case. Or I can take it out if we really want to do that.

jeromy-cannon (approved):

PR #10952, opened 2024-01-12
fix:  Logging configuration updates test 

### LINKED ISSUE #10877: Logging - Parallel execution exception
java.lang.RuntimeException: Error in parallel execution
at com.swirlds.base.test.fixtures.concurrent.internal.ConcurrentTestSupport.submitAndWait(ConcurrentTestSupport.java:158)
at com.swirlds.base.test.fixtures.concurrent.internal.ConcurrentTestSupport.submitAndWait(ConcurrentTestSupport.java:115)
at com.swirlds.base.test.fixtures.concurrent.internal.ConcurrentTestSupport.executeAndWait(ConcurrentTestSupport.java:92)
at com.swirlds.logging.HandlerLoggingLevelConfigStressTest.testWithConfig(HandlerLoggingLevelConfigStressTest.java:98)

mxtartaglia-sl:
**Description**:
Removed HandlerLoggingLevelConfigStressTest
Added HandlerLoggingLevelConfigConcurrentUpdateTest and framework for concurrently test the change of configurations in the logging framework
**Related issue(s)**:

Fixes #10877


platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/util/LoggingTestOrchestrator.java
```diff
@@ -0,0 +1,195 @@
...
+        this.configIndexReference = new AtomicInteger(0);
+        this.reloadingConfigAndAssertionRules = List.of(testScenarios);
+    }
+
+    /**
+     * Randomly selects a scenario
+     */
+    public void randomlySetScenario(HandlerLoggingLevelConfig config) {
+        // Leaves out the first default scenario given that is only for the initial setting
+        int chosenIndex = ThreadLocalRandom.current().nextInt(reloadingConfigAndAssertionRules.size() - 1) + 1;
```
hendrikebbers, line 96: I do not like how random is used here. By doing so in theory we can even miss a test. Maybe we can shuffle the list of scenarios everytime we ran through the complete list?

mxtartaglia-sl: That is a fair point. As a dev, when adding new scenarios, it is hard to know if my scenario was executed or not and be completely sure if passing locally means that it will pass remotely. 

But another thing to factor in is having a time limit for tests... that also would occasionette the possibility of leaving out scenarios if other scenarios take long enough.

What if, like you said, I start by a randomly shuffled list, only execute scenarios once (but all gets executed), and what is randomized is the time where the configuration reloading happens (instead of having a fixed frequence), what do you think of that idea?

mxtartaglia-sl: @hendrikebbers After thinking for a while, I have a few considerations regarding the test:

*Concurrency*:
There isn't a necessity for concurrency. When we load a configuration, asserting things specific to that configuration is necessary. If a configuration is loaded in the middle of the assertions, we must halt assertions related to stale scenarios. To address this, I initially suggested using a lock. However, the code becomes sequential with no benefits from concurrency. Alternatively, we can cancel the checks when the configuration is stale. But, identifying stale scenarios and canceling the related assertions introduces unnecessary complexity.

*Aditional Scenario*
Creating a test scenario where a logger logs a simple statement concurrently while the configuration is changed under the hood would be beneficial. Unfortunately, I haven't yet found a straightforward way to implement this scenario.

*Volume*
The replaced test focused on volume, even though there were no specific assertions. The intention behind the replaced test was to introduce a significant volume while methods were invoked.

Considering these things, I've rewritten the test, removed the concurrency support, introduced a controlled volume with a duration cap, and relocated the code to an existing test class since the test no longer serves as a stress test and no longer involves concurrency.

hendrikebbers (approved):

PR #10965, opened 2024-01-12
fix: (mod-service)Use concurrent read cache

Neeharika-Sompalli:
Root cause is the [use of a ](http://java/com/hedera/node/app/spi/state/ReadableKVStateBase.java#L41)HashMap readCache[ in ](http://java/com/hedera/node/app/spi/state/ReadableKVStateBase.java#L41)ReadableKVStateBase results a race between (1) thread A that is trying to get an already-read key and (2) thread B that is trying to add a newly-read key and hence resizing the HashMap. This can lead to ISS. 

To solve the issue used `ConcurrentHashMap` or `SynchronizedMap` instead of `HashMap` for readCache. Since we want to store null values in readCache if the entity doesn't exist in state, and `ConcurrentHashMap` doesn't allow null values decided to use `SynchronizedMap`.

This will be changed to use a Marker object instead of null value in the future based on performance testing.


tinker-michaelj (approved):

Issue #9886, opened 2023-11-14
Implement HIP-796 `HapiSpecOperation` primitives


tinker-michaelj:
### Problem

We need to implement the `HapiSpecOperations` that facilitate the new HIP-796 `rpc` calls in HAPI:
- [ ] `rpc lockToken (Transaction) returns (TransactionResponse)`
- [ ] `rpc unlockToken (Transaction) returns (TransactionResponse)`
- [ ] `rpc createTokenPartitionDefinition (Transaction) returns (TransactionResponse)`
- [ ] `rpc updateTokenPartitionDefinition (Transaction) returns (TransactionResponse)`
- [ ] `rpc deleteTokenPartitionDefinition (Transaction) returns (TransactionResponse)`

### Solution

Implement these.

PR #10966, opened 2024-01-15
feat(HIP796): Implement HIP-796 HapiSpecOperation primitives #9886 (updasteTokenPartitionDefinition only)

vvairava:
**Description**
To update TokenCreatePartitionDefinition for the particular TokenId.

**The PR has the following files changes**

**Existing File**
hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/TxnVerbs.java

**New File**
hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiPartitionUpdate.java

**Review comments changes for PR# - 10759**
- Added final keyword on token declaration
- Used memo element from from the parent HapiSpecOperation class

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiPartitionCreate.java

**Compile**
BUILD is SUCCESS

### DIFF
--- a/hedera-node/hapi/build.gradle.kts
+++ b/hedera-node/hapi/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
  *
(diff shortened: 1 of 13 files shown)


thomas-swirlds-labs (approved):
LGTM 👍

PR #10975, opened 2024-01-16
fix: Enable tests from Issue2319Spec

### LINKED ISSUE #8772: Resolve `Issue2319Spec` errors
### Problem

We have the following tests suite `Issue2319Suite` that has problems when it's running against modularization codebase.

### Solution

### Alternatives

_No response_

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #8772

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/issues/Issue2319Spec.java
```diff
@@ -44,28 +44,28 @@
...
 
     @HapiTest
-    @Disabled("Failing or intermittently failing HAPI Test")
     final HapiSpec propsPermissionsSigReqsWaivedForAddressBookAdmin() {
-        var pemLoc = "<PEM>";
-
         return defaultHapiSpec("PropsPermissionsSigReqsWaivedForAddressBookAdmin")
                 .given(
-                        keyFromPem(pemLoc).name("persistent").simpleWacl().passphrase("<SECRET>"),
+                        keyFromPem(pemLoc).name("persistent").passphrase(passphrase),
```
mhess-swl, line 68: Curious, why was `.simpleWacl()` removed?

Ivo-Yankov: I could not get it to properly load the key from the pem file with `.simpleWacl()` and the test failed with `INCORRECT_SIGNATURE`. It's possible that the format of the pem file is wrong, but I could not find documentation on how it should be generated.

tinker-michaelj (changes requested):
I can't see why we need to load a _specific_ key to use as the non-default special file admin key here.

We can just use,
```
     newKeyNamed(NON_TREASURY_KEY),
     newKeyListNamed(NON_TREASURY_ADMIN_KEY, List.of(NON_TREASURY_KEY)),
```
and then update the system files to verify signing requirements are waived as expected.

We _should_ then replace the admin key on these files to the default key, however. Note the default file admin key is a `KeyList` containing just the `GENESIS` key, so you can get it like,
```
newKeyListNamed(DEFAULT_ADMIN_KEY, List.of(GENESIS)),
```

Let me know if more details are needed. 💯

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/issues/Issue2319Spec.java
mhess-swl: @tinker-michaelj or @Neeharika-Sompalli this test fails against mono service locally. Not sure if it's run in CI or not. Is it ok to change these tests?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/issues/Issue2319Spec.java
```diff
@@ -44,28 +44,28 @@
 public class Issue2319Spec extends HapiSuite {
     private static final Logger log = LogManager.getLogger(Issue2319Spec.class);
 
+    private String pemLoc = "src/main/resource/randomCivilianKey.pem";
+    private String passphrase = "passphrase";
```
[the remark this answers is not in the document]
georgi-l95: Thanks. Addressed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/issues/Issue2319Spec.java
georgi-l95: Thanks. Addressed.

tinker-michaelj (approved):
LGTM,  ty @Ivo-Yankov !

PR #10977, opened 2024-01-16
test: enable `TokenAndTypeCheckSuite` fuzzy matching

### LINKED ISSUE #10974: Enable fuzzy record matching for TokenAndTypeCheckSuite

natanasow:
**Description**:

**Related issue(s)**:

Fixes #10974

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/HtsSystemContract.java
```diff
@@ -113,11 +113,7 @@ private static FullResult resultOfExecuting(
                             .systemOperations()
                             .externalizeResult(
                                     contractFunctionResultSuccessFor(
-                                            pricedResult.fullResult().gasRequirement(),
-                                            output,
-                                            frame.getRemainingGas(),
-                                            frame.getInputData(),
-                                            attempt.senderId()),
+                                            pricedResult.fullResult().gasRequirement(), output, attempt.senderId()),
```
tinker-michaelj, line 116: ⚠️ We sometimes want to include `gas`, `amount`, `functionParameters`; and sometimes not. (See mono-service [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/Precompile.java#L118) and [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/utils/PrecompileUtils.java#L57.).)

This was fixed in [this PR](https://github.com/hashgraph/hedera-services/commit/a05019567c0abdb58003f9bfe8b3ba359686eb09).

You should be able to revert this change as well as in `SystemContractUtils`, merge `develop` and this should pass.

tinker-michaelj (changes requested):
See comment---we want to include the extra fields in `contractCallResult` for some system contracts (those that can't run in a static frame); but exclude in others.

tinker-michaelj (approved):
LGTM, ty @natanasow !

PR #11000, opened 2024-01-16
fix: Flush transaction handling after replay

### LINKED ISSUE #10923: Flush transaction handling after PCES replay
- Reconnect tests [sometimes fail](http://35.247.76.217:8095/alittley/10841-direct-hashing/4N/Reconnect/20240110-222006-GCP-Daily-Reconnect-4N/VMerkleDual-NDReconnect-500-60m/swirlds-logs.html) when clearing the transaction handling thread for reconnect after a large replay
- After a discussion with @cody-littley, we came to the conclusion that it makes sense to flush transaction handling as part of the replay process, so that a node coming online is fresh and ready to go when replay completes

alittley:
- closes #10923


lpetrovic05:
Does this fix the JRS test?

alittley:
> Does this fix the JRS test?

@lpetrovic05 Yes. I am still working on getting to the bottom of why exactly the problem is arising. I think this is a good change regardless, but I want to make sure it isn't papering over a problem

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
```diff
@@ -831,6 +832,17 @@ public class SwirldsPlatform implements Platform {
...
 
+    /**
+     * Wait until the consensus round handler is not busy.
+     */
+    private void waitUntilTransactionHandlingThreadIsNotBusy() {
+        try {
+            consensusRoundHandler.waitUntilNotBusy();
+        } catch (final InterruptedException e) {
+            throw new RuntimeException("Interrupted waiting for transaction handling thread to not be busy", e);
+        }
```
cody-littley, line 843:
```suggestion
        } catch (final InterruptedException e) {
            Thread.currentThread().interrupt();
            throw new RuntimeException("Interrupted waiting for transaction handling thread to not be busy", e);
        }
```

It's best practice to always re-interrupt the current thread if you catch an interrupted exception and rethrow it as something else.

alittley: Done

cody-littley (approved):

lpetrovic05:
> > Does this fix the JRS test?
> 
> @lpetrovic05 Yes. I am still working on getting to the bottom of why exactly the problem is arising. I think this is a good change regardless, but I want to make sure it isn't papering over a problem

In that case also close the JRS ticket I opened

lpetrovic05 (approved):

Issue #10961, opened 2024-01-12
Virtual hashing is the next bottleneck


artemananiev:
Current performance bottleneck in low level virtual map benchmarks is `VirtualHasher`. It's a smart and somewhat complicated algorithm, but it has a few known drawbacks:

* Sub-trees are hashed one by one. In reality, the next sub-tree can often be hashed before the previous one is completed
* Higher ranks in every sub-tree are processed by just a few threads, not all available hashing threads
* If hash chunks (#5286) are ever implemented, the current approach is not very chunk-friendly

@OlegMazurov proposed a new approach to virtual hashing, task based. It allows max parallelism and can be made chunk friendly

PR #11001, opened 2024-01-16
feat: 10961: Virtual hashing is the next bottleneck

artemananiev:
Fixes: https://github.com/hashgraph/hedera-services/issues/10961


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -97,78 +95,57 @@ protected AbstractHashListener(
...
-        Stream<VirtualLeafRecord<K, V>> sortedDirtyLeaves = Stream.of();
-        for (int i = batchLeaves.size() - 1; i >= 0; i--) {
-            final List<VirtualLeafRecord<K, V>> batch = batchLeaves.get(i);
-            if (!batch.isEmpty()) {
-                sortedDirtyLeaves = Stream.concat(sortedDirtyLeaves, batch.stream());
-                maxPath = Math.max(maxPath, batch.get(batch.size() - 1).getPath());
-            }
+    public synchronized void onHashingCompleted() {
+        if (!nodes.isEmpty() || !leaves.isEmpty()) {
+            flush(nodes, leaves);
```
anthony-swirldslabs, line 136: At line 124 you call flush() outside of the `synchronized (this)` block, while here you call it implicitly inside this block because the method is synchronized. I think it would be safer if the synchronization contract for this method is the same for every call. This would ensure that we take a lock if one is required, or otherwise, avoid locking if it's unnecessary for running this method. I see that the method performs some I/O, so I'd guess you may want to release the lock before calling it. If my guess is wrong though, then it would help to add a comment in code here explaining why the lock is needed here specifically.

anthony-swirldslabs (commented):
I'll review the VirtualHasher.java once the PR is out of the Draft state.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -198,602 +145,312 @@ public Hash hash(
...
+        void push() {
+            if (count.decrementAndGet() == 0) {
+                forkIt();
+            }
+        }
 
-        // Let the listener know we have started hashing.
-        listener.onHashingStarted();
+        void push(final int p) {
+            if (count.updateAndGet(value -> value - p) == 0) {
```
OlegMazurov, line 178: use `addAndGet(-p)`

artemananiev: Done

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -97,78 +95,57 @@ protected AbstractHashListener(
...
-    /**
-     * {@inheritDoc}
-     */
     @Override
-    public void onRankStarted() {
-        rankLeaves = new ArrayList<>(INITIAL_BATCH_ARRAY_SIZE);
-        rankNodes = new ArrayList<>(INITIAL_BATCH_ARRAY_SIZE);
+    public void onHashingStarted() {
+        nodes = new ArrayList<>();
+        leaves = new ArrayList<>();
```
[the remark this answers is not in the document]
artemananiev: Agree. Fixed

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
artemananiev: I see no requirements for flush() to be called inside a synchronized block or not, other than performance. In `onNodeHashed()` it is called without synchronization to release the monitor (shared across all hashing tasks, as all of them use the same hashing listener) as soon as possible. In `onHashingCompleted()` this is not an issue, since hashing is completed, and no hashing tasks are running. I can change it to be consistent with `onNodeHashed()`, though...

artemananiev: Done

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -198,62 +141,189 @@ public Hash hash(
...
+            builder.update(right);
+            return builder.build();
+        }
+
+        private int getIndexInOut() {
+            if (path == 0) {
+                return 0;
+            }
+            long outPath = out.path;
+            for (int i = 0; i < out.height; i++) {
```
OlegMazurov, line 307: This loop could and should be avoided. `getLeftChildPath(path)` is just `(path << 1) + 1` or equivalently `(path << 2) - 1`. Child's path at depth n is thus `(path << (n + 1)) - 1`. This can be added to `Path` as a new API: `Path.getLeftChildPath(path, depth)`.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -198,62 +141,189 @@ public Hash hash(
...
+        private ChunkHashTask out;
+
+        private final Hash[] ins;
+
+        private VirtualLeafRecord<K, V> leaf;
+
+        // isLeaf is just an optimization: if the task is known to be used as a leaf task,
+        // there is no need to allocate an array of hashes that will always be empty
+        ChunkHashTask(long path, int height, boolean isLeaf) {
+            super(1 + (1 << height));
```
OlegMazurov, line 203: We can set the number of dependencies conditionally:
```
            super(1 + (isLeaf ? 1 : (1 << height)));
```
and use `push()` in `setLeaf(leaf)` eliminating a need in `push(p)`.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -265,535 +335,134 @@ public Hash hash(
...
-
-            // Setup and hash the subtree that we have accumulated. If I am hashing the minLeafRank,
-            // then the results go into the minRankStopQueue. If I am hashing the maxLeafRank, then
-            // the results go into the maxRankStopQueue.
-            final HashingQueue<K, V> pq = queue2.reset();
-            final HashingQueue<K, V> sq = rank == maxLeafRank ? maxRankStopQueue : minRankStopQueue;
-            listener.onBatchStarted();
-            hashSubTree(hashReader, listener, wq, pq, null, sq, firstLeafPath, lastLeafPath, rank, stopRank);
-            listener.onBatchCompleted();
+        final HashMap<Long, ChunkHashTask> map = new HashMap<>();
```
OlegMazurov, line 348: We can try to run this code as a task in the pool ensuring all created tasks will be run in the pool as well.
Currently, profiling shows that we `execute()` many tasks created here and that is expensive. By making the current thread to be part of the pool we'll effectively use `fork()` instead.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -131,21 +131,36 @@ public synchronized void onLeafHashed(final VirtualLeafRecord<K, V> leaf) {
...
+        synchronized (this) {
+            if (!nodes.isEmpty() || !leaves.isEmpty()) {
+                needFinalFlush = true;
+            }
+        }
+        if (needFinalFlush) {
             flush(nodes, leaves);
         }
         nodes = null;
         leaves = null;
```
anthony-swirldslabs, line 145: Since elsewhere in this class we always assign values to these fields under the `this` lock, I suggest to modify the implementation as follows:

```suggestion
        final List<VirtualLeafRecord<K, V>> leaves = this.leaves;
        final List<VirtualHashRecord> nodes = this.nodes;
        synchronized (this) {
            nodes = null;
            leaves = null;
        }
        if (!nodes.isEmpty() || !leaves.isEmpty()) {
            flush(nodes, leaves);
        }
```

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: `push(p)` is now eliminated.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: I introduced two new methods to `Path`: to get grand-parent path and to get left-most grand-child path, both are used in virtual hasher. Unit tests for both new methods are added, too.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: As we discussed offline, we will get this for free, once merkle tree hashing is changed to task based with its own fork-join pool.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
artemananiev: OK, I will fix this. Note that this.nodes and this.leaves should be accessed and modified in a synchronized block.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/config/VirtualMapConfig.java
```diff
@@ -82,6 +89,8 @@ public record VirtualMapConfig(
         @Min(0) @Max(100) @ConfigProperty(defaultValue = "50.0")
                 double percentHashThreads, // FUTURE WORK: We need to add min/max support for double values
         @Min(-1) @ConfigProperty(defaultValue = "-1") int numHashThreads,
+        @Min(1) @Max(64) @ConfigProperty(defaultValue = "6") int virtualHasherChunkHeight,
+        @Min(0) @ConfigProperty(defaultValue = "500000") int reconnectFlushInterval,
```
anthony-swirldslabs, line 93: It's best to include the units of the value in the name of a variable. E.g. `reconnectFlushIntervalInNodes'. This is especially important for this specific parameter because the noun "interval" is usually reserved for temporal units, so the current name of this parameter is very confusing.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -150,19 +109,7 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
...
-    public VirtualHasher() {
-        // These queues are used for the "workingQueue" (wq), "pendingQueue" (pq), "stopQueue" (sq),
-        // and "lastQueue" (lq). Which queue is which changes during execution.
-        this.queue1 = new ArrayHashingQueue<>();
-        this.queue2 = new ArrayHashingQueue<>();
-        this.maxRankStopQueue = new ArrayHashingQueue<>();
-        this.lastQueue = new ArrayHashingQueue<>();
-        this.minRankStopQueue = new ArrayHashingQueue<>();
-    }
+    private static final ForkJoinPool hashingPool = new ForkJoinPool(HASHING_THREAD_COUNT);
```
imalygin, line 112: Canonically static variables should be capitalized.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -198,62 +145,142 @@ public Hash hash(
...
+
+        private final int height; // 1 for 3-node chunk, 2 for 7-node chunk, and so on
+
+        private ChunkHashTask out;
+
+        private final Hash[] ins;
+
+        private VirtualLeafRecord<K, V> leaf;
+
+        // isLeaf is just an optimization: if the task is known to be used as a leaf task,
```
imalygin, line 160: This comment seems to be no longer relevant

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -97,80 +107,85 @@ protected AbstractHashListener(
...
+    }
 
-        // flush it down
+    // Since flushes may take quite some time, this method is called outside synchronized blocks,
+    // otherwise all hashing tasks would be blocked on listener calls until flush is completed.
+    private void flush(
+            @NonNull final List<VirtualHashRecord> hashesToFlush,
+            @NonNull final List<VirtualLeafRecord<K, V>> leavesToFlush) {
+        if (!flushInProgress.compareAndSet(false, true)) {
+            throw new IllegalStateException("Cannot start flushing, flush already in progress?");
```
anthony-swirldslabs, line 169: Minor: `onNodeHashed` above checks the atomic flag w/o setting it. If the `onNodeHashed` is allowed to be called from multiple threads, then one of the threads might end up throwing an exception here even though a prior check in `onNodeHashed` succeeded because another thread might have started the flush in the meantime.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -97,80 +107,85 @@ protected AbstractHashListener(
...
+        }
         try {
-            dataSource.saveRecords(
-                    firstLeafPath, lastLeafPath, sortedDirtyHashes, sortedDirtyLeaves, findLeavesToRemove(maxPath));
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+            final long maxPath = leavesToFlush.stream()
```
anthony-swirldslabs, line 172: Minor: `parallelStream()` ?

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -76,72 +74,33 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
...
     /**
-     * The min stop queue. It is kept and passed between {@link VirtualMap} copies, with the ASSUMPTION
-     * that only a single copy of the {@link VirtualMap} is being hashed at a time (which should be true).
-     * The reason for keeping the queue and passing it around is to cut down on garbage and temporary allocations.
-     * <p>
-     * There are two stop queues. One is used for jobs that are ultimately the result of processing starting
-     * from the max rank, and the other is for jobs that are ultimately the result of processing starting from
-     * the min rank.
+     * A listener to notify about hashing events. This listener is stored in a class field to
+     * acoid passing it as an arg to every hashing task.
```
anthony-swirldslabs, line 91: s/acoid/avoid/

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -198,62 +145,142 @@ public Hash hash(
...
+    }
+
     public Hash hash(
             final LongFunction<Hash> hashReader,
             final Iterator<VirtualLeafRecord<K, V>> sortedDirtyLeaves,
             final long firstLeafPath,
             final long lastLeafPath,
             VirtualHashListener<K, V> listener) {
 
-        // Hashing algo v5. This version of the hashing algorithm is designed to optimally process
```
anthony-swirldslabs, line 227: It would be nice to have some text describing how this new implementation actually works.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -265,535 +292,130 @@ public Hash hash(
...
+        int rootTaskHeight = Math.min(firstLeafRank, chunkHeight);
+        ChunkHashTask rootTask = new ChunkHashTask(hashingPool, ROOT_PATH, rootTaskHeight);
+        rootTask.setOut(resultTask);
+        map.put(ROOT_PATH, rootTask);
+
+        boolean firstLeaf = true;
+        final long[] stack = new long[lastLeafRank + 1];
+        Arrays.fill(stack, INVALID_PATH);
+
+        final int[] parentRankHeights = new int[256];
```
anthony-swirldslabs, line 318: Minor: not sure where this magic number comes from. Would be nice to either have a comment here in code, or put the number in a well-named constant.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -265,535 +292,130 @@ public Hash hash(
...
-        while (queue.size() < segmentSize) {
-            readLeavesInSegment(itr, queue, Math.min(segmentStart + segmentSize, eofPath));
-            // If we have read as much as is available to read, then bail.
-            if (!itr.hasNext() || itr.peek().getPath() >= eofPath) {
-                return;
-            } else {
-                segmentStart += segmentSize;
-            }
-        }
+        listener.onHashingCompleted();
```
anthony-swirldslabs, line 409: Minor: should this be in a sort of a `finally{}` block of some kind to ensure the listener's state isn't corrupt in case the above code throws an exception or otherwise fails the task?

anthony-swirldslabs (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: Good catch. I'll fix this

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
artemananiev: OK

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
artemananiev: This flag is not to protect from two (or more) threads handling `onNodeHashed()` in parallel. For this purpose, there is a synchronized block: only one of the threads may have `hashes.size() > reconnectFlushInterval`, while the other will have `hashes` empty. The flag is to handle a different case. Assume a flush is started (outside of the synchronized block), but it takes so long that there are reconnectFlushInterval (or more) hashes ready to flush again. The underlying MerkleDb is not ready to handle two parallel flushes yet, this is why the flag was introduced

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/config/VirtualMapConfig.java
artemananiev: There is a similar config `flushInterval` (which should be deprecated and eventually removed), which is also not in any time units, but in rounds. I added a JavaDoc comment for this new param, please, check if it's confusing and/or not enough

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
anthony-swirldslabs: Yes, I understand that. My point is that that synchronized block calls the atomic.get() method only. However, the actual call to flush() happens outside of the block. So theoretically, there's a chance of calling flush() while another thread executing the onNodeHashed() would call it again because the flag hasn't been set yet.

imalygin (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: OK, I'll add a short comment

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: This is no different from the current implementation. My observation is if hashing fails, this is an unrecoverable error, and there is no need to call `onHashingCompleted()`

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/config/VirtualMapConfig.java
anthony-swirldslabs: I have seen the javadoc. My comment is specifically about the name of the parameter(s).

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
anthony-swirldslabs: Oh, I see. Sounds good then.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -297,50 +297,106 @@ public Hash hash(
...
 
         boolean firstLeaf = true;
         final long[] stack = new long[lastLeafRank + 1];
         Arrays.fill(stack, INVALID_PATH);
 
-        final int[] parentRankHeights = new int[256];
+        // Tasks may have different heights. The root task has a default height. If the whole
+        // virtual tree has fewer ranks than the default height, the root task will cover all
+        // the tree (almost all, see comments below about leaf task heights)
+        final int[] parentRankHeights = new int[256]; // assuming there may be no more than 256 ranks in the tree
```
anthony-swirldslabs, line 350: That's an interesting assumption. I hope we never get to this depth of the tree. On the other hand, I'm wondering if lowering this assumed value even further (128? 64? 32? ...16? ......8?) could bring further performance improvements to the algorithm?

Also, is this assumption enforced anywhere? Or at least asserted (preferably, with a real if/throw) anywhere? If/when it is asserted, that would be a time to move the number to a shared constant then.

artemananiev: No, there are no benefits (from algorithm perspective) to have this array smaller. There are no loops like from 0 to array length.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: Some comments are provided

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
anthony-swirldslabs: Sounds good.

imalygin (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -156,6 +158,8 @@ public void onHashingCompleted() {
             leaves = null;
         }
         if (!finalNodesToFlush.isEmpty() || !finalLeavesToFlush.isEmpty()) {
+            assert !flushInProgress.get() : "Flush must not be in progress when hashing is complete";
+            flushInProgress.set(true);
             flush(finalNodesToFlush, finalLeavesToFlush);
```
anthony-swirldslabs, line 163: This should probably use compareAndSet or similar. It seems a bit fragile to just set the flag unconditionally.

anthony-swirldslabs (approved):

imalygin (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/config/VirtualMapConfig.java
anthony-swirldslabs: I'm not sure if this is actually resolved, but oh well...

OlegMazurov (approved):

PR #11004, opened 2024-01-17
chore: skip unwanted parts of init during mono event stream recovery

tinker-michaelj:
**Description**:
 - Reduces friction in running mono-service `EVENT_STREAM_RECOVERY` by not trying to create a GUI console or start Netty.

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/grpc/GrpcStarter.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/grpc/GrpcStarter.java
@@ -16,9 +16,14 @@
 
 package com.hedera.node.app.service.mono.grpc;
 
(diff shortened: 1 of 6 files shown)


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/grpc/GrpcStarter.java
```diff
@@ -42,18 +48,24 @@ public class GrpcStarter {
...
         this.grpc = grpc;
         this.addressBook = addressBook;
         this.nodeLocalProperties = nodeLocalProperties;
     }
 
     public void startIfAppropriate() {
+        // Never any reason to start Netty during event stream recovery
+        if (initTrigger == EVENT_STREAM_RECOVERY) {
+            return;
+        }
```
mhess-swl, line 68: good idea 👍

mhess-swl (approved):
LGTM! Here's to making event recovery less painful 🥂

PR #11005, opened 2024-01-17
chore: refactor to remove dependencies to contractMustBePresent

### LINKED ISSUE #10985: Refactor ProxyWorldUpdater to remove `contractMustBePresent`
### Problem

The field `contractMustBePresent` was introduced in the proxy updater to track whether calls to contract addresses which do not exist are allowed.  This defuses the state and thus can be difficult to track.

### Solution

Refactor to use a dagger provided VersionFeatureFlag to centralize control.

### Alternatives

_No response_

lukelee-sl:
**Description**:
A new field was introduced to the proxyWorldUpdater.  Remove this field and use the dagger injected FeatureFlag to determine if invalid addresses can be called.

**Related issue(s)**:

Fixes #10985 

Lots of files touched but most of it is boilerplate updates to pass in `FeatureFlag` to various evm operations that needs to check for address validity.

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCallOperation.java
```diff
@@ -88,8 +88,7 @@ private boolean mustBePresent(@NonNull final MessageFrame frame, @NonNull final
         }
         // Let system accounts calls or if configured to allow calls to non-existing contract address calls
         // go through so the message call processor can fail in a more legible way
-        return !addressChecks.isSystemAccount(toAddress)
-                && proxyUpdaterFor(frame).contractMustBePresent();
+        return !addressChecks.isSystemAccount(toAddress) && contractRequired(frame, toAddress, featureFlags);
```
david-bakin-sl, line 91: so ... if it _is_ a system account does _not_ need to be present
and then
if it _is_ a "contract required" address then does need to be present

had to write that down to figure out what was going on, and if it matched the comment

I found this rather tricky to read.  In fact I'm still not sure I'm reading it correctly.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomExtCodeSizeOperation.java
```diff
@@ -53,7 +58,7 @@ public OperationResult execute(@NonNull final MessageFrame frame, @NonNull final
                 frame.pushStackItem(UInt256.ZERO);
                 return new OperationResult(cost(true), null);
             }
-            if (proxyUpdaterFor(frame).contractMustBePresent() && !addressChecks.isPresent(address, frame)) {
+            if (contractRequired(frame, address, featureFlags) && !addressChecks.isPresent(address, frame)) {
```
david-bakin-sl, line 61: by this point I'm beginning to wish that `contractRequired(...) && !addressChecks.isPresent(...)` had been given a method with an informative name ....

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
```diff
@@ -155,9 +156,11 @@ private MessageFrame finishedAsCall(
             @NonNull final FeatureFlags featureFlags,
             @NonNull final Configuration config) {
         Code code = CodeV0.EMPTY_CODE;
-        if (canLoadCodeFromAccount(transaction, worldUpdater)) {
+        final var contractId = transaction.contractIdOrThrow();
+
+        if (canLoadCodeFromAccount(transaction, worldUpdater, contractId, config, featureFlags)) {
             final var account = worldUpdater.getHederaAccount(to);
-            if (account == null && worldUpdater.contractMustBePresent()) {
+            if (account == null && contractMustBePresent(config, featureFlags, contractId)) {
```
david-bakin-sl, line 163: i'm losing track - how is `contractMustBePresent` different from `contractRequired`?  I mean, I see they take different parameters, but given that they compute the same predicate - perhaps they should actually have the same name?  Or would _that_ be more confusing?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameUtils.java
```diff
@@ -257,6 +260,32 @@ public static boolean unqualifiedDelegateDetected(final MessageFrame frame) {
...
+
+        Long maybeGrandfatheredNumber = null;
+        if (isLongZero(address)) {
+            try {
+                maybeGrandfatheredNumber = asNumberedContractId(address).contractNum();
+            } catch (final ArithmeticException ignore) {
+                // Not a valid numbered contract id
+            }
+        }
+        return !featureFlags.isAllowCallsToNonContractAccountsEnabled(configOf(frame), maybeGrandfatheredNumber);
```
david-bakin-sl, line 286: rather unfortunate here, and in `FrameBuilder` above - that the feature flag has the opposite sense of the thing you are really interested in, expressed positively.  Consider a very simple method with a good name that just "reverses" the sense of the feature flag, to wrap usages of the feature flag?

david-bakin-sl (approved):
Boy there's a lot of annoying plumbing you had to have in this PR.

For some reason I found the conditionals in this PR (everything left over after the plumbing was done) to be really tricky to wrap my mind around.  Not sure why.

LGTM.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
lukelee-sl: Really, the reason I needed this is because I don't have a `messageFrame` to work with because I'm in the middle of creating one for the transaction

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameUtils.java
lukelee-sl: I didn't want to put the `!` in the if statements which I would need to if I reverse the meaning of this method

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomExtCodeSizeOperation.java
lukelee-sl: true I could name it something like `possiblyInvalidSolidityAddress`.  I'll make this change if there are other comments that need fixing.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameUtils.java
```diff
@@ -257,6 +260,32 @@ public static boolean unqualifiedDelegateDetected(final MessageFrame frame) {
...
     }
 
+    /**
+     * Returns true if the given frame is a call to a contract that must be present based on feature flag settings.
+     * @param frame
+     * @param address to check for possible grandfathering
+     * @param featureFlags
+     * @return
+     */
+    public static boolean contractRequired(
```
Nana-EC, line 270: Q: `contractMustBePresent` and `contractRequired` seem similar in purpose and naming but are in different classes.
Are they essentially overloads with one taking a `ContractId` and the other taking an `address` and should they be collocated?

lukelee-sl: `contractMustBePresent` is private to FrameBuilder.  I needed this method to construct a `messageFrame` which is then used everywhere else because it is available when running the evm.

PR #11014, opened 2024-01-17
test: enable fuzzy match on TokenInfoHTSSuite

vtronkov:
**Description**:
Enable fuzzy matching for TokenInfoHTSSuite

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/10895


tinker-michaelj (changes requested):
Child records for HTS system contract calls are not removable; so if this change was necessary to make `TokenInfoHTSSuite` match, there is something else going on here. 🤔 

Can you add the `Expected "X" but was "Y"` from the failing test that caused you to make this change?

vtronkov:
> Child records for HTS system contract calls are not removable; so if this change was necessary to make `TokenInfoHTSSuite` match, there is something else going on here. 🤔
> 
> Can you add the `Expected "X" but was "Y"` from the failing test that caused you to make this change?

The only test that need to be fixed is `getInfoOnInvalidNonFungibleTokenFails`(the others are passing without any changes). The error I get is: "_Instead of 18 items, 19 were generated_". I was able to pinpoint the problem to the second contractCall from the test:
```
contractCall(
        TOKEN_INFO_CONTRACT,
        GET_INFORMATION_FOR_NON_FUNGIBLE_TOKEN,
        HapiParserUtil.asHeadlongAddress(
                asAddress(spec.registry().getTokenID(NON_FUNGIBLE_TOKEN_NAME))),
        2L)
.via(NON_FUNGIBLE_TOKEN_INFO_TXN + 2)
.gas(1_000_000L)
.hasKnownStatus(ResponseCodeEnum.CONTRACT_REVERT_EXECUTED)
```

The flow is as follows - ContractCallHandler -> _(call the contract created in the test)_ -> _(call HTSSystemContract)_
The HTSSystemContract call is successful and we create the child record(), then we get back to FrameRunner.runToCompletion and we call runToCompletion(the private method). Then following the code into the besu it triggers a REVERT command. When the REVERT command is triggered we end up in RecordListBuilder.revertChildrenOf. There we can see the following code:
```
for (int i = index; i < count; i++) {
            final var child = childRecordBuilders.get(i);
            if (child.reversingBehavior() == ReversingBehavior.REMOVABLE) {
                // Remove
                ....
            } else {
                if (child.reversingBehavior() == ReversingBehavior.REVERSIBLE) {
                    child.nullOutSideEffectFields();
                    if (SUCCESSES.contains(child.status())) {
                        child.status(ResponseCodeEnum.REVERTED_SUCCESS);
                    }
                }
```
And since the child is REVERSIBLE we end up keeping it in the childRecordBuilder and change its status to REVERTED_SUCCESS. And this is the child record that is different from mono.

I tried a couple of things:
1. Find a missing validation in HTSSystemContract(and not create child record at all) -couldn't find one
2. Find a missing validation on ContractCallHandler level - couldn't find one.
3. Find a bug/missing validation on RecordListBuilder.revertChildrenOf method - couldn't find one.

So the only thing I could do is to mark the child with "REMOVABLE" ReversingBehavior. I'm more than open to suggestions on how to proceed. Thank you!

vtronkov:
Just to have it as a record, here is a response from Michael T:

> I am 95% sure this is caused by a bug in mono-service where NonFungibleTokenInfoPrecompile fails to override the Precompile.getFailureResult() method [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/Precompile.java#L114)---so instead of encoding return bytes of the same "shape" as the success result (see [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/impl/NonFungibleTokenInfoPrecompile.java#L77)), it returns just a (int64) tuple
> This means that when the EVM runs the bytecode generated to decode the output from the call ingetInfoOnInvalidNonFungibleTokenFails(), it already fails----there should be a NonFungibleTokenInfo to decode (recall signature [here](https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L692))----but there isn't!
> :arrow_right: So we can add FULLY_NONDETERMINISTIC to that spec, since we need to fix a serious bug in mono-service before we can even get a correct snapshot (edited)

tinker-michaelj (approved):
LGTM, ty @vtronkov !

PR #11031, opened 2024-01-18
chore: remove check which performs receiver signature check 

lukelee-sl:
**Description**:
Remove a check which performs receiver signature checks during the precheck phase of `ContractCall` transactions.


tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

PR #11032, opened 2024-01-18
fix: Enable fuzzy matching for HRCPrecompileSuite

stoyanov-st:
**Description**:
To match the `mono` behaviour I added a condition in `FinalizeChildRecordHandler` to set `null` to `transferList` when the record failed.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/10874

**Notes for reviewer**:

**Checklist**


tinker-michaelj (changes requested):
Quite a resourceful match mode implementation @stoyanov-st!

I think empty transfer lists are likely to appear in production record streams, however; and we should try to make mod-service match mono-service in this case.

Note that transfer lists are constructed and set just in sub-classes of [`RecordFinalizerBase`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java); so the fix should be pretty targeted.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeChildRecordHandler.java
```diff
@@ -66,6 +67,8 @@ public void finalizeChildRecord(@NonNull final ChildFinalizeContext context) {
             recordBuilder.transferList(TransferList.newBuilder()
                     .accountAmounts(asAccountAmounts(hbarChanges))
                     .build());
+        } else if (recordBuilder.status() != ResponseCodeEnum.SUCCESS) {
+            recordBuilder.transferList(null);
```
mhess-swl, line 71: `finalizeChildRecord` feels like a weird place to make sure the _user transaction_ transfer list has nothing in it. I'm not sure a better place for it though 🤔

stoyanov-st: In `mono` when there are multiple failing child records we observe nulled out fields like `transferList` while in `mod` we provide the default values. This difference (mono null vs mod empty objet)  caused the fuzzy-match to fail so I added this change to match the mono behaviour.

mhess-swl: Nevermind, I was thinking that there was also a `childRecordBuilder` available

mhess-swl (commented):
Looks good from services side

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeChildRecordHandler.java
```diff
@@ -72,6 +72,9 @@ public void finalizeChildRecord(@NonNull final ChildFinalizeContext context, fin
             // set an empty transfer list even if there are no hbar changes but the parent record succeeded
             // to be compatible with mono-service
             recordBuilder.transferList(TransferList.DEFAULT);
+        } else if (recordBuilder.status() != ResponseCodeEnum.SUCCESS) {
```
tinker-michaelj, line 75: Is this still required after merging `develop`? C.f. [this change](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java#L629) which initializes child `transferList` to `null`.

stoyanov-st: Even with the recent changes the `transferList` is still present in the result.

tinker-michaelj (approved):
Not sure why this is necessary, but seems reasonable, LGTM @stoyanov-st

PR #11043, opened 2024-01-18
fix: state unchanged on restart without upgrade

### LINKED ISSUE #11041: AddressBookInitializer should not indicate a change when there is no software upgrade.

edward-swirldslabs:
**Description**:

A previous PR made a change to the AddressBookInitializer that caused the address books in the state to be upgrade when there was not a software upgrade taking place.  This caused the nodes to ISS when restarted without a software upgrade.   

This PR fixes that problem.  

**Related issue(s)**:

Fixes #11041

**Notes for reviewer**:

I adjusted unit tests to match the correct behavior. 
I manually verified that the state is no longer modified on restart using the state that lead to the ISS.

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/AddressBookInitializerTest.java
+++ b/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/AddressBookInitializerTest.java
@@ -143,7 +143,9 @@ class AddressBookInitializerTest {
         assertNull(initializer.getPreviousAddressBook(), "The previous address book should be null.");
         assertAddressBookFileContent(
                 initializer, configAddressBook, signedState.getAddressBook(), inititializedAddressBook);
-        assertFalse(initializer.hasAddressBookChanged());
+        // Even when the genesis state has the correct address book, we always adopt the config.txt address book and
+        // indicate an address book change.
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/AddressBookInitializerTest.java
```diff
@@ -152,7 +154,7 @@ void noStateLoadedFromDiskGenesisStateChangedAddressBook() throws IOException {
         clearTestDirectory();
         final AddressBook configAddressBook = getRandomAddressBook();
         // initial state has currentAddressBook set to configAddressBook
```
alittley, line 156: is this comment still relevant?

alittley (dismissed):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/AddressBookInitializerTest.java
edward-swirldslabs: I swapped out which unit test is setting the addressbook to null in the genesis state.  The name of the unit test sounds like it implies having an address book in the genesis state.

alittley (approved):

Issue #9728, opened 2023-11-07
HIP-844 Work Package 1/1: Address account nonce discrepancies


stoqnkpL:
### Abstract / Motivation

Introduce clear rules for when to update an Ethereum transaction signer's nonce and a new field for externalising the nonce update in the record stream. This will help resolve issues where the Consensus nodes and the Mirror nodes are out of sync in regards to what the current value of an account's nonce is.

### Specification

1. The nonce of a signer of an Ethereum transaction should be increased for all transactions that are executed _in the EVM_.
2. Transactions that have reached consensus and are handled by the node but have failed additional validations performed by the node _prior to being executed in the EVM_ should have **no effect** on the Ethereum transaction signer nonce.
3. A new protobuf field in the record stream - `signerNonce` - should be populated by the node with the current value of the signer account nonce _after_ executing the transaction. This value should be used by the Mirror nodes when importing record stream files.

### Protobuf update

Add a new protobuf field to contract_call_local.proto:
```
message ContractFunctionResult {
    
    [...]

    /**
     * If not null this field specifies what the value of the signer account nonce is post transaction execution. 
     * For transactions that don't update the signer nonce, this field should be null.
     */
    Int64Value signer_nonce = 15;
}
```

### Services update

1. Move the signer nonce update to be right before this code block that starts the EVM:
```
while (!messageFrameStack.isEmpty()) {
   process(messageFrameStack.peekFirst(), tracer);
}
```
2. Populate the new protobuf field when externalising the transaction.
3. Fix in mono and mod codebase

### Mirror node update

1. When importing transactions from the record stream, update the value of the signer account's nonce that's persisted on the mirror node side with the value externalised in the new protobuf field.
2. When importing legacy transactions that don't have this field, the Mirror node should consider whether the transaction result is `SUCCESS` or `CONTRACT_REVERT_EXECUTED` and increment the signer nonce in these cases. Some statuses such as `INSUFFICIENT_GAS` can be returned both before and after entering the EVM, there is no way for the Mirror node to know whether the signer nonce should be updated. This was the root cause of multiple issues and the reason for the new protobuf field.

### Testing

Different scenarios should be covered with transactions failing due to 
- precheck validations
- handle validations
- EVM reverts from smart contract logic or insufficient gas, account amounts
- EVM reverts due to Hedera-specific validations like system account validations
- EVM reverts due to Hedera-specific validations like resource validations (e.g. preceding child record slots availability)
And different successful scenarios as well:
[...]

PR #11045, opened 2024-01-18
feat: address account nonce discrepancies mono

natanasow:
**Description**:

Introduce clear rules for when to update an Ethereum transaction signer's nonce and a new field for externalising the nonce update in the record stream. This will help resolve issues where the Consensus nodes and the Mirror nodes are out of sync in regards to what the current value of an account's nonce is.

**Specification**:

1. The nonce of a signer of an Ethereum transaction should be increased for all transactions that are executed _in the EVM_.
2. Transactions that have reached consensus and are handled by the node but have failed additional validations performed by the node _prior to being executed in the EVM_ should have **no effect** on the Ethereum transaction signer nonce.
3. A new protobuf field in the record stream - `signerNonce` - should be populated by the node with the current value of the signer account nonce _after_ executing the transaction. This value should be used by the Mirror nodes when importing record stream files.

**Protobuf update**:

Add a new protobuf field to contract_call_local.proto:
```
message ContractFunctionResult {
    
    [...]

    /**
     * If not null this field specifies what the value of the signer account nonce is post transaction execution. 
     * For transactions that don't update the signer nonce, this field should be null.
     */
    Int64Value signer_nonce = 15;
}
```

**Services update**:

1. Move the signer nonce update to be right before this code block that starts the EVM:
2. Populate the new protobuf field when externalizing the transaction.

**Fixes**: #9728

**Notes for reviewer**:

**Checklist**


stoqnkpL (dismissed):
LGTM

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
```diff
@@ -210,9 +210,6 @@ public void doStateTransitionOperation(
                         codeWithConstructorArgs,
                         consensusTime);
             } else {
-                sender.incrementEthereumNonce();
-                accountStore.commitAccount(sender);
-
                 result = evmTxProcessor.executeEth(
```
tinker-michaelj, line 213: Why don't we set the result's `signerNonce` following this `executeEth()` call (in analogy to `ContractCallTransitionLogic`)?

tinker-michaelj:
> The nonce of a signer of an Ethereum transaction should be increased for all transactions that are executed in the EVM.

What counts as "executed in the EVM"? As long as we could charge gas, it counts?

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
natanasow: Good catch, thank you! Fixed it and also added a test for ethereum contract deployment.

I haven't added in `EvmTxProcessor.executeEth()`, because there are many more places that have to be handled. E.g. both `ContractCreateTransitionLogic` and `ContractCallTransitionLogic` are leading to `CreateEvmTxProcessor.executeEth()` and `CallEvmProcessor.executeEth()`. And again, they both lead to `EvmTxProcessor.execute()` where I can potentially add the `signer_nonce` to the `TransactionProcessingResult`, but there are several different `return` places that must be handled. So, I took the path that was to add them, once the result had been already built. If you think that the better approach is to move `result.setSignerNonce()` to `EvmTxProcessor`, I'll refactor the current implementation :pray:.

natanasow:
> > The nonce of a signer of an Ethereum transaction should be increased for all transactions that are executed in the EVM.
> 
> What counts as "executed in the EVM"? As long as we could charge gas, it counts?

IMO "executed in the EVM" means that the transaction has reached [that point](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java#L160) and there are no other "blockers". The last validations between prechecks and "evm execution" is [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/EvmTxProcessor.java#L310). So, once the transaction reaches that point, we are safe to increase the signer's nonce because there are no other potential validations that could lead to increasing the nonce and not executing the transaction like it was before. 
The previous implementation increases the nonce [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L161) which causes a nonce discrepancy between services and mirror node because if the transaction fails in any of [these checks](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/EvmTxProcessor.java#L270), the nonce in services will be incremented but no record file will be exported and the new signer's nonce will not be populated in the mirror node.
These are my thoughts and feel free to correct me if I misunderstood something of that.

stoqnkpL (approved):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java
tinker-michaelj: Ah np, this looks fine to me. 🙏

Issue #11054, opened 2024-01-18
API for abstract tasks


artemananiev:
Current wiring framework in swirlds-common uses tasks internally, but they aren't exposed from the module. Other modules, e.g. swirlds-virtualmap, may also need tasks, it would be great to have single API shared across all modules.

PR #11056, opened 2024-01-18
fix: 11054: API for abstract tasks

artemananiev:
Fix summary:

* `AbstractTask` and its subclasses are moved to `com.swirlds.common.wiring.tasks` package
* The classes are now public, a few methods are also changed to public
* `AbstractTask.send()` is kept protected. Calls to this method from `ConcurrentTaskScheduler` are eliminated, now the tasks are started automatically from constructor rather than explicitly by the scheduler
* `AbstractTask.send()` is improved. It now has a shortcut to use `fork()` rather than to submit the task to the pool, if the current thread is already a member of the tasks's pool
* `AbstractTask.send(int)` is added. It's similar to `send()`, but decrements the number of dependencies by n rather than by 1

Fixes: https://github.com/hashgraph/hedera-services/issues/11054

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/AbstractTask.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/AbstractTask.java
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-package com.swirlds.common.wiring.schedulers.internal;
+package com.swirlds.common.wiring.tasks;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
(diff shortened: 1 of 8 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/AbstractTask.java
```diff
@@ -66,8 +67,32 @@ protected final void setRawResult(Void value) {}
...
+            runTask();
+        }
+    }
+
+    /*
+     * Schedules this task for execution. If the current thread is a member of this task's {@link pool},
+     * the task is forked, otherwise submitted to the pool using {@link ForkJoinPool#execute} method.
+     */
+    private final void runTask() {
+        if ((Thread.currentThread() instanceof ForkJoinWorkerThread t) && (t.getPool() == pool)) {
```
cody-littley, line 93: @OlegMazurov is this `instanceof` operation going to be expensive?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/AbstractTask.java
```diff
@@ -66,8 +67,32 @@ protected final void setRawResult(Void value) {}
...
+        }
+    }
+
+    /**
+     * If the task has no dependencies then execute it. If the task has dependencies, decrement the dependency
+     * count by {@code count} and execute if the resulting number of dependencies is zero.
+     *
+     * @param count dependency count to decrement
+     */
+    void send(final int count) {
```
OlegMazurov, line 82: I'm against this extension. It's not actually needed (I left another comment in the relevant PR).

OlegMazurov (changes requested):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/AbstractTask.java
artemananiev: OK, let's not add it for now.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/tasks/ConcurrentTask.java
```diff
@@ -53,6 +54,7 @@ protected ConcurrentTask(
         this.data = data;
         this.offRamp = offRamp;
         this.uncaughtExceptionHandler = uncaughtExceptionHandler;
+        send();
```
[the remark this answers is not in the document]
artemananiev: OK, I'm reverting this part of the fix. Let these tasks be started explicitly by the caller (e.g. ConcurrentTaskScheduler)

anthony-swirldslabs (approved):

cody-littley (approved):

cody-littley (approved):

anthony-swirldslabs (approved):

OlegMazurov (approved):

jeromy-cannon (approved):

PR #11060, opened 2024-01-18
feat: add support for industry standard key storage

### LINKED ISSUE #10211: Support Multiple .pfx Files for Node Keys
### Problem

I was trying to write a piece of code in Javascript/Golang (for one of our CLI tool) that will generate gossip keys  (`private-<nodeID>.pfx` and `public.pfx` file) for our platform node similar to this script [hedera-node/data/keys/generate.sh](https://github.com/hashgraph/hedera-services/blob/a2e0e80f1714ca8be709945e9d1dc49e64c161e8/hedera-node/data/keys/generate.sh#L30).

That script generates and pack the following key-pairs for a node in a single `private-<nodeID>.pfx` file:
- `s-<nodeID>` key-pair
- `a-<nodeID>` key-pair
- `e-<nodeID>` key-pair
- `k-<nodeID>` key-pair
and then a `public.pfx` with all the certificates that is shared across the nodes

**Problem** We don't want to rely on `keytool` as it requires user to have JDK installed on their cloud vm or machine. On the other hand, `openssl` only allows a single private key to be in the `*.pfx` file and a chain of certificates. 

As it says here `PKCS #12 files are usually created using OpenSSL, which only supports a single private key from the command line interface. The Java keytool can be used to create multiple "entries" since Java 8, but that may be incompatible with many other systems. As of Java 9, PKCS #12 is the default keystore format` (Ref: https://en.wikipedia.org/wiki/PKCS_12)

So it appears users don't have an easy way to programmatically generate compatible `*.pfx` files for Hedera Platform Nodes where they can include multiple key-pairs using open-source "crypto libraries" or `openssl` CLI.

### Solution

If we split the single `private-<node ID>.pfx` files into separate `pfx` files with single key-pair as below:
- `private-<node ID>-s.pfx` for `s-<nodeID>` key-pair
- `private-<node ID>-a.pfx` for `a-<nodeID>` key-pair
- `private-<node ID>-e.pfx` for `e-<nodeID>` key-pair
- `private-<node ID>-k.pfx` for `k-<nodeID>` key-pair

then it would allow users to generate compatible `pfx` files either using `openssl` CLI or any other crypto libraries in other languages.

However, it would require that our platform supports accepting such multiple `pfx` files for node-keys and certificates.

Here is our Slack thread where such proposal was made: https://swirldslabs.slack.com/archives/C03DU0QGE2G/p1701386136167299

Also, we need to support separate `public.pfx` keys for each node instead of a single file.

### Alternatives

User could use a docker container to run the [hedera-node/data/keys/generate.sh](https://github.com/hashgraph/hedera-services/blob/a2e0e80f1714ca8be709945e9d1dc49e64c161e8/hedera-node/data/keys/generate.sh#L30).

This is what I had to implement as part of our `node-mgmt-tools` CLI to generate keys.

### LINKED ISSUE #10666: mTLS Keys: Add support for an alternate disk storage format which is OpenSSL compatible

nathanklick:
## Description

- Adds support for industry standard key and certificate storage in an improved on-disk format. 
- Adds a new configuration option to enable and disable the enhanced key loading mechanism (`crypto.enableNewKeyStoreModel`) which defaults to a disabled state by default.
- The `EnhancedKeyStoreLoader` provides backwards compatibility and is able to load the legacy storage format.
   - It is able to detect the storage format currently in use and load using the appropriate mechanism based solely on the key files on disk.
   - It supports a hybrid loading model where some keys are in the legacy format and some are in the new industry standard format.
- Makes some existing legacy methods `package-private` to support DRY code and avoid duplication. 

### Related Issues

- Closes #10666 
- Closes #10211 

## Known Caveats

- ~~No unit tests are currently provided, but are in progress.~~
- ~~There are several javadoc comments missing, but will be provided.~~


edward-swirldslabs (approved):

nathanklick:
> The legacy code paths are executed frequently in testing and production, but I am wondering if it would be worth adding unit test for the new code paths that are sitting unused behind the feature flag.

Yes, I am working on unit tests for both code pathways. Additionally, the goal is to eventually remove the legacy code path and use the `EnhancedKeyStoreLoader` solely since it supports both formats.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -0,0 +1,1094 @@
...
+            final InputDecryptorProvider inputDecrypter =
+                    new JceInputDecryptorProviderBuilder().build(new String(keyStorePassphrase).getBytes());
+
+            return switch (entry) {
+                case PrivateKeyInfo pki -> converter.getPrivateKey(pki);
+                case PKCS8EncryptedPrivateKeyInfo epki -> converter.getPrivateKey(
+                        epki.decryptPrivateKeyInfo(inputDecrypter));
+                case PEMKeyPair kp -> converter.getPrivateKey(kp.getPrivateKeyInfo());
+                case PEMEncryptedKeyPair ekp -> converter.getPrivateKey(
+                        ekp.decryptKeyPair(decrypter).getPrivateKeyInfo());
```
leninmehedy, line 989: It looks good that this function can extract different formats. 

However, since `keyStorePassphrase` cannot be `null`, essentially the loader only supports `PEMEncryptedKeyPair`. Am I right?

Just wondering if we should avoid supporting other formats.

leninmehedy (approved):
Thank you! Such an important feature!!!

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
nathanklick: @leninmehedy The loader supports all formats listed. Restricting the array parameter to non-null simply means the are not allowed to pass in a `null` value but they may still pass in a zero length `char` array.

edward-swirldslabs (approved):

Neeharika-Sompalli (approved):
LGTM from hedera-base changes

timo0 (approved):

mhess-swl (approved):

anthony-swirldslabs (approved):
LGTM

PR #11061, opened 2024-01-19
chore: use fixed missing contract numbers in `Evm46ValidationSuite`

tinker-michaelj:
**Description**:
 - Fix intermittent failures like [this one](https://scans.gradle.com/s/fwddvtgvwx6cw/tests/task/:test-clients:itest/details/AllIntegrationTests/concurrentSpecs()%5B2%5D?top-execution=1) by using fixed missing contract numbers so that intrinsic gas cost doesn't vary with the number of binary zeros in the EVM payload.


kimbor (approved):
lgtm

PR #11064, opened 2024-01-19
fix: check throttle usage on committing changes 

### LINKED ISSUE #10215: Implement throttling on committing changes
In mono's `HederaWorldState.commit()` method we have some throttling logic, more specifically [this](https://github.com/hashgraph/hedera-services/blob/8edfb51c44b55c59369fc3031307e11f62f71a3e/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/HederaWorldState.java#L334) and [this](https://github.com/hashgraph/hedera-services/blob/8edfb51c44b55c59369fc3031307e11f62f71a3e/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/HederaWorldState.java#L340) calls.

We need to implement this logic in modularized codebase as well. A suggestion for a possible place to add it would be `RootProxyWorldUpdater.commit()`.

petreze:
**Description**:
This PR adds the checks if the auto-account creation capacity is reached as well as child records capacity inside `RootProxyWorldUpdater.commit()` method

**Related issue(s)**:

Fixes #10215 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/Context.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.spi.workflows;
+
+import com.hedera.hapi.node.base.HederaFunctionality;
+
+public interface Context {
```
tinker-michaelj, line 21: Nice idea! 🥂 

What about naming `OperationContext` to add more domain meaning? (I generally use HAPI "operations" as am umbrella term to include both transactions and queries.)

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/Context.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Verifies if the frontend throttle has enough capacity to handle the given number of the
```
tinker-michaelj, line 23:
```
Verifies if the throttle for this operation has enough capacity...
```

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/QueryContext.java
```diff
@@ -29,7 +29,7 @@
 /**
  * Context of a single query. Contains all query specific information.
  */
-public interface QueryContext {
+public interface QueryContext extends OperationContext {
```
MiroslavGatsanoga, line 32: **question:** ‏Why do we need to consider `QueryContext`? Isn't `HandleContext` enough?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/impl/NetworkUtilizationManagerImpl.java
```diff
@@ -201,4 +201,9 @@ public boolean wasLastTxnGasThrottled() {
     public void leakUnusedGasPreviouslyReserved(@NonNull final TransactionInfo txnInfo, long value) {
         backendThrottle.leakUnusedGasPreviouslyReserved(txnInfo, value);
     }
+
+    public boolean shouldThrottle(@NonNull TransactionInfo txnInfo, HederaState state) {
+        final var now = Instant.now();
```
MiroslavGatsanoga, line 206: **issue:** ‏Using `Instant.now()` doesn't seem right, in order to be deterministic we should be using consensus time in the backend throttle (and during handle in general). In mono we are passing it [here](https://github.com/hashgraph/hedera-services/blob/5a68acf8bdda2a8656104490d774bfa4afd20223/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/throttling/TxnAwareHandleThrottling.java#L45).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -896,6 +907,40 @@ public void reclaimPreviouslyReservedThrottle(int n, HederaFunctionality functio
...
+
+    @Override
+    public boolean hasThrottleCapacityForChildTransactions() {
+        var isAllowed = true;
+        final var childRecords = recordListBuilder.childRecordBuilders();
+
+        for (int i = 0, n = childRecords.size(); i < n && isAllowed; i++) {
+            final var childRecord = childRecords.get(i);
+            if (Objects.equals(childRecord.status(), SUCCESS)) {
+                final var tx = childRecord.build().transaction();
```
MiroslavGatsanoga, line 927: **thought:** ‏Not sure if we don't need a new getter for the `transaction` instead of calling `build()` 🤔.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -896,6 +907,40 @@ public void reclaimPreviouslyReservedThrottle(int n, HederaFunctionality functio
...
+                final var txInfo = new TransactionInfo(
+                        tx, childRecord.transactionBody(), tx.sigMap(), tx.signedTransactionBytes(), functionality);
```
MiroslavGatsanoga, line 929: **issue:** ‏I think `functionality` here is the one of the of the context, but the one we need should be derived from the transaction body of the child record probably using `functionOf(childRecord.transactionBody())`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleContextImpl.java
```diff
@@ -896,6 +907,40 @@ public void reclaimPreviouslyReservedThrottle(int n, HederaFunctionality functio
...
+                if (txInfo.functionality() == CONTRACT_CREATE || txInfo.functionality() == CONTRACT_CALL) {
+                    continue;
+                }
+
+                if (shouldThrottleTxn(txInfo)) {
+                    isAllowed = false;
+                }
+            }
+        }
+
```
MiroslavGatsanoga, line 940: **question:** ‏Here we are missing the reseting of the throttles from snapshots in the case of which throttling occurs i.e. [this from mono](https://github.com/hashgraph/hedera-services/blob/5a68acf8bdda2a8656104490d774bfa4afd20223/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/records/TxnAwareRecordsHistorian.java#L197). Is that on purpose?

petreze: Definitely not on purpose, it is updated

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/QueryContext.java
MiroslavGatsanoga: I thought `ContextQueryProcessor` calls `ProxyWorldUpdater.commit()` and `ContextTransactionProcessor` calls `RootProxyWorldUpdater.commit()` but I might be wrong.

Having the more general `OperationContext` interface looks good.

MiroslavGatsanoga (commented):
LGMT regarding comments I left above, PR checks need to be fixed though.

MiroslavGatsanoga (dismissed):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/RootProxyWorldUpdater.java
```diff
@@ -96,6 +100,17 @@ public void commit() {
         final var contractChangeSummary = enhancement.operations().summarizeContractChanges();
         createdContractIds = contractChangeSummary.newContractIds();
 
+        context().ifPresent(context -> {
```
tinker-michaelj, line 103: There's a feature flag we should check here, `contracts.enforceCreationThrottle`, has a default value of `false`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java
```diff
@@ -441,6 +448,22 @@ public void commit() {
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void setupContext(@NonNull final OperationContext context) {
```
tinker-michaelj, line 455: Is there a reason not to just inject the `HandleContext` into the `RootProxyWorldUpdater`?

tinker-michaelj (approved):
LGTM, ty @petreze !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyWorldUpdater.java
petreze: After the realization that this check is only needed for txs, there isn't. I updated the implementation

PR #11073, opened 2024-01-22
test: Enable fuzzy matching for ContractKeysStillWorkAsExpectedSuite

vtronkov:
**Description**:
- Enable fuzzy matching for ContractKeysStillWorkAsExpectedSuite

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/10869

**Notes for reviewer**:
The first fuzzy matching contractCall in `approvalFallbacksRequiredWithoutTopLevelSigAccess` is failing due to gasUsed differences. Debugging the code the difference is between the `unitAdjusts` size(for mod in `ClassicTransferCall.minimumTinybarPriceGiven`) and `transfer.fungibleTransfers().size()`(for mono in `TransferPrecompile.getMinimumFeeInTinybars`). In mod is 2 in mono is 1. Thus the multiplier for the gasUsed calculation is different.

Looking at where the transfer list is added we can see that in mod(`ClassicTransferDecoder.sendingUnitsFromTo`) we are adding the credit and debt transfers. In mono(`TransferPrecompile.decodeTransferToken`) we are only adding the credit. I assume that somewhere down the line, in mono, we are handling the debt, but up to this point where the gasUsed calculation happens we only count the credit.

I tried removing the debit from `ClassicTransferDecoder.sendingUnitsFromTo` but it seems that there is a code depending on that so that doesn't work. That's why I decided to change the gasUsed calculator instead of using the unitAdjusts size to count the **credit** transfers. And this seemed to work.

However when we have multiple token transfers:
```
sourcing(() -> contractCall(
                WELL_KNOWN_TREASURY_CONTRACT,
                "transferTokenUnitsFromToOthers",
                fungibleTokenMirrorAddr.get(),
                aSenderAddr.get(),
                aReceiverAddr.get(),
                bSenderAddr.get(),
                bReceiverAddr.get())
    .gas(2_000_123)
        .alsoSigningWithFullPrefix(A_WELL_KNOWN_SENDER, B_WELL_KNOWN_SENDER)
    .hasKnownStatus(expectedStatus))
```
    
we have another inconsistency. In `TransferPrecompile.decodeTransferToken` we always have one record for the transfer list(the credit one) but in the `TransferPrecompile.decodeTransferTokens` we have logic for **credit** and **debit**. So now the multiplier instead of 4 is 2.

I'm open to ideas of how I can proceed.
- Should I change mono to have credit and debit when we are transferring a single token(`TransferPrecompile.decodeTransferToken`)?
- Should I change mod to remove the debit from ClassicTransferDecoder.sendingUnitsFromTo and change the following code that is dependent on debit?
- Something else?
cc: @tinker-michaelj


vtronkov:
Enabled in https://github.com/hashgraph/hedera-services/pull/11073

PR #11074, opened 2024-01-22
fix: Nonce Discrepancies in Modularization

Ivo-Yankov:
**Description**:
- Populates the `signerNonce` field when externalizing transactions. It has a `long` value for `EthereumTransaction` and `null` in all other cases.
- Enables all Hapi tests in the new `NonceSuite`
- Enables fuzzy matching for those tests

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -205,6 +205,12 @@ public BigInteger getMaxGasAsBigInteger() {
      * @return the effective offered gas price
      */
     public long effectiveOfferedGasPriceInTinybars() {
+
+        // prevent division by zero
+        if (gasLimit == 0) {
```
tinker-michaelj, line 210: Is it possible to call this with `gasLimit = 0`? If so, what test surfaced that?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java
```diff
@@ -129,6 +129,11 @@ public GasCharges chargeForGas(
         if (transaction.isEthereumTransaction()) {
             requireNonNull(relayer);
             final var allowanceUsed = chargeWithRelayer(sender, relayer, context, worldUpdater, transaction);
+
+            // Increment nonce right before the gas is charged
```
tinker-michaelj, line 133: This says to increment "right before"; but it's clearly being incremented right _after_ charging. 😁 

Is the comment wrong, or the order of operations wrong?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -116,7 +117,11 @@ public CallOutcome call() {
         } catch (AbortException e) {
             // Commit any HAPI fees that were charged before aborting
             rootProxyWorldUpdater.commit();
-            final var result = HederaEvmTransactionResult.fromAborted(e.senderId(), hevmTransaction, e.getStatus());
+            final var result = HederaEvmTransactionResult.fromAborted(
+                    e.senderId(),
+                    hevmTransaction,
+                    e.getStatus(),
+                    context.recordBuilder(ContractOperationRecordBuilder.class).getSignerNonce());
```
tinker-michaelj, line 124: I think it's safer to re-read the sender account (if present) and call `getNonce()` rather than relying on the `recordBuilder` always having  the right value. 

(What value do we want to externalize if the sender doesn't exist?)

hedera-node/test-clients/src/main/resource/spec-default.properties
```diff
@@ -110,7 +110,7 @@ num.opFinisher.threads=8
...
...LD_NOT_ALLOWED,TRANSACTION_OVERSIZE,TRANSFER_ACCOUNT_SAME_AS_DELETE_ACCOUNT,EMPTY_ALLOWANCES,REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT,TOKEN_HAS_NO_FREEZE_KEY,TOKEN_HAS_NO_SUPPLY_KEY,INVALID_TOKEN_INITIAL_SUPPLY,INVALID_TOKEN_DECIMALS,INVALID_TOKEN_MAX_SUPPLY,ACCOUNT_REPEATED_IN_ACCOUNT_AMOUNTS,TRANSFERS_NOT_ZERO_SUM_FOR_TOKEN,INVALID_ACCOUNT_AMOUNTS,TOKEN_NAME_TOO_LONG,TOKEN_SYMBOL_TOO_LONG,INVALID_TOKEN_NFT_SERIAL_NUMBER,PERMANENT_REMOVAL_REQUIRES_SYSTEM_INITIATION,MISSING_TOKEN_SYMBOL,MISSING_TOKEN_NAME,INVALID_EXPIRATION_TIME,EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS,INVALID_ALLOWANCE_OWNER_ID,FUNGIBLE_TOKEN_IN_NFT_ALLOWANCES,TOKEN_NOT_ASSOCIATED_TO_ACCOUNT,MAX_ALLOWANCES_EXCEEDED,INVALID_ALLOWANCE_SPENDER_ID,AMOUNT_EXCEEDS_TOKEN_MAX_SUPPLY,NFT_IN_FUNGIBLE_TOKEN_ALLOWANCES,NEGATIVE_ALLOWANCE_AMOUNT,DELEGATING_SPENDER_DOES_NOT_HAVE_APPROVE_FOR_ALL,DELEGATING_SPENDER_CANNOT_GRANT_APPROVE_FOR_ALL,INVALID_TOKEN_MINT_AMOUNT,INVALID_TOKEN_BURN_AMOUNT,INVALID_WIPING_AMOUNT,INVALID_NFT_ID,BATCH_SIZE_LIMIT_EXCEEDED,METADATA_TOO_LONG,INVALID_RENEWAL_PERIOD,INVALID_CUSTOM_FEE_SCHEDULE_KEY,MAX_GAS_LIMIT_EXCEEDED,CONTRACT_DELETED,INVALID_ETHEREUM_TRANSACTION,INSUFFICIENT_GAS
```
tinker-michaelj, line 113: Why do we need to add `INSUFFICIENT_GAS` here?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/NonceSuite.java
```diff
@@ -442,8 +497,9 @@ private HapiSpec nonceUpdatedAfterEvmReversionDueContractLogic() {
                                         .contractCallResult(resultWith().signerNonce(1L))));
     }
 
+    @HapiTest
     private HapiSpec nonceUpdatedAfterEvmReversionDueInsufficientGas() {
-        return defaultHapiSpec("nonceUpdatedAfterEvmReversionDueInsufficientGas")
+        return defaultHapiSpec("nonceUpdatedAfterEvmReversionDueInsufficientGas", NONDETERMINISTIC_NONCE)
```
tinker-michaelj, line 502: ⚠️ Isn't the entire point of this suite to validate `nonce`s? 🤯

Ivo-Yankov: I found that some of the tests sometimes failed with a fuzzy matching error for the nonce of `0.0.2`. The actual nonces that matter for the context of the tests are explicitly asserted. To reduce the flakiness I added the `NONDETERMINISTIC_NONCE`, although I see why it looks wierd. I can try to find another solution to the flakiness if needed.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
Ivo-Yankov: `nonceNotUpdatedWhenInsufficientIntrinsicGasPrecheckFailed` catches this edge case. WIthout the above check it fails with `FAIL_INVALID`

hedera-node/test-clients/src/main/resource/spec-default.properties
Ivo-Yankov: It allows `INSUFFICIENT_GAS` to be thrown as a precheck error as per this test: `nonceNotUpdatedWhenInsufficientIntrinsicGasPrecheckFailed`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -161,6 +161,9 @@ public class SingleTransactionRecordBuilderImpl
     private List<AbstractMap.SimpleEntry<ContractActions, Boolean>> contractActions = new LinkedList<>();
     private List<AbstractMap.SimpleEntry<ContractBytecode, Boolean>> contractBytecodes = new LinkedList<>();
 
+    @Nullable
+    private Long signerNonce = null;
```
tinker-michaelj, line 165: Where is this field used? Doesn't it need to be set on an underlying PBJ `TransactionRecord.Builder` somewhere?

Ivo-Yankov: You are right, this field isn't really needed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/NonceSuite.java
Ivo-Yankov: All usages of `NONDETERMINISTIC_NONCE` are removed

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
tinker-michaelj: Would it make more sense then to avoid the division by zero on L#214? I think it's confusing to return an offered gas **price** of zero just because the gas **limit** is zero.

I don't know what the point of dividing by `gasLimit` was in the first place. Seems like we can just do,
```
        return BigInteger.valueOf(Long.MAX_VALUE)
                .min(getMaxGasAsBigInteger().divide(WEIBARS_TO_TINYBARS))
                .longValueExact();
```
Which will result in `HederaEvmTransaction.offeredGasCost()` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransaction.java#L119) always overflowing for a non-zero `gasLimit`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
```diff
@@ -101,11 +103,23 @@ public HederaEvmTransactionResult runToCompletion(
 
         // And return the result, success or failure
         final var gasUsed = effectiveGasUsed(gasLimit, frame);
+        Long signerNonce = null;
+        if (hydratedEthTxData != null && sender != null) {
```
tinker-michaelj, line 107: All these changes are confusing imo:
 - The `sender != null` condition is dead code---i.e., it can never be true because of the `requireNonNull(sender)` on L#87. 
 - Adding `HydratedEthTxData` to the argument list is misleading because the method doesn't actually use it for anything other than deciding to set `signerNonce` in its return value.
 - **But** even that is confusing, because the client code (i.e., the code that calls this method) already knows both (1) whether `HydratedEthTxData` is non-null and (2) what the `signerNonce` is. So I'd expect the client code to either just pass in the `signerNonce` to use; or even better, for this method to know nothing about nonces at all, since the client code can already manage it.

hedera-node/test-clients/src/main/resource/spec-default.properties
tinker-michaelj: Hmm...but this list is the response codes that:
 - mono-service rejected in precheck; and,
 - mod-service does not reject in preheck.

So I don't understand.

tinker-michaelj (commented):
I would expect this PR to look up the final `senderNonce` for the record in exactly one place, probably `ContextTransactionProcessor`. I would also expect when it _is_ looked up to have a `requireNonNull(accountStore.getByAccountId(senderId))` since we should have failed much earlier on a missing sender.

So I'm just broadly confused by how many places are touched in this PR, and what invariants we are enforcing at each stage in processing the transaction.

hedera-node/test-clients/src/main/resource/spec-default.properties
MiroslavGatsanoga: @tinker-michaelj I think the behavior of `nonceNotUpdatedWhenInsufficientIntrinsicGasPrecheckFailed` is exactly that:
- for mono-service txn is rejected in precheck with `INSUFFICIENT_GAS` status; and
- for mod-service txn is not rejected in preheck (i.e. status is `OK`) and we get the `INSUFFICIENT_GAS` status during handle.

Do you mean instead of adding it to this test-clients config we should update mod-service to return in precheck with `INSUFFICIENT_GAS` status i.e. same as mono? Possibly include `EthereumTransaction` [here](https://github.com/hashgraph/hedera-services/blob/17a0bc21145fa8741c2ab5a21d35eeececfec08c/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java#L168)?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -129,7 +135,16 @@ public CallOutcome call() {
             }
             // Commit any HAPI fees that were charged before aborting
             rootProxyWorldUpdater.commit();
-            final var result = HederaEvmTransactionResult.fromAborted(e.senderId(), hevmTransaction, e.getStatus());
+
+            var result = HederaEvmTransactionResult.fromAborted(e.senderId(), hevmTransaction, e.getStatus());
+
+            if (context.body().hasEthereumTransaction()) {
+                final var sender = rootProxyWorldUpdater.getHederaAccount(e.senderId());
+                if (sender != null) {
```
Ivo-Yankov, line 143: This check is required because it is possible that the sender does not exist. In this case a `INVALID_ACCOUNT_ID` error is thrown in `processor.processTransaction(...)` and it is handled in this `catch` statement. This is shown in the test `revertsWhenSenderDoesNotExist`

tinker-michaelj (dismissed):
LGTM, tyvm @Ivo-Yankov !

PR #11081, opened 2024-01-22
feat: dumper for signed state `topics` from mono

### LINKED ISSUE #10916: Differential testing: Store dumper for Topics store needed, mono-service representation
A store dumper for Topics is needed, similar to the other dumpers under `cli-clients...com.hedera.services.cli.signedstate`

agadzhalov:
**Description**:
Dumps all `topics` entities from `mono` in deterministic order.

1. Added `topics` subcommand in [DumpStateCommand](https://github.com/hashgraph/hedera-services/blob/10916-topics-dumper/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpStateCommand.java#L337)
2. Created new file [DumpTopicsSubcommand](https://github.com/hashgraph/hedera-services/blob/10916-topics-dumper/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java) responsible for dumping the signed state from mono
    2.1. [gatherTopics](https://github.com/hashgraph/hedera-services/blob/10916-topics-dumper/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java#L137)  takes `MerkleMapLike<EntityNum, MerkleTopic>` as an input parameter and returns `Map<Long, Topic>`
4. Created getter for `number` in [MerkleTopic](https://github.com/hashgraph/hedera-services/blob/10916-topics-dumper/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleTopic.java#L307)
5. Sorted by `EntityNum.longValue()` in [DumpTopicsSubcommand](https://github.com/hashgraph/hedera-services/blob/10916-topics-dumper/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java#L150)
6. How to run
    3.1 `./services-cli.sh signed-state -f ../../../../Snapshot/SignedState.swh dump topics --topic ./topics.csv -s`

**Related issue(s)**:

Fixes #10916


hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpContractStoresSubcommand.java
```diff
@@ -260,7 +261,7 @@ WritableKVState<SlotKey, SlotValue> getMigratedContractStore() {
 
         // Start the migration with a clean, writable KV store.  Using the in-memory store here.
 
-        final var contractSchema = new InitialModServiceContractSchema(RELEASE_045_VERSION);
+        final var contractSchema = new InitialModServiceContractSchema(CURRENT_VERSION);
```
david-bakin-sl, line 264: Good fix here.

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
```diff
@@ -0,0 +1,200 @@
...
+                    topic.isDeleted(),
+                    topic.getAdminKey(),
+                    topic.getSubmitKey(),
+                    null != topic.getRunningHash() ? topic.getRunningHash() : EMPTY_BYTES,
+                    topic.getSequenceNumber(),
+                    topic.getAutoRenewDurationSeconds(),
+                    topic.getAutoRenewAccountId());
+            Objects.requireNonNull(memo, "memo");
+            Objects.requireNonNull(adminKey, "adminKey");
+            Objects.requireNonNull(submitKey, "submitKey");
```
david-bakin-sl, line 128: is the submit key really require non null?  I thought they were optional for the user to supply.  (Though that's independent of whether the _representation_ is null.)

also is memo really require non null?  other memo fields in other semantic entities are optional and can be null.  (IIRC)

(for that matter I'm not even sure admin key is require non null! is that enforced by services when you create/udpate the key? wish I knew ...)

david-bakin-sl (commented):
I only reviewed the stuff under `cli-clients/.../signedstate` - the rest of course comes from the branch that has the migrations and (I assume) isn't changed.

This looks exactly as I imagined it, or better.  Good (re)use - if I say so myself - of field formatters from the other existing dumpers.

Only question to be resolved is the "requires no null" comment.  This is still draft so you don't need an "approve" yet, so I'll mark this "comment".)

david-bakin-sl:
BTW, since this is a _mono-representation only_ dumper it can pretty much go into develop as is, not dependent on the migration branch, right?

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
agadzhalov: referring to [hedera docs(create a topic)](https://docs.hedera.com/hedera/sdks-and-apis/sdks/consensus-service/create-a-topic) looks like neither of them is required, so assume we have to remove all of the `requireNonNull`s?
image

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -602,16 +599,22 @@ public void onStateInitialized(
         } else {
             deserializedVersion = new HederaSoftwareVersion(SemanticVersion.DEFAULT, SemanticVersion.DEFAULT);
         }
-        logger.info("Deserialized version is {}, version {}", deserializedVersion, version);
+
+        // This is the *FIRST* time in the initialization sequence that we have access to the platform. Grab it!
+        // This instance should never change on us, once it has been set
+        assert this.platform == null || this.platform == platform : "Platform should never change once set";
+        this.platform = platform;
```
vtronkov, line 606: Is this needed? I can see that on lines 576-579 we have the exact block of code. Maybe there was a merge conflict? Or something from the parent branch you branched off? Are the changes from this class and `ServicesMain.java` needed?

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.stream.Collectors;
+
+/** Dump all topics from a signed state file to a text file in a deterministic order  */
+public class DumpTopicsSubcommand {
+
+    static void doit(
+            @NonNull final SignedStateHolder state,
+            @NonNull final Path topicsPath,
+            @NonNull final EmitSummary emitSummary,
+            @NonNull final Verbosity verbosity) {
```
vtronkov, line 53: Do we need the `verbosity` field? Looks like it's not currently used.

agadzhalov: I can relate, but it looks like the same approach where verbosity isn't used is also applied in [DumpTokensSubcommand](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTokensSubcommand.java#L70), [DumpTokenAssociationsSubcommand](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTokenAssociationsSubcommand.java), [DumpUniqueTokensSubcommand](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpUniqueTokensSubcommand.java#L54).

@david-bakin-sl  What do you think should we leave it for future uses?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
david-bakin-sl: this file is from the branch

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
david-bakin-sl: yes, I'd leave it - I like the option for all the dumpers to have the option to dump stuff verbosely - even if any given dumper doesn't do it.  I think the uniformity is easier to read, if you're reading more than one of them.  IMO.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleTopic.java
```diff
@@ -304,6 +304,10 @@ public void setMemo(final @Nullable String memo) {
         this.memo = ((null != memo) && !memo.isEmpty()) ? memo : null;
     }
 
+    public int getNumber() {
```
david-bakin-sl, line 307: Instead of exposing `getNumber` pull it out of the `EntityNum` via `getKey`.  (Better to not have to modify this class.)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
david-bakin-sl: yeah on second look I don't think these two files (`Hedera.java` and `ServicesMain.java`) are really part of this PR.  Right?

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
david-bakin-sl: Remember to use `getNullableFormatter` down at the definition of `FieldFormatters` too...

david-bakin-sl: Actually, now looking at `MerkleTopic` again, I see the getters for admin key and submit key will _not_ return `null` but instead will always return something - an empty key list in the case no key was supplied.  (Sorry, should have looked earlier when I originally made the comment.)

(This is different from, say, `MerkleToken`, which returns `Optional<JKey>` (which IMO is a better API, but whatever).)

So you can leave the `requireNonNull` there.

(This may change when the modular representation is added.  I haven't looked but it's possible the modular representation may return null, or may. return `Optional`.  We'll have to see ...)

david-bakin-sl (changes requested):
Need to look at `Hedera.java` and `ServicesMain.java` I think: Are they really part of this PR?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
agadzhalov: Fixed

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleTopic.java
agadzhalov: Fixed by adding [topic.getKey().intValue()](https://github.com/hashgraph/hedera-services/blob/6ff5a597ead7233a987408ecc4f13fa4d09b3b47/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java#L116). Removed getNumber from `MerkleTopic`

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpStateCommand.java
```diff
@@ -334,6 +334,27 @@ void associations(
...
 
+    @Command(name = "topics", description = "Dump topics")
+    void topics(
+            @Option(
+                            names = {"--topic"},
+                            required = true,
+                            arity = "1",
+                            description = "Output file for topics dump")
+                    @NonNull
+                    final Path associationsPath,
```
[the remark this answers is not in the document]
agadzhalov: Fixed

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpTopicsSubcommand.java
agadzhalov: Left the requireNonNulls and fixed by using `getNullableFormatter` for `adminKey` and `supplyKey`

david-bakin-sl (approved):
LGTM!

mhess-swl (approved):
LGTM

kimbor (approved):

PR #11096, opened 2024-01-22
feat: charge gas if contract call fails before calling the evm

lukelee-sl:
**Description**:

It is possible that an exception can be thrown before the evm has a chance to execute during `ContractCall` transactions.  In these cases, we need to charge the sender gas.


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -95,7 +99,10 @@ public CallOutcome call() {
         assertEthTxDataValidIfApplicable();
 
         // Try to translate the HAPI operation to a Hedera EVM transaction, throw HandleException on failure
```
david-bakin-sl, line 101: This (existing) comment might stand improvement.  Because now there's success, `isException`, and failure.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -122,6 +129,32 @@ public CallOutcome call() {
         }
     }
 
+    private HederaEvmTransaction safeCreateHevmTransaction() {
```
david-bakin-sl, line 132: I like the way these two new private methods were separated out.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java
```diff
@@ -160,6 +162,23 @@ private HederaEvmTransactionResult processTransactionWithParties(
...
 
+    private InvolvedParties safeComputeInvolvedParties(
+            @NonNull final HederaEvmTransaction transaction,
+            @NonNull final HederaWorldUpdater updater,
+            @NonNull final Configuration config,
+            @NonNull final HederaEvmContext context) {
+        try {
+            return computeInvolvedParties(transaction, updater, config);
+        } catch (HandleException e) {
+            // Aborted ContractCall transactions before processing in the evm result in charging all gas sent
```
david-bakin-sl, line 173: Comment out of date?  Only intrinsic gas is charged?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java
```diff
@@ -160,6 +162,23 @@ private HederaEvmTransactionResult processTransactionWithParties(
...
+            if (transaction.isContractCall()) {
+                gasCharging.chargeGasForAbortedTransaction(transaction.senderId(), context, updater, transaction);
+                throw new AbortException(e.getStatus(), transaction.senderId());
```
david-bakin-sl, line 176: Up there in `ContextTransactionProcessor` it was necessary to immediately commit the `rootProxyWorldUpdater` after calling `chargeGasForAbortedTransaction`.

Is a commit of the `updater` necessary here? (If not, where is it done (for my information))

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
```diff
@@ -162,6 +179,90 @@ void stillChargesHapiFeesOnAbort() {
...
+
+        given(context.body()).willReturn(transactionBody);
+        given(hevmTransactionFactory.fromHapiTransaction(transactionBody)).willReturn(HEVM_Exception);
+        given(transactionBody.transactionIDOrThrow()).willReturn(transactionID);
+        given(transactionID.accountIDOrThrow()).willReturn(SENDER_ID);
+
+        final var outcome = subject.call();
+
+        verify(rootProxyWorldUpdater).commit();
+        assertEquals(INVALID_CONTRACT_ID, outcome.status());
```
david-bakin-sl, line 207: I still have trouble understanding hapi tests.  Test is (by name) trying to confirm that fees are charged.  Exactly where does that check happen?  (Next test too.).

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/hevm/HederaEvmTransactionTest.java
```diff
@@ -52,4 +52,11 @@ void computesOfferedGasCostWithOverflow() {
         final var subject = TestHelpers.wellKnownRelayedHapiCallWithGasLimit(Long.MAX_VALUE);
         assertEquals(Long.MAX_VALUE, subject.offeredGasCost());
     }
+
+    @Test
+    void validateContractCallState() {
+        final var subject = TestHelpers.wellKnownHapiCall();
+        assertTrue(subject.isContractCall());
+        assertFalse(subject.isException());
+    }
```
david-bakin-sl, line 61: I guess it's ok to _not_ test when it's _not_ a contract call, or _is_ an exception?

david-bakin-sl (commented):
Generally liked this.  Have a couple of questions though; I guess the important one is about the commit.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -122,6 +129,32 @@ public CallOutcome call() {
...
+                // Return a HederaEvmTransaction that represents the error in order to charge fees to the sender
+                return hevmTransactionFactory.fromContractCallException(
+                        context.body().contractCallOrThrow(), e);
+            }
+            throw e;
+        }
+    }
+
+    private CallOutcome chargeFeesAndReturnOutcome(HederaEvmTransaction hevmTransaction) {
+        // if there was an exception while creating the HederaEvmTransaction and the transaction is a ContractCall
```
Nana-EC, line 146: Q: should this also be utilized by `ContractCreate` and `EthereumTransaction`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java
```diff
@@ -136,6 +137,37 @@ public GasCharges chargeForGas(
...
+            @NonNull final HederaEvmTransaction transaction) {
+        requireNonNull(sender);
+        requireNonNull(context);
+        requireNonNull(worldUpdater);
+        requireNonNull(transaction);
+
+        final var senderAccount = worldUpdater.getHederaAccount(sender);
+        requireNonNull(senderAccount);
+        final var intrinsicGas = gasCalculator.transactionIntrinsicGasCost(transaction.evmPayload(), false);
+        validateTrue(
```
Nana-EC, line 164: In this case should we throw or should we charge the balance left even if less than intrinsic. 
I guess right now that's minimal hbar but still

Nana-EC (commented):
Some questions but also consideration on if some logic should be extended to `ContractCreate` and maybe even `EthereumTransaction`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
lukelee-sl: The hole we are covering with this pr is specific to `ContractCall` as these calls uniquely do not charge other hbar denominated fees.  We could charge for the `ContractCreate` and `EthereumTransaction` but as far as I know there hasn't been a decision made to charge such fees.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java
lukelee-sl: [Here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java#L118).  This `catch` does not encompass the exception possibly thrown up above.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
lukelee-sl: These are unit tests?  Checking that the correct handling of `HederaEvmTransaction` with an exception. Next test checks to see if the correct type of `HederaEvmTransaction` gets created when an exception occurs.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCharging.java
lukelee-sl: We could do that.  There should have been checks to prevent this from occurring earlier but it is possible that the balance has changed in the mean time.

Nana-EC (approved):

Neeharika-Sompalli (approved):
LGTM ! For the files owned by hedera-base

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
david-bakin-sl: oh yeah, unit test, my bad.  but still my question is: name of test is `stillChargesHapiFeesOnHevmException` but I don't see a check in the test body that any fee was charged.  I see only a check that `commit` was called and that the status is as expected.  `commit` being called doesn't mean a fee was charged, or does it?

david-bakin-sl (approved):
LGTM!  (I still don't get how that unit test works but you can explain it to me offline.)

Issue #11083, opened 2024-01-22
Fix Restart tests when enabling `Validators`


Neeharika-Sompalli:
In https://github.com/hashgraph/hedera-services/pull/10944, BlockNumber Validator fails on Restart tests. Understand why Block number is not incremented sequentially in this scenario

PR #11097, opened 2024-01-22
fix: (mod-service)Fix block number closing issue on restart

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/11083

Like mono-service [closes](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/StatusChangeListener.java#L65) `multiStreams` when platform status changes to `FREEZE_COMPLETE`. Added similar logic in `Hedera` to close record streams.

Tested with restart test and the `BlockNumberValidator` added in https://github.com/hashgraph/hedera-services/pull/10944


thenswan (dismissed):
LGTM, thanks 🙂

tinker-michaelj (approved):

PR #11098, opened 2024-01-22
fix: 11030: replace ProtoUtils with PBJ APIs

### LINKED ISSUE #11030: Use classes and methods from PBJ instead of custom protobuf related code in MerkleDb
MerkleDb has a class named `ProtoUtils`, which is a collection of utility methods to work with protobuf. When the class was introduced, the corresponding methods didn't exist in PBJ or were not exposed publicly. Several tickets were filed against PBJ about that.

Now PBJ is more mature, most of the filed tickets are implemented / fixed. Instead of a custom utility class, MerkleDb should call PBJ directly.

Here are a few examples:

* Writing protobuf tags and wire type constants: implemented in https://github.com/hashgraph/pbj/issues/75
* Several sizeOf method not exposed: implemented in https://github.com/hashgraph/pbj/issues/74
* BufferedData over a file input stream, limit not set: fixed in https://github.com/hashgraph/pbj/issues/110
* IO exceptions in ProtoWriterTools: related https://github.com/hashgraph/pbj/issues/73 is fixed, but need to check if it may help simplifying code in OnDiskKeySerializer and OnDiskValueSerializer or not (IO exceptions may still be thrown at the codec level)
* ProtoUtils.writeDelimited(): very handy utility method, need to check what's in PBJ to replace it

anthony-swirldslabs:
**Description**:
* Removing the legacy ProtoUtils class and replacing its usages with public APIs exposed by PBJ directly.
* Upgrading the PBJ dependency to 0.7.13 to access a couple of the public APIs.

**Related issue(s)**:

Fixes #11030

**Notes for reviewer**:
Build and unit tests pass (in platform-sdk/swirlds-jasperdb/ as well as a few repos under hedera-node/ that use PBJ.)

**Checklist**


imalygin (dismissed):

imalygin (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
```diff
@@ -884,23 +885,24 @@ public MerkleDbTableConfig getTableConfig() {
         public int pbjSizeInBytes() {
             int size = 0;
             if (tableId != 0) {
-                size += ProtoUtils.sizeOfTag(FIELD_TABLEMETADATA_TABLEID, ProtoUtils.WIRE_TYPE_VARINT);
-                size += ProtoUtils.sizeOfVarInt32(tableId);
+                size += ProtoWriterTools.sizeOfTag(
+                        FIELD_TABLEMETADATA_TABLEID, ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG);
```
jsync-swirlds, line 889: Changing the wire type changes semantics.  Is this intended?

anthony-swirldslabs: There's not a change actually. The `ProtoUtils.WIRE_TYPE_VARINT` was defined as `0`:

https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/utilities/ProtoUtils.java#L30

and `ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG` is defined as 0 ordinal in the enum:

https://github.com/hashgraph/pbj/blob/main/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoConstants.java#L6-L9

So the referenced wire type is exactly the same, and the new code is identical to the old one from the semantics perspective.

jsync-swirlds: Interesting.  The names imply different encoding (and the change wasn't mentioned), so I thought it worth asking.

jsync-swirlds (dismissed):

artemananiev:
Looks good to me. Could you also fix another issue mentioned in #11030: there is at least one place (maybe more than one) in `DataFileMetadata`, when a readable input stream is opened for a file, and stream limit is explicitly set in MerkleDb code. Instead, another `ReadableStreamingData` constructor was introduced, it can be used here.

anthony-swirldslabs:
@artemananiev :
> Looks good to me. Could you also fix another issue mentioned in #11030: there is at least one place (maybe more than one) in `DataFileMetadata`, when a readable input stream is opened for a file, and stream limit is explicitly set in MerkleDb code. Instead, another `ReadableStreamingData` constructor was introduced, it can be used here.

Just did that in the latest commit. Please have a look.

nathanklick (commented):
LGTM wrt RE owned files

nathanklick (approved):

jeromy-cannon (dismissed):

jsync-swirlds (dismissed):

jsync-swirlds (approved):

jeromy-cannon (approved):

PR #11130, opened 2024-01-24
feat: unique tokens modular signed state dumper

### LINKED ISSUE #10917: Differential testing: Enhance unique tokens store dumper to handle modular representation
Unique tokens store dumper needs to handle the migration + dumping of the modular representation.

See the `DumpContractStoresSubcommand` for how this can be done.

See https://www.notion.so/swirldslabs/Signed-State-Store-Dumpers-57d4412fa4c84459a0ca699ce1446c98 for further information.

Depends on #10912

dikel:
**Description**:
Dumps all `uniqueTokens` entities from `modular-services` in deterministic way

**Related issue(s)**:

Fixes #10917 

**Notes for reviewer**:
- Change to using `.virtualMap().extractVirtualMapData()` instead of `.virtualMap().extractVirtualMapDataC()` because the latter was non-deterministic. My tests shows no performance impact. With an average of 3 runs both implementations took ~42 seconds to execute the state dump.
- Currently we pass 5 null parameters to `InitialModServiceTokenSchema`'s constructor in `getMigratedUniques()`. Didn't think of a better way for now
- The `withValidation` option only changes the summary for now
- Do we need to implement the migration if the `state.getUniqueNFTTokens()` is not virtual?

**Checklist**

### DIFF
--- a/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpStateCommand.java
+++ b/hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpStateCommand.java
@@ -304,12 +304,30 @@ public class DumpStateCommand extends AbstractCommand {
             @Option(
                             names = {"-s", "--summary"},
                             description = "Emit a summary information")
-                    final boolean emitSummary) {
+                    final boolean emitSummary,
+            @Option(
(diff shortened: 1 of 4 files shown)


petreze:
Done inside https://github.com/hashgraph/hedera-services/pull/11148

PR #11131, opened 2024-01-24
test: add `revertedAutoCreationRollsBackEvenIfTopLevelSucceeds` to `LazyCreateThroughPrecompileSuite` and enable fuzzy-matching on it

thenswan:
This PR adds `revertedAutoCreationRollsBackEvenIfTopLevelSucceeds` to the `LazyCreateThroughPrecompileSuite` and enables fuzzy-matching on it, which validates the failure scenario from #9423: a `ContractCall` that performs a valid lazy creation in a frame at stack depth 2, and then reverts that frame, generates no preceding children even when the frame at stack depth 1 deals with its child's reversion via a try-catch.

**Notes for reviewer**:
`LazyCreateThroughPrecompileSuite` and `LazyCreateThroughPrecompileV1SecurityModelSuite` have a couple of the same suites. Should the following suites be removed from `LazyCreateThroughPrecompileV1SecurityModelSuite`?
- `htsTransferFromFungibleTokenLazyCreate`
- `revertedAutoCreationRollsBackEvenIfTopLevelSucceeds`


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/LazyCreateThroughPrecompileSuite.java
```diff
@@ -655,4 +660,62 @@ final HapiSpec htsTransferFromForNFTLazyCreate() {
...
+                                                    asAddress(spec.registry().getAccountID(OWNER))),
+                                            HapiParserUtil.asHeadlongAddress(addressBytes),
+                                            2L)
+                                    .gas(GAS_TO_OFFER)
+                                    .via(TRANSFER_TXN)
+                                    .alsoSigningWithFullPrefix(OWNER)
+                                    .hasKnownStatus(SUCCESS));
+                }))
+                .then(
+                        getTxnRecord(TRANSFER_TXN).hasChildRecordCount(1),
```
tinker-michaelj, line 711: ⚠️ We should use `.hasNonStakingChildRecordCount()` here to void flakiness in CI caused by the `TRANSFER_TXN` exporting an end-of-day node stake update transaction.

Actually it doesn't make sense to have two versions any more 😝 ...we should just delete `hasChildRecordCount()` and rename `hasNonStakingChildRecordCount()` to it.

tinker-michaelj (changes requested):
Test looks great @thenswan , but `hasChildRecordCount()` is flaky in CI...see comment.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/LazyCreateThroughPrecompileSuite.java
thenswan: @tinker-michaelj I tried deleting `hasChildRecordCount()` and renaming `hasNonStakingChildRecordCount()` to it and this caused some specs to fail (please see below which). So I just used `.hasNonStakingChildRecordCount()` on the spec that I've added and propose to create another issue for `hasChildRecordCount()`.

Integration & HAPI Tests failures in `StakingSuite` for:
- `stakingMetadataUpdateIsRewardOpportunity`
- `SecondOrderRewardSituationsWork`
- `EndOfStakingPeriodRecTest`
- `RewardsOfDeletedAreRedirectedToBeneficiary`

Additional HAPI Test failure for `CongestionPricingSuite.java#canUpdateMultipliersDynamically2()`

[^ CI run](https://github.com/hashgraph/hedera-services/runs/21263076968)

tinker-michaelj: Ah that's a great point @thenswan , `StakingSuite` specifically wants to count both. 💯

tinker-michaelj (approved):
LGTM, ty @thenswan !

PR #11133, opened 2024-01-24
fix: matching the signature for NonFungibleTokenInfo and FungibleTokenInfo on failure

### LINKED ISSUE #11132: Matching the signature for NonFungibleTokenInfo and FungibleTokenInfo on failure
If there is a failure when executing the precompile, we seem to be just returning a responseCode to the user by calling Precompile.getFailureResultFor instead of matching the signature for those precompiles which also include a NonFungibleTokenInfo and FungibleTokenInfo.

mustafauzunn:
**Description**:

If there is a failure when executing the precompile, we seem to be just returning a responseCode to the user by calling Precompile.getFailureResultFor instead of matching the signature for those precompiles which also include a NonFungibleTokenInfo and FungibleTokenInfo.

**Related issue(s)**:

Fixes #11132 

**Notes for reviewer**:

In the modularised codebase, it seems that now we are not filling `contractCallResult` on failure. 
For now we comment out these assertions until we have the mod-service contractCallResult management fixed  

**Checklist**

### DIFF
--- a/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/store/contracts/precompile/codec/EvmEncodingFacade.java
+++ b/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/store/contracts/precompile/codec/EvmEncodingFacade.java
@@ -61,14 +61,10 @@ public class EvmEncodingFacade {
     }
 
     public Bytes encodeGetTokenType(final int tokenType) {
-        return functionResultBuilder()
-                .forFunction(FunctionType.HAPI_GET_TOKEN_TYPE)
-                .withStatus(SUCCESS.getNumber())
-                .withGetTokenType(tokenType)
-                .build();
+        return encodeGetTokenType(SUCCESS, tokenType);
     }
 
-    public Bytes encodeGetTokenTypeFailure(final ResponseCodeEnum status, final int tokenType) {
+    public Bytes encodeGetTokenType(final ResponseCodeEnum status, final int tokenType) {
         return functionResultBuilder()
                 .forFunction(FunctionType.HAPI_GET_TOKEN_TYPE)
                 .withStatus(status.getNumber())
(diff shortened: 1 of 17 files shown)


mhess-swl (approved):
Makes sense to me 👍

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ApproveAllowanceSuite.java
```diff
@@ -286,7 +287,20 @@ final HapiSpec htsTokenAllowance() {
...
+                        //                                asHeadlongAddress(
+                        //
+                        // asAddress(spec.registry().getTokenID(FUNGIBLE_TOKEN))),
+                        //                                HapiParserUtil.asHeadlongAddress(notAnAddress),
+                        //                                        asHeadlongAddress(
+                        //
+                        // asAddress(spec.registry().getAccountID(theSpender))))
+                        //                                        .payingWith(OWNER)
+                        //                                        .via("fakeAddressAllowance")
+                        )))
```
mhess-swl, line 303: Is commenting this (and other assertion code blocks) intentional?

tinker-michaelj (approved):
LGTM, tyvm @mustafauzunn !

stoqnkpL (approved):

mhess-swl (approved):

stoqnkpL (approved):

tinker-michaelj (approved):
LGTM, nice work @mustafauzunn !

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/store/contracts/precompile/codec/EvmEncodingFacade.java
```diff
@@ -67,6 +68,14 @@ public Bytes encodeGetTokenType(final int tokenType) {
                 .build();
     }
 
+    public Bytes encodeGetTokenTypeFailure(final ResponseCodeEnum status, final int tokenType) {
```
Nana-EC, line 71: Q: would a smaller footprint and a more general name be an overload or a replacement of the method
```suggestion
   public Bytes encodeGetTokenType(final int tokenType) {
        return encodeGetTokenType(tokenType, SUCCESS.getNumber());
    }

    public Bytes encodeGetTokenType(final int tokenType, final ResponseCodeEnum status) {
```

Same for all the others to reduce duplication and reuse code.

Nana-EC (commented):
Maybe consider code reuse

lukelee-sl (approved):
LGTM.  ty

mhess-swl (approved):

Nana-EC (approved):

PR #11142, opened 2024-01-24
chore: 5967 validate metric naming rules

### LINKED ISSUE #5967: Enforce metric name rules
There are currently no restrictions on what characters can be used in metric names. Some characters cause problems for systems like Prometheus and Elastic to ingest. There are currently scripts that modify metric names to make them compatible with other systems.

Metric names should be validated at the time of creation. There are at least two options for this:

1. Throw an exception when a metric is created with an invalid name
2. Modify the name of the metric to conform to the standards when publishing to externals 

**Elastic converter:** https://github.com/swirlds/infrastructure/blob/master/hedera-utils/PipelineGenerator.py
**Prometheus converter:** `com.swirlds.common.metrics.platform.prometheus.NameConverter`

mxtartaglia-sl:
**Description**:
Since now, developer provided metric names, categories, and units are checked against offending Prometheus naming limitations listed here: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels, in case of finding an offending naming an error level message is logged with an EXCEPTION marker that will cause the JRS test to fail. Still, as it was being done, the matric name gets modified in order to be able to expose the metric regardless.
Unit field has been incorated to the check.   

**Related issue(s)**:

Fixes #5967


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -16,19 +16,53 @@
...
+     * @throws NullPointerException in case {@code adapterType} parameter is {@code null}
+     * @throws NullPointerException in case {@code metric} parameter is {@code null}
+     */
+    protected AbstractMetricAdapter(
+            @NonNull final PrometheusEndpoint.AdapterType adapterType,
+            @NonNull final Metric metric,
+            final boolean unitless) {
         this.adapterType = Objects.requireNonNull(adapterType, "adapterType must not be null");
+        this.values = new CommonMetricValues(unitless, Objects.requireNonNull(metric, "metric must not be null"));
+        this.values.checkAgainst(metric);
```
anthony-swirldslabs, line 60: At first glance this looked like a no-op because the call on line 59 would create the `values` object using the given `metric` object already. However, I see that the implementation of the `CommonMetricValues` constructor actually calls `fix()` on each of the individual values, while the `checkAgainst` uses a simple Objects.equals() to compare the individual values.

So essentially, in these two lines of code we're trying to assert that `fix()` returns its argument verbatim, or else we log a bunch of errors.

This brings the question as to
1. whether this is the intended purpose of this code,
2. and/or whether the calls to `fix()` are required (because this check will fail the moment the `fix()` returns anything different from its input argument),
3. and if they are required, then shouldn't the checkAgainst() take the `fix()` into account?

Could you please clarify?

mxtartaglia-sl: Sure, the "fix()" method is the existing mechanism to replace a metric name component (category/name/unit), so it is valid against some predefined naming rules Prometheus requires. If the result of invoking fix() for a string returns the exact string, the value is a valid name for Prometheus.   
The requirement is that we want to avoid/identify when the metric name is changed under the hood by the framework and inform developers that they should adjust the metric name. That is why it compares the actual parameter vs. the value after being #fix()ed.
Let me rethink the design so it is more explicit; I see how having a check method that only performs log statements collaborates with the idea that something must be wrong.
The initial design had all the code in the constructor and exposed the values after being fixed to the subclasses, but that seemed weak encapsulation.

anthony-swirldslabs: Got it. Thanks for the clarification. This makes sense. I'd suggest to add a code comment (either in checkAgainst() javadoc, or above the call to the method) to briefly describe the intention of the check. This would make the code a lot clearer.

mxtartaglia-sl: @anthony-swirldslabs Let me know if the logic + name change + additional javadoc did a better job expressing the intention

anthony-swirldslabs: That looks perfect! Thank you!

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -40,4 +73,83 @@ public int incAndGetReferenceCount() {
...
+
+        @NonNull
+        private final String unit;
+
+        @NonNull
+        private final String help;
+
+        private ConvertedMetricValues(boolean unitless, final @NonNull Metric metric) {
+            Objects.requireNonNull(metric, "metric must not be null");
+            this.unitless = unitless;
```
hendrikebbers, line 104: I do not get that flag. Metric defines the type field. Why not just check if that field is null?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/threading/framework/internal/MeasuredBlockingQueue.java
```diff
@@ -33,8 +33,8 @@
  */
 class MeasuredBlockingQueue<T> extends AbstractBlockingQueue<T> {
 
-    static final String QUEUE_MAX_SIZE_SUFFIX = "-queueMaxSize";
-    static final String QUEUE_MIN_SIZE_SUFFIX = "-queueMinSize";
+    static final String QUEUE_MAX_SIZE_SUFFIX = "_queueMaxSize";
+    static final String QUEUE_MIN_SIZE_SUFFIX = "_queueMinSize";
```
hendrikebbers, line 37: Will this change metrics names that are current used?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/threading/framework/internal/QueueThreadMetrics.java
```diff
@@ -56,7 +56,7 @@ public QueueThreadMetrics(@NonNull final AbstractQueueThreadConfiguration<?, ?>
      * @return the name of the busy time metric
      */
     public static String buildBusyTimeMetricName(@NonNull final String threadName) {
-        return "thread-busy-" + threadName;
+        return "thread_busy_" + threadName;
```
hendrikebbers, line 59: Will this change metrics names that are current used?

platform-sdk/swirlds-metrics-api/src/main/java/com/swirlds/metrics/api/Metrics.java
```diff
@@ -36,7 +36,7 @@ public interface Metrics extends Startable {
 
     String INTERNAL_CATEGORY = "internal";
     String PLATFORM_CATEGORY = "platform";
-    String INFO_CATEGORY = "platform.info";
+    String INFO_CATEGORY = "platform:info";
```
hendrikebbers, line 39: Will this change metrics names that are current used?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/status/PlatformStatusManager.java
```diff
@@ -75,7 +75,7 @@ public PlatformStatusManager(
 
         this.queue = new QueueThreadConfiguration<PlatformStatusAction>(threadManager)
                 .setComponent("platform")
-                .setThreadName("status-state-machine")
+                .setThreadName("status_state_machine")
```
hendrikebbers, line 78: This does not only change the metrics name. I do not know if platform uses the name somehow else. cc @cody-littley @lpetrovic05

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeMetrics.java
```diff
@@ -61,8 +61,7 @@ class UptimeMetrics {
             .withDescription("The time, in microseconds, required to compute uptime information each round.");
     private final RunningAverageMetric uptimeComputationTime;
 
-    private static final String ROUNDS_SINCE_LAST_CONSENSUS_EVENT = "roundsSinceLastConsensusEvent-";
-    private static final String ROUNDS_SINCE_LAST_JUDGE = "roundsSinceLastJudge-";
+    private static final String ROUNDS_SINCE_LAST_CONSENSUS_EVENT = "roundsSinceLastConsensusEvent_";
```
hendrikebbers, line 64: We really need to check after the merge that all the grafana dashboards still work

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
mxtartaglia-sl: The problem is that prometheus client will fail if we set a unit to a metric that should not have a unit set.
But our code doesn't prevent this from happening; we allow users to set units in any metric type even when Prometheus fails (we have one occurrence that sets "boolean" and other values in tests), which makes it impossible to rely only on a check against null. What was happening before is that the adapters knew not to set the unit field when necessary. I needed that information in the superclass instead of each adapter.
I guess that the final fix would be not to allow to set units on invalid metrics, and we could remove the flag, just that the change was way bigger.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/threading/framework/internal/MeasuredBlockingQueue.java
mxtartaglia-sl: It will not change the name on the Prometheus side; they were already called this after the adapter fixed the name.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/threading/framework/internal/QueueThreadMetrics.java
mxtartaglia-sl: It will not change the name on the Prometheus side; they were already called this after the adapter fixed the name.

platform-sdk/swirlds-metrics-api/src/main/java/com/swirlds/metrics/api/Metrics.java
mxtartaglia-sl: It will not change the name on the Prometheus side; they were already called this after the adapter fixed the name.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/status/PlatformStatusManager.java
mxtartaglia-sl: I checked with @poulok before pushing the change... the naming is less readable but at least consistent with the metrics created under the hood.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeMetrics.java
mxtartaglia-sl: True that, although I haven't found any difference between the resulting metric names and the values in graphana json other than the title of the metric.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -16,19 +16,52 @@
...
+     */
+    protected AbstractMetricAdapter(
+            @NonNull final PrometheusEndpoint.AdapterType adapterType,
+            @NonNull final Metric metric,
+            final boolean unitless) {
         this.adapterType = Objects.requireNonNull(adapterType, "adapterType must not be null");
+        this.values = new ConvertedMetricValues(unitless, metric);
+    }
+
+    protected <C extends SimpleCollector<?>, T extends SimpleCollector.Builder<T, C>> T setCommonValues(
```
cody-littley, line 62: nullity annotation for return type

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -40,4 +73,83 @@ public int incAndGetReferenceCount() {
...
+        @NonNull
+        private final String name;
+
+        @NonNull
+        private final String unit;
+
+        @NonNull
+        private final String help;
+
+        private ConvertedMetricValues(boolean unitless, final @NonNull Metric metric) {
```
cody-littley, line 102:
```suggestion
        private ConvertedMetricValues(final boolean unitless, final @NonNull Metric metric) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -40,4 +73,83 @@ public int incAndGetReferenceCount() {
...
+            Objects.requireNonNull(metric, "metric must not be null");
+            this.unitless = unitless;
+            this.subSystem = fix(metric.getCategory());
+            this.name = fix(metric.getName());
+            this.unit = fix(metric.getUnit());
+            this.help = metric.getDescription();
+            verifyMetricNamingComponents(metric);
+        }
+
+        <C extends SimpleCollector<?>, T extends SimpleCollector.Builder<T, C>> T fill(
```
cody-littley, line 112: nullity annotation for return type

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -40,4 +73,83 @@ public int incAndGetReferenceCount() {
...
+         * identifies changes in the metrics name components (category, name, and unit). If a change is detected, error
+         * log statements with the purpose of failing JRS are generated to inform developers that adjustments to the
+         * metric name may be required.
+         * </p>
+         * It is not throwing exceptions in order to minimize the possibility for runtime errors produced by
+         * non-critical misconfigurations
+         *
+         * @param metric the metric to check against
+         */
+        private void verifyMetricNamingComponents(@NonNull Metric metric) {
```
cody-littley, line 128:
```suggestion
        private void verifyMetricNamingComponents(@NonNull final Metric metric) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/status/PlatformStatusManager.java
cody-littley: Thread names are purely used for understanding thread dumps and for logging. There are no hard dependencies I'm aware of. Anyways, we are very soon going to completely drop support for this style of threads, so naming is not super important.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeMetrics.java
cody-littley: From my perspective, it's not the end of the world if a metric name changes. Most of our current metric names are not so great, and I have intentions to change many of them. 😜

anthony-swirldslabs (approved):

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
```diff
@@ -16,19 +16,52 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.prometheus.client.SimpleCollector;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public abstract class AbstractMetricAdapter implements MetricAdapter {
-
+    private static final Logger log = LogManager.getLogger(AbstractMetricAdapter.class);
```
OlegMazurov, line 31: `Logger logger` is predominantly used in platform code.

OlegMazurov (approved):

cody-littley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
mxtartaglia-sl: Thanks, @OlegMazurov  I should have searched what the pattern was. If it is ok with you, if there is any other suggestion that I should commit, I'll change the name in this PR, but if not, since platform-base is the owner of this file, I can create a follow-up task to replace the name that will require fewer approvals for merging it...

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
hendrikebbers: ok, if I understood it correctly "supportsUnit" would be a better param from my point of view. That one would define if the metric supports a unit. Next to that the Javadoc is missing that would help to understand the param

mxtartaglia-sl: Agreed, `supportsUnit` sounds better. I had the intention that the default false value set the unit instead of not setting it, but `unitless` ended up being a strange name. 
I didn't add Javadoc in this method given that it is a private method on a private inner class, the inner class has a Javadoc, and the containing class also has one explaining the receiving parameters. I'll add one here too, but I'm checking to be on the same page for the future: Should we always add Javadoc or only in public interfaces?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/metrics/platform/prometheus/AbstractMetricAdapter.java
mxtartaglia-sl: Soved

cody-littley (approved):

anthony-swirldslabs (approved):

OlegMazurov (approved):

hendrikebbers (approved):

PR #11144, opened 2024-01-24
feat: Migrate transaction handling to framework

### LINKED ISSUE #8127: Replace consensus handler cycle metrics with fractional timer

### LINKED ISSUE #10984: Componentize Round Handler
- This issue represents the first step of componetizing the round handler, which is to use the framework for the round handler inputs, but leave the outputs as-is (outside the framework)
    - inputs are consensus rounds, and updates to the running hash 
- The [second step](https://github.com/hashgraph/hedera-services/issues/11280) will be done later, and involves using the framework to carry the output of the round handler
    - the output is a signed state

alittley:
- closes #10984 
- closes #8127 
- must merge immediately after https://github.com/swirlds/swirlds-platform-regression/pull/3902

[diagram](https://mermaid.ink/svg/Zmxvd2NoYXJ0IFRECnYwKCgiQCIpKQp2MVsvIkNvbnNlbnN1cyBFdmVudCBTdHJlYW08YnIgLz7CpyIvXQpzdWJncmFwaCB2MlsiQ29uc2Vuc3VzIFBpcGVsaW5lIl0KdjNbIkxpbmtlZCBFdmVudCBJbnRha2UiXQp2NCgoImciKSkKdjVbImluT3JkZXJMaW5rZXI8YnIgLz7KmCJdCmVuZApzdWJncmFwaCB2NlsiRXZlbnQgQ3JlYXRpb24iXQp2N1siZXZlbnRDcmVhdGlvbk1hbmFnZXI8YnIgLz7KmOKZoSJdCnY4WyJmdXR1cmVFdmVudEJ1ZmZlcjxiciAvPsqYIl0Kdjl7eyJmdXR1cmVFdmVudEJ1ZmZlclNwbGl0dGVyIn19CmVuZApzdWJncmFwaCB2MTBbIkdvc3NpcCJdCnYxMXt7Imdvc3NpcCJ9fQp2MTJbInNoYWRvd2dyYXBoPGJyIC8+Z8qYIl0KZW5kCnN1YmdyYXBoIHYxM1siSW50YWtlIFBpcGVsaW5lIl0Kc3ViZ3JhcGggdjE0WyJFdmVudCBIYXNoaW5nIl0KdjE1W1siZXZlbnRIYXNoZXIiXV0KdjE2WyJwb3N0SGFzaENvbGxlY3RvciJdCmVuZApzdWJncmFwaCB2MTdbIkV2ZW50IFZhbGlkYXRpb24iXQp2MThbImV2ZW50RGVkdXBsaWNhdG9yPGJyIC8+ypgiXQp2MTlbImV2ZW50U2lnbmF0dXJlVmFsaWRhdG9yPGJyIC8+ypgiXQp2MjBbImludGVybmFsRXZlbnRWYWxpZGF0b3I8YnIgLz7igKAiXQplbmQKdjIxWyJPcnBoYW4gQnVmZmVyPGJyIC8+ypgiXQplbmQKc3ViZ3JhcGggdjIyWyJQcmVjb25zZW5zdXMgRXZlbnQgU3RyZWFtIl0KdjIzWy8iZXZlbnREdXJhYmlsaXR5TmV4dXMiL10KdjI0Wy8icGNlc1NlcXVlbmNlciIvXQp2MjVbInBjZXNXcml0ZXI8YnIgLz5AypjOniJdCmVuZAp2MjZbIlN0YXRlIEZpbGUgTWFuYWdlbWVudCJdCnYyN1siU3RhdGUgU2lnbmF0dXJlIENvbGxlY3Rpb24iXQp2MjhbWyJhcHBsaWNhdGlvblRyYW5zYWN0aW9uUHJlaGFuZGxlciJdXQp2MjlbImNvbnNlbnN1c1JvdW5kSGFuZGxlcjxiciAvPsKn4piF8J+VkSJdCnYzMFsiaGVhcnRiZWF0Il0KdjMxWy8icGNlc1JlcGxheWVyIi9dCnYzMnt7InJ1bm5pbmdIYXNoVXBkYXRlIn19CnYzM1sic3RhdGVTaWduZXIiXQp2MzQoKCLCpyIpKQp2MzUoKCLKmCIpKQp2MzYoKCLOniIpKQp2MzcoKCLigKAiKSkKdjM4KCgi4piFIikpCnYzOSgoIuKZoSIpKQp2NDAoKCLwn5WRIikpCnYzIC0tICJyb3VuZHMiIC0tPiB2MQp2MyAtLSAicm91bmRzIiAtLT4gdjI5CnYzIC0uICJub24tYW5jaWVudCBldmVudCB3aW5kb3ciIC4tPiB2MzUKdjMgLS0gImZsdXNoIHJlcXVlc3QiIC0tPiB2MzYKdjIxIC0tICJwcmVjb25zZW5zdXMgZXZlbnRzIiAtLT4gdjI3CnYyMSAtLSAicHJlY29uc2Vuc3VzIGV2ZW50cyIgLS0+IHYyOAp2MjEgLS0gInByZWNvbnNlbnN1cyBldmVudHMiIC0tPiB2OAp2MjEgLS0gImV2ZW50cyB0byBzZXF1ZW5jZSIgLS0+IHYyNAp2MjYgLS0gIm1pbmltdW0gaWRlbnRpZmllciB0byBzdG9yZSIgLS0+IHYyNQp2MjcgLS0gInN0YXRlcyIgLS0+IHYyNgp2MjggLS4gImZ1dHVyZXMiIC4tbyB2MzgKdjcgLS4gIm5vbi12YWxpZGF0ZWQgZXZlbnRzIiAuLT4gdjM3CnYxOCAtLSAiZXZlbnRzIHdpdGggdW52YWxpZGF0ZWQgc2lnbmF0dXJlcyIgLS0+IHYxOQp2MjMgLS4gIndhaXQgZm9yIGR1cmFiaWxpdHkiIC4tbyB2NDAKdjE1IC0tICJoYXNoZWQgZXZlbnRzIiAtLT4gdjE2CnYxOSAtLSAidW5vcmRlcmVkIGV2ZW50cyIgLS0+IHYyMQp2OCAtLSAicG9zc2libGUgcGFyZW50IGxpc3RzIiAtLT4gdjkKdjkgLS0gInBvc3NpYmxlIHBhcmVudHMiIC0tPiB2Nwp2MTEgLS0gImV2ZW50cyB0byBoYXNoIiAtLT4gdjE1CnYzMCAtLSAiaGVhcnRiZWF0IiAtLT4gdjM5CnY1IC0tICJsaW5rZWQgZXZlbnRzIiAtLT4gdjMKdjUgLS0gImV2ZW50cyB0byBnb3NzaXAiIC0tPiB2NAp2MjAgLS0gIm5vbi1kZWR1cGxpY2F0ZWQgZXZlbnRzIiAtLT4gdjE4CnYzMSAtLSAiZG9uZSBzdHJlYW1pbmcgcGNlcyIgLS0+IHYwCnYzMSAtLSAiZXZlbnRzIHRvIGhhc2giIC0tPiB2MTUKdjI0IC0tICJ1bmxpbmtlZCBldmVudHMiIC0tPiB2NQp2MjQgLS0gImV2ZW50cyB0byB3cml0ZSIgLS0+IHYyNQp2MjUgLS0gImxhdGVzdCBkdXJhYmxlIHNlcXVlbmNlIG51bWJlciIgLS0+IHYyMwp2MTYgLS0gIm5vbi12YWxpZGF0ZWQgZXZlbnRzIiAtLT4gdjIwCnYzMiAtLSAicnVubmluZyBoYXNoIHVwZGF0ZSIgLS0+IHYzNApjbGFzc0RlZiBzMCBmaWxsOiNmODgsc3Ryb2tlOiMwMDAsc3Ryb2tlLXdpZHRoOjJweApjbGFzcyB2MCx2MzQsdjM1LHYzNix2MzcsdjM4LHYzOSx2NCx2NDAgczAKY2xhc3NEZWYgczEgZmlsbDojY2NjLHN0cm9rZTojMDAwLHN0cm9rZS13aWR0aDoycHgKY2xhc3MgdjEsdjExLHYyMyx2MjQsdjMxLHYzMix2OSBzMQpjbGFzc0RlZiBzMiBmaWxsOiM5Q0Ysc3Ryb2tlOiMwMDAsc3Ryb2tlLXdpZHRoOjJweApjbGFzcyB2MTAsdjEzLHYyLHYyMix2NiBzMgpjbGFzc0RlZiBzMyBmaWxsOiNmZjksc3Ryb2tlOiMwMDAsc3Ryb2tlLXdpZHRoOjJweApjbGFzcyB2MTIsdjE1LHYxNix2MTgsdjE5LHYyMCx2MjEsdjI1LHYyNix2MjcsdjI4LHYyOSx2Myx2MzAsdjMzLHY1LHY3LHY4IHMzCmNsYXNzRGVmIHM0IGZpbGw6I0JFRixzdHJva2U6IzAwMCxzdHJva2Utd2lkdGg6MnB4CmNsYXNzIHYxNCx2MTcgczQK?bgColor=e8e8e8)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/stream/EventStreamManager.java
```diff
@@ -197,119 +202,60 @@ public void stop() {
...
      */
-    public void setStartWriteAtCompleteWindow(final boolean startWriteAtCompleteWindow) {
+    public void updateRunningHash(@NonNull final RunningEventHashUpdate runningEventHashUpdate) {
         if (streamFileWriter != null) {
-            streamFileWriter.setStartWriteAtCompleteWindow(startWriteAtCompleteWindow);
+            streamFileWriter.setStartWriteAtCompleteWindow(runningEventHashUpdate.isReconnect());
         }
+
+        initialHash = new Hash(runningEventHashUpdate.runningEventHash());
+        logger.info(EVENT_STREAM.getMarker(), "EventStreamManager::updateRunningHash: {}", () -> initialHash);
```
anthony-swirldslabs, line 240: Are you sure the `{}` will capture the return value of the lambda and not the lambda itself? If not, then I believe that `toString()` on a lambda itself is hardly useful for diagnostics. But if yes, then why not pass the `initialHash` value directly as a parameter here?

lpetrovic05: I actually see no reason for a lambda if we are just passing an object

lpetrovic05: Yeah, I think the object should just be passed directly, no reason for a lambda

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/StateConfig.java
```diff
@@ -98,7 +95,6 @@ public record StateConfig(
         @ConfigProperty(defaultValue = "false") boolean stateHistoryEnabled,
         @ConfigProperty(defaultValue = "false") boolean debugStackTracesEnabled,
         @ConfigProperty(defaultValue = "emergencyRecovery.yaml") String emergencyStateFileName,
-        @ConfigProperty(defaultValue = "1") int signedStateFreq,
```
lpetrovic05, line 101: why remove this config?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/RoundHandlingMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+            .withUnit("milliseconds");
+    private final LongGauge consensusTimeDeviation;
+
+    private static final LongGauge.Config eventsPerRoundConfig = new LongGauge.Config(
+                    INTERNAL_CATEGORY, "eventsPerRound")
+            .withDescription("The number of events per round")
+            .withUnit("count");
+    private final LongGauge eventsPerRound;
+
+    private final PhaseTimer<ConsensusRoundHandlerPhase> roundHandlerPhase;
```
lpetrovic05, line 57: can you please verify manually whether this works as expected, we don't have any automated checks for metrics

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
```diff
@@ -102,4 +108,8 @@ public record PlatformSchedulersConfig(
         @ConfigProperty(defaultValue = "SEQUENTIAL") TaskSchedulerType stateSignatureCollectorSchedulerType,
         @ConfigProperty(defaultValue = "500") int stateSignatureCollectorUnhandledCapacity,
         @ConfigProperty(defaultValue = "SEQUENTIAL") TaskSchedulerType shadowgraphSchedulerType,
-        @ConfigProperty(defaultValue = "500") int shadowgraphUnhandledCapacity) {}
+        @ConfigProperty(defaultValue = "500") int shadowgraphUnhandledCapacity,
+        @ConfigProperty(defaultValue = "SEQUENTIAL_THREAD") TaskSchedulerType consensusRoundHandlerSchedulerType,
+        @ConfigProperty(defaultValue = "5") int consensusRoundHandlerUnhandledCapacity,
+        @ConfigProperty(defaultValue = "SEQUENTIAL") TaskSchedulerType eventStreamManagerSchedulerType,
```
lpetrovic05, line 114: event stream manager already has a queue and a thread, this should probably be direct?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/stream/EventStreamManager.java
alittley: This was old code just reorganized - removing the lambda

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/RoundHandlingMetrics.java
alittley: Yes - manually verified that this phase metric works correctly.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
alittley: The problem is that the component has multiple input wires, so it isn't possible to do direct scheduling

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/StateConfig.java
alittley: We discussed this and came to consensus to remove the config

anthony-swirldslabs (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/logging/TransactionStateLogger.java
```diff
@@ -235,7 +232,7 @@ public static void logEndTransactionRecord(
      * @param <T> The type of the singleton
      */
     public static <T> void logSingletonRead(@NonNull final String label, @Nullable final ValueLeaf<T> value) {
-        if (logger.isDebugEnabled() && Thread.currentThread().getName().startsWith(HANDLE_THREAD_NAME)) {
+        if (logger.isDebugEnabled() && Thread.currentThread().getName().equals(TRANSACTION_HANDLING_THREAD_NAME)) {
             logger.debug("      READ singleton {} value {}", label, value == null ? "null" : value.getValue());
         }
```
cody-littley, line 237: This statement will always evaluate to `false` after this PR merges. The thread where we handle transactions is from the wiring framework, meaning that the thread name will never contain the string we are checking against.

IMO, adding strong dependencies on the naming schema of platform threads is a code smell. Platform thread names are not public API, and so it's dangerous to depend on the naming schema. But since this code is preexisting, we should probably double check with the services team of its importance before making it defunct.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/RoundHandlingMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+    public void recordEventsPerRound(final int eventCount) {
+        eventsPerRound.set(eventCount);
+    }
+
+    /**
+     * Records the consensus time.
+     *
+     * @param consensusTime the consensus time of the last transaction in the round that is currently being handled
+     */
+    public void recordConsensusTime(final Instant consensusTime) {
```
cody-littley, line 97:
```suggestion
    public void recordConsensusTime(@NonNull final Instant consensusTime) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
cody-littley: You can lie and call it `DIRECT_STATELESS`, but only if it is safe for multiple threads to call into it simultaneously.

I've been thinking we should probably rename `DIRECT_STATELESS` into `DIRECT_THREAD_SAFE` or something similar. The important thing for this kind of scheduler is to be thread safe, and something that is stateless is trivially thread safe. But it isn't the only way you can be thread safe.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/ConsensusRoundHandlerWiring.java
```diff
@@ -0,0 +1,61 @@
...
+        @NonNull InputWire<ConsensusRound> roundInput,
+        @NonNull InputWire<RunningEventHashUpdate> runningHashUpdateInput,
+        @NonNull Runnable flushRunnable) {
+    /**
+     * Create a new instance of this wiring.
+     *
+     * @param taskScheduler the task scheduler for this wiring object
+     * @return the new wiring instance
+     */
+    public static ConsensusRoundHandlerWiring create(@NonNull final TaskScheduler<ConsensusRound> taskScheduler) {
```
cody-littley, line 44:
```suggestion
    @NonNull
    public static ConsensusRoundHandlerWiring create(@NonNull final TaskScheduler<ConsensusRound> taskScheduler) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/EventStreamManagerWiring.java
```diff
@@ -0,0 +1,63 @@
...
+        @NonNull InputWire<RunningEventHashUpdate> runningHashUpdateInput,
+        @NonNull Runnable flushRunnable) {
+
+    /**
+     * Create a new wiring object
+     *
+     * @param taskScheduler the task scheduler to use
+     * @return the new wiring object
+     */
+    public static EventStreamManagerWiring create(@NonNull final TaskScheduler<Void> taskScheduler) {
```
cody-littley, line 46:
```suggestion
    @NonNull
    public static EventStreamManagerWiring create(@NonNull final TaskScheduler<Void> taskScheduler) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/RunningHashUpdaterWiring.java
```diff
@@ -0,0 +1,52 @@
...
+        @NonNull InputWire<RunningEventHashUpdate> runningHashUpdateInput,
+        @NonNull OutputWire<RunningEventHashUpdate> runningHashUpdateOutput) {
+
+    /**
+     * Create a new wiring object
+     *
+     * @param taskScheduler the task scheduler to use
+     * @return the new wiring object
+     */
+    public static RunningHashUpdaterWiring create(@NonNull final TaskScheduler<RunningEventHashUpdate> taskScheduler) {
```
cody-littley, line 42:
```suggestion
    @NonNull
    public static RunningHashUpdaterWiring create(@NonNull final TaskScheduler<RunningEventHashUpdate> taskScheduler) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/generate-platform-diagram.sh
```diff
@@ -0,0 +1,30 @@
+########################################################################################################################
+## Creates high level groupings of platform components.
+## Substitutes spammy wires, and collapses groupings with transformers.
```
cody-littley, line 3: I suggest removing this comment, not really necessary when there is only one command in this file.

Also, please add `#!/usr/bin/env bash` as the first line in this script.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/logging/TransactionStateLogger.java
alittley: > This statement will always evaluate to false after this PR merges

This PR changes the value of the constant to match the new name from the framework, so the functionality of this condition doesn't change post merge. 

> adding strong dependencies on the naming schema of platform threads is a code smell

I agree, but I think it's too late for now. This was recently implemented, https://github.com/hashgraph/hedera-services/pull/8633. If we remove the link, it would need to be paired with a replacement solution

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
alittley: @cody-littley Hmm, it looks to me like the streams used in `EventStreamManager` actually *aren't* threadsafe. There wasn't any protection before, but I suspect this might be thread unsafe code that happens to not conflict, based on when the various methods are called.

We could go the `DIRECT_STATELESS` route if we want to keep trusting the old behavior patterns, or leave this as a `SEQUENTIAL` scheduler (potentially with a much smaller capacity). I don't think touching the multistream code to make it threadsafe is a route we would ever want to take

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
cody-littley: If there is a thread safety concern, better to stay with `SEQUENTIAL`, in my opinion.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/logging/TransactionStateLogger.java
cody-littley: Ah, ok. This is a sequential thread scheduler, so we actually do set a name for it. I'm with this for the time being.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
lpetrovic05: Based on recent issues and Kelly's message, we should try not to change the threading model if possible, I will take a look at the event stream manager and see what the conflicts could be

lpetrovic05: > I will take a look at the event stream manager
 
I regret this now

lpetrovic05: Even sequential would not be thread safe, since this thing has internal threads

lpetrovic05: Based on the discussion, we will add simple thread safety to this class

cody-littley (approved):

poulok:
This PR also closes #8127

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/eventhandling/ConsensusRoundHandler.java
```diff
@@ -456,43 +254,48 @@ private void updatePlatformState(final ConsensusRound round) {
...
     private void createSignedState() throws InterruptedException {
-        final CycleTimingStat ssTimingStat = consensusHandlingMetrics.getNewSignedStateCycleStat();
-        ssTimingStat.startCycle();
+        if (freezeRoundReceived) {
+            // Let the swirld state manager know we are about to write the saved state for the freeze period
+            swirldStateManager.savedStateInFreezePeriod();
+        }
 
+        handlerMetrics.setPhase(GETTING_STATE_TO_SIGN);
         // create a new signed state, sign it, and send out a new transaction with the signature
```
poulok, line 292: out of date comment - this method does not sign the state or send out a transaction

alittley: Fixed

edward-swirldslabs (dismissed):

jeromy-cannon (dismissed):

cody-littley (approved):

JeffreyDallas (approved):

mxtartaglia-sl (approved):

Neeharika-Sompalli (approved):
hedera-base changes LGTM! Thanks @alittley

anthony-swirldslabs (dismissed):

lpetrovic05 (approved):

jeromy-cannon (approved):

cody-littley (approved):

Neeharika-Sompalli (approved):
LGTM !

Issue #11141, opened 2024-01-24
Replay events from round `160290501` and validate record files for differential testing


Neeharika-Sompalli:
Replay events from round `160290501` and validate record files for differential testing.

Neeharika-Sompalli:
Pushed fixes to this PR https://github.com/hashgraph/hedera-services/pull/11148
```
Parsing stream @ /Users/neeharikasompalli/Documents/Hedera/Repos/repo2/hedera-services/platform-sdk/swirlds-cli/hedera-node/data/recordStreams/record0.0.30
 ➡️  Read 68900 entries
These streams are identical ☺️
```

PR #11148, opened 2024-01-24
chore: add `RcDiff` tool

### LINKED ISSUE #10955: Automate the comparison of a mono record file with a modular record file

tinker-michaelj:
**Description**:
 - Closes #10955 
 - Closes https://github.com/hashgraph/hedera-services/issues/11141
 - Extract the `com.hedera.node.app.service.mono.utils.forensics` package into `hapi-utils/` for reuse.
 - Create an `RcDiff` picocli tool that basically wraps `OrderedComparison` and uses a new `FULLY_DETERMINSTIC` snapshot match mode for faster analysis.
 - Introduce `DumpCheckpoint`s at which mono or mod states can be dumped to a uniform text representation.

**And also**:
 - Migrate the next mono-service entity number `n` to `n-1` since mod-service increments **before** assigning entity id numbers.
 - Avoid `currentRecordFileWriter` NPE in the case of event stream recovery or reconnect (see comment).
 - Update `signedTransactionBytes` when syncing body id from record.
 - Use `null` for treasury-owned NFT `ownerId` and `spenderId` for a NFT without approved spender.
 - Always put debit first in synthetic `transferToken()` dispatch.
 - Use `ContractID` for migrated bytecode map.
 - Don't externalize `paid_staking_rewards` if all rewards are 0.
 - Initial `Nft#NftID` field correctly during migration.
 - Migrate aliases into `ALIASES` state during account migration.
 - Add `ifNotInCi()` util verb to "escape" single flaky operations---(use to stabilize `approvalFallbacksRequiredWithoutTopLevelSigAccess()`). 
 - Use copy-and-release pattern when migrating `ContractService` `VirtualMap`s.
 - Deal with a special case in `TRANSFER_TOKEN` gas calculation for mono-service fidelity.
 - Set the view gas requirement in `TotalSupplyCall`.
 - Fix `GRANT_APPROVAL` output encoding.
 - Don't charge service fee on resolved `DUPLICATE_TRANSACTION` status.
 - Waive target file signing requirements for special file updates.
 - Ensure receipt exchange rates immediately reflect the result of a `0.0.112` `FileUpdate`.
 - Add `onNewRecoveredState()` callback to `HederaLifecycles`, register during `EVENT_STREAM_RECOVERY` and use to ensure the final record stream file is closed after recovery finishes.


Neeharika-Sompalli (commented):
Are the record files inside `forensics` deleted intentionally ?

Neeharika-Sompalli (approved):
One question on the deleted files, if any unit tests will break. LGTM otherwise. tyvm @tinker-michaelj !

hedera-node/test-clients/rcdiff/check.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+if [ $# -lt 2 ]; then
+  EXPECTED_LOC='/Users/michaeltinker/Dev/hedera-services/platform-sdk/swirlds-cli/mono-record-streams'
+  ACTUAL_LOC='/Users/michaeltinker/Dev/hedera-services/platform-sdk/swirlds-cli/hedera-node/data/recordStreams/record0.0.30'
```
mhess-swl, line 5: We probably want at least a `TODO` comment to change these

Neeharika-Sompalli (dismissed):

iwsimon (approved):
LGTM! Thanks @tinker-michaelj

hedera-node/test-clients/build.gradle.kts
```diff
@@ -230,6 +230,30 @@ val yahCliJar =
...
     }
 
+val rcdiffConfig = "rcdiffConfig"
+
+configurations { create(rcdiffConfig) { extendsFrom(configurations.runtimeClasspath.get()) } }
+
+dependencies {
+    // Example: Add a dependency to your custom configuration
+    "rcdiffConfig"("info.picocli:picocli:4.6.3")
+}
```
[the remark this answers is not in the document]
tinker-michaelj: This is a one-off configuration needed to tell [the `ShadowJar` plugin](https://github.com/johnrengelman/shadow) to bundle `picocli` in the executable JAR; don't know how might be accomplished with the other plugins in the mix. 😅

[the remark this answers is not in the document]
nathanklick: @tinker-michaelj The better way to handle this is to declare a separate source set for rcdiff and move the code to said source set.

nathanklick (changes requested):
We should separate the code for rcdiff into a separate source set and stop mixing everything in the main source set.

hedera-node/test-clients/build.gradle.kts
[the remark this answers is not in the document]
tinker-michaelj: Done.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleHederaState.java
```diff
@@ -177,6 +179,14 @@ public void init(
...
             final InitTrigger trigger,
             final SoftwareVersion deserializedVersion) {
+        // If we are initialized for event stream recovery, we have to register an
+        // extra listener to make sure we call all the required Hedera lifecycles
+        if (trigger == EVENT_STREAM_RECOVERY) {
+            final var notificationEngine = platform.getNotificationEngine();
+            notificationEngine.register(
+                    NewRecoveredStateListener.class,
+                    notification -> lifecycles.onNewRecoveredState(notification.getSwirldState()));
+        }
```
mhess-swl, line 189: cool :)

mhess-swl (approved):

Neeharika-Sompalli (dismissed):
LGTM !

mhess-swl (approved):

hedera-node/test-clients/build.gradle.kts
```diff
@@ -230,6 +232,21 @@ val yahCliJar =
         }
     }
 
+val rcdiffJar =
+    tasks.register<ShadowJar>("rcdiffJar") {
+        exclude(listOf("META-INF/*.DSA", "META-INF/*.RSA", "META-INF/*.SF", "META-INF/INDEX.LIST"))
+        from(sourceSets["rcdiff"].output)
```
nathanklick, line 238: @jjohannes Is this the correct way to reference the source set?

hedera-node/test-clients/build.gradle.kts
```diff
@@ -230,6 +232,21 @@ val yahCliJar =
...
+        archiveFileName.set("rcdiff.jar")
+        configurations = listOf(project.configurations.getByName("rcdiffRuntimeClasspath"))
```
nathanklick, line 240: @jjohannes Please take a look at this as well.

hedera-node/test-clients/build.gradle.kts
```diff
@@ -46,6 +46,8 @@ sourceSets {
     // Needed because "resource" directory is misnamed. See
     // https://github.com/hashgraph/hedera-services/issues/3361
     main { resources { srcDir("src/main/resource") } }
+
+    create("rcdiff") { java.srcDir("src/rcdiff/java") }
```
nathanklick, line 50: @jjohannes Can this be simplified at all or the hardcoded path specification removed?

jjohannes:
```suggestion
    create("rcdiff")
```

Yes. We do not need the `srcDir` part as this this is the default.

hedera-node/test-clients/build.gradle.kts
jjohannes: Note that we configure all _shadowJar_ tasks are configured to have `from(sourceSets.main.output)` as default. If this should be replaced in his case, you should use `setFrom()` (instead of `from()`).

```suggestion
        setFrom(sourceSets["rcdiff"].output)
```

hedera-node/test-clients/build.gradle.kts
jjohannes: This is okay. You could also use the `[...]` notation if you want to be consistent – `project.configurations["rcdiffRuntimeClasspath"]` – but that us just a different syntax.

hedera-node/test-clients/build.gradle.kts
```diff
@@ -258,6 +275,19 @@ val cleanValidation =
         delete(File(project.file("validation-scenarios"), "ValidationScenarios.jar"))
     }
 
+val cleanRcdiff =
+    tasks.register<Delete>("cleanRcdiff") {
+        group = "build"
+        delete(File(project.file("rcdiff"), "rcdiff.jar"))
```
jjohannes, line 281: Works, but cleaner would be:

```suggestion
        delete(layout.projectDirectory.dir("rcdiff"))
```

I think the `"rcdiff.jar"` here has no effect. There is no such file in the root of the proejct.

hedera-node/test-clients/build.gradle.kts
```diff
@@ -258,6 +275,19 @@ val cleanValidation =
...
+    }
+
+val copyRcdiff =
+    tasks.register<Copy>("copyRcdiff") {
+        group = "copy"
+        from(rcdiffJar)
+        into(project.file("rcdiff"))
```
jjohannes, line 288: Clearer:

```suggestion
        into(layout.projectDirectory.dir("rcdiff"))
```

I wonder if the additional _copy_ and _clean_ tasks are needed at all. The Jar task could be configured to write the Jar directly to the desired location. And when is the clean needed? (does not need to be addressed now)

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: `setFrom()` doesn't seem to exist.

jjohannes: Sorry my bad. I mixed that up with other APIs. So I think we can't override it IIRC.
Maybe it is okay for this case to leave it like this?
Otherwise we would need to do adjustments in `com.hedera.hashgraph.shadow-jar.gradle.kts` and then check every of the other `ShadowJar` tasks. Maybe do this in a follow up, if we care.

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: Removed.

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: The JAR is in _.gitignore_ but is built here.

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: I'll leave it for a follow-up, this is already past the limits of my Gradle knowledge. 😁

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: Removed both _copy_ and _clean_ and set `destinationDirectory` in Jar.

mhess-swl (approved):

nathanklick (approved):

PR #11151, opened 2024-01-25
fix: 114: handle ParseException from PBJ Codec.parse()

anthony-swirldslabs:
**Description**:
* Upgrading PBJ dependency to 0.7.14
* Replacing handling of the `IOException` thrown by methods in the Codec interface with equivalent handling of the new `ParseException` introduced in PBJ 0.7.14

**Related issue(s)**:

Fixes #https://github.com/hashgraph/pbj/issues/114

**Notes for reviewer**:
The Gradle `test` target passes in each of the modified repositories.

**Checklist**


hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/codecs/EntityNumCodec.java
jsync-swirlds, line 1: hmmm, I'm not sure this class even has a purpose anymore.
Probably it should be removed, but perhaps not in this PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/StateUtils.java
```diff
@@ -70,7 +71,7 @@ public static <T> int writeToStream(
      */
     @NonNull
     public static <T> T readFromStream(@NonNull final InputStream in, @NonNull final Codec<T> codec)
```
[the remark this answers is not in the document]
anthony-swirldslabs: That's a good catch. I missed the semantics of this particular method/class while I was preparing this massive change. Let me fix this in the next commit.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/codecs/EntityNumCodec.java
anthony-swirldslabs: I'm not sure about this. Please file an issue against the team that owns this particular class to decide this.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/StateUtils.java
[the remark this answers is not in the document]
anthony-swirldslabs: I just fixed that in the latest commit. Please have a look.

jsync-swirlds (approved):

nathanklick (approved):
LGTM wrt RE owned files

povolev15 (approved):
LGTM

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/query/QueryWorkflowImpl.java
```diff
@@ -297,15 +298,15 @@ public void handleQuery(@NonNull final Bytes requestBuffer, @NonNull final Buffe
...
-            // This should technically not be possible. The data buffer supplied
-            // is either based on a byte[] or a byte buffer, in both cases all data
-            // is available and a generic IO exception shouldn't happen. If it does,
-            // it indicates the data could not be parsed, but for a reason other than
-            // those causing an MalformedProtobufException or UnknownFieldException.
-            logger.warn("Unexpected IO exception while parsing protobuf", e);
+        } catch (ParseException e) {
+            switch (e.getCause()) {
+                case MalformedProtobufException ex:
+                    break;
```
Neeharika-Sompalli, line 304: We should probably throw exception in both these cases here to keep same behavior
```
throw new StatusRuntimeException(Status.INVALID_ARGUMENT);
```

anthony-swirldslabs: This exception is thrown in all of these cases right below the `switch()` at line 310. This `switch` statement is only used to decide if we want to log a warning. It's coded to implement the exact same logic as was present before with individual `catch()` clauses.

Neeharika-Sompalli (approved):
`hedera-base` changes LGTM , except for one comment. Tyhanks @anthony-swirldslabs

Nana-EC (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/query/QueryWorkflowImpl.java
Neeharika-Sompalli: ah yes ! makes sense. Thanks @anthony-swirldslabs

Issue #11103, opened 2024-01-23
Analyze mono-service behavior for `AtomicCryptoTransferHTSSuite`


MrValioBg:
In the AtomicCryptoTransferHTSSuite there are 3 tests that are failing against mono-service & modularized codebase:
- cryptoTransferAllowanceFungibleToken
- cryptoTransferAllowanceHbarToken
- cryptoTransferForHbarOnly

PR #11181, opened 2024-01-26
fix: Fuzzy match AtomicCryptoTransferSuite

### LINKED ISSUE #10976: Enable fuzzy record matching for AtomicCryptoTransferHTSSuite

MrValioBg:
**Description**:
Fuzzy match AtomicCryptoTransferSuite.

The disabled tests were initially failing in mono, but passing in modularized codebase.

After adding the CRYPTO_TRANSFER_V2 in the isClassicCall method, the tests failing in mono also started to fail in modularized code.

There is a new open issue to revisit those tests - #11103

**Related issue(s)**:

Fixes #10976 

**Checklist**


mhess-swl (commented):
Looks good from the services side 👍

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
```diff
@@ -97,12 +97,19 @@ public TransactionBody decodeCryptoTransfer(
...
-                .transfers(convertingMaybeApprovedAdjustments(((Tuple) call.get(0)).get(0), addressIdConverter)));
+        final var transferList = convertingMaybeApprovedAdjustments(((Tuple) call.get(0)).get(0), addressIdConverter);
+
+        final var cryptoTransfersBody = tokenTransfers(convertTokenTransfers(
+                call.get(1),
+                this::convertingMaybeApprovedAdjustments,
+                this::convertingMaybeApprovedOwnershipChanges,
+                addressIdConverter));
+
+        if (transferList.accountAmounts() != null
```
tinker-michaelj, line 108: ⚠️ Shouldn't it be impossible for this to be `null`? 

Even if _currently_ there's no risk of hiding a bug by working around an incorrect `null` here, adding such checks makes it easier for bugs to hide in the future.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AtomicCryptoTransferHTSSuite.java
```diff
@@ -133,7 +137,8 @@ public List<HapiSpec> getSpecsInSuite() {
         });
     }
 
-    @HapiTest
+    // This test is failing against mono-service.
```
tinker-michaelj, line 140: In the PR description, you wrote,

> After adding the CRYPTO_TRANSFER_V2 in the isClassicCall method, the tests failing in mono also started to fail in modularized code.

Does that mean this test now fails as a `@HapiTest` too?

MrValioBg: Yes, exactly

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
MrValioBg: Yes, you are right, since in the Builder accountAmounts() method in TransferList, we use List.of() for the input.

I fixed it, great catch!

mhess-swl (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
tinker-michaelj: Awesome, thanks @MrValioBg !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AtomicCryptoTransferHTSSuite.java
tinker-michaelj: Ah got it 💯

tinker-michaelj (approved):
LGTM, thanks @MrValioBg !

PR #11185, opened 2024-01-26
fix: 10315 Add ALLOW_EMPTY_ERROR_MSG snapshot mode (ERCPrecompileSuite fuzzy match)

### LINKED ISSUE #10315: Enable fuzzy record matching for ERCPrecompileSuite

JivkoKelchev:
**Description**:

This PR modifies SnapshotModeOp in order to allow empty error messages from mono .
* Add ALLOW_EMPTY_ERROR_MSG
* Add TransactionID nonce normalisation.

**Related issue(s)**:

#10315 

Fixes specs:
erc20TransferFromSelf
transferErc20TokenFromContractWithNoApproval
someErc20NegativeTransferFromScenariosPass

**Notes for reviewer**:

When revert precompile calls in mod we have [this](https://github.com/hashgraph/hedera-services/blob/89bd799b91e687f60f8812e5e3f409d52f5a9fec/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java#L205) to propagate the error msg to the parent frame, but in mono there is no such functionality and the parent records has no error message.
So instead of stoping the propagation of the revert reason, I think it is better to skip this check by introducing ALLOW_EMPTY_ERROR_MSG match mode.

When Node stake update transaction occur all fowling transaction nonces are offset by 1.


tinker-michaelj (approved):
The root cause here was mod-service failure to set a revert reason when a system contract returned a revert result...not _entirely_ clear every detail of how it caused the observed failure mode; but we don't want this new snapshot mode, so reverted.

Also reverted the `NodeStakeUpdate` nonce manipulation, since they would have applied to _every_ record being matched from a snapshot, instead of a single `TransactionID` and its children.

tinker-michaelj (approved):
LGTM

Issue #11067, opened 2024-01-19
Metadata HIPs - Update TokenInfo with the new fields


Neeharika-Sompalli:
Depends on #10932 
Confirm if this is needed or is there a work around to test the transaction changes.

- [x] Update `TokenGetInfoHandler.infoForToken()` to include the newly created fields
- [x] Update `GetTokenInfoResourceUsage.usageGiven`  to include metadata and metadataKey

Tests:

- [x] Update `HapiGetTokenInfo` to include new fields and its setters. Add expectations similar to other fields in the class
- [x] Validate `getTokenInfo` returns the information when `TokenCreateHandler` changes are complete.

PR #11199, opened 2024-01-26
fix: 111067: Metadata HIPs - Update Token Info with new fields

thomas-swirlds-labs:
**Description**:

Metadata HIPs - Update TokenInfo with the new fields to support metadata changes for TokenCreate, TokenUpdate and associated tests

- Update TokenGetInfoHandler.infoForToken() to include the newly created fields
- Update GetTokenInfoResourceUsage.usageGiven to include metadata and metadataKey

**Related issue(s)**:

Fixes: https://github.com/hashgraph/hedera-services/issues/11067


Neeharika-Sompalli (approved):
This all makes sense to me @thomas-swirlds-labs ! 💯  Now you should be able to assert the changes in `TokenCreateSpecs` tests.

PR #11213, opened 2024-01-27
fix: Ensure that the pending creation customizer applies to the address being created

lukelee-sl:
**Description**:
Ensure that the pending creation customizer applies to the address being created


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCreateSuite.java
```diff
@@ -108,7 +107,7 @@
 import org.apache.tuweni.bytes.Bytes32;
 import org.junit.jupiter.api.Tag;
 
-@HapiTestSuite
+// @HapiTestSuite
```
Nana-EC, line 110: Q: SHould this be reverted?
```suggestion
 @HapiTestSuite
```

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/ContractCustomizer.java
```diff
@@ -65,7 +71,10 @@ public ContractCustomizer(final @Nullable JKey cryptoAdminKey, final HederaAccou
      * @return an appropriate top-level customizer
      */
     public static ContractCustomizer fromHapiCreation(
-            final JKey decodedKey, final Instant consensusTime, final ContractCreateTransactionBody op) {
+            final JKey decodedKey,
+            final Instant consensusTime,
+            final ContractCreateTransactionBody op,
+            final Address customizerAppliesToAddress) {
```
Nana-EC, line 77: nit: need to update method comment

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCreateSuite.java
lukelee-sl: Fixed

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ledger/accounts/ContractCustomizer.java
```diff
@@ -99,13 +108,15 @@ public static ContractCustomizer fromHapiCreation(
      * @return an appropriate child customizer
      */
     public static ContractCustomizer fromSponsorContract(
-            final AccountID sponsor, final TransactionalLedger<AccountID, AccountProperty, HederaAccount> ledger) {
+            final AccountID sponsor,
+            final TransactionalLedger<AccountID, AccountProperty, HederaAccount> ledger,
+            final Address customizerAppliesToAddress) {
```
Nana-EC, line 113: Q: this is the address of the contract to be created? 
Presumably this address is the result of id to long zero or CREATE2 right? Or is it a smaller set of options?

lukelee-sl: Yes, the new address where we should be creating the contract.

tinker-michaelj (approved):
LGTM, tyvm @lukelee-sl !

Nana-EC (approved):

tinker-michaelj (approved):
Awesome, LGTM @lukelee-sl !

Nana-EC (approved):

Nana-EC:
Single test failure was the new test added here and only affected modularized code.
The failure will be fixed in develop but if covered by the mono code here. @lukelee-sl to add more details

lukelee-sl:
There is one failing spec in the `HAPI Tests (Smart Contract)` check above.  It is a spec added to test the particular issue that this pr looks to address.  The test passes in mono service but not in modularized service.  Looking deeper, the contract creation fails in modularized service because the address it is trying to call `0x000000000000000000636F6e736F6c652e6c6f67` is not recognized as a valid address and thus the ContractCreate fails.  As this pr is addressing an issue in mono service,  the failing test is less alarming because the transaction fails.  It is an undesirable but not harmful outcome.

PR #11214, opened 2024-01-27
fix: Ensure that the pending creation customizer applies to the address being created

lukelee-sl:
**Description**:
Ensure that the pending creation customizer applies to the address being created.  Otherwise create a ghost contract.


hedera-node/test-clients/src/main/resource/contract/contracts/CallInConstructor/CallInConstructor.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.8.0;
+
+contract CallInConstructor {
+    address constant CONSOLE_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
+
+    constructor() {
```
Nana-EC, line 6: Q: Can we expand the coverage with a method (not the constructor) doing the same in a follow up execution

Nana-EC (commented):
LG
Added a suggestion for an additional scenario

hedera-node/test-clients/src/main/resource/contract/contracts/CallInConstructor/CallInConstructor.sol
lukelee-sl: added

Nana-EC (approved):

tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

PR #11216, opened 2024-01-27
perf: Execute each cache warming on individual virtual thread

### LINKED ISSUE #11215: Do cache warming on different thread
The initial implementation of cache warming can result in I/O operations carried out in the fork-join pool, which is dangerous. We should do the actual cache warming on a virtual thread.

netopyr:
This PR moves the warming of the cache to a virtual thread.

Closes #11215


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #11222, opened 2024-01-29
perf: Turn workflows into Singletons

netopyr:
This PR turns `HandleWorkflow` and `QueryWorkflowImpl` into Singletons as originally intended.


Neeharika-Sompalli (approved):
LGTM !

PR #11243, opened 2024-01-30
fix: static call with selfdestruct to system account between 0.0.751 and 0.0.999 results in FAIL_INVALID

### LINKED ISSUE #11034: Setting beneficiary to system account[0.0.751-0.0.999] via selfdestruct results in fail invalid
### Description

Executing selfdestruct via ContractCallQuery/ContractCallLocal and setting beneficiary address to system account(i.e. 0.0.999) fails in precheck with `FAIL_INVALID`.

**Note**: This also fails for accounts above 1000.

### Steps to reproduce

1. Create a contract that consists of selfdestruct function:
```
function destroyContract(address payable beneficiary) public {
        selfdestruct(beneficiary);
    }
```
2. Execute the function against 0.0.999 address

Expected result: This should fail with `LOCAL_CALL_MODIFICATION_EXCEPTION`
Actual result: The call fails with `FAIL_INVALID` in precheck.
image

### Additional context

_No response_

### Hedera network

other

### Version

0.46.0

### Operating system

None

natanasow:
**Description**:

**Related issue(s)**:

Fixes #11034

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaSelfDestructOperationV046.java
```diff
@@ -99,14 +99,16 @@ private ExceptionalHaltReason reasonToHalt(
...
+        if (updater.trackingAccounts() != null) {
+            if (updater.contractIsTokenTreasury(toBeDeleted)) {
+                return HederaExceptionalHaltReason.CONTRACT_IS_TREASURY;
+            }
+            if (updater.contractHasAnyBalance(toBeDeleted)) {
+                return HederaExceptionalHaltReason.TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES;
+            }
+            if (updater.contractOwnsNfts(toBeDeleted)) {
+                return HederaExceptionalHaltReason.CONTRACT_STILL_OWNS_NFTS;
+            }
```
stoqnkpL, line 111: AFAICT we're only using the updaters here for read-only operations. 
In theory, we should be able to detect a static message frame by calling `frame.isStatic()` and allow the frame to read from the state without being able to write. Something similar is implemented in `HTSPrecompiledContract` in order to allow query operations to execute successfully [here](https://github.com/hashgraph/hedera-services/blob/83333db0e2a23a33ad999e7b85f7ed8d7e58bfba/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java#L201). I think we can use the tranckingLedgers from the updater which _should_ be instantiated at this point.
This way, queries and simulations would return a more correct result.

stoqnkpL (changes requested):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaSelfDestructOperationV046.java
stoqnkpL: After discussing this with Nikolay, we realized that there's a cleaner solution --- we should add a check that the original execute() method [performs](https://github.com/hyperledger/besu/blob/750580dcca349d22d024cc14a8171b2fa74b505a/evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java#L51) but is missing from our implementation

natanasow: :rocket:

stoqnkpL (approved):
LGTM

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaSelfDestructOperationV046.java
```diff
@@ -74,6 +74,9 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
         final var updater = (HederaStackedWorldStateUpdater) frame.getWorldUpdater();
         final var beneficiaryAddress = Words.toAddress(frame.getStackItem(0));
         final var toBeDeleted = frame.getRecipientAddress();
+        if (frame.isStatic()) {
```
tinker-michaelj, line 77: This check is already done during `super.execute()`. Why does it need to be added here?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomSelfDestructOperation.java
```diff
@@ -71,11 +71,13 @@ public OperationResult execute(@NonNull final MessageFrame frame, @NonNull final
 
             final var tbdAddress = frame.getRecipientAddress();
             final var proxyWorldUpdater = (ProxyWorldUpdater) frame.getWorldUpdater();
-            // Enforce Hedera-specific restrictions on account deletion
-            final var maybeHaltReason =
-                    proxyWorldUpdater.tryTrackingSelfDestructBeneficiary(tbdAddress, beneficiaryAddress, frame);
-            if (maybeHaltReason.isPresent()) {
-                return haltFor(null, 0, maybeHaltReason.get());
+            // Enforce Hedera-specific restrictions on account deletion for non-static frames
+            if (!frame.isStatic()) {
```
tinker-michaelj, line 75: There's no obvious reason we need to wrap this in this `if`. Why  was this added?

tinker-michaelj (changes requested):
If these redundant `frame.isStatic()` checks are necessary, I am skeptical that the existing code is structured correctly. 

Would like to understand why these are added, and what fails if they aren't there.

natanasow:
> If these redundant `frame.isStatic()` checks are necessary, I am skeptical that the existing code is structured correctly.
> 
> Would like to understand why these are added, and what fails if they aren't there.

Hey @tinker-michaelj, without that preliminary check, a null pointer error will be thrown before executing `super.execute()` for static calls. [Here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaSelfDestructOperationV046.java#L82) we're executing `reasonToHalt()` which checks whether the deleted contract is a treasury (there are several similar checks afterward). Under the hood, `updater.contractIsTokenTreasury` uses `trackingAccounts().get()` which in static call context is null and NPE is thrown. 

Two approaches could resolve this issue in the mono codebase:
- copy the [besu](https://github.com/hyperledger/besu/blob/750580dcca349d22d024cc14a8171b2fa74b505a/evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java#L51) check before [these specific checks](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/operation/HederaSelfDestructOperationV046.java#L102) in `reasonToHalt`
- edit both `contractIsTokenTreasury`, `contractHasAnyBalance`, `contractOwnsNfts` to return false if `trackingAccounts()` is null

In the modularized codebase a `java.lang.UnsupportedOperationException: Cannot track deletion in query context` has been thrown [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomSelfDestructOperation.java#L76) for `tryTrackingSelfDestructBeneficiary`.

tinker-michaelj (approved):
LGTM, ty @natanasow !

PR #11244, opened 2024-01-30
fix: contracts.evm.allowCallsToNonContractAccounts flag misbehaviour

### LINKED ISSUE #11223: Feature flag misbehaviour (contracts.evm.allowCallsToNonContractAccounts)
The feature flag contracts.evm.allowCallsToNonContractAccounts should only control whether we allow calls to non existing accounts that result in no-op success. 
Currently, switching the flag also changes the behaviour when calls are made to token addresses.
The tests in the HRCPrecompile suite reproduce the issue.

thenswan:
This PR fixes #11223 -- `contracts.evm.allowCallsToNonContractAccounts` flag misbehavior.
The problem was when we implemented [this feature](https://github.com/hashgraph/hedera-services/pull/9628), we refactored [this code ](https://github.com/hashgraph/hedera-services/blob/83333db0e2a23a33ad999e7b85f7ed8d7e58bfba/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L321)and now the calls to token addresses fail validation. I have added `isTokenAccount` to validation in order to correspond to pre-refactor logic.


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
```diff
@@ -359,10 +359,10 @@ private Account extractAndValidateReceiver(
             } else {
                 if (!OK.equals(isUsableContract)) {
                     // call to non-existing contract flow
-                    validateTrue(!EVM_VERSION_0_30.equals(properties.evmVersion()), isUsableContract);
-                    validateTrue(!EVM_VERSION_0_34.equals(properties.evmVersion()), isUsableContract);
-                    validateTrue(!EVM_VERSION_0_38.equals(properties.evmVersion()), isUsableContract);
-                    validateTrue(properties.allowCallsToNonContractAccounts(), isUsableContract);
+                    validateTrue(isTokenAccount || !EVM_VERSION_0_30.equals(properties.evmVersion()), isUsableContract);
```
lukelee-sl, line 362: do we care about direct calls to system contract addresses?

lukelee-sl: so direct calls to 0x167 is not allowed?  I always forget whether we allow this or not.

stoqnkpL: it not allowed, only internal calls to hedera precompiles work + token redirect calls but we allow those with this fix

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
```diff
@@ -359,10 +359,10 @@ private Account extractAndValidateReceiver(
...
+                    validateTrue(isTokenAccount || !EVM_VERSION_0_34.equals(properties.evmVersion()), isUsableContract);
+                    validateTrue(isTokenAccount || !EVM_VERSION_0_38.equals(properties.evmVersion()), isUsableContract);
+                    validateTrue(isTokenAccount || properties.allowCallsToNonContractAccounts(), isUsableContract);
```
mhess-swl, line 365: Is the "misbehavior" mentioned in this PR's title the fact that the flag doesn't behave correctly in mono service, or that it doesn't behave correctly in mod service (or both)? I'm surprised to see changes to mono service code _and_ mod service in the same PR.

stoqnkpL: Initially we thought there was only an issue with the mono codebase. 
But then we opened https://github.com/hashgraph/hedera-services/pull/11245 which is based on this branch to see if this PR resolves the issue and we identified issues with mod codebase as well.

stoqnkpL (approved):
LGTM

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java
mhess-swl: Ok, thanks!

mhess-swl (approved):
Looks good for services

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
```diff
@@ -204,4 +208,15 @@ private boolean contractMustBePresent(
...
         return !featureFlags.isAllowCallsToNonContractAccountsEnabled(config, possiblyGrandFatheredEntityNumOf);
     }
+
+    private boolean emptyCodePossiblyAllowed(
+            @NonNull final Configuration config,
+            @NonNull final FeatureFlags featureFlags,
+            @NonNull final ContractID contractId,
+            @NonNull final HederaEvmTransaction transaction,
+            @NonNull final Code code) {
+        return !(contractMustBePresent(config, featureFlags, contractId) && code.equals(CodeV0.EMPTY_CODE))
```
tinker-michaelj, line 218: ⚠️ How we are going to keep complexity under control with compound conditionals like this being sprinkled in? 😵‍💫 

Shouldn't we have already enforced the contract existence check somewhere upstream in `TransactionProcessor#computeInvolvedParties()` (in any case, there is lots of code there related to contract-must-exist)?

tinker-michaelj (changes requested):
Let's revert the mod-service changes here, I want to make sure we keep each concern (e.g. contract-existence-checks) as isolated as possible in mod-service. 

I can't see how it's a `FrameBuilder` responsibility to be doing a check like the one added here.

stoqnkpL:
> Let's revert the mod-service changes here, I want to make sure we keep each concern (e.g. contract existence checks) as isolated as possible in mod-service.
> 
> I can't see how it's a `FrameBuilder` responsibility to be doing a check like the one added here.

Adding some context as to why there are mod-service changes here. Once we resolved the mono issues, we opened this [Draft PR](https://github.com/hashgraph/hedera-services/pull/11245) that we used to execute CI and make sure all tests are behaving as we expect now. Then @thenswan noticed there's a different behaviour between mono and mod codebase and @lukelee-sl helped us resolve that.
If we revert the mod-codebase change, wouldn't that break mono-mod equivalence? @lukelee-sl wdyt?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
lukelee-sl: I must admit that I'm responsible for this monstrosity as I needed this check in order to maintain equivalency with mono service as @stoqnkpL says.  But this code doesn't seem to fit well with `TransactionProcessor#computeInvolvedParties` either.  Maybe it'll be better to create an additional step like`TransactionProcessor#validateContract` after we compute the parties.  That way we can isolate all of the logic for whether we care about contract/account/contract code existence there as private methods or static methods in a utility class?

tinker-michaelj (approved):
LGTM, ty @thenswan !

PR #11249, opened 2024-01-30
test: 11136 - Create hapi tests for selfDestruct operation with system accounts as beneficiaries

### LINKED ISSUE #11136: System accounts disambiguation: Create hapi tests for selfDestruct operation for system accounts
### Problem

### Solution

Add tests for **selfDestruct** operation involving the system accounts and ensure that these tests conform to the following expected behavior.
image

### Alternatives

_No response_

0xivanov:
**Description**:

This PR adds tests for `selfDestruct` operation involving the system accounts and ensure that these tests conform to the following expected behavior:

image

**Related issue(s)**:

Fixes #11136

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
```diff
@@ -1516,16 +1523,81 @@ contract, BALANCE_OF, mirrorAddrWith(systemAccounts.get(i)))
...
+        final var opsArray = new HapiSpecOperation[nonExistingAccountsOps.length + existingAccountsOps.length];
+
+        System.arraycopy(nonExistingAccountsOps, 0, opsArray, 0, nonExistingAccountsOps.length);
+        System.arraycopy(existingAccountsOps, 0, opsArray, nonExistingAccountsOps.length, existingAccountsOps.length);
+
+        return defaultHapiSpec("testSelfDestructForSystemAccountsWithSigRequiredTrue")
+                .given(
+                        cryptoCreate(BENEFICIARY)
+                                .balance(ONE_HUNDRED_HBARS)
+                                .receiverSigRequired(true)
```
0xivanov, line 1583: Some of the ops should probably fail here, need to sync with @stoqnkpL about this, as I am not 100% sure.

0xivanov: Turns out all the system accounts that exist (800-999 from what i've seen) have `receiverSigRequired=false` so the transaction would always succeed.

tinker-michaelj (approved):
Isn't this actually testing whether system accounts can be used as _beneficiaries_ of `selfdestruct`?

Assuming that's the goal, LGTM!

0xivanov:
> Isn't this actually testing whether system accounts can be used as _beneficiaries_ of `selfdestruct`?
> 
> Assuming that's the goal, LGTM!
Yes, this is the goal :) . The PR title might be misleading, will modify it.

PR #11258, opened 2024-01-30
chore: stabilize a few specs

tinker-michaelj:
**Description**:
-  Stabilize `internalCallToExistingMirrorAddressResultsInSuccessfulCall()` by allowing for varying numbers of zero bytes in EVM payloads and hence different intrinsic gas costs.
- Stabilize `updateWithOutOfDateKeyFails()` by removing a pointless sleep-then-get-receipt pattern with the default poll-until-receipt-available behavior.
- Remove an explicit `snapshotModeOp()` from `htsTransferFromForNFTViaContractCreateLazyCreate()`.


Neeharika-Sompalli (approved):

PR #11259, opened 2024-01-30
feat: Make updates to dashboards, and add new ones

### LINKED ISSUE #11256: Update grafana dashboards
- Update grafana dashboards to include recently added metrics

alittley:
- closes #11256 
- *must be cherry picked into 46*

To review, please examine the following preprod dashboards:
- `Event Creation`
- `Gossip`
- `Intake Pipeline`
- `Threads and Queues`
- `Virtual Map`


edward-swirldslabs (dismissed):
I went to preproduction dashboards to create this feedback.   If this is not the right thing to do, please provide links to where I can see the visual product of your dashboards here. 

[Event Creation](https://preproduction.grafana.hedera-ops.com/d/c621a05a-1e2a-4685-925b-0c28c0880baa/event-creation?orgId=1): 
* This looks like a mixing of stats types.   Is the collection of 8 fraction metrics all partitions of the same thread utilization?  More text explaining  metrics and how to interpret them would be helpful. 

[Gossip](https://preproduction.grafana.hedera-ops.com/d/e857b70d-9a9f-485b-bd3d-a359c55c8efc/gossip?orgId=1&refresh=5s)
* There's an empty placeholder panel at the top, but the metrics below do not have a place holder panel.  Is there a rhyme or reason for when to use them and when not to? 
* The label on the y-axis says seconds, but I expect it to be Y for metrics that are Y/second.   Time is the x-axis

[Event Intake](https://preproduction.grafana.hedera-ops.com/d/c3475af5-4598-4279-a024-494eb4c4d66d/intake-pipeline?orgId=1)
* Some of the panels have an `(i)` symbol next to the title with slightly more description.  The text looks it might be coming from the code, but it is not on all the panels.  Should we be putting more thorough description in these information boxes and have them be on all panels?  (Explaining how to interpret the metrics and what is good or bad). 

[Threads and Queues](https://preproduction.grafana.hedera-ops.com/d/daf603b7-4578-4641-92b3-aa4f0f146876/threads-and-queues?orgId=1)
* A lot of panels with no data in them. 


[Virtual Map](https://preproduction.grafana.hedera-ops.com/d/d4dfeec8-23dc-48b9-931a-71e07a963ffc/virtualmap?orgId=1)
* all panels say no data.

alittley:
@edward-swirldslabs 

 > This looks like a mixing of stats types. Is the collection of 8 fraction metrics all partitions of the same thread utilization? More text explaining metrics and how to interpret them would be helpful.

I added some descriptions to the previously existing metrics, to hopefully make this more clear. As to mixed stat types - the idea is to put anything relevant to event creation here, which will involve various types.

> There's an empty placeholder panel at the top, but the metrics below do not have a place holder panel. Is there a rhyme or reason for when to use them and when not to?

Placeholder metrics exist to line the speedometers up with the line graph version of any given metrics. They are only there where necessary for the alignment. I personally didn't add any placeholders with these changes, they were already around.

> The label on the y-axis says seconds, but I expect it to be Y for metrics that are Y/second. Time is the x-axis

You're right, this was a copy/paste error on my part. I updated all of the y axis labels on this dashboard to be correct.

> Some of the panels have an (i) symbol next to the title with slightly more description

I added descriptions where they were lacking on this new dashboard. I agree that it makes sense to have some sort of description to provide orientation looking at these stats. 

> A lot of panels with no data in them.

This is a consequence of the fact that there isn't a separate grafana instance for each network. Rather, preprod networks all have the same instance, and production networks also share an instance. Therefore, the dashboards have to include all extant metrics, even though some are only applicable to particular versions. As our metrics stabilize, we will have to continue removing metrics from these dashboards once they age out of all networks.

cody-littley (approved):

edward-swirldslabs (approved):

beeradb (approved):
I also noted some of the missing data panels, but your explanation to @edward-swirldslabs is sufficient.

LGTM, Thanks!

Issue #11255, opened 2024-01-30
Discrepancy protobuf bucket bytes between Bucket and ParsedBucket if bucket index is zero


artemananiev:
This problem may be the cause of https://github.com/hashgraph/hedera-testing/issues/427.

There are a few issues in `Bucket` and `ParsedBucket` classes after MerkleDb was changed to support PBJ. One of the issues is critical:
* At node startup from a previously saved snapshot (in the old JDB format), HalfDiskHashMap buckets are deserialized from disk using `ParsedBucket` class
* When these `ParsedBucket` objects are then serialized (to the new PBJ format), they don't write bucket index field, if it's value is default (zero)
* Then such a bucket is loaded back, this time using `Bucket` class, as this time the format is PBJ both on disk and in config. `Bucket` implementation assumes the bucket index is always present in the serialized form, even if zero
* It may result in `getBucketIndex()` to return weird values, and in rare cases - when the bucket with index 0 is empty - may lead to the BufferUnderflowException in the issue linked above:

```
Caused by: java.nio.BufferUnderflowException
        at com.hedera.pbj.runtime.io.buffer.BufferedData.checkUnderflow(BufferedData.java:500) ~[pbj-runtime-0.7.14.jar:?]
        at com.hedera.pbj.runtime.io.buffer.BufferedData.getInt(BufferedData.java:420) ~[pbj-runtime-0.7.14.jar:?]
        at com.swirlds.merkledb.files.hashmap.Bucket.getBucketIndex(Bucket.java:172) ~[swirlds-merkledb-0.47.0-develop.xc7c823e.jar:?]
        at com.swirlds.merkledb.files.hashmap.HalfDiskHashMap.endWriting(HalfDiskHashMap.java:398) ~[swirlds-merkledb-0.47.0-develop.xc7c823e.jar:?]
        at com.swirlds.merkledb.MerkleDbDataSource.writeLeavesToPathToKeyValue(MerkleDbDataSource.java:1256) ~[swirlds-merkledb-0.47.0-develop.xc7c823e.jar:?]
        at com.swirlds.merkledb.MerkleDbDataSource.saveRecords(MerkleDbDataSource.java:547) ~[swirlds-merkledb-0.47.0-develop.xc7c823e.jar:?]
        ...
```

Other, less critical, issues in related code:
* In `Bucket`, `METADATA_SIZE` field uses a wrong wire type for `FIELD_BUCKET_INDEX` field. It should be fixed32, but now is varint
[...]

PR #11261, opened 2024-01-30
fix: 11255: Discrepancy protobuf bucket bytes between Bucket and ParsedBucket if bucket index is zero

artemananiev:
Fix summary:

* Changed `ParsedBucket` to always write bucket index field during serialization, so assumptions in `Bucket` (e.g. header size) are always met
* Added a new unit test
* Fixed a couple other minor issues

Fixes: https://github.com/hashgraph/hedera-services/issues/11255


imalygin (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/Bucket.java
```diff
@@ -241,8 +240,14 @@ public void putValue(final K key, final long value) {
                     bucketData.position(result.entryOffset());
                     bucketData.writeBytes(remainder);
                 }
+                if (bucketIndexFieldOffset > result.entryOffset()) {
+                    // It should not happen with default implementation, but if buckets are serialized
+                    // using 3rd-party tools, field order may be arbitrary, and "bucket index" field
+                    // may be after the deleted entry
+                    bucketIndexFieldOffset -= result.entrySize();
+                }
```
anthony-swirldslabs, line 248: Seems like this code is making a very particular assumption about where exactly that offset is located. I guess this should be fine as long as we only use a limited number of "third-party tools."

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/Bucket.java
artemananiev: No particular assumptions. There used to be an assumption that "bucket index" field was always written first, but with this change the assumption is gone

anthony-swirldslabs: Sounds good.

imalygin: I'll updated it in the subsequent PR

PR #11267, opened 2024-01-31
fix: platform should not use common fork join pool

### LINKED ISSUE #11265: Don't use common fork join pool for wiring
The wiring framework should not be using the common fork join pool.

cody-littley:
closes #11265

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
@@ -24,9 +24,9 @@ import com.swirlds.common.wiring.model.internal.StandardWiringModel;
 import com.swirlds.common.wiring.schedulers.builders.TaskSchedulerBuilder;
 import com.swirlds.common.wiring.schedulers.builders.TaskSchedulerMetricsBuilder;
 import com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType;
-import com.swirlds.metrics.api.Metrics;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.List;
+import java.util.concurrent.ForkJoinPool;
 
 /**
  * A wiring model is a collection of task schedulers and the wires connecting them. It can be used to analyze the wiring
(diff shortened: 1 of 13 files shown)


anthony-swirldslabs (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerBuilder.java
```diff
@@ -86,6 +91,7 @@ public TaskSchedulerBuilder(@NonNull final StandardWiringModel model, @NonNull f
             throw new IllegalArgumentException("TaskScheduler name must not be empty");
         }
         this.name = name;
+        this.pool = Objects.requireNonNull(defaultPool);
```
OlegMazurov, line 94: Should `ForkJoinPool.commonPool()` still be used to initialize `this.pool` above? Looks like it's always overwritten here.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
```diff
@@ -39,23 +39,15 @@ public interface WiringModel extends Startable, Stoppable {
...
-     * Build a new wiring model instance.
-     *
-     * @param metrics provides metrics
-     * @param time    provides wall clock time
-     * @return a new wiring model instance
-     */
-    @NonNull
-    static WiringModel create(@NonNull final Metrics metrics, @NonNull final Time time) {
-        return new StandardWiringModel(metrics, time);
+    static WiringModel create(
```
lpetrovic05, line 46: Overloading this method with the default pool seems like a better approach

lpetrovic05 (approved):

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
cody-littley: I'm not a fan of using overloaded constructors for optional arguments. Very quickly the constructors for a class grow in count and complexity. IMO, the best way of handling optional parameters is to use the builder model. 

If this is the path we decide to go down, I can open up a ticket and change this to a builder pattern as a stand alone PR. It's a small amount of work, probably a 1 or 2 point task, and I can get it done this week.

edward-swirldslabs (approved):

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
cody-littley: Marking this as resolved since it's minor, once we work out our plan I'm willing to make changes immediately as a follow up.

PR #11273, opened 2024-01-31
perf: Move token on disk

### LINKED ISSUE #11271: Tokens are created in memory
### Description

Creating tokens in memory creates an undesirable limitation for scaling the number of entities to hundreds of millions.
Affects testing, benchmarking. Will create unnecessary migration in the future if shipped as is. 

### Steps to reproduce

`InitialModServiceTokenSchema.statesToCreate()`:
```
                StateDefinition.inMemory(TOKENS_KEY, TokenID.PROTOBUF, Token.PROTOBUF),
```

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

netopyr:
Move schema definition of tokens on disk

Closes #11271

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
@@ -98,6 +98,7 @@ public class InitialModServiceTokenSchema extends Schema {
     // These need to be big so databases are created at right scale. If they are too small then the on disk hash map
     // buckets will be too full which results in very poor performance. Have chosen 10 billion as should give us
     // plenty of runway.
+    private static final long MAX_TOKENS = 10_000_000_000L;
     private static final long MAX_ACCOUNTS = 10_000_000_000L;
     private static final long MAX_TOKEN_RELS = 10_000_000_000L;
     private static final long MAX_MINTABLE_NFTS = 10_000_000_000L;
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -145,7 +146,7 @@ public InitialModServiceTokenSchema(
     @Override
     public Set<StateDefinition> statesToCreate() {
         return Set.of(
-                StateDefinition.inMemory(TOKENS_KEY, TokenID.PROTOBUF, Token.PROTOBUF),
+                StateDefinition.onDisk(TOKENS_KEY, TokenID.PROTOBUF, Token.PROTOBUF, MAX_TOKENS),
```
Neeharika-Sompalli, line 149: Only `tokenRels` are on disk and not tokens as far as I know from [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/ServicesState.java#L526)

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
Neeharika-Sompalli: Discussed as suggestion from Oleg.

mhess-swl (approved):

PR #11274, opened 2024-01-31
perf: Ensure TransactionStateLogger is not logging on info level

### LINKED ISSUE #11270: TransactionStateLogger is still too verbose
### Problem

TransactionStateLogger is set to log at `info`. At high TPS, logging on the critical transaction handling thread becomes a significant performance factor.

### Solution

Either switch to `warn` in logger configuration or reconsider the level of messages logged during transaction handling.

### Alternatives

_No response_

netopyr:
This PR changes the last logging statement on the info level to a debug statement. As the methods in `TransactionStateLogger` are called for each transaction, the info level is not appropriate.

Closes #11270


Neeharika-Sompalli (approved):
LGTM ! 🙂

Issue #11198, opened 2024-01-26
Differential testing: Create a store dumper for the "singleton" stores for modular representation


MiroslavGatsanoga:
The "singleton" stores dumper need to handle dumping from the modular representation.
Should be implemented in Services node and leverage the existing migration logic already implemented there.

vtronkov:
Merged https://github.com/hashgraph/hedera-services/pull/11275

PR #11275, opened 2024-01-31
fix: differential testing: add singleton stores mono and modular representation

### LINKED ISSUE #10999: Differential testing: Create a store dumper for the "singleton" stores, both mono- and modular- representations
New "singletons" store dumper needs to handle the mono-representation and migration + dumping of the modular representation.

See the DumpContractStoresSubcommand for how this can be done.

The dumper(s) needed here should match the design choice made for the singleton store migrator classes.

See https://www.notion.so/swirldslabs/Signed-State-Store-Dumpers-57d4412fa4c84459a0ca699ce1446c98 for further information.

Depends on https://github.com/hashgraph/hedera-services/issues/10997

petreze:
**Description**:

- Add cli commands for "singleton" stores dumping for mono.
- Adding dumpers for "singleton" stores for modular representation

**Related issue(s)**:

Fixes #10999 and #11198

**Notes for reviewer**:

- The changes from the `StateDumper` file were removed due to constant merge conflicts. All the changes will be committed at the end from a new branch called `StateDumper-final-changes`

- Entity id is combined with block info, as it contains only one value and it may be not appropriate to build separate report for it.

- [Here](https://github.com/hashgraph/hedera-services/pull/11275/files#diff-28afb026425ab4152195ef12202e5b396c85e1e4c44a549268d93e18a42adee0R68-R69) I'm not sure how to obtain the history hash for modular representation.


hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleStakingInfo.java
```diff
@@ -525,7 +525,7 @@ public MerkleStakingInfo(
 
     @Nullable
     @VisibleForTesting
-    byte[] getHistoryHash() {
+    public byte[] getHistoryHash() {
```
mhess-swl, line 528: Hmm, I'm not sure we want to expose this. But maybe it's not a big deal. @tinker-michaelj thoughts?

mhess-swl (approved):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleStakingInfo.java
david-bakin-sl: IMO it's ok as the method is (already) annotated with `@VisibleForTesting` which to me means in this case it is acceptable to use in `cli-clients` - utility tools for programmers.  Users (e.g., this dumper) are then on notice that it could change out from under them in the future and it's on them to deal with it, while in the meantime it's totally ok for a utility tool to use.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleNetworkContext.java
```diff
@@ -819,7 +819,7 @@ private String reprOf(final Instant consensusTime) {
         return consensusTime == null ? NOT_AVAILABLE : consensusTime.toString();
     }
 
-    private String stringifiedBlockHashes() {
+    public String stringifiedBlockHashes() {
```
david-bakin-sl, line 822: Please add the `@VisibleForTesting` annotation here (see comments at `MerkleStakingInfo.getHistoryHash()` below).

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/virtual/BlockInfo.java
```diff
@@ -0,0 +1,50 @@
+/*
```
david-bakin-sl, line 1: IntelliJ doesn't find any uses of this class?

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/signedstate/DumpBlockInfoSubcommand.java
```diff
@@ -0,0 +1,150 @@
...
+            long lastBlockNumber,
+            @NonNull String blockHashes,
+            @Nullable RichInstant consTimeOfLastHandledTxn,
+            boolean migrationRecordsStreamed,
+            @Nullable RichInstant firstConsTimeOfCurrentBlock) {
+        BlockInfo(@NonNull final MerkleNetworkContext networkContext) {
+            this(
+                    networkContext.getAlignmentBlockNo(),
+                    networkContext.stringifiedBlockHashes(),
+                    fromJava(networkContext.consensusTimeOfLastHandledTxn()),
```
david-bakin-sl, line 88: I had to look up `fromJava` ... wasn't obvious to me where it came from ... perhaps it shouldn't be a static import and instead fully qualify it `RichInstant.fromJava(...)`?

david-bakin-sl (approved):
LGTM.  (Sorry for the delay)

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/virtual/BlockInfo.java
petreze: It was WIP this week because the EIP155 support was prioritized for the 0.47 release

Neeharika-Sompalli (dismissed):
LGTM !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/bbm/singleton/StakingInfo.java
```diff
@@ -0,0 +1,73 @@
...
+                stakingInfo.stakeToNotReward(),
+                stakingInfo.stakeRewardStart(),
+                stakingInfo.unclaimedStakeRewardStart(),
+                stakingInfo.stake(),
+                stakingInfo.rewardSumHistory().stream()
+                        .mapToLong(Long::longValue)
+                        .toArray(),
+                stakingInfo.weight(),
+                // todo how to map history hash?
+                EMPTY_BYTES);
```
vtronkov, line 69: Anyone?

david-bakin-sl: I think it is not used.  I'll check further but for now let's remove it from the mono side.

david-bakin-sl (dismissed):
Geez, these things we've been calling "Singletons" are probably misnamed huh?  That name makes them seem simple/trivial (there's only one of each of them right?) but they turn out to be more work than any 3 of the other dumpers put together!  TY!

tinker-michaelj (approved):
Did not review in _detail_, but LGTM! tyvm @petreze 🥂

Issue #11101, opened 2024-01-23
Metadata HIPs - Implement changes for TokenUpdateHandler, and its tests


Neeharika-Sompalli:
Depends on #11065

3️⃣ **TokenUpdateHandler:**

- [x] Validate the metadata key that it can be set only if metadata key is already present on the token [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/util/TokenUpdateValidator.java#L62) similar to other keys
- [x] Also validate the `metadata` to be updated has size is less than `tokens.maxMetadataBytes`(if metadata field is present)
- [x] Set the updated fields on `Token` entity from the `TokenUpdateTransactionBody` when we [build](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java#L179) new token.

4️⃣ **Signing Requirements**:

- No changes

5️⃣  **Fees:**

- We can do all fees changes in different issue 😄 

6️⃣  **Tests**

- [x] Unit tests

- [x] E2E tests

  HapiTokenUpdate: 
  - [x] Add new fields similar to other fields and add setters to set them.
  - [x] Update `opBodyDef()` to use these fields to construct the `TokenUpdateTransactionBody`
  - [x] Update `updateStateOf()` to save the `metadataKey` to registry after token is created
  - [x] Update `TokenUpdateSpecs.creationHappyPath()` test to update the new fields on `TokenUpdate`  and validate using `TokenInfo`

PR #11279, opened 2024-01-31
fix: 11101: Metadata HIPs - Implement Changes for TokenUpdate Handling 

thomas-swirlds-labs:
**Description**:

- Validate the metadata key that it can be set only if metadata key is already present on the token similar to other keys
-  Also validate the metadata to be updated has size is less than tokens.maxMetadataBytes(if metadata field is present)
-  Set the updated fields on Token entity from the TokenUpdateTransactionBody when we build new token.

**Tests** 
- Add new fields and setters.
-  Update opBodyDef() to use these fields to construct the TokenUpdateTransactionBody
-  Update updateStateOf() to save the metadataKey to registry after token is created
- Update TokenUpdateSpecs.creationHappyPath() test to update the new fields on TokenUpdate and validate using TokenInfo

**Related issue(s)**:

Fixes #https://github.com/hashgraph/hedera-services/issues/11101 

**Notes for reviewer**:
We will handle all the fee changes in a seperate issue

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateValidator.java
```diff
@@ -62,6 +62,10 @@ public ValidationResult validateSemantics(
         if (op.hasMemo()) {
             context.attributeValidator().validateMemo(op.memo());
         }
+        // validate metadata
+        if (op.metadata() != null && !op.metadata().toString().isEmpty()) {
+            validator.validateTokenMetadata(String.valueOf(op.metadata()), tokensConfig);
+        }
```
mhess-swl, line 68: Is the `toString()` of metadata dependable, and should metadata always be interpretable as text? Based on the protobuf definition (`Metadata of the created token definition`), I would expect this to be something more like
```
// use `Arrays.isNullOrEmpty(byte[])` to check for an effectively empty byte array, 
// which - unlike toString() – won't use a potentially lossy transform of the original bytes 

if (op.metadata() != null && Arrays.isNullOrEmpty(op.getMetadata().toByteArray()) {
    // no potentially lossy conversion to a string
    validator.validateTokenMetadata(op.metadata(), tokensConfig);
}
```

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/pbj/PbjConverter.java
```diff
@@ -1303,10 +1303,10 @@ public static com.hederahashgraph.api.proto.java.ResponseCodeEnum fromPbj(@NonNu
                     .TRANSACTION_HAS_UNKNOWN_FIELDS;
             case ACCOUNT_IS_IMMUTABLE -> com.hederahashgraph.api.proto.java.ResponseCodeEnum.ACCOUNT_IS_IMMUTABLE;
             case ALIAS_ALREADY_ASSIGNED -> com.hederahashgraph.api.proto.java.ResponseCodeEnum.ALIAS_ALREADY_ASSIGNED;
-            case INVALID_METADATA_KEY -> com.hederahashgraph.api.proto.java.ResponseCodeEnum.INVALID_METADATA_KEY;
             case TOKEN_HAS_NO_METADATA_KEY -> com.hederahashgraph.api.proto.java.ResponseCodeEnum
                     .TOKEN_HAS_NO_METADATA_KEY;
             case MISSING_TOKEN_METADATA -> com.hederahashgraph.api.proto.java.ResponseCodeEnum.MISSING_TOKEN_METADATA;
+            case INVALID_METADATA_KEY -> com.hederahashgraph.api.proto.java.ResponseCodeEnum.INVALID_METADATA_KEY;
```
mhess-swl, line 1309: Nit: did this need to change position?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -301,7 +305,7 @@ private Token.Builder customizeToken(
         // If these keys did not exist on the token already, they can't be changed on update
         updateKeys(op, token, copyToken);
         updateExpiryFields(op, resolvedExpiry, copyToken);
-        updateNameSymbolMemoAndTreasury(op, copyToken, token);
+        updateNameSymbolMetadataMemoAndTreasury(op, copyToken, token);
```
mhess-swl, line 308: Ok, even though both the old and the new method names hurt my brain a little 😝

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -381,6 +388,10 @@ private void updateKeys(
             validateTrue(originalToken.hasPauseKey(), TOKEN_HAS_NO_PAUSE_KEY);
             builder.pauseKey(op.pauseKey());
         }
+        if (op.hasMetadataKey()) {
+            validateTrue(originalToken.hasMetadataKey(), TOKEN_HAS_NO_METADATA_KEY);
+            builder.metadataKey(op.metadataKey());
```
mhess-swl, line 393: `OrThrow()` here (to match `op.metadataOrThrow()`)? 
`builder.metadataKey(op.metadataKeyOrThrow())`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -73,6 +75,14 @@ public void validateTokenName(@Nullable final String name, @NonNull final Tokens
         tokenStringCheck(name, tokensConfig.maxTokenNameUtf8Bytes(), MISSING_TOKEN_NAME, TOKEN_NAME_TOO_LONG);
     }
 
+    /**
+     * Validates the token metadata, if it is exists and is not empty or not too long.
+     * @param metadata the token metadata to validate
+     */
+    public void validateTokenMetadata(@Nullable final String metadata, @NonNull final TokensConfig tokensConfig) {
```
mhess-swl, line 82: Related to a previous comment, is it ok in _all_ cases to convert the metadata bytes to a String?

Also for this class, would like to see unit tests for this method. Possible cases to exercise, with the following as metadata:
* a random string of bytes
* an empty byte array/Bytes.EMPTY
* a null value
* UTF-8 text, possibly including emojis :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenMetadataSpecs.java
```diff
@@ -0,0 +1,265 @@
...
+    public boolean canRunConcurrent() {
+        return true;
+    }
+
+    @Override
+    public List<HapiSpec> getSpecsInSuite() {
+        return List.of(
+                rejectsMetadataTooLong(),
+                creationRequiresAppropriateSigsHappyPath(),
+                creationRequiresAppropriateSigs(),
```
mhess-swl, line 93: This is more subjective and picky, but maybe name `creationRequiresAppropriateSigs()` to `creationDoesNotHaveRequiredSigs()`, in order to differentiate from `creationRequiresAppropriateSigsHappyPath()`? I had to check both methods to see what the difference was.

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerTest.java
```diff
@@ -221,6 +230,28 @@ void happyPathForNonFungibleTokenUpdate() {
...
+                .build();
+        assertThatNoException().isThrownBy(() -> subject.pureChecks(txn));
+        assertThat(txn.data().value()).toString().contains("test metadata");
+        assertThat(txn.data().value()).hasNoNullFieldsOrProperties();
+    }
+
+    @Test
+    void failsForInvalidMetaDataKey() {
+        setUpTxnContext();
+        txn = new TokenUpdateBuilder().withMetadataKey(Key.DEFAULT).build();
```
mhess-swl, line 248: Good call to test `Key.DEFAULT` here 👍

mhess-swl (commented):
Tyvm @thomas-swirlds-labs! A few details to address, but overall very solid first pass! 🙌

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateValidator.java
thomas-swirlds-labs: Good suggestion @mhess-swl ! 💯
Will modify accordingly

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
thomas-swirlds-labs: Removing metadataOrThrow() and orThrow() as well so we keep it consistent with the other keys / code

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
thomas-swirlds-labs: Good call. I've added all the test cases now including exercising byte arrays with null, randoms and empty values.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenMetadataSpecs.java
thomas-swirlds-labs: Done! 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateValidator.java
```diff
@@ -62,6 +63,10 @@ public ValidationResult validateSemantics(
         if (op.hasMemo()) {
             context.attributeValidator().validateMemo(op.memo());
         }
+        // validate metadata
+        if (op.metadata() != null && !Arrays.isNullOrEmpty(op.metadata().toByteArray())) {
```
Neeharika-Sompalli, line 67: I think we should just check null. If metadata is present and is empty we should set empty in my opinion

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -93,6 +103,25 @@ private void tokenStringCheck(
...
+    private void tokenByteArrayCheck(
+            @Nullable final byte[] bytes,
+            final int maxLen,
+            @NonNull final ResponseCodeEnum onMissing,
+            @NonNull final ResponseCodeEnum onTooLong) {
+        validateTrue(bytes != null, onMissing);
+        final int length = bytes.length;
+        validateTrue(length != 0, onMissing);
+        validateTrue(length <= maxLen, onTooLong);
+        validateTrue(!bytes.toString().contains("\u0000"), INVALID_ZERO_BYTE_IN_STRING);
```
mhess-swl, line 122: I would use something like [this](https://github.com/hashgraph/hedera-services/blob/a71e2fbe413bf9c97bfba0d6725f9f75bc2b8a45/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/validation/AttributeValidatorImpl.java#L138) instead

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
mhess-swl: Looks great, thanks!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
thomas-swirlds-labs: Sounds good. Done 👍

PR #11283, opened 2024-01-31
fix: don't verify signature transactions asyncronously

### LINKED ISSUE #11282: Don't verify transaction signatures asyncronously
There is unnecessary parallelism in transaction prehandling. It's currently calling `CryptoEngine.verifySync` and submitting a task to an executor service via the prefetchProcessor, but since verification is already in a thread pool, there is no need to spawn background tasks.

cody-littley:
closes #11282


cody-littley:
Decision was made not to modify mono code.

PR #11293, opened 2024-01-31
fix: rejected execution in wiring framework

### LINKED ISSUE #11290: Catch Rejected Execution Exceptions
In the wiring framework, we should not explode if the fork join pool throws a `RejectedExecutionException` from `ForkJoinPool.managedBlock()`.

cody-littley:
closes #11290

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
@@ -109,6 +109,7 @@ public class BackpressureObjectCounter extends ObjectCounter {
                 throw new IllegalStateException("Interrupted while blocking on an onRamp() for " + name);
             } catch (final RejectedExecutionException ex) {
                 // We've exhausted our supply of background threads, we have no choice but to busy wait.
+                Thread.onSpinWait();
             }
         }
     }
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
@@ -21,6 +21,7 @@ import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinPool.ManagedBlocker;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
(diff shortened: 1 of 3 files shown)


alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -72,38 +73,43 @@ public BackpressureObjectCounter(
      */
     @Override
     public void onRamp() {
-        final long currentCount = count.get();
-        if (currentCount < capacity) {
-            final boolean success = count.compareAndSet(currentCount, currentCount + 1);
-            if (success) {
-                return;
+        while (true) {
```
edward-swirldslabs, line 76: This is basically a spin until the condition is met and loop indefinitely until it is?  Does this burn 1 cpu while waiting?

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
edward-swirldslabs: nevermind, I think I misread where the scope ends.   The brain read the closing brace, but it's in a red zone, not a green zone.

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: It will burn a CPU either until the condition is met (i.e. there is capacity in the queue) or the fork join pool has sufficient resources to offload the blocked thread. This is intended as a last resort as opposed to just crashing when the exception is thrown. In practice, we shouldn't see this unless there is something hogging the fork join pool resources.

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/schedulers/SequentialTaskSchedulerTests.java
```diff
@@ -108,7 +108,7 @@ void orderOfOperationsTest(final String typeString) {
         assertEventuallyEquals(
                 value,
                 wireValue::get,
-                Duration.ofSeconds(1),
+                Duration.ofSeconds(10),
```
cody-littley, line 111: Note to reviewers:github test runner has been super flaky of late. Tests that used to work just fine are now having bizzare timing issues. I suspect the computer running tests is overloaded.

Increasing the timeouts here will allow more leniency when the host runner is under extreme load, but will add no additional time to the test if the runner is running at normal speed.

edward-swirldslabs: The timing issues are in no way related to this code change?

edward-swirldslabs (dismissed):

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/schedulers/SequentialTaskSchedulerTests.java
cody-littley: I can't rule that out. However, I'm unable to reproduce the failures locally. Also, I've seen lots of time sensistive tests get flaky recently.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -72,38 +73,43 @@ public BackpressureObjectCounter(
...
+                // The managedBlock() pattern allows us to block while yielding the physical CPU core to other
+                // tasks.
+                ForkJoinPool.managedBlock(onRampBlocker);
+                return;
+            } catch (final InterruptedException ex) {
+                // This should be impossible.
+                Thread.currentThread().interrupt();
+                throw new IllegalStateException("Interrupted while blocking on an onRamp() for " + name);
+            } catch (final RejectedExecutionException ex) {
+                // We've exhausted our supply of background threads, we have no choice but to busy wait.
```
lpetrovic05, line 111: Should we sleep a bit?

alittley (dismissed):

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: Possibly. @OlegMazurov what would you recommend that we do inside this loop?

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/wiring/schedulers/SequentialTaskSchedulerTests.java
cody-littley: Additional information I just found out recently (repeating it here for others, we discussed this privately on slack). There was a recent CI pipeline change that causes our tests to all run in parallel. These tests spin up lots of threads, and I'm thinking they are interfering with each other and fighting over CPU resources. IMO, increasing timeouts for stability is better than using the tag that forces the tests to run serially.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
OlegMazurov: Add `Thread.onSpinWait()`. Instantiating and throwing the exception is expensive enough to pass for a delay between unsuccessful attempts. Will be clearly visible in profiles (we'll reassess if and when that happens).
The real issue is that we use back-pressure inside a ForkJoinPool.

cody-littley: done

alittley (approved):

PR #11311, opened 2024-02-01
perf: Avoid excessive creation of Executors

### LINKED ISSUE #11299: Calling Executors.newVirtualThreadPerTaskExecutor() from the OnDiskReadableKVState constructor is too expensive
### Problem

The executor is only used to warm up keys. Previously, it was a static thread pool. The overhead of creating a new executor (even if virtual) every time `OnDiskReadableKVState` is instantiated is not justified.
In fact, the whole VirtualThread business is moot here.

### Solution

My stance is that we should avoid using VirtualThreads, especially in light of nested parallelism issues we now encounter when parallel execution explicitly coded in services interferes with the parallel framework in platform.

### Alternatives

_No response_

netopyr:
Replace usage of Executors.newVirtualThreadPerTaskExecutor() with simple static Runnable.

Closes #11299


OlegMazurov (approved):

Neeharika-Sompalli (approved):
LGTM !

PR #11318, opened 2024-02-01
feat: Address stores X509Certificate

### LINKED ISSUE #10837: Addresses store 509Certificate instead of PublicKey
requires changing the serialization and deserialization of Address. 

Does not require a HIP. 

Needed to satisfy the Roster API. 

At the same time we change the AB to use the Certificate, we also need to remove the encryption and agreement keys from the address book. That way we only have to do a migration once.

edward-swirldslabs:
**Description**:
As a pre-requisite step towards Dynamic Address Book Phase 3, this PR does the following: 
* removes the `encryption key` from the `Address`
* stores, serializes, and deserializes an `X509Certificate` instead of the `PublicKey`
* drops the `PublicKey` private variable and retrieves it from the certificate.  
* adds a `SerializableX509Certificate` class to replace `SerializablePublicKey`
* deprecates `SerializablePublicKey` for removal
* bumps address serialization class version to 6
* removes support for deserializing address versions prior to class version 5

**Related issue(s)**:

Fixes #10837 

**Notes for reviewer**:
* I made the crypto libraries a little more accessible.  Please verify I did not do anything verboten. 
  * I did this so that I could generate keys and certs dynamically in the random address book generator.  
* Some of the above permanent changes are pending the next software migration.   
* I  added `@MockitoSettings(strictness = Strictness.LENIENT)` to the Hedera life cycle unit test because Mockito claimed that the stubbed methods were not used, but if I don't include the stubs I get an exception that a crypto key cannot be rendered. 
  * It would be cleaner if I could just depend on the test artifacts I create at the platform level for pre-generated keys, but I could not import the class and I didn't want to mess with dependencies for Hedera.  
 
**Checklist**


platform-sdk/swirlds-common/build.gradle.kts
```diff
@@ -32,6 +32,7 @@ testModuleInfo {
     requires("com.swirlds.base.test.fixtures")
     requires("com.swirlds.config.api.test.fixtures")
     requires("com.swirlds.config.extensions")
+    requires("com.swirlds.platform.core")
```
lpetrovic05, line 35: common requires core? that does not sound right...

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/SerializableX509Certificate.java
```diff
@@ -0,0 +1,105 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.security.PublicKey;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Objects;
+
+public class SerializableX509Certificate implements SelfSerializable {
+    public static final int MAX_CERT_LENGTH = 65_536;
```
lpetrovic05, line 32: why this number?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/SerializableX509Certificate.java
```diff
@@ -0,0 +1,105 @@
...
+import com.swirlds.common.io.streams.SerializableDataOutputStream;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.security.PublicKey;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Objects;
+
```
lpetrovic05, line 30: no javadocs

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedX509Certs.java
```diff
@@ -0,0 +1,398 @@
...
+    private static final Map<NodeId, SerializableX509Certificate> agreeCerts = new HashMap<>();
+
+    /**
+     * Utility class
+     */
+    private PreGeneratedX509Certs() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    //    This method is commented out because it creates a dependency to swirlds-platform-core
```
lpetrovic05, line 50: i think we need to find a better solution for this

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedX509Certs.java
```diff
@@ -0,0 +1,398 @@
...
+    //    Uncomment this method to generate more certs.
+    //    /**
+    //     * Generates a set of X.509 certificates for testing purposes.
+    //     *
+    //     * @param numCerts the number of certificates to generate
+    //     * @param random   the random number generator to use
+    //     */
+    //    public static void generateCerts(final int numCerts, @NonNull final Random random) {
+    //        final Path sigCertPath = Path.of(SIG_CERT_FILE);
```
lpetrovic05, line 59: the resource loader should be used for this

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedX509Certs.java
```diff
@@ -0,0 +1,398 @@
...
+            }
+        } catch (final Exception e) {
+            throw new IllegalStateException("critical failure in loading certificates", e);
+        }
+        if (sigCerts.isEmpty() || agreeCerts.isEmpty()) {
+            throw new IllegalStateException("critical failure in loading certificates, no certs found.");
+        }
+    }
+
+    private static final List<String> sigCertsHexEncoded = List.of(
```
lpetrovic05, line 195: I think the encoded certs should be in a separate file

platform-sdk/swirlds-common/build.gradle.kts
edward-swirldslabs: In the Tests and TestFixtures it does.   The `RandomAddressBookGenerator` [https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/RandomAddressBookGenerator.java](link) lives in the common `TestFixtures` and that has references to `Address` and `AddressBook` in platform-core. 

As long as Common Main does not have a dependency Platform Core, I can understand the value of Common Test having a dependency on modules that rely on Common Main.  It makes it easier to use the code in testing that relies on the code being tested.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/SerializableX509Certificate.java
edward-swirldslabs: According to this [link](https://groups.google.com/g/mailing.openssl.users/c/awvGfcOwfZM) there is no max upper bound on the size of encoded certs since they may encode chains of certs.  For a self-signed cert, I was still uncertain.  The EC TLS Agreement key is signed by the RSA cert.  Given that the name and other fields in the cert are variable length, I wanted to give enough room that we would be safe to deserialize in.  I am treating the encoded cert as a blob of bytes.   

Since the encoding is DER/ASN.1, we could parse the beast and not rely on a WAG (wild ass guess) for its max size. 

Googling it looks like 4K might be enough for a single PEM file, [in the average case.](https://serverfault.com/questions/793563/what-is-the-common-size-of-ssl-certificate-give-or-take)  

I'm ignorant.  I was picking a safe value.  Would love for someone to give a more reasonable tighter bound.

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedX509Certs.java
edward-swirldslabs: There is precedent for doing this: [PreGeneratedPublicKeys](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedPublicKeys.java)

I originally had the keys in a separate file in the `swirld-common/src/testFixtures/resources` directory.   The problem came in when using `RandomAddressBookGenerator` from other modules.  I can't use an absolute PATH to the files to read them because other people may have them in different locations on their machines.  I also cannot use a relative PATH because the working directly in which people start the unit tests may be different.  

I agree having them in a separate file would be better, but I am ignorant of how to make that work so that it is insensitive to the current working directory that people initiate unit tests from.

edward-swirldslabs: I will see if I can get the class loader resources to work despite the directory of execution.

edward-swirldslabs: So far my efforts have failed to load the files in the resources directory.  The `PreGeneratedX509Certs` file is in `swirlds-common`, but the unit tests that need to have it load the certs are running in `swirlds-platform-core`.  When performing `PreGeneratedX509Certs.class.getClassLoader()`, null is returned.   I believe `Thread.currentThread.getContextClassLoader()` works, but its package binding is not going to be to `swirlds-common`, but whatever project's unit tests are executing.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/SerializableX509Certificate.java
lpetrovic05: can you add a comment explaining the reasoning

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/crypto/PreGeneratedX509Certs.java
lpetrovic05: our module structure is still quite messy, it seems that it needs some cleanup in order to accommodate this

jeromy-cannon (dismissed):

cody-littley (approved):

mhess-swl (approved):
Looks good for services

jeromy-cannon (dismissed):

mhess-swl (dismissed):
Good for services 👍

cody-littley (approved):

jeromy-cannon (dismissed):

mxtartaglia-sl (dismissed):

hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/HederaLifecyclesImplTest.java
```diff
@@ -44,8 +44,11 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
tinker-michaelj, line 51: How much work would it be to avoid needing this? Ime lenient mocking can hide real problems.

edward-swirldslabs: I was able to remove the mock public key and the stubbing that is being complained about in the `HederaLifecyclesTest.java` unit tests without problem and can remove the lenient marking in this file. 

When I do that for the `ServicesStateTest.java` I get an exception thrown on this [line of code](https://github.com/hashgraph/hedera-services/blob/c2ac301a51281361f8ac9517b19dc680e768d264/hedera-node/hedera-mono-service/src/test/java/com/hedera/node/app/service/mono/ServicesStateTest.java#L944).   Somehow this code is requesting the `PublicKey` and is generating a null pointer exception.  If I mock the method retrieving the public key, the exception is not thrown, but mockito complains about an unused mock.  I am left scratching my head. 

One unit test accesses the code path of getting the public key while 3 do not.  This is a damned if you do and damned if you don't situation.  

This i my ugly solution, to add a boolean for when the method needs to be stubeed: 
```
 public static AddressBook createPretendBookFrom(final Platform platform, final boolean withKeyDetails, boolean mockPublicKey) {
        final var publicKey = mock(PublicKey.class);
        final var cert = mock(X509Certificate.class);
        if(mockPublicKey){
            Mockito.when(cert.getPublicKey()).thenReturn(publicKey);
        }
```

tinker-michaelj: Hmm interesting 😵‍💫 ...so not easy.

tinker-michaelj (approved):
LGTM, tyvm @edward-swirldslabs !

jeromy-cannon (approved):

mxtartaglia-sl (approved):
LGTM

mhess-swl (approved):
LGTM for services

alittley (approved):

iwsimon (approved):
LGTM

lpetrovic05 (approved):

tinker-michaelj (approved):
LGTM, ty @edward-swirldslabs !

PR #11323, opened 2024-02-02
feat: address cold read issue in ExtCodeHash operation

lukelee-sl:
**Description**:
Address the issue of cold read potentially costing no gas by checking for sufficient gas at the beginning of the `execute()` method.

Fixes: #11332 

**Checklist**


lukelee-sl:
I tried to think of a way to test this code change as a bdd test but really could not as the behavior from an external perspective would be the same pre and post change.

david-bakin-sl (approved):
LGTM!

(Except for Github's code `diff`er.  That suboptimal.)

PR #11346, opened 2024-02-03
docs: update throttles section of config readme

ElijahLynn:
**Description**:
* Tried to capture some useful info in the config readme that @rbair23 shared with me about how throttle overrides work. Probably isn't 100% right so please provide feedback to make it more accurate

### DIFF
--- a/hedera-node/configuration/README.md
+++ b/hedera-node/configuration/README.md
@@ -43,7 +43,7 @@ path is given, the file belongs in the top-level directory with the JAR.
   - For example, more permissive Netty HTTP/2 settings
 
 ## upgrade/throttles.json
-* If present, will be used, post-upgrade, as input to a yahcli `sysfiles upload throttles` command. It will not layer on top of other throttles, it will be exactly what is in this file. If not present it will use https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/resources/throttles.json.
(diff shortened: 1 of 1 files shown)


ElijahLynn:
Hrm, this was just a readme.md change and we had a ton of checks run that did not provide any value. I think we need some improvements here. 

image

Nana-EC (commented):
LG

hedera-node/configuration/README.md
```diff
@@ -43,7 +43,7 @@ path is given, the file belongs in the top-level directory with the JAR.
   - For example, more permissive Netty HTTP/2 settings
 
 ## upgrade/throttles.json
-* If present, should be used, post-upgrade, as input to a yahcli `sysfiles upload throttles` command
+* If present, will be used, post-upgrade, as input to a yahcli `sysfiles upload throttles` command. It will not layer on top of other throttles, it will be exactly what is in this file. If not present it will use https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/resources/throttles.json.
```
Neeharika-Sompalli, line 46: Default file is under `hedera-file-service-impl/src/main/resources/genesis` for mod service

ElijahLynn, line 46:
```suggestion
* If specified, this file serves as the sole input for the `yahcli sysfiles upload throttles` command post-upgrade, replacing any existing throttles with its contents. In its absence, the default is sourced from `:/hedera-node/hedera-file-service-impl/src/main/resources/genesis/throttles.json`.
```

netopyr (approved):

ElijahLynn:
This is failing on unrelated things. How to re-trigger an entire run of all these tests?

PR #11352, opened 2024-02-05
chore: remove hashgraph demo

### LINKED ISSUE #11351: Remove HashgraphDemo
HashgraphDemo uses event fields which should only be visible to consensus and are not thread safe. Since it has purpose, it is better to remove it than to find a way to support it. Additionally, all of its functionality is already in the browser UI.

lpetrovic05:
closes  #11351


mxtartaglia-sl (approved):

JeffreyDallas (approved):

cody-littley (approved):

jsync-swirlds (approved):

netopyr (approved):

jasperpotts (commented):
Looks good

PR #11359, opened 2024-02-05
test: Create hapi tests for transfer and send operations for system accounts Part 1

### LINKED ISSUE #11138: System accounts disambiguation: Create hapi tests for transfer and send operations for system accounts
### Problem

### Solution

Add tests for **transfer** and **send** operations involving the system accounts and ensure that these tests conform to the following expected behavior.
image

### Alternatives

_No response_

0xivanov:
**Description**:

Add tests for transfer and send operations involving the system accounts and ensure that these tests conform to the expected behavior.

**Related issue(s)**:

Fixes #11138

**Notes for reviewer**:

**Checklist**


stoqnkpL (approved):

stoqnkpL (commented):
Is there a reason `.bitLength()` was used instead of something like `.size()`? 
Are you sure a bitLength of 4 is enough to cover 4 bytes?

0xivanov:
> Is there a reason `.bitLength()` was used instead of something like `.size()`? Are you sure a bitLength of 4 is enough to cover 4 bytes?

You are right, `bitLength` is the bits, so it should be 32 bits or `.size() = 4` which is in bytes.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/ExchangeRateSystemContract.java
```diff
@@ -57,6 +59,7 @@ public FullResult computeFully(@NonNull Bytes input, @NonNull MessageFrame messa
         requireNonNull(input);
         requireNonNull(messageFrame);
         try {
+            validateTrue(input.bitLength() >= 4, INVALID_TRANSACTION_BODY);
```
0xivanov, line 62: This validates the bit length, we need to check if the bytes length >= 4

tinker-michaelj (changes requested):
We should use the `PrecompileContractResult` API to communicate results from our `PrecompileContract`s to our `MessageCallProcessor`; throwing an `InvalidTransactionException` increases the API surface with no benefit.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/ExchangeRatePrecompiledContract.java
```diff
@@ -82,6 +85,7 @@ public long gasRequirement(Bytes bytes) {
     @Override
     public Bytes compute(final Bytes input, final MessageFrame frame) {
         try {
+            validateTrue(input.size() >= 4, INVALID_TRANSACTION_BODY);
```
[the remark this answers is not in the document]
0xivanov: Hey @tinker-michaelj 
In all 3 precompiled contracts in the mono, there are catch blocks that will catch `InvalidTransactionException` and ultimately return `PrecompileContractResult `.
For reference :
`HtsPrecompiledContract` line 240
`PrngSystemPrecompiledContract` line 144
for `ExchangeRatePrecompiledContract` we go into `PrecompiledContract` default logic which still returns `PrecompiledContract.PrecompileContractResult.halt`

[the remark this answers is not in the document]
0xivanov: Not related to the issue but we only do `frame.setExceptionalHaltReason(Optional.of(haltReason));`
in `HtsPrecompiledContract` and not the other 2. Is there a reason for that?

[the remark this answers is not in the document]
tinker-michaelj: Hmm you're right @0xivanov ... I definitely prefer not to introduce `validateTrue()` anywhere in mod-service system contracts; but we already have to refactor everything there, so it can be cleaned up later.

I suspect only `HtsPrecompiledContract` sets a halt reason explicitly because the other two precompiles will return `null` if they fail; and then the inherited `computePrecompile()` implementation [here](https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/precompile/PrecompiledContract.java#L61) sets the halt reason automatically.

tinker-michaelj (approved):
LGTM, ty @0xivanov !

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java
```diff
@@ -313,7 +314,7 @@ void prepareFields(final MessageFrame frame) {
     void prepareComputation(Bytes input, final UnaryOperator<byte[]> aliasResolver) {
         this.precompile = null;
         this.transactionBody = null;
-
+        validateTrue(input.size() >= 4, INVALID_TRANSACTION_BODY);
```
lukelee-sl, line 317: Adding on to @tinker-michaelj comments above, can we not simple return `null` to accomplish the same results?

lukelee-sl (commented):
General question about the checks for 4 bytes that you added in the system contracts - were they necessary to implement the main point of this pr - that is to show that transfers to system contracts would fail.  Or were they added on as an additional check because it was something you noticed while doing the primary task?

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/HTSPrecompiledContract.java
0xivanov: You are right. Are we talking only about the `HtsPrecompliledContract` here?

lukelee-sl: Yes, just in the `HtsPrecompliledContract`.  I did have a question that I posted below which is were these new checks necessary to show that transfers to system contracts would fail?

lukelee-sl (dismissed):
LGTM

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (dismissed):
LGTM, ty @0xivanov !

0xivanov:
All CI checks should be fixed now.

tinker-michaelj (approved):
LGTM, ty @0xivanov !

PR #11361, opened 2024-02-05
chore: add `-l` option to diff limited interval sizes

### LINKED ISSUE #11333: Update record stream matching to compare 5 minutes at a time
### Problem

For long time periods, we might get OOM if we read the entire contents of the stream directories with `RcDiff`.

### Solution

Instead of comparing full record directories, break the full interval into 5 minute periods and compares those one by one.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #11333
 - Add `-l`/`--len-of-diff-secs` option to `RcDiff` utility to diff smaller intervals and avoid OOM.
 
Fixed: #11372


iwsimon (dismissed):
LGTM, some failed unit tests.

PR #11381, opened 2024-02-05
chore: migrate per node pending rewards

### LINKED ISSUE #11358: (mod) Migrate per-node pending rewards
### Problem

Only total pending rewards are being migrated, although mod-service tracks per-node pending rewards.

### Solution

Migrate the amount of staking rewards pending for each node.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #11358 
 - Avoid an NPE in `FileCreateHandler` when shard or realm are not set explicitly.
 - Compute per-node pending rewards in `InitialModServiceTokenSchema` BBM migration code by summing the pending rewards for each non-deleted account staked to each node id (that doesn't have `declineReward=true`).


Neeharika-Sompalli (dismissed):
LGTM !

PR #11385, opened 2024-02-06
feat: Enhance storage (file) store dumper to handle modular representation

vtronkov:
**Description**:
* Adding modular store dumper for the files

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/10920


petreze (approved):

david-bakin-sl:
Hmm.  The dumps from the mono code here definitely won't match the dumps from the modular code here because in the mono code file contents and file metadata are in two separate tree nodes and are merged for the dump while in modular the contents + metadata are in one tree node and no merge needed.

We already _have_ a mono dumper that works - as a separate tool.  We have to think about whether the flow we're looking for requires the mono dumper to be _here_ in the node (it might very well) or if we can run the separate cli tool that exists.

Either the _full_ mono code needs to be here - see `DumpFilesSubcommand.collectFiles` where first it gets files + metadata nodes into separate hashmaps (L435-473) and then does the merge (L480-490) - _or_ I'd prefer that mono code just write a one line file saying "this is invalid, you need to run the cli tool!".

Let's discuss this in (internal) Slack.  I'll start the conversation there, then we can update the issue.

(I'll review the new modular code shortly.)

david-bakin-sl (changes requested):
Modular traversal looks fine.  Mono traversal needs work - see previous comment please!

david-bakin-sl (approved):
TY for that mono traversal fixup.  LGTM

tinker-michaelj (approved):
This looks great to me, ty @vtronkov !

PR #11397, opened 2024-02-06
chore: Update PBJ dependency

netopyr:
This PR updates the PBJ dependency to the latest version because it includes several performance fixes. In addition, it removes a workaround for a bug in PBJ that was fixed.


tinker-michaelj (approved):
LGTM, ty @netopyr !

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

jeromy-cannon (approved):

PR #11398, opened 2024-02-06
feat: Implement task scheduler squelching

### LINKED ISSUE #11140: Add squelching feature
- Squelch is a feature to be added to `TaskScheduler`s
- "squelching" means setting a flag such that all currently enqueued tasks and any newly enqueued tasks are dumped into the void
- when a component is put into "squelch" mode, it remains in that state until it is commanded to stop squelching
- the purpose of squelching is to put the system into a state where cyclic data flows are broken, and all outstanding tasks can be flushed without causing additional data to start flowing
- squelching will be disabled on schedulers by default, since it incurs the additional cost of checking a "squelch" flag before handling each task

Existing framework components that will use squelching:
- `OrphanBuffer`
- `LinkedEventIntake`
- `ConsensusRoundHandler`

alittley:
- closes #11140


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/ConcurrentTask.java
```diff
@@ -63,6 +67,9 @@ class ConcurrentTask extends AbstractTask {
     @Override
     protected boolean exec() {
         try {
+            if (squelcher.shouldSquelch()) {
```
lpetrovic05, line 70: I think @cody-littley was mentioning that we would want to wrap the handler with the squelcher. That way, if there is no squelching, it would call the handler directly and there would be no overhead.

alittley: @lpetrovic05 Do you mean wrap the handler into a lambda that manages squelching? I avoided that solution, since it would add a lot of overhead creating a bunch of lambdas.

The solution I went with doesn't add any meaningful overhead if squelching is disabled for a scheduler. Schedulers with no squelching ability will have a `NoopSquelcher`, which automatically returns `false` for `shouldSquelch`.

cody-littley: There is a way to do it that doesn't require a large number of lambdas. If you wrap the lambda in `BindableInputWire`, then you will only create two lambdas per input wire at start up time.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/ConcurrentTaskScheduler.java
```diff
@@ -74,18 +76,27 @@ public ConcurrentTaskScheduler(
...
     /**
      * {@inheritDoc}
      */
     @Override
     protected boolean offer(@NonNull final Consumer<Object> handler, @NonNull final Object data) {
-        boolean accepted = onRamp.attemptOnRamp();
+        if (getSquelcher().shouldSquelch()) {
+            // say that the data was accepted, even though it will be thrown into the void
+            return true;
+        }
```
cody-littley, line 95: At first glance this seemed strange, but when I thought about it I think it makes sense. We return true because we want to accept and discard the data, as opposed to preventing the sender from sending it. Might be nice to expand this comment a little to explain.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/ConcurrentTaskScheduler.java
```diff
@@ -74,18 +76,27 @@ public ConcurrentTaskScheduler(
...
+
+        final boolean accepted = onRamp.attemptOnRamp();
         if (accepted) {
-            new ConcurrentTask(pool, offRamp, uncaughtExceptionHandler, handler, data).send();
+            new ConcurrentTask(pool, offRamp, uncaughtExceptionHandler, handler, data, getSquelcher()).send();
```
cody-littley, line 99: If you move squelching to happen in a lambda in bindable input wire, then the tasks will not need to be aware of the squelcher.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/SequentialTask.java
```diff
@@ -36,6 +36,20 @@ class SequentialTask extends AbstractTask {
...
     private final UncaughtExceptionHandler uncaughtExceptionHandler;
 
+    /**
+     * If true, then this task will be executed even if scheduler is currently squelching.
+     * <p>
+     * This flag is necessary for flushing, since we want the flush task to execute even if the task scheduler is
+     * squelching normal handling. It isn't known at the time of construction whether a given task will be responsible
+     * for flushing, so we need a dynamic flag that can be set when the flush work is submitted.
+     */
+    private boolean overrideSquelch = false;
```
cody-littley, line 46: Something you can consider in order to avoid this ugly override is to change sequential tasks to flush in the same way as concurrent schedulers, i.e. by calling `ObjectCounter.waitUntilEmpty()`. It's less efficient, but we don't need to be super efficient when flushing since we only flush at startup and reconnect.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/LinkedEventIntakeWiring.java
```diff
@@ -32,20 +31,22 @@
...
 public record LinkedEventIntakeWiring(
         @NonNull InputWire<EventImpl> eventInput,
-        @NonNull InputWire<Boolean> pauseInput,
         @NonNull OutputWire<ConsensusRound> consensusRoundOutput,
         @NonNull OutputWire<List<EventImpl>> consensusEventsOutput,
         @NonNull StandardOutputWire<Long> keystoneEventSequenceNumberOutput,
-        @NonNull Runnable flushRunnable) {
+        @NonNull Runnable flushRunnable,
+        @NonNull Runnable startSquelchingRunnable,
+        @NonNull Runnable stopSquelchingRunnable) {
```
cody-littley, line 49: Not something to be solved in this PR, but I'm really disliking that we have to bind a bunch of lambdas in a record like this. IMO, when an object becomes complex like this, the right thing to do is to make it into a class. But this is something that probably requires discussion as a larger group.

cody-littley (commented):
Possible follow up task(s):

- add a `WiringModle.squelch()` method that automatically squelches and flushes all schedulers with squelching enabled
- add validation logic to the wiring model that checks for cycles in a squelched system
- once prior steps are completed, we can create fully automated flush that flushes all flushable schedulers in topological order

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/LinkedEventIntakeWiring.java
alittley: I agree. We should discuss as a group and come up with a plan

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/ConcurrentTask.java
alittley: Reworked

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/ConcurrentTaskScheduler.java
alittley: Done

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/SequentialTask.java
alittley: Done, for both sequential and sequential thread schedulers

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/internal/SequentialTaskScheduler.java
```diff
@@ -83,16 +82,16 @@ public SequentialTaskScheduler(
...
 
     /**
      * {@inheritDoc}
      */
     @Override
     protected void put(@NonNull final Consumer<Object> handler, @NonNull final Object data) {
-        if (getSquelcher().shouldSquelch()) {
+        if (currentlySquelching()) {
             return;
         }
```
cody-littley, line 96: For this scheduler (and for the other types as well), I'd suggest we drop the squelch checks for put/offer/inject, and insert into the queue regardless of squelching status. It causes us to do extra work while in squelching mode, but it saves us from having to check the atomic boolean once per object during normal operation.

alittley: Done

cody-littley (approved):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
```diff
@@ -71,12 +81,20 @@ protected TaskScheduler(
...
 
         this.model = Objects.requireNonNull(model);
         this.name = Objects.requireNonNull(name);
         this.type = Objects.requireNonNull(type);
         this.flushEnabled = flushEnabled;
+
+        if (squelchingEnabled) {
+            this.squelcher = new DefaultSquelcher();
+        } else {
+            this.squelcher = new NoopSquelcher();
```
lpetrovic05, line 95: Im wondering if maybe we should throw an exception instead of a noop. if you try to squelch something that wont do it, this would mean we silently ignore it

lpetrovic05: i now see that it does throw. maybe we rename it to something like `ThrowingSquelcher`

lpetrovic05 (approved):

cody-littley (approved):

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
alittley: Done

PR #11412, opened 2024-02-07
perf: Warm tokens as they are also on-disk now

### LINKED ISSUE #11410: Load tokens during warmup
Now that we have tokens on-disk, we need to add them to warm-ups as well (didn't make sense previously so no relevant code existed in mono).

```
        for (final var transfers : op.tokenTransfersOrElse(emptyList())) {
            final var tokenMeta = tokenStore.getTokenMeta(transfers.tokenOrElse(TokenID.DEFAULT)); 
```

netopyr:
This PR adds the warming of tokens to `CryptoTransfer` as tokens are also on-disk now.

Closes #11410


Neeharika-Sompalli (approved):
LGTM !

PR #11423, opened 2024-02-07
fix: Fix generated metric names

### LINKED ISSUE #11417: Fix metrics name
Some of our metric names are assembled during runtime and were missed during the name cleanup and need to be fixed.

netopyr:
This PR fixes metric names that are generated.

Closes #11417

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
@@ -167,7 +167,7 @@ public abstract class MethodBase implements ServerCalls.UnaryMethod<BufferedData
             @NonNull final Metrics metrics,
             @NonNull final String nameTemplate,
             @NonNull final String descriptionTemplate) {
-        final var baseName = serviceName + "/" + methodName;
+        final var baseName = serviceName + ":" + methodName;
         final var name = String.format(nameTemplate, baseName);
         final var desc = String.format(descriptionTemplate, baseName);
         return metrics.getOrCreate(new Counter.Config("app", name).withDescription(desc));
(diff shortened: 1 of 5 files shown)


iwsimon (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
```diff
@@ -185,7 +185,7 @@ public void invoke(
             @NonNull final Metrics metrics,
             @NonNull final String nameTemplate,
             @NonNull final String descriptionTemplate) {
-        final var baseName = serviceName + "/" + methodName;
+        final var baseName = serviceName + ":" + methodName;
```
Neeharika-Sompalli, line 188: Why do we need this ? Because only "/" is not an acceptable character ?

Neeharika-Sompalli, line 188: Also don't we need to change lines 
```
private static final String SPEEDOMETER_HANDLED_NAME_TPL = "%sHdl/sec";
private static final String SPEEDOMETER_RECEIVED_NAME_TPL = "%sRcv/sec";
```

netopyr: Yes. Prometheus is very strict and allows only alphanumeric characters plus ':' and '_'.

netopyr: I think you are looking at outdated code. This has already been fixed recently.

Neeharika-Sompalli: Ah yes! thanks

Neeharika-Sompalli (approved):

PR #11429, opened 2024-02-07
feat: Check platform status before syncing

### LINKED ISSUE #11418: Incorrect platform status should prevent sync
## Problem

- In the current system, platform status doesn't have an affect on whether a node syncs
    - Instead, we rely on gossip being started or stopped
- This is unsafe, as we could end up syncing at the wrong time 

## Solution

- Introduce logic that prevents a node from syncing if the platform status isn't in the permissible list

- Eventually, it would be good if we could always have gossip "started", and control whether or not we are syncing via platform status only
    - This would allow us to discard the racy gossip start/stop logic

alittley:
- this PR is the short term solution mentioned in #11418

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncProtocol.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncProtocol.java
@@ -18,7 +18,6 @@ package com.swirlds.platform.gossip.sync.protocol;
 
 import static com.swirlds.common.utility.CompareTo.isGreaterThanOrEqualTo;
 
-import com.swirlds.base.time.Time;
 import com.swirlds.common.context.PlatformContext;
 import com.swirlds.common.platform.NodeId;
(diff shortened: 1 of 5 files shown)


edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncProtocol.java
```diff
@@ -46,6 +49,17 @@
  * This object will be instantiated once per peer, and is bidirectional
  */
 public class SyncProtocol implements Protocol {
+    /**
+     * The platform statuses that permit syncing. If the platform isn't in one of these statuses, no syncs will be
+     * initiated or accepted
+     */
+    public static final Collection<PlatformStatus> STATUSES_THAT_PERMIT_SYNC = Set.of(
```
lpetrovic05, line 56: Can we have a separate class that encapsulates this logic? That way, the sync protocol only does `shouldSync()`

cody-littley (approved):

edward-swirldslabs (approved):

PR #11435, opened 2024-02-07
fix: 11298: VirtualMapReconnectTest fails intermittently with path not in range log message

artemananiev:
Another MerkleTestUtils method is changed to propagate underlying exceptions properly.


imalygin (dismissed):

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
```diff
@@ -1054,8 +1056,13 @@ public static <T extends MerkleNode> T testSynchronization(
                         reconnectConfig);
             }
 
+            final AtomicReference<Throwable> firstReconnectException = new AtomicReference<>();
+            final Function<Throwable, Boolean> exceptionListener = t -> {
+                firstReconnectException.compareAndSet(null, t);
```
anthony-swirldslabs, line 1061: It may be tricky if the first exception doesn't reveal the actual root cause, but a subsequent one could. It would be best to collect and report all of them.

anthony-swirldslabs (dismissed):

anthony-swirldslabs (approved):

imalygin (approved):

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
artemananiev: My experience is the very first exception does matter. Note that `StandardWorkGroup` terminates its executor service on the first exception (see its `handleError()` method for details), so more than one exceptions are not very likely.

To report all exceptions, the changes will be less compact, I wouldn't be able to use the "cause" mechanism, but will have to report all exceptions explicitly (or nest them one into another, which is weird). So I'd prefer to keep it simple for now. If in the future we'll find need in printing all exceptions, I'll implement it accordingly

anthony-swirldslabs: Sounds fair. Although I still think that all exceptions need to be logged at the very least.

PR #11441, opened 2024-02-08
feat: Ensure overwritten operations check for sufficient gas first

### LINKED ISSUE #11377: Improve gas checks for overwritten operations
### Problem

Checks for sufficient gas should occur at the beginning of all overwritten operations in order to ensure that needless work is not performed.

### Solution

Move checks for sufficient gas to the beginning of all overwritten evm operations. 

### Alternatives

_No response_

lukelee-sl:
**Description**:
Make sure that all overwritten operations check for gas before performing any other tasks.  

These checks are largely redundant as they are also performed in the base operations in Besu but these checks were added at the beginning of each operation execution to ensure that there is enough gas before doing _minimal_ work not related to determining the gas required to perform the operation.

**Related issue(s)**:

Fixes #11377 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/BasicCustomCallOperation.java
```diff
@@ -91,6 +93,11 @@ default Operation.OperationResult executeChecked(@NonNull final MessageFrame fra
         requireNonNull(evm);
         requireNonNull(frame);
         try {
+            final long cost = cost(frame);
```
david-bakin-sl, line 96: (Sigh.  This `cost()` takes a `Frame` whereas the original Besu `cost()` takes a boolean.  Especially noticeable in this particular PR where _both_ are used! Can't change the Besu one, so wish this one had a different name ... darn.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomBalanceOperation.java
```diff
@@ -47,8 +48,12 @@ public CustomBalanceOperation(
     @Override
     public OperationResult execute(@NonNull final MessageFrame frame, @NonNull final EVM evm) {
         try {
-            final var address = Words.toAddress(frame.getStackItem(0));
+            final long cost = cost(false);
```
david-bakin-sl, line 51: Is it ok that we want sufficient gas for a _cold_ operation but we charge (L64 & L68) for the (less expensive?) _warm_ operation?

(Same question for some of the other classes too.)

david-bakin-sl (commented):
I have a question about warm vs cold gas.  So ... this is a comment.  Ping me!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomBalanceOperation.java
lukelee-sl: Good point.  I think the right thing to do is to charge for cold operation as it is unlikely that the address has been warmed by `super.execute()`

david-bakin-sl (approved):
LGTM

Issue #6640, opened 2023-05-17
Remove module swirlds-platform-test


hendrikebbers:
Once we have test-fixtures enabled the `swirlds-platform-test` module should be removed. All sources of the module can be migrated to unit tests in the corresponding modules (`swirlds-platform-core`) and test-fixtures.

PR #11447, opened 2024-02-08
Platform test fixtures moved

hendrikebbers:
fixes [#6640](https://github.com/hashgraph/hedera-services/issues/6640)


mxtartaglia-sl (dismissed):

cody-littley (approved):

imalygin (dismissed):

cody-littley (approved):

mxtartaglia-sl (approved):

nathanklick (commented):
@hendrikebbers If we are to going to these lengths, we should probably just move the tests such as the timeConsuming ones to the separate source set as planned instead of introducing an annotation which uses the tag.

jeromy-cannon (approved):

imalygin (approved):

hendrikebbers:
@nathanklick fine to do that in a future PR. First I want to get rid of all that test-only-modules.

PR #11455, opened 2024-02-08
fix: Fix service names

### LINKED ISSUE #11417: Fix metrics name

netopyr:
Service names are prepended with "proto." which needs to be changed, too.

Closes #11417


netopyr:
Typo in branch name

PR #11456, opened 2024-02-08
fix: Fix service names for metrics

netopyr:
Service names are prepended with "proto.". This PR replaces the '.' as it is not accepted in metric names.


tinker-michaelj (approved):
LGTM!

Issue #11298, opened 2024-01-31
VirtualMapReconnectTest fails intermittently with path not in range log message


artemananiev:
Exception:

```
org.opentest4j.AssertionFailedError: Should not throw a Exception ==> Unexpected exception thrown: org.opentest4j.AssertionFailedError: We did not expect an exception on this reconnect attempt! java.lang.IllegalArgumentException: path (8) is not valid; must be in range (-1,-1) ==> expected: <true> but was: <false>	
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)	
•••
at com.swirlds.virtual.merkle.reconnect.VirtualMapReconnectTest.fullTeacherEmptyLearner(VirtualMapReconnectTest.java:103)	
•••
Caused by: org.opentest4j.AssertionFailedError: We did not expect an exception on this reconnect attempt! java.lang.IllegalArgumentException: path (8) is not valid; must be in range (-1,-1) ==> expected: <true> but was: <false>	
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)	
•••
at com.swirlds.virtual.merkle.reconnect.VirtualMapReconnectTestBase.reconnectMultipleTimes(VirtualMapReconnectTestBase.java:228)	
at com.swirlds.virtual.merkle.reconnect.VirtualMapReconnectTestBase.reconnectMultipleTimes(VirtualMapReconnectTestBase.java:174)	
at com.swirlds.virtual.merkle.reconnect.VirtualMapReconnectTestBase.reconnect(VirtualMapReconnectTestBase.java:170)	
•••
at com.swirlds.virtual.merkle.reconnect.VirtualMapReconnectTest.fullTeacherEmptyLearner(VirtualMapReconnectTest.java:103)	
•••
```

poulok:
We are waiting for a reproduction so that we have more information for debugging.

artemananiev:
May be related: #6991

artemananiev:
PR to print the full stack trace, if an unexpected exception is observed in reconnect tests: https://github.com/hashgraph/hedera-services/pull/11370. It doesn't fix the problem, but will help debug this issue in the future.

artemananiev:
Another PR for this ticket: https://github.com/hashgraph/hedera-services/pull/11435 It propagates reconnect exceptions from the work group to the caller.

poulok:
[...]

PR #11459, opened 2024-02-08
fix: 11298: VirtualMapReconnectTest fails intermittently with path not in range log message

artemananiev:
A new PR to help with https://github.com/hashgraph/hedera-services/issues/11298: virtual map reconnect unit tests are unstable. One of possible reasons is they are resource sensitive and should not be run in parallel with other tests. This PR adds TIMING_SENSITIVE tag to them, let's see if it will make the tests more stable. 

### DIFF
--- a/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTest.java
+++ b/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTest.java
@@ -16,7 +16,6 @@
 
 package com.swirlds.virtual.merkle.reconnect;
 
-import static com.swirlds.common.test.fixtures.junit.tags.TestQualifierTags.TIME_CONSUMING;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTest.java
```diff
@@ -52,6 +52,7 @@
 import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.MethodSource;
 
+@Tag(TestQualifierTags.TIMING_SENSITIVE)
```
anthony-swirldslabs, line 55: That's not great. Presently, this is equivalent to `@Ignored` because the `test` target simply skips tests, and all these tests are expected to be migrated under `src/timeConsuming/` - but even then, it's unclear who and when (if ever) would run `gradlew timeConsuming`. I feel that this a way to permanently disable the test and forget about it, which isn't great because the test may be useful.

It would be best to try and find a way to either make the test run w/o parallelism (but still do run, as opposed to now when it'll be disabled), or put some effort into stabilizing it as is.

anthony-swirldslabs (approved):

platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTest.java
artemananiev: This is a temporary workaround to unblock other people. The test is currently very unstable (not clear why), which leads to many intermittent PR check failures and delays all developers' PRs. That's why the tests are tagged as timing sensitive for now, until we find the reason of failures and fix it (and then remove the tag)

anthony-swirldslabs: Sounds fair. We probably should file an issue to fix the test though.

artemananiev:
Closing this PR as the corresponding changes have been already merged: https://github.com/hashgraph/hedera-services/pull/11475

Issue #11016, opened 2024-01-17
[Security V2 Tests] Add tests for scenarios Burn Token Fungible/NFT


anastasiya-kovaliova:
### Problem

We lack tests covering the V2 security model.

### Solution

Add tests covering the V2 security model functionalities.
Add tests for the following scenarios:

Add positive and negative cases with validation for:

Note: existing V1 tests for the BurnToken functionality will be used as basis for the V2 tests (from classes ContractMintHTSSuite, ContractKeysHTSSuite).

Test Plan: https://www.notion.so/limechain/V2-Security-Model-Test-Plan-Specification-98472925ed534c6d91950273975bd74c?pvs=4

PR #11474, opened 2024-02-09
test: Add tests for V2 token burn with delegate calls

nickeynikolovv:
Description:
Add suggested test from https://github.com/hashgraph/hedera-services/issues/11016
SIGNER → call →CONTRACT → delegatecall → PRECOMPILE

negative scenarios for token burn with delegatecall

Test Plan: https://www.notion.so/limechain/V2-Security-Model-Test-Plan-Specification-98472925ed534c6d91950273975bd74c?pvs=4


Nana-EC:
@attest08 what should be done with this?

PR #11485, opened 2024-02-09
fix: Migration only sets node ID under specific condition

mhess-swl:


tinker-michaelj (approved):
LGTM, tyvm @mhess-swl !

povolev15 (approved):

kimbor (approved):

PR #11487, opened 2024-02-09
feat: 11347: introduce ReconnectHalfMillionNodesBench

### LINKED ISSUE #11347: Need a JMH benchmark for virtual map reconnects
We have a few unit tests (see `VirtualMapReconnectTest` and `MerkleTestUtils`) that simulate reconnects for very simple merkle trees, just a root node + a single virtual map. A new benchmark can be created off this code to test low-level virtual map reconnect performance. It won't be end-to-end (no network, no gossip, no incoming transactions, etc.), but it will still be very useful for debugging and performance assessments. Something similar to `VirtualMapBench` and `CryptoBench`, which are also low-level, yet very powerful.
```[tasklist]
### Tasks
- [ ] Add metrics support to the benchmark
```

anthony-swirldslabs:
**Description**:
This PR introduces:
* `ReconnectHalfMillionNodesBench` - a new benchmark to measure the reconnect performance for a state with half a million nodes running in a local JVM in memory (i.e. w/o any network latencies, disk I/O delays, etc.)
* `StateBuilder` - a utility class that helps build random states for reconnect tests.
* `VirtualMapReconnectBenchBase` (which is a modified copy of `VirtualMapReconnectTestBase` from `swirlds-virtualmap/src/test/`) to support this and future reconnect benchmarks.
* `jmhReconnect` - a Gradle target to run reconnect benchmarks

**Related issue(s)**:

Related: #11347

**Notes for reviewer**:
```
$ g jmhReconnect
...
Result "com.swirlds.benchmark.reconnect.ReconnectHalfMillionNodesBench.reconnectHalfMillionNodes":
  7.044 ±(99.9%) 0.251 s/op [Average]
  (min, avg, max) = (6.945, 7.044, 7.120), stdev = 0.065
  CI (99.9%): [6.794, 7.295] (assumes normal distribution)
...
Benchmark                                                 Mode  Cnt  Score   Error  Units
ReconnectHalfMillionNodesBench.reconnectHalfMillionNodes  avgt    5  7.044 ± 0.251   s/op
...
BUILD SUCCESSFUL in 2m 33s
```

**Checklist**


OlegMazurov:
Why use `BrokenBuilder`, `BreakableDataSource` in the benchmark?
Why not a real `MerkleDbDataSourceBuilder` like in `VirtualMapBench`?

imalygin (dismissed):

jasperpotts (approved):
Exciting to see more JMH tests :-)

anthony-swirldslabs:
@OlegMazurov wrote:
> Why use `BrokenBuilder`, `BreakableDataSource` in the benchmark? Why not a real `MerkleDbDataSourceBuilder` like in `VirtualMapBench`?

There wasn't any specific reason. As mentioned in the Description above, the `VirtualMapReconnectBenchBase` was basically copied over from `VirtualMapReconnectTestBase.java` in `swirlds-virtualmap/src/test/`, and I assumed that this code should be good for the benchmark if it's good for a unit test.

I looked into the `VirtualMapBench` and I don't believe it has direct usages of `MerkleDbDataSourceBuilder` currently. Also, it seems to de-serialize a map from a file instead of building it dynamically, while I'm introducing a randomized StateBuilder in this PR which allows one to build maps in memory dynamically with any given distribution of clean/dirty/missing leaves. So I'm not entirely sure what exactly you're referring to.

However, now that I took a closer look at those `BrokenBuilder` and `BreakableDataSource` implemented at `VirtualMapReconnectBenchBase` (really, they are verbatim copies from VirtualMapReconnectTestBase.java currently), I think that they could be useful in performance testing. For instance, I think that the `BreakableDataSource` could implement randomized delays to simulate network or disk I/O latencies in future reconnect benchmarks, so that we could test the performance for more realistic scenarios in the future. (We could of course give them better names, but I'd defer that until we actually implement the features. Since they're verbatim copies currently, I'd prefer to keep the names intact for now.)

Please let me know if you see a better way to implement such features, or if there's specific reasons why these implementations could interfere with the performance testing of the reconnect procedure that I'm implementing in this PR.

OlegMazurov (changes requested):
`swirlds-benchmark` is for standalone integration-like benchmarks. The idea is to build a single self-contained artifact that can be copied to an arbitrary machine/environment and run there. Testing with actual hardware is a goal, not a nuisance. For this benchmark to be part of that suite, it needs to use real components (i.e. `VirtualMap`, `MerkleDbDataSourceBuilder`), not mocks or stubs.
If, on the other hand, this benchmark is considered a "performance unit test", it should be moved to `platform-sdk/swirlds-virtualmap/src/jmh`.

platform-sdk/swirlds-benchmarks/build.gradle.kts
```diff
@@ -45,3 +49,18 @@ jmh {
...
+    // or run `./gradlew jmhReconnect` to run the override below:
+    doFirst {
+        jmh.includes.set(listOf("Reconnect.*"))
+        jmh.jvmArgs.set(listOf("-Xmx16g"))
+        jmh.fork.set(1)
+        jmh.warmupIterations.set(2)
+        jmh.iterations.set(5)
+    }
+    finalizedBy("jmh")
+}
```
jjohannes, line 66: This is probably just not well documented. You can register additional `JMHTask` instances and have a cleaner configuration:

```suggestion
tasks.register<JMHTask>("jmhReconnect") {
    group = "jmh"
    jarArchive.set(tasks.jmhJar.flatMap { it.archiveFile })
    
    includes.set(listOf("Reconnect.*"))
    jvmArgs.set(listOf("-Xmx16g"))
    fork.set(1)
    warmupIterations.set(2)
    iterations.set(5)
    
    resultsFile.convention(layout.buildDirectory.file("results/jmh/results-reconnect.txt"))
}
```

Note that you still need to run JMH tests with `--no-configuration-cache` until the next release of the JMH plugin which will fix that issue.

jjohannes (changes requested):

anthony-swirldslabs:
@OlegMazurov wrote:
> If, on the other hand, this benchmark is considered a "performance unit test", it should be moved to platform-sdk/swirlds-virtualmap/src/jmh.

Yes, this is a "performance unit test" by design. Since I haven't heard from Artem who suggested to put all benchmarks into swirlds-benchmark, I went ahead and moved the new benchmark to swirlds-virtualmap. Please have a look.

platform-sdk/swirlds-benchmarks/build.gradle.kts
anthony-swirldslabs: Thanks for the suggestion! In fact, I started implementing this using a JMHTask, but I got stuck with the `group` and `jarArchive` definitions, and then I found the `finalizedBy` solution specifically for jmh and went with it.

However, your code looks a bit cleaner than the previous solution, so I've applied this suggestion. Note that I had to also specify the `jvm` parameter for jmh, otherwise it defaulted to an older JVM version that happens to be installed on my computer, and then the benchmark wouldn't run. With the new definition, it runs just fine.

As for the `--no-configuration-cache`, I found that the hard way and I simply set `org.gradle.configuration-cache=false` in my `~/.gradle/gradle.properties`. I know this might be inefficient. However, I prefer when things work rather than fail because of an "optimization". :)

platform-sdk/swirlds-virtualmap/src/jmh/java/com/swirlds/virtualmap/benchmark/reconnect/ReconnectHalfMillionNodesBench.java
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            VirtualMapReconnectBenchBase.startup();
+        } catch (ConstructableRegistryException e) {
+            throw new RuntimeException(e);
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+
+        // Create a state to be reused in every run
+        StateBuilder.buildState(new Random(9823452658L), 500_000, 0.15, 0.15, testTeacherMap::put, testLearnerMap::put);
```
artemananiev, line 52: All these hard-coded values should be changed to JMH params, so they can be changed without recompiling code

platform-sdk/swirlds-virtualmap/src/jmh/java/com/swirlds/virtualmap/benchmark/reconnect/VirtualMapReconnectBenchBase.java
```diff
@@ -0,0 +1,323 @@
...
+    protected final Configuration configuration = new TestConfigBuilder().getOrCreateConfig();
+    protected final ReconnectConfig reconnectConfig = configuration.getConfigData(ReconnectConfig.class);
+
+    protected VirtualDataSourceBuilder<TestKey, TestValue> createBuilder() {
+        return new InMemoryBuilder();
+    }
+
+    protected void setupEach() {
+        final VirtualDataSourceBuilder<TestKey, TestValue> dataSourceBuilder = createBuilder();
+        teacherBuilder = new BrokenBuilder(dataSourceBuilder);
```
artemananiev, line 82: Why do we need BrokenBuilder here? This is a performance benchmark, not a unit test to check how resilient virtual map reconnects are to I/O or network issues

platform-sdk/swirlds-benchmarks/build.gradle.kts
jjohannes: Thanks. I missed the `jvm` part. We could move `group`, `jarArchive`, `jvm` to `com.hedera.hashgraph.benchmark-conventions.gradle.kts` as follows. Then we have this configured centrally for this and all future JMHTasks.

```
tasks.withType<JMHTask>().configureEach {
    group = "jmh"
    jarArchive = tasks.jmhJar.flatMap { it.archiveFile }
    jvm = javaToolchains.launcherFor(java.toolchain).map { it.executablePath }.get().asFile.path
}

```

jjohannes (approved):
👍 for the Gradle-related changes

OlegMazurov (commented):
I'm fine with whatever is deemed useful for a performance unit test.

platform-sdk/swirlds-benchmarks/build.gradle.kts
anthony-swirldslabs: Good idea. I moved these parameters to `com.hedera.hashgraph.benchmark-conventions.gradle.kts`.

platform-sdk/swirlds-virtualmap/src/jmh/java/com/swirlds/virtualmap/benchmark/reconnect/ReconnectHalfMillionNodesBench.java
anthony-swirldslabs: The JMH parameter values would only be available at setupEach() which runs per every invocation of the benchmark. I moved the construction of the state into the static{} block by design in order to avoid wasting time reconstructing the same key/values for the map for every test invocation. So this particular benchmark would continue to use these hard-coded parameters, and potentially we'll even increase the state size here which will make it even more important for this particular benchmark.

At a later time, we can create a separate test/gradle target using the same base class where the values would be parametrized. You'd probably need to use smaller states, or reduce the number of iterations (and hence reduce the confidence of the measurements) in order to support that though. But we can figure this out when we introduce that separate benchmark implementation.

platform-sdk/swirlds-virtualmap/src/jmh/java/com/swirlds/virtualmap/benchmark/reconnect/VirtualMapReconnectBenchBase.java
anthony-swirldslabs: I removed the unnecessary entities.

jjohannes (approved):

jjohannes (approved):

OlegMazurov (approved):

JeffreyDallas (approved):

PR #11500, opened 2024-02-12
fix: 11498 Fix MapTest.

### LINKED ISSUE #11496: Rewrite `VirtualPipelineTests#flushThrottle`
### Description

Current implementation of `VirtualPipelineTests#flushThrottle` is way too sensitive to the environment load and, therefore, unstable. 
It has to be re-written to eliminate dependency on `Thread.sleep`. 

### Steps to reproduce

Run `VirtualPipelineTests#flushThrottle`

### Additional context

_No response_

### Hedera network

other

### Version

v0.47

### Operating system

None

### LINKED ISSUE #11498: Fix `MapTest#insertRemoveAndModifyOneMillion`
A problem was found while investigating one of `MapTest` failures. It's a regression introduced by https://github.com/hashgraph/hedera-services/pull/11268. In that fix, virtual map flushes were changed to check for the old path values before updating key to path mappings. That was the core of the algorithm for improved node remover.

However, same code, `VirtualDataSource.saveRecords()`, is also used for regular flushes, not only to flush data during reconnects. Regular flushes are slightly different, they contain real changes to virtual maps, so at the data source level they must be applied unconditionally, regardless of the old path values.

One possible solution is to introduce a flag to `saveRecords()` that indicates whether this method is called from `VirtualPipeline` (regular flushes) or `AbstractHashListener` (reconnect flushes). Based on this flag, deleted leaves are processed unconditionally, or only if the old path value matches the path in a record to delete.

imalygin:
**Description**:

**Related issue(s)**:

Fixes #11498 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
@@ -493,6 +493,7 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
      * @param leafRecordsToAddOrUpdate stream of new leaf nodes and updated leaf nodes
      * @param leafRecordsToDelete stream of new leaf nodes to delete, The leaf record's key and path
      *     have to be populated, all other data can be null.
+     * @param isReconnectContext if true, the method called in the context of reconnect
      * @throws IOException If there was a problem saving changes to data source
      */
     @Override
@@ -501,7 +502,8 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
             final long lastLeafPath,
             final Stream<VirtualHashRecord> hashRecordsToUpdate,
             final Stream<VirtualLeafRecord<K, V>> leafRecordsToAddOrUpdate,
-            final Stream<VirtualLeafRecord<K, V>> leafRecordsToDelete)
+            final Stream<VirtualLeafRecord<K, V>> leafRecordsToDelete,
+            final boolean isReconnectContext)
             throws IOException {
         try {
             validLeafPathRange = new KeyRange(firstLeafPath, lastLeafPath);
@@ -522,7 +524,8 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
             }
 
             // we might as well do this in the archive thread rather than leaving it waiting
-            writeLeavesToPathToKeyValue(firstLeafPath, lastLeafPath, leafRecordsToAddOrUpdate, leafRecordsToDelete);
+            writeLeavesToPathToKeyValue(
+                    firstLeafPath, lastLeafPath, leafRecordsToAddOrUpdate, leafRecordsToDelete, isReconnectContext);
             // wait for the other threads in the rare case they are not finished yet. We need to
             // have all writing
             // done before we return as when we return the state version we are writing is deleted
@@ -610,18 +613,7 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
         // Go ahead and lookup the value.
         VirtualLeafRecord<K, V> leafRecord = pathToKeyValue.get(path);
 
-        // FUTURE WORK: once the reconnect key leak bug is fixed, this block should be removed
-        if (!leafRecord.getKey().equals(key)) {
-            if (database.getConfig().reconnectKeyLeakMitigationEnabled()) {
-                logger.warn(MERKLE_DB.getMarker(), "leaked key {} encountered, mitigation is enabled", key);
-                return null;
-            } else {
-                logger.error(
-                        EXCEPTION.getMarker(),
-                        "leaked key {} encountered, mitigation is disabled, expect problems",
-                        key);
-            }
-        }
+        assert leafRecord != null && leafRecord.getKey().equals(key);
 
         if (leafRecordCache != null) {
             // No synchronization is needed here, see the comment above
@@ -1170,7 +1162,8 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
             final long firstLeafPath,
             final long lastLeafPath,
             final Stream<VirtualLeafRecord<K, V>> dirtyLeaves,
-            final Stream<VirtualLeafRecord<K, V>> deletedLeaves)
+            final Stream<VirtualLeafRecord<K, V>> deletedLeaves,
+            boolean isReconnect)
             throws IOException {
         if ((dirtyLeaves == null) || (firstLeafPath <= 0)) {
             // nothing to do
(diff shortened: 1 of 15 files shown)


jsync-swirlds (dismissed):

jsync-swirlds (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -611,17 +645,8 @@ public VirtualLeafRecord<K, V> loadLeafRecord(final K key) throws IOException {
         VirtualLeafRecord<K, V> leafRecord = pathToKeyValue.get(path);
 
         // FUTURE WORK: once the reconnect key leak bug is fixed, this block should be removed
```
artemananiev, line 647: I assume the comment may be deleted, too

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/datasource/VirtualDataSource.java
```diff
@@ -87,6 +87,33 @@ void saveRecords(
             final Stream<VirtualLeafRecord<K, V>> leafRecordsToDelete)
             throws IOException;
```
artemananiev, line 89: Does it make sense to change the existing `saveRecords()` method in `VirtualDataSource` to be a default method (calling the new method with "false" as the last arg)?

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -840,6 +841,7 @@ public Stream<VirtualLeafRecord<K, V>> deletedLeaves() {
                 final K key = element.key;
                 final Mutation<K, VirtualLeafRecord<K, V>> mutation = lookup(keyToDirtyLeafIndex.get(key));
                 if (mutation != null && mutation.isDeleted()) {
+                    element.value.setPath(Path.INVALID_PATH);
```
artemananiev, line 844: This is not needed with the new approach

platform-sdk/swirlds-virtualmap/src/test/java/com/swirlds/virtualmap/internal/reconnect/ReconnectHashListenerTest.java
```diff
@@ -164,6 +164,22 @@ public void close() throws IOException {
...
+                Stream<VirtualHashRecord> pathHashRecordsToUpdate,
+                Stream<VirtualLeafRecord<TestKey, TestValue>> leafRecordsToAddOrUpdate,
+                Stream<VirtualLeafRecord<TestKey, TestValue>> leafRecordsToDelete,
+                boolean isReconnectContext)
+                throws IOException {
+            final var ir = pathHashRecordsToUpdate.toList();
+            this.internalRecords.add(ir);
+            final var lr = leafRecordsToAddOrUpdate.toList();
+            this.leafRecords.add(lr);
+            delegate.saveRecords(firstLeafPath, lastLeafPath, ir.stream(), lr.stream(), leafRecordsToDelete);
```
artemananiev, line 180: Should this test use the new method with "true" as the last arg?

imalygin: Yes. It should be. Fixed.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
imalygin: Removed.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/datasource/VirtualDataSource.java
imalygin: Yes, it's a good idea, thank you

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
imalygin: Removed

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/reconnect/VirtualMapReconnectTestBase.java
```diff
@@ -247,13 +247,13 @@ public BreakableDataSource(final BrokenBuilder builder, final VirtualDataSource<
 
         @Override
         public void saveRecords(
-                final long firstLeafPath,
-                final long lastLeafPath,
-                final Stream<VirtualHashRecord> pathHashRecordsToUpdate,
-                final Stream<VirtualLeafRecord<TestKey, TestValue>> leafRecordsToAddOrUpdate,
-                final Stream<VirtualLeafRecord<TestKey, TestValue>> leafRecordsToDelete)
+                long firstLeafPath,
```
alittley, line 250: Is there a reason that you removed the `final`s on these parameters?

alittley (approved):
Approved WRT hashgraph owned file

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -1218,7 +1211,11 @@ private void writeLeavesToPathToKeyValue(
             // dirtyLeaves stream above
             if (isLongKeyMode) {
                 final long key = ((VirtualLongKey) leafRecord.getKey()).getKeyAsLong();
-                longKeyToPath.putIfEqual(key, path, INVALID_PATH);
+                if (isReconnect) {
+                    longKeyToPath.putIfEqual(key, path, INVALID_PATH);
+                } else {
+                    longKeyToPath.put(key, INVALID_PATH);
+                }
```
jsync-swirlds, line 1218: Just curious; it looks like this is the essential fix.
What behavior did we uncover that reconnect should only replace a known/matching path with Invalid; while in other cases we replace or add a path as Invalid, regardless of previous value?

jsync-swirlds (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
imalygin: As `isReconnect` flag suggests, there are two possible contexts:
- Regular flush. There we know for sure that if a tree node got into `deletedLeaves` set, it is deleted for sure and, therefore, the index should be updated unconditionally
- Flush on reconnect. In this case things may get tricky if a path of the key in question changed. In this case we delete conditionally (only if there were no changes), because if the path changed, it's a valid new path and should not be deleted.

jsync-swirlds (dismissed):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/reconnect/VirtualMapReconnectTestBase.java
imalygin: That's a refactoring artifact. Thank you for pointing it out.

jsync-swirlds (dismissed):

jsync-swirlds (approved):

alittley (approved):

alittley (dismissed):

alittley (dismissed):

alittley (approved):

artemananiev (approved):

jsync-swirlds (approved):

PR #11501, opened 2024-02-13
chore: fix yahcli build, `activate-staking` bug

### LINKED ISSUE #11494: yahcli `activate-staking` should not use `0.0.55` for properties override
### Description

Since yahcli only supports loading the key for a single payer account, it should not use any hard-coded payer (e.g., `0.0.55`) whose key is not going to be available.

### Steps to reproduce

Use the current `0.4.1` yahcli tag to `activate-staking` using payer `0.0.2` when the `0.0.55` key has been changed already.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #11494 
 - Extract `com.hedera.services.yahcli` package to a new `SourceSet`.
     * Refactor in small ways to eliminate dependencies in `main` on `yahcli` sources (e.g. replace `COMMON_MESSAGES.info()` with `System.out.println()` in a few places).
     * Move some utilities used by both `main` and `yahcli` to an `AccessoryUtils` class.
 - Use `Dockerfile` base image with a JDK 21.


jeromy-cannon (approved):

mhess-swl (approved):
LGTM

PR #11504, opened 2024-02-13
fix: add legacy etx support before EIP155

### LINKED ISSUE #11401: EIP 155 default v support
### Problem

Hedera provides support for EIP 155.
However, we were a bit to eager and are [blocking support for v=27 and v=28 scenarios](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java#L281)

### Solution

Update logic for legacy transactions EIP 155 to support for v=27 and v=28 scenarios

### Alternatives

_No response_

petreze:
**Description**:
This PR adds support for legacy (unprotected) Ethereum transactions before EIP155

**Related issue(s)**:

Fixes #11401 

**Notes for reviewer**:
- An additional check is added when trying to populate a legacy ethereum transaction to handle the case of `v` field of the transaction being either `27` or `28`
- Added `legacyUnprotectedEtxBeforeEIP155()` and `legacyEtxAfterEIP155()` suite tests to watch the behaviour when `v` field of the transaction is `27` or `28` - which basically happens when the `chainId` is `0`
- Separate those tests in a new `LeakyEthereumTestsSuite` as leaky and run it sequentially in integration tests as they change the network's EVM `chainId` and any concurrent `EthereumCall` with standard `chainId = 298` will fail
- For this reason, a `chainId()` setter is created in `HapiEthereumCall` in order to dynamically set this field
- The first approach to handle WRONG_CHAIN_ID was to add handling in `ConversionUtils.asChainIdBytes()` method used in modu codebase because without it, if we try to construct the `chainId` as bytearray, we generate `[0,0]` which yields `WRONG_CHAIN_ID` when compared to `[]` (byte[0]) which is the `chainId` before EIP155. But after some testing I wanted to assure that we also support `v = 37` and `v = 38` which are corner cases and happen when the `chainId = 1` in the legacy format so instead of tweaking `ConversionUtils.asChainIdBytes()` I used `com.esaulpaugh.headlong.util.Integers.toBytes()` method and it handles everything (I saw that the same method is used in mono as well). I'm not sure if it was performance wise or something else with the custom `asChainIdBytes()` method

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/contract/HapiEthereumCall.java
```diff
@@ -90,6 +90,7 @@ public class HapiEthereumCall extends HapiBaseCall<HapiEthereumCall> {
     private String account = null;
     private ByteString alias = null;
     private byte[] explicitTo = null;
+    private Integer chainId = 298;
```
IvanKavaldzhiev, line 93: `private Integer chainId = CHAIN_ID;` probably would be better

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/contract/HapiEthereumContractCreate.java
```diff
@@ -64,6 +63,7 @@ public class HapiEthereumContractCreate extends HapiBaseContractCreate<HapiEther
     private boolean invalidateEthData = false;
     private Optional<Long> maxGasAllowance = Optional.of(ONE_HUNDRED_HBARS);
     private String privateKeyRef = SECP_256K1_SOURCE_KEY;
+    private Integer chainId = 298;
```
IvanKavaldzhiev, line 66: The same as above

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -353,6 +353,8 @@ private static EthTxData populateLegacyEthTxData(RLPItem rlpItem, byte[] rawTx)
         byte recId = vBI.testBit(0) ? (byte) 0 : 1;
         if (vBI.compareTo(BigInteger.valueOf(34)) > 0) {
             chainId = vBI.subtract(BigInteger.valueOf(35)).shiftRight(1).toByteArray();
+        } else if (vBI.compareTo(BigInteger.valueOf(27)) == 0 || vBI.compareTo(BigInteger.valueOf(28)) == 0) {
+            chainId = new byte[0];
```
david-bakin-sl, line 357: Don't know if this is correct or not (outside of my domain knowledge at this time) but in any event certainly needs a comment to explain the circumstances.

(Also, is an empty array for `chainId` correct, instead of, say, a 2-byte array of `0`?  Not sure how this is used, or why empty array is better than null.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
```diff
@@ -875,6 +879,90 @@ HapiSpec directTransferWorksForERC20() {
                                                         .withErcFungibleTransferStatus(true)))))));
     }
 
+    // test unprotected legacy ethereum transactions before EIP155
+    // this tests the behaviour when the `v` field is 27 or 28
```
david-bakin-sl, line 883: Can this comment on this test - also the next test - point to some documentation somewhere that explains what's going on here?  I.e., so you can know what the behavior is _in words_ before reading these tests and trying to reverse-engineer the desired behavior _from_ the tests just assuming they're correct and complete?

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -353,6 +353,8 @@ private static EthTxData populateLegacyEthTxData(RLPItem rlpItem, byte[] rawTx)
         byte recId = vBI.testBit(0) ? (byte) 0 : 1;
```
david-bakin-sl, line 353: I know this is a `private` method - but that just means we have no unit test for it, apparently.  Is that a good idea?  Aren't there at least 3 things in here that we could be more confident in if there was a unit test?

david-bakin-sl (commented):
Just leaving comments.  This is a fiddly corner case that, if you don't really know what's going on (from, say, reading internal slack messages and remembering them forever) then you'll never know what's going on by reading the code, later.  Really suggest more documentation _in the code_ plus a bit more test code.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -353,6 +353,8 @@ private static EthTxData populateLegacyEthTxData(RLPItem rlpItem, byte[] rawTx)
...
         if (vBI.compareTo(BigInteger.valueOf(34)) > 0) {
```
lukelee-sl, line 354: What happens if vBI is outside of the ranges tested?  Should we throw an exception or would something catch this case down the line?

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
petreze: later on check [here](https://github.com/hashgraph/hedera-services/blob/11401-eip155-support/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java#L281) the two values if they match and null won't be a good option

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
petreze: when we are in the case of `LEGACY_ETHEREUM` txs we go through [one of these two cases](https://github.com/hashgraph/hedera-services/blob/11401-eip155-support/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxSigs.java#L71-L80) where we set the `v` value

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
david-bakin-sl: 1) Can't we then have a named constant for the 0-length byte array? 
2) Similarly, are there any names we can use for the magic numbers `27` and 35`?
3) Can any of this be abstracted behind (short 1-line) methods with reasonable names?

Goal in all these questions is to arrange things so that in the future, when someone is trying to read this, they aren't working down at the level of whether a byte array is 0-length or `null` or something else, or whether the number is supposed to be `27` or maybe `26` or even `28` but are using terms that will be familiar to them (assuming they're familiar with the Ethereum semantics)?

shemnon: 1. `UNPROTECTED_TRANSACTION` would be a good name as it will communicate the effect as well.
2. There are no names for these magic numbers, even in the yellow paper and EIP-155 where they are defined.
3. I view this class as the abstraction. You pass in transaction bytes and out comes `recId` and `chainId` as fields in the object.  There is literally no where else this should need to be know.

Nana-EC: Agree with sentiments above
Here's a combination of them for inspiration
```suggestion
     ...
     // Both Ethereum  yellow paper and https://eips.ethereum.org/EIPS/eip-155 note support for v = 27|28 cases in unprotected transaction cases
     static final BigInteger LEGACY_V_BYTE_SIGNATURE_0 = BigInteger.valueOf(27);
     static final BigInteger LEGACY_V_BYTE_SIGNATURE_1 = BigInteger.valueOf(28);
        ...
        } else if (isVBIFromUnprotectedTransaction(vBI)) {
            chainId = new byte[0];
        ....
        private static boolean isVBIOfUnprotectedTransactionType(BigInteger vBI) {
            vBI.compareTo(LEGACY_V_BYTE_SIGNATURE_0) == 0 || vBI.compareTo(LEGACY_V_BYTE_SIGNATURE_1) == 0
        }
```

Nana-EC (commented):
Agree with much of the commentary and provided a suggestion to clean up approach and improve understanding

Nana-EC (commented):
Move V byte options to class scope

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -444,4 +450,14 @@ private static EthTxData populateEip2390EthTxData(RLPItem rlpItem, byte[] rawTx)
                 rlpList.get(10).data() // s
                 );
     }
+
+    // before EIP155 the value of v in
+    // (unprotected) ethereum transactions is either 27 or 28
+    private static boolean isLegacyUnprotectedEtx(@NonNull BigInteger vBI) {
+        // EIP155 note support for v = 27|28 cases in unprotected transaction cases
+        final BigInteger LEGACY_V_BYTE_SIGNATURE_0 = BigInteger.valueOf(27);
```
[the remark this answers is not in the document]
Nana-EC: I think there's an optimization consideration here.
Moving it to a private static final will avoid this final variable being managed within the method scope each time the method is being called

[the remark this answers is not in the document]
petreze: Agreed!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
```diff
@@ -278,7 +278,7 @@ public HederaEvmTransaction fromContractTxException(
             throw new HandleException(hydratedEthTxData.status());
         }
         final var ethTxData = requireNonNull(hydratedEthTxData.ethTxData());
-        validateTrue(ethTxData.matchesChainId(asChainIdBytes(contractsConfig.chainId())), WRONG_CHAIN_ID);
+        validateTrue(ethTxData.matchesChainId(Integers.toBytes(contractsConfig.chainId())), WRONG_CHAIN_ID);
```
lukelee-sl, line 281: I'd like a better explanation of why this will not throw a `HandleException` with `WRONG_CHAIN_ID` than simply a thumbs down.  In your test in LeakyEthereumTestsSuite you had to override the `contracts.chain_id` property and set the value to 0 in order for your test `legacyUnprotectedEtxBeforeEIP155` to run.  That property will not be set to 0 in mainnet.  With this change will legacy transactions work without changing `contracts.chain_id`?

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
david-bakin-sl: TY Nana, this is in fact what I had in mind to make this code more clear in the future.  (And also I see that comments have been added to point to EIP-155 where they're needed.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
petreze: Thanks for pointing that out! Added another test as well as additional handle in `EthTxData.matchesChainId()` method

lukelee-sl: Nice! Update looks good to me.

lukelee-sl (approved):
LGTM

Nana-EC (approved):
LG

tinker-michaelj (approved):
LGTM!

PR #11525, opened 2024-02-14
feat: Add design doc for atomic crypto transfer

lukelee-sl:
**Description**:
Add a design document for the atomic crypto transfer feature in modularized code base.


hedera-node/docs/design/services/smart-contract-service/atomic-crypto-transfer.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. Validate the semantic correctness of the transaction.
+2. Handle any aliases found and potentially create hollow accounts as necessary.
+3. Handle auto associations
+4. Add transfers to the transfer list
+5. Handle custom fees and add the resulting transfers to the transfer list
+6. Perform the transfers between accounts using the transfer list
+7. Create the resulting record and return to the caller. 
+
+The implementation can be found starting in the `CryptoTransferHandler` class.
```
Nana-EC, line 48: You should highlight the feature flag

hedera-node/docs/design/services/smart-contract-service/hbar-allowance-approval.md
```diff
@@ -0,0 +1,73 @@
...
+Define the classes necessary to complete implementation of the hbar allowance and approval functionality in the Hedera Smart Contract Service.  Only the implementation in the modularized code will be described in this document.
+
+## Non Goals
+
+The implementation for crypto allowance and approval in the Token Service Module will only be described at a high level as this implementation is already completed.
+
+## Architecture
+
+In order to support both EOA and contract callers to these functions, a new interface `IHRC<###>` will be supported and implemented using the pattern established for token proxy contracts as discussed in the `Specification` section of [HIP-719](https://hips.hedera.com/hip/hip-719). 
+The difference is that `IHRC[###]` will act on an account address rather than a token address will be the account in view with respect to the allowance to the `spender` account for the specified amount.  A new proxy contract and a new system contract to handle these transaction requests will be implemented.
```
Nana-EC, line 19:
```suggestion
The difference is that `IHRC[###]` will act on an account address rather than a token address will be the account in view with respect to the allowance to the `spender` account for the specified amount.  A new proxy contract and a new system contract to handle these transaction requests will be implemented similar to HIP 632 logic.
```

Nana-EC (commented):
Looking good.
Flow of pathways in services is very useful.
Might want to consider splitting of docs

hedera-node/docs/design/services/smart-contract-service/atomic-crypto-transfer.md
```diff
@@ -0,0 +1,78 @@
...
+```cryptoTransfer(((address,int64,bool)[]),(address,(address,int64,bool)[],(address,address,int64,bool)[])[])```
+4.  The `ClassicTransferTranslator` class will call the `ClassicTransferDecoder` class to decode the parameters of the call and translating the encoded parameter into a `TransactionBody` object.
+5.  A class called  `ClassicTransfersCall` will then take the created `TransactionBody` object and dispatches a new transaction to the Token Service Module for processing.  A `VerificationStrategy` class will be provided to the token service during dispatch in order to ensure that the security model is adhered to.
+It is also responsible for other miscellaneous tasks such as checking for sufficient gas and encoding the response.
+
+![image info](./class_diagram.drawio.png)
+
+### Token Service Module
+
+Once the smart contract service dispatches the transaction to the Hedera Token Service the following steps are performed:
```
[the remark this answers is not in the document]
lukelee-sl: Not really beyond the those enumerated in item #5 above.  All of the transfer logic is isolated to the token service.

Nana-EC (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @lukelee-sl

hedera-node/docs/design/services/smart-contract-service/hbar-allowance-approval.md
lukelee-sl: Done in new pr for hbar approval

Burstall:
I am very excited to see this upgrade being worked on, as this item is required for several decentralized NFT infrastructure items, with the most obvious example being a Smart contract-based decentralized marketplace. 

The use case is captured in the proposed test; however, I wanted to flesh it out to be sure. Given the network-enforced royalties mechanism, transferring NFTs with a fallback fee to (or by) a Smart Contract is hard to achieve. The workaround we have settled on is to use the existing precompile cryptoTransfer() [hash: 189a554c] method to exchange the NFT for a FT as 'value'. This works but requires all royalty collector accounts to have the designated FT associated (even if there is no royalty to pay, as we use a minimum denomination of the FT as it appears the network checks that the royalty could be paid in the FT before checking if there is any royalty to pay). 

Once we have the new version of this function [hash: 0x0e71804f], it will unlock the ability to build decentralized infrastructure that I see as necessary, if not sufficient, to expand NFTs on Hedera; another essential condition, in my opinion, is for hardware wallets to support Hedera NFTs such that higher value items can be held in cold storage vs. hot wallets but that is outside of scope here. Every step forward is a significant step!

PR #11533, opened 2024-02-14
feat: Create new status panel, and add to existing dashboards

### LINKED ISSUE #11531: Create new platform status grafana panel
- Create a new grafana panel that better visualizes platform status over time
- Add to all existing platform dashboards

alittley:
- closes #11531 

For reviewers: the dashboard changes can be examined in preprod grafana

### DIFF
--- a/hedera-node/infrastructure/grafana/dashboards/production/platform/consensusRoundHandling.json
+++ b/hedera-node/infrastructure/grafana/dashboards/production/platform/consensusRoundHandling.json
@@ -170,12 +170,6 @@
   },
   "__requires": [
     {
-      "type": "panel",
-      "id": "gauge",
-      "name": "Gauge",
-      "version": ""
-    },
-    {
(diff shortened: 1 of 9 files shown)


edward-swirldslabs (approved):
Looking Good

JeffreyDallas (approved):

alittley:
> The platform status panel looks great - thanks for putting them together.
> 
> In our highest resolution environments, we collect metrics from the node software every 3s. Is there any danger in transitions happening faster than that, and thus, we could miss some? Would how fast the node software transitions through states like checking or observing be dependent on state size?

@beeradb There are times when transitions happen fast enough that a 3 second window could miss statuses, but I think only during startup or reconnect, where other indicators make apparent what the platform is up to. If these rapid fire transitions ever were to be meaningful to a debug effort, the platform does log each transition, and we'd be able to drill down that way.

PR #11550, opened 2024-02-15
fix: reduce sync permit count (#11545)

### LINKED ISSUE #11544: Reduce sync permit count
The current number of sync permits is fast in the average case, but we are experiencing some thrashing when the system is under load. Reduce the permit count until the thrashing problem is resolved.

cody-littley:
closes #11544 

Observed test flake related to not enough signatures being collected, known issue.


poulok (approved):

OlegMazurov:
According to this code:
```
        if (syncConfig.onePermitPerPeer()) {
            permitCount = addressBook.getSize() - 1;
        } else {
            permitCount = syncConfig.syncProtocolPermitCount();    // default == 17
        }. 
```
the number of permits will be reduced for large networks (29-node) but increased for smaller networks (7-node). How is performance testing going to be consistent with that change?

PR #11563, opened 2024-02-15
feat: use real birth rounds

### LINKED ISSUE #11555: Events have accurate birthRound
The `TipsetEventCreator` needs to set accurate birthRound (the pending consensus round) in events in the release prior to when we want to migrate to using the birthRound as ancient threshold.   We likely want to set it earlier than that to allow for development and testing of the migration before the version it is released in.

cody-littley:
closes #11555


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/BaseEventHashedData.java
```diff
@@ -321,7 +314,8 @@ public NodeId getCreatorId() {
      * @return the birth round of the event
      */
     public long getBirthRound() {
-        return birthRound;
+        // TODO don't merge until this is reverted
```
edward-swirldslabs, line 317: TODO says don't merge.

cody-littley: Right! Good catch, reverted.

edward-swirldslabs (approved):

alittley (approved):

PR #11577, opened 2024-02-16
fix: Specify transfer precedence to match mono behavior

mhess-swl:


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
```diff
@@ -328,15 +330,33 @@ private TokenTransferList adjustingUnits(
...
             @NonNull final AccountID from,
             @NonNull final AccountID to,
             final long amount,
-            final IsApproval isApproval) {
+            final IsApproval isApproval,
+            @NonNull final TransferPrecedence precedence) {
         final var accountAmounts = new ArrayList<AccountAmount>();
-        accountAmounts.add(debit(from, amount, isApproval));
-        accountAmounts.add(credit(to, amount));
+        if (precedence == TransferPrecedence.DEBIT) {
```
lukelee-sl, line 353: yikes.  this is scary that this matters.

lukelee-sl (approved):
LGTM

tinker-michaelj (approved):
LGTM, thanks @mhess-swl esp for adding the unit test!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
mhess-swl: The only difference it made in the record stream was a different ordering of account amounts. This is purely to match mono service behavior; we can probably remove this pretty quickly after deployment (and I suspect we probably should, cuz agreed – this shouldn't matter)

lukelee-sl (approved):
LGTM

tinker-michaelj (approved):
LGTM, ty @mhess-swl !

Issue #11389, opened 2024-02-06
Event building/mocking in unit tests


lpetrovic05:
A LOT of unit tests build or mock events. They also have very different approaches to doing so. It would be good to standardize all of this for two reasons:
- we end up with cleaner and better code
- it would make the event refactor easier

A class called `GossipEventBuilder` already exists, it would be a good starting point. It might need to be extended to support all use cases.

lpetrovic05:
done

PR #11579, opened 2024-02-16
chore: cleanup event building and mocking in tests

lpetrovic05:
part of #11389


edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/testFixtures/java/com/swirlds/platform/test/fixtures/event/TestingEventBuilder.java
```diff
@@ -0,0 +1,383 @@
...
+            for (int i = 0; i < tr.length; ++i) {
+                final byte[] bytes = new byte[transactionSize];
+                random.nextBytes(bytes);
+                tr[i] = new SwirldTransaction(bytes);
+            }
+        } else {
+            tr = transactions;
+        }
+
+        final long selfParentGen = fakeGeneration != null
```
alittley, line 312: The javadoc for `setGeneration` says "If parents are set, then the generation will be based on the parents provided.", but this implementation looks like a non-null `fakeGeneration` actually is taking precedent over the original generation of a defined parent.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/SerializationTests.java
```diff
@@ -53,20 +51,7 @@ public <T extends SelfSerializable> void serializeDeserializeTest(T generated) t
     }
 
     static Stream<Arguments> selfSerializableProvider() {
```
alittley, line 53: Is it still necessary to have the parameterized test, now that the parameter stream only has 1 element?

alittley (approved):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/SerializationTests.java
lpetrovic05: I don't know who wrote the class, but it looks like the idea was for it to work with any kind of serializable object, its just that this is the only class that was added. I would leave it like this, but I can change it if you feel strongly about it

PR #11584, opened 2024-02-16
feat: Add file handler

### LINKED ISSUE #10908: Implement console handler

### LINKED ISSUE #10909: Implement file handler

timo0:
Fix #10909 
Fix #10908


mxtartaglia-sl (dismissed):

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+import com.swirlds.logging.api.extensions.handler.AbstractSyncedHandler;
+import com.swirlds.logging.api.internal.format.LineBasedFormat;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.BufferedWriter;
+import java.io.FileOutputStream;
+import java.io.OutputStreamWriter;
+import java.nio.file.Path;
+import java.util.Objects;
+
+public class FileHandler extends AbstractSyncedHandler {
```
alittley, line 33: Could you add a class doc explaining what a `FileHandler` is?

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+    private final OutputStreamWriter outputStreamWriter;
+    private final BufferedWriter bufferedWriter;
+
+    /**
+     * Creates a new file handler.
+     *
+     * @param configKey     the configuration key
+     * @param configuration the configuration
+     */
+    public FileHandler(@NonNull String configKey, @NonNull Configuration configuration) {
```
alittley, line 55: nit: parameters may be `final`

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Handles a log event by appending it to the file using the {@link LineBasedFormat},
+     * followed by flushing the console output.
+     *
+     * @param event The log event to be printed.
+     */
+    @Override
+    protected void handleEvent(@NonNull LogEvent event) {
```
alittley, line 90: nit: `event` param may be `final`

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        this.fileOutputStream = fileOutputStream;
+        this.outputStreamWriter = outputStreamWriter;
+        this.bufferedWriter = bufferedWriter;
+    }
+
+    /**
+     * Handles a log event by appending it to the file using the {@link LineBasedFormat},
+     * followed by flushing the console output.
```
alittley, line 85: Is the flush happening in the `print` method call?

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandlerFactory.java
```diff
@@ -0,0 +1,56 @@
...
+package com.swirlds.logging.file;
+
+import com.google.auto.service.AutoService;
+import com.swirlds.config.api.Configuration;
+import com.swirlds.logging.api.extensions.handler.LogHandler;
+import com.swirlds.logging.api.extensions.handler.LogHandlerFactory;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+@AutoService(LogHandlerFactory.class)
+public class FileHandlerFactory implements LogHandlerFactory {
```
alittley, line 26: Could you add a class doc here?

alittley (dismissed):

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final String FILE_NAME_PROPERTY = "%s.file";
+    private static final String BUFFER_SIZE_PROPERTY = "%s.bufferSize";
+    private static final String APPEND_PROPERTY = "%s.append";
+    private static final String DEFAULT_FILE_NAME = "swirlds-log.log";
+
+    /**
+     * The emergency logger.
+     */
+    private static final EmergencyLogger EMERGENCY_LOGGER = EmergencyLoggerProvider.getEmergencyLogger();
```
jsync-swirlds, line 43: This seems like it would (eventually) be common across multiple handlers.  Would it make sense to move this static value to a protected attribute in the `AbstractLogHandler` class?

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -0,0 +1,114 @@
...
+        FileOutputStream fileOutputStream = null;
+        OutputStreamWriter outputStreamWriter = null;
+        BufferedWriter bufferedWriter = null;
+        try {
+            fileOutputStream = new FileOutputStream(filePath.toFile(), append);
+            outputStreamWriter = new OutputStreamWriter(fileOutputStream);
+            bufferedWriter = new BufferedWriter(outputStreamWriter, bufferSize);
+        } catch (final Exception exception) {
+            EMERGENCY_LOGGER.log(Level.ERROR, "Failed to create FileHandler", exception);
+        }
```
jsync-swirlds, line 76: Hmmm...  From what I can tell we only actually need/use the `BufferedWriter` object (the other two are underlying details).
Would it be reasonable, here, to use something _similar to_ the following and leave those underlying details to the JVM classes?
```suggestion
        try {
            if (!Files.exists(path) || Files.isWritable(path)) {
                if (append) {
                    bufferedWriter = Files.newBufferedWriter(filePath, StandardOpenOption.CREATE, StandardOpenOption.APPEND, StandardOpenOption.WRITE, StandardOpenOption.DSYNC);
                } else {
                    bufferedWriter = Files.newBufferedWriter(filePath, StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.DSYNC);
                }
            } else {
                // Log appropriately that the file exists and is not writable
            }
        } catch (final Exception exception) {
            EMERGENCY_LOGGER.log(Level.ERROR, "Failed to create FileHandler", exception);
        }
```

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/extensions/handler/AbstractLogHandler.java
```diff
@@ -29,9 +29,8 @@
  */
 public abstract class AbstractLogHandler implements LogHandler {
 
-    private static final String PROPERTY_HANDLER = "logging.handler.%s";
+    public static final String PROPERTY_HANDLER = "logging.handler.%s";
```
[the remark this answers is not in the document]
timo0: Good catch, moved

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: Added JavaDoc

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: Good catch, moved

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: done

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: Copilot Copy & Paste, thanks :D

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: done

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandlerFactory.java
timo0: Done

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
timo0: I wanted to provide an easy way to specify the buffer size. From what I can see, your approach does not allow for this.

jsync-swirlds: True, we wouldn't be able to specify the buffer size, then again are we certain we actually need to?  In the extreme majority of situations I've seen, changing the buffer size from the default isn't necessary or helpful.
With DSYNC (or flush-after-write to manually do the same thing), which the documentation suggests is used here, a larger or smaller buffer is almost irrelevant.

alittley (dismissed):

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
hendrikebbers: I would say that this is a good point for JMH or something to check what is the best solution. From an api point of view I'm with @jsync-swirlds . But based on the comment from @timo0 other logging frameworks do not use it. So we should find the reason behind that.

Next to that I would prefer to provide config variables only if we really need them. Based on that I would remove the buffer size as a config variable + set it as a fixed value in the code.

platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/api/internal/LoggingSystemTest.java
```diff
@@ -785,6 +803,28 @@ public void accept(LogEvent logEvent) {
...
+                .getOrCreateConfig();
+
+        loggingSystem = new LoggingSystem(configuration);
+
+        // when
+        loggingSystem.installHandlers();
+
+        // then
+        assertThat(loggingSystem.getHandlers()).hasSize(2);
+    }
```
hendrikebbers, line 826: Let's create a test that logging really ends in a file ;)

timo0: Done

alittley (approved):

mxtartaglia-sl (approved):

PR #11589, opened 2024-02-16
feat: 11347: introduce ReconnectBench

### LINKED ISSUE #11347: Need a JMH benchmark for virtual map reconnects

anthony-swirldslabs:
**Description**:
Introducing an integration-test-like benchmark for `VirtualMap` reconnect in `swirlds-benchmarks`. In particular:
* The new `ReconnectBench` that by default builds a state of 5M nodes on disk with a fixed random seed and 5% probabilities of nodes missing or being different on the learner side, and then measures the reconnect time on the state. It's parametrizable using JMH parameters defined at the top of the class.
* A new Gradle target `jmhReconnect` in `swirlds-benchmarks` to simplify running the benchmark for quick testing. Engineers who prefer to use the crude `java -jar ...` approach can still do so the same way they always do.
* Minor refactoring of `VirtualMapBench` into `VirtualMapBaseBench` to extract some common code.
* Introducing `com.swirlds.benchmark.reconnect` with modified/reduced copies of existing test fixtures. Per @OlegMazurov , `Once org.jupiter runtime is enabled with everything it does under the hood it breaks things` and `I thought test fixtures could be accepted but since they are using assertions from org.jupiter they are out`, so I was asked not to take a dependency on the test fixtures directly and instead adapt the code and copy it to the `swirlds-benchmarks` for using in performance tests. While a prior JMH benchmark that I pushed to `swirlds-virtualmap` recently uses the text fixtures directly and works just fine and so personally, I may disagree with the decision to create a copy of the fixtures code, I'm just following the request from @OlegMazurov here.

**Related issue(s)**:

Fixes #11347

**Notes for reviewer**:
```
$ g jmhReconnect
...
> Task :swirlds-benchmarks:jmhReconnect
# JMH version: 1.36
# VM version: JDK 21.0.1, OpenJDK 64-Bit Server VM, 21.0.1+12-LTS
# VM invoker: /Users/anthony/.sdkman/candidates/java/21.0.1-tem/bin/java
# VM options: -Xmx16g
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 1 iterations, 10 s each
# Measurement: 3 iterations, 10 s each
# Timeout: 2147483647 s per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: com.swirlds.benchmark.ReconnectBench.reconnect
# Parameters: (keySize = 16, learnerDifferentProbability = 0.05, learnerMissingProbability = 0.05, maxKey = 1000000, numFiles = 10, numRecords = 100000, numThreads = 32, randomSeed = 9823452658, recordSize = 128, stateSize = 5000000)
...
Result "com.swirlds.benchmark.ReconnectBench.reconnect":
  28.324 ±(99.9%) 8.931 s/op [Average]
  (min, avg, max) = (27.786, 28.324, 28.744), stdev = 0.490
  CI (99.9%): [19.392, 37.255] (assumes normal distribution)


# Run complete. Total time: 00:04:55
...

Benchmark                 (keySize)  (learnerDifferentProbability)  (learnerMissingProbability)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (stateSize)  Mode  Cnt   Score   Error  Units
ReconnectBench.reconnect         16                           0.05                         0.05   1000000          10        100000            32    9823452658           128      5000000  avgt    3  28.324 ± 8.931   s/op

BUILD SUCCESSFUL in 5m 8s
```

**Checklist**


platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -0,0 +1,169 @@
...
+    private VirtualMap<BenchmarkKey, BenchmarkValue> learnerMap;
+
+    private int dbIndex = 0;
+
+    String benchmarkName() {
+        return "ReconnectBench";
+    }
+
+    @Setup
+    public static void setupBenchmark() throws Exception {
```
artemananiev, line 86: I see similar code in `BaseBench` (which this class extends from). Should these additional constructor pairs be moved there instead of introducing a new method?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.IntStream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.openjdk.jmh.annotations.TearDown;
+
+public abstract class VirtualMapBaseBench extends BaseBench {
```
artemananiev, line 40: All these fields/methods can be kept in `VirtualMapBench`, no hard to have them there and inherit in `ReconnectBench`. This is exactly what happens with `CryptoBench`: it extends from `VirtualMapBench` and adds a few more benchmarks

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+            final BiConsumer<K, V> teacherPopulator,
+            final BiConsumer<K, V> learnerPopulator) {
+        LongStream.range(1, size).forEach(i -> {
+            final K key = keyBuilder.apply(i);
+
+            final V teacherValue = valueBuilder.apply(i);
+            teacherPopulator.accept(key, teacherValue);
+
+            if (isRandomOutcome(random, learnerMissingProbability)) {
+                return;
```
artemananiev, line 64: This doesn't correspond very well to what happens with real reconnects. This code emulates a scenario, when a learner is missing some transaction, which was applied to a teacher, yet the subsequent transactions are still applied to the learner. This is not possible with the current hashgraph algorithm.

So let's rewrite this class as follows:
* First, two identical virtual maps are generated of the given size
* The learner map is then used as is
* On the teacher side, a few additional operations are performed (to add, remove, or update an entity in the map), according to the specified probabilities

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: Good point. I moved those calls to the base class.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
anthony-swirldslabs: Unless a specific benchmark method name is requested when running JMH, this approach leads to running all the methods annotated with `@Benchmark`, including those defined in base classes. So I refactored this code on purpose to avoid inheriting from a class that has other benchmarks defined already, so as to simplify launching JMH.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
anthony-swirldslabs: Good catch. I refactored this part in the latest commit and replaced the old learner* probabilities with the three new teacher* probabilities. Please have a look.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+            final double learnerDifferentProbability,
+            final BiConsumer<K, V> teacherPopulator,
+            final BiConsumer<K, V> learnerPopulator) {
+        LongStream.range(1, size).forEach(i -> {
+            final K key = keyBuilder.apply(i);
+
+            final V teacherValue = valueBuilder.apply(i);
+            teacherPopulator.accept(key, teacherValue);
+
+            if (isRandomOutcome(random, learnerMissingProbability)) {
```
OlegMazurov: This is not a realistic scenario for a diverged learner. The Merkle structure of the two trees will be completely different. Basically, the learner state should be replicated in the teacher and then modified with `add()`, `put()` and `remove()`.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/PairedStreams.java
```diff
@@ -0,0 +1,108 @@
...
+    protected BufferedInputStream learnerInputBuffer;
+    protected MerkleDataInputStream learnerInput;
+
+    protected Socket teacherSocket;
+    protected Socket learnerSocket;
+    protected ServerSocket server;
+
+    public PairedStreams() throws IOException {
+
+        server = new ServerSocket(0);
```
OlegMazurov, line 49: Platform socket configuration should be replicated here. See `com.swirlds.platform.network.SocketConfig`

OlegMazurov: Can `TcpFactory` be reused here?

OlegMazurov (commented):
Overall, the approach seems appropriate. Better integration with the benchmarking framework is desired.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -0,0 +1,169 @@
...
+    /** A random seed for the StateBuilder. */
+    @Param({"9823452658"})
+    public long randomSeed;
+
+    /**
+     * The size of the state, aka the number of nodes in the teacher virtual map.
+     * The learner map will have the same number of nodes, or less than that
+     * depending on the {@code learnerMissingProbability} value.
+     */
+    @Param({"5000000"})
```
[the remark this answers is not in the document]
anthony-swirldslabs: I've replaced the custom `stateSize` with using the `numRecords` defined in the BaseBench.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -0,0 +1,169 @@
...
+        final Path merkleDbPath = getTestDir().resolve("merkledb" + dbIndex++);
+        MerkleDb.setDefaultPath(merkleDbPath);
+    }
+
+    @Override
+    protected VirtualMap<BenchmarkKey, BenchmarkValue> createEmptyMap() {
+        throw new UnsupportedOperationException();
+    }
+
+    private VirtualMap<BenchmarkKey, BenchmarkValue> createEmptyMap(String label) {
```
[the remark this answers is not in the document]
anthony-swirldslabs: No, it shouldn't. This was done on purpose here. I've provided more details in the `UnsupportedOperationException` message in the overridden `createEmptyMap()` method.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -0,0 +1,169 @@
...
+    @TearDown(Level.Invocation)
+    public void tearDownInvocation() throws Exception {
+        final var finalTeacherMap = teacherMap;
+        final var finalLearnerMap = learnerMap;
+
+        teacherMap = null;
+        learnerMap = null;
+
+        afterTest(() -> {
+            finalTeacherMap.getDataSource().close();
```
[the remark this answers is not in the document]
anthony-swirldslabs: The maps are getting released in the `finally{}` block in the `ReconnectRunner.reconnect(...)`.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -0,0 +1,169 @@
...
+        MerkleDbDataSourceBuilder<BenchmarkKey, BenchmarkValue> dataSourceBuilder =
+                new MerkleDbDataSourceBuilder<>(tableConfig);
+        return new VirtualMap<>(label, dataSourceBuilder);
+    }
+
+    @Setup(Level.Invocation)
+    public void setupInvocation() {
+        beforeTest("reconnect");
+
+        teacherMap = createEmptyMap("teacher");
```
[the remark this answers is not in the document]
anthony-swirldslabs: That's true. However, the current benchmark with 5 million records runs 1 warmup and 5 measurement iterations and completes in about 5 minutes, so the current time looks reasonable to me. So this isn't necessary at the moment.

We can implement this optimization at a later time, in a separate PR, if/when we start testing with a lot larger states.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
```diff
@@ -0,0 +1,177 @@
...
+     *
+     * <pre>
+     *             root
+     *           / |   \
+     *          A  i0  i1
+     *             /\  /\
+     *            B C D null
+     * </pre>
+     */
+    public static BenchmarkMerkleInternal buildLessSimpleTree() {
```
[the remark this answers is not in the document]
anthony-swirldslabs: The benchmark is designed to delegate as much of the reconnect functionality as possible to the platform (aka production) code. This is essentially achieved by passing the states to the Learning- and Teaching-Synchronizers, both of which operate with full state trees rather than with VirtualMaps directly. So the BenchmarkMerkleLeaf is used to construct the full state with a few nodes where the only most important node is the VirtualMap. I'd prefer not to have even this current boilerplate code here in the benchmark, but my understanding is that there's not a more top-level entry point to start a reconnect. There's no reason to use any lower-level APIs to try and only run the process for the VirtualMap instances because the lower we go, the more dependencies we create between the benchmark and the platform code, and this would make the maintenance of this benchmark more complex. For this reason, the BenchmarkMerkleLeaf and friends stay.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/BenchmarkMerkleInternal.java
```diff
@@ -0,0 +1,139 @@
...
+
+package com.swirlds.benchmark.reconnect;
+
+import com.swirlds.common.merkle.MerkleInternal;
+import com.swirlds.common.merkle.MerkleNode;
+import com.swirlds.common.merkle.impl.PartialNaryMerkleInternal;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiFunction;
+
+public class BenchmarkMerkleInternal extends PartialNaryMerkleInternal implements BenchmarkMerkleNode, MerkleInternal {
```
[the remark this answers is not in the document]
anthony-swirldslabs: Please see above.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/BenchmarkMerkleLeaf.java
```diff
@@ -0,0 +1,143 @@
...
+import com.swirlds.common.io.streams.SerializableDataOutputStream;
+import com.swirlds.common.merkle.MerkleLeaf;
+import com.swirlds.common.merkle.MerkleNode;
+import com.swirlds.common.merkle.impl.PartialMerkleLeaf;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiFunction;
+
+public class BenchmarkMerkleLeaf extends PartialMerkleLeaf implements BenchmarkMerkleNode, MerkleLeaf {
```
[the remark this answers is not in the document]
anthony-swirldslabs: Please see above.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/BenchmarkPlatformContextBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+import com.swirlds.config.api.ConfigurationBuilder;
+import com.swirlds.metrics.api.Metrics;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * A simple builder to create a {@link PlatformContext} for tests.
+ */
+public final class BenchmarkPlatformContextBuilder {
+
+    private static final Metrics defaultMetrics = new NoOpMetrics();
```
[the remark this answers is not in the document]
anthony-swirldslabs: We will implement metrics support in a separate PR. This is outside of scope of this introductory PR. I'll create a task for this.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
anthony-swirldslabs: Yes, this has been addressed already in a commit pushed just before you posted your comments.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/PairedStreams.java
anthony-swirldslabs: Not sure how critical this is given that the test runs on 127.0.0.1, but I did that.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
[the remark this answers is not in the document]
anthony-swirldslabs: Sounds good! I added the argument to it and moved its implementation from CryptoBenchMerkleDb to VirtualMapBaseBench, so there's no need to override this method anywhere else.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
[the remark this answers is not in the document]
anthony-swirldslabs: Yeah, I've seen those messages too. However, all these files are getting deleted nevertheless and after the test finishes, only the metrics CSV files are found in the data/ directory. I'm not sure why the messages are printed but let me spend some time on this...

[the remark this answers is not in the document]
anthony-swirldslabs: This appears to be a thread race between the MerkleDB/VirtualMap implementation and the benchmark's `@TearDown` that tries to delete the test directory recursively. Calling release()/close() would delete the DB files eventually but not immediately. There's even a comment in the ReconnectBench stating that deleted directories may be re-created in background.

So I'm solving this by replacing Files.delete() with Files.deleteIfExists() in the Utils, and also adding a small Thread.sleep() right after closing the datasources to help prevent the messages from appearing.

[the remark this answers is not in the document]
OlegMazurov: According to `swirlds-vmap.log` all databases are closed by platform code (two teacher's and two learner's). I get a clean run when I comment out `tearDownInvocation()` altogether.

OlegMazurov (dismissed):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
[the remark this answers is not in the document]
anthony-swirldslabs: Interesting. I believe the code here releases()/closes() the databases explicitly anyway. The close() on the data source is quoted above, and the release() call happens in the ReconnectRunner. I kept these calls in their sequence similar to what happens in e.g. the VirtualMapBench. Further, if nothing else, the `tearDownInvocation()` resets the state of the ReconnectBench object by clearing the maps references. So I'll keep the method as is.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/BaseBench.java
```diff
@@ -126,6 +139,18 @@ public void setup() throws IOException {
...
+            registry.registerConstructable(
+                    new ClassConstructorPair(BenchmarkMerkleInternal.class, BenchmarkMerkleInternal::new));
+            registry.registerConstructable(
+                    new ClassConstructorPair(BenchmarkMerkleLeaf.class, BenchmarkMerkleLeaf::new));
+            registry.registerConstructable(new ClassConstructorPair(VirtualLeafRecord.class, VirtualLeafRecord::new));
+            registry.registerConstructable(new ClassConstructorPair(VirtualMap.class, VirtualMap::new));
+            registry.registerConstructable(new ClassConstructorPair(VirtualMapState.class, VirtualMapState::new));
+            registry.registerConstructable(new ClassConstructorPair(VirtualRootNode.class, VirtualRootNode::new));
+            registry.registerConstructable(new ClassConstructorPair(BenchmarkKey.class, BenchmarkKey::new));
+            registry.registerConstructable(new ClassConstructorPair(BenchmarkValue.class, BenchmarkValue::new));
```
artemananiev, line 153: I wonder if `registerConstructable()` with a package name can be used instead of these individual constructor pairs? For example, `VirtualLeafRecord` is in `com.switlds.virtualmap.datasource`, so would `registerConstructables("com.swirlds.virtualmap.datasource")` work by any chance?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * <pre>
+     *             root
+     *           / |   \
+     *          A  i0  i1
+     *             /\  /\
+     *            B C D null
+     * </pre>
+     */
+    static BenchmarkMerkleInternal buildLessSimpleTree() {
```
artemananiev, line 54: This method is used only in `ReconnectRunner`, and only to replace one of the (grand)children with a virtual map. What if this method accepts a `VirtualMap` instance and returns the simplest tree possible: a root (merkle internal) with a single leaf (virtual map)?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+
+            if (teacherAdd) {
+                final K key = keyBuilder.apply(i + size);
+                final V value = valueBuilder.apply(i + size);
+                teacherPopulator.accept(key, value);
+            }
+
+            // Don't bother modifying if we're about to remove it
+            if (teacherRemove) {
+                final K key = keyBuilder.apply(i);
```
artemananiev, line 91: It means, only original nodes (in range 1..size) may be removed from the map. It works just fine, but it would be even better to track the number of added nodes (in the teacherAdd branch above) and select a node to remove from range 1..size+added

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
```diff
@@ -0,0 +1,185 @@
...
+
+        initializeTreeAfterCopy(root);
+        return root;
+    }
+
+    static <T extends MerkleNode> T hashAndTestSynchronization(
+            final MerkleNode startingTree,
+            final MerkleNode desiredTree,
+            final Configuration configuration,
+            final ReconnectConfig reconnectConfig)
```
artemananiev, line 81: I may be wrong, but reconnectConfig is redundant here. It can be obtained from configuration using
```
final ReconnectConfig reconnectConfig = configuration.getConfigData(ReconnectConfig.class);
```

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
```diff
@@ -0,0 +1,185 @@
...
+                        } catch (final IOException e) {
+                            // test code, no danger
+                            e.printStackTrace();
+                        }
+                    },
+                    reconnectConfig);
+            final PlatformContext platformContext =
+                    BenchmarkPlatformContextBuilder.create().build(configuration);
+            teacher = new TeachingSynchronizer(
+                    platformContext.getConfiguration(),
```
artemananiev, line 127: I wonder if `configuration` can be used here instead of `platformContext.getConfiguration()`? If so, the whole BenchmarkPlatformContextBuilder class is not needed

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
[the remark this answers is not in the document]
artemananiev: I agree we need a merkle tree in the benchmark, not just a virtual map. However, it's enough to have this tree minimal: just a root with a single child (virtual map). It would eliminate `BenchmarkMerkleLeaf` class, at least.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/ReconnectRunner.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class ReconnectRunner {
+    /**
+     * Run reconnect with the given teacher and learner maps.
+     *
+     * @param teacherMap a teacher map aka a good, desired state
+     * @param learnerMap a learner map aka a stale, outdated, "wrong" state
+     * @throws Exception thrown if the reconnect code throws any exceptions
+     */
+    public static void reconnect(
```
artemananiev, line 39: Why isn't this method a part of ReconnectBench? This ReconnectRunner class looks redundant (especially if its `createTreeForMap()` method is eliminated in favor of `MerkleBenchmarkUtils.buildLessSimpleTree()`)

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/BaseBench.java
anthony-swirldslabs: They sure would. However, we're losing compile-time checks (javac wouldn't bother checking if the string literals refer to actual Java packages), and we're potentially getting runtime errors if someone adds classes to these packages that are incompatible with the constructable registry for whatever reasons. But okay.

artemananiev (approved):

jeromy-cannon (approved):

PR #11606, opened 2024-02-19
feat: Add metrics that measure the duration for each transaction type

### LINKED ISSUE #11603: Add metrics to measure transaction duration
Add metrics that measure the maximum and average duration per transaction type

netopyr:
This PR adds two metrics for each transaction type to measure the average and maximum duration in ns.

Closes #11603


tinker-michaelj (approved):
Makes sense to me!

PR #11608, opened 2024-02-19
feat: limit number of events sent during a sync

### LINKED ISSUE #11607: Limit number of events in a sync
Set an upper limit for the number of events to send during a sync. Will protect nodes that are behind from getting massive syncs.

cody-littley:
closes #11607


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/shadowgraph/ShadowgraphSynchronizer.java
```diff
@@ -382,6 +388,10 @@ private List<EventImpl> createSendList(
             sendList = eventsTheyMayNeed;
         }
 
+        if (maximumEventsPerSync > 0 && sendList.size() > maximumEventsPerSync) {
+            sendList = sendList.subList(0, maximumEventsPerSync);
```
lpetrovic05, line 392: I think its a lot better if we don't even add them to the list in the first place

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/config/SyncConfig.java
```diff
@@ -52,4 +53,5 @@ public record SyncConfig(
         @ConfigProperty(defaultValue = "true") boolean filterLikelyDuplicates,
         @ConfigProperty(defaultValue = "3s") Duration nonAncestorFilterThreshold,
         @ConfigProperty(defaultValue = "500ms") Duration syncKeepalivePeriod,
-        @ConfigProperty(defaultValue = "1m") Duration maxSyncTime) {}
+        @ConfigProperty(defaultValue = "1m") Duration maxSyncTime,
+        @ConfigProperty(defaultValue = "100") int maxSyncEventCount) {}
```
lpetrovic05, line 57: I think we should start with a really high number, and then lower it when we are able to adequately test this

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/shadowgraph/ShadowgraphSynchronizer.java
cody-littley: Challenge with that approach is that we gather events out of order, and sort them at the end. It's not obvious to me how to just gather the right amount.

alittley (dismissed):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/config/SyncConfig.java
cody-littley: I've disabled this feature by default (0 == no size limit). We can enable it when it gets to the perf environments where we can test it.

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/shadowgraph/ShadowgraphSynchronizer.java
lpetrovic05: yeah, lets not complicate it now

lpetrovic05 (approved):

Issue #10100, opened 2023-11-24
Transferring tokens with HTS precompile to system accounts between 0.0.350 and 0.0.361 returns INVALID_ALIAS_KEY


anastasiya-kovaliova:
### Description

When we call the HTS precompile to execute a token transfer function from EOA to a system account (for an address in the range reserved for system accounts but with no initiallized account in it) for all addresses in the range 0.0.350 and 0.0.361 and the address 0.0.0 we expect the child record transaction to fail with the INVALID_RECEIVING_NODE_ACCOUNT error. The error we get is INVALID_ALIAS_KEY.

### Steps to reproduce

The test reproducing this behaviour is in the following PR: https://github.com/hashgraph/hedera-mirror-node/pull/7204 in the equivalence.feature file in "Scenario Outline: Validate in-equivalence tests for HTS Transfers" - all addresses for which we have this error have comments.

### Additional context

- In-equivalence test plan: https://www.notion.so/limechain/In-Equivalance-Test-plan-2c3f12807b6c4f85b061e922e7b3aede?pvs=4
- Hedera/Ethereum precompiles equivalence document: https://www.notion.so/limechain/Hedera-Ethereum-precompiles-equivalence-shared-26657bbc66394b4b92622ac99d0b907d?pvs=4
- PR with the tests validating the returned error: https://github.com/hashgraph/hedera-mirror-node/pull/7204

### Hedera network

mainnet

### Version

0.44

### Operating system

None

0xivanov:
Note: `testHtsTokenTransferToNonExistingSystemAccount` in test-clients

PR #11643, opened 2024-02-21
fix: Transferring tokens with HTS precompile to system accounts between 0.0.350 and 0.0.361 returns INVALID_ALIAS_KEY

stefan-stefanooov:
**Description**:
When we call the HTS precompile to execute a token transfer function from EOA to a system account (for an address in the range reserved for system accounts but with no initiallized account in it) for all addresses in the range 0.0.350 and 0.0.361 and the address 0.0.0 we expect the child record transaction to fail with the INVALID_RECEIVING_NODE_ACCOUNT error. The error we get is INVALID_ALIAS_KEY.

**Related issue(s)**:
[Transferring tokens with HTS precompile to system accounts between 0.0.350 and 0.0.361 returns INVALID_ALIAS_KEY #10100](https://github.com/hashgraph/hedera-services/issues/10100)

Fixes #
[10100](https://github.com/hashgraph/hedera-services/issues/10100)

Hedera network
mainnet

Version
0.44


stefan-stefanooov:
@rbair23 & @thomas-swirlds-labs 
Hello Guys,
Can you please have a look here and give your feedback ?

tinker-michaelj (approved):
LGTM, tyvm @stefan-stefanooov !

PR #11646, opened 2024-02-21
feat: [Block Node] Add entry point and gradle task

### LINKED ISSUE #11645: Block Node: Add entry point and gradle task
### Problem

Currently Block Node is empty project with no entry point.

### Solution

Add entry point and gradle task for easy startup

georgi-l95:
**Description**:
This PR adds enty point class and gradle task to start the application, as well as configuration for log4j and ground work for configuration files for the block node itself.

**Related issue(s)**:

Fixes #11645 

**Notes for reviewer**:

**Checklist**


block-node/blocknode-core/src/test/java/com/hedera/node/blocknode/core/test/ExampleTest.java
```diff
@@ -1,75 +0,0 @@
...
-                return null;
-            }
-
-            @Override
-            public DummyCoreSpi spi() {
-                return null;
-            }
-
-            @Override
-            public DummyGrpcApi grpcApi() {
```
nickpoorman, line 52: Nathan put these in here because we want to make sure we conform to good patterns. We should have our SPIs and their implementations separate. I don't think we should remove these yet until we have laid the groundwork for these things.

block-node/blocknode-core/src/main/java/com/hedera/node/blocknode/core/grpc/impl/BlockNodeService.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class BlockNodeService // extends BlockServiceGrpc.BlockServiceImplBase
+ {
+    private static final Logger logger = LogManager.getLogger(BlockNodeServer.class);
+
+    public BlockNodeService() {}
+
+    //    @Override
```
nickpoorman, line 33: Let's not commit a bunch of commented out stuff.

block-node/blocknode-core/src/main/java/module-info.java
```diff
@@ -2,16 +2,17 @@
...
-    requires com.hedera.storage.blocknode.filesystem.local;
-    requires com.hedera.storage.blocknode.filesystem.s3;
-
+    requires com.hedera.storage.blocknode.config;
+    requires com.swirlds.config.api;
+    requires io.grpc;
+    requires org.apache.logging.log4j;
     // Require modules which are needed for compilation and should be available to all modules that depend on this
     // module (including tests and other source sets).
-    requires transitive com.hedera.storage.blocknode.core.spi;
```
nickpoorman, line 12: Again, I don't think we should be removing these.

nickpoorman (changes requested):
@georgi-l95 I think we need to get a solid foundation here before we begin layering in the complexities of the business logic. We spent a great deal of time on modularization to get us to a point where things were properly broken out into modules and we should strive to have clean interfaces and implementations from day 1 on block nodes.

Issue #11304, opened 2024-02-01
Unit Test Failure: `VirtualMapHashingTest#fullLeavesRehash` fails intermittently 


imalygin:
### Description

[com.swirlds.virtualmap.VirtualMapHashingTest](https://scans.gradle.com/s/7ctqo3b42kwww/tests/task/:swirlds-virtualmap:test/details/com.swirlds.virtualmap.VirtualMapHashingTest.fullLeavesRehash(int)[3]) is failing intermittently. 

It looks like it happens under the load. I found this in the logs
```
18:34:18.747  59595 Test worker Flushed in 180 ms 
18:34:18.748  59596 Test worker Doing full rehash for the path range: 7775 - 15550  in the VirtualMap at [1]
```
Normally it takes 10-20 ms. 

See [this link ](https://scans.gradle.com/s/7ctqo3b42kwww/tests/task/:swirlds-virtualmap:test/details/com.swirlds.virtualmap.VirtualMapHashingTest/fullLeavesRehash(int)%5B3%5D?top-execution=1) for more details

### Steps to reproduce

1. Run make sure that CPU resources and disk I/O are limited
2. Run `VirtualMapHashingTest.fullLeavesRehash`

### Hedera network

other

### Version

v0.47

### Operating system

None

artemananiev:
Another JRS failure with the same symptoms: http://35.247.76.217:8095/swirlds-automation/release/0.47/4N/Reconnect/20240220-094051-GCP-Daily-Reconnect-4N/VMerkleDual-NDReconnect-500-60m/

PR #11660, opened 2024-02-21
fix: 11304: Unit Test Failure: VirtualMapHashingTest#fullLeavesRehash fails intermittently

artemananiev:
Fix summary: fixed a thread race in `ConcurrentBlockingIterator` between `close()` and `hasNext()`. In some rare cases, a consumer thread calling `hasNext()` could return false, if a supplier thread supplied an element to the iterator and immediately after that closed the iterator.

A new unit test is provided.

Fixes: https://github.com/hashgraph/hedera-services/issues/11304


anthony-swirldslabs (dismissed):

platform-sdk/swirlds-virtualmap/src/test/java/com/swirlds/virtualmap/internal/reconnect/ConcurrentBlockingIteratorTest.java
```diff
@@ -220,6 +221,28 @@ void multiThreadIteration() throws ExecutionException, InterruptedException {
...
+            } catch (final InterruptedException e) {
+                fail("Supplier interrupted");
+            }
+        });
+        supplier.start();
+        assertTrue(iterator.hasNext(), "Iterator must have more than zero elements");
+        assertEquals(1, iterator.next());
+        assertTrue(iterator.hasNext(), "Iterator must have more than one element");
+        assertEquals(2, iterator.next());
+        assertFalse(iterator.hasNext(), "Iterator must not have more than one elements");
```
OlegMazurov, line 242: s/one/two/

OlegMazurov (dismissed):

platform-sdk/swirlds-virtualmap/src/test/java/com/swirlds/virtualmap/internal/reconnect/ConcurrentBlockingIteratorTest.java
artemananiev: Nice catch. I will fix it

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/ConcurrentBlockingIterator.java
```diff
@@ -109,17 +109,15 @@ public boolean hasNext() {
...
         final long timeOutWhenMillisAre = System.currentTimeMillis() + waitMillis;
-        while ((next = buffer.poll()) == null) {
-            if (closed.get()) {
-                return false;
-            } else {
-                if (System.currentTimeMillis() > timeOutWhenMillisAre) {
-                    throw new RuntimeException(new TimeoutException("Timed out trying to read from buffer"));
-                }
+        boolean isOpen = !closed.get();
+        while (((next = buffer.poll()) == null) && isOpen) {
```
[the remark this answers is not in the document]
artemananiev: This is a good question. I am personally not sure this code is so performance critical as to justify the busy loop above. I'd expect this loop to help with local reconnect tests (where node transfers between teacher and learner are very fast), but may not be very helpful in real networks. Needs further investigation

OlegMazurov (approved):

imalygin (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/ConcurrentBlockingIterator.java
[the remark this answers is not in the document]
anthony-swirldslabs: Sounds good.

anthony-swirldslabs (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/ConcurrentBlockingIterator.java
jsync-swirlds, line 113: Given the time resolution for the wait is milliseconds, it _might_ be worth considering using the timeout version of [poll](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/LinkedBlockingQueue.html#poll(long,java.util.concurrent.TimeUnit)) with a 1 millisecond timeout.  That would still check the closed value every millisecond, but avoid the thousands of additional loops in between.

Also, could this be rewritten as follows (I think it makes sense, but you might have a reason for splitting it out)?
```
        while (((next = buffer.poll()) == null) && !closed.get()) {
```

artemananiev: There is a reason. If written like above, thread context switch may happen after the call to `poll()`, but before the call to `closed.get()`. This is exactly the problem addressed in this PR. With the way how it's currently in the PR, a call to `poll()` is guaranteed after `closed.get()` returns true.

artemananiev: Regarding the `poll()` calls, let me file a low-priority ticket about it. As I wrote above in response to @anthony-swirldslabs comments, the chances are we don't need a busy loop here at all

artemananiev: Filed https://github.com/hashgraph/hedera-services/issues/11691

PR #11667, opened 2024-02-21
perf: improve VirtualHasher performance

### LINKED ISSUE #11636: Reduce object allocations in the new virtual hasher
This is a follow-up for https://github.com/hashgraph/hedera-services/issues/10961.

A few optimizations in the new `VirtualHasher` implementation that may help with Java object allocations:

1. When a task is created for a leaf node, there is no need to create its `ins` array. It will not be used anyway. Either a null or a real leaf will be used as a task input

2. Sometimes a sibling task is created only to call `setParent()` and `setData(null)` on it. Instead, a null input may be set on the parent task, bypassing creating a task object

Another issue is not directly related to object allocations, but still looks like a valuable optimization:

3. There is a loop to process existing tasks starting from `stack[curRank]` to `curPath`. After the loop, `stack` for the current rank is invalidated, i.e. set to `INVALID_PATH`. However, it may happen that `curPath` is actually in the same chunk as `stack[curRank]`. In this case, all existing tasks between `curPath` and the end of the chunk will hang in the tasks map and will be processed after the last leaf only (in the loop to set null data for all tasks remaining in the map), despite these tasks are known to be clear. To fix it, `stack[curRank]` should be set to `curPath - 1`, if `curPath` is in the same rank

OlegMazurov:
**Description**:
* reduce object allocation
* remove synchronization around notification and make it task-based
* tune virtualHasherChunkHeight

**Related issue(s)**:

Fixes #11636

**Notes for reviewer**:
```
Benchmark                             (keySize)   (maxKey)  (numFiles)  (numRecords)  (numThreads)  (recordSize)  Mode  Cnt      Score   Error  Units
CryptoBenchMerkleDb.transferPrefetch         24  500000000        6000        100000            32          1024  avgt       16815.734           s/op
```
Previous best score - 18268 (the lower the better).

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
+++ b/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.LongFunction;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -93,6 +94,16 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
     private VirtualHashListener<K, V> listener;
 
     /**
+     * A task holding the resulting hash. Used to synchronize all parallel computations.
+     */
+    private HashHoldingTask resultTask;
+
+    /**
+     * This atomic reference is used to serialize (non-deterministically) all notification tasks.
+     */
+    private final AtomicReference<NotificationTask> nextTaskPlaceHolder = new AtomicReference<>();
+
+    /**
      * An instance of {@link Cryptography} used to hash leaves. This should be a static final
      * field, but it doesn't work very well as platform configs aren't loaded at the time when
      * this class is initialized. It would result in a cryptography instance with default (and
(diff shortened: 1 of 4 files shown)


alittley (dismissed):
Approved, with respect to the hashgraph owned file

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -200,20 +246,29 @@ public void completeExceptionally(Throwable ex) {
...
+            if (path == ROOT_PATH) {
+                currentTask.setData(leaf, path, hash, resultTask);
+            } else {
+                final NotificationTask nextTask = new NotificationTask(HASHING_POOL, 2);
+                while (!nextTaskPlaceHolder.compareAndSet(currentTask, nextTask)) {
+                    currentTask = nextTaskPlaceHolder.get();
+                }
+                currentTask.setData(leaf, path, hash, nextTask);
+            }
+        }
```
anthony-swirldslabs, line 260: It's a bit difficult to understand what exactly this code is trying to achieve and how exactly (and why) it's building a linked list of tasks here, and why the "head" of this linked list (pointed to by the `nextTaskPlaceHolder`) shouldn't have any data set initially. And how the "linked list" is structured, and what the dependencies between its elements are. And how this custom "linked list" is different from using, say, a plain LinkedList<> from JDK.

I realize that there may be performance-related reasons to maintain this data structure here this way, but they are not immediately clear from reading this code.

Also, given that the currentTask at line 250 is fetched from an atomic reference and then is modified by calling its `setData()` at line 252, it's unclear whether there may be a thread race where another thread would run this exact method and would either fall into the same branch or the `else` branch, and then do something to this same currentTask value previously fetched by the first thread at line 250.

It would help to add some comments (preferably, a javadoc for this method) explaining this a bit.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -145,51 +156,86 @@ public Hash hash(
...
+            this.leaf = leaf;
+            send();
         }
 
-        void setHash(final int index, final Hash hash) {
-            assert index >= 0 && index < (1 << height);
-            ins[index] = hash;
-            send();
+        void complete() {
+            out.send();
```
artemananiev, line 238: It would be great to have a couple assertions here. For "leaf" tasks make sure no leaf data is set. For "internal" tasks make sure no input hashes are set.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -436,29 +496,30 @@ public Hash hash(
                         continue;
                     }
                     if (siblingPath > lastLeafPath) {
+                        assert siblingPath == 2;
```
artemananiev, line 499: I am not sure this is right. Assume chunk height is 3, there is only one chunk (starting off the root node), and leaf path range is 2-4. In this case, siblings for paths 5 and 6 would fall into this code branch, but the assertion is for path 2 only...

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -400,7 +460,7 @@ public Hash hash(
                     while (curStackPath < Math.min(curPath, lastPathInCurStackChunk)) {
                         final ChunkHashTask t = map.remove(curStackPath);
                         assert t != null;
-                        t.setData(null);
+                        t.complete();
                         curStackPath++;
                     }
                     stack[curRank] = INVALID_PATH;
```
artemananiev, line 466: As I wrote in #11636 (see item 3 in the bug description), this may be improved, if curPath is less than lastPathInCurStackChunk (that is, the current path is in the same chunk as the path in the stack), stack[curRank] may be set to curPath + 1

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/AbstractHashListener.java
```diff
@@ -121,70 +115,41 @@ public synchronized void onHashingStarted() {
...
-                dirtyLeavesToFlush = leaves;
-                leaves = new ArrayList<>();
-            } else {
-                dirtyHashesToFlush = null;
-                dirtyLeavesToFlush = null;
-            }
-        }
-        if ((dirtyHashesToFlush != null) && (dirtyLeavesToFlush != null)) {
-            flush(dirtyHashesToFlush, dirtyLeavesToFlush);
+        hashes.add(new VirtualHashRecord(path, hash));
```
artemananiev, line 118: Two concerns:

1. Here in this class (which is used during reconnects): synchronization in hash listener methods is not needed any more as notifications are now serialized. Fine. However, when a flush happens, it would block the current notification task and all subsequent tasks until the flush is complete. Previously, it would block only one task, but other tasks could accumulate hashes/leaves in parallel

2. During regular hashing in the end of each round, when hashes are put to `VirtualNodeCache`, it was done in parallel (and the cache is ready to handle that), but with this new approach it will be done sequentially. I wonder how copy hash time is affected by this change?

alittley (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
OlegMazurov: It works because `parentRankHeights[lastLeafRank] == 1` so the only possibility for `siblingPath` to be greater than `lastLeafPath` is when it's a degenerate tree with `lastLeafPath == 1`.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
anthony-swirldslabs: We've discussed this offline, and this point is now clear. However, I'd still suggest for @OlegMazurov to prepare and publish some documentation about the framework in general which would be helpful to engineers unfamiliar with how it works.

jsync-swirlds:
This PR appears to be abandoned (replaced by #11787).
Could this be closed?

PR #11669, opened 2024-02-21
fix: Use an alias payer account to submit a transaction should be rejected.

### LINKED ISSUE #11651: (mod) `TransactionChecker.checkTransactionID()` permits `0.0.<alias>` payer id
### Description

In a perf environment, a throttle check in ingest failed because the payer `AccountID` was not of form `0.0.X`.
```
2024-02-21 00:19:05.607 ERROR 143  MethodBase - Possibly CATASTROPHIC failure while handling a call and running the ingest workflow
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "com.hedera.hapi.node.base.AccountID.accountNum()" is null
	at com.hedera.node.app.throttle.ThrottleAccumulator.throttleExempt(ThrottleAccumulator.java:504) ~[HederaNode.jar:?]
	at com.hedera.node.app.throttle.ThrottleAccumulator.shouldThrottleTxn(ThrottleAccumulator.java:316) ~[HederaNode.jar:?]
	at com.hedera.node.app.throttle.ThrottleAccumulator.shouldThrottle(ThrottleAccumulator.java:132) ~[HederaNode.jar:?]
	at com.hedera.node.app.throttle.SynchronizedThrottleAccumulator.shouldThrottle(SynchronizedThrottleAccumulator.java:57) ~[HederaNode.jar:?]
	at com.hedera.node.app.workflows.ingest.IngestChecker.runAllChecks(IngestChecker.java:206) ~[HederaNode.jar:?]
	at com.hedera.node.app.workflows.ingest.IngestWorkflowImpl.submitTransaction(IngestWorkflowImpl.java:86) ~[HederaNode.jar:?]
	at com.hedera.node.app.grpc.impl.TransactionMethod.handle(TransactionMethod.java:56) ~[HederaNode.jar:?]
```

We would expect to reject such a `TransactionID#accountID` field before checking throttles.

### Steps to reproduce

Use an aliased payer account to submit a transaction.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

iwsimon:
**Description**:

**Related issue(s)**:

Fixes #11651 

**Notes for reviewer**:

**Checklist**


tinker-michaelj (commented):
Hmm did you consider just changing `checkTransactionID()` to no longer accept an aliased payer account [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/TransactionChecker.java#L371)?

If we leave `checkTransactionID()` as-is, there might be another place that re-uses `TransactionChecker` and then _also_ fails to reject an aliased payer.

iwsimon:
> Hmm did you consider just changing `checkTransactionID()` to no longer accept an aliased payer account [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/TransactionChecker.java#L371)?
> 
> If we leave `checkTransactionID()` as-is, there might be another place that re-uses `TransactionChecker` and then _also_ fails to reject an aliased payer.

I was following the exception chain on the issue. You are right. Changed the fix to checkTransactionID().

tinker-michaelj (approved):
LGTM, tyvm @iwsimon !

PR #11672, opened 2024-02-22
chore: refactor mod `ThrottleService, fix restart ISS

### LINKED ISSUE #11652: (mod) `CongestionThrottleService.CONGESTION_LEVEL_STARTS` ISS post-reconnect
### Description

After reconnect in a perf environment, node encounters self-ISS isolated to,
```
platform-sdk/swirlds-cli % diff e14.txt a14.txt 
1,2c1,2
< (root) State                                                                                           /          patch-asthma-day-pass
<    0 MerkleHederaState                                                                                 /0         aware-sound-borrow-oak
---
> (root) State                                                                                           /          train-spring-second-gloom
>    0 MerkleHederaState                                                                                 /0         donor-cube-design-vicious
12c12
<       3 SingletonNode         CongestionThrottleService.CONGESTION_LEVEL_STARTS                        /0/3       pizza-true-nose-scare
---
>       3 SingletonNode         CongestionThrottleService.CONGESTION_LEVEL_STARTS                        /0/3       smile-song-reveal-gym
14c14
<          1 ValueLeaf                                                                                   /0/3/1     hungry-party-tobacco-exhaust
---
>          1 ValueLeaf 
```

### Steps to reproduce

Reconnect a node.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #11652 
 - Avoid ISS on restart with active congestion level starts by ensuring configuration reflects system file overrides _before_ initializing `ThrottleService` components.
 - Replace a lot of hand-wired object creation in `Hedera` with Dagger2 DI; especially, use DI for a new `ThrottleServiceManager` that isolates the responsibility of managing the major events in the `ThrottleService` lifecycle: 
     1. Rebuilding all components from `HederaState`.
     2. Recreating throttles and congestion multipliers on an update to system file `0.0.123`.
     3. Recreating gas throttles and congestion multipliers.
     4. Saving usage snapshots and congestion start times to a `HederaState`.
 - Remove all references to throttle service from the `MigrationContext` (makes no sense for throttling-specific infrastructure components to be there).
    * Instead of doing some elaborate machinations in `migrate()` of `ThrottlingService` schema, simply init singletons to default values of `ThrottleUsageSnaphots` and `CongestionLevelStarts` at genessis.
 - Add a [`SchemaApplicationType`](https://github.com/hashgraph/hedera-services/blob/8322e70e6b93c989923b4054db26d92773e3dcd5/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java#L281) enum in `MerkleSchemaRegistry` to hopefully make it easier to understand how applicable schemas are used.
 - Rename `EntityUtilizationMultiplier` as `UtilizationScaledThrottleMultiplier` to hopefully clarify it is, in fact, a `ThrotttleMultiplier` that just has additional scaling based on state utilization.
 - Allow a `GasLimitDeterministicThrottle` to have capacity `0` to eliminate a pointless special case (no `gas`-throttled operation can ever request less than ~21k `gas`).
 - Stop trying to work around a null `gasThrottle` field in `ThrottleAccumulator`---this is an invalid state, and made multiple unit tests extremely confusing to interpret.
 - Rename `ThrottleManager` as `ThrottleParser` and make it a stateless helper, previous stateful behavior was misleading and confusing.


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/throttle/HandleThrottleParser.java
```diff
@@ -1,57 +0,0 @@
-/*
```
mhess-swl, line 1: I'm concerned that removing this interface goes against the whole point of the modularization project, which was to split components into more manageable pieces. At least in theory, the SPI plays an important part in defining what those pieces are supposed to be, and their interactions with each other. By removing this interface, and in turn creating `ThrottleServiceManager` (which isn't a Service as defined by the Service interface), we go back to the mono model of handling things by essentially injecting singletons everywhere as an implementation detail, instead of a defined interface that the application uses to assign its implementation. As a result, the code is extremely difficult to reason about – even at a high level – because all the major pieces depend not on behaviors with boundaries, but instead are entirely subject to implementation details. 

That's isn't to say that the current throttle implementation is ideal either – I'm not a throttles expert for sure – maybe the way it is now adds unnecessary complexity that I'm not seeing. But this feels like a step towards making the system harder to understand.

tinker-michaelj: This interface was simply unnecessary, was only used in the `MigrationContext` and didn't actually need to be there.

tinker-michaelj: I.e. this type wasn't replaced by `ThrottleServiceManager` (which isolates the logic for the major `ThrottleService` "lifecycle" events instead of spreading them across `Hedera`, `NetworkUtilizationManager`, and a `ThrottleManager`); it simply had no purpose.

mhess-swl: > This interface was simply unnecessary; it was only used in the MigrationContext and didn't actually need to be there.

Missed this, you're right, thanks. 

I agree it's nice to have all the logic for throttle lifecycle events in one place; I may even agree that spreading them across `NetworkUtilizationManager` and `ThrottleManager` seems messy (I don't know enough to make this argument); but it also seems incorrect to entirely encapsulate these previously public-ish (or at least not private) API methods inside of an implementation and remove them from the SPI entirely... Would it make sense to instead put them in a separate (new) interface in the SPI?

tinker-michaelj: I think there will eventually be _some_ throttling analogue to the `FeeCalculator` that the `HandleContext` makes available to services.

At that point, it will make sense to define some throttle-related types in SPI.

But right now, throttling is managed entirely by the workflows; so there is no reason to have any throttling types in SPI.

mhess-swl (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
```diff
@@ -155,25 +160,29 @@ public void migrate(
...
         requireNonNull(currentVersion);
         requireNonNull(config);
         requireNonNull(networkInfo);
-        requireNonNull(handleThrottling);
 
         // Figure out which schemas need to be applied based on the previous and current versions, and then for each
         // of those schemas, create the new states and remove the old states and migrate the data.
         final var schemasToApply = computeApplicableSchemas(previousVersion, currentVersion);
+        if (schemasToApply.isEmpty()) {
+            logger.info("Service {} does not use state", serviceName);
```
Neeharika-Sompalli, line 173: Service doesn't upgrade right ?

iwsimon (approved):
LGTM. Thanks @tinker-michaelj !

Neeharika-Sompalli (approved):

PR #11675, opened 2024-02-22
build: rename 'com.hedera.node.blocknode' -> 'com.hedera.storage'

jjohannes:
**Description**:

When introduce, _Block Node projects_ all got a `com.hedera.node.blocknode...` Module Name. This was because of an issue in the Module Dependencies Gradle plugin - https://github.com/gradlex-org/java-module-dependencies/issues/90

This PR updates the plugin and renames the modules to  use the desired `com.hedera.storage` Module Names.

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/11236

**Notes for reviewer**:

**Checklist**


jjohannes:
Hey @georgi-l95, the dependencies are managed purely by adding `requires` to the `module-info.java` files. In your cases that would be:

`requires org.apache.logging.log4j;` (or `requires transitive` if used in public api)

`requires io.grpc;` (or `requires transitive` if used in public api)

`requires static com.github.spotbugs.annotations;` (name of module is different than the package in the module 🤷 )

The versions for the dependencies are managed in [hedera-dependency-versions/build.gradle.kts](https://github.com/hashgraph/hedera-services/blob/develop/hedera-dependency-versions/build.gradle.kts), but as these three are already in use in other places, there are already versions defined

We have _dependency analysis tooling_ in place as part of the build and the CI pipeline, which means:
- We can't add the dependencies now, because it will fail telling use they are unused and can be removed
- Once you start using the dependencies, you can add them and the check will tell you if  `requires` vs `requires transitive` is correct

Locally, you can run `./gradlew qualityGate` (or start the `qualityGate` task from IntelliJ). This includes running the dependency analysis.

georgi-l95:
@jjohannes Thank you very much 🙏

georgi-l95 (approved):

nathanklick (changes requested):
@jjohannes The team originally had requested the groupId `com.hedera.storage` and the prefix of `com.hedera.storage.blocknode` such that the individual modules should be `blocknode-*`, but this PR eliminates the `blocknode-*` module prefix. 

Now that the gradlex module is fixed, would it be possible to refactor this to `com.hedera.blocknode` with a module prefix of `blocknode-*`?

jjohannes:
@nathanklick thanks for clarifying. Adjusted the rename as requested.

The package names in the modules should probably be adjusted accordingly as well by the team once this is merged.

nathanklick (approved):

PR #11690, opened 2024-02-22
feat: faster wiring backpressure

### LINKED ISSUE #11689: Faster Back Pressure Implementation
We've observed situations where the wiring framework backpressure is slow under high contention. Make it faster.

cody-littley:
closes #11689 

I ran the following test on my laptop:
- spin up 10 threads (i.e. the number of cores I have)
- on each thread, attempt to call `onRamp()` many times in a loop
- count how long it takes for all threads to complete

I observed a 4x speedup in this test when I switched to the new implementation in this PR.


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
```diff
@@ -73,17 +77,19 @@ public BackpressureObjectCounter(
...
     public void onRamp() {
-        while (true) {
-            final long currentCount = count.get();
-            if (currentCount < capacity) {
-                final boolean success = count.compareAndSet(currentCount, currentCount + 1);
-                if (success) {
-                    return;
-                }
-            }
+        final long resultingCount = count.incrementAndGet();
```
lpetrovic05, line 80: how did you test the performance of this?

cody-littley: I performed the following test on my laptop.

- spin up one thread per core (10)
- create an object counter with a very high capacity (not attempting to measure back pressure performance, want to see performance when back pressure is not yet kicking in)
- on each thread, call `BackpressureObjectCounter.onRamp()` millions of time in a loop
- measure the amount of time required for all threads to finish

In this test, I observed a 4x speedup given the code changes in this branch.

alittley (dismissed):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
lpetrovic05: it can be very useful if more work gets done on this in the future, or if Oleg wants to analyze it. I think committing it makes more sense than throwing it away, but up to you

lpetrovic05 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/counters/BackpressureObjectCounter.java
cody-littley: Main problem I have now is that I didn't save the test, so I'd have to rewrite it. Willing to do so if you think it's worth it.

alittley (approved):

edward-swirldslabs (approved):

PR #11718, opened 2024-02-23
fix: 11715 Properly handle closed `PairedStreams` object

### LINKED ISSUE #11715: Unit Test Failure: paired streams may be closed twice
Many reconnect tests use `MerkleTestUtils.testSynchronization()` to emulate reconnects. This method creates a `PairedStreams` object, which is used for communications between learner and teacher. The object is auto-closeable and used in try-with-resources block. In addition to that, it's sometimes closed explicitly from within learner/teacher synchronizers. It may result in an unexpected test failure.

An option is to not use try-with-resources, but close the streams explicitly after synchronization, and only if they haven't been closed yet.

imalygin:
**Description**:

This PR prevents possible failures in attempt of closing `PairedStreams` object which was previously closed. If `close` method throws an expection in this case the test shouldn't fail.

**Related issue(s)**:

Fixes #11715

### DIFF
--- a/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
+++ b/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
@@ -984,7 +984,8 @@ public final class MerkleTestUtils {
             final int latencyMilliseconds,
             final ReconnectConfig reconnectConfig)
             throws Exception {
-        try (PairedStreams streams = new PairedStreams()) {
+        final PairedStreams streams = new PairedStreams();
+        try {
 
             final LearningSynchronizer learner;
             final TeachingSynchronizer teacher;
@@ -1083,6 +1084,13 @@ public final class MerkleTestUtils {
             assertReconnectValidity(startingTree, desiredTree, generatedTree);
 
             return (T) generatedTree;
+        } finally {
+            try {
+                streams.close();
+            } catch (IOException e) {
+                // test code, no danger. The stream could have been closed previously.
+                e.printStackTrace();
+            }
         }
     }
 
--- a/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
+++ b/platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
@@ -984,8 +984,7 @@ public final class MerkleTestUtils {
             final int latencyMilliseconds,
             final ReconnectConfig reconnectConfig)
             throws Exception {
-        final PairedStreams streams = new PairedStreams();
-        try {
+        try (PairedStreams streams = new PairedStreams()) {
 
             final LearningSynchronizer learner;
             final TeachingSynchronizer teacher;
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/MerkleTestUtils.java
```diff
@@ -1083,6 +1084,13 @@ public static <T extends MerkleNode> T testSynchronization(
             assertReconnectValidity(startingTree, desiredTree, generatedTree);
 
             return (T) generatedTree;
+        } finally {
+            try {
+                streams.close();
+            } catch (IOException e) {
+                // test code, no danger. The stream could have been closed previously.
+                e.printStackTrace();
+            }
```
anthony-swirldslabs, line 1093: A safer approach would be to check if the stream is closed before calling `close()`. This would prevent the test from swallowing exceptions when it shouldn't swallow them.

anthony-swirldslabs: Alternatively, the stream implementation shouldn't error-out when `close()` is called on an already closed stream. This would fix the root cause of the issue instead of implementing workarounds such as suggested above or presented in the current commit here.

imalygin: Fair enough. I moved exception handling to `PairedStreams`. It's a test fixture anyway, so I think it should be fine.

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/PairedStreams.java
```diff
@@ -80,29 +80,39 @@ public MerkleDataInputStream getLearnerInput() {
...
+            teacherOutputBuffer.close();
+            teacherInputBuffer.close();
+            learnerOutputBuffer.close();
+            learnerInputBuffer.close();
+
+            server.close();
+            teacherSocket.close();
+            learnerSocket.close();
+        } catch (IOException e) {
+            // this is the test code, and we don't want the test to fail because of a close error
```
anthony-swirldslabs, line 99: We actually do want it to fail if the stream was opened and the close() fails for some reason. It's better to catch this in a test than in production. But okay.

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/PairedStreams.java
artemananiev: This `PairedStreams` class is only used in tests, not in production

artemananiev (approved):

platform-sdk/swirlds-common/src/testFixtures/java/com/swirlds/common/test/fixtures/merkle/util/PairedStreams.java
anthony-swirldslabs: The `close()` method in the stream implementations runs in production, and this fix swallows an exception thrown by that production method. So my point still stands.

PR #11722, opened 2024-02-23
feat: split transaction prehandle out of `SwirldStateManager`

### LINKED ISSUE #11113: Split app transaction prehandle into separate class
App transaction prehandle is currently part of `SwirldStateManager`, but I don't think there is any reason that it needs to be. Prehandle isn't synchronized in any way with normal handling.

App transaction prehandle should be split into a separate class.

alittley:
- closes #11113
- closes https://github.com/hashgraph/hedera-services/issues/12144


cody-littley (approved):

lpetrovic05:
From @poulok :

> is it possible that the latest immutable state will be null for a period of time at startup when attempting to handle transactions? When we load a state from disk, we copy it (or at least we used to) so that there is an immutable version for prehandle. As Austin pointed out, that is synchronous so it’s not a problem. But if it becomes asynchronous, could that be a problem?

alittley:
Addressing the comment from @poulok :

First, the `latestImmutableState` nexus in `SwirldsPlatform` is provided with an `initialState` in the constructor of `SwirldsPlatform`, so it shouldn't ever be `null` by the time transactions are being handled.

```
latestImmutableState.setState(initialState.reserve("set latest immutable to initial state"));
```

Second, even if there *were* to be a race condition with a temporarily `null` `latestImmutableState`, the new `TransactionPrehandler` contains logic to wait until a valid `latestImmutableState` is available

```
            latestImmutableState = latestImmutableStateNexus.getState("transaction prehandle");
            while (latestImmutableState == null) {
                latestImmutableState = latestImmutableStateNexus.getState("transaction prehandle");
            }
```

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/eventhandling/TransactionPrehandler.java
```diff
@@ -0,0 +1,107 @@
...
+        final EventImpl eventImpl = new EventImpl(event, null, null);
+
+        final long startTime = time.nanoTime();
+
+        ReservedSignedState latestImmutableState = null;
+        try {
+            latestImmutableState = latestImmutableStateNexus.getState("transaction prehandle");
+            while (latestImmutableState == null) {
+                latestImmutableState = latestImmutableStateNexus.getState("transaction prehandle");
+            }
```
lpetrovic05, line 89: this has the potential to spin indefinitely and not provide any feedback. maybe after like 100 tries we should log something and sleep, then try again

lpetrovic05 (approved):

edward-swirldslabs (changes requested):
Lazar's comment is PR blocking.  I'm testing the change requested status to see if it gets cleared after pushing a commit to the PR.

lpetrovic05:
> Lazar's comment is PR blocking. I'm testing the change requested status to see if it gets cleared after pushing a commit to the PR.

I did not understand this

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/eventhandling/TransactionPrehandler.java
edward-swirldslabs: spinning indefinitely consumes a thread/cpu for useless work.  I presume this needs to be fixed before the PR can merge.

alittley: I can improve this if there is consensus on correct strategy- but note that the spinny wait is just carrying forward the previous functionality.

alittley: I lean towards this being outside the scope of this PR

edward-swirldslabs: I've removed the block from the PR being merged.   

If the solution had stayed with the AtomicReference being updated by an asynchronous method call, I would have suggested using notify() and wait().  But with the introduction of the nexus,  what about a blocking get from the nexus?

alittley: I think it would be a good future issue to add a blocking get from the nexus.

I created one here https://github.com/hashgraph/hedera-services/issues/11989

edward-swirldslabs (dismissed):

edward-swirldslabs (dismissed):

edward-swirldslabs (approved):

cody-littley (dismissed):

cody-littley (approved):

edward-swirldslabs (dismissed):

cody-littley (approved):

edward-swirldslabs (dismissed):

lpetrovic05 (approved):

cody-littley (approved):

cody-littley (approved):

PR #11731, opened 2024-02-26
build: automate publishing to Maven Central

### LINKED ISSUE #11237: Add support for automatic Maven Central releases
## Description

We need to add support to automatically release the existing artifacts published to Maven Central currently. Additionally, we want to enable publishing of the remaining `hedera-*` artifacts with the exception of `hedera-mono-service` and `test-clients`. No changes required to the `com.swirlds` artifacts which are being published except to automatically release them. The new `block-node` folder and it's projects should be published and automatically released as well.

jjohannes:
**Description**:
- Fully automate publishing to Maven Central (by utilizing https://github.com/gradle-nexus/publish-plugin)
- Publish all _Modules_ of _services_ (hedera-node)
- Publish all _Modules_ of _storage_ (block-node)

As a side effect, `assemble` and `qualityGate` now also build the Javadoc and find Javadoc errors.
This PR fixes the errors that were found.

**Note:** new/reworked convention plugin files follow the naming pattern that will be used consistently with #12997

**Related issue(s)**:

Fixes #11237

**Notes for reviewer**:

**Checklist**


.github/workflows/node-zxc-build-release-artifact.yaml
```diff
@@ -610,7 +610,7 @@ jobs:
           OSSRH_PASSWORD: ${{ secrets.svcs-ossrh-password }}
         with:
           gradle-version: ${{ inputs.gradle-version }}
-          arguments: "releaseEvmMavenCentral --scan -PpublishSigningEnabled=true --no-configuration-cache --no-parallel"
+          arguments: "releaseEvmMavenCentral --scan -PpublishSigningEnabled=true --no-configuration-cache"
```
jjohannes, line 613: `--no-parallel` no longer needed as we now properly communicate to nexus that all publications "belong together" by using the nexus plugin.

block-node/blocknode-filesystem-s3/build.gradle.kts
```diff
@@ -14,4 +14,7 @@
  * limitations under the License.
  */
 
-plugins { id("com.hedera.hashgraph.blocknode.conventions") }
+plugins {
+    id("com.hedera.hashgraph.blocknode.java")
+    id("com.hedera.hashgraph.blocknode.publish")
```
jjohannes, line 19: I started to unify a bit more how we name things in the Gradle plugins. Right now, some things have a `convention` suffix and some don't. I would drop the suffix, as everything here is a "convention" in the Gradle sense.

I now have for each set of projects a ".java" (saying this is a Java project with specific conventions) and a ".publish" (this is a published library with specific conventions)

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.blocknode.publish.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+plugins {
+    id("java")
+    id("com.hedera.hashgraph.maven-publish")
+}
+
+publishing {
+    publications {
+        named<MavenPublication>("maven") {
+            pom.developers {
+                // TODO fill this
```
jjohannes, line 26: Should something be changed here directly? Can also do this as follow up.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.nexus-publish.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+            username = System.getenv("OSSRH_USERNAME")
+            password = System.getenv("OSSRH_PASSWORD")
+        }
+    }
+}
+
+tasks.named("closeSonatypeStagingRepository") {
+    // The publishing of all components to Maven Central is automatically done before close (which
+    // is done before release).
+    dependsOn(subprojects.map { ":${it.name}:releaseMavenCentral" })
```
jjohannes, line 34: This is the important "tweak" - all projects are published before releasing

Following what we did and tested here: https://github.com/hashgraph/pbj/blob/main/pbj-core/buildSrc/src/main/kotlin/com.hedera.pbj.root.gradle.kts#L41-L45

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.platform.publish.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+// Register one 'release*' task for each publishing repository
+publishing.repositories.all {
+    val ucName = name.replaceFirstChar { it.titlecase() }
+    tasks.register("release$ucName") {
+        group = "release"
+        dependsOn(tasks.named("publishMavenPublicationTo${ucName}Repository"))
+    }
```
jjohannes, line 94: I removed duplication from this file by using a loop here. I also renamed the file and reformatted. That's why Git things this is all new (it isn't).

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.platform.publish.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+        developer {
+            name = "Release Engineering Team"
+            email = "release-engineering@swirldslabs.com"
+            organization = "Hedera Hashgraph"
+            organizationUrl = "https://www.hedera.com"
+        }
+    }
+}
+
+publishing.repositories {
```
jjohannes, line 70: ❓ Is it correct to keep these additional repos only for "platform"? Are they all still used at all? Should they be available for all components and not only "platform"?

hedera-node/hedera-evm/build.gradle.kts
```diff
@@ -31,11 +29,11 @@ testModuleInfo {
...
-            artifactId = "hedera-evm"
-        }
-    }
+// Second publication using the 'group' and 'name' under wich 'evm'
+// was published before
+publishing.publications.create<MavenPublication>("mavenLegacy") {
+    from(components["java"])
+
+    groupId = "com.hedera.evm"
+    artifactId = "hedera-evm"
```
jjohannes, line 38: ❓ If we change the coordinates, other services components that we now publish would break as they assume the unfied coordinates in their dependencies. The way to keep things working is too publish this component twice with the new and "old" coordinates. Is this fine? Maybe this second publication can then be dropped late (1.0 release?).

nathanklick: @jjohannes We must be able to support publishing different components into different maven central group IDs. Publishing into two different coordinates is not an acceptable solution. The EVM should be isolated into the `com.hedera.evm` Group ID and all other components should be referencing the EVM via the `com.hedera.evm` Group ID.

nathanklick: @jjohannes I have applied a set of changes which maintain the EVM MC coordinates. This is necessary as well because other projects depend on and use these coordinates.

iwsimon (dismissed):
LGTM.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.evm.publish.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+                    name = "Hedera Smart Contracts Team"
+                    email = "hedera-smart-contracts@swirldslabs.com"
+                    organization = "Hedera Hashgraph"
+                    organizationUrl = "https://www.hedera.com"
+                }
+                developer {
+                    name = "Release Engineering Team"
+                    email = "release-engineering@swirldslabs.com"
+                    organization = "Hedera Hashgraph"
+                    organizationUrl = "https://www.hedera.com"
```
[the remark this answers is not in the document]
nathanklick: Yes

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.platform.publish.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+    //    https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools/issues/85
+    // 2. It does not interact well with the 'gradle-nexus.publish-plugin' plugin, causing:
+    //    'No staging repository with name sonatype created' during IDE import
+    publishing.repositories.remove(publishing.repositories.getByName("sonatype"))
+    apply(plugin = "com.google.cloud.artifactregistry.gradle-plugin")
+}
+
+publishing.publications.named<MavenPublication>("maven") {
+    pom.description =
+        "Swirlds is a software platform designed to build fully-distributed " +
```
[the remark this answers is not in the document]
nathanklick: This is one of those legacy hold overs and we don't currently have approval to change it.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.platform.publish.gradle.kts
nathanklick: @jjohannes These should be deprecated soon in favor of JFrog Artifactory. Let's just leave these as Platform only at this time.

nathanklick (approved):

nathanklick (dismissed):

nathanklick (approved):

jeromy-cannon (approved):

Neeharika-Sompalli (approved):
LGTM from hedera-base

build-logic/project-plugins/src/main/kotlin/com.hedera.gradle.blocknode-publish.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+plugins {
+    id("java")
+    id("com.hedera.hashgraph.maven-publish")
+}
+
+publishing {
+    publications {
+        named<MavenPublication>("maven") {
+            pom.developers {
+                // TODO fill this
```
david-bakin-sl, line 26: TODO?

hedera-node/cli-clients/build.gradle.kts
```diff
@@ -15,7 +15,7 @@
  */
 
 plugins {
-    id("com.hedera.hashgraph.conventions")
+    id("com.hedera.gradle.services")
```
david-bakin-sl, line 18: Speaking of `cli-clients` - over at the issue this comment https://github.com/hashgraph/hedera-services/issues/11237#issuecomment-1970580413 says `cli-clients` won't be published.  I don't know the pros/cons - but why not?  These are useful tools (for developers), though not part of any running production code.  Or is _that_ the distinction?

hedera-node/hedera-evm/build.gradle.kts
```diff
@@ -15,12 +15,10 @@
  */
 
 plugins {
-    id("com.hedera.hashgraph.conventions")
-    id("com.hedera.hashgraph.evm-maven-publish")
+    id("com.hedera.gradle.evm")
```
david-bakin-sl, line 18: I'm not complaining about renaming this module, I think it's a good idea, but where'd it come from?  Is this part of the move to make it easier to get the EVM module over to the mirror node?  Or just consistency?  Or something else?

kimbor (approved):

david-bakin-sl (approved):
LGTM, only comment I'd like to seen an answer to is why `cli-clients` isn't published.  I don't know the issues - but those are tools useful to developers, so not knowing any reason to say no, I say instead: why not publish them?  Let me know.

build-logic/project-plugins/src/main/kotlin/com.hedera.gradle.blocknode-publish.gradle.kts
jjohannes: Will remove this in a follow up.

hedera-node/cli-clients/build.gradle.kts
jjohannes: The project is different in the sense that it does not provide a "library" following clear rules in what is packaged and which dependencies are published (and which module-info is published).
I agree that we can think about publishing this If we would do so, we have to think about the setup. Should it really be fat Jars? Or better follow the same approach as everywhere else to publish Jars with dependencies? How do we expect users to use this?
I'll take this with me to discuss as a follow up.

hedera-node/hedera-evm/build.gradle.kts
jjohannes: This is for consistency and backward compatibility. EVM is already published now with a different `group` and we want to keep that. (See other discussions with Nathan on this PR.)

PR #11742, opened 2024-02-26
perf: Minor performance improvements

netopyr:
This PR implements two small performance improvements:

- Cache index lookup
- Warm payer account


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #11743, opened 2024-02-26
fix: (mod) fix alias management

tinker-michaelj:
**Description**:
 - During BBM, add `alias-to-id` mappings for _both_ the in-state alias (which could be a "key alias") and any implied EVM address.
 - When checking if an EVM address is the "priority" address for an account, also detect the case the account's key alias maps to the given EVM address.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
@@ -62,6 +62,7 @@ import com.hedera.node.app.service.mono.state.virtual.UniqueTokenValue;
 import com.hedera.node.app.service.mono.state.virtual.entities.OnDiskAccount;
 import com.hedera.node.app.service.mono.state.virtual.entities.OnDiskTokenRel;
 import com.hedera.node.app.service.mono.utils.EntityNum;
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -341,6 +342,12 @@ public void migrate(@NonNull final MigrationContext ctx) {
                                         aliasesState
                                                 .get()
                                                 .put(new ProtoBytes(toAcct.alias()), toAcct.accountIdOrThrow());
+                                        if (toAcct.alias().toByteArray().length > 20) {
```
david-bakin-sl, line 345: I'd like to see the PR comment "During BBM, add alias-to-id mappings for both the in-state alias (which could be a "key alias") and any implied EVM address" as a code comment here.

david-bakin-sl (approved):
I added a suggestion to add a comment to the code.  Otherwise LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
mhess-swl: Sure, I'll make sure to add that in my next PR. We've got a clean run on CI so I'm going to merge this one if that's alright

PR #11755, opened 2024-02-26
feat: 11425: level by level reconnect

### LINKED ISSUE #11425: Level by level traversal with pauses to wait for confirmations

anthony-swirldslabs:
**Description**:
Implementing an optimization of the reconnect process where a teacher pauses after sending the entire level of a VirtualMap tree to a learner and awaits a response from the learner for the last node which status isn't clear to the teacher. Only after the status is reported (either directly for the node in question, or about one of its ancestors from which the status of this node can be inferred), will the teacher resume sending the next level of the tree.

This is done by splitting the `handleQueue` in the `VirtualTeacherTreeView` class into two `accumulatingHandleQueue` and `processingHandleQueue` queues, and flipping between them when it's time to move on to the next level.

The idea of the optimization is to prevent the teacher from sending nodes to the learner when the learner already knows the nodes. The original implementation goes ahead, iterates the entire tree, and sends whatever it finds based on the statuses that the leraner has managed to report so far, hoping that the learner keeps up with the teacher, sends responses about already received nodes, and hence probably making the teacher aware about whether they are clean or dirty. This fix tries to eliminate the probabilistic/optimistic nature of the process by instead ensuring that the learner has reported statuses of a prior level completely. This way, when the teacher is sending the next level, it already knows whether the ancestors of the current level nodes are clean or not, and hence it can skip sending clean subtrees altogether. This reduces the load on the disk and network (and RAM and CPU) on both the teacher and the learner. And this is believed to be the source of the performance boost with this fix.

**Related issue(s)**:

Fixes #11425

**Notes for reviewer**:
The performance improvement has been measured using the `ReconnectBench` introduced recently in `swirlds-benchmarks`, and the fix seems to improve the performance by about 4%. Here's the results obtained on a bare-metal Ubuntu server:

```
Without the fix:
Benchmark                 (keySize)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (teacherAddProbability)  (teacherModifyProbability)  (teacherRemoveProbability)  Mode  Cnt  Score   Error  Units
ReconnectBench.reconnect          8   1000000         100       5000000            32    9823452658           128                     0.05                        0.05                        0.05  avgt    7  6.888 ± 0.418   s/op


With the fix:
Benchmark                 (keySize)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (teacherAddProbability)  (teacherModifyProbability)  (teacherRemoveProbability)  Mode  Cnt  Score   Error  Units
ReconnectBench.reconnect          8   1000000         100       5000000            32    9823452658           128                     0.05                        0.05                        0.05  avgt    7  6.566 ± 0.397   s/op
```

So the improvement is (6.888-6.566)/6.888 = ~4.6% .

**Checklist**


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/VirtualTeacherTreeView.java
```diff
@@ -188,6 +251,11 @@ public void registerResponseForNode(final Long node, final boolean learnerHasNod
                 ? ConcurrentNodeStatusTracker.Status.KNOWN
                 : ConcurrentNodeStatusTracker.Status.NOT_KNOWN;
         nodeStatusTracker.set(node, status);
+        if (node == lastNodeAwaitingReporting) {
+            synchronized (node) {
+                node.notify();
```
imalygin, line 256: Shouldn't it be `notifyAll`?

anthony-swirldslabs: Currently, the code calling wait() on this object is single-threaded, so technically the notify() should suffice. However, I replaced it with a call to notifyAll() in the latest commit in case future implementations would make use of multiple threads.

imalygin (dismissed):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/VirtualTeacherTreeView.java
```diff
@@ -54,11 +54,52 @@ public final class VirtualTeacherTreeView<K extends VirtualKey, V extends Virtua
...
+
+    /** A queue to which we add items for the next level currently. */
+    private ConcurrentBitSetQueue accumulatingHandleQueue = new ConcurrentBitSetQueue();
+
+    /** A queue which we're processing currently (aka sending nodes to the learner). */
+    private ConcurrentBitSetQueue processingHandleQueue = new ConcurrentBitSetQueue();
+
+    /** Flip the queues by exchanging the accumulating and processing queues. */
+    private synchronized void flipQueues() {
+        final ConcurrentBitSetQueue temp = accumulatingHandleQueue;
```
artemananiev, line 69: An assertion like this would be helpful here:
```java
    assert processingHandleQueue.isEmpty();
```

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/VirtualTeacherTreeView.java
```diff
@@ -54,11 +54,52 @@ public final class VirtualTeacherTreeView<K extends VirtualKey, V extends Virtua
...
+     */
+    private volatile Long lastNodeAwaitingReporting = null;
+
+    /**
+     * A node status may become implicitly reported by a response about its ancestors,
+     * in which case we can retest the condition and go ahead and send the next level
+     * without waiting for this particular node to report its status since it's already known/inferred.
+     * So we use a timeout to recheck the condition periodically.
+     */
+    private static final long AWAIT_FOR_REPORT_TIMEOUT_MILLIS = 100;
```
artemananiev, line 96: 100ms is too coarse-grained. Since `wait()` is called in a loop, and total wait time is checked against MAX_TOTAL_... anyway, 1ms can be used here.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/VirtualTeacherTreeView.java
anthony-swirldslabs: Added.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/VirtualTeacherTreeView.java
anthony-swirldslabs: Done, and also decreased the MAX_ timeout to 1 second.

OlegMazurov (approved):

artemananiev (approved):

PR #11756, opened 2024-02-26
fix: 11750 Fixed synchronization in   `BreakableDataSource.saveRecords`

### LINKED ISSUE #11750: BrokenBuilder / BreakableDataSource synchronization needs to be improved
`BreakableDataSource` is test machinery to simulate data source (disk) problems during reconnects. Together with `BrokenBuilder` it's used to throw IO exceptions, when certain conditions are met. These conditions are set in the tests. Typical scenario is: configure the data source to fail (throw exceptions) on the first 4 flushes, then run the test 5 times, check that 4 times it failed and the last time it succeeded.

The exceptions are thrown from `saveRecords()` method, which is a part of `VirtualDataSource` interface. During reconnects, this method is called periodically by the hasher to flush received data to disk. Previously it was called by the hasher on the main hashing thread only, but in the new virtual hasher implementation it may now be called on different threads (not at the same time, though). However, the test classes haven't been updated for multi-threaded use. In particular, `numCalls`, `numCallsBeforeThrow`, `numTimesBroken`, `numTimesToBreak` and probable other fields in `BrokenBuilder` aren't synchronized in any way. It may result in dirty reads and, eventually, intermittent failures. I suspect test failures like described in #11298 and #11362 may be caused by this problem.

imalygin:
**Description**:

This PR fixed synchronization issues stemming from the fact that `saveRecords` method is no longer called from a single thread. 

**Related issue(s)**:

Fixes #11750 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/BreakableDataSource.java
+++ b/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/BreakableDataSource.java
@@ -54,14 +54,12 @@ public final class BreakableDataSource implements VirtualDataSource<TestKey, Tes
         if (builder.numTimesBroken < builder.numTimesToBreak) {
             // Syncronization block is not required here, as this code is never called in parallel
             // (though from different threads). `volatile` modifier is sufficient to ensure visibility.
-            if (builder.numTimesBroken < builder.numTimesToBreak) {
-                builder.numCalls += leaves.size();
-                if (builder.numCalls > builder.numCallsBeforeThrow) {
-                    builder.numCalls = 0;
-                    builder.numTimesBroken++;
-                    delegate.close();
-                    throw new IOException("Something bad on the DB!");
-                }
+            builder.numCalls += leaves.size();
+            if (builder.numCalls > builder.numCallsBeforeThrow) {
+                builder.numCalls = 0;
+                builder.numTimesBroken++;
+                delegate.close();
+                throw new IOException("Something bad on the DB!");
             }
         }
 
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/BreakableDataSource.java
```diff
@@ -52,12 +53,16 @@ public void saveRecords(
         final List<VirtualLeafRecord<TestKey, TestValue>> leaves = leafRecordsToAddOrUpdate.toList();
 
         if (builder.numTimesBroken < builder.numTimesToBreak) {
```
artemananiev, line 55: This is not performance critical code, so let's avoid double-check (anti-)pattern. Actually, no synchronization is needed here at all, if all BrokenBuilder fields are volatile

platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/reconnect/BrokenBuilder.java
```diff
@@ -32,9 +32,12 @@ public final class BrokenBuilder implements VirtualDataSourceBuilder<TestKey, Te
...
 
     int numCallsBeforeThrow = Integer.MAX_VALUE;
-    int numCalls = 0;
     int numTimesToBreak = 0;
-    int numTimesBroken = 0;
+
+    // the following fields are volatile to ensure visibility,
+    // as BreakableDataSource.saveRecords called from multiple threads.
+    volatile int numCalls = 0;
+    volatile int numTimesBroken = 0;
```
artemananiev, line 40: Why not just make all these 4 fields volatile? There is no need to actually synchronize anything in BreakableDataSource, as its `saveRecords()` may not be called on two threads in parallel. The only issue is variable visibility, and to fix it, it's enough to change all fields to volatiles.

imalygin: >Why not just make all these 4 fields volatile? 
Only `numTimesBroken` and `numCalls` are being changed. But if you see value in marking them all as `volatile`, I'll change that. 

>`saveRecords()`  may not be called on two threads in parallel.
Agreed. I'll add the comment about why the syncrhonized block is not required.

artemananiev (approved):

artemananiev (approved):

Issue #10931, opened 2024-01-11
Metadata HIPs


netopyr:
**Protobufs:**
- [x] #10932

**Implementation:** 
- [x] #10987 
- [x] https://github.com/hashgraph/hedera-services/issues/11067
- [x] #11065
- [x] #11101
- [x] #11353
- [x] #11102

**Questions:** 
- [x] Clarify how to remove keys

PR #11766, opened 2024-02-27
feat: Metadata HIPs

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/10931

**Protobuf changes** : https://github.com/hashgraph/hedera-protobufs/pull/326

- Updated `TokenCreateHandler` to include `metadata` and `metadataKey` fields for newly created token. And added validations.
- Updated `TokenUpdateHandler` to update `metadata` and `metadataKey` fields for existing token. And added validations.
- Updated `TokenGetInfohandler` to return information about `metadata` and `metadataKey` fields for existing token.
- Since we are currently depending on `feeSchedules.json` and `throttles.json` from `mono-service`, added a default `feeSchedules.json` and `throttles.json` mod-service can use under [hedera-app/src/main/resources/genesis](https://github.com/hashgraph/hedera-services/tree/ed2dba391ebda2ade60377f26657ce8733c0b9cd/hedera-node/hedera-app/src/main/resources/genesis). Also changed the defaults in `BootstrapConfig` and `hedera-node/configuration/dev/application.properties`
- Updated all `feeSchedules.json` under `configuration/` with the feeSchedules for `TokenUpdateNfts`
- Added a dynamic property [nfts.maxBatchSizeUpdate](https://github.com/hashgraph/hedera-services/blob/ed2dba391ebda2ade60377f26657ce8733c0b9cd/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TokensConfig.java#L55) with default value as 10 to validate the size of allowed NFTs that can be updated in a single transaction.
- Added a response code `MISSING_SERIAL_NUMBERS` needed for validating if serial numbers are empty in given transaction body
- All smart contract tests were failing since we added `METADATA_KEY` to `TokenKeyType`. Ignored that key in `HTSPrecompileResult.java`

**Testing:**
- Unit tests
- E2E tests
- JRS Tests - Passed https://swirldslabs.slack.com/archives/C03ELH5MUHK/p1710444111428079


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateValidator.java
```diff
@@ -62,6 +62,10 @@ public ValidationResult validateSemantics(
         if (op.hasMemo()) {
             context.attributeValidator().validateMemo(op.memo());
         }
+        // validate metadata
+        if (op.hasMetadata()) {
+            validator.validateTokenMetadata(op.metadata(), tokensConfig);
```
Neeharika-Sompalli, line 67: We could generalize `context.attributeValidator().validateMemo(` and re-use it here

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -92,6 +103,22 @@ private void tokenStringCheck(
...
+     */
+    private void tokenByteArrayCheck(
+            @Nullable final Bytes bytes, final int maxLen, @NonNull final ResponseCodeEnum onTooLong) {
+        if (bytes != null && bytes.length() > 0) {
+            final long length = bytes.length();
+            validateTrue(length <= maxLen, onTooLong);
+            // Remove this restriction once Mirror Node gets fixed to handle zero bytes correctly.
+            validateTrue(!containsZeroByte(bytes), INVALID_ZERO_BYTE_IN_STRING);
+        }
+    }
```
Neeharika-Sompalli, line 120: We could generalize `context.attributeValidator().validateMemo(` and re-use that here right ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,138 @@
...
+
+        // Wrap in Set to de-duplicate serial numbers
+        final var nftSerialNums = new ArrayList<>(new LinkedHashSet<>(op.serialNumbers()));
+        var nftCount = nftSerialNums.size();
+        validateTrue(nftCount <= nftStore.sizeOfState(), INVALID_TRANSACTION_BODY);
+        validateTrue(!nftSerialNums.isEmpty(), INVALID_TRANSACTION_BODY);
+        final var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
+        final var nftsAreEnabled = tokensConfig.nftsAreEnabled();
+        if (nftCount > 1) {
+            validateTrue(nftsAreEnabled, NOT_SUPPORTED);
```
steven-sheehy, line 100: Why would this check depend upon the nftCount? If NFTs are disabled it should be checked first unconditionally I would think.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -159,4 +191,18 @@ public static boolean isKeyRemoval(@NonNull final Key source) {
         requireNonNull(source);
         return IMMUTABILITY_SENTINEL_KEY.equals(source);
     }
+    /**
+     * Checks if the given Bytes contains a zero
+     * @param bytes the Bytes to check
+     * @return true Bytes contains zero, false otherwise
+     */
+    public boolean containsZeroByte(@NonNull final Bytes bytes) {
+        byte[] byteArray = bytes.toByteArray();
```
steven-sheehy, line 200: I'm not familiar with PBJ, but believe this will cause a new array allocation. If so, that would be very inefficient just to check for a certain character. Should use `getByte()` to check or see if there's an unsafe way to get access to the backing array.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -92,6 +103,22 @@ private void tokenStringCheck(
...
+     * @param maxLen the maximum number of UTF-8 bytes allowed
+     * @param onTooLong the response code to use if the bytes array is too long
+     */
+    private void tokenByteArrayCheck(
+            @Nullable final Bytes bytes, final int maxLen, @NonNull final ResponseCodeEnum onTooLong) {
+        if (bytes != null && bytes.length() > 0) {
+            final long length = bytes.length();
+            validateTrue(length <= maxLen, onTooLong);
+            // Remove this restriction once Mirror Node gets fixed to handle zero bytes correctly.
+            validateTrue(!containsZeroByte(bytes), INVALID_ZERO_BYTE_IN_STRING);
```
steven-sheehy, line 118: Remove this check. As the `INVALID_ZERO_BYTE_IN_STRING` error indicates this only applies if the field is a string. The mirror node supports zeros in byte arrays.

steven-sheehy (changes requested):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
thomas-swirlds-labs: Agreed 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,138 @@
...
+
+        final var validationResult = tokenUpdateNftValidator.validateSemantics(context, op);
+        final var token = validationResult.token();
+        final var nftStore = context.writableStore(WritableNftStore.class);
+
+        // Wrap in Set to de-duplicate serial numbers
+        final var nftSerialNums = new ArrayList<>(new LinkedHashSet<>(op.serialNumbers()));
+        var nftCount = nftSerialNums.size();
+        validateTrue(nftCount <= nftStore.sizeOfState(), INVALID_TRANSACTION_BODY);
+        validateTrue(!nftSerialNums.isEmpty(), INVALID_TRANSACTION_BODY);
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Added `nfts.maxBatchSizeUpdate` property and validated the max batch size. Thanks @steven-sheehy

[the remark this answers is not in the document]
Neeharika-Sompalli: Added `MISSING_SERIAL_NUMBERS`. Thanks @steven-sheehy

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,138 @@
...
+        final var txnBody = context.body();
+        final var op = txnBody.tokenUpdateNftsOrThrow();
+        final var tokenNftId = op.tokenOrThrow();
+
+        final var validationResult = tokenUpdateNftValidator.validateSemantics(context, op);
+        final var token = validationResult.token();
+        final var nftStore = context.writableStore(WritableNftStore.class);
+
+        // Wrap in Set to de-duplicate serial numbers
+        final var nftSerialNums = new ArrayList<>(new LinkedHashSet<>(op.serialNumbers()));
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
Neeharika-Sompalli: I don't think this check is needed here, since we are not creating any NFTs. Removed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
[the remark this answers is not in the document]
steven-sheehy: We should have an API like tokenwipe that already takes in a list of serials. Those APIs should also hopefully have such a limit property. We should be able to reuse that limit.

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/pbj/PbjConverter.java
```diff
@@ -766,6 +768,10 @@ public final class PbjConverter {
             case TRANSACTION_HAS_UNKNOWN_FIELDS -> ResponseCodeEnum.TRANSACTION_HAS_UNKNOWN_FIELDS;
             case ACCOUNT_IS_IMMUTABLE -> ResponseCodeEnum.ACCOUNT_IS_IMMUTABLE;
             case ALIAS_ALREADY_ASSIGNED -> ResponseCodeEnum.ALIAS_ALREADY_ASSIGNED;
+            case INVALID_METADATA_KEY -> ResponseCodeEnum.INVALID_METADATA_KEY;
+            case MISSING_TOKEN_METADATA -> ResponseCodeEnum.MISSING_TOKEN_METADATA;
+            case TOKEN_HAS_NO_METADATA_KEY -> ResponseCodeEnum.TOKEN_HAS_NO_METADATA_KEY;
+            case MISSING_SERIAL_NUMBERS -> ResponseCodeEnum.MISSING_SERIAL_NUMBERS;
```
steven-sheehy, line 774: Outside scope of ticket, but we could easily create a dynamic EnumMap on startup using by comparing by `Enum.name()` if all entries always have the same enum name. That would rid us of this brittle mapping.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,140 @@
...
+        updateNftMetadata(nftSerialNums, nftStore, token, op);
+    }
+
+    private void updateNftMetadata(
+            @NonNull final Set<Long> nftSerialNums,
+            @NonNull final WritableNftStore nftStore,
+            @NonNull final TokenID tokenNftId,
+            @NonNull final TokenUpdateNftsTransactionBody op) {
+        final var hasMetadata = op.hasMetadata();
+        final var metadata = hasMetadata ? op.metadata() : Bytes.EMPTY;
```
steven-sheehy, line 100: Is it possible to clear a `nft.metadata` or why would we not fail if metadata is empty? Is it because in the future we might allow updating of some other field so we want to support the no-op transaction of updating without a metadata?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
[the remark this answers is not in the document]
Neeharika-Sompalli: @steven-sheehy  We don't have a generic property to re-use. We have specific properties for Mint, wipe and burn .
`nfts.maxBatchSizeWipe`, `nfts.maxBatchSizeMint` and `nfts.maxBatchSizeBurn`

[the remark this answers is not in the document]
steven-sheehy: Makes sense. Thanks.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
steven-sheehy: As discussed offline, we should not allow clear since nft.metadata is required. We should throw if empty.

netopyr: Makes sense. Good catch!

thomas-swirlds-labs: Fixed. Thanks @steven-sheehy 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn);
+        final var op = txn.tokenUpdateNftsOrThrow();
+        validateTruePreCheck(op.hasToken(), INVALID_TOKEN_ID);
+        validateTrue(!op.serialNumbers().isEmpty(), MISSING_SERIAL_NUMBERS);
+        if (op.hasMetadata()) {
+            validateFalsePreCheck(Objects.equals(op.metadata(), Bytes.EMPTY), MISSING_TOKEN_METADATA);
+        }
```
steven-sheehy, line 75: Don't think this was what's discussed. If metadata wrapper is not present or if it's present and empty bytes it should throw an exception. This only handles latter.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn);
+        final var op = txn.tokenUpdateNftsOrThrow();
+        validateTruePreCheck(op.hasToken(), INVALID_TOKEN_ID);
+        validateTrue(!op.serialNumbers().isEmpty(), MISSING_SERIAL_NUMBERS);
+        if (op.hasMetadata()) {
+            validateFalsePreCheck(Objects.equals(op.metadata(), Bytes.EMPTY), MISSING_TOKEN_METADATA);
```
steven-sheehy, line 74: This should be MISSING_NFT_METADATA or whatever token mint throws if metadata is missing.

steven-sheehy (changes requested):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
netopyr: What usually happens if a user sends an update but no fields are set? Is it an error or a no-op?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
thomas-swirlds-labs: It throws INVALID_TOKEN_MINT_METADATA. Not sure this is correct here. 
Should we create a new response code ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
steven-sheehy: Think it's a no-op, but this one has only one field at this time. Later when more fields are added we can remove the validation. Think since there's one field it makes more sense to notify the users they're missing the metadata since there's no other possibility and they must've made a mistake. It also ensures mirror node doesn't duplicate logic to check for empty. It's also more performant since the current code is still looking up the nft even for empty metadata.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
netopyr: I assume we did not enforce it in Mono either, right? Let's create an issue and fix it after modularization. Don't want to break diff testing.

Neeharika-Sompalli: We did not enforce in mono also. 
Makes sense. Will do

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
Neeharika-Sompalli: Made this a no-op if metadata field is not set.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        final var tokenId = op.tokenOrThrow();
+
+        // Ensure that the token has metadataKey
+        final var tokenStore = context.readableStore(ReadableTokenStore.class);
+        final var token = getIfUsable(tokenId, tokenStore);
+        validateTrue(token.hasMetadataKey(), TOKEN_HAS_NO_METADATA_KEY);
+
+        if (!op.hasMetadata()) {
+            return;
+        }
```
steven-sheehy, line 99: q: If a user sets metadata to empty and serials to invalid values (-1, max long, etc) or non-existing serials, is it okay to not validate the serials and return success for no-op transaction?

netopyr: The same we do in similar cases elsewhere. I can check once I am back at my desk.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -72,6 +75,18 @@ public void validateTokenName(@Nullable final String name, @NonNull final Tokens
         tokenStringCheck(name, tokensConfig.maxTokenNameUtf8Bytes(), MISSING_TOKEN_NAME, TOKEN_NAME_TOO_LONG);
     }
 
+    /**
+     * Validates the token metadata, if it exists and is not too long. Since nft.metadata is required we should throw if empty.
```
steven-sheehy, line 79: Comment needs to be updated

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
Neeharika-Sompalli: I think we validate in other handlers. But we were thinking if we should lookup state. I will update to fail if the nft is not valid.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

steven-sheehy (dismissed):
LGTM

nathanklick (approved):

JeffreyDallas (approved):

hedera-node/hapi/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ description = "Hedera API"
 
 // Add downloaded HAPI repo protobuf files into build directory and add to sources to build them
 tasks.cloneHederaProtobufs {
-    branchOrTag = "add-pbj-types-for-state"
+    branchOrTag = "metadata-hips-646-657-765"
```
jsync-swirlds, line 28: I think these changes are already in `main`.

Neeharika-Sompalli: Yes agreed @jsync-swirlds. Thank you !
We should change develop to use `main` also. It has `add-pbj-types-for-state`. Will change in a different PR.

iwsimon (commented):
LGTM, need to fix failed unit tests.

hedera-node/hapi/build.gradle.kts
jsync-swirlds: If you change it to main in this PR, that will fix develop when this merges....

Neeharika-Sompalli: Changed Thanks @jsync-swirlds

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/fees/calculation/token/queries/GetTokenInfoResourceUsage.java
```diff
@@ -96,6 +96,8 @@ public FeeData usageGiven(final Query query, final Token token) {
                     .givenCurrentKycKey(token.hasKycKey() ? Optional.of(fromPbj(token.kycKey())) : Optional.empty())
                     .givenCurrentPauseKey(
                             token.hasPauseKey() ? Optional.of(fromPbj(token.pauseKey())) : Optional.empty())
+                    .givenCurrentPauseKey(
```
tinker-michaelj, line 99: ⚠️ This looks like a typo; should it it be deleted?

Neeharika-Sompalli: Should be `givenCurrentMetadataKey`, I will change it. Thanks @tinker-michaelj

Neeharika-Sompalli: Fixed. Thanks @tinker-michaelj

tinker-michaelj (approved):
LGTM, tyvm @Neeharika-Sompalli and @thomas-swirlds-labs ! 🎉

iwsimon (approved):
LGTM. Thanks @Neeharika-Sompalli .

steven-sheehy (dismissed):

hedera-node/configuration/dev/application.properties
```diff
@@ -8,4 +8,4 @@ staking.periodMins=1
 staking.fees.nodeRewardPercentage=10
 staking.fees.stakingRewardPercentage=10
 # Needed for Restart and Reconnect HapiTests that run many transactions of each type
-bootstrap.throttleDefsJson.resource=throttles-dev.json
+bootstrap.throttleDefsJson.resource=genesis/throttles-dev.json
```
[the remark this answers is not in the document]
Neeharika-Sompalli: @nathanklick This is the default file packaged with the jar.
Devops have this file under `hedera-node/configuration/` as usual under all environments.
We were still depending on `mono-service` file. So, changed it to a different location under `hedera-app`.

iwsimon (approved):
LGTM

nathanklick (approved):

tinker-michaelj (approved):
LGTM! 🥂

steven-sheehy (approved):

PR #11787, opened 2024-02-28
fix: 11636 VirtualHasher performance improvements

### LINKED ISSUE #11636: Reduce object allocations in the new virtual hasher

imalygin:
VirtualHasher performance improvements:

- updated chunk height to 3
- introduced `complete` method to `ChunkHashTask` to handle clean nodes optimally

**Description**:
VirtualHasher performance improvements:

- updated chunk height to 3
- introduced `complete` method to `ChunkHashTask` to handle clean nodes optimally

**Related issue(s)**:

Fixes #11636

### DIFF
--- a/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
+++ b/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
@@ -453,7 +453,8 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
                         assert siblingPath == 2;
                         parentTask.setHash((int) (siblingPath - firstSiblingPath), NULL_HASH);
                     } else if ((siblingPath < curPath) && !firstLeaf) {
-                        curTask.complete();
+                        // Mark the sibling as clean, reducing the number of dependencies
+                        parentTask.send();
                     } else {
                         // Get or create a sibling task
                         final int siblingHeight;
--- a/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
+++ b/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
@@ -412,14 +412,15 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
                         t.complete();
                         curStackPath++;
                     }
-                    /*
-                       It may happen that curPath is actually in the same chunk as lastPathInCurStackChunk.
-                       In this case, stack[curRank] should be set to curPath + 1 to prevent a situation in which all
-                       existing tasks between curPath and the end of the chunk will hang in the tasks map and will be
-                       processed only after the last leaf (in the loop to set null data for all tasks remaining in the map),
-                        despite these tasks being known to be clear.
-                    */
-                    if (curPath < lastPathInCurStackChunk) {
+
+                    //  It may happen that curPath is actually in the same chunk as stack[curRank].
+                    //  In this case, stack[curRank] should be set to curPath + 1 to prevent a situation in which all
+                    //  existing tasks between curPath and the end of the chunk will hang in the tasks map and will be
+                    //  processed only after the last leaf (in the loop to set null data for all tasks remaining in the
+                    // map),
+                    //   despite these tasks being known to be clear.
+
+                    if (curPath > curStackPath && curPath < lastPathInCurStackChunk) {
                         stack[curRank] = curPath + 1;
                     } else {
                         stack[curRank] = INVALID_PATH;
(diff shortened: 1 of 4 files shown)


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -145,51 +151,58 @@ public Hash hash(
...
+            super(pool, dependencies);
+            ins = numHashes > 0 ? new Hash[numHashes] : null;
+        }
+
+        @Override
+        protected boolean exec() {
+            return true;
+        }
+
+        void setHash(int index, Hash hash) {
```
artemananiev, line 169: Please, make all args final

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -334,7 +345,10 @@ public Hash hash(
         // the root task below. When the root task is done executing, that is it produced
         // a root hash, this hash is set as an input dependency for this result task, where
         // it's read and returned in the end of this method
-        ChunkHashTask resultTask = new ChunkHashTask(HASHING_POOL, INVALID_PATH, 1);
+        /**
+         * A task holding the resulting hash. Used to synchronize all parallel computations.
+         */
+        HashHoldingTask resultTask = new HashHoldingTask(HASHING_POOL, 1, 1);
```
artemananiev, line 351: final

OlegMazurov (dismissed):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
imalygin: Done

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
imalygin: Done

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -400,10 +409,21 @@ public Hash hash(
                     while (curStackPath < Math.min(curPath, lastPathInCurStackChunk)) {
                         final ChunkHashTask t = map.remove(curStackPath);
                         assert t != null;
-                        t.setData(null);
+                        t.complete();
                         curStackPath++;
                     }
-                    stack[curRank] = INVALID_PATH;
+                    /*
+                       It may happen that curPath is actually in the same chunk as lastPathInCurStackChunk.
```
artemananiev, line 416: "... in the same chunk as stack[curRank]"

OlegMazurov (dismissed):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
imalygin: Done

alittley (approved):
Approved WRT hashgraph owned files

artemananiev (dismissed):

alittley (approved):

OlegMazurov:
`VirtualMapBench` result in GCP:
```
Benchmark                             (keySize)   (maxKey)  (numFiles)  (numRecords)  (numThreads)  (recordSize)  Mode  Cnt      Score   Error  Units
CryptoBenchMerkleDb.transferPrefetch         24  500000000        6000        100000            32          1024  avgt       17921.621           s/op
```
Baseline score: 19066.948 (lower is better).

OlegMazurov (approved):

PR #11789, opened 2024-02-28
fix: fix ERC-20 log events and custom fee calculations

### LINKED ISSUE #11630: Replay events from round 162848838 and validate record files for differential testing - 15 min
Replay events from round 162848838 and validate record files for differential testing - 15 min

### LINKED ISSUE #11778: Missing token transfer record in a contract call.
### Description

The follow contract call [transaction](https://hashscan.io/mainnet/transaction/1708276974.036466003):
For body,
transactionID {
  transactionValidStart {
    seconds: 1708276958
    nanos: 334090293
  }
  accountID {
    accountNum: [4348507](https://hashscan.io/mainnet/account/0.0.4348507?app=false&p1=1&k1=1708824205.210638494&p2=1&k2=1708822193.613645003&ph=1&pt=1&kt=0.0.1739095)
  }
}
nodeAccountID {
  accountNum: 19
}
transactionFee: 200000000
transactionValidDuration {
  seconds: 120
}
contractCall {
  contractID {
    contractNum: [3696885](https://hashscan.io/mainnet/contract/0.0.3696885)
  }
  gas: 5000000
  functionParameters: ".\032}M\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\0003\202)(\310"
}

The #4 [crypto transfer](https://hashscan.io/mainnet/transaction/1708276974.036466007) record is missing.

cryptoTransfer {
  tokenTransfers {
    token {
      tokenNum: 1937609
    }
    transfers {
      accountID {
        accountNum: 4348507
      }
      amount: 221227067592
    }
    transfers {
      accountID {
        accountNum: 3696885
      }
      amount: -221227067592
    }
  }

Debug the code, the #4 crypto transfer was not called in CryptoTransferHandler.preHandle()/handle(). Maybe FrameRunner.runToCompletion() missed creating this in stack?

tinker-michaelj:
**Description**:
 - Closes #11630 
 - Fixes #11778
 - Fix two small mono-service issues that we have no reason to reproduce in mod-service:
    1. Use the correct resource prices for scheduled `CryptoTransfer`'s that trigger custom fee payments.
    2. Set the receiver log topic in an ERC-20 `TRANSFER_EVENT` even when the transferred amount is `0`.
 - Fix mod-service to only emit ERC-20 events on `SUCCESS`; and to still distinguish the logical sender and receiver when the amount is `0`.

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/StandardProcessLogic.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/StandardProcessLogic.java
@@ -33,8 +33,10 @@ import com.hedera.node.app.service.mono.stats.ExecutionTimeTracker;
 import com.hedera.node.app.service.mono.txns.ProcessLogic;
 import com.hedera.node.app.service.mono.txns.schedule.ScheduleProcessing;
 import com.hedera.node.app.service.mono.txns.span.ExpandHandleSpan;
+import com.hedera.node.app.service.mono.txns.span.SpanMapManager;
 import com.hedera.node.app.service.mono.utils.accessors.SwirldsTxnAccessor;
 import com.hedera.node.app.service.mono.utils.accessors.TxnAccessor;
+import com.hederahashgraph.api.proto.java.HederaFunctionality;
 import com.swirlds.platform.system.InitTrigger;
 import com.swirlds.platform.system.SoftwareVersion;
 import com.swirlds.platform.system.transaction.ConsensusTransaction;
@@ -63,6 +65,7 @@ public class StandardProcessLogic implements ProcessLogic {
     private final RecordStreaming recordStreaming;
     private final RecordCache recordCache;
     private final InitTrigger initTrigger;
+    private final SpanMapManager spanMapManager;
 
     @Inject
     public StandardProcessLogic(
@@ -79,7 +82,8 @@ public class StandardProcessLogic implements ProcessLogic {
             final RecordStreaming recordStreaming,
             final StateView workingView,
             final RecordCache recordCache,
-            @NonNull final InitTrigger initTrigger) {
+            @NonNull final InitTrigger initTrigger,
+            SpanMapManager spanMapManager) {
         this.expiries = expiries;
         this.invariantChecks = invariantChecks;
         this.expandHandleSpan = expandHandleSpan;
@@ -94,6 +98,7 @@ public class StandardProcessLogic implements ProcessLogic {
         this.workingView = workingView;
         this.recordCache = recordCache;
         this.initTrigger = requireNonNull(initTrigger);
+        this.spanMapManager = spanMapManager;
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


iwsimon (commented):
I tested on this branch that all issues of 11630 have been fixed. Thanks @tinker-michaelj ! There are some failed unit tests because HtsCall.execute() throws UnsupportedOperationException.

iwsimon (approved):
LGTM. Thanks @tinker-michaelj !

Neeharika-Sompalli (approved):

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/StandardProcessLogic.java
```diff
@@ -79,7 +82,8 @@ public StandardProcessLogic(
             final RecordStreaming recordStreaming,
             final StateView workingView,
             final RecordCache recordCache,
-            @NonNull final InitTrigger initTrigger) {
+            @NonNull final InitTrigger initTrigger,
+            SpanMapManager spanMapManager) {
```
david-bakin-sl, line 86: `@NonNull final`?  (See usage at L195)

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/span/SpanMapManager.java
```diff
@@ -318,7 +318,7 @@ private void rationalizeImpliedTransfers(final TxnAccessor accessor) {
         }
     }
 
-    private void expandImpliedTransfers(final TxnAccessor accessor) {
+    public void expandImpliedTransfers(final TxnAccessor accessor) {
```
david-bakin-sl, line 321: Needs a comment saying what "expanding implied transfers" really means.  Especially now that it's public.

david-bakin-sl: And now that it's public a unit test would be appropriate but probably you think it unnecessary.

Plus argument `accessor` needs a `@NonNull`

hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/store/contracts/precompile/codec/DecodingFacade.java
```diff
@@ -250,8 +250,9 @@ public static void addSignedAdjustment(
             final TokenID tokenType,
             final AccountID accountID,
             final long amount,
-            final boolean isApproval) {
-        if (amount > 0) {
+            final boolean isApproval,
+            final boolean zeroAmountIsReceiver) {
+        if (amount > 0 || (amount == 0 && zeroAmountIsReceiver)) {
```
david-bakin-sl, line 255: `zeroAmountIsReceiver` is an odd name; is this supposed to mean that the receiver account has zero balance?  (It's not that the 0 balance itself is the receiver ....)

david-bakin-sl (commented):
I guess I better finish this review now before it gets merged.

PR #11792, opened 2024-02-28
build: fix visibility of JMH classes in IDEA

jjohannes:
**Description**:

The fact that we used different Jars on the 'compile classpath' and 'runtime classpath' confuses IDEA. To prevent this, we now use, only for JMH, the non-module Jars consistently.

image

We might change this in the future to always use the Module Jar once the referenced Gradle issue is fixed.

**Additional chores**
- Remove `platform-sdk/swirlds-base/gradle.properties` - this file and the entry in it is not needed
- Update JMH to `1.37` (latest)

**Checklist**


timo0 (approved):
Works for me now

anthony-swirldslabs (approved):

jjohannes:
@tomzhenghedera or @leninmehedy can I get a ✅ for this IDE usability improvement.

jeromy-cannon (approved):

PR #11818, opened 2024-02-29
chore: Executor factory for the platform

### LINKED ISSUE #12023: ExecutorFactory as central place to create executors/threads/...
The draft https://github.com/hashgraph/hedera-services/pull/11818 introduces the `ExecutorFactory` that should be used in future to provide executors to all parts of the platform + services. As a next step we need a clean implementation of that factory as part of the base code.

hendrikebbers:
Fixes #12023 
Provide a executor factory api for platform code


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultForkJoinWorkerThread.java
```diff
@@ -0,0 +1,49 @@
...
+        setName(name);
+        this.onStartup = onStartup;
+    }
+
+    @Override
+    protected void onStart() {
+        if (onStartup != null) {
+            onStartup.run();
+        }
+        super.onStart();
```
poulok, line 47: The documentation says that `super.onStart()` must be called in this method FIRST

mxtartaglia-sl (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Create an ExecutorService with the given thread count.
+     *
+     * @param threadCount the thread count
+     * @return the ExecutorService
+     */
+    @NonNull
+    ExecutorService createExecutorService(int threadCount);
+
+    ScheduledExecutorService createScheduledExecutorService(int threadCount);
```
cody-littley, line 47: javadoc

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Create a not started thread with the given runnable.
+     *
+     * @param runnable the runnable
+     * @return the not started thread
+     * @deprecated all useage should be migrated to {@link #createExecutorService(int)}
+     */
+    @Deprecated
+    @NonNull
+    Thread createThread(@NonNull Runnable runnable);
```
cody-littley, line 58: Within the wiring framework, there is a type of task scheduler that wants to do its work on a long lived thread. In that use case, is the preferred solution to create a fixed executor service of size 1?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultExecutorFactory.java
```diff
@@ -0,0 +1,156 @@
...
+    @Override
+    public ForkJoinPool createForkJoinPool(int parallelism) {
+        if (parallelism <= 0) {
+            throw new IllegalArgumentException("parallelism must be greater than 0");
+        }
+        return new ForkJoinPool(parallelism, forkJoinWorkerThreadFactory, handler, false);
+    }
+
+    @Override
+    public ExecutorService createExecutorService(int threadCount) {
```
cody-littley, line 96: nullity annotations for return types within this class

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultForkJoinWorkerThreadFactory.java
```diff
@@ -0,0 +1,84 @@
...
+            @NonNull final ThreadGroup threadGroup,
+            @NonNull final Supplier<String> threadNameFactory,
+            @Nullable final Runnable onStartup) {
+        this.threadGroup = Objects.requireNonNull(threadGroup, "threadGroup must not be null");
+        this.threadNameFactory = Objects.requireNonNull(threadNameFactory, "threadNameFactory must not be null");
+        this.onStartup = onStartup;
+    }
+
+    @Override
+    public ForkJoinWorkerThread newThread(@NonNull final ForkJoinPool pool) {
```
cody-littley, line 69:
```suggestion
    @NonNull
    public ForkJoinWorkerThread newThread(@NonNull final ForkJoinPool pool) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultThreadFactory.java
```diff
@@ -0,0 +1,96 @@
...
+            @NonNull final UncaughtExceptionHandler exceptionHandler,
+            @Nullable final Runnable onStartup) {
+        this.group = Objects.requireNonNull(group, "group must not be null");
+        this.threadNameFactory = Objects.requireNonNull(threadNameFactory, "name must not be null");
+        this.exceptionHandler = Objects.requireNonNull(exceptionHandler, "exceptionHandler must not be null");
+        this.onStartup = onStartup;
+    }
+
+    @Override
+    public Thread newThread(@NonNull final Runnable innerRunnable) {
```
cody-littley, line 72:
```suggestion
    @NonNull
    public Thread newThread(@NonNull final Runnable innerRunnable) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * A factory for creating executors and threads.
+ */
+public interface ExecutorFactory {
```
cody-littley, line 27: Can we put this into the platform context as a part of this PR? Outside the context it's not very easy to use. Put one of these in the context and the entire platform will exclusively use this utility for creating threads/executors within 48 hours. 😄

cody-littley, line 27: Something worth considering is how we shut down the platform. It would be convenient if there were `shutdown()` and `join()` methods on this interface. In unit tests (in the long run), we should require that any executor factory is properly shut down and that we block on `join()` at the end of each test. I'm pretty sure we leak a ton of threads in our current suite of tests.

cody-littley, line 27: Can we put this into the platform context as a part of this PR? Not having it in the platform context probably means it either won't be adopted quickly (if at all).

anthony-swirldslabs (dismissed):
Looks good to me overall.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
hendrikebbers: For that it might make sense to define an Executor that can extend its thread pool. So instead of having a single-thread-executor instance for each long lived task we might have 1 executor for all. Or does long lived means forever?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
hendrikebbers: Let's do that once the PR is merged. But I'm with you: Should be part of platform context. But for that case we need to fix the parameters of the creation of the real instance. I want to discuss that a little more (especially with turtle in mind)

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultExecutorFactory.java
hendrikebbers: The interface already has that definitions.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/ExecutorFactory.java
hendrikebbers: Good point. I do not know if it is a good idea to have it on the interface. The platform should not be able to kill all threads. Whatever uses the platform should be able to do so. So it won't be on the interface but we need a place to do so. Created https://github.com/hashgraph/hedera-services/issues/12731

cody-littley: I agree that the platform doesn't need to shut down the threads. As long as when we shut down the platform we are able to also shut down the threads in a verifiable manner, I am ok with that.

cody-littley (dismissed):

anthony-swirldslabs (dismissed):

cody-littley (approved):

edward-swirldslabs (approved):

mxtartaglia-sl (approved):

jeromy-cannon (approved):

PR #11827, opened 2024-02-29
refactor: Flatten gossip class 11506

### LINKED ISSUE #11506: Flatten `Gossip` class
Currently, gossip is a monolith instance split across a few different classes. As a first step, we should flatten the structure to make is easier to work with.
NOTE: This ticket can be broken up into multiple if the work load ends up being large.

- `SingleNodeSyncGossip` should be removed. there is no need for a special case where the network has a single node. the `SyncGossip` class should be able to work with a single node, this should be tested.
- `Gossip`, `AbstractGossip` & `SyncGossip` should be flattened into a single class, this will make it easier to break apart later on
- anything related to unidirectional networks can be removed, since we don't support it anymore

kfa-aguda:
Closes #11506


lpetrovic05 (approved):
approved with the condition that single node network has been tested manually since we don't have automated tests for it

cody-littley (commented):
For this change, did you test a one node network? `SingleNodeSyncGossip` used to be required to get a one node network to function properly. (Feel free to ping me if you are unsure how to test this, or if it breaks and you are unsure how to debug the problem.)

PR #11836, opened 2024-03-01
chore: decouple socket factory

### LINKED ISSUE #11732: Decouple socket factory
SocketFactory and sub-classes should be refactored
- SocketFactory.createServerSocket should only have a port as input, the bind IP should be default
- The TlsFactory constructor should not have KeysAndCerts as input, instead it should have
  - Agreement private key
  - Agreement certificate
  - List of peers with public certificates to accept

### LINKED ISSUE #11835: Create PeerInfo record
In order to decouple the networking stack from the rest of the system, we need to create a record that will describe how to connect to a peer without having additional information.

lpetrovic05:
closes #11835
closes #11732


anthony-swirldslabs (dismissed):
Platform Data changes looks good to me.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -51,21 +53,29 @@ public class TlsFactory implements SocketFactory {
...
                     CertificateException, IOException {
-        Objects.requireNonNull(keysAndCerts);
-        Objects.requireNonNull(cryptoConfig);
+        Objects.requireNonNull(agrCert);
+        Objects.requireNonNull(agrKey);
+        Objects.requireNonNull(peers);
         this.socketConfig = Objects.requireNonNull(socketConfig);
+        Objects.requireNonNull(cryptoConfig);
+
+        // final KeyStore signingTrustStore = keysAndCerts.publicStores().sigTrustStore();
```
edward-swirldslabs, line 77: commented line of code remains.  Explain why it is there, or delete it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -51,21 +53,29 @@ public class TlsFactory implements SocketFactory {
...
+        final KeyStore signingTrustStore = CryptoStatic.createPublicKeyStore(peers);
```
edward-swirldslabs, line 78: Not a blocking comment, I assume this is where dynamism comes in?   The `signingTrustStore` is derived from the address book's public keys.  If the address book changes, the then this trust store changes when we rebuild the TLSFactory.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
lpetrovic05: left it there by mistake, removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
lpetrovic05: yes, the main goal here is to create the factory from the AB. in the future, when the AB changes, we will need to rebuild the factory I think. maybe its possible to just add another cert to the trust store, we will have to experiment to find out. but the important thing is that we don't depend on KeysAndCerts anymore

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/PeerInfo.java
```diff
@@ -0,0 +1,35 @@
...
+ * A record representing a peer's network information.
+ *
+ * @param nodeId             the ID of the peer
+ * @param nodeName           the name of the peer
+ * @param hostname           the hostname (or IP address) of the peer
+ * @param signingCertificate the certificate used to validate the peer's TLS certificate
+ */
+public record PeerInfo(
+        @NonNull NodeId nodeId,
+        @NonNull String nodeName,
```
cody-littley, line 33: What is a node name? Can we just use the node ID to uniquely identify the peer?

Issue #11569, opened 2024-02-15
Address Java Compiler warnings


nathanklick:
## Description

We need to cleanup the Java compiler warnings and then take steps to ensure future warnings are caught/resolved in a timely manner. 

### Example Warnings

```
warning: [module] module not found: com.hedera.node.app
```

PR #11838, opened 2024-03-01
build: (de)activate selection of javac lint features

jjohannes:
**Description**:
This PR activates most checks of `javac` as errors to run as part of `assemble`. For the active checks, this PR fixes the violations that were found.

The following checks are deactivated:

- **javac** - deactivated as we cannot use or do not want:  `module`, `serial`, `processing`, `try`, `missing-explicit-ctor`
- **javac** - deactivated because we use deprecations as a valid (intermediate) state:  `removal`, `deprecation`
- **javac** - possibly activate in the future, but deactivated for now because there are violations:
  - `this-escape` - calling public/protected method in constructor
  - `overrides` - overrides equals, but neither it ... overrides hashCode method
  - `unchecked` and `rawtypes` - this should be fixed (in my opinion), but there are some things off right now in very central interfaces like `Copyable` and `FastCopyable` and thus cross-cutting the whole code base
  - `exports` - validates that no internal types are used in the public interfaces in exported packages. I deactivated this individually in the projects that have violations right now.

See also comments in code for more details.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/11569

**Notes for reviewer**:

**Checklist**


build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.jpms-modules.gradle.kts
```diff
@@ -145,8 +144,13 @@ extraJavaModuleInfo {
     module("io.grpc:grpc-services", "grpc.services")
     module("io.grpc:grpc-protobuf", "grpc.protobuf")
     module("io.grpc:grpc-protobuf-lite", "grpc.protobuf.lite")
-    module("javax.annotation:javax.annotation-api", "java.annotation")
     module("com.github.spotbugs:spotbugs-annotations", "com.github.spotbugs.annotations")
+    module("com.google.code.findbugs:jsr305", "java.annotation") {
+        exportAllPackages()
+        mergeJar("javax.annotation:javax.annotation-api")
```
jjohannes, line 150: The changes here are necessary because the strict evaluation of annotation processors revealed some missing "compile time only" classes that annotation processors see.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/signed/SignedState.java
```diff
@@ -222,7 +222,7 @@ public long getRound() {
      * @return true if this is the genesis state
      */
     public boolean isGenesisState() {
-        return state.getPlatformState().getRound() == GENESIS_ROUND;
+        return state.getPlatformState().getRound() == PlatformState.GENESIS_ROUND;
```
cody-littley, line 225: Why was this change needed?

jjohannes: It's not needed. Thank you. Reverted.

hedera-node/test-clients/build.gradle.kts
```diff
@@ -23,6 +23,23 @@ plugins {
 
 description = "Hedera Services Test Clients for End to End Tests (EET)"
 
+sourceSets {
+    // Needed because "resource" directory is misnamed. See
+    // https://github.com/hashgraph/hedera-services/issues/3361
+    main { resources { srcDir("src/main/resource") } }
```
hendrikebbers, line 29: uff, @netopyr when can that be done?

hendrikebbers (approved):
Some great findings

hedera-node/cli-clients/src/main/java/com/hedera/services/cli/utils/Range.java
```diff
@@ -30,7 +30,7 @@
  *
  * <p>A range of integers specified by inclusive index at the low end, exclusive at the high end,
  * thus an empty range is from == to. (There is no check _on construction_ that this range is
- * "valid", i.e., 0 <= from, from <= to.)
+ * "valid", i.e., 0 &#60;= from, from &#60;= to.)
```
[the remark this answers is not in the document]
jjohannes: Maybe we can put this into an `{@code}` block instead?

build-logic/project-plugins/src/main/kotlin/com.hedera.gradle.services.gradle.kts
```diff
@@ -20,3 +20,11 @@ plugins {
 }
 
 group = "com.hedera.hashgraph"
+
+// TODO jjohannes: remove and fix issues in services code base
```
jjohannes, line 24: To reduce the numbers of required code owner reviews for this one, I will do the Javadoc fixes in the services code base as a follow up.

lpetrovic05 (approved):

hendrikebbers (commented):
For me it is really hard to review the PR since it does way to much. Can we maybe split this PR?

gradle/plugins/src/main/kotlin/com.hedera.gradle.root.gradle.kts
```diff
@@ -21,7 +21,7 @@ import net.swiftzer.semver.SemVer
 plugins {
     id("com.hedera.gradle.lifecycle")
     id("com.hedera.gradle.repositories")
-    id("com.hedera.hashgraph.nexus-publish")
+    id("com.hedera.gradle.nexus-publish")
```
jjohannes, line 24: Renamed for consistency (was missed in earlier PRs)

nathanklick (approved):

PR #11823, opened 2024-02-29
fix: fix schedules-by-equality migration, fractional custom fees, and special reward situations

tinker-michaelj:
**Description**:
 - Closes #11805 
 - Closes #11813
 - Migrate to correct key for schedule-by-equality map
 - Assess multiple fractional fees to a debit _always in terms of the initial debit_; and not in terms of what remained after applying earlier fractional fees.
 - Include `rewardRate` and `SUCCESS` status in `NodeStakeUpdateTransactionBody`.
 - Support two mono-service special cases for reward situations (on top-level `SUCCESS`): 
    1. Accounts in a `CryptoTransfer` transfer list with a `0` adjustment amount.
    2. Called contracts in a contract operation.

PR #11861, opened 2024-03-02
fix: match mono dissoc semantics

### LINKED ISSUE #11840: Match mono semantics for dissociation from deleted/expired tokens
### Problem

We need to implement [this TODO](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java#L264) for diff testing.

### Solution

Implement it.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #11840 
 - When a qualified contract uses `delegatecall` for an HTS system contract, use the frame's `receiver` address as the _authorizing_ sender, instead of the frame's EVM `sender`.
 - Strip out some confused and/or duplicate code throughout `HtsCall` implementations and `HtsSystemContract`.
 - Make `resultOfViewingToken()` return a `PricedResult` instead of a `FullResult`...the resulting design is problematic, but at least this provides the flexibility needed to simulate mono-service.
 - Fix `TokenDissociateFromAccountHandler` to match mono-service semantics for expired tokens.
 - Remove some redundant checks in the `AssociateTokenRecipientsStep` that should only be enforced later in `NFTOwnersChangeStep` for mono-service fidelity.
 - Don't check token pause status when computing custom fees.


Neeharika-Sompalli (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -609,6 +615,64 @@ private void handleUserTransaction(
         handleWorkflowMetrics.update(transactionInfo.functionality(), handleDuration);
     }
 
+    /**
```
lukelee-sl, line 618: Deja vu.  Were these change not also [here](https://github.com/hashgraph/hedera-services/pull/11823/files)?

tinker-michaelj: For `CryptoTransfer`'s with zero-adjustments, yes; this includes further cases.

lukelee-sl (commented):
LGTM.  I went once more through the logic for the exception contracts that are allowed to make a delegate transfer call and this seems sound to me.  The logic for staking contracts was also in the other pr so I believe I already looked at it.  Thanks for the explanation!

lukelee-sl (approved):

tinker-michaelj:
Superseded by https://github.com/hashgraph/hedera-services/pull/12006

Issue #11737, opened 2024-02-26
Fix failing upgrade JRS tests running against mod service


Neeharika-Sompalli:
Failing tests results https://swirldslabs.slack.com/archives/C03ELH5MUHK/p1708711884816649

- [x] Crypto-Update-Jar-1.5k-25m - `execute_immediate.mf` marker is not being detected by NMT 1.2.4
- [x] Crypto-Update-Setting-1.5k-15m - `execute_immediate.mf` marker is not being detected by NMT 1.2.4
- [x] Crypto-Update-Setting-Config-1.5k-25m - The correct `settings.txt` is not getting updated and stops at `Waiting for upgrade to finish to load new settings.txt`

Neeharika-Sompalli:
Now hitting deserialization failure
```
2024-02-29 20:35:43.566 17       ERROR EXCEPTION        <main> StreamDebugUtils: Deserialization failure. Will re-attempt deserialization with extra debug information.
com.swirlds.common.io.exceptions.ClassNotFoundException: No class with id=-7039950634678584517(0x9E4D12326FB27B3B) found
	at com.swirlds.common.io.streams.SerializableDataInputStream.registryConstructor(SerializableDataInputStream.java:453) ~[swirlds-common-0.80.0.jar:?]
	at com.swirlds.common.io.streams.SerializableDataInputStream.readSerializable(SerializableDataInputStream.java:163) ~[swirlds-common-0.80.0.jar:?]
	at com.swirlds.common.io.streams.SerializableDataInputStream.readSerializable(SerializableDataInputStream.java:88) ~[swirlds-common-0.80.0.jar:?]
	at com.swirlds.merkledb.MerkleDbTableConfig.deserialize(MerkleDbTableConfig.java:464) ~[swirlds-merkledb-0.80.0.jar:?]
	at com.swirlds.common.io.streams.SerializableDataInputStream.readSerializable(SerializableDataInputStream.java:167) ~[swirlds-common-0.80.0.jar:?]
	at com.swirlds.common.io.streams.SerializableDataInputStream.readSerializable(SerializableDataInputStream.java:108) ~[swirlds-common-0.80.0.jar:?]
	at com.swirlds.merkledb.MerkleDbDataSourceBuilder.deserialize(MerkleDbDataSourceBuilder.java:194) ~[swirlds-merkledb-0.80.0.jar:?]
[...]

PR #11884, opened 2024-03-04
chore: Fix issues related to upgrade

Neeharika-Sompalli:
Related to #11737 

- Fixes all issues related to Upgrade
- Adds and registers missing listeners needed from mono-service `ReconnectListener` and `WriteStateToDiskListener` 
- Writes `now_frozen` marker file needed for upgrade in `WriteStateToDiskListener` when freeze is submitted
- Changes the order when `MigrationContext.previousStates` is created to correctly identify a genesis restart or restart from saved state. Currently `MigrationContext.previousStates` is always empty
- Avoids writing null values for `freezeTime` and `updateFileHash` into state as that causes network to shut down, while hashing state
- Splits readable methods of `FreezeUpgradeActions` into `ReadableFreezeUpgradeActions`
- Fixes an issue to calculate totalBalance of all accounts in `InitialModServiceTokenSchema`
- Changes detecting genesis start by checking `previousVersion` instead of `MigrationContext.previousStates` is empty

**Testing :** 
Update tests pass except for query validator(not a blocker, the queries.log is empty) https://swirldslabs.slack.com/archives/C03ELH5MUHK/p1709775669765769


hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/PlatformStateAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.hedera.node.app.state;
+
+import com.swirlds.platform.state.PlatformState;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class PlatformStateAccessor {
```
mhess-swl, line 24: What is the point of wrapping the platform state with this class?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
```diff
@@ -89,7 +89,7 @@ public class MerkleSchemaRegistry implements SchemaRegistry {
     /**
      * The ordered set of all schemas registered by the service
      */
-    private final Set<Schema> schemas = new TreeSet<>();
+    private final SortedSet<Schema> schemas = new TreeSet<>();
```
mhess-swl, line 92: good call!

hedera-node/hedera-app/src/main/java/module-info.java
```diff
@@ -99,4 +99,7 @@
             com.swirlds.common,
             com.swirlds.platform;
     exports com.hedera.node.app.validation;
+    exports com.hedera.node.app.state.listeners to
+            com.hedera.node.app.test.fixtures,
+            com.swirlds.common;
```
jsync-swirlds, line 104: As best I can tell `com.swirlds.common` doesn't use anything from this module (which is good, given this module `requires transitive com.swirlds.common` on line 17).
We should remove all exports to `com.swirlds.common` here to eliminate the potential for confusion.
`com.swirlds.platform.core` appears to be the same.
Removing both of those significantly simplifies this file.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/PlatformStateAccessor.java
Neeharika-Sompalli: I just used the same one from mono-service. We need this to be easily injected by Dagger as far as I know.

hedera-node/hedera-app/src/main/java/module-info.java
Neeharika-Sompalli: Yes @jsync-swirlds . Thank you. Will change

Neeharika-Sompalli: Fixed. Thanks @jsync-swirlds

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/PlatformStateAccessor.java
mhess-swl: Ok. I prefer `Holder` instead of `Accessor`, but it's subjective and shouldn't block this PR

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.commons.io.FileUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Provides all the read-only actions that need to take place during upgrade
+ */
+public class ReadableFreezeUpgradeActions {
```
mhess-swl, line 50: Nit: `Readable` implies that this class can't do writes, but it writes marker files. Just a thought, but not a blocker

Neeharika-Sompalli: Ah yes, I just thought doesn't update state. Like we have `ReadableStores`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -897,6 +893,7 @@ public void onHandleConsensusRound(
             @NonNull final Round round, @NonNull final PlatformState platformState, @NonNull final HederaState state) {
         daggerApp.workingStateAccessor().setHederaState(state);
         daggerApp.handleWorkflow().handleRound(state, platformState, round);
+        daggerApp.platformStateAccessor().setPlatformState(platformState);
```
tinker-michaelj, line 896: Only the listener _currently_ needs access to `PlatformState` through this, right?

So it won't matter now if we set `platformState` before or after `handleRound()`.

But for clarity and just in case some handle component needs to use this accessor, setting the state first seems better.

Neeharika-Sompalli: Yes, makes sense. Fixed. Thanks @tinker-michaelj

mhess-swl (approved):
LGTM! This represents an insane amount of work and headache, well done @Neeharika-Sompalli! 🙌

mhess-swl (approved):

jsync-swirlds (approved):

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/WritableFreezeStore.java
```diff
@@ -87,14 +87,18 @@ public Timestamp lastFrozenTime() {
      *
      * @param updateFileHash The update file hash to set. If null, clears the update file hash.
```
kimbor, line 88: Need to update the javadoc here

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/WritableFreezeStore.java
```diff
@@ -57,9 +57,9 @@ public WritableFreezeStore(@NonNull final WritableStates states) {
      *
      * @param freezeTime the freeze time to set; if null, clears the freeze time
```
kimbor, line 58: Update javadoc

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/WritableFreezeStore.java
Neeharika-Sompalli: Thanks @kimbor . Will do in next PR

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/WritableFreezeStore.java
Neeharika-Sompalli: Thanks @kimbor . Will do in next PR

tinker-michaelj (approved):
LGTM, tyvm @Neeharika-Sompalli 🥂

Neeharika-Sompalli:
Update tests pass except for query validator(not a blocker, the queries.log is empty) https://swirldslabs.slack.com/archives/C03ELH5MUHK/p1709775669765769

JeffreyDallas (approved):

PR #11902, opened 2024-03-05
feat: Combine ISS detector inputs

### LINKED ISSUE #11901: Combine ISS detector inputs
- Currently, there are several input wires feeding data to the `IssDetector` from the same location

```
            issDetectorWiring.roundCompletedInput().put(roundNumber);
            issDetectorWiring.newStateHashed().put(state.getAndReserve("issDetector"));
            issDetectorWiring.handleConsensusRound().put(consensusRound);
```

These inputs should be combined into a single input.

alittley:
- closes #11901


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/iss/IssDetector.java
```diff
@@ -194,10 +210,44 @@ public void signalEndOfPreconsensusReplay(@Nullable final Object ignored) {
...
+     *
+     * @param stateAndRound the round and state to be handled
+     * @return a list of ISS notifications, or null if no ISS occurred
+     */
+    public @Nullable List<IssNotification> handleStateAndRound(@NonNull final StateAndRound stateAndRound) {
+        try (final ReservedSignedState state = stateAndRound.reservedSignedState(); ) {
+            final long roundNumber = stateAndRound.round().getRoundNum();
+            if (isRoundToIgnore(roundNumber)) {
+                return null;
+            }
```
cody-littley, line 237: I don't think the logic for ignored rounds is correct. If we want to ignore ISS detection during a specific round, it's still ok to collect signature transactions that reach consensus during that round. Also, we should still be shifting the window even if we want to ignore the round.

Would it be possible to implement this so that we do everything like normal when we are given a round we want to ignore except that we throw away ISS notifications for an ignored round? (regardless of the round during which the ISS notification was raised)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/iss/IssDetector.java
```diff
@@ -232,13 +282,21 @@ public void signalEndOfPreconsensusReplay(@Nullable final Object ignored) {
...
+     * @param round the round that may contain state signatures
+     * @return a list of ISS notifications, which may be empty
      */
-    public @Nullable List<IssNotification> handlePostconsensusSignatures(
-            @NonNull final List<ScopedSystemTransaction<StateSignatureTransaction>> transactions) {
-        return listOrNull(
-                transactions.stream().map(this::handlePostconsensusSignature).toList());
+    private @NonNull List<IssNotification> handlePostconsensusSignatures(@NonNull final ConsensusRound round) {
+        final List<ScopedSystemTransaction<StateSignatureTransaction>> stateSignatureTransactions =
+                new SystemTransactionExtractor<>(StateSignatureTransaction.class).handleRound(round);
```
cody-littley, line 290: SystemTransactionExtractor should just be a static function, instantiating a new object shouldn't be needed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/StateAndRound.java
```diff
@@ -26,4 +26,16 @@
...
+     * Make an additional reservation on the reserved signed state
+     *
+     * @param reservationReason the reason for the reservation
+     * @return this
+     */
+    public StateAndRound makeAdditionalReservation(@NonNull final String reservationReason) {
+        reservedSignedState.getAndReserve(reservationReason);
+
+        return this;
+    }
```
cody-littley, line 40: This is a bit of an odd pattern. The idea behind the state reservation pattern is that each reservation should have an object that tracks the reservation specifically. That way if we ever have a state leak, we can turn on the debug flag and see exactly which reservation we never released. With the pattern used here, we will see some `ReservedSignedState` objects that never get released, and others that get released multiple times. Even if the reference count ends up being the same, the debug trace will show something different.

A better way to do this would be to have this method return a copy of itself but with its own reservation.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/StateAndRound.java
```diff
@@ -26,4 +26,16 @@
...
+    /**
+     * Make an additional reservation on the reserved signed state
+     *
+     * @param reservationReason the reason for the reservation
+     * @return this
+     */
+    public StateAndRound makeAdditionalReservation(@NonNull final String reservationReason) {
+        reservedSignedState.getAndReserve(reservationReason);
+
+        return this;
```
cody-littley, line 39:
```suggestion
        return new StateAndRound(reservedSignedState.getAndReserve(reservationReason), round);
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/components/StateAndRound.java
```diff
@@ -26,4 +26,16 @@
...
  */
-public record StateAndRound(@NonNull ReservedSignedState reservedSignedState, @NonNull ConsensusRound round) {}
+public record StateAndRound(@NonNull ReservedSignedState reservedSignedState, @NonNull ConsensusRound round) {
+    /**
+     * Make an additional reservation on the reserved signed state
+     *
+     * @param reservationReason the reason for the reservation
+     * @return this
+     */
+    public StateAndRound makeAdditionalReservation(@NonNull final String reservationReason) {
```
cody-littley, line 36:
```suggestion
    @NonNull
    public StateAndRound makeAdditionalReservation(@NonNull final String reservationReason) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/iss/IssDetector.java
alittley: will be fixed in https://github.com/hashgraph/hedera-services/pull/11962

cody-littley (approved):

edward-swirldslabs (approved):

cody-littley (approved):

PR #11848, opened 2024-03-01
fix: fixed diff test issue 11822

iwsimon:
**Description**:

added checkTokenAllowances() in AssociateTokenRecipientsStep

**Related issue(s)**:

Fixes #11822

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #11907, opened 2024-03-05
fix: cherry-pick fixed diff test issue 11822

iwsimon:
**Description**:

Cherry pick fix from release/0.48 pr #11848
**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -637,4 +645,43 @@ private SubType getSubType(
         }
         return DEFAULT;
     }
+
+    private void validateTopLevelAllowances(
```
mhess-swl, line 649: This method looks like a good candidate to put in a validator, where it can have a bunch of unit tests around it

Neeharika-Sompalli: Agree with @mhess-swl . But not blocker for this PR merge

Neeharika-Sompalli (approved):

tinker-michaelj (approved):
LGTM, ty @iwsimon !

PR #11918, opened 2024-03-06
chore: update design doc for atomic crypto transfer

lukelee-sl:
**Description**:
The design doc for atomic crypto transfer included a positive test which is not possible.  Remove this test case.


netopyr (approved):

iwsimon (approved):
LGTM

Issue #11938, opened 2024-03-07
Disable fuzzy matching test


netopyr:
The test `LazyCreateThroughPrecompileSuite.revertedAutoCreationRollsBackEvenIfTopLevelSucceeds()` has fuzzy matching explicitly enabled. This has to be changed because we want to control fuzzy matching globally.

PR #11943, opened 2024-03-07
chore: Remove explicit fuzzy matching (cherry-pick)

netopyr:
Cherry-pick that removes the explicit fuzzy matching definition in `develop`.

Fixes https://github.com/hashgraph/hedera-services/issues/11938


Neeharika-Sompalli (approved):
LGTM !

PR #11958, opened 2024-03-07
fix: pbj-220: upgrade PBJ dependency to 0.7.23

anthony-swirldslabs:
**Description**:
This fix upgrades the PBJ dependency to v0.7.23 which contains a new implementation for a `null Bytes and Strings` fix where PBJ continues to accept `null` as an argument to model and Builder constructors, but internally ensures that model fields of these types are never `null` and instead get assigned default values (`Bytes.EMPTY` and `""` respectively), so as to prevent any NPEs when serializing such models and be compatible with Google Protobuf specification: https://github.com/hashgraph/pbj/pull/224

**Related issue(s)**:

Fixes #https://github.com/hashgraph/pbj/issues/220

**Notes for reviewer**:
PBJ v0.7.23 is released: https://central.sonatype.com/artifact/com.hedera.pbj/pbj-runtime/versions

**Checklist**

### DIFF
--- a/hedera-dependency-versions/build.gradle.kts
+++ b/hedera-dependency-versions/build.gradle.kts
@@ -58,7 +58,7 @@ moduleInfo {
     version("com.google.jimfs", "1.2")
     version("com.google.protobuf", protobufVersion)
     version("com.google.protobuf.util", protobufVersion)
-    version("com.hedera.pbj.runtime", "0.7.22")
+    version("com.hedera.pbj.runtime", "0.7.23")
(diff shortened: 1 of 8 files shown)


jeromy-cannon (dismissed):

nathanklick (approved):

Nana-EC (approved):

povolev15 (approved):
LGTM

iwsimon (approved):
LGTM

jeromy-cannon (approved):

jasperpotts (approved):

PR #11970, opened 2024-03-07
fix: avoid NPE in `ThrottlesManager` post-BBM

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/11981

Fix NPE in `ThrottlesManager` post BBM


tinker-michaelj (approved):

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceAdminSchema.java
```diff
@@ -61,15 +63,16 @@ public void migrate(@NonNull final MigrationContext ctx) {
...
-                ctx.newStates().<Timestamp>getSingleton(FreezeServiceImpl.LAST_FROZEN_TIME_KEY);
 
-        if (isGenesis) {
+        if (isGenesis || mnc != null) {
             upgradeFileHashKeyState.put(ProtoBytes.DEFAULT);
             freezeTimeKeyState.put(Timestamp.DEFAULT);
-            lastFrozenTimeKeyState.put(Timestamp.DEFAULT);
         }
 
         log.info("BBM: no migration actions necessary for admin service");
```
mhess-swl, line 72: Two things here:
1. Null the reference to the network state, so it can be released (critical). I.e. set `mnc = null;` 
2. We'll probably want to change this logging statement :)

mhess-swl (changes requested):
Looks good overall! Just need to null out the network context reference

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/schemas/InitialModServiceAdminSchema.java
Neeharika-Sompalli: Yes, thanks @mhess-swl . Fixed

mhess-swl (dismissed):

tinker-michaelj (approved):

PR #11972, opened 2024-03-07
chore: Minor clean up and fix failing Hapi tests

Neeharika-Sompalli:
- Removed all changes done in ``mono-service` files
- Since we depend on the `ApiPermissionConfig`, removed all changes to `api-permission.properties`
- Since we are currently depending on `feeSchedules.json` and `throttles.json` from `mono-service`, added a default `feeSchedules.json` and `throttles.json` mod-service can use under `hedera-app/src/main/resources/genesis` [here](https://github.com/hashgraph/hedera-services/tree/ed2dba391ebda2ade60377f26657ce8733c0b9cd/hedera-node/hedera-app/src/main/resources/genesis). Also changed the defaults in `BootstrapConfig` and `hedera-node/configuration/dev/application.properties`
- Added a dynamic property [nfts.maxBatchSizeUpdate](https://github.com/hashgraph/hedera-services/blob/ed2dba391ebda2ade60377f26657ce8733c0b9cd/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TokensConfig.java#L55) with default value as 10 to validate the size of allowed NFTs that can be updated in a single transaction.
- Added a response code `MISSING_SERIAL_NUMBERS` needed for validating if serial numbers are empty in given transaction body
- All smart contract tests were failing since we added `METADATA_KEY` to `TokenKeyType`. Ignored that key in `HTSPrecompileResult.java`


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateNftValidator.java
```diff
@@ -35,22 +26,4 @@ public class TokenUpdateNftValidator {
     public TokenUpdateNftValidator(@NonNull final TokenAttributesValidator validator) {
         this.validator = validator;
     }
-
```
thomas-swirlds-labs, line 38: We can probably do away with this class completely now ?

Neeharika-Sompalli: yes agreed 💯 . Will remove

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateNftsHandler.java
```diff
@@ -52,71 +54,58 @@
...
-        validateTrue(nftCount <= nftStore.sizeOfState(), INVALID_TRANSACTION_BODY);
-        validateTrue(!nftSerialNums.isEmpty(), INVALID_TRANSACTION_BODY);
-        final var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
-        final var nftsAreEnabled = tokensConfig.nftsAreEnabled();
-        if (nftCount > 1) {
-            validateTrue(nftsAreEnabled, NOT_SUPPORTED);
-        }
-        updateNftMetadata(nftSerialNums, nftStore, tokenNftId, op);
+        final var nftSerialNums = new LinkedHashSet<>(op.serialNumbers());
+        validateTrue(nftSerialNums.size() <= nftStore.sizeOfState(), INVALID_NFT_ID);
```
thomas-swirlds-labs, line 90: Should this be MISSING_SERIAL_NUMBERS ?

Neeharika-Sompalli: No, I think the serials provided are invalid if they are less what are in state. I used `MISSING_SERIAL_NUMBERS` when there are no serial numbers provided in transaction body. I am still not sure if we should fail there. But since we are charging based on serial numbers, I thought we should.

thomas-swirlds-labs (approved):
LGTM - Great job thank you! @Neeharika-Sompalli 🎉

Issue #11966, opened 2024-03-07
Revisit MerkleDbConfig initialization in MerkleDb classes


artemananiev:
During investigation of #11964 the following issue was found. In `MerkleDbTableConfig` there is a static field for `MerkleDbConfig`:

```java
    private static final MerkleDbConfig config = ConfigurationHolder.getConfigData(MerkleDbConfig.class);
```

This config is later used to initialize several fields to default values:

```java
    private long maxNumberOfKeys = config.maxNumOfKeys();
    private long hashesRamToDiskThreshold = config.hashesRamToDiskThreshold();
```

In reality, the `config` field is initialized before platform configuration is loaded. It means, if there is a `MerkleDbConfig` config set to a non-default value (e.g. in `settings.txt`), it will not be visible in `MerkleDbTableConfig`. There is a few other classes that use the same (anti-)pattern.

This fix will need a backport to 0.48

PR #11973, opened 2024-03-08
fix: 11966: Revisit MerkleDbConfig initialization in MerkleDb classes

artemananiev:
Fix summary:

* All static config fields are removed
* In some classes, they are replaced with non-static fields, which are initialized using `ConfigurationHolder`
* In other places, a config instance is passed as an arg to the constructor
* In a few places, where configs were used to create static thread pools / executors / etc., these objects are still left static, but initialized lazily
* A new unit test about `MerkleDbTableConfig` serialization is added

Fixes: https://github.com/hashgraph/hedera-services/issues/11966


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -62,25 +63,14 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
...
-     */
-    private static final int HASHING_THREAD_COUNT = CONFIG.getNumHashThreads();
-
     /**
      * This thread-local gets a HashBuilder that can be used for hashing on a per-thread basis.
      */
     private static final ThreadLocal<HashBuilder> HASH_BUILDER_THREAD_LOCAL =
             ThreadLocal.withInitial(() -> new HashBuilder(Cryptography.DEFAULT_DIGEST_TYPE));
 
+    private final VirtualMapConfig vmConfig = ConfigurationHolder.getConfigData(VirtualMapConfig.class);
```
poulok, line 72: Should this still be here?

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
```diff
@@ -54,31 +55,34 @@ class MerkleDbCompactionCoordinator {
...
-                new ThreadConfiguration(getStaticThreadManager())
-                        .setThreadGroup(new ThreadGroup("Compaction"))
-                        .setComponent(MERKLEDB_COMPONENT)
-                        .setThreadName("Compacting")
-                        .setExceptionHandler(
-                                (t, ex) -> logger.error(EXCEPTION.getMarker(), "Uncaught exception during merging", ex))
-                        .buildFactory());
+    static ExecutorService getCompactionExecutor() {
+        ExecutorService exec = compactionExecutor.get();
+        if (exec == null) {
```
anthony-swirldslabs, line 65: Minor: I hope there's not a thread race, today or in the future. I'd put `synchronized` on the method to be safe. While I see that you're calling compareAndSet() below, but then you have to shut down the newly created one. This seems like an extra complexity that could be avoided altogether. If the `synchronized` is being avoided for speed, then perhaps using the holder pattern here would be a better and cleaner choice (which in its simplest form would move the `synchronized{}` inside this method and use a double-check pattern.)

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -127,18 +128,28 @@ public class HalfDiskHashMap<K extends VirtualKey>
...
-                    .setComponent(MERKLEDB_COMPONENT)
-                    .setThreadName("HalfDiskHashMap Flushing")
-                    .setExceptionHandler((t, ex) ->
-                            logger.error(EXCEPTION.getMarker(), "Uncaught exception during HDHM flushing", ex))
-                    .buildFactory());
+    private static final AtomicReference<ExecutorService> flushExecutor = new AtomicReference<>();
+
+    private static ExecutorService getFlushExecutor() {
+        ExecutorService exec = flushExecutor.get();
+        if (exec == null) {
```
anthony-swirldslabs, line 136: Minor: same comment as above. Since we use this pattern in a few places, perhaps we could create an utility holder class that accepts a Supplier<> and implements this behavior in a good way, and then just reuse it.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -148,32 +148,38 @@ private static final class ClassVersion {
...
-                            .setComponent("virtual-map")
-                            .setThreadName("cache-cleaner")
-                            .setExceptionHandler((t, ex) -> logger.error(
-                                    EXCEPTION.getMarker(), "Failed to purge unneeded key/mutationList pairs", ex))
-                            .buildFactory());
+    private static final AtomicReference<Executor> cleaningPool = new AtomicReference<>();
+
+    private static Executor getCleaningPool() {
+        Executor exec = cleaningPool.get();
+        if (exec == null) {
```
anthony-swirldslabs, line 155: Minor: again same as above. I'd definitely recommend to create an utility holder class at this point as suggested above.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
```diff
@@ -110,7 +100,21 @@ public final class VirtualHasher<K extends VirtualKey, V extends VirtualValue> {
...
+    private static ForkJoinPool getHashingPool() {
+        ForkJoinPool pool = hashingPool.get();
+        if (pool == null) {
+            final VirtualMapConfig vmConfig = ConfigurationHolder.getConfigData(VirtualMapConfig.class);
+            final int hashingThreadCount = vmConfig.getNumHashThreads();
+            pool = new ForkJoinPool(hashingThreadCount);
+            if (!hashingPool.compareAndSet(null, pool)) {
+                pool.close();
+                pool = hashingPool.get();
+            }
```
anthony-swirldslabs, line 114: Minor: and this is #4 as above.

anthony-swirldslabs (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/hash/VirtualHasher.java
artemananiev: It is no longer static, so the original problem is not applicable here. However, I see your point. There is no need to have a field for this config, it can be queried as needed in the `hash()` method

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
artemananiev: You're right, it isn't synchronized because of `compareAndSet()`, and it's done for performance reasons. It's better to deal with very unlikely event of shutting down a redundant executor than to pay for synchronization on every call.

On the other hand... This method is called just a couple times for every flush, multiplied by the number of virtual maps. In other words, not performance critical at all. The chances are very low that the method is called in parallel from multiple threads, too.

@OlegMazurov What do you think? Should it be synchronized (and the field can be changed from an atomic reference to just a reference), or the current approach with CAS is better?

OlegMazurov: `volatile compactionExecutor` and double checking is the usual pattern here.

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -430,7 +442,7 @@ public DataFileReader<Bucket<K>> endWriting() throws IOException {
                     IntObjectPair<BucketMutation<K>> keyValue = iterator.next();
                     final int bucketIndex = keyValue.getOne();
                     final BucketMutation<K> bucketMap = keyValue.getTwo();
-                    flushExecutor.execute(() -> readUpdateQueueBucket(bucketIndex, bucketMap, queue));
+                    getFlushExecutor().execute(() -> readUpdateQueueBucket(bucketIndex, bucketMap, queue));
```
OlegMazurov, line 445: `getFlushExecutor()` call could be made into a `final` local variable outside the `while` loops. It could be just a `static synchronized` method then,

platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/files/DataFileReaderCloseTest.java
```diff
@@ -48,11 +48,12 @@ class DataFileReaderCloseTest {
...
     private static final DataItemSerializer<long[]> serializer = new TwoLongSerializer();
 
+    private static MerkleDbConfig dbConfig = ConfigurationHolder.getConfigData(MerkleDbConfig.class);
+
     @BeforeAll
     static void setup() throws IOException {
         final Path dir = TemporaryFileBuilder.buildTemporaryFile("readerIsOpenTest");
-        collection = new DataFileCollection<>(
-                ConfigurationHolder.getConfigData(MerkleDbConfig.class), dir, "store", serializer, null);
+        collection = new DataFileCollection<>(dbConfig, dir, "store", serializer, null);
```
OlegMazurov, line 56: Is it better to initialize `static dbConfig` here (even if for each test)?

OlegMazurov (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Done

platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/files/DataFileReaderCloseTest.java
artemananiev: Done

anthony-swirldslabs (approved):

PR #11974, opened 2024-03-08
Move remaining in-memory maps on disk

### LINKED ISSUE #11851: Move remaining inMemory Maps to onDisk
### Problem

We would like to deprecate the in memory map support at the platform level. Also the in memory map implementation is not as well optimized as on disk. Small maps will stay in cache even with on disk maps so will act like in memory maps. There are 3 places we still use in memory maps:

- Scheduled Transactions
- Topics
- Staking

### Solution

Ideally we will switch these over before 48 so we do not have to migrate twice.

### Alternatives

_No response_

netopyr:
This PR moves the remaining maps on disk.

Fixes #11851


Neeharika-Sompalli (dismissed):
LGTM !

Neeharika-Sompalli (approved):
LGTM !

PR #11995, opened 2024-03-08
fix: use Bytes instead of String for SCHEDULES_BY_EQUALITY_KEY

### LINKED ISSUE #11971: SCHEDULES_BY_EQUALITY_KEY should not be string key
### Description

The map in state for `SCHEDULES_BY_EQUALITY_KEY` should not by keyed by String. String just seems odd. Talking with Joseph it should be a hash, in which case it should be Bytes not String.

Also the migration code seems wrong as it is converting Long to String and storing that in there which will never work.
```
                schedulesByEquality.put(
                        ProtoString.newBuilder()
                                .value(String.valueOf(scheduleEqualityVirtualKey.getKeyAsLong()))
                                .build(),
                        ScheduleList.newBuilder().schedules(schedules).build());
```

### Steps to reproduce

Problem in code

### Additional context

_No response_

### Hedera network

other

### Version

0.0.48

### Operating system

None

kimbor:
**Description**:
Update the map in state for SCHEDULES_BY_EQUALITY_KEY to use Bytes instead of String for modularized code.

**Related issue(s)**:
Fixes #11971

### DIFF
--- a/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ScheduleStoreUtility.java
+++ b/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ScheduleStoreUtility.java
@@ -18,10 +18,10 @@ package com.hedera.node.app.service.schedule.impl;
 
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
-import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.Key;
 import com.hedera.hapi.node.scheduled.SchedulableTransactionBody;
 import com.hedera.hapi.node.state.schedule.Schedule;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.nio.charset.StandardCharsets;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ScheduleStoreUtility.java
```diff
@@ -59,13 +59,6 @@ private static void addToHash(final Hasher hasher, final Key keyToAdd) {
         hasher.putBytes(keyBytes);
     }
 
-    @SuppressWarnings("UnstableApiUsage")
```
povolev15, line 62: why do you delete it?

povolev15 (approved):
other than the comment LGTM

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ScheduleStoreUtility.java
kimbor: It wasn't being used anywhere. Since it uses AccountID, it doesn't really belong in schedule service anyway. My guess is that it was only there due to a copy-pasta mistake.

tinker-michaelj (approved):
LGTM, ty @kimbor !

Issue #11996, opened 2024-03-08
The fix for #11498 doesn't cover generic object keys


artemananiev:
This is a follow-up for #11498. In that fix, a new flag was added to `VirtualDataSource.saveRecords()` to indicate whether the method is called during data flushes, or from `VirtualNodeRemover` during reconnects. The flag is then used to alternate between delete() and deleteIfEqual().

However, this fix only covers long keys. Generic object keys are still deleted using `deleteIfEqual()`:

```java
            if (isLongKeyMode) {
                final long key = ((VirtualLongKey) leafRecord.getKey()).getKeyAsLong();
                if (isReconnect) {
                    longKeyToPath.putIfEqual(key, path, INVALID_PATH);
                } else {
                    longKeyToPath.put(key, INVALID_PATH);
                }
            } else {
                objectKeyToPath.deleteIfEqual(leafRecord.getKey(), path);
            }
```

This may break data flushes and result in stale keys in the database.

PR #11998, opened 2024-03-08
fix: 11996: The fix for #11498 doesn't cover generic object keys

artemananiev:
Fix summary: extended the original fix for #11498 to cover generic (object) keys in addition to long keys. A new unit test (actually, a regression test) is provided.

Fixes: https://github.com/hashgraph/hedera-services/issues/11996

### DIFF
--- a/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/map/MapTest.java
+++ b/platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/map/MapTest.java
@@ -19,6 +19,8 @@ package com.swirlds.virtual.merkle.map;
 import static com.swirlds.common.test.fixtures.junit.tags.TestQualifierTags.TIME_CONSUMING;
 import static com.swirlds.common.test.fixtures.junit.tags.TestQualifierTags.TIMING_SENSITIVE;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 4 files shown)


OlegMazurov (approved):
Very good addition to testing.

imalygin (approved):

platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/map/MapTest.java
```diff
@@ -104,4 +123,51 @@ void insertRemoveAndModifyOneMillion() throws InterruptedException {
...
+        rootNode = map.getRight();
+        rootNode.enableFlush();
+
+        copy = map.copy();
+        map.release();
+        map = copy;
+        rootNode.waitUntilFlushed();
+
+        // During this second flush, key/value 0 must be deleted from the map despite it's
+        // path the virtual tree doesn't match the path in the data source
```
anthony-swirldslabs, line 167: Nit: this might need rephrasing: "...despite it's path the virtual..."

anthony-swirldslabs (approved):

platform-sdk/swirlds-merkle/src/test/java/com/swirlds/virtual/merkle/map/MapTest.java
artemananiev: Looks like "in" is missed...

PR #11914, opened 2024-03-05
feat: pbj-212: upgrade PBJ dependency to 0.8.1

anthony-swirldslabs:
**Description**:
PBJ v0.8.1 modifies the `Codec` interface in order to support passing the `maxDepth` parameter to the `parse()` method which limits the maximum depth of nested messages that can be parsed without throwing a `ParseException`. This change also makes the `default Codec.parseStrict()` delegate to the new `parse()` method, so explicit `parseStrict()` implementations can now be removed. See https://github.com/hashgraph/pbj/pull/219 for details.

There's a few implementations of the `Codec` interface in hedera-services and they need to be updated together with upgrading the PBJ dependency to 0.8.1. This is what this fix does.

PBJ v0.8.1 has already been released: https://central.sonatype.com/artifact/com.hedera.pbj/pbj-runtime/versions

**Related issue(s)**:

Fixes #https://github.com/hashgraph/pbj/issues/212

**Notes for reviewer**:
Compilation passes. Related unit tests seem to pass as well.

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #12003, opened 2024-03-09
feat(pbj): upgrade PBJ dependency to 0.8.1

### LINKED ISSUE #12002: Back-port PBJ enhancements to 0.48
To backport PBJ `maxSize`, `null`, and `maxDepth` fixes from `develop` to `0.48`.

Essentially, we need to cherry-pick these two commits:
* https://github.com/hashgraph/hedera-services/commit/173f419619af3a289d2be114568e3665d0416827
* https://github.com/hashgraph/hedera-services/commit/a3b64270684b0d10081a2c3420442b9f82ddbc14

anthony-swirldslabs:
**Description**:
Back-porting PBJ enhancements introduced in versions up to v0.8.1, specifically:
1. `maxSize` where PBJ limits the maximum size of a message that it would parse w/o exceptions
2. `null` where PBJ continues to accept `null` values for `Bytes` and `String` types as input to model and Builder constructors, but internally ensures that all model fields of these types are assigned default values instead of the `null` - `Bytes.EMPTY` and `""` respectively
3. `maxDepth` where PBJ allows clients to limit the maximum depth of nested messages that it would parse.

Essentially, I'm cherry-picking the following two commits/pull-requests from `develop`:
* https://github.com/hashgraph/hedera-services/commit/173f419619af3a289d2be114568e3665d0416827 - https://github.com/hashgraph/hedera-services/pull/11958
* https://github.com/hashgraph/hedera-services/commit/a3b64270684b0d10081a2c3420442b9f82ddbc14 - https://github.com/hashgraph/hedera-services/pull/11914

Note that the `maxSize` fix didn't require any changes in the `hedera-services` other than upgrading the PBJ dependency version, so there's not a separate commit to back-port - the v0.8.1 contains that fix already, and that's all that is needed. The commits backported here cover changes related to the `null` and the `maxDepth` changes which required modifications in `hedera-services`.

**Related issue(s)**:

Fixes #12002

**Notes for reviewer**:
This is a verbatim back-porting of the aforementioned commits w/o any additional changes. Please refer to the pull-requests linked above for any additional details if needed.

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
@@ -116,8 +116,9 @@ public final class StreamFileProducerConcurrent implements BlockRecordStreamProd
                 throw new IllegalStateException("initRunningHash() can only be called once");
             }
 
-            if (runningHashes.runningHash() == null) {
-                throw new IllegalArgumentException("The initial running hash cannot be null");
+            if (runningHashes.runningHash() == null
+                    || runningHashes.runningHash().equals(Bytes.EMPTY)) {
+                throw new IllegalArgumentException("The initial running hash cannot be null or empty");
             }
 
             lastRecordHashingResult = completedFuture(runningHashes.runningHash());
--- a/hedera-dependency-versions/build.gradle.kts
+++ b/hedera-dependency-versions/build.gradle.kts
@@ -58,7 +58,7 @@ moduleInfo {
     version("com.google.jimfs", "1.2")
     version("com.google.protobuf", protobufVersion)
     version("com.google.protobuf.util", protobufVersion)
-    version("com.hedera.pbj.runtime", "0.7.19")
+    version("com.hedera.pbj.runtime", "0.7.23")
     version("com.squareup.javapoet", "1.13.0")
     version("com.sun.jna", "5.12.1")
     version("dagger", daggerVersion)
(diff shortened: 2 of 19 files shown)


Nana-EC (approved):

nathanklick (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/StreamFileProducerConcurrent.java
```diff
@@ -116,8 +116,9 @@ public void initRunningHash(@NonNull final RunningHashes runningHashes) {
                 throw new IllegalStateException("initRunningHash() can only be called once");
             }
 
-            if (runningHashes.runningHash() == null) {
-                throw new IllegalArgumentException("The initial running hash cannot be null");
+            if (runningHashes.runningHash() == null
```
tinker-michaelj, line 119: If it should be impossible for `runningHash()` to return `null` I would strongly prefer we delete this check.

anthony-swirldslabs: Do you suggest to do that in this back-port to `0.48`, or in a future fix in `develop`?

tinker-michaelj: `develop` is fine with me. 💯

anthony-swirldslabs: I filed https://github.com/hashgraph/hedera-services/issues/12031 to address this in `develop`.

PR #12014, opened 2024-03-11
chore: add JMH event serialization benchmark

### LINKED ISSUE #11864: Create a JMH serialization test
In order to track any potential performance regression when moving events to protobuf, create a JMH test to set a performance baseline. The test should include at least 1 system transaction.

lpetrovic05:
closes #11864

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventSerialization.java
+++ b/platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventSerialization.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventSerialization.java
```diff
@@ -0,0 +1,81 @@
...
+                .setNumberOfSystemTransactions(1)
+                .setSeed(seed)
+                .buildEvent();
+        StaticSoftwareVersion.setSoftwareVersion(event.getHashedData().getSoftwareVersion());
+        ConstructableRegistry.getInstance().registerConstructables("com.swirlds.platform.system");
+        final PipedInputStream inputStream = new PipedInputStream();
+        final PipedOutputStream outputStream = new PipedOutputStream(inputStream);
+        outStream = new MerkleDataOutputStream(outputStream);
+        inStream = new MerkleDataInputStream(inputStream);
+    }
```
cody-littley, line 68: I wonder if it might be more accurate to have two benchmarks: one that writes to a byte array output stream, and one that reads from one. Not sure if pipes are always going to be low overhead compared to the code.

cody-littley (approved):
Approved with an optional suggestion.

alittley (approved):

platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventSerialization.java
lpetrovic05: it might be better, but there are so many things related to events that need to be done that I would rather not waste more time

PR #12036, opened 2024-03-11
fix: remove null checks for non-nullable runningHash()

### LINKED ISSUE #12031: Clean up null checks for fields that can no longer be null
Recent PBJ v0.8.1 improvements made it impossible for Bytes and Strings fields to be null. The fix was back=ported to 0.48 at https://github.com/hashgraph/hedera-services/pull/12003 . The fix updated the existing `if (null)` conditions with `|| Bytes.EMPTY` and similar.

A review comment in that PR suggested that existing `null` checks should be cleaned up now that it's impossible for the fields to be null.

This issue tracks the removal of the legacy `null` checks.

anthony-swirldslabs:
**Description**:
As suggested at https://github.com/hashgraph/hedera-services/pull/12003 , I'm removing redundant `null` checks for the `runningHash()` values because with the latest PBJ (v0.7.23+), this field of type `Bytes` can never be `null`.

**Related issue(s)**:

Fixes #12031

**Notes for reviewer**:
None.

**Checklist**


tinker-michaelj (approved):
LGTM, ty @anthony-swirldslabs !

Issue #11507, opened 2024-02-13
Unit Test Failure: Stabilize VirtualMap reconnect tests


imalygin:
### Description

The following tests are failing intermittently: 

```
VirtualMapReconnectTest » teacherAbortsReconnectOnFirstInternal

VirtualMapLargeReconnectTest » multipleAbortedReconnectsCanSucceed(int, int, int, int)[1]
VirtualMapLargeReconnectTest » multipleAbortedReconnectsCanSucceed(int, int, int, int)[3]
```

https://scans.gradle.com/s/p7b5275t7a4o2
https://scans.gradle.com/s/d6alj3i3wcreg

### Steps to reproduce

Run abovementioned tests. 

### Hedera network

other

### Version

v0.48

### Operating system

None

imalygin:
Unfortunately, the issue reoccurred: 

See [here ](https://scans.gradle.com/s/apc5p6355msvm/console-log/task/:swirlds-merkle:timingSensitive?anchor=217&page=1) 
Failed on this PR - https://github.com/hashgraph/hedera-services/pull/11798/files

imalygin:
This test had multiple reasons to fail. 
This one

https://scans.gradle.com/s/d6alj3i3wcreg/tests/task/:swirlds-merkle:timingSensitive/details/com.swirlds.virtual.merkle.reconnect.VirtualMapLargeReconnectTest/multipleAbortedReconnectsCanSucceed(int%2C%20int%2C%20int%2C%20int)%5B1%5D?top-execution=1

is fixed by this PR https://github.com/hashgraph/hedera-services/pull/11718

This one 

https://scans.gradle.com/s/d6alj3i3wcreg/tests/task/:swirlds-merkle:timingSensitive/details/com.swirlds.virtual.merkle.reconnect.VirtualMapLargeReconnectTest/multipleAbortedReconnectsCanSucceed(int%2C%20int%2C%20int%2C%20int)%5B1%5D?top-execution=1

requires further investigation. During the reconnect a teacher suddenly receives `SocketClosedException`:
```
    Caused by: com.swirlds.common.merkle.synchronization.utility.MerkleSynchronizationException: Synchronization failed with exceptions	
    	at com.swirlds.common.merkle.synchronization.TeachingSynchronizer.sendTree(TeachingSynchronizer.java:199)	
    	at com.swirlds.common.merkle.synchronization.TeachingSynchronizer.synchronize(TeachingSynchronizer.java:136)	
[...]

PR #12039, opened 2024-03-11
fix: 11507 Catch possible exception thrown by `PairedStreams.close`

imalygin:
**Description**:

In some cases, there may be unfortunate circumstances under which PairedStreams have to deal with underlying streams that are already closed before the PairedStreams.close call. The test shouldn't fail in this case.

**Related issue(s)**:

Fixes (hopefully) #11507 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
+++ b/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
@@ -78,8 +78,7 @@ public class MerkleBenchmarkUtils {
             final Configuration configuration,
             final ReconnectConfig reconnectConfig)
             throws Exception {
-        PairedStreams streams = new PairedStreams(configuration.getConfigData(SocketConfig.class));
-        try {
+        try (PairedStreams streams = new PairedStreams(configuration.getConfigData(SocketConfig.class))) {
             final LearningSynchronizer learner;
             final TeachingSynchronizer teacher;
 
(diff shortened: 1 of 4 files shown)


jsync-swirlds (dismissed):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
```diff
@@ -137,6 +138,13 @@ private static <T extends MerkleNode> T testSynchronization(
 
             final MerkleNode generatedTree = learner.getRoot();
             return (T) generatedTree;
+        } finally {
+            try {
+                streams.close();
+            } catch (IOException e) {
+                // test code, no danger
+                e.printStackTrace();
+            }
```
anthony-swirldslabs, line 147: Similar to a prior fix surrounding stream.close(), I'd really prefer the case of double-closing to be handled in the stream implementation, rather than being worked around everywhere we deal with streams.

Note that I still want exceptions unrelated to double-closing to be revealed in tests.

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/MerkleBenchmarkUtils.java
imalygin: Fair enough. I changed the fix.

anthony-swirldslabs (approved):

PR #12041, opened 2024-03-11
feat: 12040 logging framework: NotNull & Nullable changes for consistency

### LINKED ISSUE #12040: Multiple Log API + Tests Enhancements

mxtartaglia-sl:
Fixes #12040

### DIFF
--- a/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggerImpl.java
+++ b/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggerImpl.java
@@ -76,26 +76,18 @@ public class LoggerImpl implements Logger {
      * @param marker           the initial marker of the logger (if present)
      * @param context          the initial context of the logger
      * @param logEventConsumer the consumer that is used to consume the log events
-     * @throws NullPointerException if the logEventConsumer is null. For all other use cases fallbacks are implemented
+     * @throws NullPointerException if the logEventFactory logEventConsumer or name is null.
      */
     protected LoggerImpl(
             @NonNull final String name,
             @Nullable final Marker marker,
-            @NonNull final Map<String, String> context,
+            @Nullable final Map<String, String> context,
             @NonNull final LogEventFactory logEventFactory,
             @NonNull final LogEventConsumer logEventConsumer) {
-        if (name == null) {
-            EMERGENCY_LOGGER.logNPE("name");
-            this.name = "";
-        } else {
-            this.name = name;
-        }
+        this.name = Objects.requireNonNull(
+                name, "name must not be null"); // Callers of this method need to make sure this is never possible
         this.marker = marker;
-        if (context == null) {
-            this.context = Collections.emptyMap();
-        } else {
-            this.context = Collections.unmodifiableMap(context);
-        }
+        this.context = (context == null) ? Collections.emptyMap() : Collections.unmodifiableMap(context);
         this.logEventFactory = Objects.requireNonNull(logEventFactory, "logEventFactory must not be null");
         this.logEventConsumer = Objects.requireNonNull(logEventConsumer, "logEventConsumer must not be null");
     }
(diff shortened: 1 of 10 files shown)


platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
```diff
@@ -120,9 +118,7 @@ public void addHandler(@NonNull final LogHandler handler) {
      * @param handler the handler to remove
      */
     public void removeHandler(@NonNull final LogHandler handler) {
-        if (handler == null) {
-            EMERGENCY_LOGGER.logNPE("handler");
-        } else {
+        if (!LoggerUtils.checkIsNull(handler, "handler")) {
```
hendrikebbers, line 121: I can not find LoggerUtils. Looks like that is new but it is not part of the PR

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/DefaultLoggingSystem.java
```diff
@@ -121,13 +115,21 @@ private static Configuration createConfiguration() {
...
-     * Returns the singleton instance.
+     * Returns the singleton instance. Initializes it if it wasn't initialized already
      *
      * @return The singleton instance.
      */
     @NonNull
     public static DefaultLoggingSystem getInstance() {
-        return InstanceHolder.INSTANCE;
+        // Double-checked locking for thread safety
+        if (instance == null) {
```
hendrikebbers, line 125: While I'm fine with this and many other changes in the PR I would say that mixing that with the test changes makes the PR too complex to review. Let's split it up in 2 please

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggerImpl.java
```diff
@@ -81,21 +74,12 @@ public class LoggerImpl implements Logger {
...
+            @Nullable final Map<String, String> context,
             @NonNull final LogEventFactory logEventFactory,
             @NonNull final LogEventConsumer logEventConsumer) {
-        if (name == null) {
-            EMERGENCY_LOGGER.logNPE("name");
-            this.name = "";
-        } else {
-            this.name = name;
-        }
+        this.name = Objects.requireNonNull(name, "name must not be null");
```
hendrikebbers, line 80: Please do not do that. This would end in throwing an exception at logger creation. The logging API should never throw an exception when used.

hendrikebbers (commented):
Let's split this in 2 PRs: 1 for changes in the logging API and 1 for changing the tests. I would like to discuss the second one (how we change tests to produce less system out)

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggerImpl.java
mxtartaglia-sl: Perfect one discussion topic I was interested in having:
It makes sense, but what is the difference with the other parameters on the method, e.g., `logEventFactory` and `logEventConsumer`? Can't we assume that if those will not be null (because they need to be handled beforehand and sent a valid parameter before it reaches this method), the same will happen to "name"?

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
mxtartaglia-sl: My bad, thanks for pointing it out

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/DefaultLoggingSystem.java
mxtartaglia-sl: This change, in particular, can be rollbacked. It was trying to determine why the default logger was logging initialization errors for some tests even when we instantiated a custom logger for the test. 
The reason was that whenever a `loggingMirror` was instantiated, it automatically added itself as a handler in the default logger, which, therefore, was initialized without need and without having log.properties configured, producing some logs to the emergency logger.
So in a way, there are changes in both the API and the test that solve the issue of reducing the amount of console statements in the tests.

cody-littley:
Friendly reminder to mark yourself as the assignee for PRs 😉  (I marked this one for you).

cody-littley:
This doesn't touch hashgraph owned code, removed myself from the reviewers list. Feel free to add me back if you'd like me to review.

timo0 (approved):

PR #12048, opened 2024-03-12
chore: add mainnet event migration test

### LINKED ISSUE #12021: Add migration mainnet event test
Add a unit test to test event migration as changes are being to events to switch to protobuf. The test should make sure that the hash of the event is consistent across migration.

lpetrovic05:
closes #12021


hedera-node/hedera-app/src/test/java/com/hedera/node/app/platform/event/EventMigrationTest.java
```diff
@@ -0,0 +1,94 @@
...
+                new File(this.getClass()
+                                .getClassLoader()
+                                .getResource("eventFiles/sdk0.46.3/2024-03-05T00_10_55.002129867Z.events")
+                                .toURI())
+                        .toPath(),
+                false)) {
+            while (iterator.hasNext()) {
+                final BaseEventHashedData hashedData = iterator.next().getBaseEventHashedData();
+                numEvents++;
+                CryptographyHolder.get().digestSync(hashedData);
```
edward-swirldslabs, line 74: How is this test performing the migration between event serialization schemes?

lpetrovic05: i explain everything in the javadoc. I have not changed the event serialization yet, I'm about to. the main thing this is testing is that the hash is maintained after migration

tinker-michaelj (approved):
LGTM, ty @lpetrovic05 !

hedera-node/hedera-app/src/test/java/com/hedera/node/app/platform/event/EventMigrationTest.java
edward-swirldslabs: I'm trying to make sure I understand how this tests the migration.  The events on disk are in the old format, they are read into memory.  When they are hashed, the hashing serializes them in the new format?    Since all but 9 events act as parents to other events in the file, it is sufficient to compare the newly hashed values (derived from new serialization) to the stored hashed values in the parent descriptors (derived from the old serialization)?

edward-swirldslabs (approved):

hedera-node/hedera-app/src/test/java/com/hedera/node/app/platform/event/EventMigrationTest.java
lpetrovic05: right now, there is no new format yet, this is what I'm currently working on.
what you said is correct. Ideally, we would have events that are stored together with their hashes. but we don't store event hashes, but we do store parent descriptors which do contain hashes, so we use those to determine if the hash has diverged

PR #12067, opened 2024-03-12
chore: Base executor

### LINKED ISSUE #12095: Executor for base services
The base modules contain several services that will need async tasks in future. Based on that we should provide an executor for the base modules.

hendrikebbers:
Fixes #12095
This PR adds an 1-thread-low-priority-executor that can be used by the base modules (and is hidden for all other modules).


platform-sdk/swirlds-base/src/main/java/com/swirlds/base/internal/impl/BaseScheduledExecutorService.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * A scheduled executor service for the base modules. The executor is based on a single thread and is a daemon thread
+ * with a low priority.
+ * <p>
+ * This class is a singleton and should only be used by code in the base modules that highly needs an asynchronous
+ * executor.
+ */
+public class BaseScheduledExecutorService implements ScheduledExecutorService {
```
mxtartaglia-sl, line 40: Should we register a `ShutdownHook` to make sure we close this executor at some point?

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/internal/impl/BaseScheduledExecutorService.java
```diff
@@ -0,0 +1,160 @@
...
+        final ThreadFactory threadFactory = BaseExecutorThreadFactory.getInstance();
+        this.innerService = Executors.newScheduledThreadPool(CORE_POOL_SIZE, threadFactory);
+    }
+
+    /**
+     * Returns the singleton instance of this executor.
+     *
+     * @return the instance
+     */
+    public static BaseScheduledExecutorService getInstance() {
```
mxtartaglia-sl, line 60: Out of curiosity, in which cases do you prefer a holder inner class, and in which cases do you prefer this synchronized with a double check mechanism?

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/internal/impl/BaseScheduledExecutorService.java
```diff
@@ -0,0 +1,160 @@
...
+ * with a low priority.
+ * <p>
+ * This class is a singleton and should only be used by code in the base modules that highly needs an asynchronous
+ * executor.
+ */
+public class BaseScheduledExecutorService implements ScheduledExecutorService {
+
+    public static final int CORE_POOL_SIZE = 1;
+
+    private static BaseScheduledExecutorService instance;
```
mxtartaglia-sl, line 44:
```suggestion
    private static volatile BaseScheduledExecutorService instance;
```

cody-littley:
This doesn't touch hashgraph owned code, removed myself from the reviewers list. Feel free to add me back if you'd like me to review.

mxtartaglia-sl (commented):
I would only suggest to move the singletons to the inner class holder pattern:
```
class Singleton {
    private static Singleton instance = new Singleton();

    private Singleton() { }

    public static Singleton getInstance() {
        return instance;
    }
}
```

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/internal/impl/BaseScheduledExecutorService.java
hendrikebbers: Good point

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/internal/impl/BaseScheduledExecutorService.java
hendrikebbers: Good point. In general inner class holder is solid and does everything that is needed.

hendrikebbers: and yes, we should define a best practice pattern for us.

mxtartaglia-sl (approved):

PR #12069, opened 2024-03-12
feat: better bind API

### LINKED ISSUE #12068: Improve bind API
`BindableInputWire.bind()` currently has two variations of the `bind()` method: one for wires that connect to the output wire, and another for those that don't connect to the output wire.

There is currently a bug (PR is open to fix) where the wrong variation of the function is accidentally used due to an incorrect type casting. This was hard to notice. It would be much better if these functions weren't named the same thing, so that the choice of which `bind()` method to call is an explicit choice and not up to the compiler to decide.

cody-littley:
closes #12068

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
@@ -101,7 +101,7 @@ public abstract class TaskScheduler<OUT> extends TaskSchedulerInput<OUT> {
 
     /**
      * Build an input wire for passing data to this task scheduler. In order to use this wire, a handler must be bound
-     * via {@link BindableInputWire#bind(Consumer)}.
+     * via {@link BindableInputWire#bindConsumer(Consumer)}.
      *
      * @param name the name of the input wire
      * @param <I>  the type of data that is inserted via this input wire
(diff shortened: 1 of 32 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
```diff
@@ -101,7 +101,7 @@ protected TaskScheduler(
 
     /**
      * Build an input wire for passing data to this task scheduler. In order to use this wire, a handler must be bound
-     * via {@link BindableInputWire#bind(Consumer)}.
+     * via {@link BindableInputWire#bindConsumer(Consumer)}.
```
alittley, line 104: or via `bind`

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
```diff
@@ -229,7 +229,7 @@ public final <X> TaskScheduler<X> cast() {
 
     /**
      * Get the number of unprocessed tasks. A task is considered to be unprocessed until the data has been passed to the
-     * handler method (i.e. the one given to {@link BindableInputWire#bind(Consumer)}) and that handler method has
+     * handler method (i.e. the one given to {@link BindableInputWire#bindConsumer(Consumer)}) and that handler method has
```
alittley, line 232: or given to `bind`

edward-swirldslabs (approved):

alittley (approved):

alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
cody-littley: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/TaskScheduler.java
cody-littley: fixed

alittley (approved):

PR #12080, opened 2024-03-12
fix: 12032 Handle Log framework System.err output in tests

### LINKED ISSUE #12032: Temporally disable Log tests that produce too much console output
The files it's failing on are:
platform-sdk/swirlds-logging/build/test-results/test/TEST-com.swirlds.logging.LoggerImplTest.xml
platform-sdk/swirlds-logging/build/test-results/test/TEST-com.swirlds.logging.api.internal.LoggingSystemTest.xml

mxtartaglia-sl:
Fixes #12032

### DIFF
--- a/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
+++ b/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
@@ -217,7 +217,7 @@ public class LoggingSystem implements LogEventConsumer {
                 .filter(LogProvider::isActive)
                 .toList();
         providers.forEach(p -> p.install(getLogEventFactory(), this));
-        EMERGENCY_LOGGER.log(Level.DEBUG, providers.size() + " logging providers installed: " + providers);
+        EMERGENCY_LOGGER.log(Level.TRACE, providers.size() + " logging providers installed: " + providers);
     }
 
     /**
@@ -262,7 +262,8 @@ public class LoggingSystem implements LogEventConsumer {
                 .toList();
 
         handlers.forEach(this::addHandler);
-        EMERGENCY_LOGGER.log(Level.DEBUG, handlers.size() + " logging handlers installed: " + handlers);
+
+        EMERGENCY_LOGGER.log(Level.TRACE, handlers.size() + " logging handlers installed: " + handlers);
     }
 
     /**
--- a/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
+++ b/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
@@ -217,7 +217,7 @@ public class LoggingSystem implements LogEventConsumer {
                 .filter(LogProvider::isActive)
                 .toList();
         providers.forEach(p -> p.install(getLogEventFactory(), this));
-        EMERGENCY_LOGGER.log(Level.TRACE, providers.size() + " logging providers installed: " + providers);
+        EMERGENCY_LOGGER.log(Level.DEBUG, providers.size() + " logging providers installed: " + providers);
     }
 
     /**
@@ -263,7 +263,7 @@ public class LoggingSystem implements LogEventConsumer {
 
         handlers.forEach(this::addHandler);
 
-        EMERGENCY_LOGGER.log(Level.TRACE, handlers.size() + " logging handlers installed: " + handlers);
+        EMERGENCY_LOGGER.log(Level.DEBUG, handlers.size() + " logging handlers installed: " + handlers);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
```diff
@@ -217,7 +217,7 @@ public void installProviders() {
                 .filter(LogProvider::isActive)
                 .toList();
         providers.forEach(p -> p.install(getLogEventFactory(), this));
-        EMERGENCY_LOGGER.log(Level.DEBUG, providers.size() + " logging providers installed: " + providers);
+        EMERGENCY_LOGGER.log(Level.TRACE, providers.size() + " logging providers installed: " + providers);
```
thomas-swirlds-labs, line 220: Won't this increase the volume of log data (once tests are re-enabled). 
Might there also be any performance impact ?

hendrikebbers: TRACE is a lower level than DEBUG and therefore the log won't be shown by default.
That is called only at instantiation so there is no performance impact.
@mxtartaglia-sl next to that I do not get why you reduced the level. That line will maybe called 1 time per test and based on that it should not be a problem.

platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/api/internal/format/EpochFormatUtilsTest.java
```diff
@@ -19,10 +19,12 @@
 import static java.time.ZoneOffset.UTC;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.swirlds.base.test.fixtures.io.WithSystemError;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import org.junit.jupiter.api.Test;
 
+@WithSystemError
```
hendrikebbers, line 27: Why? Does the EpochFormatUtils create sys out? Or is that the emergency logger that is used in the util.

platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/util/HandlerLoggingLevelConfigTestOrchestrator.java
```diff
@@ -75,7 +75,6 @@ private HandlerLoggingLevelConfigTestOrchestrator(TestScenario... testScenarios)
      * Performs the verification of scenario given by its index on the list
      */
     private void testScenario(int scenario, HandlerLoggingLevelConfig config) {
-        System.out.printf("Testing scenario %d%n", scenario);
```
hendrikebbers, line 78: I do not get it. It can't be a problem to do 1 line of output per test (why I would like to use the emergency logger instead of sys out)

platform-sdk/swirlds-logging/src/testFixtures/java/com/swirlds/logging/test/fixtures/internal/LoggerMirrorExtension.java
```diff
@@ -32,9 +33,11 @@ public void interceptTestMethod(
             ExtensionContext extensionContext)
             throws Throwable {
         try (final LoggingMirrorImpl loggingMirror = new LoggingMirrorImpl()) {
+            DefaultLoggingSystem.getInstance().addHandler(loggingMirror);
```
hendrikebbers, line 36: we should do a try-finally for adding/removing the handler

alittley (approved):

platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/api/internal/format/EpochFormatUtilsTest.java
mxtartaglia-sl: Yes the utils logs to the emergency logger when it can't parse the date.

platform-sdk/swirlds-logging/src/test/java/com/swirlds/logging/util/HandlerLoggingLevelConfigTestOrchestrator.java
mxtartaglia-sl: It is not, and it also doesn't add any value. And when a scenario fails, it does say which one.
I didn't use the emergency logger, so we don't have circular dependencies. The tool is a helper for testing the logging framework, and I didn't want it to use the logging framework for its internal activities.

platform-sdk/swirlds-logging/src/testFixtures/java/com/swirlds/logging/test/fixtures/internal/LoggerMirrorExtension.java
mxtartaglia-sl: Ended, try-finally inside a try seemed ugly, but it leaves a lot of scenarios without removing the handler

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/LoggingSystem.java
mxtartaglia-sl: Yes, in a second thought this change is not needed. The only problem is that many tests have the debug level enabled by default, making this line repeat often enough. Maybe we should address the test later and leave  the level as it is.

mxtartaglia-sl: Rollbacked

alittley (approved):

timo0 (approved):

thomas-swirlds-labs (approved):
LGTM! @mxtartaglia-sl

PR #12093, opened 2024-03-13
feat: Map HapiContractCreate to HapiEthereumContractCreate

### LINKED ISSUE #11728: Get the nonce and the privateKeyRef
The two fields need to be passed in the `HapiEthereumContractCreate`'s constructor (created in https://github.com/hashgraph/hedera-services/issues/11729). The constructor
will be called inside the `convertHapiCallsToEthereumCalls` method. To be able to pass the variables in the constructor
(in https://github.com/hashgraph/hedera-services/issues/11729) in this story we will make sure that they are available for passing.

**nonce** - we want to implement it just like `HapiEthereumCall`. Introduce `useSpecNonce` that it's false if we set the _nonce_ otherwise in opBodyDef just before constructing the `EthTxData` we need something like:
```
if (useSpecNonce) {
    nonce = spec.getNonce(privateKeyRef);
}
```

**privateKeyRef** needs to be passed in `convertHapiCallsToEthereumCalls`. Currently, it's called from two places:
- `HapiSpec.run` - if `isEthereumAccountCreatedForSpec(this)` is _true_ then get the existing one. To do that we will probably need to create a new static method inside `UtilStateChange`. If it's false - then get the key created from `initializeEthereumAccountForSpec`
- `CustomSpecAssertion` - get from `UtilStateChange`'s static method created above

### LINKED ISSUE #11729: Map from HapiContractCreate to HapiEthereumContractCreate
* In `HapiContractCreate` implement `isConvertableToEthCreate` and `refusingEthConversion` just like `HapiContractCall`

- In `UtilVerbs.convertHapiCallsToEthereumCalls` add:
```
else if (op instanceof HapiContractCreate callOp && callOp.isConvertableToEthCreate()) {
    convertedOps.add(new HapiEthereumContractCreate(callOp));
}
```

- Add a constructor to **HapiEthereumContractCreate** to map from **HapiContractCreate**
```
public HapiEthereumContractCreate(HapiContractCreate contractCreate) {
    // TODO: implement
}
```
* Most of the mapping will be straightforward since both `HapiEthereumContractCreate` and `HapiContractCreate` implement `HapiBaseContractCreate`
* the nonce and privateKeyRef will come from https://github.com/hashgraph/hedera-services/issues/11728.
* type - hardcode **EIP1559** - the same as **HapiEthereumCall**
* gasPrice, maxFeePerGas, maxPriorityGas, invalidateEthData, ethFileID and maxGasAllowance - use the default/skip

vtronkov:
**Description**:
* Adding the mapping from HapiContractCreate to HapiEthereumContractCreate when we want to execute a suite with an Ethereum context

**Related issue(s)**:

Fixes #11728
Fixes #11729


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpecOperation.java
```diff
@@ -354,7 +356,9 @@ protected Transaction finalizedTxn(
                 .andThen(opDef);
 
         setKeyControlOverrides(spec);
-        final List<Key> keys = signersToUseFor(spec);
+        List<Key> keys = signersToUseFor(spec);
+        Iterables.removeIf(keys, Predicates.isNull());
```
tinker-michaelj, line 360: Why would there be `null` keys here?

vtronkov: While testing I had the problem of having `null` keys here. I really can't remember the exact case but this was the reason I added this line.

Dealing with the nulls inside `signersToUseFor` makes more sense, I agree. Fixed 👍

Neeharika-Sompalli (approved):
LGTM !

tinker-michaelj (approved):
LGTM, ty @vtronkov !

PR #12107, opened 2024-03-13
feat: pces birth round migration

### LINKED ISSUE #11612: PCES birth round migration
When we enable birth rounds, we need to perform a migration on the PCES. This will involve reading part of the PCES and writing it back in a different format.

cody-littley:
closes #11612


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesBirthRoundMigration.java
```diff
@@ -0,0 +1,262 @@
...
+                migrationRound,
+                minimumJudgeGenerationInMigrationRound);
+
+        makeBackupFiles(recycleBin, databaseDirectory);
+
+        final List<GossipEvent> eventsToMigrate = readEventsToBeMigrated(
+                platformContext, recycleBin, selfId, minimumJudgeGenerationInMigrationRound, migrationRound);
+
+        if (eventsToMigrate.isEmpty()) {
+            logger.error(EXCEPTION.getMarker(), "No events to migrate. PCES birth round migration aborted.");
```
alittley, line 113: Is this case supposed to be impossible (it isn't being tested)? If it isn't impossible, should old files be cleaned up?

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/preconsensus/PcesBirthRoundMigrationTests.java
```diff
@@ -0,0 +1,424 @@
...
+        }
+
+        Files.createDirectories(testDirectory);
+        recycleBinPath = testDirectory.resolve("recycle-bin");
+        pcesPath = testDirectory.resolve("pces");
+        temporaryFilePath = testDirectory.resolve("tmp");
+
+        StaticSoftwareVersion.setSoftwareVersion(new BasicSoftwareVersion(0));
+
+        System.out.println(testDirectory); // TODO remove
```
alittley, line 103: TODO

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/preconsensus/PcesBirthRoundMigrationTests.java
```diff
@@ -0,0 +1,424 @@
...
+        StaticSoftwareVersion.reset();
+    }
+
+    private record PcesFilesWritten(@NonNull List<PcesFile> files, @NonNull List<GossipEvent> events) {}
+
+    /**
+     * Generate a bunch of PCES files in the legacy format.
+     *
+     * @param random        the random number generator
+     * @param discontinuity whether to introduce a discontinuity in the stream. If true, discontinuity is placed at the
```
alittley, line 121: What if a discontinuity were to occur *after* the migration point?

alittley (approved):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesBirthRoundMigration.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Find all PCES files beneath a given directory. Unlike the normal process of reading PCES files via
+     * {@link PcesFileReader#readFilesFromDisk(PlatformContext, RecycleBin, Path, long, boolean, AncientMode)}, this
+     * method ignores discontinuities and returns all files.
+     *
+     * @param ancientMode only return files that conform to this ancient mode
+     * @return all PCES files beneath the given directory
+     */
+    @NonNull
+    public static List<PcesFile> findPcesFiles(@NonNull final Path path, @NonNull final AncientMode ancientMode) {
```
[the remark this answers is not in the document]
cody-littley: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesBirthRoundMigration.java
cody-littley: This should be impossible. More of a sanity check than a real scenario I expect to encounter.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/preconsensus/PcesBirthRoundMigrationTests.java
cody-littley: removed

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/preconsensus/PcesBirthRoundMigrationTests.java
cody-littley: In theory, we might see a discontinuity after the migration point if a node reconnected and got the freeze state.

The short story is that all events in the PCES prior to the discontinuity will be ignored, and only events after the discontinuity will be eligible for migration. File cleanup ignores discontinuities/origin rounds, so that will work the same as normal.

When reading events to be migrated, we use the method `PcesFileTracker.getFileIterator(final long lowerBound, final long originRound)`, which is designed only to traverse events without crossing discontinuities. This is well unit tested in the non-migration unit tests.

Suppose a node reconnects and gets the freeze round state. There are two possible scenarios:

1) the node gets all non-ancient events through gossip prior to its peers shutting down
2) the node does not get all non-ancient events through gossip prior to its peers shutting down

Scenario 1 is easy, the migration will happen just like normal.

Scenario 2 is a bit more complex. If the network shuts down before the reconnecting node receives non-ancient events through gossip, and if the node is eventually power cycled to upgrade it (currently I think it might get stuck waiting for signatures), then it may not migrate some events. It will receive those events through gossip when it starts back up (it is impossible for this to happen to >=1/3 nodes). It will then write those events to its PCES in birth round mode. No long term harm.

Are these scenarios something you think it worthwhile to unit test?

alittley: I think it might be worth adding to unit tests, just to cover all the edges

cody-littley: Added requested test coverage.

edward-swirldslabs (approved):

Issue #11582, opened 2024-02-16
Self ISS during reconnect


netopyr:
While we performed a reconnect in one of our tests, the node encountered a self ISS. More details can be found here: https://github.com/hashgraph/hedera-testing/issues/434

tomzhenghedera:
see latest update in [Test issue 438](https://github.com/hashgraph/hedera-testing/issues/438#issuecomment-1976065772)

tomzhenghedera:
auto-reconnect regressed with latest build, see update in [test issue 434](https://github.com/hashgraph/hedera-testing/issues/434#issuecomment-1986951352)

tomzhenghedera:
With the latest **fix-entityid-issue** branch build, the auto-reconnect Self-ISS and node restart Self-ISS didn't not show and can be considered the fix worked.  This ticket could be closed whence it's merged.

PR #12109, opened 2024-03-13
fix: (mod-service) Fix a case for identifying `RESTART` scenario

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/11582

When there are no applicable schemas if the version to be migrated is same, we apply the last registered schema. In this scenario, the `ApplicationType` should be considered as `RESTART_ONLY`.

If it is considered as `MIGRATE_THEN_RESTART`, it could lead to ISS if node does a auto-reconnect since the `EntityId` service commits a different EntityId number from state.


tinker-michaelj (approved):
LGTM, ty @netopyr !

netopyr:
> LGTM, ty @netopyr !

Well, thank you, @Neeharika-Sompalli ! I just did the cherry-pick. 😄

PR #12112, opened 2024-03-13
feat: node starts without checking internal hostname

### LINKED ISSUE #11751: Use commandline or environment variable to determine node ids to start locally.
`ServicesMain.java` already has commandline argument parsing to determine local nodes to start up.  Likely this can be leveraged into a common util / process.  

The hard work is going to shifting the code that uses `AddressBookNetworkUtils.isLocal()` to using the above for determining the locality of nodes. 
* `CryptoStatic` checks `isLocal` to determine if the private keys should be loaded from disk
* `EnhancedKeyStoreLoader` checks `isLocal` to determine if the private keys should be loaded from disk. 
    * Are the above two alternate forms of doing the same thing? 
* `EnhancedKeyStoreLoader` uses `isLocal` in validation phase for extra checks. 
* `OutBoundConnectionCreator` checks `isLocal` to use loopback address instead of going through network address. 
* `SwirldsGUI` uses `isLocal` to get the count of the number of loads that are local to create GUIs for them. 

The environment variables need to be queried for a specification of which nodeId(s) to start locally.   The commandline arguments, if provided, take precedence over the environment variables. If neither commandline or environment variables are provided, start all nodes locally. 

NOTE: merging the PR for this requires coordination with SOLO and DevOps.  They need to provide environment variables to indicate the specific nodes to start.

edward-swirldslabs:
Must not merge until https://github.com/swirlds/swirlds-platform-regression/pull/4032 is merged

**Description**:
This PR removes the use of the internal ip address for determining which nodes to start.   The nodes to start locally must be specified on the command line or through environment variable.  
* If no nodes are specified:
   * `Browser` starts all nodes locally.
   * `ServicesMain` logs an error and exits.     
* If multiple nodes are specified
  * `Browser` starts just the nodes specified locally
  * `ServicesMain` exits with an error indicating exactly 1 node must be started. 
* If specified, the commandline arguments take precedence over environment variables. 

The methods for loading the cryptography now take parameters indicating which nodes will be started locally. 

CLI: comma separated list at end of java command. 
```
java <jar> -local 0,1,2
```

Environment Variable:  comma separated list for the `nodesToRun` environment variable. 
```
nodesToRun=0,1,2
```

**!!!! This PR impacts all deployments of nodes, from JRS, to SOLO, to DevOps managed test and production networks.**
* To start a single node, provide a single node id through commandline or environment variable.  

I've removed `ServicesMainTest.java` as it only tested the behavior of mismatched addresses.  The logic has significantly changed.  These unit tests were actually misbehaving.  legacyConfigProperties was returning a null address book instead of an empty address book.    All code paths are used in testing and production and will fail critically if the code paths are not working as expected. Maintaining complicated static mocked scenarios is not necessary. 
1. HAPI tests specify nodes to run through the command line.  
2. SOLO uses environment variables to configure nodes to run. 
3. The Browser runs all nodes locally. 
4. ServicesMain later ensures there is exactly 1 node running. 

**Related issue(s)**:

Fixes #11751 

**Testing**

This was manually tested locally with the Browser and verified to work with both CLI and ENV methods.   The same logic is at play in ServicesMain.  The proof is when the code is executed in testing environments and it works. 

**Checklist**

Possible Impacts?
* NMT
* JRS
* Solo / FST
* HAPI tests
* Local Node Testing
* Gradle Build
* Perfnet Testing

Sign Offs
* [ ] Services
* [ ] Testing
* [x] Release Engineering
* [ ] DevOps


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -506,9 +512,12 @@ public static SecureRandom getNonDetRandom() {
      * @return a map of KeysAndCerts objects, one for each node
      */
     public static Map<NodeId, KeysAndCerts> initNodeSecurity(
-            @NonNull final AddressBook addressBook, @NonNull final Configuration configuration) {
+            @NonNull final AddressBook addressBook,
+            @NonNull final Configuration configuration,
+            @NonNull final Set<NodeId> nodesToStart) {
```
alittley, line 517: missing javadoc for `nodesToStart`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
```diff
@@ -379,21 +378,26 @@ public static void writeSettingsUsed(@NonNull final Configuration configuration)
...
-        for (final Address address : addressBook) {
-            // if the local nodes to start are not specified, start all local nodes. Otherwise, start specified.
-            if (AddressBookNetworkUtils.isLocal(address)
-                    && (localNodesToStart.isEmpty() || localNodesToStart.contains(address.getNodeId()))) {
-                nodesToRun.add(address.getNodeId());
+        // if the nodes to start are not specified, start all nodes. Otherwise, start specified.
+        if (nodesToStart.isEmpty()) {
+            nodesToRun.addAll(addressBook.getNodeIdSet());
+        } else {
+            for (NodeId nodeId : nodesToStart) {
```
alittley, line 394: `nodeId` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
edward-swirldslabs: added

alittley (approved):

edward-swirldslabs:
Lets of work to do before this can be merged.    This PR removes a capability without introducing the capability's replacement.   JRS and other tools such as NMT will need to be re-worked to support either (1) command line identification of which node to start or (2) setting of an environment variable which indicates which node to start.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -213,41 +217,47 @@ public class EnhancedKeyStoreLoader {
...
+            @NonNull final Set<NodeId> nodesToStart) {
         this.addressBook = Objects.requireNonNull(addressBook, MSG_ADDRESS_BOOK_NON_NULL);
         this.keyStoreDirectory = Objects.requireNonNull(keyStoreDirectory, MSG_KEY_STORE_DIRECTORY_NON_NULL);
         this.keyStorePassphrase = Objects.requireNonNull(keyStorePassphrase, MSG_KEY_STORE_PASSPHRASE_NON_NULL);
         this.sigPrivateKeys = HashMap.newHashMap(addressBook.getSize());
         this.sigCertificates = HashMap.newHashMap(addressBook.getSize());
         this.agrPrivateKeys = HashMap.newHashMap(addressBook.getSize());
         this.agrCertificates = HashMap.newHashMap(addressBook.getSize());
-        this.localNodes = HashSet.newHashSet(addressBook.getSize());
+        this.localNodes = new HashSet<>(Objects.requireNonNull(nodesToStart, MSG_NODES_TO_START_NON_NULL));
```
nathanklick, line 240:
```suggestion
        this.localNodes = Collections.unmodifiableSet(Objects.requireNonNull(nodesToStart, MSG_NODES_TO_START_NON_NULL));
```

cody-littley (approved):
LGTM, although this should not be merged until all downstream dependents give the thumbs up.

rbair23 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
```diff
@@ -318,20 +318,39 @@ public static void writeSettingsUsed(@NonNull final Configuration configuration)
...
-            // if the local nodes to start are not specified, start all local nodes. Otherwise, start specified.
-            if (AddressBookNetworkUtils.isLocal(address)
-                    && (localNodesToStart.isEmpty() || localNodesToStart.contains(address.getNodeId()))) {
-                nodesToRun.add(address.getNodeId());
+        final Set<NodeId> addressBookNodeIds = addressBook.getNodeIdSet();
+
+        if (cliNodesToRun.isEmpty()) {
+            if (envNodesToRun.isEmpty()) {
+                // if no node ids are provided by cli or env, run all nodes from the address book.
+                return new ArrayList<>(addressBookNodeIds);
```
anthony-swirldslabs, line 340: This logic modification seems to require a change for all node operators (mainnet, testnet, all the perf/preview networks, and all testing frameworks that start a node via this code path), which is a lot of affected parties.

Is this being communicated to all the affected parties and configuration changes are being made to ensure that all the systems are configured before this code is released?

anthony-swirldslabs: Also, I don't see where ServicesMain would log an error if both the env vars and CLI args are missing. It only complains if more than one node is specified on the command line. I think you'd want to verify that logic.


If verified and/or fixed, so that the node indeed errors out when the config is totally missing everywhere, then this concern can be resolved. (Although you'd still want to communicate the need to update the configuration to all the parties, so that they don't get surprised with unexpected exceptions.)

edward-swirldslabs: It's in the BootstrapUtils call after the code I modified.  https://github.com/hashgraph/hedera-services/blob/d0888fc879a8ff49089f41e80a4227a5bf568e0f/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java#L346-L356

anthony-swirldslabs: That looks good. Thanks! However, we still need to communicate the change to everyone who needs to update their nodes configurations before releasing this code.

lpetrovic05:
If no nodes are specified, I feel like we should not run all nodes. We could either die, or we could add a mechanism to check which private keys we have on disk and determine the node to run based on that. Wdyt?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/SwirldsGui.java
```diff
@@ -45,15 +46,20 @@ private SwirldsGui() {}
...
      * @param visible  should the window be initially visible? If not, call setVisible(true) later.
+     * @param nodesToStart the set of nodes to start windows for
      * @return the new window
      */
-    public static Console createConsole(final Platform platform, final int winNum, final boolean visible) {
+    public static Console createConsole(
+            @NonNull final Platform platform,
+            final int winNum,
+            final boolean visible,
+            @NonNull final Set<NodeId> nodesToStart) {
```
lpetrovic05, line 56: feel like this variable in this context should be something like `localNodes`

lpetrovic05 (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -230,12 +232,16 @@ public static void main(final String... args) throws Exception {
...
+        final var envNodesToRun = platformConfig.getConfigData(BasicConfig.class).nodesToRun().stream()
+                .map(NodeId::of)
+                .toList();
+        final List<NodeId> nodesToRun = getNodesToRun(diskAddressBook, cliNodesToRun, envNodesToRun);
         checkNodesToRun(nodesToRun);
         final var selfId = ensureSingleNode(nodesToRun, commandLineArgs.localNodesToStart());
-        final var platformConfig = buildPlatformConfig();
         BootstrapUtils.setupConstructableRegistryWithConfiguration(platformConfig);
-        final var networkKeysAndCerts = initNodeSecurity(diskAddressBook, platformConfig);
+        final var networkKeysAndCerts = initNodeSecurity(diskAddressBook, platformConfig, new HashSet<>(nodesToRun));
```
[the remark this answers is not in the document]
edward-swirldslabs: updated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -206,19 +216,8 @@ private static void launchUnhandled(@NonNull final CommandLineArgs commandLineAr
...
-            final Configuration guiConfig = guiConfigBuilder.build();
-
-            final ConfigurationBuilder configBuilder = ConfigurationBuilder.create();
-            rethrowIO(() ->
-                    BootstrapUtils.setupConfigBuilder(configBuilder, getAbsolutePath(DEFAULT_SETTINGS_FILE_NAME)));
-            final Configuration configuration = configBuilder.build();
-
-            initNodeSecurity(appDefinition.getConfigAddressBook(), configuration);
-            guiEventStorage = new GuiEventStorage(guiConfig, appDefinition.getConfigAddressBook());
+            initNodeSecurity(appDefinition.getConfigAddressBook(), bootstrapConfiguration, new HashSet<>(nodesToRun));
```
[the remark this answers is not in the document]
edward-swirldslabs: updated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Browser.java
```diff
@@ -253,7 +252,8 @@ private static void launchUnhandled(@NonNull final CommandLineArgs commandLineAr
                     nodeId);
             final var cryptography = CryptographyFactory.create();
             CryptographyHolder.set(cryptography);
-            final KeysAndCerts keysAndCerts = initNodeSecurity(appDefinition.getConfigAddressBook(), configuration)
+            final KeysAndCerts keysAndCerts = initNodeSecurity(
+                            appDefinition.getConfigAddressBook(), configuration, new HashSet<>(nodesToRun))
```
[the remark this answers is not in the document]
edward-swirldslabs: updated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -351,12 +353,16 @@ private static String getPrivateKeysFileName(final String memberName) {
     @NonNull
     @Deprecated(since = "0.47.0", forRemoval = false)
     static Map<NodeId, KeysAndCerts> loadKeysAndCerts(
-            @NonNull final AddressBook addressBook, @NonNull final Path keysDirPath, @NonNull final char[] password)
+            @NonNull final AddressBook addressBook,
+            @NonNull final Path keysDirPath,
+            @NonNull final char[] password,
+            @NonNull Set<NodeId> nodesToStart)
```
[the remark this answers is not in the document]
edward-swirldslabs: made final

edward-swirldslabs:
> If no nodes are specified, I feel like we should not run all nodes. We could either die, or we could add a mechanism to check which private keys we have on disk and determine the node to run based on that. Wdyt?

When we run platform alternate apps locally, the default behavior is to run all nodes unless specified otherwise.  I've preserved this behavior.  I know the GUI and alternate apps are going the way of the dodo, but not yet. 

There is a check in services to ensure exactly 1 node is started.   I think that's fine for now.   I don't want to add code to check for local keys.  Someone could add additional keys by mistake or we might come up with key rotation policies that have multiple keys on disk for a short time.  It makes things more complicated to tie those two things together.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
```diff
@@ -349,31 +352,16 @@ public static void writeSettingsUsed(@NonNull final Configuration configuration)
 
         for (final NodeId nodeId : nodesToRun) {
             if (!addressBook.contains(nodeId)) {
-                // all nodes to start must exist in the address book.
-                throw new IllegalArgumentException("Node " + nodeId + " is not in the address book");
+                // all nodes to start must exist in the address book
+                exitSystem(NODE_ADDRESS_MISMATCH);
+                throw new ConfigurationException(
+                        "Node " + nodeId + " is not in the address book and cannot be started.");
```
anthony-swirldslabs, line 358: Minor q: wouldn't the `exitSystem` exit before the exception is thrown? If yes, perhaps we should log the error first so as to provide more diagnostics for debugging?

anthony-swirldslabs (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -375,7 +393,9 @@ public static Hedera newHedera(@NonNull final NodeId selfNodeId, @NonNull final
     public static Configuration buildPlatformConfig() {
         final ConfigurationBuilder configurationBuilder = ConfigurationBuilder.create()
                 .withSource(SystemEnvironmentConfigSource.getInstance())
-                .withSource(SystemPropertiesConfigSource.getInstance());
+                .withSource(SystemPropertiesConfigSource.getInstance())
+                .withConfigDataType(BasicConfig.class);
```
timo0, line 397: not sure if this is needed.

`BootstrapUtils.setupConfigBuilder` uses `autoDiscoverExtensions()`

lpetrovic05:
> > If no nodes are specified, I feel like we should not run all nodes. We could either die, or we could add a mechanism to check which private keys we have on disk and determine the node to run based on that. Wdyt?
> 
> When we run platform alternate apps locally, the default behavior is to run all nodes unless specified otherwise. I've preserved this behavior. I know the GUI and alternate apps are going the way of the dodo, but not yet.
> 
> There is a check in services to ensure exactly 1 node is started. I think that's fine for now. I don't want to add code to check for local keys. Someone could add additional keys by mistake or we might come up with key rotation policies that have multiple keys on disk for a short time. It makes things more complicated to tie those two things together.

makes sense

lpetrovic05 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/BasicConfig.java
```diff
@@ -56,7 +58,8 @@ public record BasicConfig(
         @ConfigProperty(defaultValue = "1000") int jvmPauseReportMs,
         @ConfigProperty(defaultValue = "60s") Duration hangingThreadDuration,
         @ConfigProperty(defaultValue = "data/saved") String emergencyRecoveryFileLoadDir,
-        @ConfigProperty(defaultValue = "0") long genesisFreezeTime) {
+        @ConfigProperty(defaultValue = "0") long genesisFreezeTime,
+        @ConfigProperty(defaultValue = Configuration.EMPTY_LIST) List<Long> nodesToRun) {
```
timo0, line 62: Since you always use this list as `NodeId` you could store these as `List<NodeId>` and create a `ConfigConverter` for `NodeId`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
timo0: you can add the msg to `exitSystem` and remove the `throw` since it's gets never executed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -214,12 +229,15 @@ public static void main(final String... args) throws Exception {
                     "Multiple nodes were supplied via the command line. Only one node can be started per java process.");
             exitSystem(NODE_ADDRESS_MISMATCH);
         }
-        final List<NodeId> nodesToRun = getNodesToRun(diskAddressBook, commandLineArgs.localNodesToStart());
-        checkNodesToRun(nodesToRun);
-        final var selfId = ensureSingleNode(nodesToRun, commandLineArgs.localNodesToStart());
         final var platformConfig = buildPlatformConfig();
+        final var cliNodesToRun = commandLineArgs.localNodesToStart();
+        final var envNodesToRun = platformConfig.getConfigData(BasicConfig.class).nodesToRun().stream()
```
timo0, line 234: `envNodesToRun` is a little "short" `configNodesToRun` would fit better since the the value could possible come from a config file as well

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -384,44 +404,29 @@ public static Configuration buildPlatformConfig() {
...
+        if (nodesToRun.size() > 1) {
             logger.error(
                     EXCEPTION.getMarker(),
-                    "The requested node id {} is not configured to run. Please check the address book.",
-                    requestedNodeId);
-            exitSystem(NODE_ADDRESS_MISMATCH);
-            throw new ConfigurationException(String.format(
-                    "The requested node id %s is not configured to run. Please check the address book.",
-                    requestedNodeId));
+                    "Multiple nodes are configured to run. Exactly one node must be started per java process.");
```
timo0, line 425: minor: Why not the same msg?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/SwirldsGui.java
```diff
@@ -45,15 +45,19 @@ private SwirldsGui() {}
...
      * @param visible  should the window be initially visible? If not, call setVisible(true) later.
+     * @param localNodes the set of nodes running locally need content rendered in the gui.
      * @return the new window
      */
-    public static Console createConsole(final Platform platform, final int winNum, final boolean visible) {
+    public static Console createConsole(
+            @NonNull final Platform platform,
+            final int winNum,
+            final boolean visible,
+            @NonNull final Set<NodeId> localNodes) {
```
timo0, line 55: You repeated for all calls the
```
platform.getRoster().rosterEntries().stream()
                        .map(RosterEntry::nodeId)
                        .map(NodeId::of)
                        .collect(Collectors.toSet()
```
maybe the parameter should be removed and called here internally?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
edward-swirldslabs: I am removing the throw after the exit.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/BasicConfig.java
edward-swirldslabs: Creating config converter for NodeId

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
edward-swirldslabs: makes sense.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
edward-swirldslabs: making the error message the same.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/SwirldsGui.java
edward-swirldslabs: Using the config converter removes this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/SwirldsGui.java
```diff
@@ -44,16 +43,16 @@ private SwirldsGui() {}
      * Create a new window with a text console, of the recommended size and location, including the Swirlds menu.
      *
      * @param platform the platform to create the console with
+     * @param winNum the number of the window, starting at 0
      * @param visible  should the window be initially visible? If not, call setVisible(true) later.
      * @return the new window
      */
-    public static Console createConsole(final Platform platform, final int winNum, final boolean visible) {
+    public static Console createConsole(@NonNull final Platform platform, final int winNum, final boolean visible) {
```
timo0, line 50: Minor: there is no null-check for platform

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoaderTest.java
```diff
@@ -124,7 +124,10 @@ void keyStoreLoaderPositiveTest(final String directoryName)
             throws IOException, KeyLoadingException, KeyStoreException {
         final Path keyDirectory = testDataDirectory.resolve(directoryName);
         final AddressBook addressBook = addressBook();
-        final EnhancedKeyStoreLoader loader = EnhancedKeyStoreLoader.using(addressBook, configure(keyDirectory));
+        Set<NodeId> nodesToStart = addressBook.getNodeIdSet();
```
timo0, line 127: minor
```suggestion
        final Set<NodeId> nodesToStart = addressBook.getNodeIdSet();
```

timo0 (approved):
just minor comments

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/SwirldsGui.java
edward-swirldslabs: The platform is accessed in the method. we've been leaving off null checks because a null pointer exception would be generated in the method that the argument is passed.   We add null checks when the null pointer exception is not generated in the same call stack.  (Setting local variables that get used later).

lpetrovic05 (dismissed):

timo0 (dismissed):

kimbor (approved):

jeromy-cannon:
@edward-swirldslabs , this is confusing:
CLI: comma separated list at end of java command.
```
java <jar> -local 0,1,2
```
Environment Variable: comma separated list for the nodesToRun environment variable.
```
nodesToRun=0,1,2
```
It looks like the code and expectation is that you only supply a single node ID, but your example here shows multiple.

edward-swirldslabs:
> @edward-swirldslabs , this is confusing: CLI: comma separated list at end of java command.
> 
> ```
> java <jar> -local 0,1,2
> ```
> 
> Environment Variable: comma separated list for the nodesToRun environment variable.
> 
> ```
> nodesToRun=0,1,2
> ```
> 
> It looks like the code and expectation is that you only supply a single node ID, but your example here shows multiple.

I have updated the PR text to clarify that ServicesMain can only start 1 node per process.  Browser can start multiple nodes per process.

hedera-node/infrastructure/docker/containers/production-next/consensus-node/entrypoint.sh
```diff
@@ -100,6 +100,10 @@ if [[ "${JCP_OVERRIDDEN}" != true && "${JAVA_MAIN_CLASS}" == "com.swirlds.platfo
   JAVA_CLASS_PATH="data/lib/*"
 fi
 
+# Setup Consensus Node Arguments
+CONSENSUS_NODE_ARGS=""
+[[ -n "${CONSENSUS_NODE_ID}" && "${CONSENSUS_NODE_ID}" -ge 0 ]] && CONSENSUS_NODE_ARGS="-local ${CONSENSUS_NODE_ID}"
```
dalvizu, line 105: is CONSENSUS_NODE_ID is not specified or less than 0 should this be treated as an error condition and not startup?

nathanklick: @dalvizu It ultimately will be at the NMT level. We could add a failure condition here; however, these images are also used for older software releases which do not have the `-local` option. 

We made it optional here for backwards compatibility with NMT 1.2.9 and `0.58.x` releases.

nathanklick: Ultimately, the `ServicesMain` entrypoint should fail fast if the required nodeId is not specified. 

@edward-swirldslabs Does the `ServicesMain` entrypoint fail if no environment variable and no `-local` switch is supplied?

edward-swirldslabs: @nathanklick Yes.  The call to `ensureSingleNode` makes sure there is no more and no less than 1 node being started.  

If no -local switch is supplied, then we're relying on the environment variable to determine the node to start.  If no environment variable is set, then the node will fail to start (from ServicesMain) because its singular identity was not specified.

nathanklick: @dalvizu Does this satisfy your concerns?

alex-kuzmin-hg (dismissed):
Reviewed - loks OK

hedera-node/infrastructure/docker/containers/production-next/consensus-node/entrypoint.sh
edward-swirldslabs: The commandline and environmment variables are parsed into `NodeId` constructs that wrap the long value.  If the long value provided is negative, the `NodeId` will fail with a java illegal argument exception.

imalygin (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -396,44 +399,28 @@ public static Configuration buildPlatformConfig() {
...
-        if (nodesToRun.size() > 1 && !nodesToRun.contains(requestedNodeId)) {
-            logger.error(
-                    EXCEPTION.getMarker(),
-                    "The requested node id {} is not configured to run. Please check the address book.",
-                    requestedNodeId);
-            exitSystem(NODE_ADDRESS_MISMATCH);
-            throw new ConfigurationException(String.format(
-                    "The requested node id %s is not configured to run. Please check the address book.",
-                    requestedNodeId));
+        logger.info(STARTUP.getMarker(), "there are {} nodes set to run locally", nodesToRun.size());
```
anthony-swirldslabs, line 411: Minor: logging the whole list itself might be helpful for debugging, too.

hedera-node/infrastructure/docker/containers/production/main-network-node/entrypoint.sh
```diff
@@ -37,7 +37,7 @@ if [[ -n "${JAVA_HEAP_MAX}" ]]; then
 fi
 
 # Setup Main Class
-[[ -z "${JAVA_MAIN_CLASS}" ]] && JAVA_MAIN_CLASS="com.swirlds.platform.Browser"
+[[ -z "${JAVA_MAIN_CLASS}" ]] && JAVA_MAIN_CLASS="com.hedera.node.app.ServicesMain"
```
anthony-swirldslabs, line 40: This looks like a huge change. I'm all for having only a single entry point class to minimize differences between testing and production environments, so +1 on the change itself.

However, I'm wondering if this is a safe change at this point? Has this been tested? Also, should this modification really be a part of this particular PR?

Also, can the Browser class be nuked now?

anthony-swirldslabs (dismissed):

hedera-node/infrastructure/docker/containers/production/main-network-node/entrypoint.sh
edward-swirldslabs: Browser is needed for local testing.     This file being changed is under release engineering / devops control.    No idea why they would have been using Browser as an entrypoint this late in the game.  It should have been ServicesMain all along.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
anthony-swirldslabs: If someone mistakenly configures a production/ServicesMain use-case to start several nodes and gets an exception thrown just a few lines below in this method, it would be immensely helpful to see which particular nodes were configured to start - this could provide clues as to where the bad configuration comes from or help search/grep the incorrect config values. With only a count of entries in this list printed in the log, it may be a little more complex to debug such a case. But I'm okay if you don't want to log it.

hedera-node/infrastructure/docker/containers/production/main-network-node/entrypoint.sh
anthony-swirldslabs: Sounds good. I was also surprised to see Browser mentioned in a "main network" (supposedly, production) file. Thanks for the clarification.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
edward-swirldslabs: I've updated the log message.   

It looks like this: `The following nodes [0, 1, 2, 3] are set to run locally`. (ran from browser.)

anthony-swirldslabs (approved):
LGTM. Thanks!

anthony-swirldslabs (approved):

nathanklick (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -228,12 +228,14 @@ public static void main(final String... args) throws Exception {
...
         }
-        final List<NodeId> nodesToRun = getNodesToRun(diskAddressBook, commandLineArgs.localNodesToStart());
-        checkNodesToRun(nodesToRun);
-        final var selfId = ensureSingleNode(nodesToRun, commandLineArgs.localNodesToStart());
         final var platformConfig = buildPlatformConfig();
+        final var cliNodesToRun = commandLineArgs.localNodesToStart();
+        final var configNodesToRun =
+                platformConfig.getConfigData(BasicConfig.class).nodesToRun();
+        final List<NodeId> nodesToRun = getNodesToRun(diskAddressBook, cliNodesToRun, configNodesToRun);
+        final var selfId = ensureSingleNode(nodesToRun);
```
mhess-swl, line 236: 😵‍💫 this took some digging. For clarity, especially when I read this code again next, I suggest:

```suggestion
        // Determine which nodes were _requested_ to run from the command line
        final var cliNodesToRun = commandLineArgs.localNodesToStart();
        // Determine which nodes are _configured_ to run from the config file(s)
        final var configNodesToRun =
                platformConfig.getConfigData(BasicConfig.class).nodesToRun();
        // Using the requested nodes to run from the command line, the nodes configured to run, and now the
        // address book on disk, reconcile the final list of nodes to run
        final List<NodeId> nodesToRun = getNodesToRun(diskAddressBook, cliNodesToRun, configNodesToRun);
        // Finally, verify that the reconciliation of above node IDs yields exactly one node to run
        final var selfId = ensureSingleNode(nodesToRun);
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/info/DiskStartupNetworks.java
```diff
@@ -301,7 +301,8 @@ private Optional<Network> genesisNetworkFromConfigTxt(@NonNull final Configurati
             final var configFile = LegacyConfigPropertiesLoader.loadConfigFile(Paths.get(DEFAULT_CONFIG_FILE_NAME));
             try {
                 legacyBook = configFile.getAddressBook();
-                CryptoStatic.initNodeSecurity(legacyBook, platformConfig);
+                // load the public keys into the address book.  No private keys need to be loaded.
```
mhess-swl, line 304: Nit:

```suggestion
                // Load the public keys into the address book. No private keys should be loaded!
```

hedera-node/hedera-app/src/test/java/com/hedera/node/app/ServicesMainTest.java
```diff
@@ -1,142 +0,0 @@
...
-import java.util.ArrayList;
-import java.util.List;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
-import org.mockito.MockedStatic;
-import org.mockito.junit.jupiter.MockitoExtension;
-
-@ExtendWith(MockitoExtension.class)
-final class ServicesMainTest {
```
mhess-swl, line 47: Why are the tests in this class deleted? They still seem relevant, even if they need to be reworked

mhess-swl (commented):
Mostly looks good from services, aside from some deleted tests. There are also a couple of cosmetic changes suggested for posterity

tinker-michaelj (approved):
I like @mhess-swl's suggested comments but otherwise LGTM, tyvm @edward-swirldslabs !

nathanklick (approved):

tinker-michaelj (approved):
LGTM, ty @edward-swirldslabs !

hedera-node/infrastructure/docker/containers/production-next/consensus-node/entrypoint.sh
edward-swirldslabs: @dalvizu  I am resolving this comment since there has been no reply

andrewb1269hg (approved):
Review and approve files associated to `devops-ci` ownership.

PR #12114, opened 2024-03-13
fix: wire `HashingOutputStream` into sidecar writing

### LINKED ISSUE #12113: (mod) `HashingOutputStream` not wired into `SidecarWriterV6` stream
### Description

I looked at the sidecar hashes in a `List<SidecarMetadata>` from a mod-service `RecordStreamFile` and all were the SHA-384 hash of the empty string.

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12113 
 - Wire the `HashingOutputStream` into the chained stream used by `SidecarWriterV6`.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v6/SidecarWriterV6.java
```diff
@@ -152,6 +159,7 @@ List<SidecarType> types() {
     @Override
     public void close() throws IOException {
         outputStream.close();
+        hashingDelegateStream.close();
```
steven-sheehy, line 162: For compressFile=false, outputStream.close() will also close hashingDelegateStream since they're both using fout. This could cause an exception on close due to already closed. Should add a test to verify. Think the solution is not not set hashingDelegateStream when compressFile=false and check if null in close.

tinker-michaelj: You would expect this given the normal behavior of `OutputStream` wrapper classes, which propagate a received `close()` to their delegate.

However, `HashingOutputStream` [does not override `close()`](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/HashingOutputStream.java); and is in fact a no-op.

So what happens with `compressFile=false` is:
 1. `WritableStreamingData` receives `close()`, propagates to `bout`
 2. `BufferedOutputStream` receives `close()`, propagates to `hashingOutputStream`
 3. _<close chaining stops>_
 4. We explicitly call `fout.close()` and now invoke the bottom `FileOutputStream`.

Neeharika-Sompalli (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v6/SidecarWriterV6.java
steven-sheehy: Ah, I see. I should've read your code comment.

steven-sheehy (approved):

PR #12130, opened 2024-03-14
fix: (mod-service) Fix a case for identifying `RESTART` scenario

Neeharika-Sompalli:
Related to https://github.com/hashgraph/hedera-services/issues/11582

Cherry-pick https://github.com/hashgraph/hedera-services/pull/12109


tinker-michaelj (approved):

PR #12134, opened 2024-03-14
fix: encode zero allowance in success response for non-fungible token

### LINKED ISSUE #12127: (mod) `allowance()` system contract returns wrong tuple length when called with non-fungible token type
### Description

I called `allowance(<token>, <owner>, <spender>)` with a non-fungible token type.

Instead of returning `(SUCCESS, 0)` it returned just `(SUCCESS)`; and thus decoding the output reverts the contract.

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12127 
 - Whenever returning a `COMPLETED_SUCCESS` precompile result from the `AllowanceCall`, encode a length-2 tuple.
 - Previously mirror node acceptance test would revert when calling [this method](https://github.com/hashgraph/hedera-mirror-node/blob/e4a28be9d61a3d40280dfe628bb992614adab2af/hedera-mirror-test/src/test/resources/solidity/contracts/EstimatePrecompileContract.sol#L370) with a non-fungible token type using a non-static call.


lukelee-sl (approved):
LGTM.   Thank you!

PR #12136, opened 2024-03-14
fix: always encode length-2 output for `AllowanceCall` success

tinker-michaelj:
**Description**:
 - Port #12134


lukelee-sl (approved):
LGTM

PR #12149, opened 2024-03-14
feat: (cherry-pick) Metadata HIPs 

Neeharika-Sompalli:
Cherry Pick #11766


hedera-node/hapi/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ description = "Hedera API"
 
 // Add downloaded HAPI repo protobuf files into build directory and add to sources to build them
 tasks.cloneHederaProtobufs {
-    branchOrTag = "add-pbj-types-for-state"
+    branchOrTag = "main"
```
mhess-swl, line 28: I agree with @netopyr that this is a dangerous thing to do. But, while the 'tag' part of this doesn't seem to work...

iwsimon (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/HapiTxnOp.java
```diff
@@ -577,6 +577,16 @@ public T memo(String text) {
         return self();
     }
 
+    public T blankMetadata() {
+        memo = Optional.of("");
```
mhess-swl, line 581: Should this be `metadata` instead of `memo`?

hedera-node/hapi-fees/src/main/java/com/hedera/node/app/hapi/fees/usage/token/TokenGetInfoUsage.java
```diff
@@ -85,4 +90,9 @@ public TokenGetInfoUsage givenCurrentlyUsingAutoRenewAccount() {
         addRb(BASIC_ENTITY_ID_SIZE);
         return this;
     }
+
+    public TokenGetInfoUsage givenCurrentMetadata(final String metadata) {
+        addRb(metadata.length());
```
mhess-swl, line 95: Do we want to add a comment here explaining why this is so different than the fees for other messages? Just a thought

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenHandlers.java
```diff
@@ -99,7 +100,8 @@ public TokenHandlers(
             @NonNull final TokenGetInfoHandler tokenGetInfoHandler,
             @NonNull final TokenGetAccountNftInfosHandler tokenGetAccountNftInfosHandler,
             @NonNull final TokenGetNftInfoHandler tokenGetNftInfoHandler,
-            @NonNull final TokenGetNftInfosHandler tokenGetNftInfosHandler) {
+            @NonNull final TokenGetNftInfosHandler tokenGetNftInfosHandler,
+            TokenUpdateNftsHandler tokenUpdateNftsHandler) {
```
mhess-swl, line 104: Nit: `@NonNull final`

JeffreyDallas (approved):

netopyr (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/HapiTxnOp.java
netopyr: I think @mhess-swl is right. Please double-check and eventually fix it in a follow-up PR.

hedera-node/hapi/build.gradle.kts
netopyr: Let's fix this as soon as the build is fixed and tags are allowed again.

hedera-node/hapi-fees/src/main/java/com/hedera/node/app/hapi/fees/usage/token/TokenGetInfoUsage.java
netopyr: @mhess-swl Is there something specific you are missing? I am afraid properly documenting the fee system is a rabbit hole. We should revisit once modularization is done.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenHandlers.java
netopyr: Good catch! @thomas-swirlds-labs @Neeharika-Sompalli Please fix this in a follow-up PR.

hedera-node/hapi-fees/src/main/java/com/hedera/node/app/hapi/fees/usage/token/TokenGetInfoUsage.java
mhess-swl: I'm referring to the fact that the pricing model for this endpoint is significantly different than the pricing models in other endpoints

PR #12155, opened 2024-03-14
feat(ReconnectBench): reuse generated state between invocations

### LINKED ISSUE #11816: Reuse generated state between iterations in ReconnectBench

anthony-swirldslabs:
**Description**:
This fix makes use of the `numFiles` parameter in the ReconnectBench, allowing the newly introduced `storageOptimizer` in the `StateBuilder` to optimize the map storage - in particular, to call the `copyMap()` method which helps the benchmark to generate larger states.

The ultimate change is the use of the new `saveMaps()` and `restoreMap(label)` methods that allow the benchmark to generate a state in the beginning and save it, and then restore it for every individual invocation and avoid re-generating the same state over and over.

I've spent some time playing with the `numRecords` and `numFiles` parameters, as well as with the `-Xmx` value. However, the state size maxes out at 50 M, and anything larger than that leads to JVM exiting with code 137 (related to memory allocation) or outright OOMs. Initially I thought that maybe the `-Xmx` could help increase the maximum state size, however, even a doubled value wouldn't help at all. I also tried to insert `System.gc()` in various locations, but that didn't help either. At this point, I ran out of ideas of what else could be toggled to help increase the maximum supported state size. Suggestions are welcome (depending on their nature, they may or may not make it into this PR, or be planned for future, separate PRs.)

Anyways, with the current settings the benchmark is able to work with a state of size 50 M, which is already a 10x improvement over the original 5M state.

**Related issue(s)**:

Fixes #11816

**Notes for reviewer**:
```
$ gr jmhReconnect
...
# Parameters: (keySize = 8, maxKey = 1000000, numFiles = 500, numRecords = 100000, numThreads = 32, randomSeed = 9823452658, recordSize = 128, teacherAddProbability = 0.05, teacherModifyProbability = 0.05, teacherRemoveProbability = 0.05)

# Run progress: 0.00% complete, ETA 00:01:20
# Fork: 1 of 1
# Warmup Iteration   1: 2024-03-14 15:02:19.195 Benchmark configuration: BenchmarkConfig[benchmarkData=data,saveDataDirectory=false,verifyResult=true,enableSnapshots=false,printHistogram=false,csvOutputFolder=data,csvMetricsFileName=BenchmarkMetrics.csv,csvMetricNamesFileName=BenchmarkMetricNames.csv,csvWriteFrequency=1000,csvAppend=true,deviceName=sda]
2024-03-14 15:02:19.197 Build: 0.49.0-SNAPSHOT (849aaa7)
Building a state of size 50,000,000
Copying files for i = 100,000
Copying files for i = 200,000
...
Benchmark                 (keySize)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (teacherAddProbability)  (teacherModifyProbability)  (teacherRemoveProbability)  Mode  Cnt    Score   Error  Units
ReconnectBench.reconnect          8   1000000         500        100000            32    9823452658           128                     0.05                        0.05                        0.05  avgt    7  124.219 ± 5.629   s/op

BUILD SUCCESSFUL in 26m 38s
```

**Checklist**


platform-sdk/swirlds-benchmarks/build.gradle.kts
```diff
@@ -54,5 +54,6 @@ tasks.register<JMHTask>("jmhReconnect") {
 
     resultsFile.convention(layout.buildDirectory.file("results/jmh/results-reconnect.txt"))
 
-    benchmarkParameters.put("numRecords", listProperty("5000000"))
+    benchmarkParameters.put("numRecords", listProperty("100000"))
```
jsync-swirlds, line 57: Really commenting a few lines up, line 53...

Have you tried a large heap (32G or 64G, or more, and set both Xms and Xmx) with a different GC, such as the ZGC we use in production?  The default G1 GC may struggle with large heap sizes.
Here is the option set we nominally use in production (NMT may be slightly different, but this is what's "requested").
`-Xmx100g -Xms32g -XX:+UnlockExperimentalVMOptions -XX:+UseZGC -XX:ConcGCThreads=14 -XX:ZMarkStackSpaceLimit=16g -XX:+UseLargePages -XX:MaxDirectMemorySize=8g -cp swirlds.jar:data/lib/* com.swirlds.platform.Browser > /opt/hgcapp/services-hedera/HapiApp2.0/stdout.log 2>&1
`

anthony-swirldslabs: Thanks for the suggestion. I'm trying

`jvmArgs.set(listOf("-Xmx32g", "-Xms32g", "-XX:+UnlockExperimentalVMOptions", "-XX:+UseZGC", "-XX:ConcGCThreads=5", "-XX:+UseLargePages", "-XX:MaxDirectMemorySize=8g"))`

and the JVM just comes to silence during the warm-up iteration after generating the state, and no progress is happening at all. Looks like a thread executor dies somewhere silently.

I'm unsure if this is the result of a merge from upstream that I did just before publishing the PR, or the result of the new JVM parameters. I'll be playing with this next week.

For now, the fix stands as it is in this PR.

imalygin (approved):

OlegMazurov (dismissed):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -71,62 +71,76 @@ public class ReconnectBench extends VirtualMapBaseBench {
...
                         teacherAddProbability,
                         teacherRemoveProbability,
                         teacherModifyProbability,
-                        buildVMPopulator(teacherMap),
-                        buildVMPopulator(learnerMap));
+                        buildVMPopulator(teacherRef),
+                        buildVMPopulator(learnerRef),
+                        i -> {
+                            if (i % numRecords == 0) {
+                                System.err.printf("Copying files for i = %,d\n", i);
```
artemananiev, line 121: Debug output?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -71,62 +71,76 @@ public class ReconnectBench extends VirtualMapBaseBench {
...
+                new AtomicReference<>(createEmptyMap("teacher"));
+        final AtomicReference<VirtualMap<BenchmarkKey, BenchmarkValue>> learnerRef =
+                new AtomicReference<>(createEmptyMap("learner"));
 
         final Random random = new Random(randomSeed);
         new StateBuilder<>(BenchmarkKey::new, BenchmarkValue::new)
                 .buildState(
                         random,
-                        numRecords,
+                        (long) numRecords * (long) numFiles,
```
artemananiev, line 113: No need to cast both to longs, one cast should be enough.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
```diff
@@ -167,6 +196,7 @@ protected VirtualMap<BenchmarkKey, BenchmarkValue> copyMap(
      */
     protected VirtualMap<BenchmarkKey, BenchmarkValue> flushMap(
             final VirtualMap<BenchmarkKey, BenchmarkValue> virtualMap) {
+        logger.info("Flushing map {}...", virtualMap.getLabel());
```
artemananiev, line 199: Debug output?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -71,62 +71,76 @@ public class ReconnectBench extends VirtualMapBaseBench {
...
+                saveMaps(List.of(teacherRef.get(), learnerRef.get()));
+        mapCopies.forEach(this::releaseAndCloseMap);
+    }
 
+    /** Restore the saved state from disk as a new test on-disk copy for each iteration. */
+    @Setup(Level.Invocation)
+    public void setupInvocation() {
+        updateMerkleDbPath();
+
+        teacherMap = restoreMap("teacher");
```
artemananiev, line 140: `restoreMap()` may return null, if no previously saved state is found... I may be wrong, but looking into code, I'd expect an NPE in this case

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
```diff
@@ -71,62 +71,76 @@ public class ReconnectBench extends VirtualMapBaseBench {
...
+                                teacherRef.set(copyMap(teacherRef.get()));
+                                learnerRef.set(copyMap(learnerRef.get()));
+                            }
+                        });
+
+        teacherRef.set(flushMap(teacherRef.get()));
+        learnerRef.set(flushMap(learnerRef.get()));
+
+        final List<VirtualMap<BenchmarkKey, BenchmarkValue>> mapCopies =
+                saveMaps(List.of(teacherRef.get(), learnerRef.get()));
```
artemananiev, line 131: There is a benchmark config named `saveDataDirectory`. It makes sense to check it here

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
```diff
@@ -227,6 +257,50 @@ protected void verifyMap(long[] map, VirtualMap<BenchmarkKey, BenchmarkValue> vi
         }
     }
 
+    protected List<VirtualMap<BenchmarkKey, BenchmarkValue>> saveMaps(
```
artemananiev, line 260: I wonder if the existing `saveMap()` method may be used instead of adding this new method?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
```diff
@@ -67,8 +71,13 @@ public void buildState(
             final double teacherRemoveProbability,
             final double teacherModifyProbability,
             final BiConsumer<K, V> teacherPopulator,
-            final BiConsumer<K, V> learnerPopulator) {
+            final BiConsumer<K, V> learnerPopulator,
+            final Consumer<Long> storageOptimizer) {
+        System.err.printf("Building a state of size %,d\n", size);
```
artemananiev, line 76: Debug output?

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: I left it here intentionally. This output helps track the progress of building a state which is useful for larger states. E.g. the current 50M state takes around 10-15 minutes to build on my Mac, so it's nice to have a visual feedback when running this benchmark.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: Sure, I can remove one of the casts.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
anthony-swirldslabs: Similar to above, this extra output helps track the progress of the benchmark. Also, this one is simply improving the already existing debug output by adding an opening log in addition to the closing one. This way, if the operation freezes, one knows what was going on just before.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: That's a good catch. I'll handle this (although the bench would still fail with an exception.)

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: While the parameter is undocumented, I believe its purpose if clear from reading

https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/BaseBench.java#L164-L170

Basically it controls whether the benchmark would remove the data directory **after** the benchmark is complete. So it seems irrelevant to the saveMaps() calls - the ReconnectBench would still call this method regardless of the boolean value of that config parameter.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/VirtualMapBaseBench.java
anthony-swirldslabs: No, and this is a part of the reason why the fix took a bit of time. I tried using the existing method and after a couple of days of trying synced up with Oleg, spent some time debugging it, and figured it's designed to save a single map only. The ReconnectBench needs two maps, and the existing method would save them separately. This would cause the existing restoreMap() logic to create a MerkleDb that only knows about the first map being restored and refusing to restore the second map.

After evaluating options, I decided to implement the new `saveMaps` method that allows one to save several maps and be able to restore them using the existing code.

We could consider removing the legacy `saveMap` at a later time because the new one can replace it.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/StateBuilder.java
anthony-swirldslabs: Same as above. This is left here on purpose.

imalygin (approved):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
artemananiev: If `saveDataDirectory` is true, a copy is saved after each test iteration, not just after the benchmark is complete

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
artemananiev: Why would it fail with an exception? `restoreMap()` returns null, if there is nothing to restore, i.e. this is the first run without a saved state. This looks like a valid case to me. In this case, a new map should be created

nathanklick (approved):
Approved w.r.t the Gradle changes only.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: I'm unable to find this particular logic. I see that flushMap() will saveMap() if it's true though.

However, the default value for the `saveDataDirectory` parameter is false as seen at:

https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/config/BenchmarkConfig.java#L53

So this logic is never active by default. And the ReconnectBench doesn't use it explicitly, and doesn't recommend changing its default value.

So I still think that this parameter is completely irrelevant to the ReconnectBench. At least for now - until/unless we start using it in this particular benchmark.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
anthony-swirldslabs: This `restoreMap` call occurs in a `@Setup(Level.Invocation)` which runs before every invocation.

However, the benchmark is currently designed to rely on a prior `@Setup public void setupBenchmark()` call, which is implicitly `@Setup(Trial)` that runs before the entire benchmark. That's where the map is being saved. If this method fails, the the benchmarks is supposed to fail because it's not designed to bypass this first @Setup method.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/ReconnectBench.java
artemananiev: No, it is relevant. The logic is exactly the same as in virtual map benchmark: save the state, so the benchmark can be started from that state next time rather than from scratch

anthony-swirldslabs: I filed https://github.com/hashgraph/hedera-services/issues/12194 to address this at a later time when this feature becomes more relevant (which depends on the ability to generate much-much larger states.)

PR #12187, opened 2024-03-18
chore: Remove unnecessary deleted flag in `TokenRelation`

Neeharika-Sompalli:
Remove unnecessary deleted flag in `TokenRelation` state PBJ object. 

Protobufs changes are https://github.com/hashgraph/hedera-protobufs/pull/330


tinker-michaelj (dismissed):

jeromy-cannon (dismissed):

hedera-node/hapi/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ description = "Hedera API"
 
 // Add downloaded HAPI repo protobuf files into build directory and add to sources to build them
 tasks.cloneHederaProtobufs {
-    branchOrTag = "main"
+    branchOrTag = "remove-deleted-from-token-relation"
```
netopyr, line 28: I think we should just ensure the change in the protobuf repository is merged into `main` first and leave this value as it is.

Neeharika-Sompalli: @netopyr Yes, that is an option. 
But we will break `release/0.48` compilation until this PR is merged, if we merge protobufs PR first. 😃 
If we merge this PR first and then merge protobufs, we need to open a new PR changing to `main` on release/0.48. 
Do you have a preference?

kimbor (approved):
lgtm!

hedera-node/hapi/build.gradle.kts
netopyr: No preference. I'd be ok with preparing everything and merging the two PRs quickly, one after the other. But we can also play it safe and do it the correct way.

Neeharika-Sompalli: Makes sense. I will open PR to change it back to `main` immediately after this is merged. Just so, no one has issues with compilation.

iwsimon (approved):
LGTM

jeromy-cannon (approved):

PR #12196, opened 2024-03-18
fix: `numPositiveBalances` mgmt

tinker-michaelj:
**Description**:
 - Allow `TopicUpdate` to use an empty submit key.
 - In `TokenMintHandler`, ensure the updated treasury account has incremented `numberPositiveBalances`.
 - In `TokenUpdateHandler`, ensure the previous treasury's `numberPositiveBalances` is decremented.


Neeharika-Sompalli (dismissed):
LGTM!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
```diff
@@ -214,6 +211,7 @@ private List<Long> mintNonFungible(
 
         // Change the supply on token
         changeSupply(token, treasuryRel, metadataCount, FAIL_INVALID, accountStore, tokenStore, tokenRelStore);
+        treasuryAccount = accountStore.get(treasuryRel.accountIdOrThrow());
```
mhess-swl, line 214: Why assign `treasuryAccount` twice?

Neeharika-Sompalli: We are updating `treasuryAccount` in `changeSupply` and were not getting latest account from store again.
This was causing `numPositiveBalances` to be set incorrectly.

iwsimon (dismissed):
LGTM!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenMintHandler.java
netopyr: We should add a comment. I was confused, too.

Neeharika-Sompalli: Yes, will add now.

Neeharika-Sompalli: Added comments. Thanks @mhess-swl @netopyr

iwsimon (dismissed):
LGTM

steven-sheehy (commented):
Shouldn't there be tests that verify these changes? Especially at the Hapi spec level as we may refactor units

tinker-michaelj:
> Shouldn't there be tests that verify these changes? Especially at the Hapi spec level as we may refactor units

Added two EETs [here](https://github.com/hashgraph/hedera-services/pull/12196/files#diff-e66f78f615a9ef8db967bb09893c31958771b5d258f471ac29421e0ffcae1e77R131).

Neeharika-Sompalli (approved):
LGTM !

PR #12200, opened 2024-03-18
feat: Add maximum entity count metrics

netopyr:
This PR adds metrics that report the configured maximum number of entities. This will allow us to calculate the utilization.


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

Issue #11882, opened 2024-03-04
Accommodate any node removals in `configAddressBook` on `updateWeight()` call


Neeharika-Sompalli:
When testing with removal of nodes, some tests caught this error. 
Accommodate the assumption for removal of nodes in `configAddressBook` on call to `updateweight()`
Fix this in both mono and modular service

```
2024-03-04 18:04:55.933 63       ERROR EXCEPTION        <main> Browser: Unable to start Browser
java.util.NoSuchElementException: no address with id 29 exists
	at com.swirlds.platform.system.address.AddressBook.getAddress(AddressBook.java:333) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.system.address.AddressBook.updateWeight(AddressBook.java:390) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.hedera.node.app.service.mono.ServicesState.lambda$updateWeight$0(ServicesState.java:318) ~[app-service-mono-0.46.3.jar:?]
	at java.base/java.util.Map.forEach(Map.java:733) ~[?:?]
	at com.hedera.node.app.service.mono.state.adapters.MerkleMapLike$1.forEach(MerkleMapLike.java:75) ~[app-service-mono-0.46.3.jar:?]
	at com.hedera.node.app.service.mono.ServicesState.updateWeight(ServicesState.java:314) ~[app-service-mono-0.46.3.jar:?]
	at com.swirlds.platform.state.address.AddressBookInitializer.initialize(AddressBookInitializer.java:217) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.state.address.AddressBookInitializer.<init>(AddressBookInitializer.java:137) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.PlatformBuilder.build(PlatformBuilder.java:245) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.Browser.launchUnhandled(Browser.java:185) ~[swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.Browser.launch(Browser.java:129) [swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.Browser.parseCommandLineArgsAndLaunch(Browser.java:103) [swirlds-platform-core-0.46.3.jar:?]
	at com.swirlds.platform.Browser.main(Browser.java:92) [swirlds-platform-core-0.46.3.jar:?]
[...]

PR #12122, opened 2024-03-14
fix: Fix `updateWeight` for removed nodes

Neeharika-Sompalli:
Fixes #11882 

Since it is possible for nodes to be removed in `configAddressBook` passed into `updateWeight()` call from platform, we need to check if the nodeId exists in `configAddressBook` and update weights only for them.

PR #12216, opened 2024-03-19
fix: (cherry-pick) Fix updateWeight for removed nodes

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/11882
Cherry-pick https://github.com/hashgraph/hedera-services/pull/12122

Since it is possible for nodes to be removed in configAddressBook passed into updateWeight() call from platform, we need to check if the nodeId exists in configAddressBook and update weights only for them.


jeromy-cannon (approved):

iwsimon (approved):
LGTM

tinker-michaelj (approved):

PR #12228, opened 2024-03-19
fix: make `HederaSoftwareVersion` match Services upgrade semantics

### LINKED ISSUE #12217: (mod) Support config-only updates via `hedera.config.version`
### Problem

We still want to be able to re-deploy the same JAR in a way that will be recognized as a software upgrade. (For example, if we want to change the address book before completion of dynamic address book.)

### Solution

Replicate mono-service behavior of replacing the semver `build` part with the `hedera.config.version` value whenever it is overridden.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #12217 
 - Only use a `pre` semver part if it is an `alpha-X` literal.
 - Ignore all `build` parts from environment; use **only** if `hedera.config.version` is overridden to non-zero value.


mhess-swl (approved):

Neeharika-Sompalli (approved):
Makes sense to me. Thanks @tinker-michaelj

PR #12275, opened 2024-03-21
feat: clickable wiring diagram

### LINKED ISSUE #12274: Clickable wiring diagram
Add hyperlinks to the wiring diagram. When you click on a component, it should take you to the source code for that component on github.

cody-littley:
closes #12274


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
```diff
@@ -113,6 +140,7 @@ public static PlatformSchedulers create(
                         .withOnRamp(hashingObjectCounter)
                         .withExternalBackPressure(true)
                         .withMetricsBuilder(model.metricsBuilder().withUnhandledTaskMetricEnabled(true))
+                        .withHyperlink(platformCoreHyperlink(EventHasher.class))
```
cody-littley, line 143: A note to reviewers: this will become 100% automated once we construct schedulers automatically inside the proxy `ComponentWiring` classes (this is a planned feature).

edward-swirldslabs (approved):

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/WiringFlowchart.java
```diff
@@ -238,9 +243,17 @@ private void collapseGroup(@NonNull final GroupVertex group) {
         final List<ModelEdge> edges = collectEdges();
         final List<ModelVertex> groupVertices = collectGroupVertices(group);
 
+        final Set<String> hyperlinks = new HashSet<>();
+        for (final ModelVertex vertex : groupVertices) {
+            if (vertex.getHyperlink() != null) {
+                hyperlinks.add(vertex.getHyperlink());
+            }
+        }
+        final String hyperlink = hyperlinks.size() == 1 ? hyperlinks.iterator().next() : null;
```
anthony-swirldslabs, line 252: Minor: we go into the `else` branch if there's either zero or more than one element in the set (also, sets are unsorted, btw.) I'm not entirely sure if this is the desired logic.

If we're certain there's either zero or 1 element, then we could use something like `MoreCollectors.toOptional()` to enforce the contract and make the code robust.

If, on the other hand, more than 1 elements is allowed, then it's unclear why an unsorted set is used, and also why this is treated the same as if there's no elements at all. In this case, either the logic should be fixed, or some clarifying comments need to be added.

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/WiringFlowchart.java
cody-littley: This is intentional. When grouping things on a graph, the parent group should only have a hyperlink from a child if there is exactly one hyperlink within its set of children. If there are no child hyperlinks then there is obviously nothing we can do. But if there are multiple child hyperlinks, we have no way to decide which child hyperlink to choose. In the indeterminate case, better to have no hyperlink than to have a confusing and potentially misleading hyperlink.

anthony-swirldslabs: Makes sense. Thanks for clarifying this. However, like I mentioned, it would be best to add a comment about this into the code above this logic so that future maintainers of the code understand the intention.

alittley (commented):
<mistakenly submitted comment on wrong PR>

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java
```diff
@@ -16,19 +16,46 @@
 
 package com.swirlds.platform.wiring;
 
+import static com.swirlds.common.wiring.model.HyperlinkBuilder.platformCommonHyperlink;
+import static com.swirlds.common.wiring.model.HyperlinkBuilder.platformCoreHyperlink;
+
 import com.swirlds.common.context.PlatformContext;
+import com.swirlds.common.stream.EventStreamManager;
```
anthony-swirldslabs, line 23: This fails with

```
/runner/_work/hedera-services/hedera-services/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulers.java:23: error: cannot find symbol
import com.swirlds.common.stream.EventStreamManager;
                                ^
  symbol:   class EventStreamManager
  location: package com.swirlds.common.stream
```

https://github.com/hashgraph/hedera-services/actions/runs/8381299589/job/22952492704#step:6:4673

anthony-swirldslabs: I take that back. This seems to be a regression of https://github.com/hashgraph/hedera-services/commit/6e1dc367c629e212932e4163d691b1f64e69329c.

PR #12282, opened 2024-03-21
chore: Increase max token associations to 50M

### LINKED ISSUE #12094: Move configuration from application.properties to code
The `application.properties` file used in MainNet contained a number of configurations. These need to become the new defaults (if they are not already), and the `application.properties` file has to be cleared.

### LINKED ISSUE #12218: Increase max. number of token associations
Increase the maximum number of token associations to 50 million.

thomas-swirlds-labs:
**Description**:

Increase max token relations / associations to 50M for release 48 (has been decided as the number required).
Sync properties with values on MainNet and TestNet
Set properties for PreviewNet

**Related issue(s)**:

Fixes #12218 
Fixes #12094 

**Notes for reviewer**:


iwsimon (dismissed):
LGTM

netopyr (dismissed):

poulok (dismissed):

Nana-EC (dismissed):

rbair23 (dismissed):

dalvizu (dismissed):

hedera-node/configuration/previewnet/application.properties
```diff
@@ -1 +1,4 @@
-
+bootstrap.genesisPublicKey=c249a323c878f5b5e2daccda6d731e6fdc32f870228d1cd4fae559d947dbc36c
+contracts.chainId=297
+ledger.id=0x02
+contracts.evm.version.dynamic=true
```
netopyr, line 4: The properties in this file were copied from `hedera-node/configuration/previewnet/bootstrap.properties`.

File `0.0.121` on previewnet appears to be empty

hedera-node/configuration/testnet/application.properties
```diff
@@ -4,7 +4,6 @@
 # It does not have to match bootstrap.properties once that file is deprecated by the switch from mono-service to hedera-config.
 
 #Overrides that differ based on the network
-ledger.id=0x01
-contracts.chainId=296
 bootstrap.genesisPublicKey=e06b22e0966108fa5d63fc6ae53f9824319b891cd4d6050dbf2b242be7e13344
-contracts.knownBlockHash=
+contracts.chainId=296
+ledger.id=0x01
```
netopyr, line 9: `contracts.knownBlockHash` is not used anymore. Reordered the other properties to match file `0.0.121` on testnet.

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -33,7 +33,8 @@ public record ContractsConfig(
         // @ConfigProperty(defaultValue = "") KnownBlockValues knownBlockHash,
         @ConfigProperty(value = "keys.legacyActivations", defaultValue = "1058134by[1062784]")
                 String keysLegacyActivations,
-        @ConfigProperty(value = "localCall.estRetBytes", defaultValue = "32") @NetworkProperty int localCallEstRetBytes,
+        @ConfigProperty(value = "localCall.estRetBytes", defaultValue = "4096") @NetworkProperty
+                int localCallEstRetBytes,
```
netopyr, line 37: This value was stored in file `0.0.121` on mainnet.

netopyr (dismissed):

iwsimon (dismissed):
LGTM.

Nana-EC (dismissed):

hedera-node/configuration/previewnet/application.properties
lukelee-sl, line 4: I wonder if there should be a similar comment as in the beginning of the testnet version of this file?  Also, is there a mainnet version of this file yet?

david-bakin-sl, line 4: We need to learn why this is set in previewnet - it appears only useful in test suites where you can override, for one test, the evm version used.

david-bakin-sl (changes requested):
Questioning setting `contracts.evm.version.dynamic=true` in previewnet.  "request changes" because I think this needs resolution.

hedera-node/configuration/previewnet/application.properties
```diff
@@ -1 +1,9 @@
+# Except for network-specific properties (e.g. previewnet vs. mainnet), this file is intended to be an
+# exact replica of file 121 on testnet and should match bootstrap.properties exactly.
+# It will be used by modular code for all the property overrides.
+# It does not have to match bootstrap.properties once that file is deprecated by the switch from mono-service to hedera-config.
 
+#Overrides that differ based on the network
+bootstrap.genesisPublicKey=c249a323c878f5b5e2daccda6d731e6fdc32f870228d1cd4fae559d947dbc36c
+contracts.chainId=297
+ledger.id=0x02
```
netopyr, line 9: Added comments similar to `testnet/application.properties`. 
Removed `contracts.evm.version.dynamic` as requested by the smart contract team.
Otherwise, this file is an exact copy of the sibling `bootstrap.properties` file (minus the properties that match mainnet).

netopyr (approved):

david-bakin-sl (changes requested):
LGTM!

To preserve for posterity:  Turns out `contracts.evm.version.dynamic` is broken in that in only changes contracts' call processing and not other overrides, e.g., opcodes.  Thus it leaves you inconsistent.  There's no way to use it in production, and we'll have to get rid of it (or make it work properly).

lukelee-sl (approved):
LGTM

SimiHunjan (approved):

poulok (approved):

iwsimon (approved):
LGTM

Nana-EC (approved):

hedera-node/configuration/testnet/application.properties
dalvizu: Should we issue this update immediately after this is merged?

dalvizu:
I see there are no changes to mainnet application.properties - there was a PR (an earlier version of this PR) to match the actual 0.0.121l, which is not empty. Is the intent of this that during the 0.48 upgrade we should issue an empty application.properties?

hedera-node/configuration/testnet/application.properties
netopyr: No, we need to wait until 0.48 was deployed. I have not checked the code of 0.47.

netopyr:
Yes, this is the intent. Once we have done this, we can follow the new process for application.properties changes.

dalvizu (approved):

PR #12285, opened 2024-03-21
refactor: proxy wiring for EventWindowManager

### LINKED ISSUE #12257: Use proxy wiring for `EventWindowManagerWiring`
This will need to use the new `PassThroughWiring` that was created in this PR: https://github.com/hashgraph/hedera-services/pull/12192

edward-swirldslabs:
**Description**

This PR creates a component interface and default implementation of the `EventWindowManager` for use with proxy wiring. 

Fixes #12257 

**Notes for reviewer**:
* The wiring and default implementation inputs are unit tested. 
* Consensus Tests pass with using the new code. 
* Manual launch of a local test network works.


alittley (dismissed):

cody-littley:
friendly reminder to mark yourself as asignee

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
```diff
@@ -253,7 +253,13 @@ public PlatformWiring(@NonNull final PlatformContext platformContext) {
...
 
         gossipWiring = GossipWiring.create(model);
-        eventWindowManagerWiring = EventWindowManagerWiring.create(model);
+        eventWindowManagerWiring = new ComponentWiring<>(
+                model,
+                EventWindowManager.class,
+                model.schedulerBuilder("eventWindowManager")
+                        .withType(TaskSchedulerType.DIRECT_THREADSAFE)
+                        .build()
+                        .cast());
```
cody-littley, line 262:
```suggestion
                model.schedulerBuilder("eventWindowManager")
                        .withType(TaskSchedulerType.DIRECT_THREADSAFE)
                        .withHyperlink(platformCoreHyperlink(EventWindowManager.class))
                        .build()
                        .cast());
```

cody-littley (dismissed):

alittley (dismissed):

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultEventWindowManager.java
```diff
@@ -0,0 +1,45 @@
...
+        return round.getNonAncientEventWindow();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @NonNull
+    public NonAncientEventWindow setEventWindow(@NonNull final NonAncientEventWindow eventWindow) {
+        return eventWindow;
```
anthony-swirldslabs, line 43: Minor: it's a little confusing when a setter doesn't actually set anything. This would certainly be a surprise for anyone reading or debugging this code. It might be worthwhile to add a comment to the javadoc of this method (or the class) implementation to clarify this unusual behavior and reasons behind it.

anthony-swirldslabs (approved):

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultEventWindowManager.java
lpetrovic05: I think its better to change the name of the method, I suggest `updateEventWindow`

lpetrovic05 (approved):

cody-littley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultEventWindowManager.java
edward-swirldslabs: method has been renamed.

alittley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultEventWindowManager.java
anthony-swirldslabs: Renaming a method is an option, indeed. However, it's still confusing for a method that's supposed to update something not to update anything and just return its input argument verbatim. I'd still be surprised to be debugging this code, assuming the method updates something, and then discover it actually doesn't do anything at all. I still think that either the implementation of the method here is wrong or requires clarifying comments in code, or the method's name and javadoc in the defining interface is inaccurate, or at the very minimum confusing.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultEventWindowManager.java
edward-swirldslabs: I understand the confusion as this component has no internal state that it holds onto.   All of our components are stages on a pipeline, most with internal state that needs to impact the next item in the pipeline.   Some of our components do not have internal state, they are "pass through" or entry points into the pipeline.   In this case the component is both a pass through in one data flow and an entry point in the pipeline for a differently sourced data flow.  

Calling it "set" or "update" is about more of the psychology external to the pipeline, the activity of the agent that is calling the method and their intended effect on the system all along the pipeline.   So while the is a  singular empty component with no internal state, it is the gateway into a pipeline of other components and that is what the method is named for, its down stream effects.

edward-swirldslabs: If there is a better name, happy to hear it, but I don't think we can avoid this situation of having vacuous methods and components that are internally stateless and semantically reflect effects and needs happening in other components.

anthony-swirldslabs: Got it. Thanks for the clarification. That makes sense as far as the implementation is concerned, but it still looks a bit confusing.

While I'm not suggesting this for this PR, but perhaps having different interfaces for entry points and internal components with semantically different public methods could eliminate the confusion. Alternatively, if calling this method is not expected on this particular implementation, then perhaps this implementation could throw an `UnsupportedOperationException` instead of just returning a value.  If none of this is an option, then like I mentioned above, perhaps adding a comment into the javadoc of this overridden method an stating that it doesn't actually set or update anything by design might be a good idea. But you can add this clarifying comment in a future PR if you like.

cody-littley (approved):

PR #12305, opened 2024-03-22
feat: task scheduler config

### LINKED ISSUE #11847: Task scheduler configuration
Create a complex configuration type for task schedulers.

cody-littley:
closes #11847


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+/**
+ * Configures a task scheduler.
+ *
+ * @param type                       the type of task scheduler, if null then {@link TaskSchedulerType#SEQUENTIAL} is
+ *                                   used
+ * @param unhandledTaskCapacity      the maximum number of unhandled tasks, or 0 if unbounded, if null then 0 is used
```
alittley, line 27: Maybe `1` be a safer default `unhandledTaskCapacity`? A queue that accidentally has a size of 1 when more is needed will simply have bad performance. On the other hand, a queue that is accidentally unbounded could behave fine in the normal case, but be dangerous in the edge case.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+            final String[] parts = string.split(" ");
+            for (final String part : parts) {
+                String strippedPart = part.strip();
+                if (strippedPart.isEmpty()) {
+                    continue;
+                }
+
+                try {
+                    // if the part can be parsed as a TaskSchedulerType, then it is the type
+                    type = TaskSchedulerType.valueOf(strippedPart);
```
alittley, line 87: consider throwing if `type` has already been defined

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+                if (strippedPart.startsWith(TaskSchedulerConfigOption.CAPACITY)) {
+                    try {
+                        // parse a string in the form "CAPACITY(1234)"
+                        final int openParenIndex = strippedPart.indexOf('(');
+                        final int closeParenIndex = strippedPart.indexOf(')');
+                        if (openParenIndex == -1 || closeParenIndex == -1) {
+                            throw new IllegalArgumentException("Invalid capacity \"" + strippedPart + "\"");
+                        }
+                        final String capacityString = strippedPart.substring(openParenIndex + 1, closeParenIndex);
+                        unhandledTaskCapacity = Long.parseLong(capacityString);
```
alittley, line 102: consider throwing if capacity is already defined

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+
+        try {
+            final String[] parts = string.split(" ");
+            for (final String part : parts) {
+                String strippedPart = part.strip();
+                if (strippedPart.isEmpty()) {
+                    continue;
+                }
+
+                try {
```
alittley, line 85: consider splitting this `parse` method into separate private methods: `parseType`, `parseCapacity`, and `parseOption`

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+                    boolean value = true;
+                    if (strippedPart.startsWith("!")) {
+                        value = false;
+                        strippedPart = strippedPart.substring(1);
+                    }
+
+                    final TaskSchedulerConfigOption option = TaskSchedulerConfigOption.valueOf(strippedPart);
+
+                    switch (option) {
+                        case UNHANDLED_TASK_METRIC -> unhandledTaskMetricEnabled = value;
```
alittley, line 120: consider throwing if any value already defined is parsed again

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
```diff
@@ -98,6 +99,8 @@ public record PlatformSchedulersConfig(
         @ConfigProperty(defaultValue = "1.0") double defaultPoolMultiplier,
         @ConfigProperty(defaultValue = "0") int defaultPoolConstant,
         @ConfigProperty(defaultValue = "500") int eventHasherUnhandledCapacity,
+        @ConfigProperty(defaultValue = "SEQUENTIAL CAPACITY(500) FLUSHABLE UNHANDLED_TASK_METRIC")
+                TaskSchedulerConfiguration internalEventValidator,
         @ConfigProperty(defaultValue = "SEQUENTIAL") TaskSchedulerType internalEventValidatorSchedulerType,
```
alittley, line 104: Is there a reason you left `internalEventValidatorSchedulerType` and `internalEventValidatorUnhandledCapacity` intact, despite newly defining this configuration in `TaskSchedulerConfiguration internalEventValidator`?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
alittley: Alternatively, we could force definition of unhandled task capacity, and not permit any scheduler to rely on a default

lpetrovic05: why not use a primitive for this?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfigOption.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package com.swirlds.common.wiring.schedulers.builders;
+
+/**
+ * Various configuration options for a task scheduler. Note that the task scheduler type uses values from
+ * {@link TaskSchedulerType}, and that the unhandled task capacity is represented as an integer value.
+ */
+public enum TaskSchedulerConfigOption {
```
lpetrovic05, line 23: Not a requirement, just a suggestion.
I might have been better if all were enums, and based on the enum itself, it might have an argument or not.

lpetrovic05: then when parsing, you check if the enum has arguments or no, and then optionally parse the arguments

lpetrovic05 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
cody-littley: I'm ok using 1 as the default for now. In the medium to long term, I expect we are going to make all of them unbounded -- relying on some sort of health monitor to detect when we need to stop gossiping. At that point in time the default will be changed back to unlimited.

cody-littley: > why not use a primitive for this?

I'm using a boxed primitive intentionally. This allows `null` to represent when a value was not specified (i.e. when configured to use the default).

cody-littley: Change executed. Looking at existing scheduler configuration, the only component utilizing the previous default was the post hash collector. I've now explicitly configured this to use an unlimited capacity.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
cody-littley: check added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
cody-littley: check added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
cody-littley: I've split a large amount of the logic into helper functions, but due to the nature of variables being local to the static method and not classmember variables, there is still a bunch of logic that has to remain in the top level method.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformSchedulersConfig.java
cody-littley: Unintentional. They are now removed.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfigOption.java
cody-littley: Are you saying you would have preferred to see syntax like `TYPE(SEQUENTIAL)`?

My personal preference would be to add features/complexity to the parser only as we need it. At the time, only the capacity has an argument, it seems like a premature feature to generalize it so that any could have an argument. Definitely something to consider if we find multiple configuration options in need of arguments.

Would it be ok if this merges as is, and we consider this type of change in follow up PRs?

lpetrovic05 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,249 @@
...
+        return null;
+    }
+
+    /**
+     * Try to parse a string as an unhandled task metric configuration.
+     *
+     * @param string the string to parse
+     * @return the parsed configuration, or null if the string is not a valid configuration
+     */
+    private static Boolean tryToParseUnhandledTaskMetric(@NonNull final String string) {
```
alittley, line 196: `tryToParseUnhandledTaskMetric`, `tryToParseBusyFractionMetric`, `tryToParseFlushing`, and `tryToParseSquelching` could all be replaced with a single method:

```
    private static Boolean tryToParseOption(
            @NonNull final String string, @NonNull final TaskSchedulerConfigOption option) {
        if (string.equals(option.toString())) {
            return true;
        } else if (string.equals("!" + option)) {
            return false;
        }
        return null;
    }
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,249 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+/**
+ * Configures a task scheduler.
+ *
+ * @param type                       the type of task scheduler, if null then {@link TaskSchedulerType#SEQUENTIAL} is
+ *                                   used
+ * @param unhandledTaskCapacity      the maximum number of unhandled tasks, or 0 if unbounded, if null then 0 is used
```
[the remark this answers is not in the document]
cody-littley: 0 == unbounded

> or 0 if unbounded, if null then 0 is used

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,249 @@
...
+                continue;
+            }
+
+            final Boolean parsedSquelching = tryToParseSquelching(strippedPart);
+            if (parsedSquelching != null) {
+                if (squelchingEnabled != null) {
+                    throw new IllegalStateException("Multiple squelching configurations specified: " + string);
+                }
+                squelchingEnabled = parsedSquelching;
+            }
```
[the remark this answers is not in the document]
cody-littley: added

alittley (dismissed):

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+            final TaskSchedulerType parsedType = tryToParseTaskSchedulerType(strippedPart);
+            if (parsedType != null) {
+                if (type != null) {
+                    throw new IllegalArgumentException("Multiple task scheduler types specified: " + string);
+                }
+                type = parsedType;
+                continue;
+            }
+
+            final Long parsedCapacity = tryToParseCapacity(strippedPart);
```
anthony-swirldslabs, line 98: Minor: perhaps using a regular expression to parse something like `"(\!)?(\w+)(?:\((\d+)\))?"` and implementing a `static TaskSchedulerConfigOption.valueOf(String)` to parse the special CAPACITY and the rest of the enum constants (using the Enum.valueOf), and then running a simple `switch (valueOf(matcher.group(2)))` here could result in a more robust parser by avoiding low-level strings parsing in the try*() methods.

platform-sdk/swirlds-config-extensions/src/testFixtures/java/com/swirlds/config/extensions/test/fixtures/ConfigUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    public static Set<Class<? extends Record>> loadAllConfigDataRecords(@NonNull final Set<String> packagePrefixes) {
+        Objects.requireNonNull(packagePrefixes, "packagePrefix must not be null");
+
+        final ClassGraph classGraph = new ClassGraph().enableAnnotationInfo();
+
+        if (!packagePrefixes.isEmpty()) {
+            classGraph.whitelistPackages(packagePrefixes.toArray(new String[0]));
+        }
+
+        try (final ScanResult result = classGraph.scan()) {
```
anthony-swirldslabs, line 69: Minor: this may be a heavy operation, especially if certain use cases (e.g. some tests, which is especially important in the context of our goal to speed them up) may not require all the classes to be loaded. Although one could consider it a good "warm up" exercise for the JVM. Also, the classgraph seems to have known issues when running with JDK 16+.

I agree that this scan seems convenient. But perhaps a more optimal solution would be to manually register all the supported/required records somewhere near the application's `main()` method or the tests's `@BeforeAll` method.

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-config-extensions/src/testFixtures/java/com/swirlds/config/extensions/test/fixtures/ConfigUtils.java
cody-littley: The only use of this method is to verify that we properly load all platform config data types, which we verify in a single test in `PlatformConfigurationExtensionTests`. This method is not called in production environments or as a regular operation within other unit tests.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerConfiguration.java
cody-littley: My personal preference for this parsing is to keep the initial implementation super simple, and to consider more complex/complete parsing solutions if we find this to be an area where we have growing pain. IMO, even if regex results in more robust code, I think manual string parsing is a lot simpler simpler given the size of the grammar we need to parse.

anthony-swirldslabs: Sounds fair.

anthony-swirldslabs (dismissed):

alittley (approved):

anthony-swirldslabs (dismissed):

alittley (dismissed):

alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/TaskSchedulerBuilder.java
```diff
@@ -56,8 +59,7 @@ public class TaskSchedulerBuilder<O> {
 
     private TaskSchedulerType type = TaskSchedulerType.SEQUENTIAL;
     private final String name;
-    private TaskSchedulerMetricsBuilder metricsBuilder;
-    private long unhandledTaskCapacity = UNLIMITED_CAPACITY;
+    private long unhandledTaskCapacity = 1;
```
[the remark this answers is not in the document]
cody-littley: This is a behavior change. Requested by @alittley. The reasoning is that a capacity of 1 may result in bad performance if you accidentally use it, but an unlimited capacity may result in an OOM if you use it accidentally.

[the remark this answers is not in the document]
cody-littley: Semi-related -- in general, I think we should always specify the capacity for each scheduler. Reliance on the default capacity risks unintentional consequences.

anthony-swirldslabs (approved):

alittley (dismissed):

alittley (approved):

alittley (approved):

timo0 (approved):

PR #12314, opened 2024-03-22
chore: unify logic for reclaiming frontend throttle capacity

tinker-michaelj:
**Description**:
 - In both mono- and mod-service, reclaim capacity used to throttle implicit creations when they fail at consensus.
     * **mono**: Move throttle reclamation into `ServicesTxnManager`.
         * Also, ensure the `TxnAccessor` for an `EthereumTransaction` always "counts" an implied lazy creation by calling `countImplicitCreations()` when "rationalizing" an `EthereumTransaction` span.
     * **mod**: Move throttle reclamation into `HandleWorkflow`.
 -  _Note:_ only the node that submitted the failed auto-creation reclaims frontend capacity, of course.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleServiceManager.java
```diff
@@ -136,6 +139,11 @@ public ResponseCodeEnum recreateThrottles(@NonNull final Bytes encoded) {
         return validatedThrottles.successStatus();
     }
 
+    public int numImplicitCreations(
+            @NonNull final TransactionBody body, @NonNull final ReadableAccountStore accountStore) {
+        return backendThrottle.getImplicitCreationsCount(body, accountStore);
```
Neeharika-Sompalli, line 144: We could probably store this value in `ThrottleAccumulator` and rest it in `resetUsage`. Since it only happens for a failed transfer, may be it is okay to re-calculate. Not a blocker for this PR. 👍

Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleServiceManager.java
tinker-michaelj: True, should get this in today. 💯

PR #12315, opened 2024-03-22
feat(ReconnectBench): emulate slow I/O

### LINKED ISSUE #11809: Emulate latencies in ReconnectBench
1. Replicate `Lagging*` classes in `swirlds-benchmarks`
2. Add a parameter to specify the latency with the default value of zero
3. Use the replicated classes and the new parameter in `MerkleBenchmarkUtils`

anthony-swirldslabs:
**Description**:
Introducing "slow" implementations for the Teaching- and Learning-Synchronizers that sleep before sending every single message during a reconnect, and enabling the ReconnectBench to enable these slow implementation by specifying a non-zero `delayMilliseconds` JMH parameter. Note that this implementation is different from the already existing Lagging- versions of these classes in that the latter only introduces a lag (i.e. a shift in time) for all the data being sent over the stream, while the BenchmarkSlow- implementations in this fix actually delay each and every message independently, thus simulating the slow behavior of real I/O operations happening for every message compared to the normal in-memory behavior when the benchmark runs very fast w/o any simulated delays.

As a minor change, I'm updating JVM parameters for the jmhReconnect target that allow it to build larger states. In particular, I was able to successfully run the benchmark on my Mac with a state of 100M. These parameters would also allow it to build even larger states, e.g. I tried building 500M and it could do that. However, the reconnect process itself after the state has been built still required a larger direct memory limit. Fixing this is outside of scope of this PR.

Finally, I'm updating the parameters of the `jmhReconnect` Gradle target to use a state of size `10,000` with `0.01` probabilities of adding/removing/modifying nodes, and running with a delay of `1` ms.

This may seem like too small a state, but it isn't. Thanks to the delay which applies to each and every message exchanged over the I/O streams, this enables us to immediately spot a performance difference between various reconnect implementations that is attributed directly to the number of nodes exchanged over the wire and/or read/written from/to disk. It is assumed that this must be the most important optimization factor for the reconnect process because anything else in the Reconnect implementation seems to be optimal as it is today already.

If desired, it's still physically possible to test with larger states and the delay. However, the running of such a test would take an unnecessarily long time, while the delay already simulates the necessary performance testing parameters even with much smaller states.

**Related issue(s)**:

Fixes #11809

**Notes for reviewer**:
Updated on 3/25 with the latest commit using 100μs storage and 50μs network delays:

Using the Level By Level Traversal fix as an example and running

`gr jmhReconnect && (grep "Finished sync" output/swirlds.log  | tail -5)`

I'm observing a 35% time improvement as can be seen:

Without the fix:
```
Benchmark                 (delayNetworkMicroseconds)  (delayStorageMicroseconds)  (keySize)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (teacherAddProbability)  (teacherModifyProbability)  (teacherRemoveProbability)  Mode  Cnt  Score   Error  Units
ReconnectBench.reconnect                          50                         100          8   1000000         100          1000            32    9823452658           128                     0.01                        0.01                        0.01  avgt    7  4.856 ± 0.429   s/op

BUILD SUCCESSFUL in 2m 13s
80 actionable tasks: 80 executed
2024-03-25 14:47:21.205 743      INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":4.912,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":23677,"leafNodes":4083,"redundantLeafNodes":1988,"internalNodes":19594,"redundantInternalNodes":7755} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:47:27.270 804      INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":4.897,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":23677,"leafNodes":4083,"redundantLeafNodes":1988,"internalNodes":19594,"redundantInternalNodes":7755} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:47:32.406 865      INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.9490000000000003,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":19520,"leafNodes":3908,"redundantLeafNodes":1813,"internalNodes":15612,"redundantInternalNodes":3773} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:47:38.423 926      INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":4.914,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":23677,"leafNodes":4083,"redundantLeafNodes":1988,"internalNodes":19594,"redundantInternalNodes":7755} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:47:44.521 987      INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":4.928,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":23677,"leafNodes":4083,"redundantLeafNodes":1988,"internalNodes":19594,"redundantInternalNodes":7755} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
```

With the Level By Level Traversal fix:
```
Benchmark                 (delayNetworkMicroseconds)  (delayStorageMicroseconds)  (keySize)  (maxKey)  (numFiles)  (numRecords)  (numThreads)  (randomSeed)  (recordSize)  (teacherAddProbability)  (teacherModifyProbability)  (teacherRemoveProbability)  Mode  Cnt  Score   Error  Units
ReconnectBench.reconnect                          50                         100          8   1000000         100          1000            32    9823452658           128                     0.01                        0.01                        0.01  avgt    7  3.175 ± 0.201   s/op

BUILD SUCCESSFUL in 2m 22s
80 actionable tasks: 80 executed
2024-03-25 14:44:18.335 1231     INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.12,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":13934,"leafNodes":2095,"redundantLeafNodes":0,"internalNodes":11839,"redundantInternalNodes":0} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:44:22.781 1292     INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.325,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":14669,"leafNodes":2095,"redundantLeafNodes":0,"internalNodes":12574,"redundantInternalNodes":735} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:44:27.127 1353     INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.228,"hashTimeInSeconds":0.001,"initializationTimeInSeconds":0.0,"totalNodes":14408,"leafNodes":2095,"redundantLeafNodes":0,"internalNodes":12313,"redundantInternalNodes":474} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:44:31.443 1414     INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.166,"hashTimeInSeconds":0.0,"initializationTimeInSeconds":0.0,"totalNodes":14602,"leafNodes":2624,"redundantLeafNodes":529,"internalNodes":11978,"redundantInternalNodes":139} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
2024-03-25 14:44:35.679 1475     INFO  RECONNECT        <<work group synchronization-test: learning-synchronizer-main #1>> LearningSynchronizer: Finished synchronization {"timeInSeconds":3.133,"hashTimeInSeconds":0.001,"initializationTimeInSeconds":0.0,"totalNodes":14342,"leafNodes":2478,"redundantLeafNodes":383,"internalNodes":11864,"redundantInternalNodes":25} [com.swirlds.logging.legacy.payload.SynchronizationCompletePayload]
```

**Checklist**


platform-sdk/swirlds-benchmarks/build.gradle.kts
```diff
@@ -50,10 +50,22 @@ fun listProperty(value: String) = objects.listProperty<String>().value(listOf(va
...
+        )
+    )
 
     resultsFile.convention(layout.buildDirectory.file("results/jmh/results-reconnect.txt"))
 
-    benchmarkParameters.put("numRecords", listProperty("100000"))
-    benchmarkParameters.put("numFiles", listProperty("500"))
+    benchmarkParameters.put("numRecords", listProperty("1000"))
+    benchmarkParameters.put("numFiles", listProperty("10"))
+    benchmarkParameters.put("delayMilliseconds", listProperty("1"))
```
jsync-swirlds, line 67: Would it be appropriate to fuzz this value?
Real networks have variable latency (and typically range from 2ms to 500ms or more, depending on route).  Perhaps config for minDelayMilliseconds and maxDelayMilliseconds then the slow reconnect implementations insert a value that's random between min and max?
Just a thought that the variability may help us find more potential issues.

jsync-swirlds (commented):
Looks good; one possible improvement.

platform-sdk/swirlds-benchmarks/build.gradle.kts
anthony-swirldslabs: I was thinking about that myself. Two reasons I decided against this at this time:
1. I already have to use a 1 ms delay. Anything larger would make the benchmark run a lot longer because the delay applies to every message exchanged over the wire. While one could use nanoTime() and a busy sleep loop for delays less than 1 ms, I don't find this elegant. Especially when it comes to performance testing.
2. It's nice when benchmark results are as reproducible as possible. Introducing randomness into the delays, especially when benchmarking the reconnect where we have multiple threads working concurrently, may not be too very desired. It would also come with an extra complexity of having to create separate Random instances with parametrized seeds for each of the threads. While certainly possible, I don't feel that this extra complexity is needed at this time.

So we can certainly do this in the future if we feel this would add a new dimension to the performance testing, but this doesn't seem necessary or justified currently.

jsync-swirlds: Entirely reasonable.  Definitely something to revisit in the future, as I have found many sources of performance degradation do not show up in steady-state or smooth situations, but become significant in the presence of sufficient variability.

Perhaps in the future we can create a _separate_ performance benchmark that runs over a much longer timeframe (so that delay can be close to real-world network delay, rather than extremely short) and tracks total "sleep" time as well as total "run" time so we can assess effective performance without noise from the simulated network delay.

jsync-swirlds (approved):

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/lag/BenchmarkSlowAsyncOutputStream.java
```diff
@@ -0,0 +1,68 @@
...
+        getOutgoingMessages().put(message);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void serializeMessage(final T message) throws IOException {
+        try {
+            Thread.sleep(delayMilliseconds);
```
artemananiev, line 62: This would be an overkill. Even 1ms delay here would result in catastrophic drop in reconnect performance (my observations: about 100x drop). Instead, an approach from `LaggingAsyncOutputStream` may be used

platform-sdk/swirlds-benchmarks/build.gradle.kts
anthony-swirldslabs: I filed https://github.com/hashgraph/hedera-services/issues/12343 to implement the delay fuzzing.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/lag/BenchmarkSlowAsyncOutputStream.java
anthony-swirldslabs: Also, I've updated the description with the results of the latest revision. Level By Level traversal wins by 32% over the current baseline when using the 100μs/50μs delays.

Please do read the description for all the supporting comments and the updated test results.

platform-sdk/swirlds-benchmarks/build.gradle.kts
anthony-swirldslabs: FYI, the microsecond-level sleep is now implemented in the latest commit here. Also, there's now separate storage and network delays that can be tweaked independently.

platform-sdk/swirlds-benchmarks/src/jmh/java/com/swirlds/benchmark/reconnect/lag/BenchmarkSlowAsyncOutputStream.java
```diff
@@ -58,11 +62,17 @@ public void sendAsync(final T message) throws InterruptedException {
...
         message.serialize(getOutputStream());
     }
+
+    /**
+     * Busy sleep for a given number of microseconds.
+     * @param micros time to sleep, in microseconds
+     */
+    private static void sleepMicros(final long micros) {
+        final long sleepUntil = System.nanoTime() + micros * 1000L;
+        while (System.nanoTime() < sleepUntil)
```
jsync-swirlds, line 75: There's no need for a spinwait here.
Thread has a static [sleep](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Thread.html#sleep(java.time.Duration)) method with sufficient precision.
```java
Thread.sleep(Duration.ofNanos(micros * 1000));
```

anthony-swirldslabs: Ah, it's available since Java 19. Good to learn about it. Let me update the fix.

anthony-swirldslabs: Note that it's not as precise as the busy wait. And it in fact slows down the benchmark by some 18%. But I guess it's still better than a busy loop. So the latest commit implements this.

artemananiev (approved):

JeffreyDallas (approved):

PR #12317, opened 2024-03-23
fix: deterministic order of effective payer numbers

### LINKED ISSUE #12300: (0.48) `TransferWithCustomFractionalFees` ISS
### Description

We have seen an ISS isolated to the records FCQ that appears to be correlated to the `TransferWithCustomFractionalFees` suite ([here](https://github.com/hashgraph/hedera-services/actions/runs/8370825854/job/22918990506?pr=12168)).

### Steps to reproduce

Rerun CI.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12300 
 - Replaces `Map.copyOf()` with `LinkedHashMap::new`, as the `ImmutableCollections.MapN` returned by the former has an [intentionally random iterator](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ImmutableCollections.java#L1266). 🤯 
     * This made the order of effective payer numbers for a fractional fee with multiple receivers non-deterministic.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AssessmentResult.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AssessmentResult.java
@@ -54,7 +54,11 @@ public class AssessmentResult {
             final List<TokenTransferList> inputTokenTransfers, final List<AccountAmount> inputHbarTransfers) {
         mutableInputBalanceAdjustments = buildFungibleTokenTransferMap(inputTokenTransfers);
         immutableInputTokenAdjustments = Collections.unmodifiableMap(mutableInputBalanceAdjustments.entrySet().stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, entry -> Map.copyOf(entry.getValue()))));
+                .collect(Collectors.toMap(
+                        Map.Entry::getKey,
+                        entry -> new LinkedHashMap<>(entry.getValue()),
+                        (a, b) -> a,
+                        LinkedHashMap::new)));
 
         immutableInputHbarAdjustments = buildHbarTransferMap(inputHbarTransfers);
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AssessmentResult.java
```diff
@@ -54,7 +54,11 @@ public AssessmentResult(
             final List<TokenTransferList> inputTokenTransfers, final List<AccountAmount> inputHbarTransfers) {
         mutableInputBalanceAdjustments = buildFungibleTokenTransferMap(inputTokenTransfers);
         immutableInputTokenAdjustments = Collections.unmodifiableMap(mutableInputBalanceAdjustments.entrySet().stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, entry -> Map.copyOf(entry.getValue()))));
+                .collect(Collectors.toMap(
+                        Map.Entry::getKey,
+                        entry -> new LinkedHashMap<>(entry.getValue()),
```
netopyr, line 59: This will create a modifiable `Map`. Is this okay?

netopyr (approved):

Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AssessmentResult.java
tinker-michaelj: Note this is wrapped in in `Collections.unmodifiableMap()`.

netopyr: Only the "outer" `Map` is wrapped. The "inner" `Map`s, created by `entry -> new LinkedHashMap<>(entry.getValue())`, are still modifiable.

PR #12271, opened 2024-03-21
feat: enable admin key to be able to remove other keys

fchitakova:
**Description**:

The purpose of these changes is to enable the admin key to be able remove itself from a token, as well as the other low privileged keys as described in [HIP-540](https://hips.hedera.com/hip/hip-540  ). 

To achieve this, the change checks:

1. If the transaction body contains IMMUTABILITY_SENTINEL_KEY value for the any key, since this is how we expect the keys to be removed. 
2. If the admin keys is present and used for signing the transaction in addition to the required payer signature

**Related issue(s)**: https://github.com/hashgraph/hedera-services/issues/12158

Fixes #

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #12327, opened 2024-03-25
feat: low priority keys can change themselves to another valid and an invalid key

### LINKED ISSUE #12160: HIP-540: Implement and test the logic that all keys can change themselves to an invalid key (such as all-zeros)
Such updates will be enabled by an explicit flag in the transaction body that will indicate if we should check if the new key is valid (by default we should be checking the validity to be backwards compatible)

petreze:
**Description**:
Currently admin key is able to change itself and also other low privilege keys.
Additionally we want to implement the ability for low privilege keys to be able to update themselves to new valid keys (i.e. without requiring admin key signature but instead only the signature of the corresponding low privilege key being updated).
This new logic also applies to "invalid" keys such as `0x0000000000000000000000000000000000000000` but not limited to

Several X tests are added in a new class `TokenUpdateReplaceKeysXTest` and a new `Hip540ChangeOrRemoveKeysSuite` is added to test cases where we need to verify signature requirements related to the HIP

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
This PR combines the logic for both issues as they share similar logic

- A new `addRequiredSigners()` method was introduced in `TokenUpdateHandler` to cover all the cases of signature requirements. The tricky part is when we have a low priority key update. There we want to either require the admin key signature or the signatures of all of the respective low priority keys, which are going to be updated, that were previously set to the token. To handle this case, a complex `Threshold` key 1/2 was created where one of the keys is the admin key and the other one is a `KeyList` (N/N must sign) of all the current respective low priority keys which are going to be updated
- Some changes were made in `HapiTokenUpdate` and `HapiGetTokenInfo` in order to test the update a key to an invalid one
- Several negative suite tests were added in `TokenUpdateSpecs` to cover some cases and a brand new `Hip540ChangeOrRemoveKeysSuite` was created to test all the new logic related to this HIP

cc: @stefan-stefanooov 

**Checklist**


stefan-stefanooov (approved):

stefan-stefanooov (dismissed):

stefan-stefanooov (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -434,6 +401,107 @@ private void updateKeys(
...
+                        op.hasFeeScheduleKey());
+                addRequiredLowPrioritySigner(
+                        lowPriorityKeys, originalToken.hasPauseKey(), originalToken.pauseKey(), op.hasPauseKey());
+                addRequiredLowPrioritySigner(
+                        lowPriorityKeys,
+                        originalToken.hasMetadataKey(),
+                        originalToken.metadataKey(),
+                        op.hasMetadataKey());
+                addRequiredLowPrioritySigner(
+                        lowPriorityKeys, originalToken.hasMetadataKey(), originalToken.metadataKey(), op.hasMetadata());
```
Neeharika-Sompalli, line 460: This update to metadata can have signature from admin or metadataKey. We need to use a threshold key in this case, like we did before.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -434,6 +401,107 @@ private void updateKeys(
...
+                    // to sign)
+                    requiredKey = Key.newBuilder()
+                            .thresholdKey(ThresholdKey.newBuilder()
+                                    .keys(KeyList.newBuilder()
+                                            .keys(keysRequired)
+                                            .build())
+                                    .threshold(1)
+                                    .build())
+                            .build();
+                }
```
Neeharika-Sompalli, line 482: If I read this correctly, one of all the keys should sign. What happens if we are updating several keys in same update transaction ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -427,6 +427,36 @@ public static boolean isMetadataOnlyUpdateOp(@NonNull final TokenUpdateTransacti
...
+                || op.hasAutoRenewPeriod()
+                || op.hasMemo()
+                || op.hasExpiry());
+    }
+
+    /**
+     * Check if a given token already has some of the low priority keys
+     * low priority keys are -> wipeKey, kycKey, supplyKey, freezeKey, feeScheduleKey, pauseKey or metadataKey
+     */
+    public static boolean hasAlreadySomeLowPriorityKeys(@NonNull final Token token) {
```
thomas-swirlds-labs, line 450: We should consider renaming LowPriorityKeys to NonAdminKeys - as it will be clearer and won't require us to call out explicitly what the other 7 keys are each time.

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerTest.java
```diff
@@ -267,6 +267,13 @@ void failsIfTokenImmutable() {
                 .get(fungibleTokenId)
                 .copyBuilder()
                 .adminKey((Key) null)
+                .wipeKey((Key) null)
```
thomas-swirlds-labs, line 270: Should add unit tests to cover scenarios other than `failsIfTokenImmutable`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -427,6 +427,36 @@ public static boolean isMetadataOnlyUpdateOp(@NonNull final TokenUpdateTransacti
         return op.equals(defaultWithMetadata);
     }
 
+    /**
+     * Check if TokenUpdateOp wants to update only some of the low priority keys or the metadata field
```
thomas-swirlds-labs, line 431: Should metadata update be part of this?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/Hip540ChangeOrRemoveKeysSuite.java
```diff
@@ -0,0 +1,615 @@
...
+                        .payingWith(supply))
+                .then(getTokenInfo(tokenName).searchKeysGlobally().hasSupplyKey(supplyKey2));
+    }
+
+    @HapiTest
+    final HapiSpec tokenUpdateFeeScheduleKey() {
+        final var admin = "admin";
+        final var feeSchedule = "feeSchedule";
+        final var feeScheduleKey2 = "feeScheduleKey2";
+        final var newFeeScheduleKey = "feeScheduleRedux";
```
thomas-swirlds-labs, line 186: nit: newFeeScheduleKey is not required and can be removed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenUpdate.java
```diff
@@ -219,13 +226,33 @@ public HapiTokenUpdate expiry(long at) {
...
-    public HapiTokenUpdate improperlyEmptyingAdminKey() {
-        useImproperEmptyKey = true;
+    public HapiTokenUpdate properlyEmptyingAdminKey() {
+        useEmptyAdminKeyList = true;
         return this;
     }
 
-    public HapiTokenUpdate properlyEmptyingAdminKey() {
-        useEmptyAdminKeyList = true;
+    public HapiTokenUpdate usingInvalidAdminKey() {
```
thomas-swirlds-labs, line 234: Handles invalid keys but what about "empty" keys?

hedera-node/hedera-app/src/xtest/java/token/update/AbstractTokenUpdateXTest.java
```diff
@@ -0,0 +1,58 @@
+/*
```
thomas-swirlds-labs, line 1: This class is never used and could be deleted

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/queries/token/HapiGetTokenInfo.java
```diff
@@ -299,6 +309,46 @@ public HapiGetTokenInfo searchKeysGlobally() {
         return this;
     }
 
+    public HapiGetTokenInfo hasInvalidAdminKey() {
```
thomas-swirlds-labs, line 312: Handles invalid keys but what about "empty" ?

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerParityTest.java
```diff
@@ -145,12 +152,23 @@ void tokenUpdateReplacingAdminKey() throws PreCheckException {
     void tokenUpdateWithSupplyKeyedToken() throws PreCheckException {
         final var txn = txnFrom(UPDATE_WITH_SUPPLY_KEYED_TOKEN);
         final var context = new FakePreHandleContext(readableAccountStore, txn);
+        final var lowPriorityKeyList = Key.newBuilder()
```
thomas-swirlds-labs, line 155: NIT: could refactor to declare lowPriorityKeyList and thresholdKey once and avoid duplication

thomas-swirlds-labs (changes requested):
Added some comments inline.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
petreze: @Neeharika-Sompalli 
Hey, not exactly. Here we have `lowPriorityKeys` list of low priority keys to be updated. After that if the list is not empty, we create `lowPriorityKeyList` KeyList (N/N signers) and only then we create the threshold key (1/2 signers) which requires either all keys from `lowPriorityKeyList` to sign or only the admin key to sign.

In that way, if we have 2 low prio keys to be updated we either require those 2 signatures or only the admin key signature

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
petreze: @thomas-swirlds-labs 
As I understood correctly the metadata HIP, we should allow the case where the `metadataKey` can update the `metadata` field by itself without requiring the admin key signature..? [Here](https://github.com/hashgraph/hedera-services/blob/12160-low-priority-to-an-invalid-key/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java#L459-L460) we allow for this when we enter the in this case `noOtherFieldThanLowPriorityKeyOrMetadataWillBeUpdated(op)`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenUpdate.java
petreze: @thomas-swirlds-labs 
We separated the logic of the HIP into 2 PRs. This one contains the implementation of low prio keys to change themselves to another valid or an invalid key. The other functionality will be [here](https://github.com/hashgraph/hedera-services/pull/12271) where the "empty" keys are considered removed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/queries/token/HapiGetTokenInfo.java
petreze: Same as [here](https://github.com/hashgraph/hedera-services/pull/12327#discussion_r1552224086)

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerParityTest.java
petreze: @thomas-swirlds-labs 
The `lowPriorityKeyList` contains different key for each keys and there are only 4 occurrences of these types of tests... do you mean to declare them inside `setUp()` which will be executed before each of the test cases?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -434,6 +401,107 @@ private void updateKeys(
...
+                } else {
+                    final Key lowPriorityKeyList = Key.newBuilder()
+                            .keyList(KeyList.newBuilder().keys(lowPriorityKeys).build())
+                            .build();
+                    final List<Key> keysRequired = List.of(lowPriorityKeyList, originalToken.adminKey());
+                    // with this we will require either admin key signature or all low priority keys signatures
+                    // (we can update several low priority keys with one TokenUpdate, so we will require all of these
+                    // keys
+                    // to sign)
+                    requiredKey = Key.newBuilder()
```
Neeharika-Sompalli, line 474: If I read this correctly, one of all the keys should sign. What happens if we are updating several keys in same update transaction ? This should just take adminKey or all the keys updated to sign the transaction

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -143,25 +143,25 @@ public void validateTokenKeys(
         if (hasAdminKey && !isKeyRemoval(adminKey)) {
             validateTrue(isValid(adminKey), INVALID_ADMIN_KEY);
         }
-        if (hasKycKey) {
+        if (hasKycKey && !isKeyRemoval(kycKey)) {
```
Neeharika-Sompalli, line 146: Why are we not checking the `NO_VALIDATION` condition here ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenUpdateSpecs.java
```diff
@@ -137,8 +139,26 @@ public List<HapiSpec> getSpecsInSuite() {
                 canUpdateExpiryOnlyOpWithoutAdminKey());
     }
 
+    private List<HapiSpec> negativeTests() {
+        return List.of(
+                updateTokenTreasuryRequiresZeroTokenBalance(),
```
Neeharika-Sompalli, line 144: Need more tests covering updating multiple keys in same transaction

Neeharika-Sompalli (changes requested):
The way we are adding required keys for `TokenUpdate` seems like it wont work. Need more tests validating all the cases.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -434,6 +401,107 @@ private void updateKeys(
...
+     * @param op token update transaction body
+     * @param originalToken original token
+     */
+    private void addRequiredSigners(
+            @NonNull PreHandleContext context,
+            @NonNull final TokenUpdateTransactionBody op,
+            @NonNull final Token originalToken)
+            throws PreCheckException {
+        if (op.hasAdminKey()) {
+            if (isValid(op.adminKey())) {
```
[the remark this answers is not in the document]
petreze: I agree but here the concern is other and the comments explain that. Basically `op.adminKey()` is valid, we require it to sign the update, and if its not, we consider it to be invalid, hence the previous admin key wants to make itself to an invalid key (i.e: long zero address) and in that case we require the already existing admin key to sign

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -427,6 +427,36 @@ public static boolean isMetadataOnlyUpdateOp(@NonNull final TokenUpdateTransacti
...
+     * low priority keys are -> wipeKey, kycKey, supplyKey, freezeKey, feeScheduleKey, pauseKey or metadataKey
+     */
+    public static boolean noOtherFieldThanLowPriorityKeyOrMetadataWillBeUpdated(
```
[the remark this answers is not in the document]
petreze: @Neeharika-Sompalli 
I was thinking about the same but could not figure it out other way...

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
petreze: Here we have `lowPriorityKeys` list of low priority keys to be updated. After that if the list is not empty, we create `lowPriorityKeyList` KeyList (N/N signers) and only then we create the `Threshold` key (1/2 signers) which requires either all keys from lowPriorityKeyList to sign or only the admin key to sign.

In that way, if we have 2 or more low prio keys to be updated we either require those low prio keys to sign or only the admin key signature

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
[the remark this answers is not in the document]
petreze: @Neeharika-Sompalli 
We handle this case in `getAllRequiredLowPrioritySigners()`. This method adds every low priority key in a `KeyList` which will be required to sign the TokenUpdateTx

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
petreze: Because method `validateTokenKeys()` is also used in `TokenCreateValidator` as well. Instead, in `TokenUpdateValidator` we are checking for other than `NO_VALIDATION` enum value and then perform key validations

hedera-node/hedera-app/src/xtest/java/token/update/AbstractTokenUpdateXTest.java
petreze: Removed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenUpdateSpecs.java
petreze: @Neeharika-Sompalli 
Those types of tests are in `Hip540ChangeOrRemoveKeysSuite` file. We split the specific tests about the HIP there for better visibility

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerTest.java
petreze: Added

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenUpdate.java
thomas-swirlds-labs: Ok. Makes sense. In that case it could also be useful to have a long running HIP-540 feature branch to merge the multiple PRs into before develop ?

petreze:
Closing this one as we created a [PR](https://github.com/hashgraph/hedera-services/pull/12736) for the entire implementation. This was made because there are quite a lot corner cases which test the entire logic of the HIP

Issue #11139, opened 2024-01-24
System accounts disambiguation: Create .md document showing differences for operations on system accounts


kselveliev:
### Problem

Calling operations on system accounts in hedera may differ from ethereum.
We need to add a new .md file that document and shows what is the expected behaviour when calling operations
on system account and how it differs from ethereum.

### Solution

Based on the results from testing the following operations on all groups of system accounts between 0.0.0 -> 0.0.1000
#11134 #11135 #11136 #11137 #11138
Create .md doc describing the expected behaviour and how it differs from ethereum.

PR #12328, opened 2024-03-25
docs: Add documentation for the differences in the operations with system accounts

bilyana-gospodinova:
**Description**:
In this PR we add documentation for the expected results from executing the different types of operations with system accounts. The differences between Ethereum and Hedera are compared and grouped.

Fixes https://github.com/hashgraph/hedera-services/issues/11139


hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,98 @@
...
+
+### SelfDestruct operation
+
+**Ethereum:** success for any address. If the beneficiary is the same as contract address, it burns the eth and destructs the contract.
+
+**Hedera:**
+1. For addresses is range 0x0 - 0x2EE (0.0.0 - 0.0.750): fail with status `INVALID_SOLIDITY_ADDRESS`.
+2. For addresses is range 0x2EF - 0x3E8 (0.0.751 - 0.0.1000):
+   - success, if account exists and has `receiverSigRequired` == false;
+   - success, if account exists and has `receiverSigRequired` == false and account is `sender`;
```
stoqnkpL, line 42: should this be `receiverSigRequired` == true?
also, do we have a test confirming this scenario works as expected (if we don't we should plan to add test coverage)?

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,98 @@
...
+   - fail in some failing scenarios such as: beneficiary same as account-to-be-deleted, `receiverSigRequired` == true and signature provided, etc.
+3. For addresses above 0.0.1000: the same as 2. but a hollow account would be created if the beneficiary doesn't exist.
+
+### Call operations (CallOp, DelegateCallOp, CallCodeOp, StaticCallOp)
```
stoqnkpL, line 46: we should specify that no value is being transferred as part of the call, or we end up in a different flow similar to the `transfer and send` section below

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,98 @@
...
+1. For address 0x0:
+   - **Ethereum:** fail, there is no contract on this address.
```
stoqnkpL, line 48: What is meant by `fail` here and how is the behavior different from the Hedera `success with no op` behavior? 
Hedera behavior should be aligned with Ethereum at this point and there should be no difference. The way it's specified in this design doc makes it look like Hedera behaves differently.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,98 @@
...
+   - **Hedera:** success with no op.
+2. For address range 0x1 - 0x9:
+   - **Ethereum:** success, the Ethereum precompiles exist on these addresses.
+   - **Hedera:** success, the Ethereum precompiles exist on these addresses.
+3. For address range 0xA - 0x166 (0.0.10 - 0.0.358):
+   - **Ethereum:**
+     - success, if the address is a contract and we are using the correct ABI;
+     - fail, if the address is a contract and we are **not** using the correct ABI;
+     - fail, if there is no contract.
+   - **Hedera:** success with no op.
```
stoqnkpL, line 58: Similar to the comment above, we should be acting in the same way as Ethereum, as if an address on Ethereum is called on which there is no contract or the wrong abi is used.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Correct, updated as per the comment suggests.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Updated as discussed internally. Ethereum and Hedera behave the same for call operations.

stoqnkpL (approved):
LGTM

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,101 @@
...
+**Ethereum:**
+1. For addresses in range 0x0 - 0x9: success, value is locked.
+2. For addresses 0xA and above: 
+    - success, if address is a payable contract;
+    - success, if address is an account;
+    - success, if address is empty + creates the account;
+    - fail, if address is a non-payable contract.
+
+**Hedera:**
+1. For addresses is range 0x0 - 0x2EE (0.0.0 - 0.0.750):
```
MrValioBg, line 86:
```suggestion
1. For addresses in range 0x0 - 0x2EE (0.0.0 - 0.0.750):
```

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,101 @@
...
+**Ethereum:** Hash of the chosen account's code, the empty hash (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) if the account has no code, or 0 if the account does not exist or has been destroyed.
+
+**Hedera:** For addresses in range 0x0 - 0x3E8 (0.0.0 - 0.0.1000) it returns the empty hash.
+
+### Balance operation
+
+**Ethereum:** Returns balance or 0 if an account or contract doesn’t exist on the address. This is valid for any address.
+
+**Hedera:**
+1. For addresses is range 0x0 - 0x2EE (0.0.0 - 0.0.750): returns 0 as if account doesn’t exist.
```
MrValioBg, line 31: Here "is" should be "in".

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,101 @@
...
+2. For addresses is range 0x2EF - 0x3E8 (0.0.751 - 0.0.1000): returns balance or 0 if an account or contract doesn’t exist on the address. This behavior is the same for addresses above 0.0.1000.
+
+### SelfDestruct operation
+
+**Ethereum:** success for any address. If the beneficiary is the same as contract address, it burns the eth and destructs the contract.
+
+**Hedera:**
+1. For addresses is range 0x0 - 0x2EE (0.0.0 - 0.0.750): fail with status `INVALID_SOLIDITY_ADDRESS`.
+2. For addresses is range 0x2EF - 0x3E8 (0.0.751 - 0.0.1000):
```
MrValioBg, line 40: Here "is" should be "in" also.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,101 @@
+# System accounts
+
+There are some differences when calling operations on system accounts in Hedera and in Ethereum.
+This doc covers these differences and describes the expected behavior.
+
+### ExtCode operations
+
+`ExtCodeSize`
+
+**Ethereum:** Get size of an account’s code. If there’s no code returns 0.
```
MrValioBg, line 10:
```suggestion
**Ethereum:** Get size of an account’s code. If there’s no code it returns 0.
```

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Fixed

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Fixed

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Fixed

david-bakin-sl:
This isn't a review because I haven't gone through line-by-line to see the table is correct.  So it's comments/suggestions.

* _This document is well needed_ and long overdue! **Thank you!**
* Could use a little section defining, to remind the reader, of the addresses of the "system accounts" - specifically that there are two regions 0..750 and 750..1000 and how they're different
  * And also in that same section there could be a mention of where our (current) system contracts are in that range
  * This would help people understand in the tables all the finicky hex addresses
* Needs a legend for the icon symbols used - ✅❌❗ (in particular I was surprised to see ❗ and can't tell the difference between that and ❌)
* Does ❌ mean where the Hedera result is _different_ than the EVM developer expects?  Seems that way in the SELFDESTRUCT table where of course the SELFDESTRUCT opcode never fails in Ethereum but does sometimes for us.  But then in the Call Operations table it is also used for, e.g., "if address is a contract and we're not using the correct ABI" in the Ethereum column so there it seems to mean something else.  I mean, that's the normal behavior if you call a contract with the wrong ABI.  So the suggestion is to distinguish the very important case of result is _different_ than the EVM developer expects vs "normal" failure.

Nana-EC (changes requested):
Nice.
A few suggestions.

Does this reflect the current logic in mono and modularized, where `INVALID_SOLIDITY_ADDRESS` or others are thrown?

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,146 @@
...
+| 0x16a → 0x2EE (0.0.361 → 0.0.750)  | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns 0 as if account doesn’t exist                                       |
+| 0x2EF → 0x3E8 (0.0.751 → 0.0.1000) | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns balance or 0 if an account or contract doesn’t exist on the address |
+| 0x3E8 → ♾️                         | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns balance or 0 if an account or contract doesn’t exist on the address |
+
+### SelfDestruct operation
+
+**Ethereum:** success for any address. If the beneficiary is the same as contract address, it burns the eth and destructs the contract.
+
+**Hedera:**
+1. For addresses in range 0x0 - 0x2EE (0.0.0 - 0.0.750): fail with status `INVALID_SOLIDITY_ADDRESS`.
```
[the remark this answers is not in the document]
stoqnkpL: We do no-ops for call operations that don't also attempt to transfer funds. For self-destructs that attempt to transfer to system accounts we have to fail, we cannot allow the transfer. This is a place where we decided to diverge from EVM built-in behavior.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,146 @@
...
+2. For addresses in range 0x2EF - 0x3E8 (0.0.751 - 0.0.1000):
+   - success, if account exists and has `receiverSigRequired` == false;
+   - success, if account exists and has `receiverSigRequired` == true and account is `sender`;
+   - fail in some failing scenarios such as: beneficiary same as account-to-be-deleted, `receiverSigRequired` == true and signature provided, etc.
```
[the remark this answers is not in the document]
stoqnkpL: We can add this information to the doc:
If the account to be deleted is the same as the beneficiary address, the operation would fail with status `SELF_DESTRUCT_TO_SELF`.
If the contract to be deleted is a token treasury, the operation would fail with status `CONTRACT_IS_TREASURY`.
If the contract to be deleted has a positive balance, the operation would fail with status `TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES`.
If the contract to be deleted owns NFTs, the operation would fail with status `CONTRACT_STILL_OWNS_NFTS`.
If the beneficiary has configured receiversigrequired = true on their account and are not the sender in the message frame, the operation would fail with status `INVALID_SIGNATURE`.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,146 @@
...
+| 0x16a → 0x2EE (0.0.362 → 0.0.750)  | ✅ - if address is a contract and we’re using the correct ABI<br/>✅ - success with no op, if there is no contract<br/>❌ - if address is a contract and we’re not using the correct ABI | ✅ - success with no op                                                                                                   |
+| 0x2EF → 0x3E8 (0.0.751 → 0.0.1000) | ✅ - if address is a contract and we’re using the correct ABI<br/>✅ - success with no op, if there is no contract<br/>❌ - if address is a contract and we’re not using the correct ABI | ✅ - success with no op                                                                                                   |
+| 0x3E8 → ♾️ | ✅ - if address is a contract and we’re using the correct ABI<br/>✅ - success with no op, if there is no contract                                                                 | ✅ - if a contract exists and we’re using the correct ABI<br/>❌ - INVALID_SOLIDITY_ADDRESS if a contract does not exist |
```
stoqnkpL, line 107:
```suggestion
| 0x3E8 → ♾️ | ✅ - if address is a contract and we’re using the correct ABI<br/>✅ - success with no op, if there is no contract                                                                 | ✅ - if a contract exists and we’re using the correct ABI<br/>✅  - no op success if a contract does not exist |
```

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,146 @@
+# System accounts
+
+There are some differences when calling operations on system accounts in Hedera and in Ethereum.
```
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-node/docs/system-accounts-operations.md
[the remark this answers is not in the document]
bilyana-gospodinova: Comment addressed.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Comment addressed.

kselveliev (dismissed):
LGTM

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+| 0x167 (0.0.359)                    | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns 0 as if account doesn’t exist                                       |
+| 0x168, 0x169 (0.0.360, 0.0.361)    | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns 0 as if account doesn’t exist                                       |
+| 0x16a → 0x2EE (0.0.361 → 0.0.750)  | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns 0 as if account doesn’t exist                                       |
+| 0x2EF → 0x3E8 (0.0.751 → 0.0.1000) | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns balance or 0 if an account or contract doesn’t exist on the address |
+| 0x3E8 → ♾️                         | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns balance or 0 if an account or contract doesn’t exist on the address |
+
+### SelfDestruct operation
+
+**Ethereum:** success for any address. If the beneficiary is the same as contract address, it burns the eth and destructs the contract.
```
david-bakin-sl, line 80: re "destructs the contract" - no longer true post Cancun (which isn't deployed yet, admittedly, but will be soon) - Cancun semantics for self destruct will apply: Sweep (if possible, otherwise fail as the notes below show) then self destruct _only if created in same transaction_)

david-bakin-sl (commented):
Comment about selfdestruct w.r.t. cancun. 

Doc can be changed now to mention it, or we can remember to change it in the release that Cancun ships with.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+- 0.0.50 - system admin;
+- 0.0.55 - address book admin;
+- 0.0.57 - exchange rates admin;
+- 0.0.58 - freeze admin;
+- 0.0.59 - system delete admin;
+- 0.0.60 - system undelete admin;
+- 0.0.800 - staking reward account;
+- 0.0.801 - node reward account.
+
+Addresses in range 1 → 9 are special addresses on which the Ethereum precompiled contracts exist. This is valid for both Ethereum and Hedera.
```
david-bakin-sl, line 22: BTW, that's the _current_ configuration of Ethereum.  This will expand in the future. (In fact `0x0A` through `0x13` are already reserved in BESU for EIP-2537 and over time there will be more.  

* (Also, to make matters more confusing, there is now with Cancun an EVM address for the new Ethereum "Beacon Root contract" - which is a contract, _not_ a precompile - that is special to Ethereum.  But this document needn't go in to it.)

Maybe also change "Addresses in range ..." to "EVM addresses in range ..."?

david-bakin-sl (commented):
One more comment.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+
+Lastly, the system accounts can be divided in these two groups:
+- addresses in range 0 → 750 - accounts that reject hbar transfers. An exception is the HTS precompiled contract that can receive transfers.
```
Nana-EC, line 25: Also why is 0x167 (0.0.359) able to receive HBAR?! 
```suggestion
- addresses in range 0 → 750 - accounts that reject hbar transfers. An exception is the HTS system contract that can receive transfers.
```

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
+# System accounts
+
+## Definition
+The Hedera network reserves the first 1000 entity numbers for its own uses, and any account with a number in this reserved range is referred to as a system account.
+
+The following addresses are reserved for the Hedera precompiled contracts:
```
Nana-EC, line 6: Please change the following 3 reserved address to not system not precompile.
The only precompiles are the EVM precompiles
```suggestion
The following addresses are reserved for the Hedera system contracts:
```

Nana-EC (commented):
Lots of good improvements.
Please correct the mentions of precompile though

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+- 0x167 (0.0.359) - HTS precompile;
+- 0x168 (0.0.360) - ExchangeRate precompile;
+- 0x169 (0.0.361) - PRNG precompile.
```
lukelee-sl, line 9: The set of addresses here can also expand in the future.   We are already working on 0x16a for HAS system contracts

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+
+**Ethereum:** Copy an account’s code to memory. If a contract doesn’t exist on the address, it copies empty bytes.
+
+**Hedera:** For addresses in range 0x0 → 0x3E8 (0.0.0 → 0.0.1000) it copies empty bytes.
+
+`ExtCodeHash`
+
+**Ethereum:** Hash of the chosen account's code, the empty hash (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) if the account has no code, or 0 if the account does not exist or has been destroyed.
+
+**Hedera:** For addresses in range 0x0 → 0x3E8 (0.0.0 → 0.0.1000) it returns the empty hash.
```
lukelee-sl, line 57: empty or 0x hash?

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,184 @@
...
+
+**Ethereum:** Returns balance or 0 if an account or contract doesn’t exist on the address. This is valid for any address.
+
+**Hedera:**
+1. For addresses in range 0x0 → 0x2EE (0.0.0 → 0.0.750): returns 0 as if account doesn’t exist.
+2. For addresses in range 0x2EF → 0x3E8 (0.0.751 → 0.0.1000): returns balance or 0 if an account or contract doesn’t exist on the address. This behavior is the same for addresses above 0.0.1000.
+
+| Address                            | in Ethereum                                                                   | in Hedera                                                                     |
+|------------------------------------|-------------------------------------------------------------------------------|-------------------------------------------------------------------------------|
+| 0x0                                | ✅ returns balance or 0 if an account or contract doesn’t exist on the address | ✅ returns 0 as if account doesn’t exist                                       |
```
lukelee-sl, line 69: I believe we always return 0x if below 0.0.751

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Only for token create cases. I will clarify that in the doc.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Done.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Updated.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Just for the token create scenario. This was explained in the last section regarding the transfers but I will clarify that here as well.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: The long-zero address actually. The current text is misleading in this case as it looks like the empty hash is the same as in Ethereum but it isn't. I will update this.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: As far as I see we always return 0 for EVM addresses below 751. We have a test for that - `Evm46ValidationSuite#testBalanceOfForSystemAccounts`.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: It seems that the Cancun upgrade went live on March 13 so I will update the documentation according to your comment. Thanks for pointing that out.

hedera-node/docs/system-accounts-operations.md
Nana-EC: Q: Why is this?
@stoqnkpL Did we miss this scenario?
For example there are some system account with balance below 751, why not return the valid balance?
This is returning incorrect data if not.

stoqnkpL: When we designed this we wanted the EVM to behave as if this account does not exist and can't be created. We intentionally decided to return 0 as if the account doesn't exist.

Nana-EC: Ah yes, so far as it can't receive balance then this is valid, but if it can this is invalid.

Nana-EC (dismissed):
Unblocking as this captures the current logic and is in the right direction.
Any updates and differences should see new tickets and changes in modularized code base

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,187 @@
...
+     - fail, if the address is a contract, and we are **not** using the correct ABI.
+   - **Hedera:** success, the HTS, ExchangeRate and PRNG system contracts exist on these addresses accordingly.
+5. For address range 0x16A → 0x3E8 (0.0.362 → 0.0.1000):
+   - **Ethereum:**
+     - success, if the address is a contract, and we are using the correct ABI;
+     - success with no op, if there is no contract;
+     - fail, if the address is a contract, and we are **not** using the correct ABI.
+   - **Hedera:** success with no op.
+
+_Please note that the expected behavior above is valid considering there is no `value` passed. If there is `value`, this falls in the next section._
```
bilyana-gospodinova, line 137: Add clarification which is the next section - "Transfer and send operations".

kselveliev (dismissed):

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Updated.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Hedera:**
+1. For addresses in range 0x0 → 0x2EE (0.0.0 → 0.0.750): fail with status `INVALID_SOLIDITY_ADDRESS`.
+2. For addresses in range 0x2EF → 0x3E8 (0.0.751 → 0.0.1000):
+   - success, if account exists and has `receiverSigRequired` == false;
+   - success, if account exists and has `receiverSigRequired` == true and account is `sender`;
+   - fail in some failing scenarios such as: 
+     - beneficiary same as account-to-be-deleted - fail with status `SELF_DESTRUCT_TO_SELF`;
+     - contract-to-be-deleted is a token treasury - fail with status `CONTRACT_IS_TREASURY`;
+     - contract-to-be-deleted has a positive balance - fail with status `TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES`;
```
IvanKavaldzhiev, line 93: Probably we can elaborate more on this case, e.g.:

contract-to-be-deleted has a positive `fungible token` balance, 

so that we are not misinterpreting this for hbar balance.

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,187 @@
...
+| 0x3E8 → ♾️ | ✅ - if address is a payable contract<br/>✅ - if address is an account<br/>✅ - if address is empty, creates account<br/>❌ - if address is a non-payable contract | ✅ - if address is a payable contract<br/>✅ - if address is an account with receiversigrequired = false<br/>✅ - if address is empty, creates account<br/>❌ - if address is a non-payable contract<br/>❌ - if address is an account with receiversigrequired=true | ✅ - if address is a payable contract<br/>✅ - if address is an account with receiversigrequired = false<br/>✅ - if address is empty, creates account<br/>✅ - if address is a non-payable contract<br/>❌ - if address is an account with receiversigrequired=true |
```
IvanKavaldzhiev, line 187: Typos in receiversigrequired -> should be receiverSigRequired

hedera-node/docs/system-accounts-operations.md
```diff
@@ -0,0 +1,187 @@
...
+**Hedera:**
+1. For addresses in range 0x0 → 0x2EE (0.0.0 → 0.0.750):
+    - Transfers using `.send` and `.transfer`: fail with status `INVALID_FEE_SUBMITTED`. Exception: `tokenCreate` to address 0x167 which is successful.
+    - Transfer of HTS tokens through the HTS system contract: fail with status `INVALID_RECEIVING_NODE_ACCOUNT`.
+2. For addresses in range 0x2EF → 0x3E8 (0.0.751 → 0.0.1000):
+    - success, if account exists and has `receiverSigRequired` == false;
+    - success, if account exists and has `receiverSigRequired` == false and account is `sender`;
+   
+_The account is not created if it doesn't exist._
+3. For addresses above 0.0.1000:
```
IvanKavaldzhiev, line 169: The formatting of point 3 is wrong. It doesn't display as a table but plain text.

bilyana-gospodinova: This line is not meant to be a part of a table. The table starts from line 178.

hedera-node/docs/system-accounts-operations.md
bilyana-gospodinova: Fixed

tinker-michaelj (approved):
LGTM, tyvm @bilyana-gospodinova - we can always refine details in future PRs, this is a great start. 🙏

edward-swirldslabs (approved):

Issue #12159, opened 2024-03-15
HIP-540: Implement and test the logic that all keys can update themselves to another valid key


MiroslavGatsanoga:
Currently admin key is able to change itself and also other low privilege keys.
Additionally we want to implement the ability for low privilege keys to be able to update themselves to new valid keys (i.e. without requiring admin key signature but instead only the signature of the corresponding low privilege key being updated).

PR #12329, opened 2024-03-25
feat: HIP-540:  Implement and test the logic that all keys can update themselves to another valid key

stefan-stefanooov:
**Description**:
Currently admin key is able to change itself and also other low privilege keys.
Additionally we want to implement the ability for low privilege keys to be able to update themselves to new valid keys (i.e. without requiring admin key signature but instead only the signature of the corresponding low privilege key being updated).

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/12159
Fixes #
https://github.com/hashgraph/hedera-services/issues/12159


petreze:
Changes are present in the following [PR](https://github.com/hashgraph/hedera-services/pull/12327) for better readability. Closing this one

PR #12366, opened 2024-03-26
chore: remove verify signature method from DTO

### LINKED ISSUE #12214: Remove usage of `com.swirlds.common.crypto.Signature.verifySignature()`
Currently, the validation of a signature is done inside `com.swirlds.common.crypto.Signature.verifySignature()`. Since signatures will be moving to protobuf, the class used will be autogenerated, so it cannot have any additional methods. 

### Tasks
- the utility that should be used for validating signatures already exists in `CryptoStatic.verifySignature()`
- use this utility instead of the method `com.swirlds.common.crypto.Signature.verifySignature()`
- check if `LinkedObjectStreamValidateUtils` is used
  - if no: remove it
  - if yes: alter it to use the new utility

lpetrovic05:
closes  #12214


alittley (dismissed):

alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/stream/internal/LinkedObjectStreamValidateUtils.java
```diff
@@ -225,4 +227,17 @@ public static <T extends SelfSerializable> Pair<StreamValidationResult, Hash> va
...
+    // Code duplicated in order to avoid wasting time on the event stream which will be removed soon
+    public static boolean verifySignature(final byte[] data, final Signature signature, final PublicKey publicKey) {
+        try {
+            final java.security.Signature sig = java.security.Signature.getInstance(
+                    SignatureType.RSA.signingAlgorithm(), SignatureType.RSA.provider());
+            sig.initVerify(publicKey);
+            sig.update(data);
+            return sig.verify(signature.getSignatureBytes());
+        } catch (final Exception ignored) {
+        }
```
anthony-swirldslabs, line 240: This may be extremely difficult to debug. Please consider re-throwing a `RuntimeException` wrapping the original exception here, or at least log the original exception. Otherwise, please add comments in code explaining why any and all exceptions are safe to silently swallow here (there's hardly ever a good reason for this, unless it's a very specific and relatively benign exception such as the `InterruptedException` or similar.)

lpetrovic05: The main reason I kind of glossed over this is that this code will be removed soon. I now added logging that was there in the original method that was removed.

timo0 (approved):
Approved for base team files

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/stream/internal/LinkedObjectStreamValidateUtils.java
anthony-swirldslabs: Thank you!

alittley (approved):

PR #12369, opened 2024-03-26
feat: Pass Metrics to created VirtualMaps

### LINKED ISSUE #12356: VirtualMap / VirtualDataSource metrics are missing, not registered
Virtual map and data source metrics aren't reported in mod-services mode. It appeared that there was a call to `VirtualMap.registerMetrics()` in mono-services code, but it's missing for some reason in mod-services.

netopyr:
This PR ensures we register `Metrics` with all `VirtualMap` instances we create.

Fixes #12356


tinker-michaelj (approved):
LGTM, ty @netopyr !

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12377, opened 2024-03-26
feat: Reloadable TLSFactory to support dynamic addressbook

### LINKED ISSUE #12299: Investigate if a new server socket needs to be created when adding/removing nodes

kfa-aguda:
Closes #12299 

This PR re-works how we establish TLS socket connections by making it possible to change the list of PeerInfo at runtime, and re-initializing the trust-store, without disconnecting existing sessions whilst allowing new trusted ones.


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
```diff
@@ -47,4 +49,9 @@ public TcpFactory(@NonNull final SocketConfig socketConfig) {
         SocketFactory.configureAndConnect(clientSocket, socketConfig, hostname, port);
         return clientSocket;
     }
+
+    @Override
+    public void handlePeerListUpdate(List<PeerInfo> peerInfoList) {
+        // not implemented for TCP Factory
```
alittley, line 55: should this throw, rather than silently doing nothing?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -123,4 +107,39 @@ public TlsFactory(
         clientSocket.startHandshake();
         return clientSocket;
     }
+
+    @Override
+    public void handlePeerListUpdate(List<PeerInfo> peers) {
```
alittley, line 112: `peers` is missing `final` and nullity annotation

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/SocketFactoryTest.java
```diff
@@ -162,6 +164,44 @@ void tlsFactoryTest(final AddressBook addressBook, final Map<NodeId, KeysAndCert
                 NetworkUtils.createSocketFactory(node2, addressBook, keysAndCerts2, TLS_IP_TOS_CONFIG));
     }
 
+    @ParameterizedTest
+    @MethodSource({"com.swirlds.platform.crypto.CryptoArgsProvider#basicTestArgs"})
+    void tlsFactoryTestDynamicPeers(final AddressBook addressBook, final Map<NodeId, KeysAndCerts> keysAndCerts)
+            throws Throwable {
+        assertTrue(addressBook.getSize() > 1, "Address book must contain at least 2 nodes");
+        // choose 2 random nodes to test
+        final Random random = new Random();
```
alittley, line 173: Random seed should be printed, so failing test can be reproduced

```
        final Random random = RandomUtils.getRandomPrintSeed();
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
kfa-aguda: It's an optional implementation. Naked tcp connections don't have a need to do this yet, and probably never will (because this entire class goes away in my next PR).

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/SocketFactoryTest.java
kfa-aguda: Nice, thank you.

alittley (commented):
Approved, with respect to java code correctness and style.

I'm not plugged into the problem enough to give feedback on implementation choices, so I'll defer to Lazar in that regard

alittley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
```diff
@@ -162,12 +158,7 @@ public static String formatException(final Throwable e) {
                     Utilities.createPeerInfoList(addressBook, selfId),
                     socketConfig,
                     cryptoConfig);
-        } catch (final NoSuchAlgorithmException
-                | UnrecoverableKeyException
-                | KeyStoreException
-                | KeyManagementException
-                | CertificateException
-                | IOException e) {
+        } catch (NoSuchAlgorithmException e) {
```
cody-littley, line 161:
```suggestion
        } catch (final NoSuchAlgorithmException e) {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/SocketFactory.java
```diff
@@ -126,4 +128,13 @@ static void configureAndConnect(
      */
     @NonNull
     Socket createClientSocket(@NonNull final String hostname, final int port) throws IOException;
+
+    /**
+     * A convenient handler for executing logic on dynamic peers
+     * e.g. it could be used to re-initialize certificate keystore, or ask peers to re-authenticate
+     *
+     * @param peerInfoList
+     *  the updated list of peers
```
cody-littley, line 137: Javadoc line wrapping is a bit funky. Suggest using intellij to auto-format.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/SocketFactory.java
```diff
@@ -126,4 +128,13 @@ static void configureAndConnect(
...
+     */
+    void handlePeerListUpdate(List<PeerInfo> peerInfoList);
```
cody-littley, line 139:
```suggestion
    void handlePeerListUpdate(@NonNull List<PeerInfo> peerInfoList);
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
```diff
@@ -47,4 +49,9 @@ public TcpFactory(@NonNull final SocketConfig socketConfig) {
         SocketFactory.configureAndConnect(clientSocket, socketConfig, hostname, port);
         return clientSocket;
     }
+
+    @Override
+    public void handlePeerListUpdate(List<PeerInfo> peerInfoList) {
```
cody-littley, line 54:
```suggestion
    public void handlePeerListUpdate(@NonNull final List<PeerInfo> peerInfoList) {
```

cody-littley, line 54: Naming suggestion. How about `updatePeers()`?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -123,4 +107,40 @@ public TlsFactory(
         clientSocket.startHandshake();
         return clientSocket;
     }
+
+    @Override
+    public void handlePeerListUpdate(final @NonNull List<PeerInfo> peers) {
```
cody-littley, line 112: Not sure if this is in the coding standard, but within our team we usually put the nullity annotation before the `final` keyword.
```suggestion
    public void handlePeerListUpdate(@NonNull final List<PeerInfo> peers) {
```

cody-littley, line 112: In some places, this method uses the parameter name `PeerInfoList`, but here it uses the parameter name `peers`. My personal preference would be to use `peers` everywhere. Whichever you choose, probably best to use the same in all places.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
```diff
@@ -39,23 +39,23 @@ public class CryptoArgsProvider {
...
-        final AddressBook genAB = createAddressBook();
+        final AddressBook genAB = createAddressBook(NUMBER_OF_ADDRESSES);
         final Map<NodeId, KeysAndCerts> genC = CryptoStatic.generateKeysAndCerts(genAB);
         System.out.println(
                 "Key generating took " + Duration.between(start, Instant.now()).toMillis());
         return Stream.of(Arguments.of(loadedAB, loadedC), Arguments.of(genAB, genC));
     }
 
-    private static AddressBook createAddressBook() {
+    public static AddressBook createAddressBook(int size) {
```
cody-littley, line 56:
```suggestion
    public static AddressBook createAddressBook(final int size) {
```

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/SocketFactoryTest.java
```diff
@@ -162,6 +165,44 @@ void tlsFactoryTest(final AddressBook addressBook, final Map<NodeId, KeysAndCert
                 NetworkUtils.createSocketFactory(node2, addressBook, keysAndCerts2, TLS_IP_TOS_CONFIG));
     }
 
+    @ParameterizedTest
+    @MethodSource({"com.swirlds.platform.crypto.CryptoArgsProvider#basicTestArgs"})
+    void tlsFactoryTestDynamicPeers(final AddressBook addressBook, final Map<NodeId, KeysAndCerts> keysAndCerts)
```
cody-littley, line 170: nullity annotations for parameters

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -123,4 +107,40 @@ public TlsFactory(
...
+            trustManagerFactory.init(signingTrustStore);
+            SSLContext.setDefault(sslContext);
+            sslContext.init(keyManagerFactory.getKeyManagers(), trustManagerFactory.getTrustManagers(), nonDetRandom);
+        } catch (final NoSuchAlgorithmException
+                | UnrecoverableKeyException
+                | KeyStoreException
+                | KeyManagementException
+                | CertificateException
+                | IOException e) {
+            throw new PlatformConstructionException("A problem occurred while creating the SocketFactory", e);
```
lpetrovic05, line 143: when updating peers, the exception would not be a construction exception

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -123,4 +107,40 @@ public TlsFactory(
...
+
+            // "PKIX" may be more interoperable than KeyManagerFactory.getDefaultAlgorithm or
+            // TrustManagerFactory.getDefaultAlgorithm(), which was "SunX509" on one system tested
+            final KeyManagerFactory keyManagerFactory =
+                    KeyManagerFactory.getInstance(CryptoConstants.KEY_MANAGER_FACTORY_TYPE);
+            keyManagerFactory.init(agrKeyStore, password);
+            final TrustManagerFactory trustManagerFactory =
+                    TrustManagerFactory.getInstance(CryptoConstants.TRUST_MANAGER_FACTORY_TYPE);
+            trustManagerFactory.init(signingTrustStore);
+            SSLContext.setDefault(sslContext);
```
lpetrovic05, line 135: I think we should avoid setting the default, as we will have multiple instances of SSLContext

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/SocketFactoryTest.java
```diff
@@ -162,6 +165,44 @@ void tlsFactoryTest(final AddressBook addressBook, final Map<NodeId, KeysAndCert
...
+        // assert they can still connect to each other
+        testSockets(socketFactory1, socketFactory2);
+
+        // this test is limited, but sufficient for this proof of concept.
+        // In production, clients from socketFactory2
+        // wouldn't successfully handshake with server from socketFactory1 as
+        // socketFactory1's peerList now excludes all addresses in socketFactory2
+    }
+
     @Test
```
lpetrovic05, line 206: I think we are missing some test cases:
- when the list of peers is updated, are the connections created before the update still open and can they still transmit data
- a peer tries to connect, but his certificate is still not whitelisted, so it fails. the context is then updated with the new peer, and this time it can connect

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/SocketFactoryTest.java
lpetrovic05, line 170: I think it would be better if these tests had only peer lists as input

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/SocketFactory.java
kfa-aguda: thanks. I've turned this on so it's automatic on commit.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
kfa-aguda: thing is it doesn't really *have to* update the peers, though it may, depending on implementation. The goal of the method is to allow the factory take action based on the updated list of peers. Still open to name suggestion though but _updatePeers_ *could be* misleading.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
kfa-aguda: it does make sense to set the default ssl context to that which belongs to the version we've chosen. The JVM may have different ssl context instances but we do* want the default to be for our chosen TLS version.

Anyway, how do you think we could end up with multiple instances?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/SocketFactory.java
```diff
@@ -126,4 +128,12 @@ static void configureAndConnect(
      */
     @NonNull
     Socket createClientSocket(@NonNull final String hostname, final int port) throws IOException;
+
+    /**
+     * A convenient handler for reloading connectivity components
+     * e.g. it could be used to reload a TLS trust store, or ask peers to re-authenticate
```
cody-littley, line 134:
```suggestion
    /**
     * Update the current list of peers with which we wish to be connect.
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
cody-littley, line 1: Are we still planning on deleting this class?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -123,4 +132,25 @@ public TlsFactory(
         clientSocket.startHandshake();
         return clientSocket;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void reload(@NonNull final List<PeerInfo> peers) {
+        lock.lock();
```
cody-littley, line 141: What is the purpose of this lock? This class as a whole is not thread safe if we have multi-threaded access, and currently `reload` is only called from this class's constructor.

cody-littley: Before we add locks to all of the methods in this class, it might be helpful to define the access pattern we intend to support with this class. It may be the case that we can guarantee an access pattern that does not require locking.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/PlatformInitializationException.java
```diff
@@ -0,0 +1,26 @@
...
+package com.swirlds.platform.system;
+
+/**
+ * Thrown when an issue occurs while initializing/re-initializing platform instances
+ */
+public class PlatformInitializationException extends RuntimeException {
+    public PlatformInitializationException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+}
```
cody-littley, line 26: I know the "conventional wisdom" is to always create specific exceptions for things. My personal take is that having a large number of exceptions that you don't intend to catch and handle doesn't really buy much. In this situation, I think a naked `RuntimeException` or `IllegalStateException` would be sufficient.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
```diff
@@ -39,23 +44,22 @@ public class CryptoArgsProvider {
      */
     static Stream<Arguments> basicTestArgs() throws Exception {
         Instant start = Instant.now();
-        final AddressBook loadedAB = createAddressBook();
-        final Map<NodeId, KeysAndCerts> loadedC =
-                CryptoStatic.loadKeysAndCerts(loadedAB, ResourceLoader.getFile("preGeneratedKeysAndCerts/"), PASSWORD);
+        final Pair<AddressBook, Map<NodeId, KeysAndCerts>> addressBookWithKeys =
+                loadAddressBookWithKeys(NUMBER_OF_ADDRESSES);
```
cody-littley, line 48: I suggest avoiding the `Pair<>` type, and to instead create a simple record. Records are simple to create and create much more readable code.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
```diff
@@ -68,6 +72,20 @@ private static AddressBook createAddressBook() {
         return addresses;
     }
 
+    /**
+     * returns the addressBook with keys loaded from file.
+     *
+     * @param size the size of the required address book
+     */
+    public static Pair<AddressBook, Map<NodeId, KeysAndCerts>> loadAddressBookWithKeys(final int size)
```
cody-littley, line 80:
```suggestion
    @NonNull
    public static Pair<AddressBook, Map<NodeId, KeysAndCerts>> loadAddressBookWithKeys(final int size)
```

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/ConnectivityTestBase.java
```diff
@@ -55,19 +60,33 @@ class ConnectivityTestBase {
...
+     * - leaves the server socket open until told to close
      *
-     * @param serverSocket the server socket to listen on
+     * @param serverSocket  the server socket
+     * @param data          the data to be read from the socket
+     * @param stopFlag      flag for stopping the thread
      */
-    static Thread createSocketThread(final ServerSocket serverSocket) {
-        return new Thread(() -> {
+    static Thread createSocketThread(
```
cody-littley, line 71:
```suggestion
    @NonNull
    static Thread createSocketThread(
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TcpFactory.java
kfa-aguda: yes. In the next PR or two.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/PlatformInitializationException.java
kfa-aguda: I actually agree that we either use one of the naked exceptions or use one of the closest exceptions we have defined e.g.`PlatformConstructionException` - which is what I initially used but @lpetrovic05 has a different view, so I created this one.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
kfa-aguda: Preferences :) I actually *really do* happen to like the Pair construct more - they read exactly like what they do, hold `a pair of things`. No strong preference so I'll remove.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
kfa-aguda: thank you.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
kfa-aguda: `reload` will eventually be called from the outside of this class, and we do need the operations to be atomic when it does.

>It may be the case that we can guarantee an access pattern that does not require locking
This is true but i don't think it's the case here. This class is doing a lot of long ish operations that may interleave easily (e.g. other threads may easily access the default SSL Context), locking seems the reasonable approach to achieving atomicity.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
lpetrovic05: I would not add a lock prematurely, I think we should avoid concurrent access if possible

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
```diff
@@ -88,19 +95,18 @@ public TlsFactory(
...
-                KeyManagerFactory.getInstance(CryptoConstants.KEY_MANAGER_FACTORY_TYPE);
+        this.keyManagerFactory = KeyManagerFactory.getInstance(CryptoConstants.KEY_MANAGER_FACTORY_TYPE);
         keyManagerFactory.init(agrKeyStore, password);
-        final TrustManagerFactory trustManagerFactory =
-                TrustManagerFactory.getInstance(CryptoConstants.TRUST_MANAGER_FACTORY_TYPE);
-        trustManagerFactory.init(signingTrustStore);
-        final SSLContext sslContext = SSLContext.getInstance(CryptoConstants.SSL_VERSION);
+        this.trustManagerFactory = TrustManagerFactory.getInstance(CryptoConstants.TRUST_MANAGER_FACTORY_TYPE);
+        this.sslContext = SSLContext.getInstance(CryptoConstants.SSL_VERSION);
         SSLContext.setDefault(sslContext);
```
lpetrovic05, line 102: we should not set the default

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/PlatformInitializationException.java
kfa-aguda: https://github.com/hashgraph/hedera-services/pull/12377#discussion_r1543003765

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/TlsFactory.java
kfa-aguda: As discussed, this isn't a premature lock. The public access to the static `SSLContext.getDefault()` outside of this class legitimizes a case for locking. @lpetrovic05 proposed exploring the removal of `SSLContext.setDefault()`. This is old code. I'm working on removing it, and subsequently the lock.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/connectivity/SocketFactory.java
kfa-aguda: on a second thought, this method isn't updating the list of peers. It's reloading the trust store with certificates presented by the peers. Maybe 
`   * reloads the trust store with peer certificates` 
is better?

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/TlsFactoryTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * The tests in this class are unidirectional server socket tests. For bidirectional tests, see
```
lpetrovic05, line 40: javadoc seems to be out of date

lpetrovic05 (dismissed):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/network/connectivity/ConnectivityTestBase.java
```diff
@@ -59,37 +58,55 @@ class ConnectivityTestBase {
         IP_TOS = configurationIpTos.getConfigData(SocketConfig.class);
     }
 
+    /**
+     * creates a server socket thread for testing purposes
+     *
+     * @param serverSocket the server socket to listen on
+     */
+    static Thread createSocketThread(@NonNull final ServerSocket serverSocket) {
```
cody-littley, line 66:
```suggestion
    @NonNull
    static Thread createSocketThread(@NonNull final ServerSocket serverSocket) {
```

cody-littley (approved):

Issue #12388, opened 2024-03-26
VirtualMap / DataSource metrics should not contain certain chars in the names


artemananiev:
New mod-services virtual maps are labeled like "Service.TYPE" The corresponding metric names are "vmal_<xyz>_Service.TYPE" Such names are rejected somewhere at JRS level.

One option is to replace "." with "_" in metric names, both for virtual map and data source metrics.

PR #12390, opened 2024-03-26
fix: 12388: VirtualMap / DataSource metrics should not contain certain chars in the names

artemananiev:
Fix summary: `.` are replaced with `_` in generated virtual map and data source metric names.

Fixes: https://github.com/hashgraph/hedera-services/issues/12388

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
@@ -125,7 +125,9 @@ public class MerkleDbStatistics {
      */
     public MerkleDbStatistics(final MerkleDbConfig dbConfig, final String label) {
         this.dbConfig = dbConfig;
-        this.label = Objects.requireNonNull(label, "label must not be null");
+        Objects.requireNonNull(label, "label must not be null");
+        // If the label contains ".", they are replaced with "_", since metric names may not contain "."
+        this.label = label.replace('.', '_');
         hashesStoreCompactionTimeMsList = new ArrayList<>();
         hashesStoreCompactionSavedSpaceMbList = new ArrayList<>();
         hashesStoreFileSizeByLevelMbList = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
```diff
@@ -125,7 +125,9 @@ public class MerkleDbStatistics {
      */
     public MerkleDbStatistics(final MerkleDbConfig dbConfig, final String label) {
         this.dbConfig = dbConfig;
-        this.label = Objects.requireNonNull(label, "label must not be null");
+        Objects.requireNonNull(label, "label must not be null");
+        // If the label contains ".", they are replaced with "_", since metric names may not contain "."
+        this.label = label.replace('.', '_');
```
anthony-swirldslabs, line 130: Minor: it may be confusing for callers of this constructor to have their metric labels modified silently. I'd suggest to fix the callers to pass valid metric names in here, and this constructor should simply throw if the label is invalid (e.g. if it contains periods.) This solution would be a lot less confusing and a lot more robust.

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatistics.java
artemananiev: @anthony-swirldslabs It can be done like you propose, but I am not sure it would look better. For example, at the virtual map level, `statistics` (of type `VirtualMapStatistics`) is created in `VirtualRootNode`:

```java
        if (statistics == null) {
            // Only create statistics instance if we don't yet have statistics. During a reconnect operation.
            // it is necessary to use the statistics object from the previous instance of the state.
            this.statistics = new VirtualMapStatistics(state.getLabel());
        }
```

This class, virtual root node, is not related to metrics at all. It has no knowledge what metrics names are right, and what are wrong. With your proposal, this line should be changed, so `state.getLabel()` is checked to contain illegal chars, corrected, and only then passed to `VirtualMapStatistics` constructor. This would better be done in the statistics classes (`VirtualMapStatistics`, `MerkleDbStatistics`), i.e. where it is done in this fix, to keep virtual root node code clean off metrics stuff.

anthony-swirldslabs: I was thinking more about either making sure that the `state.getLabel()` returns a name that is valid from metrics perspective in the first place, or having a `public static String formatMetricLabel(String)` formatter that would be called explicitly when passing the label to the `VirtualMapStatistics`/`MerkleDbStatistics` constructors, rather than implicitly relying on these constructors to do the job. As I mentioned, my main goal is to avoid a confusion where this reformatting happens silently w/o the calling code having an idea about it.

Issue #12269, opened 2024-03-21
EPIC: MATs (Minimally Accepted Test Suite)


anastasiya-kovaliova:
### Problem

The CI checks have execution time over 1h 30m which we need to reduce to maximum 30 minutes.
Analyse the CI and the Tests and in order to plan what can we optimised in order to acheve the goal of 30m execution time.
The following areas are in scope of the Analyses:
- E2E tests in the test-clients framework:
- Unit Tests in Services
- CI workflows configuration
- Smart Contracts repo 

### Solution

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-services/issues/12295
```

PR #12393, opened 2024-03-27
chore: Add github actions for uploading junit test reports as artifacts

### LINKED ISSUE #12295: Upload junit test reports as artifacts
Currently junit reports from test are generated and processed as part of the CI pipelines, but the files themselves are not available. 

To better support the optimization of tests runtime they need to be available as an artifact on job runs.

yaroslav-007:
**Description**:
Add github actions for uploading the junit test reports as artifacts.

Fixes #12295

**Checklist**


nathanklick (changes requested):
These should not be uploaded due to size concerns.

nathanklick:
@yaroslav-007 We should limit these to only uploading the artifacts when the associate test step fails.

isavov:
@nathanklick  This PR is in support of #12269 with the goal of allowing QA engineers to analyze in detail the time each test consumes. In that sense it would be more useful for them to have the junit reports for both successful and failed runs. Currently we upload the full network logs as well, which are often bigger than the reports.
To facilitate the intent of the original issue I would suggest we could do either or all of the following:
1. Always upload test reports artifacts, but limit retention time using the `retention-days` config in the action to a short time (5-7 days)
2. Upload the network node logs only on failed tests
What do you think?

yaroslav-007:
Hello @nathanklick,

I have change the upload action to be trigger when there is failure and artifacts to be kept for 7 days.

isavov (approved):

nathanklick (approved):

PR #12396, opened 2024-03-27
fix: remove admin key as defaultSigner for HapiTokenUpdate

### LINKED ISSUE #12395: HIP-540: Remove token admin key from `defaultSigners()` in `HapiTokenUpdate`
We should remove token admin key from `defaultSigners()` in `HapiTokenUpdate` because after HIP-540, this will no longer be valid assumption to require admin_key signature on every `TokenUpdateTx`

This will require a lot of suite fixes because of the change in the default assumption about this operation

petreze:
**Description**:
Removes `admin key` as default signer from `HapiTokenUpdate` since we are not expecting this to be a requirement for every `TokenUpdateTx`

**Related issue(s)**:

Fixes #12395 

**Notes for reviewer**:
After HIP-540, we drastically change the signing requirements for the `TokenUpdateTx`, hence the `admin key` is not required for every token update

**Checklist**


tinker-michaelj (dismissed):
Looks quite sensible, tyvm @petreze !

PR #12401, opened 2024-03-27
fix: Fix to deterministic iteration order

Neeharika-Sompalli:


hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/schemas/InitialModServiceConsensusSchema.java
```diff
@@ -71,8 +71,9 @@ public void migrate(@NonNull final MigrationContext ctx) {
             final var topicStoreRef = new AtomicReference<>(ctx.newStates().<TopicID, Topic>get(TOPICS_KEY));
             log.info("BBM: running consensus migration...");
 
-            fs.forEach((k, v) -> {
-                final var pbjTopic = stateToPbj(v);
+            fs.keySet().stream().sorted().forEach(k -> {
```
iwsimon, line 74: Follow @cody-littley's suggest, use this instead?
fs.treeIterator().setFilter(v -> v instanceof MerkleTopic).forEachRemaining(v -> {

Neeharika-Sompalli: yes used `MerkleMapLike.forEachNode`. Does same logic as far as I see

iwsimon (approved):
LGTM. Thanks @Neeharika-Sompalli !

PR #12407, opened 2024-03-27
fix: always finalize pending precompile `CONTRACT_ACTION` in `MessageCallProcessor.start()`

### LINKED ISSUE #12361: `tracePrecompileResult()` called for precompile call with insufficient gas
### Description

The EVM "action stack" got out of sync with the `MessageFrame` stack after calling an EVM precompile with insufficient gas.

### Steps to reproduce

As above.

### Additional context

`MessageCallProcessor.executePrecompile()` [here](https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/processor/MessageCallProcessor.java#L153) never calls `tracePrecompileCall()` in the case of insufficient gas.

So unconditionally calling `tracePrecompileResult()` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/contracts/execution/HederaMessageCallProcessorV038.java#L80) will incorrectly pop the action stack. 

### Hedera network

testnet

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12361


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -167,17 +168,23 @@ private void doExecutePrecompile(
...
-            final var result = precompile.computePrecompile(frame.getInputData(), frame);
+            result = precompile.computePrecompile(frame.getInputData(), frame);
+            // NOTE: the tracePrecompileCall() callback is a no-op with the current tracer
             tracer.tracePrecompileCall(frame, gasRequirement, result.getOutput());
             if (result.isRefundGas()) {
                 frame.incrementRemainingGas(gasRequirement);
             }
-            finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);
         }
+        // We must always call tracePrecompileResult() to ensure the tracer is in a consistent
```
lukelee-sl, line 183: Do we not have the same issue with `doExecuteSystemContract`?

david-bakin-sl: I understand the comment at L177 that currently `tracePrecompileCall` is a noop, but even so, the way that `doExecutePrecompile` and `doExecuteSystemContract` are written the calls in them to `tracePrecompileCall` happen in a different order w.r.t. the check for sufficient gas, so if it was _not_ the case it was a noop the effects would be different (called in one case, not called in the other).

david-bakin-sl: A future refactor could move tracer actions - and possibly end-of-frame cleanup - into a separate object where they could be controlled under a try-with-resources and that might make this logic - in which conditional actions flow from the multi-return `start()` method to nested conditionals in `doExecute{Precompile,SystemContract}` down to `finishPrecompileExecution` easier to understand.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCallSuite.java
```diff
@@ -287,6 +287,23 @@ final HapiSpec repeatedCreate2FailsWithInterpretableActionSidecars() {
                 .then();
     }
 
+    @HapiTest
+    final HapiSpec insufficientGasToPrecompileFailsWithInterpretableActionSidecars() {
+        final var contract = "LowLevelCall";
+        final var payload = unhex(
+                "1d19ea4313c1943e9d66830b1e05f54895b9743810ad24d831b1186a32c15fe83000fdae8d8e064d5db6ef2b9bd63baa748da3ac6c2f5b4ea372a0f16f844797198e9393920d483a7260bfb731fb5d25f1aa493335a9e71297e485b7aef312c21800deef121f1e76426a00665e5c4479674322d4f75edadd46debd5cd992f6ed275dc4a288d1afb3cbb1ac09187524c7db36395df7be3b99e673b13a075a65ec1d9befcd05a5323e6da4d435f3b617cdb3af83285c2df711ef39c01571827f9d2e1aca507b1e6203bddd296bb9b5a2021850bf2fe51011efd7ea2ef559df305727c5b75c764a9eb1f38f05fb732c3a738ebd02b01eade2a3ff4f065eebb96cf70ef864561faf2e2d8ab64de6f2403cd4335860945d2259681d9fc47508786bae186a54489576bced52e7a986603846c314b59164d18591323a739752634f40482f6e185aecd100dde233463f21e06c2aa0d61c046bebb729b891f8d4660c6cbf2f1f2e789e7364c70ecb38087cc90810c3789ef30c0dd973f0f1a572a089f81c");
```
david-bakin-sl, line 294: Can we have a comment here that gives some kind of hint as to what this blob is?   Anything?

david-bakin-sl (commented):
Would really like to see a comment provided at `ContractCallSuite` L294 as mentioned below.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
tinker-michaelj: Fixed same issue there, thanks @lukelee-sl !

lukelee-sl (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCallSuite.java
tinker-michaelj: Added comment; it's a real-world payload to the `0x08` precompile from [this transaction](https://hashscan.io/testnet/transaction/1711397022.025030483)...although I don't think it really mattered in the end, was just trying to reproduce.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
tinker-michaelj: Agreed @david-bakin-sl , you can imagine this making a very nice application of a `try`-with-resources. 🤔

david-bakin-sl (approved):
LGTM (leaving the improvement w.r.t. `tracer.tracePrecompileCall` for the future).  (But please don't merge until Lee has a second shot at this too.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
```diff
@@ -55,6 +55,9 @@ private void validateStateChangeIds(@NonNull final TransactionSidecarRecord side
 
     private void validateActionIds(@NonNull final TransactionSidecarRecord sidecar) {
         final var actions = sidecar.getActions().getContractActionsList();
+        System.out.println("------");
```
lukelee-sl, line 58: do we want these in here still?

lukelee-sl (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
tinker-michaelj: Removed.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -196,14 +203,16 @@ private void doExecuteSystemContract(
         final var fullResult = systemContract.computeFully(frame.getInputData(), frame);
         final var gasRequirement = fullResult.gasRequirement();
         tracer.tracePrecompileCall(frame, gasRequirement, fullResult.output());
```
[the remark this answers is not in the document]
tinker-michaelj: Removed this no-op.

Neeharika-Sompalli (approved):

david-bakin-sl (approved):
LGTM

PR #12416, opened 2024-03-28
tests: Add negative test cases for token info system contracts

### LINKED ISSUE #12394: Implement negative test cases for TokenInfo, NFTTokenInfo and FungibleTokenInfo System Contracts

MrValioBg:
**Description**:

**Related issue(s)**:

Fixes #12394 

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenInfoHTSSuite.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenInfoHTSSuite.java
@@ -41,6 +41,7 @@ import static com.hedera.services.bdd.spec.transactions.token.CustomFeeSpecs.fra
 import static com.hedera.services.bdd.spec.transactions.token.CustomFeeSpecs.royaltyFeeWithFallback;
 import static com.hedera.services.bdd.spec.utilops.CustomSpecAssert.allRunFor;
 import static com.hedera.services.bdd.spec.utilops.UtilVerbs.childRecordsCheck;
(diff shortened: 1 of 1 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenInfoHTSSuite.java
```diff
@@ -749,6 +754,132 @@ final HapiSpec getInfoOnInvalidNonFungibleTokenFails() {
...
+                                .andAllChildRecords()
+                                .logged(),
+                        childRecordsCheck(
+                                NON_FUNGIBLE_TOKEN_INFO_TXN,
+                                CONTRACT_REVERT_EXECUTED,
+                                recordWith().status(INVALID_TOKEN_NFT_SERIAL_NUMBER)));
+    }
+
+    @HapiTest
+    final HapiSpec getInfoForFungibleTokenByNFTTokenAddressFails() {
```
MrValioBg, line 841: This test seems reasonable to fail with INVALID_TOKEN or similar, since we use NFT. But it seems like its behaving like getTokenInfo. 
We get SUCCESS by running it against both mono and mod.

MrValioBg: Understood, then maybe we could change the name to "..ddressWorks" and put a comment for the behaviour, or maybe we can remove the test itself. What do you think?

MrValioBg: Done, updated

tinker-michaelj (approved):
LGTM, ty @MrValioBg !

PR #12418, opened 2024-03-28
chore: Added extra-checks to prevent unchecked submits on mainnet and testnet

### LINKED ISSUE #12363: Check ledger-id before sending unchecked submit
Before sending an unchecked submit in `SubmissionManager.submit()`, we check if the profile is not `PROD`. As an additional security measure, we should also check if the configured ledger-id is not mainnet or testnet.

netopyr:
This PR adds extra checks to secure unchecked submits even more.

Fixes #12363

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/SubmissionManager.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/SubmissionManager.java
@@ -73,6 +73,7 @@ public class SubmissionManager {
     private static final String SPEEDOMETER_FORMAT = "%,13.2f";
     private static final Bytes MAIN_NET_LEDGER_ID = Bytes.fromHex("00");
     private static final Bytes TEST_NET_LEDGER_ID = Bytes.fromHex("01");
+    private static final Bytes PREVIEW_NET_LEDGER_ID = Bytes.fromHex("02");
 
     // FUTURE Consider adding a metric to keep track of the number of duplicate transactions submitted by users.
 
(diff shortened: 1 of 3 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/SubmissionManager.java
```diff
@@ -130,8 +133,12 @@ public void submit(@NonNull final TransactionBody txBody, @NonNull final Bytes t
...
             // We do NOT allow this call in production!
             // check profile dynamically, this way we allow profile overriding in Hapi tests
-            final var hederaConfig = configProvider.getConfiguration().getConfigData(HederaConfig.class);
-            if (hederaConfig.activeProfile() == Profile.PROD) {
+            final var configuration = configProvider.getConfiguration();
+            final var hederaConfig = configuration.getConfigData(HederaConfig.class);
+            final var ledgerConfig = configuration.getConfigData(LedgerConfig.class);
+            if (hederaConfig.activeProfile() == Profile.PROD
+                    || MAIN_NET_LEDGER_ID.equals(ledgerConfig.id())
+                    || TEST_NET_LEDGER_ID.equals(ledgerConfig.id())) {
```
Neeharika-Sompalli, line 141: We should not allow in previewnet also right ?

netopyr: Added a check for previewnet

Neeharika-Sompalli (dismissed):
LGTM ! Thanks @netopyr

tinker-michaelj (approved):
LGTM, thx @netopyr !

SimiHunjan (approved):

PR #12436, opened 2024-03-28
chore: Added extra-checks to prevent unchecked submits

### LINKED ISSUE #12363: Check ledger-id before sending unchecked submit

netopyr:
Cherry-pick of #12418 

Closes #12363


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12486, opened 2024-04-02
fix: Setup metrics for WritableKVStateStack

### LINKED ISSUE #12485: Fix utilization metrics
Utilization metrics do not show up in Grafana. This needs to be investigated and fixed.

netopyr:
This PR ensure metrics are also setup for `WritableKVStateStack`

Fixes #12485

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/disk/OnDiskWritableKvStateMetrics.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/disk/OnDiskWritableKvStateMetrics.java
@@ -60,7 +60,7 @@ public class OnDiskWritableKvStateMetrics {
                 .withFormat("%,d");
         metrics.getOrCreate(totalUtilizationConfig);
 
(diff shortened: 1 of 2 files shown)


Neeharika-Sompalli (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/disk/OnDiskWritableKvStateMetrics.java
```diff
@@ -60,7 +60,7 @@ public OnDiskWritableKvStateMetrics(
                 .withFormat("%,d");
         metrics.getOrCreate(totalUtilizationConfig);
 
-        final double relativeFactor = 100.0 / maxCapacity;
+        final double relativeFactor = 100.0 / Math.max(1.0, maxCapacity); // in some tests, maxCapacity is 0
```
netopyr, line 63: This is just a workaround for now. We need to fix the tests in the future.

Neeharika-Sompalli: Makes sense

Issue #12491, opened 2024-04-02
HederaSoftwareVersion breaks event hash backwards compatability


cody-littley:
Recent fixes to `HederaSoftwareVersion` inadvertently broke event hashing across migration boundaries.

PR #12494, opened 2024-04-02
fix: version backwards compatability

cody-littley:
related to #12491 

Must be cherry picked to all branches after 48


Neeharika-Sompalli (approved):
LGTM ! Thanks @cody-littley

tinker-michaelj (approved):
LGTM! Is there a requirement that the `SoftwareVersion` specifically always serialize itself in a form consistent with the deserialized version?

edward-swirldslabs (approved):

cody-littley:
@tinker-michaelj 

> LGTM! Is there a requirement that the SoftwareVersion specifically always serialize itself in a form consistent with the deserialized version?

Yes, this is a requirement (although not a well documented one). The requirement is derived from the requirement that events never change their serialized bytes after a migration. Since we include a software version inside each event, we end up having to also extend that requirement to software versions.

PR #12515, opened 2024-04-02
feat: address book serializes text hostnames

### LINKED ISSUE #12508: Platform address-book serializes hostname loaded from config.txt
The following config.txt can be input with hostnames, but the names are converted to IP addresses before they are added to the address book and the address book is serialized with IP addresses.    The addresses need to preserve their hostnames.  

This is critical to support SOLO with restarting containers in a k8s environment. 

```
swirld, 298
app, HederaNode.jar
address, 0, node0, node0, 1, network-node0-0.network-node0.node-cmd-e2e-pem.svc.cluster.local, 50111, network-node0.node-cmd-e2e-pem.svc.cluster.local, 50111, 0.0.3
address, 1, node1, node1, 1, network-node1-0.network-node1.node-cmd-e2e-pem.svc.cluster.local, 50111, network-node1.node-cmd-e2e-pem.svc.cluster.local, 50111, 0.0.4
address, 2, node2, node2, 1, network-node2-0.network-node2.node-cmd-e2e-pem.svc.cluster.local, 50111, network-node2.node-cmd-e2e-pem.svc.cluster.local, 50111, 0.0.5
address, 3, node3, node3, 1, network-node3-0.network-node3.node-cmd-e2e-pem.svc.cluster.local, 50111, network-node3.node-cmd-e2e-pem.svc.cluster.local, 50111, 0.0.6
nextNodeId, 4
```

edward-swirldslabs:
**Description**:
This PR ensures that we store and serialize text based hostnames when they are given in config.txt.    Roundtrip serialization and deserialization is verified by using the string "localhost" as a hostname for an address.  

**Related issue(s)**:

Fixes #12508

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
@@ -239,9 +239,9 @@ public class AddressBookUtils {
                 nickname,
                 selfname,
                 weight,
-                internalIp.getHostAddress(),
+                parts[5], // FQDN Support: use original string for internal address
                 internalPort,
-                externalIp.getHostAddress(),
+                parts[7], // FQDN Support: use original string for external address
                 externalPort,
                 memoToUse);
     }
--- a/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/system/address/AddressBookTests.java
+++ b/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/system/address/AddressBookTests.java
@@ -463,6 +463,12 @@ class AddressBookTests {
     void roundTripSerializeAndDeserializeCompatibleWithConfigTxt() throws ParseException {
         final RandomAddressBookGenerator generator = new RandomAddressBookGenerator(getRandomPrintSeed());
         final AddressBook addressBook = generator.build();
+        // FQDN Support: modify address in address book to have a text based host name.
+        addressBook.add(addressBook
+                .getAddress(addressBook.getNodeId(0))
+                .copySetHostnameInternal("localhost")
+                .copySetHostnameExternal("localhost"));
+
         // make one of the memo fields an empty string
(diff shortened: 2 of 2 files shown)


cody-littley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
```diff
@@ -239,9 +239,9 @@ public static Address parseAddressText(@NonNull final String addressText) throws
                 nickname,
                 selfname,
                 weight,
-                internalIp.getHostAddress(),
+                parts[5], // FQDN Support: use original string for internal address
```
alittley, line 242: Is there a reason that you stopped using `internalIp` and `externalIp`, but didn't remove their parsing from the method?

edward-swirldslabs: The parsing generates a network `UnknownHostException` if the text is not network resolvable.   I wanted to leave that unchanged.    Does the parsing really do anything?  I don't know.  It made sense at the time to make sure that the hostname was network resolvable.  Perhaps it doesn't make sense here and failure at the network connection attempt is more appropriate.    What do you think?

alittley: I think you're right, it makes sense to make sure it resolves. 

I think something like the following would be good, so that the intention is clear. Otherwise, someone is likely to remove the block in the future, thinking it's no longer necessary

```
        try {
            // make sure the IP address is valid, even though we don't use the result
            final InetAddress ignored = InetAddress.getByName(parts[5]);
        } catch (UnknownHostException e) {
            throw new ParseException("Cannot parse ip address from '" + parts[5] + ",", 5);
        }
```

lpetrovic05 (approved):

jeromy-cannon (dismissed):

jeromy-cannon (approved):

nathanklick (approved):

PR #12534, opened 2024-04-03
feat: move the JVM anchor into the wiring model

### LINKED ISSUE #12533: Move JVM anchor into wiring model
Move the JVM anchor into the platform wiring model. The wiring model should ultimately be responsible for all platform threads.

cody-littley:
closes #12533

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
@@ -137,21 +137,16 @@ public interface WiringModel extends Startable, Stoppable {
     OutputWire<Instant> buildHeartbeatWire(final double frequency);
 
     /**
-     * Lower a metaphorical anchor into the JVM that prevents the JVM from shutting down. This is useful if there are no
-     * other non-daemon threads running in the JVM. Before the JVM is permitted to exit, the anchor must be raised via
-     * {@link #raiseJvmAnchor()}.
-     * <p>
-     * Calling this method when the anchor is already lowered has no effect.
+     * Prevent the JVM from exiting even if there are no non-daemon threads outside the wiring model. Calling this
+     * method while the JVM exit is already prevented has no effect.
      */
-    void lowerJvmAnchor();
+    void preventJvmExit();
 
     /**
-     * Raise the metaphorical anchor that prevents the JVM from shutting down. This is automatically called when the
-     * model is stopped via {@link #stop()}.
-     * <p>
-     * Calling this method when the anchor is already raised has no effect.
+     * If the JVM is currently being prevented from exiting due to a call to {@link #preventJvmExit()}, then this method
+     * will allow the JVM to exit. Calling this method while the JVM exit is already permitted has no effect.
      */
-    void raiseJvmAnchor();
+    void permitJvmExit();
 
     /**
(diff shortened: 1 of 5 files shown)


alittley (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/WiringModel.java
```diff
@@ -136,6 +136,23 @@ String generateWiringDiagram(
...
     OutputWire<Instant> buildHeartbeatWire(final double frequency);
 
+    /**
+     * Lower a metaphorical anchor into the JVM that prevents the JVM from shutting down. This is useful if there are no
+     * other non-daemon threads running in the JVM. Before the JVM is permitted to exit, the anchor must be raised via
+     * {@link #raiseJvmAnchor()}.
+     * <p>
+     * Calling this method when the anchor is already lowered has no effect.
+     */
+    void lowerJvmAnchor();
```
edward-swirldslabs, line 146: Anchor lowering stops the motion of that which lowered the anchor.    I don't know that lowering an anchor is the right metaphor.  Inserting wrench into gears is a better metaphor.  What about `insertWrenchIntoJVMExitGears`?   Too cumbersome.   Perhaps a straightforward purposeful method name like `preventJVMExit()` and `allowJVMExit()`.

cody-littley: I've renamed these methods as you have suggested.

I did not rename the helper class JvmAnchor... not exactly sure what to call it. Since this is more of an internal detail, I think it's less important to rename. But if you have any good suggestions, I'm still open to the idea.

alittley (approved):

kfa-aguda (approved):
lgtm.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/JvmAnchor.java
```diff
@@ -33,16 +30,10 @@ public class JvmAnchor implements Startable, Stoppable {
...
-    public JvmAnchor(@NonNull final ThreadManager threadManager) {
-        thread = new ThreadConfiguration(threadManager)
-                .setComponent("platform")
-                .setThreadName("jvm-anchor")
-                .setRunnable(this::run)
-                .setDaemon(false)
-                .build();
+    public JvmAnchor() {
+        thread = new Thread(this::run, "<jvm-anchor>");
+        thread.setDaemon(false); // important
```
kfa-aguda, line 36: Actually, looks like we don't expect to ever have more than one anchor instance. If so, and considering the contract of StandardWiringModel#preventJvmExit(), it might be worth making this a singleton to prevent inadvertently having multiple anchors?

cody-littley: In a production environment we won't have more than one anchor. In TURTLE mode we may have multiple anchors, although having more than one is not strictly necessary.

The reason why I am extremely averse to using a true singleton model is that I believe static stateful variables to be inherently "evil". We are currently in the middle of converting all static singleton objects used by the platform into objects that are scoped to the platform, not to the JVM. To me, it's worth a little waste in TURTLE style tests to avoid having to use a static variable.

PR #12535, opened 2024-04-03
fix: add pureChecks to ingest workflow

netopyr:
Call pureChecks() in ingest workflow
Added some catch-all handling


Neeharika-Sompalli (dismissed):
LGTM ! Thanks @netopyr

iwsimon (approved):
LGTM. Thanks @netopyr

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12549, opened 2024-04-04
fix: Cherry-pick: add pureChecks to ingest workflow

### LINKED ISSUE #12532: Run pureChecks in ingest workflow
Some of the implementations of `TransactionHandler.calculateFees()` expect a valid transaction (e.g. all mandatory fields exist). We decided to run `pureChecks()` ahead of it. We may revert this decision later, but this would require that all implementations of `calculateFees()` can handle invalid input data.

netopyr:
Cherry-pick of #12535 

Closes #12532


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

tinker-michaelj (approved):
LGTM, ty @netopyr !

PR #12453, opened 2024-04-01
fix: tokenAssociate throws IndexOutOfBoundsException

JivkoKelchev:
**Description**:

This PR adds additional check to ensure that newly created token relation list is not empty. This way we prevent IndexOutOfBoundsException when passing no tokens param to the request body.

**Related issue(s)**:

Fixes #12440

**Notes for reviewer**:

This is a quick fix and it match mono behaviour, and return is SUCCES, but we may need to return a proper error status when tokens are empty.

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #12163, opened 2024-03-18
test: Add negative test cases to AssociationsPrecompileSuite

stoyanov-st:
**Description**:
Added  more HapiTests to verify the behaviour of negative associations scenarios.

**Related issue(s)**:

Fixes #12061 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #12550, opened 2024-04-04
fix: tokenAssociate throws IndexOutOfBoundsException (#12453)

### LINKED ISSUE #12440: `tokenAssociate` throws `IndexOutOfBoundsException`
### Description

`tokenAssociate` throws `IndexOutOfBoundsException` when passing no tokens param to the request body.

### Steps to reproduce

-

### Additional context

_No response_

### Hedera network

other

### Version

0.48

### Operating system

macOS

JivkoKelchev:
**Description**:

This PR adds an additional check to ensure that the newly created token relation list is not empty. This way, we prevent IndexOutOfBoundsException when passing no tokens param to the request body.

**Related issue(s)**:

Fixes #12440

**Notes for reviewer**:

This is a quick fix and it match mono behaviour, and return is SUCCES, but we may need to return a proper error status when tokens are empty.

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -173,19 +173,22 @@ protected void createAndLinkTokenRels(
...
-                .headTokenId(firstOfNewTokenRels.tokenId())
-                // and also update the account's total number of token associations
-                .numberAssociations(account.numberAssociations() + newTokenRels.size())
-                .build();
-
-        // Save the results
-        accountStore.put(updatedAcct);
-        newTokenRels.forEach(tokenRelStore::put);
+        // FUTURE - We may need to return a proper error status when tokens are empty
+        if (!newTokenRels.isEmpty()) {
```
netopyr, line 177: I think we have to include the call of `linkTokenRels()` (one line above). If `newTokenRels` is empty, the remove-call in line 220 fails, doesn't it?

thomas-swirlds-labs:
Do we not want to include the `associateTokensNegativeScenarios` test in release/0.48 as well ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
JivkoKelchev: Good catch. I was using an account without previous relations, so the remove was never executed.
I will create a new PR to develop and cherry-pick here. 
Thanks

JivkoKelchev:
> Do we not want to include the `associateTokensNegativeScenarios` test in release/0.48 as well ?

I got merge conflicts when I cherry-picked it, because of this test.
And since it was part of another [PR](https://github.com/hashgraph/hedera-services/pull/12163) that was not included in release/0.48, I decided to accept changes from 0.48 (remove the test)

netopyr (approved):
LGTM - thanks @JivkoKelchev

thomas-swirlds-labs (approved):
LGTM! tyvm @JivkoKelchev

tinker-michaelj (approved):
LGTM, ty @JivkoKelchev !

PR #12554, opened 2024-04-04
chore: refactor transactions

### LINKED ISSUE #12551: Refactor transactions
In order to prepare the move to protobuf payloads/transactions, some refactoring should be done on transactions:

- Remove SystemTransaction (we will no longer have this hierarchy when we move to protobufs)
- Move metadata to `ConsensusTransactionImpl` which will become the payload wrapper
- Remove signatures from transaction (currently only used by testing apps)

lpetrovic05:
closes #12551


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransaction.java
```diff
@@ -16,9 +16,16 @@
...
  * A transaction that has reached consensus.
  */
-public sealed interface ConsensusTransaction extends Transaction, ReachedConsensus permits ConsensusTransactionImpl {}
+public sealed interface ConsensusTransaction extends Transaction permits ConsensusTransactionImpl {
+    /**
+     * Returns the community's consensus timestamp for this item.
+     *
+     * @return the consensus timestamp
+     */
+    Instant getConsensusTimestamp();
```
alittley, line 30: missing nullity annotation (also on implementations)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransactionImpl.java
```diff
@@ -16,60 +16,51 @@
...
+     * Sets the consensus timestamp of this transaction
      *
-     * @param consensusOrder
-     * 		the consensus order
+     * @param consensusTimestamp
+     * 		the consensus timestamp
      */
-    public void setConsensusOrder(final long consensusOrder) {
-        this.consensusOrder = consensusOrder;
+    public void setConsensusTimestamp(final Instant consensusTimestamp) {
```
alittley, line 47: `consensusTimestamp` missing nullity annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransactionImpl.java
```diff
@@ -16,60 +16,51 @@
...
-     */
-    private long consensusOrder = NO_CONSENSUS;
-
     /**
      * The consensus timestamp of this transaction, or null if consensus has not yet been reached.
      * NOT serialized and not part of object equality or hash code
      */
     private Instant consensusTimestamp;
+    /** An optional metadata object set by the application */
+    private Object metadata;
```
anthony-swirldslabs, line 31: "Metadata" sounds generic, and so does the plain `Object` type. Above I've seen it's used for signatures only.

Should this field (and its getters) be named and typed accordingly? If not, then its javadoc could mention a few different use-cases to clarify why it's loosely typed.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransaction.java
lpetrovic05: This is a tricky one. The idea of this interface is that transactions are presented when they reach consensus through this interface. So technically this should be `@NonNull`. But the object that implements this interface also implements the interface of the pre-consensus transactions. Meaning the overriding method would be `@Nullable`. I say we leave this as is for the time being since its not new code, and maybe we can discuss and potentially get rid of one of these interfaces.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransactionImpl.java
lpetrovic05: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransactionImpl.java
lpetrovic05: This is old code, just moved. The interface declaring `getMetadata` and `setMetadata` explains what its used for. Our test apps use it for signatures, Hedera sets a different type.

jeromy-cannon (approved):

alittley (approved):

mxtartaglia-sl (approved):
Approved for base-team owned files

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/ConsensusTransactionImpl.java
anthony-swirldslabs: Got it. Makes sense. Thanks for the pointer.

anthony-swirldslabs (approved):

PR #12556, opened 2024-04-04
feat: EventSignatureValidator cleanup

### LINKED ISSUE #12555: Build EventSignatureValidator with component builder
Convert the event signature validator to the new component builder pattern.

cody-littley:
closes #12555


alittley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformComponentBuilder.java
```diff
@@ -220,4 +224,43 @@ public EventDeduplicator buildEventDeduplicator() {
...
     }
+
+    /**
+     * Provide an event signature validator in place of the platform's default event signature validator.
+     *
+     * @param eventSignatureValidator the event signature validator to use
+     * @return this builder
+     */
+    @NonNull
+    public PlatformComponentBuilder withEventSignatureValidator(
```
kfa-aguda, line 235: A few unused methods in this class, old and new. If they're really unused (vs indirectly via, say, reflection), any reason for adding them?

cody-littley: Although I agree that in general it's important to remove unused methods, these particular methods are intentionally not removed. The purpose of this class is to allow us to swap out component implementations at platform construction time. We currently don't swap anything out, but we will do so in the near future.

Issue #12565, opened 2024-04-04
ContractDelete without contractID throws unexpected NPE


netopyr:
Discovered during Java SDK integration testing: 

ContractDeleteIntegrationTest > Cannot delete contract when contract ID is not set
[Link to the test](https://github.com/hashgraph/hedera-sdk-java/blob/df866e90bb796a23678aa6ea85227bf4e52b905b/sdk/src/integrationTest/java/ContractDeleteIntegrationTest.java#L117)

SDK Logs:
java.lang.AssertionError:
Expecting throwable message:
 "Hedera transaction `0.0.1022@1712216222.748000000` failed pre-check with the status `FAIL_INVALID`"
to contain:
 "INVALID_CONTRACT_ID"
but did not.
   
hedera-services Logs:
ERROR 106  IngestWorkflowImpl - Possibly CATASTROPHIC failure while running the ingest workflow
java.lang.NullPointerException: Field contractID is null
        at java.base/java.util.Objects.requireNonNull(Objects.java:259) ~[?:?]
        at com.hedera.hapi.node.contract.ContractDeleteTransactionBody.contractIDOrThrow(ContractDeleteTransactionBody.java:153) ~[hapi-0.48.0-alpha.8.jar:?]
        at com.hedera.node.app.authorization.PrivilegesVerifier.hasPrivileges(PrivilegesVerifier.java:114) ~[?:?]
        at com.hedera.node.app.authorization.AuthorizerImpl.hasPrivilegedAuthorization(AuthorizerImpl.java:82) ~[?:?]
        at com.hedera.node.app.spi.authorization.Authorizer.hasWaivedFees(Authorizer.java:76) ~[app-spi-0.48.0-alpha.8.jar:?]
        at com.hedera.node.app.workflows.SolvencyPreCheck.checkSolvency(SolvencyPreCheck.java:140) ~[?:?]
        at com.hedera.node.app.workflows.SolvencyPreCheck.checkSolvency(SolvencyPreCheck.java:126) ~[?:?]
        at com.hedera.node.app.workflows.ingest.IngestChecker.runAllChecks(IngestChecker.java:243) ~[?:?]
        at com.hedera.node.app.workflows.ingest.IngestWorkflowImpl.submitTransaction(IngestWorkflowImpl.java:92) ~[?:?]
        ...

netopyr:
Another example:

ContractUpdateIntegrationTest > Cannot update contract when contract ID is not set
[...]

PR #12574, opened 2024-04-04
fix: Add pure check to ContractDeleteHandler

netopyr:
Fixes https://github.com/hashgraph/hedera-services/issues/12565
Add `pureChecks()` to `ContractDeleteHandler` to ensure mandatory `contractID` is provided.

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractDeleteHandler.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractDeleteHandler.java
@@ -36,6 +36,7 @@ import com.hedera.hapi.node.base.HederaFunctionality;
 import com.hedera.hapi.node.base.SubType;
 import com.hedera.hapi.node.contract.ContractDeleteTransactionBody;
 import com.hedera.hapi.node.state.token.Account;
+import com.hedera.hapi.node.transaction.TransactionBody;
 import com.hedera.node.app.hapi.utils.fee.SmartContractFeeBuilder;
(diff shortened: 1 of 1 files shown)


Neeharika-Sompalli (approved):

tinker-michaelj (approved):
LGTM, ty @netopyr !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractDeleteHandler.java
```diff
@@ -64,6 +65,12 @@ public ContractDeleteHandler() {
         // Exists for injection
     }
 
+    @Override
+    public void pureChecks(@NonNull TransactionBody txn) throws PreCheckException {
+        final var op = txn.contractDeleteInstanceOrThrow();
+        mustExist(op.contractID(), INVALID_CONTRACT_ID);
```
david-bakin-sl, line 71: I note that at `TokenDeleteHandler.pureChecks` it does this check by calling `PreCheckException.validateTruePreCheck` and at `CryptoDeleteHandler` it directly throws a `PreCheckException` (as do several other `pureChecks` - someday this could be cleaned up to have _one_ way to do it (which would be more likely to be correct or at least easier to use as a template for future work).

david-bakin-sl (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractDeleteHandler.java
netopyr: I agree. We need to go through all handlers and standardize such patterns.

PR #12578, opened 2024-04-04
fix: make pureChecks required in all handlers

### LINKED ISSUE #12566: TokenAssociate without accountID throws unexpected NPE
Discovered during Java SDK integration testing:

TokenAssociateIntegrationTest > Cannot associate account with tokens when account ID
[Link to the test](https://github.com/hashgraph/hedera-sdk-java/blob/df866e90bb796a23678aa6ea85227bf4e52b905b/sdk/src/integrationTest/java/TokenAssociateIntegrationTest.java#L89)

SDK Logs:
is not set FAILED
  java.lang.AssertionError:
  Expecting throwable message:
    "Hedera transaction `0.0.1183@1712216328.038000000` failed pre-check with the status `FAIL_INVALID`"
  to contain:
    "INVALID_ACCOUNT_ID"
  but did not.
  
hedera-services Logs:
ERROR 106  IngestWorkflowImpl - Possibly CATASTROPHIC failure while running the ingest workflow
java.lang.NullPointerException: Field account is null
        at java.base/java.util.Objects.requireNonNull(Objects.java:259) ~[?:?]
        at com.hedera.hapi.node.token.TokenAssociateTransactionBody.accountOrThrow(TokenAssociateTransactionBody.java:145) ~[hapi-0.48.0-alpha.8.jar:?]
        at com.hedera.node.app.service.token.impl.handlers.TokenAssociateToAccountHandler.calculateFees(TokenAssociateToAccountHandler.java:188) ~[app-service-token-impl-0.48.0-alpha.8.jar:?]
        at com.hedera.node.app.workflows.dispatcher.TransactionDispatcher.dispatchComputeFees(TransactionDispatcher.java:128) ~[?:?]
        at com.hedera.node.app.workflows.ingest.IngestChecker.runAllChecks(IngestChecker.java:242) ~[?:?]
        at com.hedera.node.app.workflows.ingest.IngestWorkflowImpl.submitTransaction(IngestWorkflowImpl.java:92) ~[?:?]
        at com.hedera.node.app.grpc.impl.TransactionMethod.handle(TransactionMethod.java:56) ~[?:?]
        at com.hedera.node.app.grpc.impl.MethodBase.invoke(MethodBase.java:131) ~[?:?]
        at com.hedera.node.app.grpc.impl.netty.GrpcServiceBuilder$ListenerImpl.onMessage(GrpcServiceBuilder.java:252) ~[?:?]
        at com.hedera.node.app.grpc.impl.netty.GrpcServiceBuilder$ListenerImpl.onMessage(GrpcServiceBuilder.java:220) ~[?:?]
        ...

### LINKED ISSUE #12644: Validate pureChecks in smart contracts service
The integration tests of the SDK found two cases where invalid transactions led to unexpected Exceptions. We should make sure that each handler has a `pureChecks()` method (even if it is empty initially) and review the pure checks for all handlers of the smart contracts service.

kimbor:
**Description**:
TransactionHandler had a default empty implementation of pureChecks, which meant that all inheriting classes could optionally implement pureChecks or accept the empty default. This PR removes the default empty implementation and adds overrides for all implementing classes which previously lacked one.

**Related issue(s)**:
Fixes #12644

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
@@ -17,8 +17,10 @@
 package com.hedera.node.app.service.contract.impl.handlers;
 
 import static com.hedera.hapi.node.base.HederaFunctionality.CONTRACT_CALL;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CONTRACT_ID;
 import static com.hedera.node.app.service.contract.impl.utils.ConversionUtils.throwIfUnsuccessful;
 import static com.hedera.node.app.service.mono.pbj.PbjConverter.fromPbj;
+import static com.hedera.node.app.spi.validation.Validations.mustExist;
 import static java.util.Objects.requireNonNull;
 
 import com.hedera.hapi.node.base.HederaFunctionality;
@@ -75,8 +77,9 @@ public class ContractCallHandler implements TransactionHandler {
     }
 
     @Override
-    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
-        // nothing to do
+    public void pureChecks(@NonNull TransactionBody txn) throws PreCheckException {
+        final var op = txn.contractCallOrThrow();
+        mustExist(op.contractID(), INVALID_CONTRACT_ID);
     }
 
     @NonNull
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
@@ -23,6 +23,7 @@ import static java.util.Objects.requireNonNull;
 
 import com.hedera.hapi.node.base.HederaFunctionality;
 import com.hedera.hapi.node.base.SubType;
+import com.hedera.hapi.node.transaction.TransactionBody;
 import com.hedera.node.app.hapi.utils.fee.SmartContractFeeBuilder;
 import com.hedera.node.app.service.contract.impl.exec.CallOutcome.ExternalizeAbortResult;
 import com.hedera.node.app.service.contract.impl.exec.TransactionComponent;
@@ -32,6 +33,7 @@ import com.hedera.node.app.spi.fees.FeeContext;
 import com.hedera.node.app.spi.fees.Fees;
 import com.hedera.node.app.spi.workflows.HandleContext;
 import com.hedera.node.app.spi.workflows.HandleException;
+import com.hedera.node.app.spi.workflows.PreCheckException;
 import com.hedera.node.app.spi.workflows.PreHandleContext;
 import com.hedera.node.app.spi.workflows.TransactionHandler;
 import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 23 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java
```diff
@@ -98,6 +99,11 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         }
     }
 
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        // nothing to do
```
netopyr, line 104: Here we should check if the `contractID` is set.

kimbor: In preHandle (line 88) we call `op.contractIDOrElse(ContractID.DEFAULT)`. Is it correct to use a default contract ID if no contract ID is set?

netopyr (approved):
LGTM - thanks @kimbor

tinker-michaelj (approved):
LGTM, it seems quite sensible to have a hook where we can detect guaranteed failures at ingest.

But it's likely that we have some `calaculateFees()` implementations that should be more robust and work around missing fields with e.g. `contractIdOrElse(ContractID.DEFAULT)` patterns.

povolev15 (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCreateHandler.java
```diff
@@ -70,6 +71,11 @@ public void handle(@NonNull final HandleContext context) throws HandleException
         throwIfUnsuccessful(outcome.status());
     }
 
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        // nothing to do
```
lukelee-sl, line 76: Question - why are we doing nothing here and in the `EthereumTransactionHandler` below instead of checking the op is the correct type?  I understand that the contractID may not exist.

kimbor: The handlers are dispatched based on the op type. In order for code in this class to be called, op must already be of type ContractCreate

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCallHandler.java
```diff
@@ -72,6 +76,12 @@ public void preHandle(@NonNull final PreHandleContext context) {
         // No non-payer signatures to verify
     }
 
+    @Override
+    public void pureChecks(@NonNull TransactionBody txn) throws PreCheckException {
+        final var op = txn.contractCallOrThrow();
+        mustExist(op.contractID(), INVALID_CONTRACT_ID);
```
Neeharika-Sompalli, line 82: Not a blocker for this PR, but we should be consistent in the checks done in this method. We are validating `ContractID` exists in smart-contract service. But not doing any checks in Consensus service.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractCreateHandler.java
lukelee-sl: Makes sense but it seems we are checking the type in some of the handlers.

lukelee-sl (approved):
LGTM

JeffreyDallas (approved):

PR #12581, opened 2024-04-04
fix: unit test and add the right response code

### LINKED ISSUE #12571: Updating account without accountID returns INVALID_ACCOUNT_ID
Updating an account without accountID returns `INVALID_ACCOUNT_ID`, should return `ACCOUNT_ID_DOES_NOT_EXIST`.

Discovered during Java SDK integration testing:

[Link to the test](https://github.com/hashgraph/hedera-sdk-java/blob/df866e90bb796a23678aa6ea85227bf4e52b905b/sdk/src/integrationTest/java/AccountUpdateIntegrationTest.java#L71)

AccountUpdateIntegrationTest > Cannot update account when account ID is not set
   java.lang.AssertionError:
   Expecting actual throwable to be an instance of:
     com.hedera.hashgraph.sdk.ReceiptStatusException
   but was:
     com.hedera.hashgraph.sdk.PrecheckStatusException: Hedera transaction `0.0.1022@1712216187.819000000` failed pre-check with the status `INVALID_ACCOUNT_ID`

povolev15:
This fix supposed to provide the right response code when the account ID is not available during the crypto update

**Related issue(s)**:

Fixes #12571


netopyr (approved):
LGTM - thanks @povolev15

PR #12591, opened 2024-04-05
fix: Return REQUIRED_KEY in CryptoCreate if no alias is used

### LINKED ISSUE #12563: Trying to create an account without a key returns INVALID_ALIAS_KEY
When trying to create an account without a key, we return `INVALID_ALIAS_KEY` which is confusing. We should return `KEY_REQUIRED`.

Discovered while running Java SDK integration tests:

>>> [Link to the test](https://github.com/hashgraph/hedera-sdk-java/blob/df866e90bb796a23678aa6ea85227bf4e52b905b/sdk/src/integrationTest/java/AccountCreateIntegrationTest.java#L71)

AccountCreateIntegrationTest > Cannot create account with no key
   java.lang.AssertionError:
   Expecting throwable message:
     "Hedera transaction `0.0.1022@1712216157.828000000` failed pre-check with the status `INVALID_ALIAS_KEY`"
   to contain:
     "KEY_REQUIRED"
   but did not.

netopyr:
This PR changes the error handling of CryptoCreate slightly. When no alias is provided, we want to return `REQUIRED_KEY` instead of `INVALID_ALIAS_KEY`.

Fixes #12563


thomas-swirlds-labs (approved):
LGTM!

PR #12593, opened 2024-04-05
test: add negative tests for transferNFT, transferNFTs, transferToken, transferTokens

### LINKED ISSUE #12482: Implement negative test cases for transferTokens and transferNFT

MrValioBg:
**Description**:
Add negative tests for transferNFT, transferNFTs, transferToken, transferTokens.

**Related issue(s)**:

Fixes #12482 

**Notes for reviewer**:
Check comments on the PR

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
```diff
@@ -224,6 +229,26 @@ public TransactionBody decodeHrcTransferNftFrom(
...
+        if (Arrays.equals(attempt.selector(), ClassicTransfersTranslator.TRANSFER_TOKENS.selector())) {
+            final var call = ClassicTransfersTranslator.TRANSFER_TOKENS.decodeCall(attempt.inputBytes());
+            final Address[] accountIds = call.get(1);
+            if (accountIds.length == 0) {
+                return ResponseCodeEnum.EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS;
+            }
+            final var tokenStore = attempt.enhancement().nativeOperations().readableTokenStore();
+            final var token = tokenStore.get(ConversionUtils.asTokenId(call.get(0)));
+            if (token == null) {
+                return ResponseCodeEnum.SPENDER_DOES_NOT_HAVE_ALLOWANCE;
```
MrValioBg, line 242: The "SPENDER_DOES_NOT_HAVE_ALLOWANCE" obviously makes no sense, but this is the current mono behaviour for `transferTokens`.
Do we want to still do the matching for such cases at this point, where we are close to modularized codebase release @tinker-michaelj

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersDecoder.java
```diff
@@ -155,7 +155,8 @@ public TransactionBody decodeTransferNfts(
         final Address[] from = call.get(1);
         final Address[] to = call.get(2);
         final long[] serialNo = call.get(3);
-        if (from.length != to.length || from.length != serialNo.length) {
+
+        if ((from.length != to.length || from.length != serialNo.length) && from.length != 0) {
```
tinker-michaelj, line 159: Why do we need this `from.length != 0` check? Shouldn't we throw an exception if the `from` array is empty and the `to` array isn't?

MrValioBg: This is in order to get the EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS status in a child record, as it would never reach the logic in the handler for it, if we throw the exception here.

In Mono, we have a child record with EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS for this case. 
Previously, before this check in modularization, we had the check in the handler, but since we throw an error here, we never dispatch the child and never receive the error.

MrValioBg: Additional note:
The problem is that the [validation check](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java#L123) we have would never be reached if we throw the exception here. Mentioning Mono behavior is just for comparison. Since the difference here is not only the status (if it were, we might address it if differential testing prompts us), we may consider fixing it.

tinker-michaelj: I still don't understand. 😞 Are you saying we have a EET that:
 1. Uses a `from.length > 0`; and, 
 2. Uses a `to.length == 0`; and,
 3. Expects a child record with status `EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS`?

MrValioBg: This PR is adding such a EET.

david-bakin-sl: > Additional note: This [validation check](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java#L123) would never be reached if we throw the exception here.

BTW, the comment at that line seems wrong for the check as given:

> `// Verify that one and only one of the two types of transfers (fungible or non-fungible) is present`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
```diff
@@ -89,7 +110,12 @@ public List<HapiSpec> getSpecsInSuite() {
...
-        return List.of(nonZeroTransfersFail());
+        return List.of(
+                nonZeroTransfersFail(),
+                transferTokensNegativeCasesFail(),
+                transferTokenNegativeCasesFail(),
+                transferNFTsNegativeCasesFail(),
+                transferNFTNegativeCasesFail());
     }
 
     List<HapiSpec> positiveSpecs() {
```
david-bakin-sl, line 121: How can it be that there are _no_ positive tests?  Not a single happy path?  Is it covered in some other suite? (Not suggesting that _this_ PR change for that, just wondering if we need an issue for it.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
```diff
@@ -159,6 +185,547 @@ final HapiSpec nonZeroTransfersFail() {
                                                 htsPrecompileResult().withStatus(TRANSFERS_NOT_ZERO_SUM_FOR_TOKEN)))));
     }
 
+    @HapiTest
+    final HapiSpec transferTokensNegativeCasesFail() {
```
david-bakin-sl, line 189: This test and the next need better names that talk about what's being tested.  For example this test has something to do with the various arrays here (3 of them) not being of valid _length_ or _presence_.  The next test has something to do with invalid addresses and/or balance.

And same for the following two NFT cases.

david-bakin-sl (approved):
Suggest renaming these tests so they become self-documenting.  Also a couple of questions.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
MrValioBg: Hey!
Seems like we test the `transferNFT, transferNFTs, transferToken and transferTokens` as positive cases in:

`ContractKeysStillWorkAsExpectedSuite.approvalFallbacksRequiredWithoutTopLevelSigAccess`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
MrValioBg: I've improved the test's names to describe better the behaviour they are testing.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
david-bakin-sl: We need an index or something to the suites and the tests.  It's hard to decide where to _put_ a test, and it's hard to decide where to _find_ a test, unless you know all the suites ....

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ContractHTSSuite.java
david-bakin-sl: Yes, better now, thank you.

david-bakin-sl (approved):
LGTM

PR #12595, opened 2024-04-05
fix: Return REQUIRED_KEY in CryptoCreate if no alias is used

netopyr:
This PR changes the error handling of CryptoCreate slightly. When no alias is provided, we want to return REQUIRED_KEY instead of INVALID_ALIAS_KEY.

Fixes https://github.com/hashgraph/hedera-services/issues/12563


thomas-swirlds-labs (approved):
LGTM @netopyr !

PR #12622, opened 2024-04-05
perf: reorganize metrics initialization

netopyr:
This PR introduces the StoreMetricsService, which keeps only a single copy of each store metric and makes it available.


tinker-michaelj (dismissed):
LGTM, ty @netopyr !

tinker-michaelj (approved):

tinker-michaelj (approved):

hendrikebbers (approved):

PR #12635, opened 2024-04-08
feat: platform writes diagnostic marker files.

### LINKED ISSUE #11930: Write a marker file to disk in certain error scenarios
There are come error scenarios (some fatal, some not) for which certain information needs to be gathered to aid engineers in debugging the problem. It would be nice to eventually write a diagnostics directory on disk and put all relevant information there (like copying pertinent PCES files, consensus data dumps, etc) which could then be downloaded. But as a first step, the platform should write a marker file to disk when any of these scenarios is encountered so that if the problem occurs in a test environment, the test framework can go download the appropriate files. We do not want to download and store these files all the time because it is cost prohibitive.

Task: Make the platform write a marker file to disk in a location visible to JRS (i.e. `data/saved`) when any of the following scenarios occurs:

- coin round
- a round with no judges
- no supermajority of judges
- non-catastrophic ISS
- catastrophic ISS

edward-swirldslabs:
**Description**:
This PR adds the writing of marker files to a configured directory in the event of the following circumstances: 
* IssDetector
   * self-iss
   * catastrophic-iss
   * other-iss
* ConsensusImpl
   * Coin Round
   * No Super Majority in Voting 

The capability is turned off by default until a directory is specified as a location to write the files.   The most likely directory would be `data/saved/marker_files`

**Related issue(s)**:

Fixes #11930  

**Notes for reviewer**:

This PR introduces a base unit test class to inject the JUNIT temp directory.   Ideally we would turn on marker file writing in all unit tests and have all unit tests validate that no marker file was written by default.  Doing this really requires the creation of a FileSystemManager and a battery of default assumption checks as part of a testing framework.  These things are beyond the scope of this PR. 

If there is a better way to do what is in this PR, that feedback is very welcome. 

**Checklist**


mxtartaglia-sl (approved):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+     *     <li>The temp directory is added to the config for marker files</li>
+     *     <li>The platform context builder modifications are applied</li>
+     * </ol>
+     * Any configuration set by the platform context builder modifying method overrides the configuration created by by the config modifier. Best practice is to set configuration through the config modifier and all other platform context variables through the platform contect modifier.
+     *
+     * @param platformContextModifier the function to modify the platform context builder
+     * @param configModifier          the function to modify the test config builder
+     * @return the platform context
+     */
+    protected PlatformContext createPlatformContext(
```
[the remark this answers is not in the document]
edward-swirldslabs: Doing the extension allows for less code to be written and any changes to be applied to all unit tests at once.   Without the extension, in this PR,  we would need to add the annotation and declaration of the temp directory to each unit test.   That would be a small change here. 

As we move towards a more robust testing framework I expect that there will be more state to manage per test and per test class that might warrant using an extension of a base class that does the book keeping.    I can change it if there is consensus.  This PR will necessarily introduce tech debt that needs to be cleaned up one we have Turtle Lite and a `FileSystemManager`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private final File markerFileDirectory;
+
+    /**
+     * Creates a new {@link MarkerFileWriter} with the given {@link PlatformContext}.  If the marker file writer is
+     * enabled, files are written to the configured directory.
+     *
+     * @param platformContext the platform context containing configuration.
+     */
+    public MarkerFileWriter(@NonNull final PlatformContext platformContext) {
```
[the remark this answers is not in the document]
edward-swirldslabs: There's multiple aesthetics at play.  I find the uniformity of passing in a `PlatformContext` into constructors to be very clean and the asymmetry of sometimes passing in the `PlatformContext` and sometimes passing in specific pieces or configuration to be less clean.  

Since the `MarkerFileWriter` is constructed in multiple locations, `IssDetector` and `ConsensusImpl` for now, but there could be more locations in the future, there would be a multiplicity of locations where the file location is extracted from the `PlatformContext`.  Should we in the future change where the config value is located, then each location that extracts the directory from the `PlatformContext` would need to be modified.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Writes marker files with the given filename to disk in the configured directory.  If there is no configured
+ * directory, no marker files are written.
+ */
+public class MarkerFileWriter {
+
+    /**
+     * The logger for this class.
+     */
+    private static final Logger LOG = LogManager.getLogger(MarkerFileWriter.class);
```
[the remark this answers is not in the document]
edward-swirldslabs: while there is not a `LOG` when I do an IDE search on `final Logger`, there is a `LOGGER` used in a couple places.  `log` and `logger` seem to occur equally.   I will change it to `logger`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * The directory where the marker files are written.  If null, no marker files are written.
+     */
+    private final File markerFileDirectory;
```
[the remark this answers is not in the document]
edward-swirldslabs: I'm switching everything to Path api.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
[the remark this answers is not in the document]
edward-swirldslabs: just added a rate limited logger which uses time from the platform context.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Creates a platform context for the tests with the given builder modifications. Modifications are applied in the
+     * following order:
+     * <ol>
+     *     <li>The config modifications are applied</li>
+     *     <li>The temp directory is added to the config for marker files</li>
+     *     <li>The platform context builder modifications are applied</li>
+     * </ol>
+     * Any configuration set by the platform context builder modifying method overrides the configuration created by by the config modifier. Best practice is to set configuration through the config modifier and all other platform context variables through the platform contect modifier.
```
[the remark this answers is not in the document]
edward-swirldslabs: not sure why auto format didn't wrap the line.  Now manually wrapped.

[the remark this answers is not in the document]
edward-swirldslabs: fixed

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return platformContextBuilder.build();
+    }
+
+    /**
+     * Gets the temporary directory provided by JUnit
+     *
+     * @return the temporary directory
+     */
+    Path getTempDir() {
```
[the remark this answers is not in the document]
edward-swirldslabs: added `@NonNull` and check to ensure it is non-null.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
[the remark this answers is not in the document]
edward-swirldslabs: removed duplicate by

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Returns the directory where the marker files are written.  May be null.
+     *
+     * @return the directory where the marker files are written.  This may be null.
+     */
+    public Path getMarkerFileDirectory() {
+        if (markerFileDirectory == null) {
+            return null;
```
kfa-aguda, line 109: is it worth logging a message to this effect? "Skipping Marker file, directory not set" or something.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -42,6 +42,7 @@
 import com.swirlds.platform.internal.EventImpl;
 import com.swirlds.platform.metrics.ConsensusMetrics;
 import com.swirlds.platform.system.address.AddressBook;
+import com.swirlds.platform.util.MarkerFileWriter;
```
cody-littley, line 45: Adding this as a comment so that you have to press resolve before merging. Don't forget to set the milestone prior to merging (which milestone will depend on when it merges, and when the 0.49 branch is cut).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -141,6 +142,8 @@
  */
 public class ConsensusImpl extends ThreadSafeConsensusInfo implements Consensus {
 
+    public static final String COIN_ROUND_MARKER_FILE = "consensus-coin-round";
+    public static final String NO_SUPER_MAJORITY_MARKER_FILE = "consensus-no-super-majority";
```
cody-littley, line 146: Do we expect for whatever utility is watching for these marker files to do things differently depending on which marker file is found? If "no", then would it make more sense to instead use the same marker file type for each mode of failure?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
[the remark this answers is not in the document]
cody-littley: There are a few deviant places. A while ago there was a discussion in the coding style channel and Richard settled on `logger` as the pattern to use.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
```diff
@@ -76,4 +76,19 @@ public Path getAppsDirPath() {
     public Path getLogPath() {
         return rethrowIO(() -> getAbsolutePath(logPath));
     }
+
+    /**
+     * path to the directory where marker files are written (which might not exist).  Null is returned when the string is empty or set to "/dev/null".
+     *
+     * @return the absolute path to the directory where marker files are written or null if the path is empty or "/dev/null".
+     */
+    public Path getMarkerFilesDir() {
```
cody-littley, line 85:
```suggestion
    @NonNull
    public Path getMarkerFilesDir() {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
```diff
@@ -25,21 +25,21 @@
...
+ * @param markerFilesDir  path to the directory where marker files are written.
  */
 @ConfigData("paths")
 public record PathsConfig(
         @ConfigProperty(defaultValue = ".") String settingsUsedDir,
         @ConfigProperty(defaultValue = "data/keys") String keysDirPath,
         @ConfigProperty(defaultValue = "data/apps") String appsDirPath,
-        @ConfigProperty(defaultValue = "log4j2.xml") String logPath) {
+        @ConfigProperty(defaultValue = "log4j2.xml") String logPath,
+        @ConfigProperty(defaultValue = "none") String markerFilesDir) {
```
cody-littley, line 42: Would it make sense to set this to have a default value of `/dev/null` and to drop support for the the `none` keyword?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
cody-littley: We might want to be careful with logging. Or at the very least we'd want to rate limit the log. Some of these marker files have the potential to be written many times during error conditions, and an error followed by a gigabyte of log spam is no fun to debug. 🙃

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,113 @@
...
+            throw new RuntimeException("Failed to create marker file: " + markerFile, e);
+        }
+    }
+
+    /**
+     * Returns the directory where the marker files are written.  May be null.
+     *
+     * @return the directory where the marker files are written.  This may be null.
+     */
+    public Path getMarkerFileDirectory() {
```
cody-littley, line 107:
```suggestion
    @NonNull
    public Path getMarkerFileDirectory() {
```

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.nio.file.Path;
+import java.util.function.Function;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Platform level unit test base class for common setup and teardown.
+ */
+public abstract class BasePlatformUnitTests {
```
cody-littley, line 34: Personal preference. Happy to workshop naming via slack if you'd like.

```suggestion
public abstract class PlatformTest {
```

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Temporary directory provided by JUnit
+     */
+    @TempDir
+    protected Path tempDir;
+
+    /**
+     * Creates a default platform context for the tests
+     */
+    protected PlatformContext createDefaultPlatformContext() {
```
cody-littley, line 47:
```suggestion
    @NonNull
    protected PlatformContext createDefaultPlatformContext() {
```

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
cody-littley, line 65:
```suggestion
    @NonNull
    protected PlatformContext createPlatformContext(
```

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+     * </ol>
+     * Any configuration set by the platform context builder modifying method overrides the configuration created by by the config modifier. Best practice is to set configuration through the config modifier and all other platform context variables through the platform contect modifier.
+     *
+     * @param platformContextModifier the function to modify the platform context builder
+     * @param configModifier          the function to modify the test config builder
+     * @return the platform context
+     */
+    protected PlatformContext createPlatformContext(
+            @Nullable final Function<TestPlatformContextBuilder, TestPlatformContextBuilder> platformContextModifier,
+            @Nullable final Function<TestConfigBuilder, TestConfigBuilder> configModifier) {
```
cody-littley, line 67: As a general pattern, passing in configuration via lambdas is something I dislike.

Here's an idea. 

- Extend `TestPlatformContextBuilder` with a method `withRootDirectory(@NonNull Path rootDirectory)`
    - if set, then call into the configuration builder to update directories, as needed
    - eventually when we have a file system manager (hopefully not too far in the future), this will configure the file system manager
- Remove `TestPlatformContextBuilder.withConfiguration(Configuration)` and replace it with `withConfigurationBuilder(ConfigurationBuilder)`. Should be a straight forward refactor, but may touch many files.
- Remove the `createPlatformContext()` method from this class and replace it with `createPlatformContextBuilder()`. Within that method, call `withRootDirectory()` using the test directory.

As a part of this change, I'd strongly recommend reaching out to @hendrikebbers. I know he has put a lot of thought into how is the best way to set up unit test infrastructure. At the end of the day, hashgraph team owns hashgraph unit tests, but it's possible Hendrick may have a few tricks up his sleeves that will make our lives easier.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
cody-littley, line 34: Does junit delete the temporary directory in between tests within the same file? If not, I suggest adding an `@AfterEach` method that ensures that `tempDir` is deleted.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+     * checks for the presence of the marker file.
+     *
+     * @param markerFileName the name of the marker file
+     * @param exists         true if the marker file should exist, false otherwise
+     */
+    protected void assertMarkerFile(@NonNull final String markerFileName, final boolean exists) {
+        assertEquals(
+                exists,
+                getMarkerFileDirectory().resolve(markerFileName).toFile().exists());
+    }
```
cody-littley, line 114: If this class is intended to be a class that all of our unit tests inherit from, I'd suggest we refrain from adding methods specific to individual features. 99.99% of tests won't need to interact with marker files.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
```diff
@@ -0,0 +1,115 @@
...
+    Path getTempDir() {
+        return tempDir;
+    }
+
+    /**
+     * Gets the marker file directory
+     *
+     * @return the marker file directory
+     */
+    protected Path getMarkerFileDirectory() {
```
cody-littley, line 100:
```suggestion
    @NonNull
    protected Path getMarkerFileDirectory() {
```

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/MarkerFileWriterTests.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    void testWriteMarkerFile() {
+        final PlatformContext platformContext = TestPlatformContextBuilder.create()
+                .withConfiguration(new TestConfigBuilder()
+                        .withValue("paths.markerFilesDir", testDirectoryPath.toString())
+                        .getOrCreateConfig())
+                .build();
+
+        // create a new MarkerFileWriter
+        MarkerFileWriter markerFileWriter = new MarkerFileWriter(platformContext);
```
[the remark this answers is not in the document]
edward-swirldslabs: making it final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: good call, I will log an info message.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -477,6 +485,8 @@ private void voteInAllElections(@NonNull final EventImpl votingWitness) {
                     // no need to vote anymore until we create elections for the next round
                     return;
                 }
+            } else {
+                markerFileWriter.writeMarkerFile(NO_SUPER_MAJORITY_MARKER_FILE);
```
alittley, line 489: Leaving a comment here, to consider whether this is the correct place for this to be written

Looking at this method, the `false` case of this block doesn't *seem* to me to be something that ought to constitute an error. But I'm not familiar enough with this code to say one way or another. Maybe @lpetrovic05  can weigh in

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,133 @@
...
+     * Writes a marker file with the given filename to the configured directory.
+     *
+     * @param filename the name of the marker file to write.
+     */
+    public void writeMarkerFile(@NonNull final String filename) {
+        if (markerFileDirectory == null) {
+            // Configuration did not set a marker file directory.  No need to write marker file.
+            return;
+        }
+        Path markerFile = markerFileDirectory.resolve(filename);
```
alittley, line 109: May be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
alittley: @edward-swirldslabs I think having a static atomic boolean accessed by the constructor is more complexity than it's worth.

Only a couple of these objects are constructed, at boot, so it's not like we're contending with infinite log spam.

IMO I'd not log this at all, but if we *do* want a `null` input parameter to be logged, having 2 of that log message instead of just 1 doesn't make much of a difference

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,133 @@
...
+                Files.createDirectories(markerFileDirectoryPath);
+                directory = markerFileDirectoryPath;
+            } catch (final IOException e) {
+                logger.error(
+                        LogMarker.EXCEPTION.getMarker(),
+                        "Failed to create marker file directory: {}",
+                        markerFileDirectoryPath,
+                        e);
+            }
+            if (!Files.isDirectory(markerFileDirectoryPath)) {
```
alittley, line 82: A preexisting file will cause `createDirectories` above to fail with an IO exception.

With this in mind, I think the `isDirectory` check here is redundant, since it will only be triggered in cases where `createDirectories` *also* failed (and the message would be misleading in the case where directory creation failed for a reason other than a preexisting file)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
lpetrovic05: I think it should always be the same file, but maybe the contents can change depending on the issue

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
lpetrovic05: yeah, this is the wrong place, this will actually happen all the time until we decide a witness

cody-littley (dismissed):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/main/java/com/swirlds/platform/test/consensus/framework/TestInput.java
```diff
@@ -19,17 +19,37 @@
...
  * Holds the input to a consensus test.
+ * @param platformContextSupplier the {@link PlatformContext} to use for the test
+ * @param numberOfNodes the number of nodes in the test
+ * @param weightGenerator the {@link WeightGenerator} to use for the test
+ * @param seed the seed for the random number generator
+ * @param eventsToGenerate the number of events to generate
  */
 public record TestInput(
-        @NonNull PlatformContext platformContext,
+        @NonNull Supplier<PlatformContext> platformContextSupplier,
```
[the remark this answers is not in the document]
edward-swirldslabs: reverted supplier change

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -42,6 +46,30 @@ public static void initConfig() {
...
     }
 
+    /**
+     * Injects a platform context with the birth round ancient threshold set to the value in the test parameters and
+     * adds the temp directory for writing marker files.
+     *
+     * @param params the test parameters
+     * @return the modified test parameters
+     */
+    private ConsensusTestParams modifyParams(ConsensusTestParams params) {
```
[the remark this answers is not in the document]
edward-swirldslabs: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: I'll log an error if the feature is turned on.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
edward-swirldslabs: marker files typically do not have content.   It is simply presence / absence of a named file.   JRS will detect that the directory is non-empty, but the human triaging the failure will want to know the cause.  Without differentiation of file name, we will not know the root cause of the data dump.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
edward-swirldslabs: Added `@Nullable` marking

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
edward-swirldslabs: I've added a flag for turning on the feature, default value is off, and have a valid location set for the default directory.  I've dropped the `none` option.

edward-swirldslabs: The location of the boolean feature flag may need to be moved to a different config, I don't know which config it belongs in.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: Added `@Nullable` on return value type.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
edward-swirldslabs: Made the recommended refactor.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
edward-swirldslabs: Added `@NonNull` annotation

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
edward-swirldslabs: This PR definitely introduces tech debt that will have to be cleaned up.  Whether the method is here or in a utility file, I would expect that all unit tests that instantiate the event pipeline will want to turn marker file writing on and check that no marker files were written as the default behavior.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/BasePlatformUnitTests.java
edward-swirldslabs: Added annotation.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
edward-swirldslabs: I moved the check to the location where we construct the `ConsensusRound`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: removed the redundant check

cody-littley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
cody-littley: I'm ok with it here for now.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param filename the name of the marker file to write.
+     */
+    public void writeMarkerFile(@NonNull final String filename) {
+        if (markerFileDirectory == null) {
+            // Configuration did not set a marker file directory.  No need to write marker file.
+            return;
+        }
+        final Path markerFile = markerFileDirectory.resolve(filename);
+        if (Files.exists(markerFile)) {
```
alittley, line 107: Since this `MarkerFileWriter` object is being used on the consensus thread, I think we need to ensure that there is no possibility of having any unnecessarily repeated IO calls

What do you think about this solution:
- create a hash set in this class, that stores all `filename`s that have been passed into `writeMarkerFile`
- Before checking `Files.exists` here, add a check to see if the `filename` already exists in the set, and return if so

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -42,6 +47,30 @@ public static void initConfig() {
         new TestConfigBuilder().getOrCreateConfig();
     }
 
+    /**
+     * Injects a platform context with the birth round ancient threshold set to the value in the test parameters and
+     * adds the temp directory for writing marker files.
```
alittley, line 52: This method doesn't appear to be doing anything with marker files, I assume this comment is stale?

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -128,7 +175,7 @@ void subQuorumPartitionTests(final ConsensusTestParams params) {
     void cliqueTests(final ConsensusTestParams params) {
         ConsensusTestRunner.create()
                 .setTest(ConsensusTestDefinitions::cliqueTests)
-                .setParams(params)
+                .setParams(modifyParams(params))
                 .setIterations(NUM_ITER)
                 .run();
```
alittley, line 180: Any reason you're not asserting marker file presence on this test?

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -50,9 +79,12 @@ public static void initConfig() {
     void orderInvarianceTests(final ConsensusTestParams params) {
         ConsensusTestRunner.create()
                 .setTest(ConsensusTestDefinitions::orderInvarianceTests)
-                .setParams(params)
+                .setParams(modifyParams(params))
                 .setIterations(NUM_ITER)
                 .run();
+        assertMarkerFile(ConsensusImpl.NO_SUPER_MAJORITY_MARKER_FILE, false);
```
alittley, line 85: These 3 lines are repeated verbatim in each test, consider making a private utility method to assert that these marker files are all absent. Or add an `afterEach` method

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/state/IssDetectorTests.java
```diff
@@ -359,18 +363,24 @@ void mixedOrderTest() {
                                     expectedRoundStatus.get((int) notification.getRound()),
                                     notification.getIssType()));
         });
+        issDetectorTestHelper.getIssNotificationList().forEach(notification -> {
+            switch (notification.getIssType()) {
```
alittley, line 367: This switch could be replaced with 

```
assertMarkerFile(notification.getIssType().toString(), true);
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -461,6 +475,7 @@ private void voteInAllElections(@NonNull final EventImpl votingWitness) {
                         candidateWitness,
                         "coin-" + (countingVote.isSupermajority() ? "counting" : "sig"),
                         diff);
+                markerFileWriter.writeMarkerFile(COIN_ROUND_MARKER_FILE);
```
alittley, line 478: Maybe this should log an error, to fail `jrs` tests

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -664,6 +679,27 @@ private List<EventImpl> getStronglySeenInPreviousRound(final EventImpl event) {
                 getMinRoundGeneration(),
                 Math.max(previousRoundNonExpired, decidedRoundNumber - config.roundsExpired() + 1));
 
+        // Check for no judges or super majority conditions.
+        final long judgeWeights = judges.stream()
+                .mapToLong(event -> getWeight(event.getCreatorId()))
+                .sum();
+        consensusMetrics.judgeWeights(judgeWeights);
+        if (judges.isEmpty()) {
+            markerFileWriter.writeMarkerFile(NO_JUDGES_MARKER_FILE);
```
alittley, line 688: We *always* expect to have judges, right? If so, maybe this should log an error too

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -188,6 +195,11 @@ public class ConsensusImpl extends ThreadSafeConsensusInfo implements Consensus
      */
     private AncientMode ancientMode;
 
+    /** The marker file writer */
+    private MarkerFileWriter markerFileWriter;
+    /** The rate limited logger for rounds without a super majority of weight on judges */
+    private RateLimitedLogger noSupertMajorityLogger;
```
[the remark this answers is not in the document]
edward-swirldslabs: corrected

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -664,6 +679,27 @@ private List<EventImpl> getStronglySeenInPreviousRound(final EventImpl event) {
...
+                .sum();
+        consensusMetrics.judgeWeights(judgeWeights);
+        if (judges.isEmpty()) {
+            markerFileWriter.writeMarkerFile(NO_JUDGES_MARKER_FILE);
+        } else {
+            final boolean superMajority =
+                    Threshold.SUPER_MAJORITY.isSatisfiedBy(judgeWeights, addressBook.getTotalWeight());
+            if (!superMajority) {
+                markerFileWriter.writeMarkerFile(NO_SUPER_MAJORITY_MARKER_FILE);
+                noSupertMajorityLogger.warn(
```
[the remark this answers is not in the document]
edward-swirldslabs: turned into an error.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/PathsConfig.java
```diff
@@ -21,25 +21,28 @@
...
- *     path to data/apps/
- * @param logPath
- * 	   path to log4j2.xml (which might not exist)
+ *
+ * @param settingsUsedDir the directory where the settings used file will be created on startup if and only if
+ *                        settings.txt exists
+ * @param keysDirPath     path to data/keys/
+ * @param appsDirPath     path to data/apps/
+ * @param logPath         path to log4j2.xml (which might not exist)
+ * @param markerFilesDir  path to the directory where marker files are written.  Set the value to `none` to turn off
```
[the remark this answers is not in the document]
edward-swirldslabs: removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/metrics/ConsensusMetrics.java
```diff
@@ -97,4 +97,11 @@ public interface ConsensusMetrics {
      * @return weighted mean
      */
     double getAvgOtherReceivedTimestamp();
+
+    /**
+     * Report the weight of the judges when a round comes to consensus.
+     *
+     * @param weight    the weight of the judges
```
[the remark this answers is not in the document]
edward-swirldslabs: reformatted

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * The logger for this class.
+     */
+    private static final Logger logger = LogManager.getLogger(MarkerFileWriter.class);
+
+    /**
+     * Flag to indicate if we still need to log an info message about the marker file directory not being set.
+     */
+    private static final AtomicBoolean logMarkerFileDirectoryNotSet = new AtomicBoolean(true);
```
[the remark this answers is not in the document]
edward-swirldslabs: removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,128 @@
...
+                .getMarkerFilesDir();
+        failedToWriteMarkerFileLogger = new RateLimitedLogger(logger, platformContext.getTime(), Duration.ofMinutes(1));
+        Path directory = null;
+        if (enabled) {
+            if (markerFileDirectoryPath != null) {
+                try {
+                    Files.createDirectories(markerFileDirectoryPath);
+                    directory = markerFileDirectoryPath;
+                } catch (final IOException e) {
+                    directory = null;
```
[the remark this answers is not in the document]
edward-swirldslabs: removed setting directory to null.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: The work of writing marker files is now done inside the context of a `ConcurrentHashMap.computeIfAbsent` check.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
edward-swirldslabs: oversight, thanks for the catch.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
edward-swirldslabs: added afterEach method

edward-swirldslabs: Although I am wondering now if it indicates while method the after each is called after.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/state/IssDetectorTests.java
edward-swirldslabs: nice!  changed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
edward-swirldslabs: added rate limited logger for coin rounds

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
edward-swirldslabs: added rate limited logger for an error here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
edward-swirldslabs: the `logVote` method does log an info message.  So I don't know if a coin round is technically an error.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -56,6 +52,8 @@ public class MarkerFileWriter {
      */
     private final Path markerFileDirectory;
 
+    private final Map<String, Void> markerFilesWrittern = new ConcurrentHashMap<>();
```
alittley, line 55: "writtern"

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -0,0 +1,124 @@
...
+     * Writes a marker file with the given filename to the configured directory.
+     *
+     * @param filename the name of the marker file to write.
+     */
+    public void writeMarkerFile(@NonNull final String filename) {
+        if (markerFileDirectory == null) {
+            // Configuration did not set a marker file directory.  No need to write marker file.
+            return;
+        }
+        markerFilesWrittern.computeIfAbsent(filename, key -> {
```
alittley, line 103: If I'm interpreting this doc correctly, the way you're doing it here won't work

```
If the specified key is not already associated with a value (or is mapped to null), attempts to compute its value using the given mapping function and enters it into this map unless null.
```

Since you always return `null` as the computed value, seems like the computation will be made every time, regardless

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
edward-swirldslabs: fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
cody-littley: I'm ok resolving this. Mostly was trying to understand better how we intended to use markers.

cody-littley (approved):

PR #12647, opened 2024-04-08
Return NOT_SUPPORTED in assertThrottlingPreconditions()

netopyr:
This PR fixes tests that were failing because the wrong error code was returned

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
@@ -274,7 +274,7 @@ public final class IngestChecker {
             // at consensus, and adding them to normal throttle buckets, c.f.
             // https://github.com/hashgraph/hedera-services/issues/12559
             if (payerNum >= hederaConfig.firstUserEntity()) {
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
```diff
@@ -274,7 +274,7 @@ private void assertThrottlingPreconditions(
             // at consensus, and adding them to normal throttle buckets, c.f.
```
Neeharika-Sompalli, line 274: Comment should be modified to use `NOT_SUPPORTED`

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

tinker-michaelj (approved):
LGTM!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
netopyr: Good catch! Thanks! Fixed

thomas-swirlds-labs (approved):

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12653, opened 2024-04-08
fix: support receipt/record queries with nonce

### LINKED ISSUE #12652: (mod) Support receipt and record queries for child transactions
### Problem

We want to be able to query for a specific child receipt or record by nonce, instead of needing to query the top-level `TransactionID` with e.g. `include_child_receipts=true`.

### Solution

Look up child receipts and records from the `RecordCache` using their top-level `TransactionID`; then check the list of child records for a matching `nonce`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #12652 
 - Fix `DeduplicationCache` comparator to include `nonce`.
 - Remove some misleading comments in `RecordCacheImpl`.
 - Update receipt and record queries to fetch a specific child by `TransactionID` nonce.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
@@ -102,7 +102,8 @@ public class RecordCacheImpl implements HederaRecordCache {
      */
     private final DeduplicationCache deduplicationCache;
     /**
-     * A map of transaction IDs to the histories of all transactions that came to consensus with that ID, or their child
+     * A map of transaction IDs to the histories of all transactions that came to
+     * consensus with that ID, or their child
      * transactions. This data structure is rebuilt during reconnect or restart. Using a non-deterministic, map is
      * perfectly acceptable, as the order of these histories is not important.
      */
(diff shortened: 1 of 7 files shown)


Neeharika-Sompalli (approved):
Makes sense to me. Thanks @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -311,6 +303,11 @@ private void removeExpiredTransactions(
     @Nullable
     @Override
     public History getHistory(@NonNull TransactionID transactionID) {
+        logger.info(
+                "Getting history for transaction ID: {} (contained in dedup cache? {}) - history = {}",
+                transactionID,
+                deduplicationCache.contains(transactionID),
+                histories.get(transactionID));
```
Neeharika-Sompalli, line 310: This can be deleted

Neeharika-Sompalli (approved):

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetReceiptHandler.java
```diff
@@ -84,23 +84,40 @@ public Response findResponse(@NonNull final QueryContext context, @NonNull final
...
+                    final var maybeRecord = history.childRecords().stream()
+                            .filter(record -> transactionId.equals(record.transactionID()))
+                            .findFirst();
+                    if (maybeRecord.isEmpty()) {
+                        responseBuilder.header(header.copyBuilder()
+                                .nodeTransactionPrecheckCode(RECEIPT_NOT_FOUND)
+                                .build());
+                    } else {
+                        responseBuilder.receipt(maybeRecord.get().receipt());
+                    }
```
mhess-swl, line 120: Seems worth extracting to a common method, but definitely not a PR blocker

mhess-swl (approved):
LGTM, thanks @tinker-michaelj!

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/NetworkTransactionGetReceiptHandler.java
tinker-michaelj: True Matt, noticed the duplicated pattern as well...then thought these queries aren't likely to change any time soon, and decided to leave it for now. 🤷

PR #12654, opened 2024-04-08
fix: support receipt/record queries with nonce 

tinker-michaelj:
**Description**:
 - Cherry-pick #12653


Neeharika-Sompalli (approved):

PR #12665, opened 2024-04-08
fix: 12247 Remove incorrect assert from `reopenFileChannel`

### LINKED ISSUE #12247: Unit test failure - `DataFileCollectionTest.merge()`
### Description

`DataFileCollectionTest.merge()` is flaky and needs to be investigated. 

See https://scans.gradle.com/s/6zdhtwz2bmcoo/tests/overview?outcome=FAILED

### Steps to reproduce

Run `DataFileCollectionTest.merge()` and hope for the worst. 

### Additional context

_No response_

### Hedera network

other

### Version

v0.49

### Operating system

None

imalygin:
**Description**:

This PR addresses rare non-deterministic failure of `DataFileCollectionTest#merge` test. It may happen due to race of reading and compacting threads. It turned out that in some cases there _may be_ null in `DataFileReaderPbj.fileChannels` collection, and it's a valid case, we should not assume that it's never null in `DataFileReaderPbj#reopenFileChannel` method. 

**Related issue(s)**:

Fixes #12247 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReaderPbj.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReaderPbj.java
@@ -330,6 +330,8 @@ public class DataFileReaderPbj<D> implements DataFileReader<D> {
      * threads are calling this method to replace it with a new channel, only one of them will
      * proceed, while the channel opened in the other thread will be closed immediately and
      * not used any further.
+     * <br/>
+     * Note: If {@link #fileChannels} has null value for the given index, this method does nothing.
      *
      * @param index File channel index in {@link #fileChannels}
      * @param closedChannel Closed file channel to replace
@@ -338,10 +340,11 @@ public class DataFileReaderPbj<D> implements DataFileReader<D> {
      */
     protected void reopenFileChannel(final int index, final FileChannel closedChannel) throws IOException {
         assert index < maxFileChannels;
-        // May be closedChannel or may be already reopened in a different thread
-        assert fileChannels.get(index) != null;
         assert !closedChannel.isOpen();
         final FileChannel fileChannel = FileChannel.open(path, StandardOpenOption.READ);
+        // It may so happen that `fileChannels` collection was updated by another thread to null for this index
+        // (for example, if the compaction deletes the file)
+        // This is a valid scenario, and the method does nothing in this case
         if (!fileChannels.compareAndSet(index, closedChannel, fileChannel)) {
             fileChannel.close();
         }


artemananiev (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReaderPbj.java
```diff
@@ -338,10 +340,11 @@ protected void openNewFileChannel(final int index) throws IOException {
      */
     protected void reopenFileChannel(final int index, final FileChannel closedChannel) throws IOException {
         assert index < maxFileChannels;
-        // May be closedChannel or may be already reopened in a different thread
-        assert fileChannels.get(index) != null;
         assert !closedChannel.isOpen();
         final FileChannel fileChannel = FileChannel.open(path, StandardOpenOption.READ);
+        // It may so happen that `fileChannels` collection was updated by another thread to null for this index
+        // (for example, if the compaction deletes the file)
+        // This is a valid scenario, and the method does nothing in this case
```
anthony-swirldslabs, line 347: The fix looks good overall. I'm just uncertain about the "does nothing part". The method has a name that suggests it's supposed to leave a channel at `index` open upon return - at least that's what I, as a caller of this method, would expect.

Now if another thread set it to `null` at this index, the method will now keep the `null` at that `index`, and the caller now has to deal with the `null`.

Are you sure the caller is ready to deal with the `null` at `index` after this method returns?

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileReaderPbj.java
imalygin: >The fix looks good overall. I'm just uncertain about the "does nothing" part. 

I was not certain about that too. At first, this removal looked like a violation of the method contract, because the name suggests that it _will_ reopen the channel. However, in this particular context the reopen is not feasible because this file may no longer exist as it was compacted. So, in a conversation with Artem I suggested that may be it makes sense to rename the method to `tryRepoenFileChannel` to reflect that there is a scenario where it does nothing. However, upon further review of the code we decided to leave it as is. 

>Are you sure the caller is ready to deal with the null at index after this method returns?

Yes. It turned out that the calling code is ready for the channel not to be there. Let's consider method `DataFileReaderPbj#read`. This is where we call `reopenFileChannel`:
```
           } catch (final ClosedChannelException e) {
                // This exception may be thrown, if the channel was closed, because a different
                // thread reading from the channel was interrupted. Re-create the file channel
                // and retry
                reopenFileChannel(fcIndex, fileChannel);
            } finally {
```

As you can see, it does 3 retries. Let's assume that the root cause scenario of this issue in unfolding: 
1. One reading thread was reading from the channel and got interrupted - the file channel transitioned to the closed state
2. The next reader thread attempted to use this file channel and got `ClosedChannelException`
3. At the same time compaction thread deleted the file channel, and `reopenFileChannel` did nothing
4. On the next iteration the reader thread will execute this code:
```
            final FileChannel fileChannel = fileChannels.get(fcIndex);
            if (fileChannel == null) {
                // On rare occasions, if we have a race condition with compaction, the file channel
                // may be closed. We need to return null, so that the caller can retry with a new reader
                return null;
            }
```
As you see, the caller is aware of the possibility of getting `null` here.

anthony-swirldslabs: Thanks for the detailed explanation. That sounds good to me. Approved!

PR #12668, opened 2024-04-09
fix: use exact math for all summed balances

### LINKED ISSUE #12667: (mod) Use exact math when summing amounts
### Description

There are several places where `long` values are summed without `Math.addExact()`.

### Steps to reproduce

Submit a `CryptoTransfer` with e.g. only credits `(+Long.MAX_VALUE, +Long.MAX_VALUE, +2)` 

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
  - Fixes #12667 
  - Use `BigInteger` for net-zero adjustment calculations in `CryptoTransfer` pure checks.
  - Use exact math for all custom fee summing.
  - During record finalization,
      1. Use exact math for summing net balance adjustments.
      2. Add an extra safety check in record finalization for any balance that exceeds total hbar supply.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AdjustmentUtils.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AdjustmentUtils.java
@@ -16,10 +16,15 @@
 
 package com.hedera.node.app.service.token.impl.handlers.transfer.customfees;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE;
+
 import com.hedera.hapi.node.base.AccountID;
+import com.hedera.hapi.node.base.ResponseCodeEnum;
 import com.hedera.hapi.node.base.TokenID;
 import com.hedera.hapi.node.transaction.CustomFee;
 import com.hedera.hapi.node.transaction.FixedFee;
+import com.hedera.node.app.spi.workflows.HandleException;
+import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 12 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AdjustmentUtils.java
```diff
@@ -198,8 +203,21 @@ public static void adjustHbarFees(final AssessmentResult result, final AccountID
...
+    public static long addExactOrThrow(final long a, final long b) {
+        return addExactOrThrowReason(a, b, INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE);
+    }
+
+    public static long addExactOrThrowReason(
+            final long a, final long b, @NonNull final ResponseCodeEnum failureReason) {
+        try {
+            return Math.addExact(a, b);
+        } catch (final ArithmeticException ignore) {
+            throw new HandleException(INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE);
```
Neeharika-Sompalli, line 220: Should we be using `failureReason` here ?

Neeharika-Sompalli (approved):
LGTM ! Have just one comment. Thanks @tinker-michaelj 🙌

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/AdjustmentUtils.java
tinker-michaelj: Ah good catch Neeha, fixed.

Neeharika-Sompalli (approved):

PR #12674, opened 2024-04-09
fix: marking states to be saved

### LINKED ISSUE #12639: Fix snaphsot lifecycle
The current way we specify if a round snapshot round is ugly and error prone. These decisions should be made when the round reaches consensus, not after transactions have been handled for the round.

cody-littley:
closes #12639

This image shows the new wiring for the marking state snapshots that should be saved to disk (by default it is collapsed).

Screenshot 2024-04-11 at 8 37 39 AM


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultSavedStateController.java
```diff
@@ -55,20 +56,21 @@ public DefaultSavedStateController(@NonNull final StateConfig stateConfig) {
         this.stateConfig = Objects.requireNonNull(stateConfig);
     }
 
+    // TODO revisit synchronized keywords
```
alittley, line 59: TODO

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
```diff
@@ -477,9 +478,9 @@ private void wire() {
         // to move the logic of SavedStateController::markSavedState into the signedStateFileManager. There is no reason
```
alittley, line 478: This big `FUTURE_WORK` comment is no longer relevant

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/wiring/PlatformWiring.java
lpetrovic05: This comment is actually wrong I think. The reason this logic is not in the `signedStateFileManager` is because state saving can take minutes. And in that time, the queue can build up with states that do not need to be saved to disk. The comment should reflect that so someone else doesn't get confused in the future.

cody-littley: Removed the comment here. Added a comment to `SavedStateController`. I think this sort of thing is more specific to the component and doesn't really belong where we wire things together.

```
/**
 * Controls which signed states should be written to disk based on input from other components.
 * <p>
 * This is intentionally split out of the component that actually saves the state. Saving the state can take a long
 * time, and we don't want the queue of states waiting to be handled by that component to grow large while we are
 * waiting to write a state. It is much better simply not to put states into that component's queue if we don't want to
 * write them to disk.
 */
public interface SavedStateController {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/components/DefaultSavedStateController.java
cody-littley: Removed synchronized keywords (and TODO) since I don't think they are necessary for thread safety.

alittley (approved):

edward-swirldslabs (approved):

PR #12678, opened 2024-04-09
test: Improvements to assertions on sidecar files

### LINKED ISSUE #12629: Improvements to assertions on sidecar files in HapiSuites
### Problem

Currently, the sidecar file assertions (done through the `SidecarWatcher` class) expect you to have added `expectedSidecars` for all incoming sidecar records while the `SidecarWatcher` is running, but give you no flexibility when doing the assertions. For example, in some test scenarios you don't want to test all of them, but only assert that a specific record exists in the file and this is currently not possible.

### Solution

**Your task is to**:
- Refactor the way that the `SidecarWatcher` is doing the assertions to allow you to test only a specific record in the file instead of always testing the whole file.

**A nice addition to have**:
- Create the infrastructure to match the sidecar records using custom assertions (using something similar to the `ArgumentMatcher` class in Mockito - might need a custom implementation here)

### Alternatives

_No response_

victor-yanev:
**Description**:

Currently, the sidecar file assertions (done through the `SidecarWatcher` class) expect you to have added `expectedSidecars` for all incoming sidecar records while the `SidecarWatcher` is running, but give you no flexibility when doing the assertions.

For example, in some test scenarios you don't want to test all of them, but only assert that a specific record exists in the file and this is currently not possible. If you have missed to add some record to the `expectedSidecars` queue, the `assertIncomingSidecar` method will throw a `NullPointerException` making the test fail.

**Solution**:

This PR adds the following changes:
- Refactors the way that the `SidecarWatcher` is doing the assertions to allow you to test only a specific record in the file instead of always testing the whole file.
- Adds the infrastructure to match the sidecar records using custom assertions (using custom implementations of `hamcrest` matchers)

**Related issue(s)**:

Fixes #12629 

**Notes for reviewer**:

**Checklist**


georgi-l95 (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/assertions/matchers/TransactionSidecarRecordMatcher.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    /**
+     * @param sidecarRecord the actual {@link TransactionSidecarRecord}
+     * @return {@code true} if the consensus timestamp of the given
+     * {@link TransactionSidecarRecord} matches the expected value
+     */
+    public boolean matchesConsensusTimestampOf(final TransactionSidecarRecord sidecarRecord) {
+        final Matcher<Timestamp> matcher = timestampMatcher.apply(consensusTimestamp);
+        return consensusTimestamp == null || matcher.matches(sidecarRecord.getConsensusTimestamp());
```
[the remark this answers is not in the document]
victor-yanev: The idea is that we match only the fields which are provided. 

For example, if we want to match only the `contractActions` we set only the `contractActions` of the `TransactionSidecarRecordMatcher` and all fields which are null in the matcher won't be compared to the actual object.

But you are right about the `consensusTimestamp`, it should be a required field, since we use it as an identifier when comparing the sidecar records. Without it we cannot know which record the expected record is supposed to be matched against.

MrValioBg (dismissed):
LGTM! 💯

jeromy-cannon (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/leaky/LeakyCryptoTestsSuite.java
```diff
@@ -1221,14 +1221,14 @@ final HapiSpec lazyCreateViaEthereumCryptoTransfer() {
                                                         .setCallOperationType(CallOperationType.OP_CALL)
                                                         .setCallingAccount(senderAccountIdReference.get())
                                                         .setRecipientAccount(lazyAccountIdReference.get())
-                                                        .setGas(2_000_000L)
+                                                        .setGas(629_000L)
```
tinker-michaelj, line 1224: Interesting...was this only passing before due to limitations in the matcher?

tinker-michaelj (approved):
Awesome work, tyvm @victor-yanev !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/leaky/LeakyCryptoTestsSuite.java
victor-yanev: Nope, the `gasLimit` of `lazyCreateTxn` got changed from `2_000_000L` to `650_000L` in another PR on `05.04.24` and it caused this assertion here to fail. 

I just fixed it cause I noticed it fails while I was working on the assertions.

tinker-michaelj (approved):

PR #12679, opened 2024-04-09
feat: Check platform status before syncing (#11429)

alittley:
- cherry pick of https://github.com/hashgraph/hedera-services/pull/11429

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncStatusChecker.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncStatusChecker.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


edward-swirldslabs (approved):

poulok (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncStatusChecker.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A utility class for checking if a platform status permits syncing
+ */
+public class SyncStatusChecker {
+    /**
+     * The platform statuses that permit syncing. If the platform isn't in one of these statuses, no syncs will be
+     * initiated or accepted
+     */
+    public static final Collection<PlatformStatus> STATUSES_THAT_PERMIT_SYNC = Set.of(
```
kfa-aguda, line 32: is this intended to be public? Don't see a reference to it from outside of this class.

cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncStatusChecker.java
alittley: No strong reason to be public.

Since this code is already in `develop`, we wouldn't want to make style changes here, though, to avoid discrepancies between the release and `develop`. But if we wanted to change this to `private` in `develop`, I wouldn't have any objections

poulok: Good point @kfa-aguda, this can be changed to `private` in `develop`. However we will leave it as is in this PR due to the tight timeline required for this fix. We do not want to go through another PR gate cycle for this.

kfa-aguda: works with me.

PR #12683, opened 2024-04-09
refactor: Use `TestingEventBuilder` in `InOrderLinkerTests`

alittley:
- part of #11389

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
+++ b/platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
@@ -17,37 +17,29 @@
 package com.swirlds.platform.event.linking;
 
 import static com.swirlds.common.test.fixtures.RandomUtils.getRandomPrintSeed;
-import static com.swirlds.common.test.fixtures.RandomUtils.randomHash;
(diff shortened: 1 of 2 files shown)


lpetrovic05 (approved):

cody-littley (approved):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/event/linking/InOrderLinkerTests.java
```diff
@@ -541,31 +445,26 @@ void selfParentTimeCreatedMismatch(final boolean useBirthRoundForAncient) {
 
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
-    @DisplayName("Other parent with mismatched time created should be linked")
+    @DisplayName("Other parent with mismatched time created SHOULD be linked")
```
kfa-aguda, line 448: I think the displayName vs javadoc consideration we discussed offline applies here. As you're modifying the long name now, probably worth changing this to a javadoc, but only if you have other revisions to make to this PR. Otherwise, it can wait.

alittley: @kfa-aguda This is a good point, which I'll bear in mind moving forward.

For now, as suggested, I'll merge without fixing, since no other revisions need to be made on this PR.

PR #12688, opened 2024-04-09
fix: ensure precompile calls with `INSUFFICIENT_GAS` are fully traced

tinker-michaelj:
**Description**:
 - mono-service port of #12407 
 - Remove apparently intentional but incorrect block in `HederaTracer.tracePrecompileResult()` that skips popping the action stack when the precompile was aborted with `INSUFFICIENT_GAS`.


lukelee-sl (approved):
LGTM

netopyr (changes requested):
Does this PR go into `release/0.48` or `release/0.49`?

tinker-michaelj:
Unnecessary.

PR #12692, opened 2024-04-09
feat: FileSystemManager

### LINKED ISSUE #12690: Create FileSyetemManager v1
This is the current proposal:
```
public interface FileSystemManager {
    @NonNull
    Path resolve(@NonNull Path path);

    @NonNull
    Path createTemporaryPath(@NonNull String name);

    void recycle(@NonNull Path path) throws IOException;
}
```

Other considerations:
- we need to stop people from touching other parts of the file system
- renaming the old temporary file builder to LegacyTemporaryFileBuilder, and saving useful code into a new non-static version

mxtartaglia-sl:
**Related issue(s)**:

Fixes #12690


platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * {@inheritDoc}
+     * This implementation is nominal.
+     */
+    @Override
+    public void recycle(@NonNull final Path relativePath) throws IOException {
+        bin.recycle(subPathFromTempLocation(relativePath));
+    }
+
+    private Path subPathFromTempLocation(@NonNull final Path subpath) {
```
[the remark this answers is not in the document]
cody-littley: It is unclear what the "fromTempLocation" means in this context

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
```diff
@@ -0,0 +1,96 @@
...
+        try {
+            if (Files.isSameFile(path, rootPath)) {
+                throw new IllegalArgumentException(
+                        "Requested subPath resolves to the root temporary path '" + path + ".");
+            }
+            return path;
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Cannot assert if provided path against temporaryFileLocation", e);
+        }
+    }
```
cody-littley, line 95: I suggest we do this slightly differently.

- root
    - temporary location
    - recycle bin location
    - data 

We should put all directories managed internally by the file system manager underneath the root. There should be a `data` directory (we can workshop the name for this if you don't like it), and all paths created by `resolve(path)` should be underneath the `data` directory. That way, there is never any possibility of collision between directories managed internally by the manager and directories managed by the end user of this utility.

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param name the name to use as suffix for the directory
+     * @return the resolved path
+     * @throws IllegalArgumentException if the path cannot be relative to or scape the root directory
+     */
+    @NonNull
+    @Override
+    public Path createTemporaryPath(@NonNull final String name) {
+        final String fileName = FILE_ID.getAndIncrement() + "-" + name;
+        return subPathFromTempLocation(Paths.get(fileName));
+    }
```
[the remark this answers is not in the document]
mxtartaglia-sl: Indeed. Sorry, I forgot to push the last version yesterday!

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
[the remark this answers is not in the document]
mxtartaglia-sl: I will do that, but I'm just wondering why we need Javadocs for private methods...

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/FileSystemManager.java
```diff
@@ -0,0 +1,56 @@
...
+package com.swirlds.filesystem.manager;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.IOException;
+import java.nio.file.Path;
+
+/**
+ * Responsible for organizing and managing access to the file system.
+ */
+public interface FileSystemManager {
```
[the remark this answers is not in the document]
mxtartaglia-sl: This was a hard one, still much to do to fully use the manager, but had to replace how we access the recycleBin and test

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
mxtartaglia-sl: instead of data used usr, let me know what you thing about that

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerFactoryImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import com.swirlds.config.api.Configuration;
+import com.swirlds.metrics.api.Metrics;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.nio.file.Path;
+
+public class FileSystemManagerFactoryImpl implements FileSystemManagerFactory {
+
+    private static class InstanceHolder {
+        private static final FileSystemManagerFactoryImpl INSTANCE = new FileSystemManagerFactoryImpl();
+    }
```
cody-littley, line 37: Why are we providing static access to the file system manager? Static access to things like this is something that I think we should go out of our way to discourage.

If there are currently things that need access to things like the temporary file builder, I suggest that we rename the old static implementation to something like `LegacyTemporaryFileBuilder` and mark it as depreciated. The eventual goal for the static use cases should be to stop accessing things via static stateful variables.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerImpl.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final Logger logger = LogManager.getLogger(FileSystemManagerImpl.class);
+    private static final String TMP = "tmp";
+    private static final String USER = "usr";
+    private static final String BIN = "recycle-bin";
+    private final Path rootPath;
+    private final Path tempPath;
+    private final Path userPath;
+    private final Path recycleBinPath;
+    private final RecycleBin bin;
+    private static final AtomicLong TMP_FIELD_INDEX = new AtomicLong(0);
```
cody-littley, line 60: I don't think this should be static. We shouldn't try to prevent collisions between two file system managers. Instead, we should rely on the fact that there should only be a single file system manager that has ownership of a directory.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/utility/TemporaryFileBuilder.java
```diff
@@ -44,9 +44,10 @@ private TemporaryFileBuilder() {}
      */
     public static synchronized Path getTemporaryFileLocation() throws IOException {
         if (temporaryFileLocation == null) {
-            final TemporaryFileConfig config = ConfigurationHolder.getConfigData(TemporaryFileConfig.class);
+            final FileSystemManagerConfig config = ConfigurationHolder.getConfigData(FileSystemManagerConfig.class);
             final StateCommonConfig stateConfig = ConfigurationHolder.getConfigData(StateCommonConfig.class);
-            overrideTemporaryFileLocation(getAbsolutePath(config.getTemporaryFilePath(stateConfig)));
+            overrideTemporaryFileLocation(getAbsolutePath(
+                    stateConfig.savedStateDirectory().resolve(config.rootPath()).toString()));
         }
```
cody-littley, line 51: I suggest we leave this class as is, rename it to `LegacyTemporaryFileBuilder`, and mark it as deprecated. Then we should put pressure on the remaining users of the static version to convert to non-static access patterns.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/CompareStatesCommand.java
```diff
@@ -165,7 +165,7 @@ public Integer call() throws IOException {
         final Configuration configuration = DefaultConfiguration.buildBasicConfiguration(
                 ConfigurationBuilder.create(), getAbsolutePath("settings.txt"), configurationPaths);
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 168: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/DiagramCommand.java
```diff
@@ -121,7 +121,7 @@ private void setLessMystery(final boolean lessMystery) {
     public Integer call() throws IOException {
         final Configuration configuration = DefaultConfiguration.buildBasicConfiguration(ConfigurationBuilder.create());
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 124: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/DiagramLegendCommand.java
```diff
@@ -59,7 +59,7 @@ public Integer call() throws IOException {
         BootstrapUtils.setupConstructableRegistry();
 
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 62: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/EventStreamRecoverCommand.java
```diff
@@ -129,7 +129,7 @@ public Integer call() throws Exception {
         final Configuration configuration = DefaultConfiguration.buildBasicConfiguration(
                 ConfigurationBuilder.create(), getAbsolutePath("settings.txt"), configurationPaths);
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 132: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/GenesisPlatformStateCommand.java
```diff
@@ -73,7 +73,7 @@ public Integer call() throws IOException, ExecutionException, InterruptedExcepti
         BootstrapUtils.setupConstructableRegistry();
 
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 76: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/cli/ValidateAddressBookStateCommand.java
```diff
@@ -73,7 +73,7 @@ public Integer call() throws IOException, ExecutionException, InterruptedExcepti
         BootstrapUtils.setupConstructableRegistry();
 
         final PlatformContext platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 76: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/internal/RecoveryPlatform.java
```diff
@@ -88,7 +88,7 @@ public RecoveryPlatform(
 
         notificationEngine = NotificationEngine.buildEngine(getStaticThreadManager());
 
-        context = new DefaultPlatformContext(configuration, metrics, CryptographyHolder.get(), Time.getCurrent());
+        context = new DefaultPlatformContext(configuration, metrics, CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 91: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/editor/StateEditor.java
```diff
@@ -68,7 +68,7 @@ public StateEditor(final Path statePath) throws IOException {
         final Configuration configuration = DefaultConfiguration.buildBasicConfiguration(ConfigurationBuilder.create());
 
         platformContext = new DefaultPlatformContext(
-                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 71: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/editor/StateEditorSave.java
```diff
@@ -78,7 +78,7 @@ public void run() {
                     DefaultConfiguration.buildBasicConfiguration(ConfigurationBuilder.create());
 
             final PlatformContext platformContext = new DefaultPlatformContext(
-                    configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent());
+                    configuration, new NoOpMetrics(), CryptographyHolder.get(), Time.getCurrent(), null);
```
cody-littley, line 81: Should this be null? I think it would be ok to instantiate a file system manager in the current working directory for this.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerImpl.java
mxtartaglia-sl: Okay, but nothing prevents multiple instances of the fileSystemManager from being created at the same location. If the index is not static, those two instances could create files in the same directory with the same name. Thoughts?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/utility/TemporaryFileBuilder.java
mxtartaglia-sl: Absolutely. I hold that change because of this [pr](https://github.com/hashgraph/hedera-services/pull/12734/files), but I can also include it here.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/FileSystemManager.java
```diff
@@ -0,0 +1,64 @@
...
+    Path resolve(@NonNull Path relativePath);
+
+    /**
+     * Creates a path relative to the root directory of the file system manager.
+     * Implementations can choose the convenient subfolder inside the root directory.
+     * There is no file or directory actually being created after the invocation of this method.
+     *
+     * @param tag if indicated, will be suffixed to the returned path
+     * @return the resolved path
+     * @throws IllegalArgumentException if the path is "below" the root directory (e.g. resolve("../foo")
```
poulok, line 49: I tend to think of the example being "above" the root directory

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param selfId      the ID of this node
+     * @return the location where recycle bin files are stored
+     */
+    public Path getRootPath(@NonNull final StateCommonConfig stateConfig, @NonNull final NodeId selfId) {
+        return getAbsolutePath(stateConfig
+                .savedStateDirectory()
+                .resolve(Long.toString(
+                        selfId.id())) // Or is the node the first parameter ?? how do we write the states files?
+                .resolve(rootPath()));
+    }
```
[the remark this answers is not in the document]
mxtartaglia-sl: Currently, the files are inside the  .savedStateDirectory(), which matches how the folder structure works today. We discussed with Kelly not to break the current folder structure for the node, but if I get you correctly, then the desired structure would be:

```
* node-root (it might contain the node id)
   * savedStateDirectory
   * tmp
   * usr
   * recycle-bin
```
FS-manager root matches node-root   
   
If I do that then It impacts how we configure the node, so now is
```
* savedStateDirectory
   *fs-manager-root
      * tmp
      * usr
      * recycle-bin  
```
Do you think is ok for now

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerFactoryImpl.java
mxtartaglia-sl: As we discussed in Slack:
 * The static access is provided for the Factory, not the FileSystemManager
 * The solution uses both Singleton + Factory patterns as many places in base team code.
 * There is a point in saying that we might not need the factory, and providing the ability to change implementing types provides more flexibility than we would use.
  @hendrikebbers , can you jump in? What do you think about the use of a factory in this case?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/FileSystemManager.java
mxtartaglia-sl: Agree

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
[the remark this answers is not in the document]
cody-littley: In the short term, we absolutely must not change the directories that outside processes are utilizing. A few examples that come to mind:

- saved states
- event stream
- record stream

Directories that are managed 100% by the java code are ok to change, as log as there is sufficient testing, and if we do appropriate migration and cleanup. E.g.

- temporary files
- recycle bin
- preconsensus event stream
- address book debug files

I suggest an incremental adoption of the new file system manager. When we first deploy this, we should accept the fact that not all files will utilize this tool immediately. Then, we should target the legacy files one at a time and come up with migration solutions.

One nice to have, if we could use configuration to change the "usr" directory name to a different name, we could have the same file path for a lot of important files. Make "data" the root directory, and "saved" the usr directory. That way, anything created by `FileSystemManager.resolve()` will live beneath `data/saved` (i.e. where most stuff currently lives).

Note that we will need special cleanup logic for the old temporary file directory and recycle bin directory during the first release where this is deployed. Normally cleanup is done at system boot time. But since we are changing those paths, the regular workflow won't know where to clean up.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerImpl.java
cody-littley: I suggest we get a small group of decision makers together to discuss static vs non-static behavior. My personal preference is to take steps towards eliminating static stateful variables entirely.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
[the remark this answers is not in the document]
mxtartaglia-sl: Ok! I like this, and it is a similar approach to what has been discussed internally with the Base Team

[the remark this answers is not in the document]
mxtartaglia-sl: Would the things inside "saved" need to be deleted at startup? or would that only stand for the temp dir?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
```diff
@@ -0,0 +1,45 @@
...
+public record FileSystemManagerConfig(
+        @ConfigProperty(defaultValue = "data")
+                String rootPath, // On purpose matches root of {@link StateCommonConfig#savedStateDirectory()}
+        @ConfigProperty(defaultValue = DATA) String userDataDir,
+        @ConfigProperty(defaultValue = TMP) String tmpDir,
+        @ConfigProperty(defaultValue = BIN) String recycleBinDir,
+        @ConfigProperty(defaultValue = "7d") Duration recycleBinMaximumFileAge,
+        @ConfigProperty(defaultValue = "1d") Duration recycleBinCollectionPeriod) {
+
+    public static final String TMP = "tmp";
```
hendrikebbers, line 42: I do not know if I like this. Looks like you only use it in unit tests. I would create a TMP_DEFAULT_VALUE constant in unit tests instead of providing it as public api.

cody-littley (changes requested):
~We should not merge this until we have resolved the conversation with the use of static variables.~

resolved

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
mxtartaglia-sl: If I rename it to DEFAULT_TMP_DIR_NAME?
The intention is to be able to use the property's default value without instantiating a record through the config API (something that happens occasionally in the code).
If I create a separate constant for the test, then if the default value changes, we have to remember to change it on the test.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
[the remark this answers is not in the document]
mxtartaglia-sl: The root is now data/ and the user-data dir matches saved

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/context/DefaultPlatformContext.java
```diff
@@ -33,25 +36,51 @@ public final class DefaultPlatformContext implements PlatformContext {
...
+     * @param cryptography      the cryptography
+     * @param time              the time
+     * @param fileSystemManager the fileSystemManager
+     */
+    public DefaultPlatformContext(
+            @NonNull final Configuration configuration,
+            @NonNull final Metrics metrics,
+            @NonNull final Cryptography cryptography,
+            @NonNull final Time time,
+            @Nullable final FileSystemManager fileSystemManager) {
```
cody-littley, line 77:
```suggestion
            @NonNull final FileSystemManager fileSystemManager) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/context/DefaultPlatformContext.java
```diff
@@ -89,4 +118,13 @@ public Metrics getMetrics() {
     public Time getTime() {
         return time;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Nullable
```
cody-littley, line 125:
```suggestion
    @NonNull
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
```diff
@@ -0,0 +1,45 @@
...
+                String rootPath, // On purpose matches root of {@link StateCommonConfig#savedStateDirectory()}
+        @ConfigProperty(defaultValue = DEFAULT_DATA_DIR_NAME) String userDataDir,
+        @ConfigProperty(defaultValue = DEFAULT_TMP_DIR_NAME) String tmpDir,
+        @ConfigProperty(defaultValue = DEFAULT_RECYCLE_BIN_DIR_NAME) String recycleBinDir,
+        @ConfigProperty(defaultValue = "7d") Duration recycleBinMaximumFileAge,
+        @ConfigProperty(defaultValue = "1d") Duration recycleBinCollectionPeriod) {
+
+    public static final String DEFAULT_TMP_DIR_NAME = "tmp";
+    public static final String DEFAULT_DATA_DIR_NAME = "saved";
+    public static final String DEFAULT_RECYCLE_BIN_DIR_NAME = "recycle-bin";
```
cody-littley, line 44: The current name for this directory is `data/saved/swirlds-recycle-bin`. We probably need some migration code. All it needs to do is copy the directory `swirlds-recycle-bin` to `recycle-bin` if it exists.

We don't currently need the same for the legacy temporary file builder, since we will be running both the old and the new versions for now.

cody-littley (approved):

platform-sdk/swirlds-filesystem-manager/src/main/java/com/swirlds/filesystem/manager/internal/FileSystemManagerImpl.java
mxtartaglia-sl: As we spoke in slack, the data folder became the root of the fs manager, and saved became the user folder

cody-littley (dismissed):

hendrikebbers (dismissed):

jeromy-cannon (dismissed):

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerImplTest.java
```diff
@@ -0,0 +1,278 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
jsync-swirlds, line 8: The copyright header is duplicated in this file.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.swirlds.common.io.filesystem.FileSystemManager;
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import java.time.Duration;
+
+/**
+ * Settings for {@link FileSystemManager}
+ *
+ * @param rootPath
```
jsync-swirlds, line 27: Unnecessary newline makes this a little odd.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/context/DefaultPlatformContext.java
```diff
@@ -33,25 +35,51 @@ public final class DefaultPlatformContext implements PlatformContext {
...
      *
-     * @param configuration the configuration
-     * @param metrics       the metrics
-     * @param cryptography  the cryptography
-     * @param time          the time
+     * @param configuration     the configuration
+     * @param metrics           the metrics
+     * @param cryptography      the cryptography
+     * @param time              the time
      */
```
jsync-swirlds, line 47: We should probably stop trying to align these, it creates a lot of comment churn.

jsync-swirlds (dismissed):
Functional code seems reasonable.
There are a few minor items with newlines and a duplicated text block.

imalygin (approved):

netopyr (changes requested):
Can you please redirect the test classes in `hedera-node` to the new API instead of just redirecting it to deprecated API? Otherwise we will add several warnings that nobody will look at until we happen to touch the files.

mxtartaglia-sl:
Thanks for the review @netopyr.
Regarding your request, here is the limitation: without having the `VirtualMerkleLeafHasher` use the new API, which requires a full refactor from the data team, these tests have to use the `LegacyTemporaryFileBuilder` in some way.
In the way that it is now, it is being used consistently with VirtualMerkleHasher to create the dirs under the same expected paths; the alternative is to use a `tmpdir` from Junit, which still requires calling to `LegacyTemporaryFileBuilder#overrideTemporaryFileLocation` so that  VirtualMerkleLeafHasher doesn't fail for not being able to copy files between different folders.
We get the warnings in both. I think this should be addressed after the data team migrates the use of the tool internally. 
What do you think?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/config/FileSystemManagerConfig.java
mxtartaglia-sl: if I need to add a change in this PR, I will fix this; if not, I'll use a future PR to address it.
Thanks for spotting.

platform-sdk/swirlds-common/src/test/java/com/swirlds/common/io/filesystem/internal/FileSystemManagerImplTest.java
mxtartaglia-sl: thanks! I will address this in case a change is needed or a different PR if not!

netopyr:
If we cannot use the new API, we should not mark the existing API as deprecated, in my opinion. Priorities change quickly in this project, and refactoring may take much longer than expected. (It is also questionable if something should be labeled legacy if no alternative exists, but at least this creates no warnings. 🙂)

mxtartaglia-sl:
@netopyr  I completely agree with your point, especially avoiding any unnecessary warning (deprecated or otherwise) that cannot be fixed or may take a while to fix.

But here are the reasons why I went along with the proposed change of making it as deprecated:
* The pr includes a replacement to the api, the problem is that it is not a direct replacement, it is necessary a refactor to be able to be used, until both functionality and test are refactored, we are stuck with using the legacy API in both places. It seemed consistent with any other 3rd party library change incorporated, we would first get the new deprecated methods and then work to fix the usage with it's proposed alternative.
* By deprecating the old builder, I think it would ward off new ussages. 
 * I was planning on completely deleting LegacyTemporaryFileBuilder after all usages are removed, It wont live with the deprecated annotation after that.
 
 If I remove the '@deprecated' annotation I will have to leave a 'do not use, use the alternative ... ' comment in the javadoc, which IMO seems not strong enough information or motivation.
 
 if you still think is better not to deprecate it, I'll make the changes... If not I'll help the data-team according to it's priorities to get it changed.
lmk

jsync-swirlds:
Just a possibly helpful note.
If you remove the (definitely premature) `forRemoval = true` parameter to the deprecated annotation then the warnings are not triggered.

Setting for removal is a declaration that the API will be removed _in the next release_, and _should not_ be used for internal APIs at all.  For external APIs it should only be used when the majority of API clients are expected to have already migrated to the replacement API.

hendrikebbers:
If I see it correctly the new API provides the functionally to replace the old API for 100%. Next to that we added documentation to the old API that shows how the new API can be used. 

It is not the goal of the base team to refactor any code to use the new api. That should be done by the hashgraph / data / services / ... teams. For sure we can become better in communicating that a new API is ready to use.

Based on that `Deprecated` is the 100% correct annotation from my point of view. What can be discussed if `forRemoval = true` should be placed at the old api. I would say **yes**: I add the `Deprecated` annotation to an API that should not be used at new places. Even if the replacement is not ready to use I already add the annotation to the legacy api to make it clear that devs should try to avoid using it. If a new api exists that is already used at some places and that api provides 100% of the functionality or a good workaround for everything of the legacy api than it would always add a `forRemoval = true` since at that point we know that the api will be removed as soon as possible. It is not our goal to maintain 2 apis that have the same functionality. Based on that I disagree with @netopyr and @jsync-swirlds

jsync-swirlds:
@hendrikebbers Your position is reasonable.  I am just offering a mechanism to reduce the massive array of warnings that we cannot do anything to clear.
Getting the codebase to a state where it compiles without warnings is also reasonable, and marking items that are deprecated but _cannot_ currently be removed as `forRemoval = true` generates exactly the type of non-actionable warning we should prefer to avoid.

I don't have a hard preference either direction, but I would like to see fewer warnings about deprecated items that we simply cannot (or will not) clear because the replacement is not yet ready or we lack time and priority to take effective action.

jsync-swirlds (dismissed):

cody-littley (dismissed):

netopyr:
An API should describe what is possible right now rather than what is intended to be implemented in the future. What is the point of discouraging the usage of a method if there is no alternative?

Soon, we will start a major effort to reduce or even remove all warnings in modularized code. We must not add warnings now that cannot be resolved.

I think @jsync-swirlds 's suggestion to remove the `for Removal = true` parameter is a good compromise.

jeromy-cannon (dismissed):

mxtartaglia-sl:
I share Hendrik's point of view, but I removed the `for removal = true`  for consensus... 
 `@Deprecated(forRemoval = true)` exists in at least 72 other places in the code base... it is not a topic that we will solve in one PR... we either all get on the same page or start trusting the responsible team's technical criteria.

jsync-swirlds (approved):

jeromy-cannon (dismissed):

hendrikebbers (approved):

hendrikebbers:
> We must not add warnings now that cannot be resolved.
@mxtartaglia-sl @netopyr just for me for understanding: I assumed that each warning can "easily" be removed by switching to the new API as described in the additional javadoc of the deprecated methods. Is that wrong?

cody-littley (approved):

mxtartaglia-sl:
You are not wrong @hendrikebbers

Neeharika-Sompalli (approved):
LGTM from `hedera-base` changes

jsync-swirlds (approved):

cody-littley (approved):

jeromy-cannon (approved):

timo0 (approved):

Neeharika-Sompalli (approved):
LGTM !

imalygin (approved):

PR #12698, opened 2024-04-10
chore: Implement un-HAPI tests for TokenAssociateToAccountHandler

### LINKED ISSUE #12675: Fix pureChecks
Add HAPI-tests for all transactions and queries with invalid data (e.g. missing mandatory fields, non-existing IDs). The tests need to make sure that we do not throw unexpected Exceptions and return the correct error code.
```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-services/issues/12717
- [ ] https://github.com/hashgraph/hedera-services/issues/12644
- [ ] https://github.com/hashgraph/hedera-services/issues/12711
- [ ] https://github.com/hashgraph/hedera-services/issues/12689
- [ ] https://github.com/hashgraph/hedera-services/issues/12619
- [ ] https://github.com/hashgraph/hedera-services/issues/12702
- [ ] https://github.com/hashgraph/hedera-services/issues/12714
- [ ] https://github.com/hashgraph/hedera-services/issues/12773
```

netopyr:
**Description**:
This PR adds HAPI tests to `TokenAssociationSpec` which check invalid transactions are handled correctly.

**Related issue(s)**:
Fixes #12675

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenAssociationSpecs.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenAssociationSpecs.java
@@ -18,6 +18,7 @@ package com.hedera.services.bdd.suites.token;
 
 import static com.hedera.services.bdd.junit.TestTags.TOKEN;
 import static com.hedera.services.bdd.spec.HapiSpec.defaultHapiSpec;
+import static com.hedera.services.bdd.spec.HapiSpec.propertyPreservingHapiSpec;
 import static com.hedera.services.bdd.spec.assertions.NoTokenTransfers.emptyTokenTransfers;
 import static com.hedera.services.bdd.spec.assertions.SomeFungibleTransfers.changingFungibleBalances;
 import static com.hedera.services.bdd.spec.assertions.TransactionRecordAsserts.recordWith;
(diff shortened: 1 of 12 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenAssociationSpecs.java
```diff
@@ -119,6 +129,59 @@ public boolean canRunConcurrent() {
...
+                                .fee(DEFAULT_FEE)
+                                .signedBy(DEFAULT_PAYER)
+                                .hasPrecheck(INVALID_ACCOUNT_ID),
+                        tokenAssociate(bob, VANILLA_TOKEN).fee(DEFAULT_FEE).hasKnownStatus(ACCOUNT_DELETED),
+                        tokenAssociate(alice, List.of()).hasKnownStatus(SUCCESS),
+                        tokenAssociate(alice, VANILLA_TOKEN, VANILLA_TOKEN)
+                                .hasPrecheck(TOKEN_ID_REPEATED_IN_TOKEN_LIST),
+                        tokenAssociate(alice, unknownID).hasKnownStatus(INVALID_TOKEN_ID),
+                        tokenAssociate(alice, TBD_TOKEN).hasKnownStatus(TOKEN_WAS_DELETED),
+                        tokenAssociate(alice, KNOWABLE_TOKEN).hasKnownStatus(TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT));
```
mhess-swl, line 164: These permutations feel like they belong in unit tests. But, if they do, then why didn't the unit tests we have catch this? 🤷 

I'm nervous that adding a large number of negative-case hapi tests for some 50 transaction handlers is going to severely increase the amount of time these tests take. But I'm not sure what to do about it...

mhess-swl (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenAssociationSpecs.java
netopyr: It is a good question why no unit tests caught these errors. I believe part of the answer is that we redefined the purpose of `pureChecks()` and `preHandle()` multiple times while the handlers were already in development.

Having these tests is valuable, in my opinion, because we will move the checks around to do as little as possible during ingest. These tests will ensure that, overall, the errors are still handled as expected. Also, by introducing these tests, we fix which error code is returned in a given scenario, and we can document it.

PR #12713, opened 2024-04-10
feat: allow the deterministic deployment proxy contract transaction to succeed

lukelee-sl:
**Description**:
Foundry (amongst other tools and products) deploy a deterministic deployment contract described [here](https://github.com/Arachnid/deterministic-deployment-proxy?tab=readme-ov-file#deployment-transaction) which is deployed using a signed transaction signed by a particular evm address.

Unfortunately, this transaction does not succeed in Hedera because of the difference in gas price.  This pr is an attempt to address this issue.

**Related issue(s)**:

Fixes # [issue](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2287)

**Checklist**


hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -193,8 +207,13 @@ public long getAmount() {
     }
 
     public BigInteger getMaxGasAsBigInteger() {
+        long multiple = 1L;
+        if (Arrays.equals(rawTx, FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION)) {
```
MiroslavGatsanoga, line 211: **thought:** Perhaps we should check for the `FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION` value only for the `LEGACY_ETHEREUM` type of transactions instead of of all types?‏

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -533,4 +536,36 @@ void maxGasIsPositive() {
             }
         }
     }
+
+    @Test
+    void maxGasForFoundryDeterministicDeployerIsAsExpected() {
+        final var oneByte = new byte[] {1};
+        // high bit of most significant byte is zero
```
IvanKavaldzhiev, line 543: Isn't this comment related to the line above? If so, we can move it on top.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCreateSuite.java
```diff
@@ -162,6 +169,26 @@ public boolean canRunConcurrent() {
         return true;
     }
 
+    @HapiTest
+    final HapiSpec createFoundryDeterministicDeployer() {
+        final var creatorAddress = ByteString.copyFrom(CommonUtils.unhex(FOUNDRY_DEPLOYMENT_SIGNER));
```
IvanKavaldzhiev, line 174: Is this address used in the test? Shouldn't it be the payer of the transaction instead of the default `PAYER`?

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
lukelee-sl: makes sense.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
lukelee-sl: removed the comment.  I copied from the unit test above this one.  I don't think that comment adds much clarity.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCreateSuite.java
lukelee-sl: Used in line 183 to lazily create and fund account that signed the transaction.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -60,6 +61,19 @@ public record EthTxData(
...
     static final BigInteger LEGACY_V_BYTE_SIGNATURE_1 = BigInteger.valueOf(28);
 
+    static final int FOUNDRY_DETERMINISTIC_DEPLOYER_GAS_PRICE_MULTIPLIER = 100;
+    // The specific transaction bytes that are used to deploy the Foundry Deterministic Deployer contract
+    public static final byte[] FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION;
+
+    static {
+        try {
+            FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION = Hex.decodeHex(
+                    "f8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222");
```
david-bakin-sl, line 71: Can we reference https://github.com/Arachnid/deterministic-deployment-proxy?tab=readme-ov-file#deployment-transaction in a comment here?  (It's in the PR comment but that's disconnected from the source code.)

tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
david-bakin-sl: BTW, that readme also says:
> **Note:** as of last readme update; don't trust these to be latest!
!!

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -60,6 +61,19 @@ public record EthTxData(
     static final BigInteger LEGACY_V_BYTE_SIGNATURE_0 = BigInteger.valueOf(27);
     static final BigInteger LEGACY_V_BYTE_SIGNATURE_1 = BigInteger.valueOf(28);
 
+    static final int FOUNDRY_DETERMINISTIC_DEPLOYER_GAS_PRICE_MULTIPLIER = 100;
+    // The specific transaction bytes that are used to deploy the Foundry Deterministic Deployer contract
+    public static final byte[] FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION;
+
+    static {
+        try {
+            FOUNDRY_DETERMINISTIC_DEPLOYER_TRANSACTION = Hex.decodeHex(
```
david-bakin-sl, line 70: Can you use Java's own `HexFormat` rather than Apache's `Hex`?  (I think we're going to eventually try to get rid of the latter's library, or so I've heard.) (`HexFormat.parseHex` also doesn't throw a checked exception, which is nice.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCreateSuite.java
```diff
@@ -124,6 +125,11 @@ public class ContractCreateSuite extends HapiSuite {
     private static final String PATTERN = "0.0.%d";
 
     private static final Logger log = LogManager.getLogger(ContractCreateSuite.class);
+    private static final String FOUNDRY_DEPLOYMENT_SIGNER = "3fab184622dc19b6109349b94811493bf2a45362";
+    private static final String FOUNDRY_DEPLOYMENT_TRANSACTION =
+            "f8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222";
```
david-bakin-sl, line 130: Same comment pointing off to `deterministic-deployment-proxy` here please.

david-bakin-sl (approved):
Suggestions for improvement, but will approve anyway if you want to fix it later (or never).

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
lukelee-sl: will fix in the next release.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
lukelee-sl: Hex was already used in the class so just used what was available without importing anything new.  Will use HexFormat in a future pr

PR #12720, opened 2024-04-11
fix: Add test cases for TokenInfo single field update via contract calls and update handler logic to match the wanted behaviour

### LINKED ISSUE #12455: Implement negative test cases for tokenUpdate

stoyanov-st:
**Description**:
Added `HAPI` tests to verify that we can update single field only for a token via contract call.
Extended some validations in `TokenUpdateHandler` to cover the flows also via contract calls.
Regarding the `memo` field - To be consistent with the mono behaviour, we don't want to update the value of the field if it is empty string thus resulting in disabling a test `tokenUpdateCanClearMemo()` that is simulating that.

**Related issue(s)**:

Fixes #12455 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerTest.java
+++ b/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenUpdateHandlerTest.java
@@ -1095,6 +1095,40 @@ class TokenUpdateHandlerTest extends CryptoTokenHandlerTestBase {
                 .has(responseCode(CURRENT_TREASURY_STILL_OWNS_NFTS));
     }
 
+    @Test
+    void validateZeroTreasuryNotUpdatedForContractCalls() {
+        // We don't set transaction ID so we simulate a contract call
+        txn = TransactionBody.newBuilder()
+                .tokenUpdate(TokenUpdateTransactionBody.newBuilder()
+                        .token(fungibleTokenId)
+                        .treasury(zeroAccountId)
+                        .name("TestTokenUpdateDoesNotUpdateZeroTreasury")
+                        .build())
+                .build();
+        given(handleContext.body()).willReturn(txn);
+        assertThatNoException().isThrownBy(() -> subject.handle(handleContext));
+
+        final var token = writableTokenState.get(fungibleTokenId);
+        assertThat(token.treasuryAccountId()).isEqualTo(fungibleToken.treasuryAccountId());
+        assertThat(token.treasuryAccountId()).isNotEqualTo(zeroAccountId);
+        assertThat(token.name()).isEqualTo("TestTokenUpdateDoesNotUpdateZeroTreasury");
+    }
+
+    @Test
+    void validateZeroTreasuryIsUpdatedForHapiCalls() {
+        txn = new TokenUpdateBuilder()
+                .withToken(fungibleTokenId)
+                .withTreasury(zeroAccountId)
+                .withName("TestTokenUpdateDoesUpdateZeroTreasury")
+                .build();
+        given(handleContext.body()).willReturn(txn);
+        assertThatNoException().isThrownBy(() -> subject.handle(handleContext));
+
+        final var token = writableTokenState.get(fungibleTokenId);
+        assertThat(token.treasuryAccountId()).isEqualTo(zeroAccountId);
+        assertThat(token.name()).isEqualTo("TestTokenUpdateDoesUpdateZeroTreasury");
+    }
+
     /* --------------------------------- Helpers --------------------------------- */
     /**
      * A builder for {@link com.hedera.hapi.node.transaction.TransactionBody} instances.
--- a/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/util/CryptoTokenHandlerTestBase.java
+++ b/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/util/CryptoTokenHandlerTestBase.java
@@ -152,6 +152,8 @@ public class CryptoTokenHandlerTestBase extends StateBuilderUtil {
     protected final AccountID feeCollectorId = transferAccountId;
     protected final AccountID stakingRewardId =
             AccountID.newBuilder().accountNum(800).build();
+    protected final AccountID zeroAccountId =
+            AccountID.newBuilder().accountNum(0).build();
 
     /* ---------- Account Numbers ---------- */
     protected final Long accountNum = payerId.accountNum();
@@ -365,6 +367,7 @@ public class CryptoTokenHandlerTestBase extends StateBuilderUtil {
     protected Account stakingRewardAccount;
     protected Account tokenReceiverAccount;
     protected Account hbarReceiverAccount;
+    protected Account zeroAccount;
 
     /* ---------- Maps for updating both readable and writable stores ---------- */
     private Map<AccountID, Account> accountsMap;
@@ -415,6 +418,7 @@ public class CryptoTokenHandlerTestBase extends StateBuilderUtil {
         accountsMap.put(delegatingSpenderId, delegatingSpenderAccount);
         accountsMap.put(spenderId, spenderAccount);
         accountsMap.put(treasuryId, treasuryAccount);
+        accountsMap.put(zeroAccountId, zeroAccount);
         accountsMap.put(stakingRewardId, stakingRewardAccount);
 
         tokensMap = new HashMap<>();
@@ -790,6 +794,10 @@ public class CryptoTokenHandlerTestBase extends StateBuilderUtil {
                 .tinybarBalance(Long.MAX_VALUE)
                 .headNftId((NftID) null)
                 .headNftSerialNumber(0L)
+                .build();
+        zeroAccount = givenValidAccountBuilder()
+                .accountId(zeroAccountId)
+                .key(EMPTY_KEYLIST)
                 .build();
     }
 


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java
```diff
@@ -317,4 +327,248 @@ public HapiSpec getTokenKeyForNonFungibleNegative() {
...
+                                .signingWith(MULTI_KEY)
+                                .payingWith(ACCOUNT)
+                                .gas(GAS_TO_OFFER)
+                                .hasKnownStatus(SUCCESS),
+                        getTokenInfo(TOKEN)
+                                .logged()
+                                .hasTokenType(TokenType.FUNGIBLE_COMMON)
+                                .hasSymbol("SYM1")
+                                .hasName("New Name")
+                                .hasEntityMemo("New Memo")
```
MrValioBg, line 543: nit: name, symbol, entityMemo are repetitive, consider extracting them into constants.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -112,7 +112,7 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         if (op.hasAutoRenewAccount()) {
             context.requireKeyOrThrow(op.autoRenewAccountOrThrow(), INVALID_AUTORENEW_ACCOUNT);
         }
-        if (op.hasTreasury()) {
+        if (op.hasTreasury() && !isZeroAccount(op.treasury())) {
```
tinker-michaelj, line 115: I think we should only use these special cases **if** the `TokenUpdate` has been dispatched from a contract call.

See [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java#L336) for an example in `CryptoCreateHandler`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -112,7 +112,9 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         if (op.hasAutoRenewAccount()) {
             context.requireKeyOrThrow(op.autoRenewAccountOrThrow(), INVALID_AUTORENEW_ACCOUNT);
         }
-        if (op.hasTreasury()) {
+        // If the transactionID is missing we have contract call,
+        // so we need to verify that the provided treasury is not a zero account.
+        if (op.hasTreasury() && (context.body().hasTransactionID() || !isZeroAccount(op.treasury()))) {
```
david-bakin-sl, line 117: This test, which also appears down ~L171, is tricky in and of itself (nested opposite boolean operations), and deserves to be pulled out to a method and given a name (and maybe not be a one-liner: maybe the RHS of the `&&` _also_ needs a name).  

Also something similar but different is happening down around L364 - together there's a bit of interesting semantics going on in these rules - and these rules should be pulled out and given names so that we can identify the commonalities and differences and later we might discover other places where one or the other rule is appropriate to use.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -166,7 +168,7 @@ public void handle(@NonNull final HandleContext context) throws HandleException
         // enabled and has open slots. If so, auto-associate.
```
david-bakin-sl, line 168: It's probably not the time now, but in the future we should consider baking this comment into the code by giving these separate checks _names_ so that we can just read the names and know what checks are being performed - rather than having to map these various conditions for auto-association onto the confusing mass of code below.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -342,20 +345,25 @@ private Token.Builder customizeToken(
...
+            final Token originalToken,
+            final boolean isHapiCall) {
         if (op.symbol() != null && op.symbol().length() > 0) {
             builder.symbol(op.symbol());
         }
         if (op.name() != null && op.name().length() > 0) {
             builder.name(op.name());
         }
-        if (op.hasMemo()) {
+        if (op.hasMemo() && op.memo().length() > 0) {
```
david-bakin-sl, line 358: It is utterly stupid that Java distinguishes between `null` (i.e., no reference) and `empty()` (i.e., 0-length) in the case of strings.  And in a language for which one of the main selling points was (and is!) that "there are no pointers!".

It's also annoying that protoc doesn't generate methods that hide this from the user.  (And now pbj too.)

Although - take a look at the generated code in `TokenUpdateTransactionBody.Builder` - isn't it the case that the _setters_ there fix up this difference?  (That may be a new pbj innovation, I don't remember.)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -342,20 +345,25 @@ private Token.Builder customizeToken(
...
             builder.memo(op.memo());
         }
         if (op.hasMetadata()) {
             builder.metadata(op.metadata());
         }
-        if (op.hasTreasury() && !op.treasuryOrThrow().equals(originalToken.treasuryAccountId())) {
+        if (op.hasTreasury()
+                && (isHapiCall || !isZeroAccount(op.treasuryOrThrow()))
+                && !op.treasuryOrThrow().equals(originalToken.treasuryAccountId())) {
```
david-bakin-sl, line 366: If you can figure out this test (L364-L366) then [_you're a better man than I am, Gunga Din_](https://en.wikipedia.org/wiki/Gunga_Din).

Basic form is: `a && (b || !c) && (!d)` - uh huh.  Oh, and it _requires_ short-circuiting too.

Can't this be fixed up so even I can understand it?  Not just by adding a comment, though that is probably necessary, but by breaking it up and adding meaningful names to subexpressions?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenUpdateSpecs.java
```diff
@@ -664,7 +664,8 @@ public HapiSpec updateTokenTreasuryRequiresZeroTokenBalance() {
                         .hasKnownStatus(TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES));
     }
 
-    @HapiTest
+    // TODO: To be consistent with the mono behaviour we shouldn't change the memo when we provide empty memo value.
```
david-bakin-sl, line 667: When will this TODO be addressed?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java
```diff
@@ -89,11 +98,13 @@ public class TokenUpdatePrecompileSuite extends HapiSuite {
...
     private static final String ACCOUNT_TO_ASSOCIATE_KEY = "associateKey";
+    private static final String AUTO_RENEW_ACCOUNT = "autoRenewAccount";
     public static final String CUSTOM_NAME = "customName";
     public static final String CUSTOM_SYMBOL = "Ω";
     public static final String CUSTOM_MEMO = "Omega";
     private static final long ADMIN_KEY_TYPE = 1L;
     private static final long SUPPLY_KEY_TYPE = 16L;
+    private static final KeyShape KEY_SHAPE = KeyShape.threshOf(1, ED25519, CONTRACT);
 
     public static void main(String... args) {
```
david-bakin-sl, line 109: Another suite with only negative tests and no happy path tests.

david-bakin-sl (commented):
IMO stuff that goes into the production code - here `TokenUpdateHandler.java` - needs to be written so that later we can _read_ and _understand_ it - especially if the situation arises where we need to read and understand it under _time pressure_.  Time spent _now_ in making the code clear will pay off.  (And, IMO also, comments are insufficient for this, in some cases, as here.)

Also: PR title + description doesn't mention that there are _semantic changes_ to _production code_ in this PR; it isn't just new test cases.  See changes to `TokenUpdateHandler.java`, mentioned above.  This will be important at merge time when the final commit message is written.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -485,4 +497,12 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
                         txnEstimateFactory)
                 .usageGiven(fromPbj(body), sigValueObj, token));
     }
+
+    private boolean checkForZeroAccountIfContractCall(boolean isContractCall, TokenUpdateTransactionBody op) {
```
david-bakin-sl, line 501: Is this named right?   First, `check...` doesn't tell you by the name whether its _true_ for a zero account if contract call, or false.  Should be something that tells you definitely like `is` or `has` or `can`, or something like that. 

But also, the test seems to be "is a contract call or else it's not zero" or alternatively, "its not zero, except if it is a contract call then it can be zero".

But then OTOH the comment at L365-367 says zero is valid _unless_ it is a contract call.

Am I reading this right?   (... and I've lost track now of what it _should_ be ...)

Comment indicates name of this method could be `canBeZeroUnlessAContract`?  And then the code needs to match that.  If the comment is correct.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -485,4 +497,12 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+        return isContractCall || !isZeroAccount(op.treasury());
+    }
+
+    private boolean isZeroAccount(@NonNull AccountID accountID) {
+        return accountID.equals(AccountID.newBuilder().accountNum(0L).build());
```
david-bakin-sl, line 506: Maybe have a private static final `ZERO_ACCOUNT_ID`.  Probably doesn't make a difference in the long run, performance-wise.  Might be easier to read.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
david-bakin-sl: I also have a question about the argument named `isContractCall`.  Because at L369 the actual argument is `isHapiCall` which I don't get why that means it's a contract call (now that I look at it again), and at L172 and L117 the actual argument is `txn.hasTransactionId()` yet the comment at L115 says that if the transaction id is _missing_ then it is a contract call, so now I'm confused.

david-bakin-sl (commented):
I'm confused.  Comments don't seem to match the code.  Could be me.  Could you check again?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
stoyanov-st: Done

david-bakin-sl (dismissed):
How about calling it `canBeZeroAccountOnlyIfHapiCall` instead?  

Otherwise, LGTM.

david-bakin-sl (approved):
LGTM!

MrValioBg (dismissed):
LGTM!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenUpdateSpecs.java
```diff
@@ -684,7 +684,8 @@ public HapiSpec updateTokenTreasuryRequiresZeroTokenBalance() {
                         .hasKnownStatus(TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES));
     }
 
-    @HapiTest
+    // To be consistent with the mono behaviour we shouldn't change the memo when we provide empty memo value.
+    // We should think of a method/function to reset the memo value.
```
lukelee-sl, line 688: should this test be reenabled and show that the memo is not updated?

kimbor: +1

stoyanov-st: We actually verify that behaviour with the `HapiTest` added in this PR.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -321,20 +325,28 @@ private Token.Builder customizeToken(
...
             builder.memo(op.memo());
         }
         if (op.hasMetadata()) {
             builder.metadata(op.metadata());
         }
-        if (op.hasTreasury() && !op.treasuryOrThrow().equals(originalToken.treasuryAccountId())) {
+        // Here we check that there is a treasury to be updated,
+        // that if the transaction is a contract call the treasury shouldn't be a zero account
+        // and that the provided treasury account is different from the current one
+        if (op.hasTreasury()
```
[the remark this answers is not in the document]
stoyanov-st: Tests added.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -590,4 +604,12 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
                         txnEstimateFactory)
                 .usageGiven(fromPbj(body), sigValueObj, token));
     }
+
+    private boolean canBeZeroAccountOnlyIfHapiCall(final boolean isHapiCall, final TokenUpdateTransactionBody op) {
```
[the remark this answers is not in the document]
stoyanov-st: Thank you for the suggestion! I've renamed the method.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/TokenUpdateSpecs.java
lukelee-sl: OK.  In that case can this test be removed?

lukelee-sl (approved):
LGTM

kimbor (approved):

PR #12734, opened 2024-04-11
feat: non-static temporary path builder

### LINKED ISSUE #12732: Create non-static temporary path builder
The current TemporaryFileBuilder utility is static and stateful. We should instead create a non-static implementation that lives in the platform context.

cody-littley:
closes #12732

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/utility/TemporaryPathBuilder.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/utility/TemporaryPathBuilder.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/utility/TemporaryPathBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+     * @param temporaryDataDirectory the directory where temporary files will be created. If there are files already in
+     *                               this location they will be deleted. Directory will be created in this location if
+     *                               it does not exist.
+     * @throws IOException if there is a problem setting up the temporary directory
+     */
+    public TemporaryPathBuilder(@NonNull final Path temporaryDataDirectory) throws IOException {
+        this.temporaryDataDirectory = Objects.requireNonNull(temporaryDataDirectory);
+        if (exists(temporaryDataDirectory)) {
+            logger.info(STARTUP.getMarker(), "Deleting temporary data directory {}", temporaryDataDirectory);
+            FileUtils.deleteDirectory(temporaryDataDirectory);
```
alittley, line 54: if we gave all temp files a common prefix, we could verify that only temporary files are deleted, and protect against file system disaster.

Just a thought, since classes like this are a bit scary- up to you

edward-swirldslabs (approved):

lpetrovic05 (approved):

PR #12736, opened 2024-04-11
feat: HIP-540 full implementation

petreze:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyUtils.java
```diff
@@ -33,6 +33,10 @@ public class KeyUtils {
 
     public static final Key IMMUTABILITY_SENTINEL_KEY =
             Key.newBuilder().keyList(KeyList.DEFAULT).build();
+    public static final Key ALL_ZEROS_INVALID_KEY = Key.newBuilder()
+            .ecdsaSecp256k1(Bytes.fromHex("0000000000000000000000000000000000000000"))
```
tinker-michaelj, line 37: A compressed ECDSA key [must begin ](https://crypto.stackexchange.com/questions/108091/how-to-determine-the-prefix-of-a-secp256k1-compressed-public-key) with `0x02` or `0x03`; so this key isn't even in principle valid. 

An all-zero ED25519 key, however, **is** valid in principle; but not in practice, since there's no way to find the corresponding private key.

I suggest we use an all-zero ED25519 key here; or make this a valid ECDSA key by prefixing `0x02`.

hedera-node/hedera-app/src/xtest/java/token/update/TokenUpdateRemoveKeysXTest.java
```diff
@@ -0,0 +1,151 @@
...
+        // remove admin key
+        handleAndCommitSingleTransaction(
+                component.tokenUpdateHandler(),
+                tokenUpdate(idOfNamedToken(TOKEN_REMOVE_KEYS_ID), List.of(b -> b.adminKey(IMMUTABILITY_SENTINEL_KEY))),
+                OK);
+    }
+
+    @Override
+    protected void assertExpectedNfts(@NonNull ReadableKVState<NftID, Nft> nfts) {
+        // Here is where we can e.g. assert the NFT has been transferred to the counterparty
```
tinker-michaelj, line 100: We can delete these `assertExpectedNfts()` and `assertExpectedTokenRelations()` overrides since they are still no-ops, as in the superclass implementation.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -423,22 +424,47 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
-        if (!op.hasMetadata()) {
-            return false;
-        }
-        final var defaultWithMetadata = TokenUpdateTransactionBody.newBuilder()
-                .metadata(op.metadata())
-                .token(op.token())
-                .build();
-        return op.equals(defaultWithMetadata);
+    public static boolean containsKeyRemoval(@NonNull final TokenUpdateTransactionBody op) {
+        return op.hasAdminKey() && isKeyRemoval(op.adminKey())
```
tinker-michaelj, line 431: If we make `isKeyRemoval(Key)` return false when it receives `null`, we can remove all these `op.has<X>Key()` checks and this becomes more readable.

tinker-michaelj (changes requested):
I would prefer that we encapsulate the logic needed to manage each non-admin key in an enum value like below:

```
public enum NonAdminTokenKey {
    KYC_KEY {
        @Override
        public boolean isPresent(TokenUpdateTransactionBody update) {
            return update.hasKycKey();
        }

        @Override
        public Key getFrom(TokenUpdateTransactionBody update) {
            return update.kycKeyOrThrow();
        }

        @Override
        public void setOn(TokenUpdateTransactionBody.Builder updateBuilder, Key key) {
            updateBuilder.kycKey(key);
        }

        @Override
        ResponseCodeEnum invalidStatus() {
            return ResponseCodeEnum.INVALID_KYC_KEY;
        }
    };

    abstract public Key getFrom(TokenUpdateTransactionBody update);
    abstract public boolean isPresent(TokenUpdateTransactionBody update);
    abstract public void setOn(TokenUpdateTransactionBody.Builder updateBuilder, Key key);
    abstract ResponseCodeEnum invalidStatus();

    public void clearFrom(TokenUpdateTransactionBody.Builder updateBuilder) {
        setOn(updateBuilder, null);
    }
    public void assertValidOrMissing(TokenUpdateTransactionBody update) throws PreCheckException {
        if (isPresent(update)) {
            validateTruePreCheck(isValid(getFrom(update)), invalidStatus());
        }
    }
}
```

Then e.g. validating that all present non-admin keys are valid is simply,
```
        for (final var nonAdminTokenKey : NON_ADMIN_TOKEN_KEYS) {
            nonAdminTokenKey.assertValidOrMissing(op);
        }
```
...and almost all other repeated tasks in the handler will similarly convert into `for` or `forEach()` patterns.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyUtils.java
petreze: Agreed, ecdsa builder was used by mistake. Switched to all-zero ed25519 curve key

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -423,22 +424,47 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
-        if (!op.hasMetadata()) {
-            return false;
-        }
-        final var defaultWithMetadata = TokenUpdateTransactionBody.newBuilder()
-                .metadata(op.metadata())
-                .token(op.token())
-                .build();
-        return op.equals(defaultWithMetadata);
+    public static boolean containsKeyRemoval(@NonNull final TokenUpdateTransactionBody op) {
+        return isKeyRemoval(op.adminKey())
```
tinker-michaelj, line 431: We'd like this to be,
```
        for (final var tokenKey : TOKEN_KEYS) {
            if (isKeyRemoval(tokenKey.getFromUpdate(op))) {
                return true;
            }
        }
        return false;
```
As a main benefit of enumerating the keys in one place is that we can eliminate the risk of forgetting a key elsewhere.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -380,70 +362,145 @@ private void updateExpiryFields(
...
+                                    .build())
+                            .build();
+                }
+
+                context.requireKey(requiredKey);
+            } else if (!isExpiryOnlyUpdateOp(op)) {
+                // if we update any other field, we need the current admin key to sign
+                // note: for expiry only op admin key is not required
+                context.requireKey(originalToken.adminKey());
+            }
```
Neeharika-Sompalli, line 449: What happens if Token doesn't have adminKey and its not `expiryOnlyUpdateOp`. Do we still throw somewhere ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -380,70 +362,145 @@ private void updateExpiryFields(
...
-        if (op.hasAdminKey()) {
-            final var newAdminKey = op.adminKey();
-            if (isKeyRemoval(newAdminKey)) {
-                builder.adminKey((Key) null);
-            } else {
-                builder.adminKey(newAdminKey);
+    private List<Key> getAllRequiredLowPrioritySigners(
+            @NonNull final Token originalToken, @NonNull final TokenUpdateTransactionBody op) {
+        List<Key> lowPriorityKeys = new ArrayList<>();
+        addLowPrioritySigner(lowPriorityKeys, originalToken.hasWipeKey(), originalToken.wipeKey(), op.hasWipeKey());
```
Neeharika-Sompalli, line 462: hit : You don't need `originalToken.hasWipeKey()` parameter passed, instead check `originalToken.wipeKey() == null`

Neeharika-Sompalli: Also for this you could use `isPresentInitially` and `isPresentInUpdate` in a for loop for all `TokenKeys` and add to `lowPriorityKeys` . That will be concise and more readable I think.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenKeysHelper.java
```diff
@@ -0,0 +1,233 @@
...
+        private static Key getNewKeyValue(Key newKey) {
+            return isKeyRemoval(newKey) ? null : newKey;
+        }
+
+        public void updateKey(TokenUpdateTransactionBody update, Token originalToken, final Token.Builder builder) {
+            if (isPresentInUpdate(update)) {
+                validateTrue(isPresentInitially(originalToken), tokenHasNoKeyStatus());
+                setOn(builder, update);
+            }
+        }
```
Neeharika-Sompalli, line 231: You could also add `isValid` method here. So, we can validate all keys

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenUpdateValidator.java
```diff
@@ -56,12 +56,13 @@ public ValidationResult validateSemantics(
...
-            validateTrue(token.hasAdminKey() || token.hasMetadataKey(), TOKEN_IS_IMMUTABLE);
-        } else if (!isExpiryOnlyUpdateOp(op)) {
-            validateFalse(isEmpty(token.adminKey()), TOKEN_IS_IMMUTABLE);
+        // If the token has an empty admin key it can't be updated for any other fields other than low priority keys or
+        // expiry
+        // For updating only low priority keys - (wipeKey, kycKey, supplyKey, freezeKey, feeScheduleKey, pauseKey or
+        // metadataKey)
+        // the transaction should have admin key or the respective low priority key signature
+        if (isEmpty(token.adminKey())) {
+            validateTrue(isExpiryOnlyUpdateOp(op) || hasAlreadySomeNonAdminKeys(token), TOKEN_IS_IMMUTABLE);
```
Neeharika-Sompalli, line 65: How could we be sure that the key we are trying to modify is present using `hasAlreadySomeNonAdminKeys` ?

I think this should be taking the transaction body and validating the key we are trying to modify is present, otherwise throw `TOKEN_IS_IMMMUTABLE`

netopyr (changes requested):
I think @Neeharika-Sompalli discovered a security hole that has to be fixed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
petreze: Applied

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/validation/AttributeValidator.java
```diff
@@ -82,8 +82,10 @@ static boolean isImmutableKey(@NonNull Key key) {
      * @param source the key to check
      * @return true if the key is a key removal, false otherwise
      */
-    static boolean isKeyRemoval(@NonNull final Key source) {
-        requireNonNull(source);
+    static boolean isKeyRemoval(final Key source) {
```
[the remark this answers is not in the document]
petreze: It was changed after [this discussion](https://github.com/hashgraph/hedera-services/pull/12736#discussion_r1569454866) due to more readability and include it in `TokenKeys` util class. We return false if the input is null

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -423,22 +424,47 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
-    public static boolean isMetadataOnlyUpdateOp(@NonNull final TokenUpdateTransactionBody op) {
-        if (!op.hasMetadata()) {
-            return false;
-        }
-        final var defaultWithMetadata = TokenUpdateTransactionBody.newBuilder()
-                .metadata(op.metadata())
-                .token(op.token())
-                .build();
-        return op.equals(defaultWithMetadata);
+    public static boolean containsKeyRemoval(@NonNull final TokenUpdateTransactionBody op) {
```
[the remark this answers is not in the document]
petreze: Agreed, tyvm!

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/validation/AttributeValidator.java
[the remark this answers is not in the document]
netopyr: I am not sure I can follow. It looks like `null` is a valid argument for `source`. In such cases, we add the annotation `@Nullable`.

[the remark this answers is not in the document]
netopyr: I did not mean to change the logic here. Just add the annotation, please.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
petreze: @Neeharika-Sompalli 
I believe updating **only** expiry does not need any signatures, so it should succeed or my assumptions is not quite true?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -380,70 +362,145 @@ private void updateExpiryFields(
...
+            @NonNull PreHandleContext context,
+            @NonNull final TokenUpdateTransactionBody op,
+            @NonNull final Token originalToken)
+            throws PreCheckException {
+        if (op.hasAdminKey()) {
+            if (isValid(op.adminKey())) {
+                // if op admin key is valid we require the signature
+                context.requireKey(op.adminKeyOrThrow());
+            }
+            if (originalToken.hasAdminKey()) {
```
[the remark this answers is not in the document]
petreze: Good catch, this remained after we combined the logic for the entire HIP in one branch, ty

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenKeysHelper.java
petreze: @Neeharika-Sompalli 
That was our initial idea but method `validateTokenKeys()` is called from both `TokenCreateValidator` and `TokenUpdateValidator` which means we have 2 different operation structures and cannot unify them with the same approach..

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -423,22 +428,34 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
+                || op.hasMemo()
+                || op.hasExpiry());
+    }
+
+    /**
+     * Check if a given token already has some of the low priority keys
+     */
+    public static boolean hasAlreadySomeNonAdminKeys(@NonNull final Token token) {
+        // here we want to remove the admin key case because we need only low priority keys
+        final var NonAdminKeys =
```
Neeharika-Sompalli, line 451:
```suggestion
        final var nonAdminKeys =
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -423,22 +428,34 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
+                || op.hasMemo()
+                || op.hasExpiry());
+    }
+
+    /**
+     * Check if a given token already has some of the low priority keys
+     */
+    public static boolean hasAlreadySomeNonAdminKeys(@NonNull final Token token) {
+        // here we want to remove the admin key case because we need only low priority keys
+        final var nonAdminKeys =
```
tinker-michaelj, line 451: Instead of re-creating this every time, I think just define above:
```
private static final Set<TokenKeys> NON_ADMIN_TOKEN_KEYS = EnumSet.complementOf(EnumSet.of(TokenKeys.ADMIN_KEY));
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
Neeharika-Sompalli: No I mean, if it is **not** expiry only op and there is **no** `adminKey`

petreze: @Neeharika-Sompalli 
In this case:
- if the token has no low priority key set it will result in `TOKEN_IS_IMMUTABLE`
- otherwise check required signatures

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -42,19 +42,24 @@
...
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 public class BaseTokenHandler {
     private static final Logger log = LogManager.getLogger(BaseTokenHandler.class);
+    private static final Set<TokenKeys> TOKEN_KEYS = EnumSet.allOf(TokenKeys.class);
```
Neeharika-Sompalli, line 62: nit : This could be public static

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -81,6 +79,7 @@
 @Singleton
 public class TokenUpdateHandler extends BaseTokenHandler implements TransactionHandler {
     private final TokenUpdateValidator tokenUpdateValidator;
+    private static final Set<TokenKeys> TOKEN_KEYS = EnumSet.allOf(TokenKeys.class);
```
tinker-michaelj, line 82: We can just make the parent `Set<TokenKeys>` fields `protected` and delete this one.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -398,49 +370,132 @@ private void updateExpiryFields(
...
+                if (!isKeyRemoval(newRoleKey)) {
+                    if (op.keyVerificationMode() == NO_VALIDATION) {
+                        requireAdminOrRole(context, token, tokenKey);
+                    } else {
+                        // With "full" verification mode, our required key
+                        // structure is a 1/2 threshold with components:
+                        //   - Admin key
+                        //   - A 2/2 list including the role key and its replacement key
+                        final var key = tokenKey.getFromUpdate(op);
+                        requireAdminOrRole(context, token, tokenKey, key);
```
Neeharika-Sompalli, line 437: WE are using 2/2 list just to follow the pattern right ? Its basically like needs (admin or role key ) and replacement key right ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -398,49 +370,132 @@ private void updateExpiryFields(
...
+            @NonNull final Token token,
+            @NonNull final TokenKey roleKey,
+            @Nullable final Key replacementKey)
+            throws PreCheckException {
+        final var maybeRoleKey = roleKey.getFromToken(token);
+        // Prioritize TOKEN_IS_IMMUTABLE for completely immutable tokens
+        mustExist(maybeRoleKey, token.hasAdminKey() ? roleKey.tokenHasNoKeyStatus() : TOKEN_IS_IMMUTABLE);
+        if (token.hasAdminKey()) {
+            context.requireKey(oneOf(
+                    replacementKey == null ? maybeRoleKey : allOf(maybeRoleKey, replacementKey),
```
Neeharika-Sompalli, line 461: If there is a replacement Key, we need `replacementKey` and oneof(adminKey, roleKey) right ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -398,49 +370,132 @@ private void updateExpiryFields(
...
+                    token.adminKeyOrThrow()));
+        } else {
+            context.requireKey(maybeRoleKey);
+            if (replacementKey != null) {
+                context.requireKey(replacementKey);
+            }
```
Neeharika-Sompalli, line 467: If the replacementKey is required to sign and its not valid, it will not be added to the requiredNonPayerKeys. We should validate and throw here ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
tinker-michaelj: We require _either_,
1. The `adminKey` _or_,
2. **Both** the `existingRoleKey` and `replacementRoleKey`

tinker-michaelj: The idea is that whenever the `adminKey` signs, no other key is required---because the `adminKey` has control to change those "lower keys" to whatever it wants.

Neeharika-Sompalli: But how do we know the `replacementRoleKey` is valid ?

tinker-michaelj (approved):
LGTM, tyvm @petreze !

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
Neeharika-Sompalli: Ah, yes we validate that in `pureChecks`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
tinker-michaelj: Done in `pureChecks()`.

Neeharika-Sompalli: We validate in `pureChecks`. Nvmd

Neeharika-Sompalli (approved):
LGTM! Thanks @petreze

netopyr (approved):

PR #12746, opened 2024-04-11
refactor: Delete unneeded GossipEvent methods

### LINKED ISSUE #12745: Delete `GossipEvent` `equals` and `hashCode`
- Currently, `GossipEvent::equals` and `GossipEvent::hashCode` only consider the contents of `BaseEventHashedData`, _not_ the signature or unhashed data
- This is counterintuitive, since the signature in particular is a critical part of the object, but isn't being considered for equality
    - This constitutes a liability, since a programmer could accidentally use these methods in the wrong way
- `GossipEvent::equals` isn't used anywhere, nor do we need to use `GossipEvent::hashCode`

With all this in mind, we should just delete these methods.

alittley:
- closes #12745


alittley:
Too invasive for now. Will be fixed during event refactor effort.

PR #12751, opened 2024-04-11
feat: improve performance of tlsPeerIdentifier to O(1)

### LINKED ISSUE #12709: feat: improve performance of tlsPeerIdentifier to O(1)

kfa-aguda:
closes #12709. 

This PR improves the certificate lookup performance of a newly connected network peer.

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkPeerIdentifier.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkPeerIdentifier.java
@@ -16,7 +16,6 @@
 
 package com.swirlds.platform.network;
 
-import static com.swirlds.logging.legacy.LogMarker.EXCEPTION;
 import static com.swirlds.logging.legacy.LogMarker.SOCKET_EXCEPTIONS;
 
 import com.swirlds.common.context.PlatformContext;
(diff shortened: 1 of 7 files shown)


lpetrovic05:
@kfa-aguda @cody-littley 
This approach makes dynamic changes to the peer list more difficult. We should probably discuss an alternative approach

cody-littley:
@lpetrovic05 

> This approach makes dynamic changes to the peer list more difficult. We should probably discuss an alternative approach

When we want things to be dynamic, I'm guessing we are going to need a method that says "hey, use this new peer list".

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkPeerIdentifier.java
```diff
@@ -46,52 +45,49 @@ public class NetworkPeerIdentifier {
      */
     private final RateLimitedLogger noPeerFoundLogger;
 
-    private final RateLimitedLogger multiplePeersFoundLogger;
+    // a mapping of X500Principal and their peers
```
alittley, line 48: Member docs should use /* */ comment style (only need be addressed in this PR if there are other needed changes)

kfa-aguda: thank you. Will take care of it in a subsequent PR.

poulok:
> When we want things to be dynamic, I'm guessing we are going to need a method that says "hey, use this new peer list".

Why not just create the peer info as a Map in the first place and pass it into the method instead of the constructor?

edward-swirldslabs (approved):

kfa-aguda:
> > When we want things to be dynamic, I'm guessing we are going to need a method that says "hey, use this new peer list".
> 
> Why not just create the peer info as a Map in the first place and pass it into the method instead of the constructor?

We discussed that actually but as it's generally not the best to pass around mutable collections, I'd have to write some wrapper around the map with a get. Whilst that also works, this is simpler.

Issue #12750, opened 2024-04-11
Allow only `CryptoTransfer` to use aliased accountIds


Neeharika-Sompalli:
In mod-service since we look up aliased accountIds, we are supporting aliases out-of-the-box in all transactions. But, since we were not supporting aliases everywhere in mono-service, we would like to do the same for the initial release of modular codebase.

PR #12754, opened 2024-04-11
fix: Allow aliases only in CryptoTransfer

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/12750
Fixes https://github.com/hashgraph/hedera-services/issues/12776

- Adds a new function in `ReadableAccountStore.getAliasedAccountId()` that looks up accountId's that have aliases set
- The `ReadableAccountStore.getAccountId()` and `ReadableAccountStore.get()` don't look up accountIds that have aliases set
- Used `ReadableAccountStore.getAliasedAccountId()` in `CryptoTransferHandler` and `CryptoGetAccountInfo` and `GetAccountDetails` to permit use of aliases

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/ReadableAccountStoreImpl.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/ReadableAccountStoreImpl.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.token.impl;
 
+import static com.hedera.hapi.node.base.AccountID.AccountOneOfType.ACCOUNT_NUM;
 import static com.hedera.node.app.service.token.AliasUtils.asKeyFromAliasOrElse;
 import static com.hedera.node.app.service.token.AliasUtils.extractEvmAddress;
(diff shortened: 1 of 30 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/ReadableAccountStoreImpl.java
```diff
@@ -133,7 +135,37 @@ public boolean contains(@NonNull final AccountID accountID) {
     protected Account getAccountLeaf(@NonNull final AccountID id) {
         // The Account ID may be aliased, in which case we need to convert it to a number-based account ID first.
```
iwsimon, line 136: Remove this comment as it has changed.

Neeharika-Sompalli: Yes Iris. Thanks Will update in next PR.

iwsimon (approved):
LGTM. Only small comment change in next PR. Thanks @Neeharika-Sompalli

povolev15 (approved):
LGTM Thanks @Neeharika-Sompalli

tinker-michaelj (approved):

PR #12757, opened 2024-04-11
feat: 12108: Virtual root nodes may be synced in parallel during reconnects

### LINKED ISSUE #12108: Virtual root nodes may be synced in parallel during reconnects
Current reconnect implementation is single threaded. The teacher traverses the merkle tree, sends nodes to the learner, collects information about children. If some node is a custom reconnect root, it's handled separately. The key issue here is that all custom reconnect roots (like `VirtualRootNode`) are processes sequentially, one by one:

```java
            while (!subtrees.isEmpty()) {
                try (final TeacherSubtree subtree = subtrees.remove()) {
                    subtree.getView().waitUntilReady();
                    sendTree(subtree.getRoot(), subtree.getView());
                }
            }
```

There is no real reason to do it this way, though. Multiple virtual root nodes may be synced in parallel. Disk utilization would be much better. There is a problem, though. Today there is a single connection / single output stream / single input stream for every reconnect (teacher, learner) pair. Every request / response has implicit context (reconnect root). If multiple reconnect roots are processed in parallel, every request / response must contain information, which root it is about.

artemananiev:
Fixes: https://github.com/hashgraph/hedera-services/issues/12108


artemananiev:
This PR is obsolete, will never be merged

PR #12764, opened 2024-04-12
fix: (cherry-pick) Allow aliases only in CryptoTransfer

Neeharika-Sompalli:
Cherry-pick https://github.com/hashgraph/hedera-services/pull/12754


tinker-michaelj (approved):

Issue #12758, opened 2024-04-11
Augment `PureChecks` for TokenAccountWipeHandler, TokenAssociateToAccountHandler, and TokenBurnHandler


mhess-swl:
Add the following pure checks:
* ~~TokenAccountWipeHandler:~~
  – ~~if tokenID is not null, tokenID is valid~~
  – ~~if accountID is not null, accountID is valid~~
* ~~TokenAssociateToAccountHandler:~~
  – ~~if accountID is not null, accountID is valid~~
* TokenBurnHandler:
  – ~~if tokenID is not null, tokenID is valid~~
  – serialNums is not null

mhess-swl:
EDIT: per [this](https://github.com/hashgraph/hedera-services/issues/12711#issuecomment-2059587684) comment, we're only going to check for null objects. Therefore, the only work required for this ticket is checking `serialNums` is not null in _TokenBurnHandler_, which doesn't need its own ticket.

Issue #12711, opened 2024-04-10
Validate and Add any missing `pureChecks()` in Crypto and Token handlers


Neeharika-Sompalli:
Validate the basic fields are set correctly in `pureChecks` in all token and crypto handlers. Add any if they are missing.

- [x] CryptoAddLiveHashHandler – nothing needed; not supported
- [x] CryptoApproveAllowanceHandler – nothing needed (see comment below about operatorId)
- [x] #12714
- [x] CryptoDeleteAllowanceHandler – ~~if nftAllowances is not null, non-null checks on all contained allowances and sub-objects~~ nothing needed
- [x] CryptoDeleteHandler – ~~if transferAccountID is not null, transferAccountID is valid; if deleteAccountID is not null, deleteAccountID is valid~~ nothing needed
- [x] CryptoDeleteLiveHashHandler – nothing needed; not supported
- [x] #12715
- [x] CryptoGetAccountInfoHandler – ~~could check that accountID is valid, but `validate()` depends on looking up the account object from state??~~ nothing needed (see comment about checking null below)
- [x] CryptoGetAccountRecordsHandler – ~~could check that accountID is valid, but `validate()` depends on looking up the account object from state??~~ nothing needed (see comment about checking null below)
- [x] CryptoGetLiveHashHandler – nothing needed; not supported
- [x] CryptoGetStakersHandler – nothing needed; not supported
- [x] #12817
- [x] CryptoUpdateHandler – nothing needed
- [x] TokenAccountWipeHandler – ~~#12758~~ nothing needed
- [x] TokenAssociateToAccountHandler – ~~issue #12758 – if accountID is not null, accountID is valid~~ nothing needed
- [x] TokenBurnHandler – ~~issue #12758 – if tokenID is not null, tokenID is valid; serialNums is not null~~ nothing needed
- [x] TokenCreateHandler – nothing needed
- [x] TokenDeleteHandler – ~~if tokenID is not null, tokenID is valid~~ nothing needed
- [x] TokenDissociateFromAccountHandler – ~~if accountID is not null, accountID is valid; if tokens is not null, all contained tokenIDs are valid~~ nothing needed
- [x] TokenFeeScheduleUpdateHandler – ~~if accountID is not null, accountID is valid;~~ if customFees is not null, all contained custom fees are valid~~ nothing needed
- [x] TokenFreezeAccountHandler – ~~if accountID is not null, accountID is valid; if tokenID is not null, tokenID is valid~~ nothing needed
[...]

PR #12774, opened 2024-04-12
fix: Add pure checks for token wipe

mhess-swl:
Part of #12758

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.token.impl.validators;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ADMIN_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CUSTOM_FEE_SCHEDULE_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_FREEZE_KEY;
@@ -23,6 +24,7 @@ import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_KYC_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_METADATA_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_PAUSE_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_SUPPLY_KEY;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TOKEN_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_WIPE_KEY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ZERO_BYTE_IN_STRING;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.METADATA_TOO_LONG;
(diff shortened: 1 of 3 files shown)


thomas-swirlds-labs (dismissed):
LGTM! Thanks @mhess-swl 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -57,6 +64,36 @@ public TokenAttributesValidator() {
...
+        validateAccountID(accountID, INVALID_ACCOUNT_ID);
+    }
+
+    /**
+     * Performs a "pure" check for a given token ID. If the token ID is not in the body, is null, or is invalid, a {@link PreCheckException} is thrown.
+     */
+    public static void validateTokenId(final boolean bodyHasTokenId, @Nullable final TokenID tokenID)
+            throws PreCheckException {
+        validateTruePreCheck(bodyHasTokenId, INVALID_TOKEN_ID);
+        validateTruePreCheck(tokenID != null, INVALID_TOKEN_ID);
```
Neeharika-Sompalli, line 82: Why do we need this check again ? `bodyHasTokenID` means it is not null right ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
```diff
@@ -57,6 +64,36 @@ public TokenAttributesValidator() {
...
+        validateNullableTokenId(tokenID);
+    }
+
+    /**
+     * Performs validation checks for a token ID <b>only if</b> the tokenID is not null.
+     */
+    public static void validateNullableTokenId(@Nullable final TokenID tokenID) throws PreCheckException {
+        if (tokenID != null) {
```
Neeharika-Sompalli, line 90: Why do we need not null check ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
mhess-swl: ohhh I was thinking `hasTokenId` checked if it was the default token. I get really confused with the default token and checking for null..

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
mhess-swl: In the case where `tokenID` is null, this method won't throw any NPE's

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
mhess-swl: I removed the extra check

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAttributesValidator.java
Neeharika-Sompalli: I see. Are we using `validateNullableTokenId` outside `validateTokenId` ? If so, we can keep that. Otherwise we can make this private and make it @NonNull . What do you think ?

mhess-swl: I removed the validateNullable... method entirely. It's not necessary.

mhess-swl:
Closing per [this](https://github.com/hashgraph/hedera-services/issues/12711#issuecomment-2059587684) comment. This work isn't needed.

PR #12786, opened 2024-04-15
docs: add design document for HIP-904 token reject operation

MiroslavGatsanoga:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/docs/design/services/token-service/airdrops/token-reject.md
+++ b/hedera-node/docs/design/services/token-service/airdrops/token-reject.md
@@ -19,7 +19,7 @@ Token reject transaction can be performed only for tokens which are part of the 
 
 ## Architecture
 
-The implementation related to the new `TokenReject` transaction will be gated behind a feature flag.
+The implementation related to the new `TokenReject` transaction will be gated behind a `tokens.reject.enabled` feature flag.
 
 ### HAPI updates
 
--- a/hedera-node/docs/design/services/token-service/airdrops/token-reject.md
+++ b/hedera-node/docs/design/services/token-service/airdrops/token-reject.md
@@ -116,6 +116,14 @@ An update into the `feeSchedule` file would be needed to specify that.
         - Custom fees are not applied and no tokens are deducted from the sender balance
         - Token dissociate fees are not applied
 
+### Token allowances
+An owner can provide a token allowance for non-fungible and fungible tokens. The owner is the account that owns the tokens and grants the allowance to the spender. The spender is the account that spends tokens authorized by the owner from the owners account.
+
+If `TokenReject` is executed on the owner of a token allowance, then the allowance for that token should be canceled.
+
+If `TokenReject` is executed on the spender of a token allowance, then the allowance should not be affected. It's the responsibility of the owner to clean up any unnecessary allowances (or continue to pay rent for them).
+That would mean that after performing `TokenReject` on a spender account, then the spender can still perform `TokenAirdrop` for a given allowance even though the token was rejected.
+
 ## Acceptance Tests
 
 All of the expected behaviour described below should be present only if the new `TokenReject` feature flag is enabled. No custom fees should be assessed and the tokens should not be dissociated from the account after token reject is performed.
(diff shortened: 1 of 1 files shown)


hedera-node/docs/design/services/token-service/airdrops/token-reject.md
```diff
@@ -0,0 +1,135 @@
...
+ * Reject undesired token(s).<br/>
+ * Transfer one or more token balances held by the requesting account to the treasury for each
+ * token type.<br/>
+ * Each transfer SHALL be one of the following
+ * - A single non-fungible/unique token.
+ * - The full balance held for a fungible/common token type.
+ *
+ * A single tokenReject transaction SHALL support a maximum of 10 transfers.
+ */
+message TokenRejectTransactionBody {
```
MrValioBg, line 39: Shouldn't we also have `AccountID` in the `TokenRejectTransactionBody` ?

MiroslavGatsanoga: I think the account that is rejecting the tokens should be the payer of the transaction.
@Nana-EC is this correct?

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
```diff
@@ -0,0 +1,135 @@
+# Introduce TokenReject operation
+
+## Purpose
+
+As part of [HIP-904](https://hips.hedera.com/hip/hip-904), we would like to have the option of rejecting a token that has become part of an account's balance (no matter how e.g. via regular crypto transfer or airdrop).
+
+This would mean transferring the token (the whole amount if fungible) or the NFT (concrete serial number from a unique token type) back to its treasury account, without assessing custom fees. Using token reject an account could get rid of potential “ransom” tokens which have very expensive custom fees and only pay for the cheaper token reject operation.
```
Neeharika-Sompalli, line 7: In the current code even a `CryptoTransfer` to treasury doesn't charge custom fees. 
So it seems like, the added advantage for this transaction is we can transfer back even the token is paused or token is frozen (which is not the case with `CryptoTransfer`)

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
Neeharika-Sompalli: We should probably clarify that in the javadocs.

Nana-EC: Good point.
To support a different account paying for the transaction we should not assume the account and probable have an accountID field as @MrValioBg noted.

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Implement token rejection in system contract functions
+
+## Architecture
+
+The implementation related to the new `TokenReject` transaction will be gated behind a feature flag.
+
+### HAPI updates
+
+Create new transaction type as defined in the HIP:
```
Nana-EC, line 26: Would prefer to link to the protobuf fields in a given tag to ensure any slight changes are in sync.
Can update later or in a future PR when 0.50.0 protobug tag is made

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
MrValioBg: We don't have custom fees if the payer is the treasury account. 
However, it doesn't seem like we have a case where there is a fee exemption if the payer is sending to the treasury account.

Reference [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/CustomFeeExemptions.java#L51)

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
MiroslavGatsanoga: Noted, once protobufs is merged and tagged we can add links to the definitions.

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
MiroslavGatsanoga: @Neeharika-Sompalli Given the code Valio linked above it seems like we don't charge the custom fees only if the treasury is the sender of the transfer, but for transfers in the other direction (i.e. treasury is receiver) it looks like we charge custom fees from the sender and that's the case we can use `TokenReject` to avoid those fees.

Neeharika-Sompalli: @MiroslavGatsanoga @MrValioBg yes I mean if the sender is treasury custom fees are[ not assessed.](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CustomFeeAssessmentStep.java#L377). And if receiver is treasury royalty fees are not assessed, because thats the only fee receiver will be paying in custom fees [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/customfees/CustomRoyaltyFeeAssessor.java#L98).

hedera-node/docs/design/services/token-service/airdrops/token-reject.md
Nana-EC: LG

Neeharika-Sompalli (approved):
LGTM ! Thanks @MiroslavGatsanoga

MrValioBg (approved):
LGTM! 💯

PR #12792, opened 2024-04-15
docs: add design doc for pending state

IvanKavaldzhiev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,126 @@
...
+
+    void remove(@NonNull PendingAirdropId airdropId) {
+        // invoked by the `TokenCancelAirdrop`
+    }
+}
+```
+
+### Operations
+
+1. Fetching airdrop values from the collections. This would be applicable only for fungible airdrops.
```
Nana-EC, line 78: Q: wouldn't this also be applicable when Claiming a fungible airdrops. We should call out that the get will be done in 2 transaction scenarios
```suggestion
1. Fetching airdrop values from the collections. This would be applicable only for fungible tokens (both TokenAirdrop and TokenClaimAirdrop transaction).
```

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,126 @@
...
+   For a specific `PendingAirdropId`, we would get the corresponding `PendingAirdropValue`.
+2. Handling a new airdrop in state. We can have several cases:
+   - Only a single fungible or non-fungible airdrop for a specific token defined - we just put a new entry in the map
+   - An airdrop for an existing `PendingAirdropId` is already defined - we should replace the existing `PendingAirdropValue` for this `PendingAirdropId` with the aggregated new fungible amount
+3. Handling an airdrop `cancel` in state - we remove the `PendingAirdropId` key for the airdrop we want to cancel
```
Nana-EC, line 83: Q: are you missing the "4. Handling an airdrop `claim` in state" ?

IvanKavaldzhiev: I think yes. In the case of claim - we again should remove the corresponding `PendingAirdropId` key from the state.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,126 @@
...
+
+We can have a case, where we have multiple airdrops for the same fungible token. In this case the  `PendingAirdropValue` would be updated with the new aggregated value.
+
+### Serialization
+
+In order for the pending state to be kept during node upgrades or restarts we should write it on disk with a serialization mechanism, as we do for other context or state related information. This could be done in the `InitialModServiceTokenSchema` since the pending airdrop state would be written in the already existing token state.
+
+We should define a new entry in `StateChildIndices` called `AIRDROPS`. Based on it, we would fetch a `VirtualMap` entry for the airdrops. It would include an `AirdropKey`, which would be a `PendingAirdropId` representation as a key and an `EntityNum` , which would be a `PendingAirdropValue` representation as a value. The `AirdropKey` should implement the `VirtualKey` interface, so that the key has a very strong hash algorithm to optimize performance. `EntityNum` is an already existing numeric wrapper for long values that extend the `VirtualValue` interface and provide better efficiency for usage in `VirtualMaps`.
```
Neeharika-Sompalli, line 41: As discussed today, we don't need this section. And this doc should be updated design for traversal of all pending airdrops sent by an account (using the LinkedList pattern) during expiration or deletion

IvanKavaldzhiev: I adapted the doc following your comments.

Neeharika-Sompalli (dismissed):
LGTM!

Nana-EC (approved):

Nana-EC (approved):
LG, let's get 2 services team member sign off since it's state related.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,122 @@
...
+	PendingAirdropValue get(@NonNull final PendingAirdropId airdropId);
+	long sizeOfState();
+	void warm(@NonNull final PendingAirdropId airdropId);
+    List<PendingAirdropId> getPendingAirdropIdsBySender(final AccountId sender);
+}
+```
+
+The `ReadableAirdropStoreImpl` would keep an instance of `ReadableKVState<PendingAirdropId, PendingAirdropValue> airdropsState`.
+
+An additional method `getPendingAirdropIdsBySender` would be added to the `ReadableAirdropStore` to get all PendingAirdropIds for a specific sender. This would be useful for traversing all sender's airdrops when their account expires or gets deleted. Then we should iterate over all of their pending airdrops and cancel them.
```
netopyr, line 54: Can you please expand on how this will be implemented? I am afraid that this will not be possible with the State definition above alone.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Stores
+
+We need to define a special `WritableStore` for the airdrops, so that we can manipulate the content in them. This would be the `WritableAirdropStore`.
+
+It would extend `ReadableAirdropStoreImpl` that implements the following interface:
+
+```java
+public interface ReadableAirdropStore {
+	PendingAirdropValue get(@NonNull final PendingAirdropId airdropId);
```
netopyr, line 45: What would this method return in the case of an NFT?

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,122 @@
...
+## Goals
+
+1. Define state structure where to keep the information for incoming airdrops, so that they could be canceled or claimed
+2. Define what state transitions inside the pending state would be allowed in the scope of airdrop operations and how changes would be processed
+3. Describe how this new state would be serialized and kept after restarts
+4. Define Readable and Writable stores for accessing and manipulating the pending airdrops state
+
+## Architecture
+
+The newly defined entity state won’t be able to persist anything in the state directly. If we want to commit something from the pending state to the persisted state, we would have to delegate a given pending airdrop by creating a synthetic crypto transfer transaction, send it to a suitable handler and delete it from the pending state. The writable nature of the pending state would be limited only to adding new entries for a pending airdrop to be transferred or to remove already added ones.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: I cleared out some of the terminology there. The pending state might be misinterpreted. Of course, the pending state keeping the airdrop information will be persisted in state and will be serialized. What is pending here is that we don't directly change the balances of the sender and recipient.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,122 @@
...
+
+The `WritableStore` of the airdrops, should contain a handler method and if a given pending airdrop is claimed, it could be delegated to the `CryptoTransferHandler` and executed.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: True, I removed this logic. This delegation would be part of the `TokenAirdropClaimHandler`

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
IvanKavaldzhiev: It won't be applicable for NFTs. I renamed the method, so that it's clear it will be used only for fungible tokens.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,123 @@
...
+	PendingAirdropValue getFungibleAirdropAmount(@NonNull final PendingAirdropId airdropId);
+	long sizeOfState();
+	void warm(@NonNull final PendingAirdropId airdropId);
+    List<PendingAirdropId> getPendingAirdropIdsBySender(final AccountId sender);
+}
+```
+
+The `ReadableAirdropStoreImpl` would keep an instance of `ReadableKVState<PendingAirdropId, PendingAirdropValue> airdropsState`.
+
+An additional method `getPendingAirdropIdsBySender` would be added to the `ReadableAirdropStore` to get all `PendingAirdropIds` for a specific sender. This would be done via filtering all `PendingAirdropIds` kept in state by the `sender` that is passed as a parameter.
```
netopyr, line 55: This will not work. We cannot filter all entries in the state. Please check how this is done in similar cases (e.g., for token associations).

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,123 @@
...
+
+    void remove(@NonNull PendingAirdropId airdropId) {
+        // invoked by the `TokenCancelAirdrop`
+    }
+}
+```
+
+### Operations
+
+1. Fetching airdrop values from the collections. This would be applicable only for fungible airdrops (for both `TokenAirdrop` and `TokenClaimAirdrop` transactions).
```
netopyr, line 76: Does this mean we cannot determine if there is a pending airdrop for an NFT?

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,123 @@
...
+   For a specific `PendingAirdropId`, we would get the corresponding `PendingAirdropValue`.
+2. Handling a new airdrop in state. We can have several cases:
+   - Only a single fungible or non-fungible airdrop for a specific token defined - we just put a new entry in the map
+   - An airdrop for an existing `PendingAirdropId` is already defined - we should replace the existing `PendingAirdropValue` for this `PendingAirdropId` with the aggregated new fungible amount
+3. Handling an airdrop `cancel` in state - we remove the `PendingAirdropId` key for the airdrop we want to cancel
+4. Handling an airdrop `claim` in state - we remove the `PendingAirdropId` key for the airdrop after we claim it and create a synthetic crypto transfer for it inside the `TokenClaimAirdropHandler`
```
netopyr, line 82: Nit: I believe we agreed that we do not want to create synthetic crypto transfers in this case.

IvanKavaldzhiev: Removed

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
IvanKavaldzhiev: No, this method will return the TokenAirdropValue is applicable only for fungible airdrops. For NFT airdrops we don't have such a value. It just keeps the amount of fungible tokens we want to airdrop.

I added a clarification sentence, but basically for an NFT airdrop we would have all of the needed information in the PendingAirdropId itself and won't need to fetch something additional.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
IvanKavaldzhiev: I changed the design and aligned it with the `TokenRelationStore`. Instead of filtering inside of state, now the design doc reflects getting the airdrop keys from state and iterate over them from the outside and perform business logic there.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,202 @@
...
+
+We should define the structure of pending changes that would be kept in the `InitialModServiceTokenSchema` state, so we would reuse the token’s serialized state and add pending airdrops records in it. We could use the type defined in the protobuf records for the airdrop information itself. We should use the protobuf types since they are used for state serialization. So, the structure could look like this:
+
+`StateDefinition.*onDisk*(*AIRDROPS_KEY*, PendingAirdropId.PROTOBUF, PendingAirdropValue.PROTOBUF, MAX__AIRDROPS);`
+
+We can have a case, where we have multiple airdrops for the same fungible token. In this case the  `PendingAirdropValue` would be updated with the new aggregated value.
+
+### Models
+
+In order to be able to iterate over all pending airdrops for a specific sender account, we should enrich the account protobuf with 2 new fields:
```
netopyr, line 40: There is just one new field.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
```diff
@@ -0,0 +1,202 @@
...
+2. Airdrop a fungible token with a `Long.MAX_VALUE` amount to Alice and then make a second airdrop with the same fungible token and amount with a value of 1 again to Alice → the second airdrop attempt should fail and it’s data shouldn’t be saved in the pending state
+3. Airdrop an NFT with a negative serial number to Alice → the airdrop should fail and nothing should be saved in the pending state
+4. Airdrop an NFT with a serial number 1 to Alice and then perform another airdrop with the same token and a serial number 2 again to Alice → the airdrop should succeed and all of the airdrop data should be saved in the pending state
+5. Airdrop a fungible token with amount X to Alice and the same fungible token with amount Y to Bob → the airdrops should succeed and all airdrop data should be saved in the pending state
+6. Airdrop a fungible token with amount X to Alice and the same fungible token with amount Y again to Alice → the airdrops should succeed and the airdrop data should be aggregated having an amount of X + Y and should be saved in the pending state
```
netopyr, line 180: What is the difference between 5. and 6. Is 5. meant to be for NFTs?

netopyr (dismissed):
LGTM - thanks @IvanKavaldzhiev

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
IvanKavaldzhiev: Both are connected to fungible tokens. The difference is that in 6. we have the same recipient for separate airdrops and the end result would be aggregation of the two.

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
IvanKavaldzhiev: Fixed

hedera-node/docs/design/services/token-service/airdrops/airdrop-pending-state.md
netopyr: Ah, now I see it. Thanks!

netopyr (approved):
LGTM - thanks @IvanKavaldzhiev

PR #12794, opened 2024-04-15
docs: add a design doc for TokenAirdropTransaction

IvanKavaldzhiev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+     */
+    rpc airdropTokens (Transaction) returns (TransactionResponse);
+```
+
+### Fee types
+
+The airdrop transaction would include several types of fees, a different set of them would be applicable, based on the conditions the airdrop is performed to. The list of fees is:
+
+- Airdrop fee covering the operation itself
+- Association fee covering a pre-paid fee for associating the token being airdropped to the receiver, in case there is a missing relationship between the two
```
Neeharika-Sompalli, line 125: This is only if the `maxAutoAssociations` is not greater than `usedAutoAssociations` right?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+    - Verify that the pending airdrops list does not have any duplicate entries
+    - Pre-handle:
+        - The transaction must be signed by the sender for each entry in the `token_transfers` list
+        - In the case where the airdrop fees are covered by a separate payer, the transaction must include also the payer’s signature
+    - Handle:
+        - Any additional validation depending on config or state i.e. semantics checks
+        - Check that the airdrop sender account is a valid account. That is:
+          - an existing account, which was not deleted or expired
+          - a non-existing account, pointed by an alias of `evm_address` or public `ECDSA key`
+        - Check that the token being airdropped, does not contain a custom fee, which needs to be covered by the recipient. Such transactions should be reverted and rejected.
```
Neeharika-Sompalli, line 157: You can mention `fallbackRoyaltyFee` here also

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+        - The business logic for sending pending airdrops
+            - Check the association status and the existence of the recipient:
+
+              - It is an existing one and has `max_auto_associations` set to -1 and is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a `TokenAssociation`
+
+              - It is an existing one and has `max_auto_associations`, which is a positive number and there are free auto association slots and the recipient is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a `TokenAssociation`
```
Neeharika-Sompalli, line 165: Need to be clear here that it is added to `auto_associations` section of `TransactionRecord`

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+
+              - It is an existing one and has `max_auto_associations`, which is a positive number but there are no free auto association slots and the recipient is not associated to the token →
+                the token transfer is interpreted as a pending and a `TokenPendingAirdrop` entry is added to the `TransactionRecord`
```
Neeharika-Sompalli, line 168: Here it will be added only to `pendingAirdrops` section in `TransactionRecord` right ?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+        - Any additional validation depending on config or state i.e. semantics checks
+        - Check that the airdrop sender account is a valid account. That is:
+          - an existing account, which was not deleted or expired
+          - a non-existing account, pointed by an alias of `evm_address` or public `ECDSA key`
+        - Check that the token being airdropped, does not contain a custom fee, which needs to be covered by the recipient. Such transactions should be reverted and rejected.
+        - The business logic for sending pending airdrops
+            - Check the association status and the existence of the recipient:
+
+              - It is an existing one and has `max_auto_associations` set to -1 and is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a `TokenAssociation`
```
Neeharika-Sompalli, line 162: Need to be clear here that it is added to auto_associations section of TransactionRecord

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+              - It’s not existing and the AccountID is a public `ECDSA` key →
+                the airdrop would first auto-create a new account with `maxAutoAssociations` field of -1 and directly transfer the token being airdropped to it. This would include a `TokenTransferList` and the alias of the newly created account in the `TransactionRecord`
+
+              - It’s not existing and the AccountID is an `evm_address` →
+                the airdrop would first lazily-create a new account with `maxAutoAssociations` field of -1 and directly transfer the token being airdropped to it. This would include a `TokenTransferList` and the alias of the newly created account in the `TransactionRecord`
+    - Fee calculation logic
+        - assess all required fees for the airdrop from the payer of the transaction. This includes:
+            - airdrop fee (assessed always)
+            - first auto-renew fee for the token (assessed always)
+            - custom token fees covered by the sender (assessed always if present)
```
Neeharika-Sompalli, line 185: Do we store these custom fees somewhere? When are the custom fees charged ?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: This association fee is not related to auto associations, but for explicit association which would happen in case the airdropped is claimed by the receiver. Since we don't want to add this expense to the receiving side, the airdrop sender should pay in advance for this potential future association.

If there are free auto association slots for the recipient, the airdrop would directly be transferred to the recipient and I think no association fee should be paid in this case.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: I propose not to store them. They would be charged in the `TokenAirdrop.calculateFees` during the sending of airdrop

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
jsync-swirlds: Based on the HIP the sender will pay for the automatic association for a recipient who has an automatic association slot available.
>As an hts-token-airdrop-sender I want to cover the fees (custom, association & rent until the end of the first full auto renewal period) of an airdrop that consumes an automatic token slot to reduce the friction for a receiving account

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Oh, I thought automatic associations are not covered by fees and only a transfer fee is paid in those cases. If they are, this association fee would also cover this auto association step, only if `maxAutoAssociations` >= `usedAutoAssociations`, so there are free auto association slots available.

jsync-swirlds: Interesting, I had understood that automatic association slots had a cost (both a fee when consumed, and a "rent" when the account is renewed).

Thanks for clarifying.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
+# Airdrop Transaction
+
+## Purpose
+
+The essence of HIP-904 is introducing a missing functionality in the Hedera ecosystem that enables airdrop transactions. 
+Airdrops are frequently used operation in the web3 ecosystem and a lot of blockchain, dapp or DAO developers incentivise users to explore and use their products through it. Early adopters are ofter awarded with the so called airdrops which are native tokens for the system and a dedicated amount of them are distributed in airdrops. Hedera could benefit from introducing this functionality and this is something longed for from the whole community.
```
jsync-swirlds, line 6: This reads better with some line wrapping (which will be removed in the rendered view), and there are a couple misspelled words and awkward grammar.
```suggestion
Airdrops are frequently used operation in the web3 ecosystem and a lot of blockchains,
 dapps, or DAO developers incentivize users to explore and use their products with
 this process. Early adopters are often awarded with so-called airdrops which are
 tokens for the system and a dedicated amount of them are distributed in such airdrops.
 Hedera will benefit from introducing this functionality and this is something long
 requested by the community.
```

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+     * All transfer fees (including custom fees and royalties), as well as the rent cost for the
+     * first auto-renewal period for any automatic-association slot occupied by the airdropped
+     * tokens, SHALL be charged to the account submitting this transaction.
+     */
+    rpc airdropTokens (Transaction) returns (TransactionResponse);
+```
+
+### Fee types
+
+The airdrop transaction would include several types of fees, a different set of them would be applicable, based on the conditions the airdrop is performed to. The list of fees is:
```
jsync-swirlds, line 122: Misuse of `would` here, and a slight improvement in wording.
```suggestion
The airdrop transaction will include several types of fees, a different set of them will be applicable, based on the characteristics of the airdrop. The list of fees is:
```

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+
+- Airdrop fee covering the operation itself
+- Association fee covering a pre-paid fee for associating the token being airdropped to the receiver, in case there is a missing relationship between the two
+- Custom fees covering any token custom fees, which are required to be paid by the sender of the potential transfer in case of claim
+- Transfer fee covering a pre-paid fee for transferring the token being airdropped to the receiver, so that when they potentially claim the token to be free of charge
+- Auto-renewal fees covering a pre-paid period for a full auto-renewal cycle
+- Fee covering auto-creation/lazy-creation of accounts, if we airdrop to an alias address of non-existing account
+
+Note that for all these fees, a given sender, which is the actual token owner, could be covered by a different payer, which is willing to pay. However, in case the receiver has not claimed an airdrop for a more than a full auto-renew cycle, the sender (token owner) would be required to continue paying future rents and they cannot be covered by a separate payer.
```
jsync-swirlds, line 131: I would recommend leaving off the last clause ` and they cannot be covered by a separate payer`.  Nothing in the HIP, that I can see, requires that nobody can pay those fees on the sender's behalf; in fact it is not unreasonable, in my opinion, to allow an account identified in the `auto_renew_account_id` for the sender to cover those rent fees (although that is not required by this HIP either).

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+
+An update into the `feeSchedule` file would be needed to specify the addition of the new fee values for the airdrop transaction type itself.
+
+### Keys
+
+Since a given sender/token owner could be the payer of the airdrop transaction or they can rely on a separate payer account we have a couple of use cases related to the key verification logic.
+
+If the sender of the airdrop is also the payer of this transaction, we would need only their key to be present in the transaction context.
+
+If the airdrop transaction fees from the list above are covered by a separate payer, then both the keys of the airdrop sender and the payer should be present in the context. That is needed, so that the sender authorizes giving an airdrop from their own balance.
```
jsync-swirlds, line 141:
```suggestion
If the airdrop transaction fees from the list above are covered by a separate payer, then the signatures of both the airdrop sender and the payer must be present in the context. This is so because the sender is required to authorize sending the airdrop tokens from their balance.
```

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Service updates
+
+- Update `TokenServiceDefinition` class to include the new RPC method definition for sending airdrops
+- Implement new `TokenAirdropHandler` class which should be invoked when the gRPC server handles `TokenAirdropTransaction` transactions. The class should be responsible for:
+    - Verify that the pending airdrops list contains between 1 and 10 entries, inclusive
+    - Verify that the pending airdrops list does not have any duplicate entries
```
jsync-swirlds, line 148: Do we disbar duplicates?  What if an airdrop smart contract were to send 3 tokens from Bob to Alice for using a new dApp, and send a bonus 3 tokens from Bob to Alice for "liking" the dApp account on `X`?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+    - Pre-handle:
+        - The transaction must be signed by the sender for each entry in the `token_transfers` list
+        - In the case where the airdrop fees are covered by a separate payer, the transaction must include also the payer’s signature
+    - Handle:
+        - Any additional validation depending on config or state i.e. semantics checks
+        - Check that the airdrop sender account is a valid account. That is:
+          - an existing account, which was not deleted or expired
+          - a non-existing account, pointed by an alias of `evm_address` or public `ECDSA key`
```
jsync-swirlds, line 156: I do not think this specific line would be valid, because the non-existent account would not have any balance to send...

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+            - Check the association status and the existence of the recipient:
+
+              - It is an existing one and has `max_auto_associations` set to -1 and is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
+
+              - It is an existing one and has `max_auto_associations`, which is a positive number and there are free auto association slots and the recipient is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
+
+              - It is an existing one and has `max_auto_associations`, which is a positive number but there are no free auto association slots and the recipient is not associated to the token →
+                the token transfer is interpreted as a pending and a `TokenPendingAirdrop` entry is added to the `TransactionRecord`
```
jsync-swirlds, line 169: Do we need to specify also that the PendingAirdropId/PendingAirdropValue entry is added to the PendingAirdrop virtual map in state?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+    - Handle:
+        - Any additional validation depending on config or state i.e. semantics checks
+        - Check that the airdrop sender account is a valid account. That is:
+          - an existing account, which was not deleted or expired
+          - a non-existing account, pointed by an alias of `evm_address` or public `ECDSA key`
+        - Check that the token being airdropped, does not contain a custom fee, which needs to be covered by the recipient. This includes `fractionalFees` with `net_of_transfers=false` and `royaltyFees`, including `fallbackRoyaltyFees`.
+        - Such transactions should be reverted and rejected.
+        - The business logic for sending pending airdrops
+            - Check the association status and the existence of the recipient:
+
```
jsync-swirlds, line 161: Probably shouldn't have blank lines between bullet points.  Some markdown processing will create independent lists in that case.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+              - It is an existing one and has `max_auto_associations` set to -1 and is not associated to the token →
+                the token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
```
jsync-swirlds, line 163: It looks like you're trying to create a case/answer structure with the arrow at end of line, if that's the case, you probably need to add an actual line break ("view" this file to see how github removes the line breaks otherwise).
```suggestion
              - It is an existing one and has `max_auto_associations` set to -1 and is not associated to the token →<br/>
                The token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
```

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+* Verify that an airdrop with a token and existing recipient with `maxAutoAssociations=0` and missing token association, works correctly and the airdrop lends in the pending state
+* Verify that an airdrop with a token and existing recipient with `maxAutoAssociations` equal to a positive number and there are no free `autoAssociations` slots works correctly and airdrop lends in the pending state
+* Verify that an airdrop with a token and a missing recipient pointed by its public `ECDSA key alias`, works correctly by auto-creating the recipient with `maxAutoAssociations=-1` and the token is directly transferred to it
+* Verify that an airdrop with a token and a missing recipient pointed by its `evm_address` alias, works correctly by lazy-creating the recipient with `maxAutoAssociations=-1` and the token is directly transferred to it
+* Verify that an airdrop with a missing TokenID fails
+* Verify that an airdrop with a missing recipient pointed by an `ED25519 key alias` fails
```
jsync-swirlds, line 202: Are we certain this is not supposed to work?  As far as I know it is entirely valid to send a `CryptoTransfer` to an ED25519 public key and it will auto-create the account, setting the `alias` to the public key bytes.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: You're correct! My bad, I was probably referring to the recipient accounts, but this specific section is for the sender accounts only. I'll delete the second case.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,208 @@
...
+ *    claimed immediately, the `sender` SHALL pay the cost to renew the token association, and
+ *    the cost to maintain the pending airdrop, until the pending airdrop is claimed or cancelled.
+ *
+ * If an airdrop would create a pending airdrop for a fungible/common token, and a pending airdrop
+ * for the same sender, receiver, and token already exists, the existing pending airdrop
+ * SHALL be updated to add the new amount to the existing airdrop, rather than creating a new
+ * pending airdrop.
+ *
+ * Any airdrop that completes immediately SHALL be irreversible. Any airdrop that results in a
+ * "Pending Airdrop" MAY be canceled via a `cancelAirdrop` transaction.<
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: I agree with you. This statements makes the doc a bit restrictive and this might extend in the future. I removed the last clause.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: I added this clarification.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: You're right. That's still a valid case. I removed this scenario.

jsync-swirlds (dismissed):

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,205 @@
...
+     * <p>
+     * All token transfers MUST successfully transfer tokens or create a pending airdrop
+     * for this transaction to succeed.<br/>
+     * This list MUST contain between 1 and 10 transfers, inclusive.
+     * <p>
+     * <blockquote>Note that each transfer of fungible/common tokens requires both a debit and
+     * a credit, so each _fungible_ token transfer MUST have _balanced_ entries in the
+     * TokenTransferList for that transfer.</blockquote>
+     */
+     repeated TokenTransferList token_transfers= 1;
```
Neeharika-Sompalli, line 94: Using a `TokenTransferList` here doesn't seem right. How are we doing to enforce there is one credit and one debit in each transfer ? 

Seems like we need a different protobuf definition that has one credit and one debit.
Allowing user to set multiple credits and debits will make it confusing to create a `PendingAirdropId` that needs one sender and one receiver

Neeharika-Sompalli: @tinker-michaelj @netopyr @Nana-EC   What do you think ?

jsync-swirlds: Each Pending Airdrop is created individually from a _single_ transfer in the list, and only if _necessary_. The majority of transfers are expected to complete without creating a pending airdrop.
I would expect to see one PendingAirdropId for _each_ combination of sender and recipient in the transfer list(s), and _only_ if the recipient cannot immediately receive the airdrop.

IvanKavaldzhiev: In the case of fungible airdrops having the same sender and receiver - we would result in a single transfer with the fungible value aggregated.

In the case of NFT airdrops with different serial numbers but the same sender and receiver - we would have a single transfer in the list for each separate serial number with the same sender and receiver.

This is my understanding, having the current HIP definition in mind.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Actually in the diagram in the HIP we can see that even if the receiver has free auto associate slots, the airdrop sender should still cover an association fee, so I think @jsync-swirlds is right and the association fee charged during an airdrop covers both a direct auto association or a potential manually triggered association, during a claim by the recipient. In any case, we should charge this association fee from the airdrop sender. That's why we also change the existing `CryptoTransfer` fee logic by making the sender always paying for the token association.

@Nana-EC , @Neeharika-Sompalli , what do you think?

image

jsync-swirlds (approved):

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
Neeharika-Sompalli: Ah I see .. so, we don't charge `maxAutoAssociations` during creation anymore. Then this makes sense.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
Neeharika-Sompalli: I still don't understand, I think we will need to add validations in code. But probably don't want to block this PR. 🙂

Neeharika-Sompalli (dismissed):
LGTM ! But I am not sure adding `TokenTransferList` in protobufs is correct.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -193,13 +193,14 @@ If the airdrop transaction fees from the list above are covered by a separate pa
...
 * Verify that an airdrop with wrong input data (e.g. negative amount, negative serial number or missing mandatory field) fails
 * Verify that an airdrop with a missing signature of the sender fails
-* Verify that an airdrop, which fee expenses are covered by a separate payer account but has missing payer signature - fails
-* Verify that an airdrop with more than 10 entries in the `token_transfers` list - fails
-* Verify that an airdrop with a duplicate entry in the `token_transfers` list - fails
\ No newline at end of file
+* Verify that an airdrop, which fee expenses are covered by a separate payer account, but has missing payer signature fails
+* Verify that an airdrop with more than 10 entries in the `token_transfers` list fails
+* Verify that an airdrop with a duplicate entry for the same NFT in the `token_transfers` list fails with `PENDING_NFT_AIRDROP_ALREADY_EXISTS`
+* Verify that an airdrop with a duplicate entry for the same fungible token and recipient in the `token_transfers` list works correctly and the fungible token amount is aggregated in the final airdrop to the recipient
```
Neeharika-Sompalli, line 206: Is this valid case? In `CryptoTransfer` we reject duplicate entries. We should follow similar pattern imo.

IvanKavaldzhiev: In the case of airdrops, it's allowed for the same recipient to receive a given fungible token multiple times in the same airdrop transaction.

@Nana-EC can you confirm this?

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+                Тhe token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
+              - It is an existing one and has `max_auto_associations`, which is a positive number and there are free auto association slots and the recipient is not associated to the token →<br/>
+                Тhe token is auto associated and is directly transferred to the recipient and a `TokenTransferList` is added to the `TransactionRecord`, as well as a new `TokenAssociation` entry in the `automatic_token_associations`
+              - It is an existing one and has `max_auto_associations`, which is a positive number but there are no free auto association slots and the recipient is not associated to the token →<br/>
+                Тhe token transfer is interpreted as a pending and a `TokenPendingAirdrop` entry is added to the `TransactionRecord`. In addition, the airdrop `PendingAirdropId`/`PendingAirdropValue` entry is added to the `PendingAirdrop` VirtualMap in state
```
netopyr, line 169: `VirtualMap` is one possible implementation. In the business logic layer, we should only reference stores.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
netopyr: I agree with @Neeharika-Sompalli that we should follow the same pattern.

The HIP mentions the case of two separate airdrop transactions for the same token.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -0,0 +1,206 @@
...
+            - custom token fees covered by the sender (assessed always if present)
+            - crypto transfer fee (pre-paid always)
+            - token association fee (pre-paid if account is not associated to the token)
+            - auto account creation (assessed if the recipient does not exist and they are referred by their public `ECDSA` key or `evm_address`)
+- Interact with the pending airdrop state - if the recipient is not associated to the token, then the `TokenAirdropTransaction` responsibilities in term of interacting with the pending state would be to just add the airdrop information in the state
+- Add this new operation type to the `ThroughputLimits` throttle bucket group, so that it's included in the throttling mechanism
+
+## Acceptance Tests
+
+* Verify that an airdrop with a token and existing recipient with `maxAutoAssociations=-1` works correctly and the recipient receives the token directly
```
netopyr, line 193: Please add a test that checks the behavior when the feature flag is disabled.

netopyr:
The doc does not mention allowances at all. Please explain how allowances may modify the behavior (where applicable).

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Added

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Fixed

IvanKavaldzhiev:
> The doc does not mention allowances at all. Please explain how allowances may modify the behavior (where applicable).

Added more information about allowances and their use case. Also added test cases related to this scenario.

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
IvanKavaldzhiev: Noted

Neeharika-Sompalli (approved):
LGTM!

netopyr (approved):
LGTM - thanks @IvanKavaldzhiev

hedera-node/docs/design/services/token-service/airdrops/airdrop-transaction.md
```diff
@@ -211,4 +212,5 @@ If the airdrop transaction fees from the list above are covered by a separate pa
 * Verify that an airdrop, which fee expenses are covered by a separate payer account, but has missing payer signature fails
 * Verify that an airdrop with more than 10 entries in the `token_transfers` list fails
 * Verify that an airdrop with a duplicate entry for the same NFT in the `token_transfers` list fails with `PENDING_NFT_AIRDROP_ALREADY_EXISTS`
+* Verify that deletion of an account who has sent an airdrop that is still in pending state fails
```
jsync-swirlds: This probably should not fail (the account owner must be able to delete the account without resorting to a third-party action).  I believe it should delete the pending airdrop instead, so that a pending airdrop cannot "pin" an account in place.

IvanKavaldzhiev: We had some discussions that in order to delete an account that has sent airdrops, you should first delete all related airdrops to this account. So, I believe this should happen with explicit TokenAirdropCancel transactions beforehand.

@Nana-EC, @netopyr  what do you think?

jsync-swirlds: The recipient account cannot cancel the airdrops, so a token holder could, in effect, pin someone else's account by sending airdrops that would have to be claimed (and using a structure that makes claim unpalatable).

jsync-swirlds: I misread the line as a recipient rather than a sender.

After re-reading, I agree with the existing wording.

Neeharika-Sompalli (approved):

jsync-swirlds (approved):

PR #12797, opened 2024-04-15
feat: Enhance purechecks for CryptoCreateHandler

### LINKED ISSUE #12714: Validate and Update pureChecks(): CryptoCreateHandler
This handler could have the following additional pure checks:
* autoRenewPeriod: if not null, check for > -1
* shardID: if not null, check for > 0
* realmID: if not null, check for > 0
* newRealmAdminKey: if not null, check for empty list??
* memo:  verify not null (but empty string is allowed)
* maxAutomaticTokenAssociations: check for > -1
* stakedId: verify not null
        - if type is stakedAccountID, verify stakedAccountID is valid
        - if type is stakedNodeId, verify stakedNodeId is valid
* alias: verify not null   

Lastly, there may be a mismatch between the handler logic and the protobuf object. `autoRenewPeriod` is a `@Nullable` field, but the handler requires this field to be present [here](https://github.com/hashgraph/hedera-services/blob/7056eeacde38b686114b7bf253baf0a7a5c9bf7a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java#L115). Which is correct?

thomas-swirlds-labs:
**Description**:
Adding additional pure checks to CryptoCreateHandler and associated Hapi Tests. 

**Related issue(s)**:

Fixes #12714 

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
```diff
@@ -111,15 +112,22 @@ public CryptoCreateHandler(
     @Override
     public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
         final var op = txn.cryptoCreateAccountOrThrow();
+        validateTruePreCheck(
+                op.autoRenewPeriod() != null && op.autoRenewPeriod().seconds() > -1, INVALID_RENEWAL_PERIOD);
+        if (op.hasShardID()) {
```
jsync-swirlds, line 117: Shard and realm shouldn't be in pure checks.
Those aren't "the node failed" issues; they're user error which should be checked in handle.
autorenew period also should be in handle.

The general path (once we avoid catastrophic failures) is to get transactions to handle so we succeed in charging the payer if anything else is wrong. The payer on the transaction is responsible for submitting a transaction that can be processed, and we want to hold that payer responsible as much as possible.

Neeharika-Sompalli: Agreed with @netopyr

jsync-swirlds: If we need a temporary check, we could just check that realm and shard are exactly equal to `0`, given we only support the `0` shard and `0` realm currently.

`autoRenwePeriod` _should_ allow null (if we have trouble with null in that field it's possibly a bug), but if set I can see checking it's `>= 0`.

We definitely don't want to be checking for a non-null `newRealmAdminKey` as that's only for _creating a new realm_ and _should not_ be set.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
```diff
@@ -111,15 +112,22 @@ public CryptoCreateHandler(
...
+            validateTruePreCheck(op.realmIDOrThrow().realmNum() >= 0, INVALID_ACCOUNT_ID);
+        }
+        if (op.hasNewRealmAdminKey()) {
+            validateFalsePreCheck(isEmpty(op.newRealmAdminKeyOrThrow()), INVALID_ADMIN_KEY);
+        }
+        validateTruePreCheck(op.maxAutomaticTokenAssociations() >= 0, INVALID_TRANSACTION_BODY);
         validateTruePreCheck(op.initialBalance() >= 0L, INVALID_INITIAL_BALANCE);
         validateTruePreCheck(op.hasAutoRenewPeriod(), INVALID_RENEWAL_PERIOD);
-        // FUTURE: should this return SEND_RECORD_THRESHOLD_FIELD_IS_DEPRECATED
-        validateTruePreCheck(op.sendRecordThreshold() >= 0L, INVALID_SEND_RECORD_THRESHOLD);
```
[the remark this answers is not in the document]
thomas-swirlds-labs: I believe it was removed since it was deprecated. Have restored it now 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
thomas-swirlds-labs: Thanks @jsync-swirlds. Have addressed now.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoCreateHandler.java
```diff
@@ -111,15 +113,29 @@ public CryptoCreateHandler(
...
         validateTruePreCheck(op.initialBalance() >= 0L, INVALID_INITIAL_BALANCE);
-        validateTruePreCheck(op.hasAutoRenewPeriod(), INVALID_RENEWAL_PERIOD);
         // FUTURE: should this return SEND_RECORD_THRESHOLD_FIELD_IS_DEPRECATED
         validateTruePreCheck(op.sendRecordThreshold() >= 0L, INVALID_SEND_RECORD_THRESHOLD);
         // FUTURE: should this return RECEIVE_RECORD_THRESHOLD_FIELD_IS_DEPRECATED
         validateTruePreCheck(op.receiveRecordThreshold() >= 0L, INVALID_RECEIVE_RECORD_THRESHOLD);
         validateTruePreCheck(
                 op.proxyAccountIDOrElse(AccountID.DEFAULT).equals(AccountID.DEFAULT),
                 PROXY_ACCOUNT_ID_FIELD_IS_DEPRECATED);
+        validateTruePreCheck(op.hasAutoRenewPeriod(), INVALID_RENEWAL_PERIOD);
```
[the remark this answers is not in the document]
thomas-swirlds-labs: Good catch, have updated 👍

Neeharika-Sompalli (approved):

jsync-swirlds (approved):

PR #12821, opened 2024-04-15
feat: 12226 flush task for swirlds-log

### LINKED ISSUE #12226: Implement a Flush-Task for the loggers
Using Base thread executor created on: https://github.com/hashgraph/hedera-services/issues/12095
Implement a rolling flush task that periodically calls the flush method on the handlers.

mxtartaglia-sl:
Fixes #12226


platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/file/FileHandler.java
```diff
@@ -77,6 +80,19 @@ public FileHandler(
...
             throw new IOException("Could not create FileHandler", e);
         }
+        BaseExecutorFactory.getInstance()
+                .scheduleAtFixedRate(
+                        () -> {
+                            try {
+                                outputStream.flush();
+                            } catch (IOException e) {
+                                EMERGENCY_LOGGER.log(
+                                        Level.WARN, "Unable to flush " + handlerName + " nothing to do, will retry", e);
```
hendrikebbers, line 90: I do not get the message.

mxtartaglia-sl: Let me rethink the message
Given that it is a periodic operation that is automatically invoked, I thought it was best to try to communicate that it was not an end-of-the-world failure...

PR #12823, opened 2024-04-15
chore: 12815: restore CacheWarmer() constructor with a provided executor

### LINKED ISSUE #12815: Restore CacheWarmer() constructor with a provided executor
### Description

The change was erroneously introduced in #12753. Restore the constructor to be visible for testing and add a test to keep it used.

### Steps to reproduce

From code review.

### Additional context

_No response_

### Hedera network

other

### Version

0.49.0-SNAPSHOT

### Operating system

macOS

OlegMazurov:
**Description**:
* Restored `CacheWarmer()` constructor with provided `Executor` erroneously removed in #12753. 
* Added a unit test to keep it used.

**Related issue(s)**:

Fixes #12815

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/CacheWarmer.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/CacheWarmer.java
@@ -18,6 +18,7 @@ package com.hedera.node.app.workflows.handle;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.TransactionID;
 import com.hedera.hapi.node.transaction.TransactionBody;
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/CacheWarmer.java
```diff
@@ -59,13 +60,23 @@ public CacheWarmer(
...
+            @NonNull final TransactionDispatcher dispatcher,
+            @NonNull final Executor executor) {
         this.checker = checker;
         this.dispatcher = requireNonNull(dispatcher);
-        final int parallelism = configProvider
-                .getConfiguration()
-                .getConfigData(CacheConfig.class)
-                .cryptoTransferWarmThreads();
-        this.executor = new ForkJoinPool(parallelism);
+        this.executor = requireNonNull(executor);
```
rbair23, line 79: Rather than doing it this way, I was wondering if we should use a Dagger provider to create the executor that we use in this constructor. It is more elaborate, but improves the interface. @tinker-michaelj has a good handle on how that would work.

netopyr: Yes, makes sense. In particular, once the `ExecutorFactory` from platform-base is available, this will be much cleaner.

tinker-michaelj (approved):
LGTM, ty @netopyr !

Neeharika-Sompalli (approved):
LGTM! Thanks @OlegMazurov @netopyr

Issue #12012, opened 2024-03-11
contractCalls to etheretumContractCalls NPE


vtronkov:
https://github.com/hashgraph/hedera-services/pull/10944 introduced a bug that is breaking the conversion of contractCalls to etheretumContractCalls in the suites.
The current workaround(for local purposes only) is to short-circuit `GenesisRecordsConsensusHook.process` but I'm not sure if this is the source of the problem.

To reproduce the bug you can try running the following tests:
`com.hedera.services.bdd.suites.contract.precompile.ApproveAllowanceSuite#htsTokenApprove`
`com.hedera.services.bdd.suites.contract.precompile.ContractMintHTSSuite#transferNftAfterNestedMint`
`com.hedera.services.bdd.suites.contract.precompile.PrngPrecompileSuite#invalidLargeInputFails`
(don't forget to override the name method and add "_Eth" suffix)

The error I am recieving is `java.lang.NullPointerException: Cannot invoke "com.hederahashgraph.api.proto.java.Response.getTransactionGetRecord()" because "this.response" is null`

Digging further @JivkoKelchev  and I found that:
1. In the `GenesisRecordsConsensusHook.process` method if `(blockStore.getLastBlockInfo().consTimeOfLastHandledTxn() != null) return;` was replaced with `if (!shouldStreamRecords(blockStore, context)) return;`
1.1 Debugging the old if statement in the test cases above - we are always getting true as an outcome - short-circuiting the method. Thus we've never created the system accounts.
1.2 We are not sure if this is accurate behavior (probably not). It seems that it should run once every time a node is started.
1.3 If the old code was correct then we have the fix of the problem - we just revert it.
2. With the new code/changes we are getting `MAX_CHILD_RECORDS_EXCEEDED`. This happens only on the first test from the suite on the first transaction. What is happening is that `GenesisRecordsConsensusHook.process` is creating 500+ proceeding child records and when we execute the first transaction the child records are counted towards the system child records, thus throwing `MAX_CHILD_RECORDS_EXCEEDED`. If this is something we should fix we should somehow mark the child record as `UNLIMITED_CHILD_RECORDS` or maybe somehow split the initial system account creation from the first transaction.

PR #12832, opened 2024-04-16
fix: contractCalls to etheretumContractCalls NPE

vtronkov:
**Description**:
Fixing `MAX_CHILD_RECORDS_EXCEEDED` exception when executing tests that are converted to Ethereum.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/12012

**Notes for reviewer**:
The bug was not caused by https://github.com/hashgraph/hedera-services/pull/10944 - it was revealed by it. Till now we were short-circuiting `GenesisRecordsConsensusHook.process` but after https://github.com/hashgraph/hedera-services/pull/10944 was merged we started actually executing the method.

The tests that were converted to Ethereum were throwing `MAX_CHILD_RECORDS_EXCEEDED`. The reason for that is the 
`UtilStateChange.initializeEthereumAccountForSpec`(now `createEthereumAccountForSpec`) method with a combination of `GenesisRecordsConsensusHook.process`.

The `process` method is creating a lot of system accounts. We are marking all the records created from that with `UNLIMITED_CHILD_RECORDS` which bypass the `MAX_CHILD_RECORDS_EXCEEDED` check. The problem is that when we execute the `GenesisRecordsConsensusHook.process` it's executed together, I assume, with the system accounts creation. And since the `GenesisRecordsConsensusHook.process` is not a special process it's not marked as `UNLIMITED_CHILD_RECORDS` we are checking for `MAX_CHILD_RECORDS_EXCEEDED`. The counter is not skipping the `UNLIMITED_CHILD_RECORDS` and we are exceeding the limit and throwing the error.

I was thinking about changing the counting logic so we can skip all the `UNLIMITED_CHILD_RECORDS` records but I think it would be easier and simpler if we don't execute directly the accounts from`GenesisRecordsConsensusHook.process` and we just return them and add them as part of the test's operations.

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpec.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpec.java
@@ -35,7 +35,7 @@ import static com.hedera.services.bdd.spec.queries.QueryVerbs.getTxnRecord;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.cryptoTransfer;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.scheduleCreate;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.scheduleSign;
-import static com.hedera.services.bdd.spec.utilops.UtilStateChange.initializeEthereumAccountForSpec;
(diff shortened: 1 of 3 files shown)


Neeharika-Sompalli (approved):
LGTM !

Issue #11033, opened 2024-01-18
Direct call w/o value to system accounts [0.0.361-0.0.750] result in precompile error


yiliev0:
### Description

Executing a ContractCall w/o value to system account in the range[0.0.361-0.0.750) result in `PRECOMPILE_ERROR` 

### Steps to reproduce

1. Create an account
2. Execute ContractCall to 0.0.629 using the account from step1.

**Expected result**: The call should succeed with no-op. 
**Actual result**: The call fails with `PRECOMPILE_ERROR`

### Hedera network

other

### Version

0.46.0

### Operating system

None

stefan-stefanooov:
This will be taken care of after 0.48 version.

PR #12833, opened 2024-04-16
feat: Direct call w/o value to system accounts [0.0.361-0.0.750] change output

stefan-stefanooov:
Direct call w/o value to system accounts [0.0.361-0.0.750] result in precompile error

### **Description**:
Executing a ContractCall w/o value to system account in the range[0.0.361-0.0.750) result in PRECOMPILE_ERROR
`Expected result:` The call should `succeed` with `no-op`.

### **Steps to reproduce**

1. Create an account
2. Execute ContractCall to 0.0.629 using the account from step1.
3. `Expected` result: The call should succeed with no-op.
4. `Actual` result: The call fails with PRECOMPILE_ERROR

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/11033
Fixes #
https://github.com/hashgraph/hedera-services/issues/11033

Fixes #

### **Notes for reviewer**

This is achieved by adding a new function called `handleNonExtantSystemAccount`. It will: 

>  //    Create an empty/successful result object
> `PrecompileContractResult result = PrecompileContractResult.success(Bytes.EMPTY);`
> //    Pass it to the existing **finishPrecompileExecution** function that mutates the **frame** object with the required values.
> `finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);`

We call this new function in `CustomMessageCallProcessor.java` -> `start`, when: 

> - We don't have a Hedera system contract (precompile) with this input address **codeAddress** -> `systemContracts.containsKey(codeAddress)` returns `false`
> - the input **codeAddress** for the frame is from the reserved system accounts rage [0.0.0-0.0.750) -> `addressChecks.isSystemAccount` returns `true`
> - we don't have a holt event
> - we don't have a EVM precompile with this input address
>

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -171,6 +171,7 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
     private void handleNonExtantSystemAccount(
             @NonNull final MessageFrame frame, @NonNull final OperationTracer tracer) {
         final PrecompileContractResult result = PrecompileContractResult.success(Bytes.EMPTY);
+        frame.decrementRemainingGas(frame.getRemainingGas());
         finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);
     }
 
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -129,6 +129,16 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
             if (alreadyHalted(frame)) {
                 return;
             }
+
+            handleNonExtantSystemAccount(frame, tracer);
+            return;
+        }
+
+        // Handle evm precompiles
+        final var evmPrecompile = precompiles.get(codeAddress);
+        if (evmPrecompile != null) {
+            doExecutePrecompile(evmPrecompile, frame, tracer);
+            return;
         }
 
         // Transfer value to the contract if required and possibly halt
@@ -137,13 +147,6 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
             if (alreadyHalted(frame)) {
                 return;
             }
-        }
-
-        // Handle evm precompiles
-        final var evmPrecompile = precompiles.get(codeAddress);
-        if (evmPrecompile != null) {
-            doExecutePrecompile(evmPrecompile, frame, tracer);
-            return;
         }
 
         // For mono-service fidelity, we need to consider called contracts
@@ -160,6 +163,12 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
 
     public boolean isImplicitCreationEnabled(@NonNull Configuration config) {
         return featureFlags.isImplicitCreationEnabled(config);
+    }
+
+    private void handleNonExtantSystemAccount(
+            @NonNull final MessageFrame frame, @NonNull final OperationTracer tracer) {
+        final PrecompileContractResult result = PrecompileContractResult.success(Bytes.EMPTY);
+        finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);
     }
 
     private void doExecutePrecompile(
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -171,7 +171,7 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
     private void handleNonExtantSystemAccount(
             @NonNull final MessageFrame frame, @NonNull final OperationTracer tracer) {
         final PrecompileContractResult result = PrecompileContractResult.success(Bytes.EMPTY);
-        frame.decrementRemainingGas(frame.getRemainingGas());
+        frame.clearGasRemaining();
         finishPrecompileExecution(frame, result, PRECOMPILE, (ActionSidecarContentTracer) tracer);
     }
 
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
@@ -112,6 +112,7 @@ public class CustomMessageCallProcessor extends MessageCallProcessor {
     @Override
     public void start(@NonNull final MessageFrame frame, @NonNull final OperationTracer tracer) {
         final var codeAddress = frame.getContractAddress();
+        final var evmPrecompile = precompiles.get(codeAddress);
         // This must be done first as the system contract address range overlaps with system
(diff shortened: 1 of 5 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -94,7 +93,7 @@ public CustomMessageCallProcessor(
     }
 
     /**
-     * Starts the execution of a message call based on the contract address of the given frame,
+     * Sta§rts the execution of a message call based on the contract address of the given frame,
```
lukelee-sl, line 96: Is this change on purpose?

lukelee-sl (commented):
I wonder if this use case is handled by [this](https://github.com/hashgraph/hedera-services/pull/12842/files) pr?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -124,12 +123,23 @@ public void start(@NonNull final MessageFrame frame, @NonNull final OperationTra
             return;
         }
 
+        // Handle evm precompiles
```
lukelee-sl, line 126: Question - why did this block have to move.  If there is a good explanation maybe include a comment?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
stefan-stefanooov: Not really. My mistake ... it should have been a "$" I guess... for all the money that is going to start flowing after this fix is deployed. 💰 
I'll remove it. Thank you, Sir.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
stefan-stefanooov: I moved it because of readability and changing operation order was important. 
But since then for consolidating the work with the MR you mentioned above ... I changed back the order. Thank you.

stefan-stefanooov:
> I wonder if this use case is handled by [this](https://github.com/hashgraph/hedera-services/pull/12842/files) pr?

That was a good catch. 
We synced with the Merge Request authors and we are figuring out a way to move forward. 
We are gravitating towards  merging my Request first and then they will work on top of it. 

Thank you Sir!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -217,7 +217,8 @@ public HapiTestEngineExecutionContext before(HapiTestEngineExecutionContext cont
 
             // Allow for a simple switch to enable in-process Alice node for debugging
             final String debugEnv = System.getenv("HAPI_DEBUG_NODE");
-            final boolean debugMode = Boolean.parseBoolean(debugEnv);
+            // final boolean debugMode = Boolean.parseBoolean(debugEnv);
+            final boolean debugMode = true;
```
Neeharika-Sompalli, line 221: Is this intentional ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
```diff
@@ -4245,7 +4241,7 @@ HapiSpec traceabilityE2EScenario15() {
                                                             .setRecipientContract(
                                                                     spec.registry()
                                                                             .getContractId(contract))
-                                                            .setGasUsed(80135)
+                                                            .setGasUsed(80193)
```
Neeharika-Sompalli, line 4244: Are these gas changes different from mono-service ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
```diff
@@ -4260,7 +4256,7 @@ HapiSpec traceabilityE2EScenario15() {
                                                             .setCallingContract(
                                                                     spec.registry()
                                                                             .getContractId(contract))
-                                                            .setGas(3870609)
+                                                            .setGas(3870552)
```
Neeharika-Sompalli, line 4259: same here

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
```diff
@@ -4588,7 +4580,7 @@ final HapiSpec traceabilityE2EScenario21() {
                                                 .setCallingAccount(TxnUtils.asId(GENESIS, spec))
                                                 .setCallOperationType(CallOperationType.OP_CALL)
                                                 .setGas(978936)
-                                                .setGasUsed(963748)
+                                                .setGasUsed(3172)
```
Neeharika-Sompalli, line 4583: Does this fail mono-service and work for mod service ? Why are we changing these numbers?

stefan-stefanooov: Hey! We are changing the flow and returning different statuses to the user. I tough that was the reason for the change in gas. Am I wrong ?

stefan-stefanooov: @Neeharika-Sompalli

Neeharika-Sompalli: I am not sure. But changing `HapiSpec` seems like changing the logic from mono-service. 
 @tinker-michaelj  thoughts?

petreze (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
stefan-stefanooov: @lukelee-sl , @david-bakin-sl
Could you please review this ?

lukelee-sl: IMO, we should discourage (but still allow) this use case where the user tries to call a function on a system account.  One way to do that is if we charge all the remaining gas in the transaction but still allow the transaction to succeed.  I believe if we did this then the `gasUsed` value above in the test may not have to be updated (I haven't tested this).

david-bakin-sl: If we do this `system-accounts-operations.md` should be updated - it doesn't discuss fees for "no-op success" at all.

petreze: @Neeharika-Sompalli @tinker-michaelj @lukelee-sl @david-bakin-sl 
We are changing the logic from mono, indeed. This fix and others from the same package are bugs found in mono codebase and has to be fixed in modu

stefan-stefanooov: New version available guys :) 
In it we consume all the gas. 
https://github.com/hashgraph/hedera-services/pull/12833/files#diff-79d792e4de7dd21f5fe7e60edebaea811a3bfd113a27838a84208e9747c00cbbR174

lukelee-sl (approved):
LGTM. Thank you!

petreze (approved):

Neeharika-Sompalli (approved):
I am not sure if the chnages to Hapi tests are expected. Will defer to @tinker-michaelj  or @lukelee-sl  or @david-bakin-sl review 🙂

PR #12834, opened 2024-04-16
Eth tx data negative value test

vtronkov:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
+++ b/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
@@ -368,8 +368,8 @@ public record EthTxData(
         }
 
         final var value = rlpList.get(4).asBytes().length == 0
-            ? BigInteger.ZERO
-            : new BigInteger(rlpList.get(4).asBytes());
+                ? BigInteger.ZERO
+                : new BigInteger(rlpList.get(4).asBytes());
 
         return new EthTxData(
                 rawTx,
@@ -381,7 +381,7 @@ public record EthTxData(
                 null, // maxGas
                 rlpList.get(2).asLong(), // gasLimit
                 rlpList.get(3).data(), // to
-            rlpList.get(4).asBigInt(), // value
+                value, // value
                 rlpList.get(5).data(), // callData
                 null, // accessList
                 recId,
@@ -407,8 +407,8 @@ public record EthTxData(
         }
 
         final var value = rlpList.get(6).asBytes().length == 0
-            ? BigInteger.ZERO
-            : new BigInteger(rlpList.get(6).asBytes());
+                ? BigInteger.ZERO
+                : new BigInteger(rlpList.get(6).asBytes());
 
         return new EthTxData(
                 rawTx,
@@ -420,7 +420,7 @@ public record EthTxData(
                 rlpList.get(3).data(), // maxGas
                 rlpList.get(4).asLong(), // gasLimit
                 rlpList.get(5).data(), // to
-            rlpList.get(6).asBigInt(), // value
+                value, // value
                 rlpList.get(7).data(), // callData
                 rlpList.get(8).data(), // accessList
                 rlpList.get(9).asByte(), // recId
@@ -446,8 +446,8 @@ public record EthTxData(
         }
 
         final var value = rlpList.get(5).asBytes().length == 0
-            ? BigInteger.ZERO
-            : new BigInteger(rlpList.get(5).asBytes());
+                ? BigInteger.ZERO
+                : new BigInteger(rlpList.get(5).asBytes());
 
         return new EthTxData(
                 rawTx,
@@ -459,7 +459,7 @@ public record EthTxData(
                 null, // maxGas
                 rlpList.get(3).asLong(), // gasLimit
                 rlpList.get(4).data(), // to
-            rlpList.get(5).asBigInt(), // value
+                value, // value
                 rlpList.get(6).data(), // callData
                 rlpList.get(7).data(), // accessList
                 rlpList.get(8).asByte(), // recId
--- a/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
+++ b/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
@@ -367,6 +367,10 @@ public record EthTxData(
             chainId = new byte[0];
         }
 
+        final var value = rlpList.get(4).asBytes().length == 0
+            ? BigInteger.ZERO
+            : new BigInteger(rlpList.get(4).asBytes());
+
         return new EthTxData(
                 rawTx,
                 EthTransactionType.LEGACY_ETHEREUM,
@@ -377,7 +381,7 @@ public record EthTxData(
                 null, // maxGas
                 rlpList.get(2).asLong(), // gasLimit
                 rlpList.get(3).data(), // to
-                rlpList.get(4).asBigInt(), // value
+            rlpList.get(4).asBigInt(), // value
                 rlpList.get(5).data(), // callData
                 null, // accessList
                 recId,
@@ -402,6 +406,10 @@ public record EthTxData(
             return null;
         }
 
+        final var value = rlpList.get(6).asBytes().length == 0
+            ? BigInteger.ZERO
+            : new BigInteger(rlpList.get(6).asBytes());
+
         return new EthTxData(
                 rawTx,
                 EthTransactionType.EIP1559,
@@ -412,7 +420,7 @@ public record EthTxData(
                 rlpList.get(3).data(), // maxGas
                 rlpList.get(4).asLong(), // gasLimit
                 rlpList.get(5).data(), // to
-                rlpList.get(6).asBigInt(), // value
+            rlpList.get(6).asBigInt(), // value
                 rlpList.get(7).data(), // callData
                 rlpList.get(8).data(), // accessList
                 rlpList.get(9).asByte(), // recId
@@ -437,6 +445,10 @@ public record EthTxData(
             return null;
         }
 
+        final var value = rlpList.get(5).asBytes().length == 0
+            ? BigInteger.ZERO
+            : new BigInteger(rlpList.get(5).asBytes());
+
         return new EthTxData(
                 rawTx,
                 EthTransactionType.EIP2930,
@@ -447,7 +459,7 @@ public record EthTxData(
                 null, // maxGas
                 rlpList.get(3).asLong(), // gasLimit
                 rlpList.get(4).data(), // to
-                rlpList.get(5).asBigInt(), // value
+            rlpList.get(5).asBigInt(), // value
                 rlpList.get(6).data(), // callData
                 rlpList.get(7).data(), // accessList
                 rlpList.get(8).asByte(), // recId
--- a/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
+++ b/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
@@ -539,24 +539,24 @@ class EthTxDataTest {
     void negativeValueEIP1559() {
         final var negativeValue = BigInteger.valueOf(-1000);
 
-        final var oneByte = new byte[]{1};
+        final var oneByte = new byte[] {1};
         final EthTxData ethTxData = new EthTxData(
-            oneByte,
-            EthTransactionType.EIP1559,
-            oneByte,
-            1,
-            oneByte,
-            oneByte,
-            oneByte,
-            1,
-            oneByte,
-            negativeValue,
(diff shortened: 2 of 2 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/EthereumCallDataHydration.java
```diff
@@ -74,9 +74,19 @@ public HydratedEthTxData tryToHydrate(
...
             }
+
+            final var initcodeBytes = callDataFile.contents().toByteArray();
+            // Bytes.fromHex() doesn't appreciate a leading '0x' but we supported it in mono-service
+            final String hexedInitcode;
+            if (initcodeBytes[0] == (byte) '0' && initcodeBytes[1] == (byte) 'x') {
+                hexedInitcode = new String(initcodeBytes, 2, initcodeBytes.length - 2);
+            } else {
+                hexedInitcode = new String(initcodeBytes);
+            }
```
vtronkov, line 85: This snipped is taken from `HevmTransactionFactory.initcodeFor`. The idea is that we have tests the call data is prefixed with 0x. To be consistent we are removing the prefix.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java
```diff
@@ -198,6 +198,7 @@ private HederaEvmTransaction fromHapiEthereum(
             @NonNull final AccountID senderId,
             @NonNull final EthTxData ethTxData,
             final long maxGasAllowance) {
+        validateTrue(ethTxData.getAmount() >= 0, CONTRACT_NEGATIVE_VALUE);
```
vtronkov, line 201: Missed validation

vtronkov:
> Are there any HAPI tests that should go along with this pr?

This bug was found at https://github.com/hashgraph/hedera-services/pull/12834. The `callFailsWhenAmountIsNegativeButStillChargedFee` HAPI test, when converted to Ethereum, is testing just that. So we have it covered.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -477,6 +477,12 @@ private static EthTxData populateEip2390EthTxData(RLPItem rlpItem, byte[] rawTx)
                 );
     }
 
+    // Wrapping the bytes in a BigInteger to handle when there is a negative value. If we use the RLPItem.asBigInt()
+    // method it doesn't handle two's complement correctly.
+    private static BigInteger toValue(final byte[] value) {
+        return value.length == 0 ? BigInteger.ZERO : new BigInteger(value);
```
tinker-michaelj, line 483: Should we then use the `BigInteger` constructor with an explicit `signum` (see [here](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/math/BigInteger.java#L481)):
```
return value.length == 0 ? BigInteger.ZERO : new BigInteger(1, value);
```

tinker-michaelj (approved):
LGTM, ty @vtronkov 1

lukelee-sl (dismissed):
LGTM.  Thanks for answering my question

MrValioBg (dismissed):
LGTM! 💯 

We just need to fix the  `failsWithInvalidFileIdOnEmptyCallDataFile` unit test

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -165,7 +165,7 @@ public byte[] encodeTx() {
                     gasPrice,
                     Integers.toBytes(gasLimit),
                     to,
-                    Integers.toBytesUnsigned(value),
```
vtronkov, line 168: There was a new bug when you pass a really big number(20 billion for example), we were wrapping it in **Integer** and then when we call `populateEthTxData` we were trying to unwrap it as **BigInteger**. To make it consistent I just used the value variable(which is **BigInteger**) to be consistent with `populateEthTxData`.

I wrote a new unit test to verify this and now the test-clients tests are passing.

lukelee-sl (approved):
LGTM. tyvm

tinker-michaelj (approved):
LGTM, ty @vtronkov !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/EthereumCallDataHydration.java
```diff
@@ -74,9 +74,19 @@ public HydratedEthTxData tryToHydrate(
             if (callDataFile.deleted()) {
                 return failureFrom(FILE_DELETED);
             }
+
+            final var initcodeBytes = callDataFile.contents().toByteArray();
```
david-bakin-sl, line 78: If you step back and look at L78-L92 what you're really trying to do is convert the file contents, which is in pbj's `Byte` class, to a `byte[]`.  And the only problem is that the first two bytes of the pbj-encoded byte array may in fact be the characters `0` and `x` - and, say, who the heck allowed that?

Oh well.  Anyway, all you _really_ want to do is take a slice of the original pbj `Bytes` iff the first two characters are the hex prefix, or else the entire array if it is not.  So why not just use the `Bytes.getBytes(long,long)` overload to do the slice (very cheap) and then `Bytes.toByteArray()` to get the contents as a `byte[]`?

```java

            final var hexPrefix = new byte[]{(byte)'0', (byte)'x'};
            final var contents = callDataFile.contents();
            final var offset = contents.matchesPrefix(hexPrefix) ? hexPrefix.length : 0L;
            final var len = contents.length() - offset;
            final var callData = contents.getBytes(offset,len).toByteArray();
```

You'll need to check that actually works though - it syntax checks but that doesn't mean it's correct.

david-bakin-sl (commented):
Some suggestions.

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -477,6 +477,12 @@ private static EthTxData populateEip2390EthTxData(RLPItem rlpItem, byte[] rawTx)
                 );
     }
 
+    // Wrapping the bytes in a BigInteger to handle when there is a negative value. If we use the RLPItem.asBigInt()
```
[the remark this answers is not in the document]
vtronkov: Sounds good to me, looks more readable. But is it okay to do this refactoring into a separate PR/story? I don't want to water down this PR much 😄

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -38,13 +38,15 @@
...
+import org.junit.jupiter.params.provider.EnumSource;
 
 class EthTxDataTest {
 
     static final String SIGNATURE_ADDRESS = "a94f5374fce5edbc8e2a8697c15331677e6ebf0b";
     static final String SIGNATURE_PUBKEY = "033a514176466fa815ed481ffad09110a2d344f6c9b78c1d14afc351c3a51be33d";
     static final String RAW_TX_TYPE_0 =
-            "f864012f83018000947e3a9eaf9bcc39e2ffa38eb30bf7a93feacbc18180827653820277a0f9fbff985d374be4a55f296915002eec11ac96f1ce2df183adf992baa9390b2fa00c1e867cc960d9c74ec2e6a662b7908ec4c8cc9f3091e886bcefbeb2290fb792";
+            "f864012f83018000947e3a9eaf9bcc39e2ffa38eb30bf7a93feacbc18100827653820277a0f9fbff985d374be4a55f296915002eec11ac96f1ce2df183adf992baa9390b2fa00c1e867cc960d9c74ec2e6a662b7908ec4c8cc9f3091e886bcefbeb2290fb792";
     static final String RAW_TX_TYPE_0_TRIMMED_LAST_BYTES =
```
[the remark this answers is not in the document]
vtronkov: Well, I think the name should be the same. The change was needed because I changed the way `EthTxData.populateEthTxData` handles the **value** parsing. It wasn't handling negative values well. You can see `EthTxData.toValue`.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -78,7 +80,7 @@ class EthTxDataTest {
             "f86c098504a817c800825208943535353535353535353535353535353535353535880de0b6b3a76400008025a028ef61340bd939bc2195fe537567866003e1a15d3c71ff63e1590620aa636276a067cbe9d8997f761aecb703304b3800ccf555c9f3dc64214b297fb1966a3b6d83";
     // v = 37
     static final String EIP155_UNPROTECTED =
```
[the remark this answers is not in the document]
vtronkov: Done 👍

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/EthereumCallDataHydration.java
vtronkov: I just had to wrap the `callData` in `Hex.decode`, other then that it works perfectly 💯

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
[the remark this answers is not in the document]
david-bakin-sl: oh yeah, it's an idea for the future

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -38,13 +38,15 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
 
 class EthTxDataTest {
 
     static final String SIGNATURE_ADDRESS = "a94f5374fce5edbc8e2a8697c15331677e6ebf0b";
     static final String SIGNATURE_PUBKEY = "033a514176466fa815ed481ffad09110a2d344f6c9b78c1d14afc351c3a51be33d";
     static final String RAW_TX_TYPE_0 =
-            "f864012f83018000947e3a9eaf9bcc39e2ffa38eb30bf7a93feacbc18180827653820277a0f9fbff985d374be4a55f296915002eec11ac96f1ce2df183adf992baa9390b2fa00c1e867cc960d9c74ec2e6a662b7908ec4c8cc9f3091e886bcefbeb2290fb792";
+            "f864012f83018000947e3a9eaf9bcc39e2ffa38eb30bf7a93feacbc18100827653820277a0f9fbff985d374be4a55f296915002eec11ac96f1ce2df183adf992baa9390b2fa00c1e867cc960d9c74ec2e6a662b7908ec4c8cc9f3091e886bcefbeb2290fb792";
```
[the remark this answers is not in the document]
david-bakin-sl: too subtle for me!

david-bakin-sl (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/EthereumCallDataHydration.java
```diff
@@ -74,15 +74,23 @@ public HydratedEthTxData tryToHydrate(
             if (callDataFile.deleted()) {
                 return failureFrom(FILE_DELETED);
             }
+
+            // Bytes.fromHex() doesn't appreciate a leading '0x' but we supported it in mono-service
```
tinker-michaelj, line 78: Mmm is this duplicated from [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/HevmTransactionFactory.java#L356)? Would prefer extracting a reusable method. 😇

tinker-michaelj (approved):
LGTM, tyvm @vtronkov !

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/infra/EthereumCallDataHydration.java
vtronkov: Created a tech debt story: https://github.com/hashgraph/hedera-services/issues/13226

PR #12839, opened 2024-04-16
feat: Enhance purechecks for CryptoGetAccountBalanceHandler

### LINKED ISSUE #12715: CryptoGetAccountBalanceHandler

thomas-swirlds-labs:
**Description**:
Adding additional pure checks to CryptoGetAccountBalanceHandler and associated Unit and Hapi Tests.

**Related issue(s)**:

Fixes #12715 

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java
```diff
@@ -82,19 +86,47 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
         final var query = context.query();
         final var accountStore = context.createStore(ReadableAccountStore.class);
         final CryptoGetAccountBalanceQuery op = query.cryptogetAccountBalanceOrThrow();
+        mustExist(op.balanceSource(), INVALID_CONTRACT_ID);
```
mhess-swl, line 89: Is this error code correct? `op.balanceSource()` can be `ACCOUNT_ID` or `CONTRACT_ID`, so it seems misleading to always use `INVALID_CONTRACT_ID`. Especially since this endpoint is called _CryptoGetACCOUNTBalanceHandler_. I would expect we default to `INVALID_ACCOUNT_ID` like in the `UNSET` case just below.

EDIT: `validateContractId(...)` and `validateAccountId(...)` both call `mustExist(...)`, so I think we can remove this line here

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java
```diff
@@ -82,19 +86,47 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
...
+            validateContractId(op, accountStore);
         } else {
             throw new PreCheckException(INVALID_ACCOUNT_ID);
         }
     }
 
+    private void validateContractId(CryptoGetAccountBalanceQuery op, ReadableAccountStore accountStore)
+            throws PreCheckException {
+        mustExist(op.contractID(), INVALID_CONTRACT_ID);
+        ContractID contractId = (ContractID) op.balanceSource().value();
```
mhess-swl, line 105: nit: `final`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java
```diff
@@ -82,19 +86,47 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
...
+        validateTruePreCheck(contractId.shardNum() == 0, INVALID_CONTRACT_ID);
+        validateTruePreCheck(contractId.realmNum() == 0, INVALID_CONTRACT_ID);
+        validateTruePreCheck(contractId.contractNumOrThrow() >= 0, INVALID_CONTRACT_ID);
+        final var contract = accountStore.getContractById(requireNonNull(op.contractID()));
+        validateFalsePreCheck(contract == null || !contract.smartContract(), INVALID_CONTRACT_ID);
```
mhess-swl, line 110: Personally I find calling this with compound boolean expressions like so – `validateFalse(<condition 1> || <condition2> etc.)` – very hard to read, since it requires negating the expression with boolean algebra in order to decipher the code intent. 

```
validateFalsePreCheck(contract == null, INVALID_CONTRACT_ID);
validateTruePreCheck(contract.smartContract(), INVALID_CONTRACT_ID);
```

Just a thought though, nothing inherently incorrect about this afaict

thomas-swirlds-labs: Makes sense 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java
```diff
@@ -82,19 +87,42 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
         final var query = context.query();
         final var accountStore = context.createStore(ReadableAccountStore.class);
         final CryptoGetAccountBalanceQuery op = query.cryptogetAccountBalanceOrThrow();
+        if (op.balanceSource().kind().equals(UNSET)) {
+            throw new PreCheckException(INVALID_ACCOUNT_ID);
+        }
```
Neeharika-Sompalli, line 92: Why not use `!op.hasAccountId`?

mhess-swl (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java
thomas-swirlds-labs: Makes sense, thanks Neeha

Neeharika-Sompalli (approved):

mhess-swl (approved):

PR #12842, opened 2024-04-16
fix: internal calls to system account result in wrong behaviour

petreze:
**Description**:
This PR fixes the behaviour whenever we make internal calls to a system account both with and without sending value

**Related issue(s)**:

Fixes 

**Notes for reviewer**:
The expected behaviour is:
- if we try to send value over the call - this should result in `INVALID_FEE_SUBMITTED`
- if we do not send value - the behaviour should be a no-op success

**Checklist**


Neeharika-Sompalli (dismissed):
LGTM only from `hedera-base` changes

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
```diff
@@ -167,6 +169,8 @@ private long effectiveGasUsed(final long gasLimit, @NonNull final MessageFrame f
     private void propagateHaltException(MessageFrame frame, ExceptionalHaltReason haltReason) {
         if (haltReason.equals(INSUFFICIENT_CHILD_RECORDS)) {
             setPropagatedCallFailure(frame, RESULT_CANNOT_BE_EXTERNALIZED);
+        } else if (haltReason.equals(INVALID_FEE_SUBMITTED)) {
```
lukelee-sl, line 172: Question - is the propagation of the `INVALID_FEE_SUBMITTED` necessary for mono service fidelity?

lukelee-sl (approved):
Seems fine to me.  Just a question and a suggestion.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
petreze: It is necessary for fixing the behaviour. This bug was found in mono but the fix will be only in modu codebase, from my understanding

lukelee-sl (approved):
LGTM

Neeharika-Sompalli (approved):

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/processors/CustomMessageCallProcessorTest.java
```diff
@@ -171,7 +174,7 @@ void valueCannotBeTransferredToSystemContracts() {
         givenHaltableFrame(isHalted);
         givenCallWithCode(ADDRESS_6);
         given(addressChecks.isSystemAccount(ADDRESS_6)).willReturn(true);
-        given(registry.get(ADDRESS_6)).willReturn(nativePrecompile);
+        lenient().when(registry.get(ADDRESS_6)).thenReturn(nativePrecompile);
```
MrValioBg, line 177: Why do we want to use lenient() instead of just removing the stubbing?

MrValioBg: It seems like it was necessary before because `(precompiles.get(codeAddress) == null)` was getting called in `doHaltIfInvalidSystemCall`. But this is no longer the case.

petreze: Agreed, it is still called though but we reversed the statements so its not needed, yup

Neeharika-Sompalli (approved):
LGTM from `hedera-base`

MrValioBg (dismissed):
LGTM!

lukelee-sl (dismissed):
LGTM

jeromy-cannon (dismissed):

Neeharika-Sompalli (dismissed):
LGTM from `hedera-base`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
```diff
@@ -167,6 +169,8 @@ private long effectiveGasUsed(final long gasLimit, @NonNull final MessageFrame f
...
+            setPropagatedCallFailure(frame, HevmPropagatedCallFailure.INVALID_FEE_SUBMITTED);
```
david-bakin-sl, line 173: L171 and L173 should either _both_ have the `HevmPropagatedCallFailure.` qualification, or _neither_.  (I prefer both.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/failure/CustomExceptionalHaltReason.java
```diff
@@ -76,6 +76,8 @@ public static ResponseCodeEnum statusFor(@NonNull final ExceptionalHaltReason re
             return ResponseCodeEnum.MAX_CHILD_RECORDS_EXCEEDED;
         } else if (reason == CustomExceptionalHaltReason.INVALID_CONTRACT_ID) {
             return ResponseCodeEnum.INVALID_CONTRACT_ID;
+        } else if (reason == CustomExceptionalHaltReason.INVALID_FEE_SUBMITTED) {
```
david-bakin-sl, line 79: At some point in time, as part of a general code cleanup, this if-ladder should be replaced by something more directly readable.  E.g., each enum `CustomExceptionalHaltReason` could have a field which is its mapped status.  Then the mapped status would be, syntactically, right next to the declaration of its enum value, and thus, the whole thing would be easier to read.  

SImilarly for `errorMessageFor`.  That one is not too complicated, but still, data-driven is better, IMO.

For the future.  (Perhaps you could add a `// FUTURE:` comment around here so we can find it later when we pass through the contract service looking for cleanups.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HevmPropagatedCallFailure.java
```diff
@@ -41,7 +41,11 @@ public enum HevmPropagatedCallFailure {
     /**
      * The call failed because its externalizing its result would exceed the maximum number of child records.
      */
-    RESULT_CANNOT_BE_EXTERNALIZED(INSUFFICIENT_CHILD_RECORDS);
+    RESULT_CANNOT_BE_EXTERNALIZED(INSUFFICIENT_CHILD_RECORDS),
+    /**
+     * The call failed because invalid fee was submitted for an EVM call
+     */
+    INVALID_FEE_SUBMITTED(CustomExceptionalHaltReason.INVALID_FEE_SUBMITTED);
```
david-bakin-sl, line 48: Here again inconsistency in qualification.  I personally like having the `CustomExceptionalHaltReason.` qualifier on all these enums, but either way is ok as long as it's consistent.

Oh, whoops, I see there's no choice here due to name clash.  So ... I suggest (mildly) that the other two enum's argument be qualified as well.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/processors/CustomMessageCallProcessorTest.java
```diff
@@ -159,10 +160,11 @@ void callsToNonStandardSystemContractsAreNotSupported() {
         givenHaltableFrame(isHalted);
         givenCallWithCode(NON_EVM_PRECOMPILE_SYSTEM_ADDRESS);
         given(addressChecks.isSystemAccount(NON_EVM_PRECOMPILE_SYSTEM_ADDRESS)).willReturn(true);
+        when(frame.getValue()).thenReturn(Wei.of(2L));
```
david-bakin-sl, line 163: The _next_ test does a call-to-system-contract-with-value-transfer.  This test should be the call-to-system-contract-**without**-value-transfer one, right?  So it should mock `frame.getValue()` with `Wei.ZERO` and check for no-op success, I think?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCallSuite.java
```diff
@@ -364,7 +365,7 @@ final HapiSpec lowLevelEcrecCallBehavior() {
         final var TEST_CONTRACT = "TestContract";
         final var somebody = "somebody";
         final var account = "0.0.1";
-        return defaultHapiSpec("LowLevelEcrecCallBehavior", NONDETERMINISTIC_TRANSACTION_FEES)
+        return defaultHapiSpec("LowLevelEcrecCallBehavior", FULLY_NONDETERMINISTIC)
```
david-bakin-sl, line 368: Just for my information, what additional laxity of `FULLY_NONDETERMINISTIC` (over `NONDETERMINISTIC_TRANSACTION_FEES`) is needed by this test?  And also the next test?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
```diff
@@ -1697,7 +1696,7 @@ final HapiSpec traceabilityE2EScenario4() {
     @HapiTest
     @Order(6)
     final HapiSpec traceabilityE2EScenario5() {
-        return defaultHapiSpec("traceabilityE2EScenario5", NONDETERMINISTIC_FUNCTION_PARAMETERS)
+        return defaultHapiSpec("traceabilityE2EScenario5", FULLY_NONDETERMINISTIC)
```
david-bakin-sl, line 1699: Again, for my own information, why are these tests suddenly needing `FULLY_NONDETERMINISTIC` over `NONDETERMINISTIC_FUNCTION_PARAMETERS`?   Since only a couple of them do, the rest are happy with `NONDETERMINISTIC_FUNCTION_PARAMETERS` and `NONDETERMINISTIC_TRANSACTION_FEES` together.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/NonceSuite.java
```diff
@@ -1028,7 +1030,7 @@ HapiSpec nonceUpdatedAfterEvmReversionDueInsufficientTransferAmountEthContractCr
     HapiSpec nonceUpdatedAfterEvmReversionDueSendingValueToEthereumPrecompileEthContractCreate() {
         return defaultHapiSpec(
                         "nonceUpdatedAfterEvmReversionDueSendingValueToEthereumPrecompileEthContractCreate",
-                        NONDETERMINISTIC_ETHEREUM_DATA)
+                        FULLY_NONDETERMINISTIC)
```
david-bakin-sl, line 1033: Same question w.r.t. nondeterminism flag.

david-bakin-sl (commented):
One fix needed to a test, some style issues you might address, and some questions you might answer for me.

Also!!! Please consider fixing up `system-accounts-operations.md` as we discussed in slack, specifically:

a) In the section "Call operations (CallOp, ...)" move the "_Please note that the expected behavior ..._" sentence to right up under the section title, and
b) Change the section title to something like "Call operations _that do not transfer value_ (CallOp, ...)"

Both suggestions aimed at making that important qualification to this section more difficult to miss.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/processors/CustomMessageCallProcessorTest.java
petreze: Good catch, ty

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCallSuite.java
petreze: @david-bakin-sl 
Those inconsistencies in regards to the issues we are fixing are found in mono. And since the fuzzy matching is making a comparison between mono and modu results, they won't match because of the change in the behaviour. And as I'm aware the fuzzy match test itself was already successful and it is not a present thing that we are trying to be compatible with, if that makes sense..

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/traceability/TraceabilitySuite.java
petreze: Answered above

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/NonceSuite.java
petreze: Answered above

hedera-node/docs/system-accounts-operations.md
```diff
@@ -107,7 +107,9 @@ Legend of the symbols used in the tables below:
 | 0x3E8 → ♾️                           |                     ✅                      | ✅ - if account exists and has receiverSigRequired = false<br/>✅ - if (account exists) and (account has receiverSigRequired = true) and (account is sender)<br/>❌ - some failing scenarios described above |
 | beneficiary same as contract address | ✅ burns the eth and destructs the contract | ❌ - fails with SELF_DESTRUCT_TO_SELF                                                                                                                                                                       |
 
-### Call operations (CallOp, DelegateCallOp, CallCodeOp, StaticCallOp)
+### Call operations that do not transfer value (CallOp, DelegateCallOp, CallCodeOp, StaticCallOp)
+
+_Please note that the expected behavior above is valid considering there is no `value` passed. If there is `value`, this falls in the next section ("Transfer and send operations")._
```
david-bakin-sl, line 112: "... expected behavior **described in this section** is valid **if** there is no `value` passed" <- suggested rewording

david-bakin-sl (approved):
LGTM!  (with one minor suggestion if you feel like it)

david-bakin-sl (approved):
LGTM!

MrValioBg (approved):

netopyr (approved):
LGTM in regards to hedera-services

PR #12859, opened 2024-04-17
fix: unit test marker file writing collision

### LINKED ISSUE #12857: marker file code broke consensusTest
https://scans.gradle.com/s/gfrtkamtvpyjq/tests/task/:swirlds-platform-test:test/details/com.swirlds.platform.test.consensus.ConsensusTests/forkingTests(ConsensusTestParams)%5B2%5D?top-execution=1

edward-swirldslabs:
**Description**:
* added file system test for existence of marker file before writing.  
* added atomic boolean flags for when to ignore marker file checks at the end of consensus tests.  

Fixes #12857

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
@@ -107,16 +107,12 @@ public class MarkerFileWriter {
         markerFilesWritten.computeIfAbsent(filename, key -> {
             final Path markerFile = markerFileDirectory.resolve(filename);
             try {
-                // Prevent any other writer for the same string to check and write the file at the same time.
-                // The cost of acquiring the lock is no worse than the file system calls.
-                synchronized (filename) {
-                    if (Files.exists(markerFile)) {
-                        failedToWriteMarkerFileLogger.error(
-                                LogMarker.ERROR.getMarker(), "Marker file already exists: {}", markerFile);
-                        return true;
-                    }
-                    Files.createFile(markerFile);
+                if (Files.exists(markerFile)) {
+                    failedToWriteMarkerFileLogger.error(
+                            LogMarker.ERROR.getMarker(), "Marker file already exists: {}", markerFile);
+                    return true;
                 }
+                Files.createFile(markerFile);
             } catch (final IOException e) {
                 failedToWriteMarkerFileLogger.error(
                         LogMarker.EXCEPTION.getMarker(), "Failed to create marker file: {}", markerFile, e);
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
@@ -107,12 +107,16 @@ public class MarkerFileWriter {
         markerFilesWritten.computeIfAbsent(filename, key -> {
             final Path markerFile = markerFileDirectory.resolve(filename);
             try {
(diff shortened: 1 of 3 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/MarkerFileWriter.java
```diff
@@ -107,7 +107,16 @@ public void writeMarkerFile(@NonNull final String filename) {
         markerFilesWritten.computeIfAbsent(filename, key -> {
             final Path markerFile = markerFileDirectory.resolve(filename);
             try {
-                Files.createFile(markerFile);
+                // Prevent any other writer for the same string to check and write the file at the same time.
+                // The cost of acquiring the lock is no worse than the file system calls.
+                synchronized (filename) {
```
cody-littley, line 112: This makes me a bit uneasy. If we are using shared static constants to define the file name strings then I think it will prevent parallel writing... but this is something that seems fragile and is not part of the documented API for this method.

A possible alternate approach might be to accept the fact that there is a race condition, and to log an error in the logs if we fail to write the marker file due to the race condition.

cody-littley: If there is concern with tests writing marker files in parallel, I think we should mark these unit tests as `@TimingSensisitive`.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -44,16 +45,26 @@ class ConsensusTests extends PlatformTest {
      */
     private final int NUM_ITER = 1;
 
+    private AtomicBoolean ignoreNoSuperMajorityMarkerFile = new AtomicBoolean(false);
```
alittley, line 48: I'm not grokking why these need to be atomic

If the atomic is necessary, though, the members may be `final`

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
```diff
@@ -131,6 +142,9 @@ void forkingTests(final ConsensusTestParams params) {
                 .setParams(modifyParams(params))
                 .setIterations(NUM_ITER)
                 .run();
+        // Some forking tests make too many forkers.  When there is  > 1/3 nodes forking, no super majority can result.
```
alittley, line 145: According to Leemon's slack message, it seems it would also be possible to have no judges at all. So theoretically, that marker file too should be ignored here (though in practice, it seems unlikely that that would arise)

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
edward-swirldslabs: I've made  it a normal boolean since all methods are run in separate instances of the class.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/consensus/ConsensusTests.java
edward-swirldslabs: Adding

alittley (approved):

cody-littley (approved):

PR #12748, opened 2024-04-11
fix: refactor pure checks smart contracts network admin

kimbor:
**Description**:
Refactor pureChecks for Smart Contracts and Network Admin services. Move any checks which do not rely on state from preHandle and handle methods to pureChecks method for each transaction handler.

**Related issue(s)**:
Fixes #12689
Related to #12644

PR #12874, opened 2024-04-18
fix: Revert pureChecks exception thrown in unchecked submit

mhess-swl:
The exception newly added to the unchecked submit's `pureChecks` method is causing some hapi tests to fail. Until we can get the tests updated, or perhaps permanently, this commit returns the handler behavior to its previous state.

For reference, the exception was added in PR #12748


netopyr (approved):
LGTM - thanks @mhess-swl

PR #12878, opened 2024-04-18
chore: Fix Restart test

Neeharika-Sompalli:
- Added a condition that was missing in pureChecks. 
- [This PR](https://github.com/hashgraph/hedera-services/pull/12748/files) moved a condition to pureChecks , but missed to add a condition that should be executed only for FREEZE_UPGRADE, TELEMETRY_UPGRADE, PREPARE_UPGRADE


povolev15 (approved):
LGTM ty @Neeharika-Sompalli

PR #12881, opened 2024-04-18
perf: synchronous signature verification

### LINKED ISSUE #12237: SignatureVerifierImpl.verify() should not call verifyAsync()
### Problem

`SignatureVerifierImpl.verify()` is invoked in threads handling client API calls to verify the payer signature (IngestChecker.verifyPayerSignature()). The actual verification is delegated to a separate thread pool via `Cryptography.verifyAsync()`. The overhead of passing the call from a larger pool to a smaller pool and then waiting for the result in `SignatureVerificationFutureImpl.get()` is not justified. Client thread throughput drops significantly.
A similar case can be made for `PreHandleWorkflowImpl.verifySignatures()`.

### Solution

Verify signatures synchronously in the calling thread.

### Alternatives

_No response_

netopyr:
**Description**:

This PR changes signature verification to be synchronous

**Related issue(s)**:

Fixes #12237


hedera-node/hedera-app/src/main/java/com/hedera/node/app/signature/impl/SignatureVerifierImpl.java
```diff
@@ -75,8 +74,7 @@ public Map<Key, SignatureVerificationFuture> verify(
         }
 
         // Gather each TransactionSignature to send to the platform and the resulting SignatureVerificationFutures
-        final var platformSigs = new ArrayList<TransactionSignature>(sigs.size());
-        final var futures = new HashMap<Key, SignatureVerificationFuture>(sigs.size());
+        final var futures = HashMap.<Key, SignatureVerificationFuture>newHashMap(sigs.size());
```
Neeharika-Sompalli, line 77: I think using HashMap here is fine.

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12887, opened 2024-04-19
perf: Cherry-pick: synchronous signature verification

### LINKED ISSUE #12237: SignatureVerifierImpl.verify() should not call verifyAsync()

netopyr:
**Description**:
Cherry-pick of  #12881

**Related issue(s)**:

Fixes #12237


Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #12811, opened 2024-04-15
feat: auto-resubmit operations with modifications

tinker-michaelj:
**Description**:
 - Closes #12773 
 - _TBD_

PR #12900, opened 2024-04-19
fix: cherry-pick main fixes from #12811

Neeharika-Sompalli:
Cherry-pick main fixes from #12811

### DIFF
--- a/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusDeleteTopicHandler.java
+++ b/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusDeleteTopicHandler.java
@@ -66,6 +66,7 @@ public class ConsensusDeleteTopicHandler implements TransactionHandler {
         final var op = context.body().consensusDeleteTopicOrThrow();
         final var topicStore = context.createStore(ReadableTopicStore.class);
         // The topic ID must be present on the transaction and the topic must exist.
+        mustExist(op.topicID(), INVALID_TOPIC_ID);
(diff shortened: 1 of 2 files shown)


tinker-michaelj (approved):

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusDeleteTopicHandler.java
```diff
@@ -66,6 +66,7 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         final var op = context.body().consensusDeleteTopicOrThrow();
         final var topicStore = context.createStore(ReadableTopicStore.class);
         // The topic ID must be present on the transaction and the topic must exist.
+        mustExist(op.topicID(), INVALID_TOPIC_ID);
```
iwsimon, line 69: If op.topicID is null, topic will be null on line 70 because we used TopicID.DEFAULT instead. You will get INVALID_TOPIC_ID.

Neeharika-Sompalli: But we should fail early and not look at state with `TopicID.DEFAULT` if `TopicID` is not set.  But I agree it would have already failed in `pureChecks`

netopyr: Interesting point, @iwsimon. I will resolve here to be able to merge. Please continue the discussion here: #12811

PR #12908, opened 2024-04-19
chore: 11772 Refactored `AddressBook` - it's no longer a `MerkleLeaf`

### LINKED ISSUE #11772: Refactor `AddressBook`
Currently, `AddressBook` is an instance of `MerkleLeaf`. However, it's a part of `PlatformState` and it no longer needs to be an instance of `MerkleLeaf` as it's not used as such. The only thing that it needs to be is `SelfSerializable` and `Hashable`.

imalygin:
**Description**:

This is a clean up. `AddressBook` objects are no longer used as `MerkleLeaf` objects, because `AddressBook` is now a part of `PlatformState` object which, in it's turn, is a  `MerkleLeaf`. 

So, `AddressBook` should no longer implement this interface. It only needs to be  `Iterable`, `SelfSerializable`, and Hashable.

**Related issue(s)**:

Fixes #11772

### DIFF
--- a/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/migration/StateChildIndices.java
+++ b/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/migration/StateChildIndices.java
@@ -28,7 +28,6 @@ public final class StateChildIndices {
     public static final int SPECIAL_FILES = 7;
     public static final int SCHEDULE_TXS = 8;
     public static final int RECORD_STREAM_RUNNING_HASH = 9;
-    public static final int LEGACY_ADDRESS_BOOK = 10;
     public static final int CONTRACT_STORAGE = 11;
     public static final int STAKING_INFO = 12;
     public static final int PAYER_RECORDS_OR_CONSOLIDATED_FCQ = 13;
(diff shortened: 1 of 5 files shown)


cody-littley (approved):

edward-swirldslabs (commented):
A couple observations: 
1. the removal of the leaf index for `LEGACY_ADDRESS_BOOK` is happening in the mono repo.  If this change goes through, does something similar need to happen in the modular repo? 
2. How does this change interact with the Dynamic Address Book  (DAB) effort?   The existing `AddressBook` construct is becoming the `ConsensusRoster` and Hedera will move forward with their own definition of what an `AddressBook` means to them.   Would that new concept of a Hedera `AddressBook` live at this index that is being dropped?  

@iwsimon

imalygin:
@edward-swirldslabs 

1. Correct me if I'm wrong, but, as far as I can tell, is not used as a part of the Merkle Tree. It is still there just because it was not removed, and this PR is removing it. There is another Merkle leaf - `PlatformState` - and `AddressBook` is a part of that. 
2. From what I can tell, DAB effort is orthogonal to this change as it doesn't require the `AddressBook` object to be a part of the Merkle tree.

edward-swirldslabs (approved):

cody-littley (approved):

edward-swirldslabs (approved):

iwsimon (approved):
LGTM.

Neeharika-Sompalli (approved):
LGTM!

PR #12911, opened 2024-04-20
fix: manage `StakingInfos` in restart

### LINKED ISSUE #12910: `onUpdateWeight()` uses uninitialized states API
### Description

Because `Schema`s are not registered until `SwirldState.init()` is called, the `onUpdateWeight()` lifecycle method cannot use the "states API".

### Steps to reproduce

Upgrade 0.49 software.

### Additional context

_No response_

### Hedera network

previewnet

### Version

`develop`

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12910

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/HederaLifecyclesImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/HederaLifecyclesImpl.java
@@ -26,15 +26,14 @@ import com.hedera.node.app.service.mono.state.adapters.MerkleMapLike;
 import com.hedera.node.app.service.mono.state.merkle.MerkleStakingInfo;
 import com.hedera.node.app.service.mono.state.migration.StateChildIndices;
 import com.hedera.node.app.service.mono.utils.EntityNum;
-import com.hedera.node.app.service.token.ReadableStakingInfoStore;
 import com.hedera.node.app.service.token.TokenService;
-import com.hedera.node.app.spi.state.WritableKVState;
-import com.hedera.node.app.spi.state.WritableKVStateBase;
 import com.hedera.node.app.state.merkle.HederaLifecycles;
 import com.hedera.node.app.state.merkle.MerkleHederaState;
-import com.hedera.node.app.workflows.dispatcher.ReadableStoreFactory;
+import com.hedera.node.app.state.merkle.disk.OnDiskKey;
+import com.hedera.node.app.state.merkle.disk.OnDiskValue;
 import com.swirlds.common.context.PlatformContext;
 import com.swirlds.common.platform.NodeId;
+import com.swirlds.common.threading.manager.AdHocThreadManager;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.merkle.map.MerkleMap;
 import com.swirlds.platform.state.PlatformState;
@@ -44,6 +43,8 @@ import com.swirlds.platform.system.Round;
 import com.swirlds.platform.system.SoftwareVersion;
 import com.swirlds.platform.system.address.AddressBook;
 import com.swirlds.platform.system.events.Event;
+import com.swirlds.virtualmap.VirtualMap;
+import com.swirlds.virtualmap.VirtualMapMigration;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


Neeharika-Sompalli (approved):
Makes sense to me. Thanks @tinker-michaelj  🙏 . We should let @tomzhenghedera  verify this as well :)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -184,8 +189,44 @@ public void setStakingFs(
...
+     * @param ctx {@link MigrationContext} for this schema restart operation
+     */
     @Override
-    public void restart(@NonNull MigrationContext ctx) {}
+    public void restart(@NonNull MigrationContext ctx) {
+        // We need to validate and mark any node that are removed during upgrade as deleted.
+        // Since restart is called in the schema after an upgrade, and we don't want to depend on
+        // schema version change, validate all the nodeIds from the addressBook in state and mark
+        // them as deleted if they are not yet deleted in staking info.
+        final var stakingToState = ctx.newStates().<EntityNumber, StakingNodeInfo>get(STAKING_INFO_KEY);
```
Neeharika-Sompalli, line 208: I think all schemas will be registered by this point. So should not be an issue 🤔

Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj

tomzhenghedera (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -184,8 +190,47 @@ public void setStakingFs(
...
+                }
+            });
+        }
+        // Validate if any new nodes are added in addressBook and not in staking info.
+        // If so, add them to staking info/ with weight 0. Also update maxStake and
+        // minStake for the new nodes.
+        addAdditionalNodesIfAny(newStakingInfos, networkInfo.addressBook(), ctx.configuration());
+
+        if (newStakingInfos.isModified()) {
+            ((WritableKVStateBase<EntityNumber, StakingNodeInfo>) newStakingInfos).commit();
```
netopyr, line 231: This looks like a smell. We should never commit in a service.

Do you think this is even necessary? In `MerkleSchemaRegistry`, we call `commit()` right after calling `restart()` anyway.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -184,8 +190,47 @@ public void setStakingFs(
...
+     *     <li>For any node in the address book that is not in state, initializes its staking info.</li>
+     * </ol>
+     *
+     * @param ctx {@link MigrationContext} for this schema restart operation
+     */
     @Override
-    public void restart(@NonNull MigrationContext ctx) {}
+    public void restart(@NonNull MigrationContext ctx) {
+        final var networkInfo = ctx.networkInfo();
+        final var newStakingInfos = ctx.newStates().<EntityNumber, StakingNodeInfo>get(STAKING_INFO_KEY);
```
netopyr, line 206: We should work with the store here, IMO, and not with `WritableKVState`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
```diff
@@ -604,4 +650,37 @@ private void initializeNetworkRewards(@NonNull final MigrationContext ctx) {
...
+    private void addAdditionalNodesIfAny(
+            @NonNull final WritableKVState<EntityNumber, StakingNodeInfo> stakingInfos,
+            @NonNull final List<NodeInfo> nodeInfos,
+            @NonNull final Configuration config) {
+        final var numberOfNodesInAddressBook = nodeInfos.size();
+        final long maxStakePerNode =
+                config.getConfigData(LedgerConfig.class).totalTinyBarFloat() / numberOfNodesInAddressBook;
+        final var numRewardHistoryStoredPeriods =
+                config.getConfigData(StakingConfig.class).rewardHistoryNumStoredPeriods();
+        final var maxStakeHasChanged = stakingInfos.size() != numberOfNodesInAddressBook;
```
netopyr, line 663: I think `stakingInfos.size()` also counts deleted nodes, right? Maybe instead of this flag, we can check if `maxStakePerNode` is different from what is stored.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/HederaLifecyclesImpl.java
```diff
@@ -92,56 +125,40 @@ public void onUpdateWeight(
...
-                            EntityNumber.newBuilder().number(id.id()).build(),
-                            stakingInfo.copyBuilder().deleted(true).weight(0).build());
-                    logger.info(
-                            "Node {} is deleted from configAddressBook during upgrade "
-                                    + "and is marked deleted in state",
-                            id);
-                }
+        final var nodeIdsLeftToUpdate = configAddressBook.getNodeIdSet();
+        if (!isMonoState) {
+            final var stakingInfoIndex = state.findNodeIndex(TokenService.NAME, STAKING_INFO_KEY);
```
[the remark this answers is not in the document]
tinker-michaelj: The challenge here is that the _updateWeight()_ lifecycle is called **before** `init()`, which means that no `Schema`'s are registered and hence the States API is unavailable.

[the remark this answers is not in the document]
tinker-michaelj: As above...all `{Readable,Writable}States` are "empty" at this point in the lifecycle because no `Schema`'s have been registered.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
tinker-michaelj: Agreed, should be unnecessary, removed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/InitialModServiceTokenSchema.java
tinker-michaelj: Fixed.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/HederaLifecyclesImpl.java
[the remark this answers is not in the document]
netopyr: Ouch! I always thought the main point of an `init()` method is to initialize before anything else happens. 😃 

I guess it makes no sense to change this in the current design, but we should fix it when introducing the dynamic address book.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/TokenContextImpl.java
```diff
@@ -143,4 +149,11 @@ public boolean isScheduleDispatch() {
     public void markMigrationRecordsStreamed() {
         blockRecordManager.markMigrationRecordsStreamed();
     }
+
+    @Override
+    public Set<Long> knownNodeIds() {
+        return new ReadableStoreFactory(state)
```
netopyr, line 155: I think we can use `stack` here. Both will return the same `Set`, but I think using the root state and stack in the same class risks the introduction of hard-to-spot bugs in the future.

netopyr (approved):
LGTM - thanks @tinker-michaelj

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

iwsimon (approved):
LGTM. Thanks @tinker-michaelj

netopyr (approved):
LGTM - thanks @tinker-michaelj

tomzhenghedera (approved):

Issue #11678, opened 2024-02-22
Add more test cases for using ECDSA keys that are corresponding to a hollow account


MiroslavGatsanoga:
Add more test cases that use ECDSA keys corresponding to an existing hollow account alias in different places e.g. supply keys, admin keys, etc.

Example:
Token create which supply key is the same ECDSA key corresponding to a hollow account alias.

Note:
1. We have a similar `aliasCanBeUsedOnManyAccountsNotAsAlias` spec that is for auto-creations instead of lazy-creations
2. We have something related to lazy-creation in `accountDeletionResetsTheAliasNonce` but the spec is focused on asserting nonces not if we can use keys corresponding to hollow account aliases

PR #12916, opened 2024-04-22
test: Add hapi tests for ECDSA hollow account alias as different types of keys

bilyana-gospodinova:
**Description**:
This PR adds test cases that use ECDSA keys corresponding to an existing hollow account alias in places like:
- different types of keys on token create, e.g. admin key, supply key, etc.
- topic admin and submit keys.
Each test creates an ECDSA key, a hollow account with this public key as alias, sets the same key according to the scenario above and verifies if the key is usable.

Fixes https://github.com/hashgraph/hedera-services/issues/11678


Neeharika-Sompalli (approved):

kselveliev (approved):

anastasiya-kovaliova (dismissed):

kselveliev (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM !

PR #12918, opened 2024-04-22
fix: `computeFees` NPE in multiple queries

### LINKED ISSUE #12672: `getFileInfo` query can fail with `NullPointer`
### Description

There are cases where `getFileInfo` can cause `NullPointerException`

### Steps to reproduce

-

### Additional context

_No response_

### Hedera network

other

### Version

0.48

### Operating system

None

natanasow:
**Description**:

There is a problem in `QueryWorkflowImpl.handleQuery()`. The current implementation flow for paymentRequired queries is the following:
- run all checks via `ingestChecker.runAllChecks`
- get the payerID via `payerID = requireNonNull(transactionInfo.payerID())`
- create `QueryContext`
- validate crypto transfer via `queryChecker.validateCryptoTransfer`
- validate permissions via `queryChecker.checkPermissions`
- compute fees `handler.computeFees(context)` - There are `query.<field>OrThrow()` calls which result in NPE :disappointed: eg here in [FileGetInfoHandler](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileGetInfoHandler.java#L96).
- run query validation `handler.validate(context);`

Solution:
- We can wrap in `try/catch` or `.has<prop>()` each place in computeFees where `op.<field>OrThrow()` appears and return `Fees.FREE` to skip NPE [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/query/QueryWorkflowImpl.java#L235), then we should count on validate method to handle and throw expected PrecheckException. That approach requires refactoring in the current queries codebase. I gave examples with `FileGetInfoHandler` but the problem does exist in any query that has `op.<field>OrThrow()` in its `computeFees` method.

**Related issue(s)**:

Fixes #12672 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusGetTopicInfoHandler.java
```diff
@@ -161,7 +161,10 @@ private Optional<ConsensusTopicInfo> infoForTopic(
     public Fees computeFees(@NonNull QueryContext queryContext) {
         final var query = queryContext.query();
         final var topicStore = queryContext.createStore(ReadableTopicStore.class);
-        final var op = query.consensusGetTopicInfoOrThrow();
+        if (!query.hasConsensusGetTopicInfo()) {
+            return Fees.FREE;
+        }
```
Neeharika-Sompalli, line 166: This should fail, if we are dispatching incorrectly.

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileGetContentsHandler.java
```diff
@@ -81,8 +81,14 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
...
         final var fileStore = queryContext.createStore(ReadableFileStore.class);
-        final var op = query.fileGetContentsOrThrow();
-        final var fileId = op.fileIDOrThrow();
+        if (!query.hasFileGetContents()) {
+            return Fees.FREE;
+        }
+        final var op = query.fileGetContents();
+        if (!op.hasFileID()) {
+            return Fees.FREE;
+        }
```
Neeharika-Sompalli, line 90: We shouldn't return FREE. Instead use `op.fileIDOrElse(FileID.DEFAULT)`

Neeharika-Sompalli (commented):
I think we should fail if there is an issue with `dispatch` or log error. Also in all places we use `OrThrow` for Ids we can use `OrElse(ID.DEFAULT)` pattern

natanasow:
Resolved by https://github.com/hashgraph/hedera-services/pull/13020

PR #12919, opened 2024-04-22
fix: `ApproveAllowanceValidator` NPEs

tinker-michaelj:
**Description**:
- Cherry-pick additional fixes from #12811

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
@@ -31,6 +31,7 @@ import static com.hedera.node.app.spi.workflows.HandleException.validateTrue;
 import static java.util.Collections.emptyList;
 
 import com.hedera.hapi.node.base.AccountID;
+import com.hedera.hapi.node.base.TokenID;
 import com.hedera.hapi.node.base.TokenSupplyType;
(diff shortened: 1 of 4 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
```diff
@@ -129,7 +130,7 @@ private void validateFungibleTokenAllowances(
         for (final var allowance : tokenAllowances) {
             final var owner = allowance.owner();
             final var spender = allowance.spenderOrThrow();
-            final var token = tokenStore.get(allowance.tokenIdOrThrow());
+            final var token = tokenStore.get(allowance.tokenIdOrElse(TokenID.DEFAULT));
```
Neeharika-Sompalli, line 133: Shouldn't this fail in pureChecks ?

Neeharika-Sompalli (commented):
This change LGTM. But I think a check is missing in `pureChecks`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/ApproveAllowanceValidator.java
tinker-michaelj: Agreed that is more consistent @Neeharika-Sompalli , fixed!

Neeharika-Sompalli (approved):

PR #12949, opened 2024-04-23
chore: Update log rolling policy to size based only

### LINKED ISSUE #12921: Update log rolling policy to size based only
Update the current log4j configurations to roll files based on size, not date, and without compression. This will bring the configurations in line with what is supported by the base logger before we migrate to it.

timo0:
Updated the current log4j configurations to roll files based on size, not date, and without compression. 

This will bring the configurations in line with what is supported by the base logger before we migrate to it.

Fix #12921


hendrikebbers (approved):

rbair23:
I am thinking of two things that might be issues. First, whether there are any scripts used by node operators to gather logs, since the name and kind changes. Second, the lack of compression could be a problem in JRS tests and also on actual nodes, as I suspect these files compress well, especially if they are noisy.

povolev15 (approved):
LGTM

dalvizu (approved):
LGTM

JeffreyDallas (approved):

PR #12956, opened 2024-04-23
test: 04649 - Compare aliases from state with record stream

### LINKED ISSUE #4649: Verifying of accounts map from the final state, comparing its aliases with the record stream after fuzzing tests execution
### Problem

Implement verifying of accounts map from the final state, comparing its aliases with the record stream after fuzzing tests execution

### Solution

Implement new aliases validator and integrate it in HAPI test engine similarly to [the other existing validators](https://github.com/hashgraph/hedera-services/pull/10944).

### Alternatives

_No response_

kselveliev:
**Description**:
This PR create a new recordStreamValidator for aliases, which checks if account aliases from state
are present in the recordStream

This PR modifies:
Added `StateAccess.java` that is responsible for reading the state in HapiTestEngine.java
Added `AccountAliasValidator.java` that is responsible for verifying the account aliases are present in the recordStream
Modified `HapiTestEngine.java` to use the new validator

**Related issue(s)**:

Fixes #4649

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/HapiTestEngine.java
```diff
@@ -504,6 +507,9 @@ public HapiTestEngineExecutionContext execute(
                 }
             }
```
kselveliev, line 509: to ensure that most of the state changes will be flushed we probably could sleep here for some time

netopyr:
I am afraid this needs to be migrated to the new HAPI test framework before it can be merged.

Issue #12831, opened 2024-04-16
HIP-904: Implement the new TokenReject functionality and the integration tests for it.


MrValioBg:
### Problem

Today, if an account holds no balance (or no serial) of a token type, then the account can be dissociated from the token. However, tokens may have hefty custom fees for transferring a token. A user may not be able or willing to pay those fees, but cannot dissociate without doing so, effectively **forcing them** to pay for auto-renewal of tokens they do not want.

### Solution

This can be a significant security concern. A token with custom fees could be used by a token spammer to airdrop a token to all users with a custom fee that will charge the users some HBAR or other tokens in order to get rid of the token — essentially creating a “ransom” token. To prevent this, we introduce the new TokenRejectTransaction.

Any token (fungible or non-fungible) can be rejected, whether it was created through automatic or manual association, and no matter how long the token was held. A rejected token is transferred back to the token treasury, and all custom fees on the token are waived. The payer of the transaction will only pay the appropriate fees for the HAPI transaction type.

**_Disclaimer_**: Although this feature is introduced in the HIP-904 for frictionless airdrops, its not specifically tied only for airdrops.
Reference: [HIP-904](https://hips.hedera.com/hip/hip-904)

PR #12957, opened 2024-04-23
feat: Fee calculation for TokenReject functionality

### LINKED ISSUE #12835: HIP-904: Update feeSchedules, throttles & pricing calculation for new TokenReject transaction.

MrValioBg:
**Description**:
The values for canonical prices and feeschedules are currently copied from cryptotransfer, **those values are temporary until** there is a final decision for the exact ones.

For the `calculateFees` method implementation in TokenRejectHandler, [CryptoTransferHandler's calculateFees ](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java#L536)method was used as a reference

**Related issue(s)**: #12831 

Fixes #12835 

**Notes for reviewer**:
> [!NOTE]
> _This PR **is not aimed** to be merged into **develop**._

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -62,7 +70,51 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
+
+        return feeContext
+                .feeCalculator(getSubType(numOfNFTRejections))
+                .addBytesPerTransaction(bpt)
+                .addRamByteSeconds(rbs * USAGE_PROPERTIES.legacyReceiptStorageSecs())
+                .calculate();
+    }
+
+    private SubType getSubType(final int numOfNFTRejections) {
+        return numOfNFTRejections != 0 ? TOKEN_NON_FUNGIBLE_UNIQUE : TOKEN_FUNGIBLE_COMMON;
```
MrValioBg, line 105: In CryptoTransferHandler, we determine the SubType in a similar way. We first check if there is a NFT transfer, if not, only then we have FUNGIBLE. 
This means that if there are both Fungible and NFT tokens to be rejected, the subtype would be NON_FUNGIBLE_TOKEN.

hedera-node/hapi-fees/src/main/resources/canonical-prices.json
```diff
@@ -75,6 +75,10 @@
   "TokenUpdateNfts": {
     "DEFAULT": 0.001
   },
+  "TokenReject": {
+    "TOKEN_FUNGIBLE_COMMON": 0.001,
+    "TOKEN_NON_FUNGIBLE_UNIQUE": 0.001
+  },
```
Neeharika-Sompalli, line 81: @Nana-EC  Have we clarified with Atul the prices for these new transactions ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -62,7 +70,51 @@ public void handle(@NonNull final HandleContext context) throws HandleException
     @NonNull
     @Override
     public Fees calculateFees(final FeeContext feeContext) {
-        // Todo: Implement the logic for calculating fees
-        return Fees.FREE;
+        final var body = feeContext.body();
```
Neeharika-Sompalli, line 73: Why do we need all this logic, if we are diispatching a `CryptoTransfer` internally ? Wouldn't a `CryptoTransfer.calculateFees` be sufficient ?

MrValioBg: Hey Neeha!  
If I am getting you right about what do you refer to as "dispatching CryptoTransfer internally"
I am not sure if that is possible without breaking the flow. Isn't every handler's functionality currently confined to itself, i.e., no HAPI operation dispatches another one?

Also we have to skip some checks that are specific to the TokenReject functionality, like skipping requiredReceiverSig or transferring even if the token is paused.

Neeharika-Sompalli: Hey @MrValioBg , I think if we call `dispatchSyntheticTransaction` it computes fees also. We can extract common functionality imo.

MrValioBg: If we use  the `dispatchSyntheticTransaction` and compute fees from it, what we will do about the Fees of the TokenReject, if it fails too early before dispatching?
[Docs](https://github.com/hashgraph/hedera-services/pull/12786/files#diff-345e91521530c2849705b8d557dc66a644189322a8cc02910e1b5ea11cd73929R102) mentions that the fees should be the same as of a basic transfer without custom fees.

On another note, if we call `dispatchSyntheticTransaction`, the checks for required signatures for the receiver and the paused token, I believe, are in the handler itself, so handling it specifically for TokenReject could be challenging.

As you mentioned we could definitely extract common logic from the transfer handler logic and use it, but I am not quite sure about dispatching it, because of the cases I mentioned.

Neeharika-Sompalli: I see. We should atleast extract common logic from `CryptoTransferHandler` instead of duplicating here.

MrValioBg: Definitely.

hedera-node/configuration/testnet/upgrade/feeSchedules.json
```diff
@@ -1378,7 +1378,93 @@
                 }
             } ]
         }
-    }, {
+    },
+        {
```
Neeharika-Sompalli, line 1382: Every entry is repeated twice in this file. Need to do the same for this entry.

hedera-node/configuration/mainnet/upgrade/feeSchedules.json
```diff
@@ -1507,6 +1507,91 @@
           ]
         }
       },
+      {
+        "transactionFeeSchedule": {
```
Neeharika-Sompalli, line 1511: Every entry is repeated twice in this file. Need to do the same for this entry.

hedera-node/hedera-file-service-impl/src/main/resources/genesis/feeSchedules.json
```diff
@@ -1461,6 +1461,91 @@
           ]
         }
       },
+      {
+        "transactionFeeSchedule": {
```
Neeharika-Sompalli, line 1465: same here

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -62,7 +70,51 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
+        final var op = body.tokenRejectOrThrow();
```
Neeharika-Sompalli, line 74: We could probably have. abase class for `CryptoTransferHandler` and `TokenRejectHandler` and re-use all code that is common.

Neeharika-Sompalli (commented):
We need to re-use all the code possible. Also all the `feeSchedules.json` under `hedera-node/configuration` should be changed.

hedera-node/configuration/testnet/upgrade/feeSchedules.json
MrValioBg: Done. I've added it. Great catch.

hedera-node/configuration/mainnet/upgrade/feeSchedules.json
MrValioBg: Done. I've added it. Great catch.

hedera-node/hedera-file-service-impl/src/main/resources/genesis/feeSchedules.json
MrValioBg: Done. I've added it.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -126,7 +134,52 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
+        final int weightedFungibleTokens = tokenMultiplier * numOfFungibleTokenRejections;
+
+        final long bpt =
+                calculateBytesPerTransaction(weightedTokensInvolved, weightedFungibleTokens, numOfNFTRejections);
+        final long rbs = calculateRamByteSeconds(weightedTokensInvolved, weightedFungibleTokens, numOfNFTRejections);
+
+        return feeContext
+                .feeCalculator(getSubType(numOfNFTRejections))
+                .addBytesPerTransaction(bpt)
+                .addRamByteSeconds(rbs * USAGE_PROPERTIES.legacyReceiptStorageSecs())
```
Neeharika-Sompalli, line 164: rbs is already multiplied with `USAGE_PROPERTIES.legacyReceiptStorageSecs()` in line 181. Need not do here again.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -126,7 +134,52 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
+            } else {
+                numOfNFTRejections++;
+            }
+        }
+
+        final int weightedTokensInvolved = tokenMultiplier * totalRejections;
+        final int weightedFungibleTokens = tokenMultiplier * numOfFungibleTokenRejections;
+
+        final long bpt =
+                calculateBytesPerTransaction(weightedTokensInvolved, weightedFungibleTokens, numOfNFTRejections);
```
Neeharika-Sompalli, line 158: It is extremely important and useful to add unit tests and E2E tests in the same PR, to validate the correctness of the code.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Thats true. Great catch! I've adjusted it.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: I agree, I will close this PR and open one which contains the whole logic.

PR #12972, opened 2024-04-23
chore: delete ThreadDumpGenerator

### LINKED ISSUE #12970: Delete ThreadDumpGenerator
There is an old thread dump utility that is currently in the platform code but is not tested or suitable for production environments. I put eyes on this as I was searching for places with threads outside the wiring framework. Let's delete this for now, and we can circle back and reconsider the appropriate strategy if we ever have the time to do it properly.

cody-littley:
closes #12970


alittley (dismissed):

lpetrovic05:
I don't see why would should delete this now if we are not sure whether we will need it or not. Is it getting in the way?

cody-littley:
@lpetrovic05 

> I don't see why would should delete this now if we are not sure whether we will need it or not. Is it getting in the way?

I'm mostly just attempting to clean up unused code that uses threads. IMO, better to go back in git history to get this if we need it than to have to maintain it for many years and never end up using it.

kfa-aguda (approved):
love to see deleted code!

alittley (approved):

PR #12967, opened 2024-04-23
fix: Enable aliases in TokenWipe as per mono service

Neeharika-Sompalli:
Fixes #12953 

- In mono-service we have aliases enabled for only `TokenWipe` transaction apart from `CryptoTransfer` transaction.
- Enable the same in mod service

PR #12974, opened 2024-04-23
fix: Enable aliases in TokenWipe as per mono service (#12967)

Neeharika-Sompalli:
Cherry-pick https://github.com/hashgraph/hedera-services/pull/12967


kimbor (approved):

Neeharika-Sompalli:
The failing tests in Misc are known , there is a PR to fix this in release/0.49 and will be merged to develop

PR #12977, opened 2024-04-24
fix: use `targeted_address` for action sidecar call to missing address

### LINKED ISSUE #12641: (mod) Investigate certain sidecar contract ids
### Description

A previewnet transaction [here](https://hashscan.io/previewnet/transaction/1712267533.938791678) generated a `CONTRACT_ACTION` sidecar whose recipient was a `recipient_account` with an out-of-range `accountNum`.

### Steps to reproduce

Unclear.

### Additional context

_No response_

### Hedera network

previewnet

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #12641 
 - Only replace the `targeted_address` in an action sidecar with a `receipient_account` **IF** the targeted address exists after processing the `CALL`.
     * Otherwise leave the `targeted_address` in the final sidecar.
 - In `getHederaContractId()`, return the id of a pending creation if its address matches _before_ falling back to behavior controlled by `contractMustBePresent`.
     * This fixes `TraceabilitySuite#assertSidecars()`.
- Allow gas costs in hard-coded sidecar expectations to vary by up to 64 `gas` instead of 32.
- Remove some dead code and confusing comments.


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/ActionStack.java
```diff
@@ -192,12 +191,13 @@ private ContractAction finalFormOf(@NonNull final ContractAction action, @NonNul
...
-                        try {
-                            builder.recipientAccount(accountIdWith(hederaIdNumberIn(frame, lazyCreatedAddress)));
-                        } catch (final IllegalArgumentException | NullPointerException | ArithmeticException e) {
-                            // handle non-existing to/receiver address
-                            builder.recipientAccount((AccountID) null);
+                        final var maybeCreatedAddress = pbjToBesuAddress(action.targetedAddressOrThrow());
+                        final var maybeCreatedAccount = proxyUpdaterFor(frame).getHederaAccount(maybeCreatedAddress);
+                        // Fill in the account of id of a successful lazy creation; but just leave
+                        // the targeted address in case of a failed lazy-creation or a call to a
+                        // non-existent address
```
david-bakin-sl, line 198: How does this fix
> A previewnet transaction [here](https://hashscan.io/previewnet/transaction/1712267533.938791678) generated a CONTRACT_ACTION sidecar whose recipient was a recipient_account with an out-of-range accountNum.

Or is the out-of-range accountNum referred to the `((AccountID)null)` that was in the previous code?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
```diff
@@ -68,6 +69,10 @@ private void validateActionIds(@NonNull final TransactionSidecarRecord sidecar)
                 assertValid(action.getRecipientContract(), "action#recipientContract", sidecar, this::isValidId);
             }
 
+            final var isMissingRecipient =
```
david-bakin-sl, line 72: (It's just me, I'm sure, but I would have use De Morgan's law to make this be `hasARecipient` and done everything in the _positive_ sense.  Because I'm a positive kind of guy.   Next line too.  But don't change anything here.  Just a thought on people's differing world-views and how that's reflected in code.  Or, possibly, that's just baloney, huh?)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
```diff
@@ -117,7 +122,7 @@ private void assertValid(
     }
 
     private boolean isValidId(long shard, long realm, long num) {
-        return shard == 0L && realm == 0L && num >= 1;
+        return shard == 0L && realm == 0L && num >= 1 && num < Integer.MAX_VALUE;
```
david-bakin-sl, line 125: And at L129 below?

david-bakin-sl (approved):
LGTM (with some comments)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/ActionStack.java
```diff
@@ -192,12 +191,13 @@ private ContractAction finalFormOf(@NonNull final ContractAction action, @NonNul
                 } else {
                     builder.output(tuweniToPbjBytes(frame.getOutputData()));
                     if (action.targetedAddress() != null) {
-                        final var lazyCreatedAddress = pbjToBesuAddress(action.targetedAddressOrThrow());
-                        try {
-                            builder.recipientAccount(accountIdWith(hederaIdNumberIn(frame, lazyCreatedAddress)));
-                        } catch (final IllegalArgumentException | NullPointerException | ArithmeticException e) {
-                            // handle non-existing to/receiver address
-                            builder.recipientAccount((AccountID) null);
+                        final var maybeCreatedAddress = pbjToBesuAddress(action.targetedAddressOrThrow());
```
lukelee-sl, line 194: Makes much more sense than what tried to do.

lukelee-sl (dismissed):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
```diff
@@ -55,6 +55,7 @@ private void validateStateChangeIds(@NonNull final TransactionSidecarRecord side
 
     private void validateActionIds(@NonNull final TransactionSidecarRecord sidecar) {
         final var actions = sidecar.getActions().getContractActionsList();
+        System.out.println(actions);
```
Neeharika-Sompalli, line 58: Just debug code ?

Neeharika-Sompalli (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/ActionStack.java
tinker-michaelj: The previous code would always set a long-zero address as the `recipient_account`, **even if** that long-zero address didn't exist.

So e.g. `0xE8D4A50FFF` would become `receipient_account { account_num: 999999999999 }`.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
tinker-michaelj: Thanks Neeha, removed!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
tinker-michaelj: Switched to the positive spin. 😁

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/ValidContractIdsAssertion.java
tinker-michaelj: Good idea, done.

lukelee-sl (approved):
LGTM

Neeharika-Sompalli (approved):

PR #12978, opened 2024-04-24
feat: Create ConnectionManagerFactory

### LINKED ISSUE #11820: Create ConnectivityManager
depends on https://github.com/hashgraph/hedera-services/issues/11762

- the class `StaticConnectionManagers` should become `ConnectivityManager`
- it should be thread-safe, since it will be called by multiple threads
- it should have a method `List<ConnectionManager> updatePeers(List<PeerInfo>)` which should return a ConnectionManager for each peer
- it also needs to hold a map of all ConnectionManagers internally
- when a new incoming connection is received, the method `newConnection(SocketConnection)` should be called. this should validate the incoming connection and pass it to the appropriate connection manager

kfa-aguda:
closes #11820


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/ConnectivityManager.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Set;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Pre-builds connection managers for the supplied topology, does not allow changes at runtime
+ */
+public class ConnectivityManager {
+    private static final Logger logger = LogManager.getLogger(ConnectivityManager.class);
+    private final NetworkTopology topology;
```
lpetrovic05, line 44: topology seems to be unchanged even when we update peers, which seems to indicate that updatePeers will not work. if unit tests did not catch this, then they need to be expanded.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/ConnectivityManager.java
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * @param topology          the network topology
+     * @param connectionCreator the connection creator
+     */
+    public ConnectivityManager(
+            @NonNull final NetworkTopology topology, @NonNull final OutboundConnectionCreator connectionCreator) {
+        this.topology = Objects.requireNonNull(topology);
+        this.connectionCreator = Objects.requireNonNull(connectionCreator);
+
+        this.connectionManagers = createConnectionManagers();
```
lpetrovic05, line 62: there are 2 pathways for creating connection managers, one based on topology and the other based on the peer list. i think it would be a lot simpler if there was only 1

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/ConnectivityManager.java
kfa-aguda: Hmm this isn't clear to me yet. Is there a requirement for the topology to change in this class? Always understood the connectivity manager's responsibility begins *after* the network topology has already been decided, with that topology passed down here. Are you thinking for DAB, we have to change the topology here? If yes, It's not something we've discussed but happy to hash it out and document.

lpetrovic05: updating peers is a change in topology, it cannot remain the same

kfa-aguda: Got it. Can you expand further on what should be changed in the topology and why? Happy to hop a quick call if that works better.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/ConnectionManager.java
```diff
@@ -39,6 +41,13 @@ public interface ConnectionManager {
      */
     Connection getConnection();
 
+    /**
+     * Get the node id that this connection manager is responsible for
+     *
+     * @return the id of the node that this connection manager is responsible for
+     */
+    NodeId getNodeId();
```
hendrikebbers, line 49: Add NonNull annotation

hendrikebbers (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/ConnectivityManager.java
lpetrovic05: sure, ping me on slack and we can arrange a call

kfa-aguda:
Closing as this requires additional work to implement topology changes and DAB has been deprioritized.

PR #12993, opened 2024-04-24
perf: warm up more entities

### LINKED ISSUE #12976: Warm up more entities
### Problem

#12531 touches more entities (tokens, NFTs), which has caused performance regression in the NFT transfer benchmark.
Those entities need to be added to `CryptoTransferHandler.warm()`.

### Solution

Update `CryptoTransferHandler.warm()`.

### Alternatives

_No response_

OlegMazurov:
**Description**:
* for the NFT receiver account, warm up its `headToken`
* `nftChangesFrom()`: don't read the token unless it's needed

**Related issue(s)**:

Fixes #12976

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
@@ -171,13 +171,14 @@ public class CryptoTransferHandler implements TransactionHandler {
             }
             final List<NftTransfer> nftTransfers = tokenTransferList.nftTransfers();
             for (final NftTransfer nftTransfer : nftTransfers) {
-                warmNftTransfer(accountStore, nftStore, tokenRelationStore, tokenID, nftTransfer);
+                warmNftTransfer(accountStore, tokenStore, nftStore, tokenRelationStore, tokenID, nftTransfer);
             }
         });
     }
 
     private void warmNftTransfer(
             @NonNull final ReadableAccountStore accountStore,
+            @NonNull final ReadableTokenStore tokenStore,
             @NonNull final ReadableNftStore nftStore,
             @NonNull final ReadableTokenRelationStore tokenRelationStore,
             @NonNull final TokenID tokenID,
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -195,7 +196,10 @@ private void warmNftTransfer(
         nftTransfer.ifReceiverAccountID(receiverAccountID -> {
             final Account receiver = accountStore.getAliasedAccountById(receiverAccountID);
             if (receiver != null) {
-                receiver.ifHeadTokenId(headTokenID -> tokenRelationStore.warm(receiverAccountID, headTokenID));
+                receiver.ifHeadTokenId(headTokenID -> {
```
tinker-michaelj, line 199: The `receiver`'s head `tokenRelation` will only be updated it is being auto-associated to the token type of the NFT. 

And the token type pointed to by `headTokenID` should never be used during a `CryptoTransfer`, unless by coincidence it appears in another `tokenTransferList` in the same transaction.

If these warmings are empirically helpful, LGTM...just curious why they would be. 🤔

tinker-michaelj (approved):
Had one question about the added warmings, but LGTM!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
Neeharika-Sompalli: Seems like the test was doing auto-associations and without warming `headTokenID` it was causing performance issue.

Neeharika-Sompalli (approved):
LGTM ! Thanks @OlegMazurov

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
OlegMazurov: Well, that's exactly the case with the benchmark: it relies on auto-association as otherwise there would need to be billions of API calls to associate all tokens to their receiving accounts, which is impractical (would take many days at 10K TPS). I realize it's not reflective of real-life workloads but it may be considered the worst case scenario and thus methodologically acceptable (we know we'll get better TPS with pre-associated tokens).

iwsimon (approved):
LGTM. Thanks @OlegMazurov !

OlegMazurov:
I have to close this PR due to a broken commit signature. Please re-review #13005

Issue #11568, opened 2024-02-15
Document the Gradle build configuration


nathanklick:
## Description

The Gradle build logic needs to be documented including:

- The `build-logic` folder
- Project/module naming standards
- Use of the GradleX tooling
- Available custom tasks

jjohannes:
As part of this (or split-out into separate issue): Use unified naming and structure in **all** repositories 
- https://github.com/hashgraph/hedera-services
- https://github.com/hashgraph/pbj
- https://github.com/hashgraph/hedera-sdk-java
- ...

For example:
- Use `gradle/plugins` as "build logic" folder everywhere
- Use `com.hedera.gradlebuild...` as namespace for all convention plugin
- Use similar naming for plugins that do similar things in the different repos
- Use the **exact** same names for convention plugins that do **exactly** the same in different repos (later we could think about sharing the logic between repos)

jjohannes:
**Start with the _user_ documentation**

As part of this, move the following to `docs` folder in root, update where appropriate. Make sure we have a full "how and what to use Gradle for in daily development" guide.

- branch-naming-conventions
- docker-quickstart
- gradle-quickstart
- hedera-java-style-guide
- intellij-quickstart
- maintainers-guide

jjohannes:
I plan to document the Gradle setup itself in the README of the new repository we'll create for the Gradle setup as part of #14255

jjohannes:
Follow up work: https://github.com/hiero-ledger/hiero-gradle-conventions/issues/6

PR #12997, opened 2024-04-25
build: consistent naming of Gradle configuration files / Gradle updates

jjohannes:
**Description**:

Consistent names and locations for shared Gradle configuration (aka convention plugins). Changes:

- moved `build-source` to `graldle/plugins` and merged to just one "plugins" project. This is what most projects do now. And I like that it keeps things you do not need to touch regularly hidden in the "gradle" folder (that we already have for the wrapper).
- Changes to convention plugin IDs:
   - All now have the `com.hedera.gradle` namespace to indicate that these are "packages" for the "gradle build configurations" (and not production or test code). The idea is that the configuration (with the same naming) can be reused in other repos like pbj or hedera-sdk-java.
   - The "main" convention plugins that define to which group a project belonga are now consistently named. We have:
      - `com.hedera.gradle.platform`
      - `com.hedera.gradle.services`
      - `com.hedera.gradle.blocknode`
- The `-convention` suffix has been removed from all places where it was used. It was inconsistently used and does not add any value imo.
- Gradle and all plugins are now updated to the latest version
- The patching in `com.hedera.gradle.jpms-modules` is now expressed in a more compact way using the new features for this of the `org.gradlex.jvm-dependency-conflict-resolution` plugin.

**Related issue(s)**:

#11568 

**Notes for reviewer**:

**Checklist**


gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.jpms-modules.gradle.kts
```diff
@@ -50,86 +42,77 @@ dependencies.components {
...
-        params(grpcComponents + annotationLibraries)
-    }
-    withModule<AddDependenciesMetadataRule>("io.grpc:grpc-protobuf") { params(grpcModule) }
-    withModule<RemoveDependenciesMetadataRule>("io.grpc:grpc-protobuf-lite") {
-        params(
-            grpcComponents + annotationLibraries + listOf("com.google.protobuf:protobuf-javalite")
-        )
+    module("io.netty:netty-transport-native-epoll") {
+        addFeature("linux-x86_64")
+        addFeature("linux-aarch_64")
```
hendrikebbers, line 47: What is `addFeature` doing?

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.root.gradle.kts
```diff
@@ -14,19 +14,20 @@
...
-    id("com.hedera.hashgraph.repositories")
-    id("com.hedera.hashgraph.aggregate-reports")
-    id("com.hedera.hashgraph.spotless-conventions")
-    id("com.hedera.hashgraph.spotless-kotlin-conventions")
-    id("com.hedera.hashgraph.dependency-analysis")
+    id("com.hedera.gradlebuild.lifecycle")
+    id("com.hedera.gradlebuild.repositories")
+    id("com.hedera.gradlebuild.aggregate-reports")
+    id("com.hedera.gradlebuild.spotless")
+    id("com.hedera.gradlebuild.spotless-kotlin")
```
hendrikebbers, line 26: Do I need to add `com.hedera.gradlebuild.spotless` and `com.hedera.gradlebuild.spotless-kotlin`? I would assume  that `spotless-kotlin` already integrate the other one.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.jpms-modules.gradle.kts
```diff
@@ -50,86 +42,77 @@ dependencies.components {
...
     }
-    withModule<AddDependenciesMetadataRule>("io.grpc:grpc-protobuf-lite") {
-        params(grpcModule + listOf("com.google.protobuf:protobuf-java"))
+    module("io.grpc:grpc-netty") {
+        annotationLibraries.forEach { removeDependency(it) }
```
hendrikebbers, line 50: This file looks way more complex than the last time I looked at it. We need to create shared knowledge on the patterns and apis.

hendrikebbers (dismissed):

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.root.gradle.kts
jjohannes: :+1: removed.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.jpms-modules.gradle.kts
jjohannes: When was that last time? From my perception it became more and more readable with each iteration. 😄 But there was also a time where it was shorter before **everything** was patched to be a module.

I agree on the knowledge sharing though. In the IDE you can now click on the "methods" used here and get some documentation on what they are doing. And there are some comments in the file. But we can add more. Maybe this is something we should have a chat about to figure out what's the best way is to improve this. (in a follow up to this PR)

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.jpms-modules.gradle.kts
jjohannes: There is Javadoc on the `addFeature` that says:

> Make the Jar with the give 'classifier' known as _Feature Variant_ so that it can be selected via capability
in a dependency declaration.
See: https://docs.gradle.org/current/userguide/component_metadata_rules.html#making_different_flavors_of_a_library_available_through_capabilities
See: https://blog.gradle.org/optional-dependencies

In short: the information that the classified Jars exist is added to the metadata which then allows the _java-module-dependencies_ plugin to address the Jars as individual modules.

JeffreyDallas (approved):

JeffreyDallas (dismissed):

jjohannes:
@JeffreyDallas I cleaned up the PR and move some changes in documentation out of this one. Now there are no other approvals required. Could you re-approve?

jeromy-cannon (approved):

PR #13010, opened 2024-04-25
feat: Move to fully connected network

### LINKED ISSUE #12984: Eliminate addressBook from Network component

kfa-aguda:
closes #12984

This PR was formerly titled "Eliminate addressBook from Network component" with the goal of eliminating the dependency of network on addressBook, making peer lookups O(1); an important prep for Dynamic Address Book which will be lookup heavy.

Upon initial implementation and review, updating the PR to capture the eventual effect - make the network fully connected.


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -32,35 +31,33 @@
...
             final int numberOfNeighbors) {
-        this.addressBook = Objects.requireNonNull(addressBook, "addressBook must not be null");
-        this.selfId = Objects.requireNonNull(selfId, "selfId must not be null");
-        this.connectionGraph = new RandomGraph(random, addressBook.getSize(), numberOfNeighbors, SEED);
-
-        if (!addressBook.contains(selfId)) {
-            throw new IllegalArgumentException("Address book does not contain selfId");
-        }
+        this.peerNodes =
+                Objects.requireNonNull(peers.stream().map(PeerInfo::nodeId).toList(), "peers must not be null");
```
cody-littley, line 58: The `Objects.requireNonNull()` is not necessary. If `peers` is null then `peers.stream()` will throw an exception first.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -121,4 +115,15 @@ public boolean shouldConnectTo(final NodeId nodeId) {
...
     }
+
+    /**
+     * Returns the index of the given node in the peer list, which must not be the self index
+     *
+     * @param nodeId the node ID
+     * @return the index of the node in the peer list
+     */
+    private int getIndexOfNodeId(@NonNull final NodeId nodeId) {
+        final int index = peerNodes.indexOf(nodeId);
```
cody-littley, line 126: Instead of iterating the list to find the index, we should instead create a `Map<NodeId, Integer>` for `O(1)` lookup time.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -76,19 +73,16 @@ public List<NodeId> getNeighbors() {
      */
     @Override
     public List<NodeId> getNeighbors(final Predicate<NodeId> filter) {
```
alittley, line 75: This method used to return *neighbors*, but now it's returning all peers.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -121,4 +115,15 @@ public boolean shouldConnectTo(final NodeId nodeId) {
...
         return connectionGraph;
     }
+
+    /**
+     * Returns the index of the given node in the peer list, which must not be the self index
+     *
+     * @param nodeId the node ID
+     * @return the index of the node in the peer list
+     */
+    private int getIndexOfNodeId(@NonNull final NodeId nodeId) {
```
alittley, line 125: Won't this method have an off by 1 error for every node with index > self index? You're using position in `peerNodes`, but self is missing.

alittley (commented):
Why have the files `sigCerts.data` and `agrCerts.data` been added in this PR?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley, line 1: Why are we modifying this class right now? We discussed as a group a few weeks back, and as I recall the decision was to make the network fully connected and not to try and modify the partial connectivity code until we had the bandwidth to test it properly.

The danger of modifying this code now is that we don't actually have any integration tests where the network is not fully cunnected, meaning we may be introducing bugs we don't catch for a long time.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
```diff
@@ -188,9 +188,10 @@ protected SyncGossip(
         final ThreadConfig threadConfig = platformContext.getConfiguration().getConfigData(ThreadConfig.class);
 
         final BasicConfig basicConfig = platformContext.getConfiguration().getConfigData(BasicConfig.class);
-
-        topology = new StaticTopology(random, addressBook, selfId, basicConfig.numConnections());
         final List<PeerInfo> peers = Utilities.createPeerInfoList(addressBook, selfId);
+
+        topology =
+                new StaticTopology(random, peers, addressBook.getIndexOfNodeId(selfId), basicConfig.numConnections());
```
lpetrovic05, line 194: this index is an AB concept, so we should get rid of it

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: good point.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: the peers are all neighbors, but not necessarily adjacent.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: for every node with index > self index, we need the index returned by this method to be off by 1 (to match their position similar to the original addressBook). 
There's an edge case for when being off by 1 can lead to this method returning an index number > vertex upper bound, but we deal with that correctly in the isAdjacent check so it's ok.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: All that this change does is to get rid of the addressBook within the network code so we're operating on peers, making tests so much easier. AB is limiting to the effectiveness of writing tests, and also just a completely orthogonal concept to the network code. It's why I have made some choices to closely mimic the old functionality, like doing an index lookup via iteration vs map (for ex).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
alittley: I don't understand this. If all peers are neighbors, why do we have 2 terms here? To me, `neighbor` implies `adjacent`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
alittley: I don't understand this either. The previous code in this class did the following, to get the index of the other node in the address book. This computation *includes* self id in the indices

```
addressBook.getIndexOfNodeId(nodeId);
```

But the new code instead defines the index of the other node via its position in the peer list, which *doesn't* contain self id. So for every node with index greater than this self id, the returned node index will be 1 less.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
alittley: > like doing an index lookup via iteration vs map

The existing code uses `AddressBook::getIndexOfNodeId` to determine node index, which has a map lookup under the hood. Where are you seeing index lookup via iteration?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: O(1) here would be nice but i still have to keep a list of peers for the neighbors. To use a map, the values of the map would be a set of peers which must still be converted to a list in O(n). No good trade-off here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: `StaticTopology.getNeighbors` is one.

Anyway, doing away with node indices altogether has made this a lot easier as we no longer have to match the old behavior.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -121,4 +102,19 @@ public boolean shouldConnectTo(final NodeId nodeId) {
...
+     * Maps the list of peers to a map of node IDs to their index in the peer list
+     * and populates the peerNodesList with the node IDs
+     *
+     * @param peers the list of peers
+     * @return the map of node IDs to their peer Id
+     */
+    @NonNull
+    private Map<NodeId, Long> map(@NonNull final List<PeerInfo> peers) {
+        for (final PeerInfo peer : peers) {
+            peerNodeToIdMap.put(peer.nodeId(), peer.nodeId().id());
```
edward-swirldslabs, line 116: This definition of a map is not meaningful.  If you have the key, then you can call `.id()` and get the value without using the map.  

I think your intent may be to create a new index ordering based on PeerInfo and map to indexes for use in compatibility with RandomGraph?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -98,20 +84,15 @@ public boolean shouldConnectToMe(final NodeId nodeId) {
      * @return true if this node is my neighbor, false if not
      */
     private boolean isNeighbor(final NodeId nodeId) {
-        if (!addressBook.contains(nodeId)) {
-            return false;
-        }
-        final int selfIndex = addressBook.getIndexOfNodeId(selfId);
-        final int nodeIndex = addressBook.getIndexOfNodeId(nodeId);
-        return connectionGraph.isAdjacent(selfIndex, nodeIndex);
+        return peerNodeToIdMap.containsKey(nodeId);
```
edward-swirldslabs, line 87: You've dropped the `isAdjacent` check with connectionGraph.   If the PR is just to remove the address book dependency then you still need to query the index from the map you construct once it has the right value in the key-value pair.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -17,78 +17,64 @@
...
 
 /**
  * A bidirectional topology that never changes.
  */
 public class StaticTopology implements NetworkTopology {
     private static final long SEED = 0;
 
-    private final NodeId selfId;
+    /** nodes are mapped so lookups are efficient. **/
+    private Map<NodeId, Long> peerNodeToIdMap = new HashMap<>();
```
edward-swirldslabs, line 36: The name of the variable is more appropriately named as `nodeIdToIndexMap` assuming the intent is to replace the address book lookup for index.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -17,78 +17,64 @@
...
      *                          cryptographically secure
-     * @param addressBook       the current address book
+     * @param peers             the set of peers in the network
      * @param selfId            the ID of this node
      * @param numberOfNeighbors the number of neighbors each node should have
      */
     public StaticTopology(
             @NonNull final Random random,
-            @NonNull final AddressBook addressBook,
+            @NonNull final List<PeerInfo> peers,
```
edward-swirldslabs, line 56: Is there an invariant on this list that the PeerInfo is in monotonically ascending order with respect to NodeId?

edward-swirldslabs (changes requested):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -17,78 +17,64 @@
...
      */
     @Override
-    public List<NodeId> getNeighbors(final Predicate<NodeId> filter) {
-        final int selfIndex = addressBook.getIndexOfNodeId(selfId);
-        return Arrays.stream(connectionGraph.getNeighbors(selfIndex))
-                .mapToObj(addressBook::getNodeId)
-                .filter(filter)
-                .toList();
+    public Set<NodeId> getNeighbors() {
+        return peerNodeToIdMap.keySet();
```
edward-swirldslabs, line 69: In the previous implementation this method called the predicate argument method.  That method has been deleted.  In the implementation of that deleted method, the definition of Neighbor was scoped to the `connectIonGraph` which is a `RandomGraph` that also models the number of allowed degrees or edges between nodes.  If there are 100 nodes and each node has only 10 degrees in the graph then each node has 10 neighbors.   The old implementation would provide the 10 neighbors a node is allowed to talk to.  This implementation returns all neighbors as being allowed to talk to.

edward-swirldslabs (changes requested):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley: change not made in latest commit

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley: `O(1)` is non-optional for stuff like this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley: Ok, I think I understand why this works.

If the node ID of the neighbor is greater than ours, then this method returns a value that is one less than its actual index. If the node ID of the neighbor is equal to our node ID plus one, `getIndexOfNodeId()` increments the number so that it doesn't actually equal our index.

Although this technically yields the correct result, I feel like this is one of those times where the code is "accidentally returning the right results". At the very least, the code here does not return values consistent with the javadocs. Better to refactor this code, I can imagine that the reason why/how this works may not be apparent to a developer who looks at this in the future.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley: Ok, I think I understand why this works.

If the node ID of the neighbor is greater than ours, then this method returns a value that is one less than its actual index. If the node ID of the neighbor is equal to our node ID plus one, getIndexOfNodeId() increments the number so that it doesn't actually equal our index.

This class still returns the correct results because we only care if the other node index is greater or less than our own index, so it doesn't matter that it's off by 1 as long as it is greater than our index.

Although this technically yields the correct result, I feel like this is one of those times where the code is "accidentally returning the right results". At the very least, the code here does not return values consistent with the javadocs. Better to refactor this code I think. I can imagine that the reason why/how this works may not be apparent to a developer who looks at this in the future.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: yes there is an invariant, but it's that PeerInfo are ordered as they appear in the original addressBook. I understand they should be monotonically ascending as defined in the AB, but I don't know if that's mandated anywhere.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
cody-littley: Perhaps we can get `O(1)` with a much similar approach. Is there any reason we can't just compare node IDs directly? The index is really just a proxy for node IDs. All we really want to do is figure out if the other node comes first in the address book, and if you have a lower node ID you will always come first in the address book.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: Discussed with @edward-swirldslabs. The `StaticTopology` implementation, given nodes ranging from 1-100 returned what's equivalent to the 'peersList` (i.e. the list of peers), all the time. There's a specific test case for this. Agreed it did that based on what's returned by `connectionGraph.getNeighbors()`. I understand there's been some historic discussions around what the future of this class should look like, but the goal of this PR is to introduce an equivalent logic that does not use the AB. 

Happy to discuss alternatives/limitations.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
edward-swirldslabs: @lpetrovic05 Is the scope of this PR to simply remove the address book references or also drop our support for non-fully connected network topologies?   The `RandomGraph` class uses hard-coded arrays of size `addressbook.getSize()`.  The translation from `id` to `index` is to support the use of `RandomGraph` to decide who is a neighbor when there are more nodes in the network than the degree of edges allowed per node.  

If we drop the use of `RandomGraph`, then we can get rid of the translation from `id` to `index`.  But if we keep `RandomGraph`, then we still need to translate from `id` to `index`.  We can do it without the address book, using the `List<PeerInfo>`, but we'll have to reconstruct the `Map<id, index>`, and be sensitive to where the `selfId` falls into the sequence.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
edward-swirldslabs: To fully do away with node indeces, we have to drop use of `RandomGraph` and officially only support fully connected networks until we introduce a new implementation of connecting to less peers than the number of nodes in the address book.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
```diff
@@ -193,9 +193,9 @@ protected SyncGossip(
         final ThreadConfig threadConfig = platformContext.getConfiguration().getConfigData(ThreadConfig.class);
 
         final BasicConfig basicConfig = platformContext.getConfiguration().getConfigData(BasicConfig.class);
-
-        topology = new StaticTopology(random, addressBook, selfId, basicConfig.numConnections());
         final List<PeerInfo> peers = Utilities.createPeerInfoList(addressBook, selfId);
+
+        topology = new StaticTopology(random, peers, selfId, basicConfig.numConnections());
```
lpetrovic05, line 198: I think probably the best thing to do is to replace the topology implementation with a simpler one that has no random graph. and we keep the current one for the future when we decide we want to support it

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: resolving, as this is now outdated.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: resolving, as this is now outdated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: resolving, as this is now outdated as per subsequent discussions.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: As discussed, `RandomGraph` has now been dropped completely and the network made fully connected. Resolving this thread.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
kfa-aguda: `RandomGraph` has now been dropped with the network becoming fully connected. Resolving.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: outdated. Resolving

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: outdated. Resolving.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: updated.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: outdated, resolving.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
kfa-aguda: done. Resolving.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
kfa-aguda: This how now been implemented as a fully connected network, eliminating the use of RandomGraph completely. Resolving.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -98,27 +69,14 @@ public boolean shouldConnectToMe(final NodeId nodeId) {
      * @return true if this node is my neighbor, false if not
      */
     private boolean isNeighbor(final NodeId nodeId) {
```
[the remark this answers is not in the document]
kfa-aguda: It's still a network topology though, even if now fully connected. It's useful to have the concept of who's a neighbor vs who isn't, even if now straightforward. We still need to always check the peer is in the list of neighbors.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
```diff
@@ -17,78 +17,49 @@
...
 public class StaticTopology implements NetworkTopology {
-    private static final long SEED = 0;
 
-    private final NodeId selfId;
-    /**
-     * Two nodes are neighbors if their indexes in the address book are neighbors in the connection graph.
-     */
-    private final AddressBook addressBook;
+    /** nodes are mapped so lookups are efficient. **/
+    private final Set<NodeId> nodeIdToIndexSet = new HashSet<>();
```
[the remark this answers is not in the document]
kfa-aguda: Agreed. Modified.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/OutboundConnectionManager.java
```diff
@@ -46,7 +47,7 @@ public OutboundConnectionManager(final NodeId peerId, final OutboundConnectionCr
     @Override
     public Connection waitForConnection() {
         try (final LockedResource<Connection> resource = lock.lock()) {
-            while (!resource.getResource().connected()) {
+            while (!Objects.requireNonNull(resource.getResource()).connected()) {
```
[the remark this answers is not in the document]
kfa-aguda: thanks. fixed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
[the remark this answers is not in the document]
alittley: I'd agree if this was part of an interface or something, but this is a private implementation detail of a class described by the class doc as "A fully connected topology that never changes." IMO it doesn't make sense to leave around extra internal complexity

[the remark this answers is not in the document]
kfa-aguda: The interface this class implements has a 'getNeighbors' API, which is implemented by this class. So reasoning this using the 'neighbor' concept is necessary. This private method makes it unambiguous who is a neighbor vs who isn't. That makes all the sense to me. 

Are you requesting a change or is this a nice to have?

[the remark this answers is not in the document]
alittley: The implementation of the `getNeighbors` *itself* is what makes who is a neighbor unambiguous. `getNeighbors` simply returns all node ids, so anyone should instantly understand what that means. `getNeighbors` doesn't even call `isNeighbor`, so I don't see it being relevant.

I'm not blocking this PR on this point, but I'd like to leave this comment thread open for others to weigh in during their reviews. To me, this is a clear case of unnecessary internal complexity.

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/topology/StaticTopology.java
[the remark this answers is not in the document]
kfa-aguda: done. thanks both.

edward-swirldslabs (approved):

PR #13012, opened 2024-04-25
fix: refactor and simplify HIP-540 implementation and tests

tinker-michaelj:
**Description**:
 - Move state-independent `TOKEN_UPDATE` checks [into `pureChecks()`](https://github.com/hashgraph/hedera-services/blob/refactor-hip540/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java#L92).
 - Make signing requirements [much more explicit](https://github.com/hashgraph/hedera-services/blob/refactor-hip540/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java#L381).
 - Dramatically simplify test coverage by characterizing each scenario in a [`Hip540TestScenario`](https://github.com/hashgraph/hedera-services/blob/refactor-hip540/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/hip540/Hip540TestScenario.java#L59) value that is capable of translating itself into a sequence of `HapiSpecOperation`s; and repeat all these scenarios for each non-admin key type [here](https://github.com/hashgraph/hedera-services/blob/refactor-hip540/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/token/hip540/TestScenarios.java#L46).
 - Remove some unused methods.
 - Add some `requireNonNull()` and `OrThrow()` assertions to fail faster in some cases.


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -87,27 +89,27 @@ public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckExcept
...
-            validateTruePreCheck(isValid(op.feeScheduleKey()), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
+        // IMPORTANT: No matter the TokenKeyValidation mode, we always require keys to
+        // be structurally valid. Putting structurally invalid keys into ledger state
+        // makes no sense, and could create problems for mirror nodes and block explorers.
+        // That is, using NO_VALIDATION only lets a user set a new key without its signature;
+        // and thus set a low-entropy key that proves a role function has been "disabled"
+        for (final var tokenKey : TOKEN_KEYS) {
+            if (tokenKey.isPresentInUpdate(op)) {
+                final var key = tokenKey.getFromUpdate(op);
+                if (!isKeyRemoval(key)) {
```
petreze, line 100:
```suggestion
                if (!isKeyRemoval(key) && op.keyVerificationMode() != TokenKeyValidation.NO_VALIDATION) {
```
if we do not check for the `TokenKeyValidation` enum, a user will always end up with `INVALID_<SOME>_KEY` in prechecks when he tries to update to an invalid key. IMO, we should check if the `TokenKeyValidation` is `NO_VALIDATION` and in this case to not throw anything and allow the user to update to an invalid key

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -374,93 +376,87 @@ private void updateKeys(
...
         }
-        return lowPriorityKeys;
     }
 
-    /**
-     * Check if TokenUpdateOp tries to remove some of the token keys
-     */
+    private void requireAdmin(@NonNull final PreHandleContext context, @NonNull final Token originalToken)
+            throws PreCheckException {
+        validateTruePreCheck(originalToken.hasAdminKey(), TOKEN_IS_IMMUTABLE);
```
petreze, line 456: Previously, we were throwing `TOKEN_IS_IMMUTABLE` when the token cannot be updated under any circumstances. With this logic, we will throw it if the admin key is required for the success of the update transaction and it is not present. But this does not mean that the token is immutable. If the token has any low priority key, it can still be updated if for example the freeze key want to update itself

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -58,8 +58,8 @@
 
 public class BaseTokenHandler {
     private static final Logger log = LogManager.getLogger(BaseTokenHandler.class);
-    protected static final Set<TokenKeys> TOKEN_KEYS = EnumSet.allOf(TokenKeys.class);
-    protected static final Set<TokenKeys> NON_ADMIN_TOKEN_KEYS = EnumSet.complementOf(EnumSet.of(TokenKeys.ADMIN_KEY));
+    public static final Set<TokenKey> NON_ADMIN_TOKEN_KEYS = EnumSet.complementOf(EnumSet.of(TokenKey.ADMIN_KEY));
+    static final Set<TokenKey> TOKEN_KEYS = EnumSet.allOf(TokenKey.class);
```
petreze, line 62: I think `TOKEN_KEYS` and `NON_ADMIN_TOKEN_KEYS` can stay protected

petreze (commented):
Why we remove `Hip540ChangeOrRemoveKeysSuite`. The file contains lots of corner cases to test every combination of `TokenUpdateTx` to examine every possible behaviour for this HIP.

With the provided changes in this PR, 17 from 29 tests in `Hip540ChangeOrRemoveKeysSuite` are failing

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
```diff
@@ -374,93 +376,87 @@ private void updateKeys(
...
+        final var maybeRoleKey = roleKey.getFromToken(token);
+        mustExist(maybeRoleKey, roleKey.tokenHasNoKeyStatus());
+        if (token.hasAdminKey()) {
+            context.requireKey(oneOf(
+                    replacementKey == null ? maybeRoleKey : allOf(maybeRoleKey, replacementKey),
+                    token.adminKeyOrThrow()));
+        } else {
+            context.requireKey(maybeRoleKey);
+            if (replacementKey != null) {
+                context.requireKey(replacementKey);
```
petreze, line 449: Whenever a low priority key wants to update itself to other key. In this case, the new key is **not** required to sign the transaction
i.e: `kycKey` wants to update itself to `newKycKey` -> we require only `kycKey` signature (part of the HIP requirements)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -429,31 +429,14 @@ public static boolean isExpiryOnlyUpdateOp(@NonNull final TokenUpdateTransaction
...
      */
-    public static boolean hasAlreadySomeNonAdminKeys(@NonNull final Token token) {
-        for (final var nonAdminKey : NON_ADMIN_TOKEN_KEYS) {
-            if (nonAdminKey.isPresentInitially(token)) {
-                return true;
-            }
-        }
-        return false;
+    protected static boolean updatesAdminOnlyNonKeyTokenProperty(@NonNull final TokenUpdateTransactionBody op) {
+        return !op.symbol().isEmpty() || !op.name().isEmpty() || op.hasAutoRenewPeriod() || op.hasMemo();
```
Neeharika-Sompalli, line 439: Its better to add a comment that we are already checking autoRenew Id and treasury. That's why we are not including here even though the name of the function suggests all fields except non-key token property

Neeharika-Sompalli:
> I totally agree with enhancing the `pureChecks()` and removing some unused methods as well as `requireNonNull()` and `OrThrow()` assertions, however why we remove `Hip540ChangeOrRemoveKeysSuite`. The file contains lots of corner cases to test every combination of `TokenUpdateTx` to examine every possible behaviour for this HIP.
> 
> With the provided changes in this PR, 17 from 29 tests in `Hip540ChangeOrRemoveKeysSuite` are failing as well as some of the already existing metadata HIP tests

@petreze I think is `Hip540ChangeOrRemoveKeysSuite` renamed to `Hip540Suite`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
tinker-michaelj: I have opened a [HIP PR](https://github.com/hashgraph/hedera-improvement-proposal/pull/950) to eliminate misleading use of the term "invalid"; we definitely don't want to introduce a way to put invalid key structures in state.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
tinker-michaelj: A `TokenUpdate` that tries to change a lower-privilege key without the admin key will still succeed...it would be nice to have more targeted response codes for trying to change non-key properties without an admin key, true. 🤔

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
tinker-michaelj: I have updated the HIP to clarify that without the admin signature, the new low-privilege key also needs to sign.

(This is the only meaningful interpretation of `TokenKeyValidation.NONE`, I think.)

Neeharika-Sompalli (approved):
LGTM ! We can add a String to the `HIP540TestScenario` to describe test name. What do you think ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenUpdateHandler.java
petreze: I see, this changes the initial requirements quite a bit

tinker-michaelj: True. :/

Issue #12412, opened 2024-03-27
Scalability - Improve Reconnect Metrics


artemananiev:
After each reconnect, learners report the following metrics in logs:

* timeInSeconds - total time reconnect took
* hashTimeInSeconds - hash time
* initializationTimeInSeconds - I am not sure about it
* totalNodes - total nodes transferred
* leafNodes - leaf nodes transferred
* redundantLeafNodes - see below
* internalNodes - internal nodes transferred
* redundantInternalNodes - see below

All metrics related to node counters are slightly confusing. In the current "push" reconnect implementation, they are set to

* totalNodes - number of lessons sent by the teacher; each lesson contain node data and children hashes
* internalNodes - number of lessons for internal nodes
* redundantInternalNodes = number of internal node lessons sent by the teacher, if the node is clean (same on the learner side)
* leafNodes - number of lessons for leaf nodes
* redundantLeadNodes - similar to redundantInternalNodes above

The problem here is that if a lesson is sent for a node, the lesson contains hashes for all its children. Today such a single lesson contributes to only totalNodes and one of internalNodes/leafNodes, but in reality information about more than one node is sent. redundantLeafNodes is always close to zero. It happens because virtual maps are large, and by the time when teacher sends a leaf node, a response from learner about the node is already received. For clean leaves, it looks like they are successfully skipped, but their hashes were still read by the teacher, sent to the learner (as a part of parent node's lesson), and then verified on the learner side. This should be reflected in leafNodes, but it is not. Same observation for redundantInternalNodes.

For virtual maps, another observation is any internal node lesson is redundant. Ideally, only dirty virtual leaves should be sent, since in virtual maps all data is in leaves. It means, for virtual maps, having internalNodes and redundantInternalNodes separate doesn't make sense. Only number of sent internal nodes does matter.

Here is my proposal (names are TBD):

* totalTransfers - total number of transfers. Today, it's equal to internalNodeTransfers + leafNodeTransfers, but it may not be true in the future
* internalNodeTransfers - number of internal nodes transferred. For virtual nodes, only hashes are transferred. For merkle nodes, it can be either a hash, or hash + data, it's still counted as one node
* leafHashTransfers - number of leaf node hashes transferred
* leafDataTransfers - number of leaf node data transferred
* cleanLeafDataTransfers - number of leaf node data transfers, when the node is clean (same on both sides)

For the current push implementation:

* totalTransfers - every lesson is a transfer
[...]

PR #13027, opened 2024-04-26
feat(reconnect): introduce ReconnectMapStats interface

anthony-swirldslabs:
**Description**:
Introducing an interface to help gather reconnect-related statistics for maps per a proposal at https://github.com/hashgraph/hedera-services/issues/12412#issuecomment-2076062254

It is out of scope of this particular PR to implement this interface, or start using it.

**Related issue(s)**:

Contributes to #12412

**Notes for reviewer**:
`gr assemble` passes.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
@@ -28,34 +28,48 @@ package com.swirlds.common.merkle.synchronization.stats;
  */
 public interface ReconnectMapStats {
     /**
-     * Increment a transfers counter.
+     * Increment a transfers from teacher counter.
      * <p>
-     * Different reconnect algorithms may define the term "transfer" differently. Examples of a transfer: <br>
+     * Different reconnect algorithms may define the term "transfer" differently. Examples of a transfer from teacher: <br>
      * * a lesson from the teacher, <br>
-     * * a query response to the teacher, <br>
-     * * a request from the learner, <br>
-     * * a response from the teacher.
-     *
-     * @param fromTeacher a transfer of data from a teacher to the learner (e.g. a lesson, or a response)
-     * @param fromLearner a transfer of data from the learner to its teacher (e.g. a query response, or a request)
+     * * a response from the teacher per a prior request from the learner.
      */
-    default void incrementTransfers(int fromTeacher, int fromLearner) {}
+    default void incrementTransfersFromTeacher() {}
 
     /**
-     * Gather stats about internal nodes transfers.
+     * Increment a transfers from learner counter.
+     * <p>
+     * Different reconnect algorithms may define the term "transfer" differently. Examples of a transfer from learner: <br>
+     * * a query response to the teacher for a single hash, <br>
+     * * a request from the learner.
+     */
+    default void incrementTransfersFromLearner() {}
+
+    /**
+     * Gather stats about internal nodes hashes transfers.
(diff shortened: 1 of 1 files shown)


jasperpotts (approved):

imalygin (dismissed):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Increment a transfers counter.
+     * <p>
+     * Different reconnect algorithms may define the term "transfer" differently. Examples of a transfer: <br>
+     * * a lesson from the teacher, <br>
+     * * a query response to the teacher, <br>
+     * * a request from the learner, <br>
+     * * a response from the teacher.
+     */
+    default void incrementTransfers() {}
```
artemananiev, line 39: OK, I think I understand "transfers" now. They may or may not be derivable from internal / leaf node counters. This is fine.

My next question is: is it fair to assume that all transfers include two parts: teacher to learner and learner to teacher? If reconnect implementations with uni-directional requests are anticipated, then we should track teacher to learner and learner to teacher transfers separately.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
```diff
@@ -0,0 +1,55 @@
...
+    default void incrementInternalNodes(int num, int cleanNum) {}
+
+    /**
+     * Gather stats about leaf nodes transfers.
+     * @param hashNum the number of hashes of leaf nodes transferred
+     * @param dataNum the number of data payloads of leaf nodes transferred
+     * @param cleanDataNum the number of data payloads transferred unnecessarily because they were clean
+     */
+    default void incrementLeafNodes(int hashNum, int dataNum, int cleanDataNum) {}
+}
```
artemananiev, line 55: Rather than tracking "internals" vs "leaves", I'd propose two orthogonal categories: "hashes" and "data". Note that internal nodes may contain data, too.

For "hashes", all we need to track is the number of hash transfers. For "data", number of transfers and number of redundant transfers (aka clean nodes).

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
anthony-swirldslabs: The transfers may not always be symmetrical. For example, the current reconnect algorithm may or may not include a query into a lesson. If a query is present, then the learner has to respond to the query and this adds one more transfer. However, if there weren't any queries, then the learner wouldn't send a response for this particular lesson, so only the initial lesson transfer would be counted then.

I agree that counting transfers in different directions might be interesting, so I'll add arguments for this method.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapStats.java
anthony-swirldslabs: I see your point. I propose we count both the dimensions for both the leaves and internal nodes separately, so that we have a full set of all the counters that can possibly be non-zero for any reconnect implementation. Also, as I mentioned off-line hash transfers may be clean and not clean, and the distinction is important, so we need to count these as well. I'll make the corresponding change in the next commit shortly.

artemananiev (approved):

PR #13029, opened 2024-04-27
Too large transactions should not trigger possibly catastrophic failure

### LINKED ISSUE #13004: fix: Rewrite grpc's "possibly CATASTROPHIC failure" which actually isn't possibly catastrophic, so it is appropriately less scary
If we get a gRPC message or something that's too large (I guess larger than we've configured gRPC for) it throws a `RuntimeException` with a really scary message which then shows up in our logs and frightens innocent developers, like me.

image

I'm given to understand that it isn't actually possibly catastrophic at all, nor even an error.  It's just a warning.  (Or so I've been told.)

It would be nice to catch that exception and rewrite it to be more temperate.

netopyr:
**Description**:

This PR excludes the check if a message is too large from the catch-all branch.

**Related issue(s)**:

Fixes #13004

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
@@ -109,17 +109,17 @@ public abstract class MethodBase implements ServerCalls.UnaryMethod<BufferedData
     @Override
     public void invoke(
             @NonNull final BufferedData requestBuffer, @NonNull final StreamObserver<BufferedData> responseObserver) {
+        // Track the number of times this method has been called
+        callsReceivedCounter.increment();
+        callsReceivedSpeedometer.cycle();
+
+        // Fail-fast if the request is too large (Note that the request buffer is sized to allow exactly
+        // 1 more byte than MAX_MESSAGE_SIZE, so we can detect this case).
+        if (requestBuffer.length() > MAX_MESSAGE_SIZE) {
+            throw new RuntimeException("More than " + MAX_MESSAGE_SIZE + " received");
+        }
+
         try {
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
```diff
@@ -109,17 +109,17 @@ public abstract class MethodBase implements ServerCalls.UnaryMethod<BufferedData
...
-                throw new RuntimeException("More than " + MAX_MESSAGE_SIZE + " received");
-            }
+        // Track the number of times this method has been called
+        callsReceivedCounter.increment();
+        callsReceivedSpeedometer.cycle();
+
+        // Fail-fast if the request is too large (Note that the request buffer is sized to allow exactly
+        // 1 more byte than MAX_MESSAGE_SIZE, so we can detect this case).
+        if (requestBuffer.length() > MAX_MESSAGE_SIZE) {
+            throw new RuntimeException("More than " + MAX_MESSAGE_SIZE + " received");
```
Neeharika-Sompalli, line 119: Should we throw this exception or just add a warning ? 
If we throw a RuntimeException here is it be fine ?

Neeharika-Sompalli (dismissed):
LGTM! Thanks @netopyr

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/MethodBase.java
netopyr: You are right. The error handling is wrong. I updated the file. Thanks, @Neeharika-Sompalli!

Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr

PR #13031, opened 2024-04-27
fix: remove gas cost check

### LINKED ISSUE #13030: Remove check if gas cost are less than maximum transaction fee
In the modularized code, we added a check when the estimated gas cost exceeds the maximum transaction fee, resulting in an `INSUFFICIENT_TX_FEE` error. We believe this change is the cause of the sudden increase in this type of error that has been reported. So, we decided to remove this check for now.

netopyr:
**Description**:

This PR removes the check that failed transactions when the estimated gas cost was higher than the maximum transaction fee.

**Related issue(s)**:

Fixes #13030


Neeharika-Sompalli (dismissed):
LGTM as per discussion! Thanks @netopyr

Neeharika-Sompalli (approved):
LGTM ! Thanks @netopyr

PR #13038, opened 2024-04-29
fix: storage link migration and management

tinker-michaelj:
**Description**:
 - Closes #13037 
 - Use the same `int[]` -> `Bytes` key migration for a contract account's first key as for the other storage slots.
 - Update the existing head slot's `prev` pointer when inserting a new slot.
 - _Only_ log an error and continue if linked list management fails from a broken invariant.

PR #13039, opened 2024-04-29
fix: storage link migration and management

tinker-michaelj:
**Description**:
 - Cherry-pick #13038


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

Nana-EC (approved):
LG based on Luke's review of the same code in develop

PR #13040, opened 2024-04-29
feat(HIP796): Implement HIP-796 moveUnitsBetweenSameUserPartitions

vvairava:
Description

Move the units from one partition to another partition of the same user account

Existing Files

Compile
./gradlew assemble is SUCCESS


tinker-michaelj (approved):
LGTM, ty @vvairava !

PR #13041, opened 2024-04-29
feat: HIP-904 Introduce `unlimitedAutoAssociations` feature flag

### LINKED ISSUE #12981: HIP-904: Add feature flag for unlimited token associations. Modify TokenConfig and EntityConfig to ensure we are not limiting maxAutoAssociations.

stoyanov-st:
**Description**:
With this PR we are introducing the `unlimitedAutoAssociations` feature flag in `EntitiesConfig`.
We also add relative changes to handler logic in regards to auto associations and relative tests.
**Related issue(s)**:

Fixes #12981 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
+++ b/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
@@ -83,4 +83,5 @@ public record ContractsConfig(
                 boolean chargeGasOnPreEvmException,
         @ConfigProperty(value = "evm.nonExtantContractsFail", defaultValue = "0") @NetworkProperty
                 Set<Long> evmNonExtantContractsFail,
-        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion) {}
+        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion,
+        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
--- a/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
+++ b/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
@@ -83,5 +83,4 @@ public record ContractsConfig(
                 boolean chargeGasOnPreEvmException,
         @ConfigProperty(value = "evm.nonExtantContractsFail", defaultValue = "0") @NetworkProperty
                 Set<Long> evmNonExtantContractsFail,
-        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion,
-        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
+        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion) {}
--- a/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
+++ b/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
@@ -24,4 +24,5 @@ import com.swirlds.config.api.ConfigProperty;
 public record EntitiesConfig(
         @ConfigProperty(defaultValue = "3153600000") @NetworkProperty long maxLifetime,
         // @ConfigProperty(defaultValue = "FILE") Set<EntityType> systemDeletable
-        @ConfigProperty(defaultValue = "false") @NetworkProperty boolean limitTokenAssociations) {}
+        @ConfigProperty(defaultValue = "false") @NetworkProperty boolean limitTokenAssociations,
+        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
--- a/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
+++ b/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
@@ -25,4 +25,4 @@ public record EntitiesConfig(
         @ConfigProperty(defaultValue = "3153600000") @NetworkProperty long maxLifetime,
         // @ConfigProperty(defaultValue = "FILE") Set<EntityType> systemDeletable
         @ConfigProperty(defaultValue = "false") @NetworkProperty boolean limitTokenAssociations,
-        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
+        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociationsEnabled) {}
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
@@ -54,7 +54,6 @@ import com.hedera.node.app.spi.workflows.HandleException;
 import com.hedera.node.app.spi.workflows.PreCheckException;
 import com.hedera.node.app.spi.workflows.PreHandleContext;
 import com.hedera.node.app.spi.workflows.TransactionHandler;
-import com.hedera.node.config.data.ContractsConfig;
 import com.hedera.node.config.data.EntitiesConfig;
 import com.hedera.node.config.data.TokensConfig;
 import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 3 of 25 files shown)


MrValioBg:
It seems like this PR includes adding the feature flag for both the mono and modularized codebase. Why do we need it in mono?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
```diff
@@ -170,18 +180,22 @@ private boolean isTotalNumTokenRelsWithinMax(
...
+            @NonNull final ContractsConfig contractsConfig,
             @NonNull final Account account,
             @NonNull final List<TokenID> tokenIds) {
         final var numAssociations = requireNonNull(account).numberAssociations();
         final var tokenAssociationsLimited = entitiesConfig.limitTokenAssociations();
         final var maxTokensPerAccount = config.maxPerAccount();
-        return !tokenAssociationsLimited || (numAssociations + tokenIds.size() <= maxTokensPerAccount);
+        return !tokenAssociationsLimited
+                || ((numAssociations + tokenIds.size() <= maxTokensPerAccount)
+                        && contractsConfig.unlimitedAutoAssociations());
```
MrValioBg, line 198: Shouldn't this be 
`return !tokenAssociationsLimited || contractsConfig.unlimitedAutoAssociations()  ||
            (numAssociations + tokenIds.size() <= maxTokensPerAccount);`
So its || instead of &&

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -83,4 +83,5 @@ public record ContractsConfig(
                 boolean chargeGasOnPreEvmException,
         @ConfigProperty(value = "evm.nonExtantContractsFail", defaultValue = "0") @NetworkProperty
                 Set<Long> evmNonExtantContractsFail,
-        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion) {}
+        @ConfigProperty(value = "evm.version", defaultValue = "v0.50") @NetworkProperty String evmVersion,
+        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
```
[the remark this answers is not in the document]
stoyanov-st: Mainly because we have the `allowAutoAssociations` in the same config.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenCreateValidator.java
```diff
@@ -183,12 +184,14 @@ private void validateSupplyType(final TokenSupplyType supplyType, final long max
...
+            @NonNull final ContractsConfig contractsConfig,
             @NonNull final Account account,
             @NonNull final Token token,
             @NonNull final WritableTokenRelationStore tokenRelStore) {
         validateFalse(
-                entitiesConfig.limitTokenAssociations()
-                        && account.numberAssociations() + 1 > tokensConfig.maxPerAccount(),
+                (entitiesConfig.limitTokenAssociations()
+                                && account.numberAssociations() + 1 > tokensConfig.maxPerAccount())
+                        || !contractsConfig.unlimitedAutoAssociations(),
```
MrValioBg, line 194: Do we need the unlimitedAutoAssociations, here? Aren't we overriding the limitTokenAssociations, one?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
MrValioBg: Discussed offline, we shouldn't override the tokenAssociationsLimited flag with the feature toggle.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenCreateValidator.java
```diff
@@ -191,8 +191,8 @@ public void validateAssociation(
             @NonNull final Token token,
             @NonNull final WritableTokenRelationStore tokenRelStore) {
         validateFalse(
-                entitiesConfig.limitTokenAssociations()
-                        && account.numberAssociations() + 1 > tokensConfig.maxPerAccount(),
+                (entitiesConfig.limitTokenAssociations()
+                        && account.numberAssociations() + 1 > tokensConfig.maxPerAccount()),
```
MrValioBg, line 195: Is this change needed? It seems like it only adds `()`

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
```diff
@@ -24,4 +24,5 @@
 public record EntitiesConfig(
         @ConfigProperty(defaultValue = "3153600000") @NetworkProperty long maxLifetime,
         // @ConfigProperty(defaultValue = "FILE") Set<EntityType> systemDeletable
-        @ConfigProperty(defaultValue = "false") @NetworkProperty boolean limitTokenAssociations) {}
+        @ConfigProperty(defaultValue = "false") @NetworkProperty boolean limitTokenAssociations,
+        @ConfigProperty(defaultValue = "true") @NetworkProperty boolean unlimitedAutoAssociations) {}
```
Neeharika-Sompalli, line 28: Should this be `unlimitedAutoAssociationsEnabled` or `allowUnlimitedAutoAssociations` for clarity

Neeharika-Sompalli, line 28: Also this should be in `TokensConfig`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoCreateValidator.java
```diff
@@ -98,7 +98,7 @@ public boolean tooManyAutoAssociations(
             @NonNull final LedgerConfig ledgerConfig,
             @NonNull final EntitiesConfig entitiesConfig,
             @NonNull final TokensConfig tokensConfig) {
-        return n > ledgerConfig.maxAutoAssociations()
-                || (entitiesConfig.limitTokenAssociations() && n > tokensConfig.maxPerAccount());
+        return (entitiesConfig.limitTokenAssociations() && n > tokensConfig.maxPerAccount())
+                || (!entitiesConfig.unlimitedAutoAssociations() && n > ledgerConfig.maxAutoAssociations());
```
Neeharika-Sompalli, line 102: Why do we need to check `!entitiesConfig.unlimitedAutoAssociations()` this here ? Even if the configuration is enabled, we should not be able to set a higher number. That configuration is just allowed for `-1` to be set in `maxAutoAssociations` right ?

stoyanov-st: From my understanding we should not limit the automatic associations if the configuration is enabled.
Or is it only if the value is -1 we have unlimited auto associations and we keep limiting the positive values e.g. 5000 is valid but 5001 - invalid?
@Neeharika-Sompalli @MiroslavGatsanoga  Can you provide more info on this?

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/EntitiesConfig.java
stoyanov-st: I will change it to `allowUnlimitedAutoAssociations`

netopyr: I think it should be `unlimitedAutoAssociationsEnabled`, because we use this pattern more frequently.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoCreateValidator.java
netopyr: I agree. If we allow unlimited auto associations, it is hard to explain why setting a concrete number should be limited. But this changes the current behavior and therefore we have to double-check (and eventually update the HIP).

MiroslavGatsanoga: > Or is it only if the value is -1 we have unlimited auto associations and we keep limiting the positive values e.g. 5000 is valid but 5001 - invalid
Yes, that should be the intended behaviour. Having the feature flag enabled should allow `maxAutoAssociations` value for any account to be set to -1.
If an account has `maxAutoAssociations = -1` then we have unlimited auto associations for this account.

Issue #13043, opened 2024-04-29
Update `calculateFees` to charge non-zero fee when the entity doesn't exist in state


Neeharika-Sompalli:
In `calculateFees` when the entity doesn't exist in state we are returning `FeeData.getDefaultInstance` in few cases, which results in charging zero fees. Change it to charge a minimal amount with constant value 1 set.

PR #13044, opened 2024-04-29
chore: Update default fees to a constant value

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/13043

- Fixes the handlers that are returning `FeeData.getDefaultInstance` when an entity doesn't exist in state to use a FeeData with a constant value set.


tinker-michaelj (dismissed):

tinker-michaelj (dismissed):

tinker-michaelj (approved):

PR #13045, opened 2024-04-29
feat: add `htsCallWithInsufficientGasHasNoStateChanges()` test

tinker-michaelj:
**Description**:
 - Add a `@HapiTest` verifying that a call to `0x167` that fails with insufficient gas doesn't change state, even if the calling frame doesn't revert after the failed low-level call.


lukelee-sl (approved):
LGTM

PR #13047, opened 2024-04-29
chore: Update default fees to a constant value (#13044)

Neeharika-Sompalli:
Cherry-pick https://github.com/hashgraph/hedera-services/pull/13044


tinker-michaelj (approved):

Issue #13052, opened 2024-04-30
fix: Exchange rate precompile return error when called with value


stefan-stefanooov:
### Description
Executing an internal call with value against Exchange rate precompile addresses returns success. Only the contract that acts as middleman is credited with hbars.

Fixes the second half of this issue: https://github.com/hashgraph/hedera-services/issues/11218

### Solution

Add a fix that forces a `INVALID_FEE_SUBMITTED` error to be returned for exchange rate precompile, when its called with value.

PR #12793, opened 2024-04-15
feat: internal call against PRNG returns success instead of invalid fee submitted

natanasow:
### Description

Executing an internal call with value against PRNG/Exchange rate precompile addresses returns success. Only the contract that acts as middleman is credited with hbars.

### Steps to reproduce

1. Create a contract with the following function:
```
 function makeCallWithAmount(address _to, bytes memory _data) external payable returns (bool success, bytes memory returnData) {
        (success, returnData) = _to.call{value: msg.value}(_data);
    }
```
2. Deploy the contract
3. Execute a contract call against the contract from step1/2 using the function with parameters(address _to) using the PRNG/ExchangeRate precompile addresses 
```
var data = encodeDataToByteArray("getPseudorandomSeed()");
var parameters = new ContractFunctionParameters()
                .addAddress("0x0000000000000000000000000000000000000169")
                .addBytes(data);
```

**Expected result**: The secondary call against the precompile address should fail with `INVALID_FEE_SUBMITTED`
**Actual result**: The secondary call is successful. 

Note 1: This is also reproducible with ExchangeRate precompile address
Note 2: This fixes only PRNG calls

**Related issue(s)**:

Fixes #https://github.com/hashgraph/hedera-services/issues/11218

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

Issue #11218, opened 2024-01-29
Executing an internal call with value against PRNG/Exchange rate precompile addresses returns success


yiliev0:
### Description

Executing an internal call with value against PRNG/Exchange rate precompile addresses returns success. Only the contract that acts as middleman is credited with hbars.

### Steps to reproduce

1. Create a contract with the following function:
```
 function makeCallWithAmount(address _to, bytes memory _data) external payable returns (bool success, bytes memory returnData) {
        (success, returnData) = _to.call{value: msg.value}(_data);
    }
```
2. Deploy the contract
3. Execute a contract call against the contract from step1/2 using the function with parameters(address _to) using the PRNG/ExchangeRate precompile addresses 
```
var data = encodeDataToByteArray("getPseudorandomSeed()");
var parameters = new ContractFunctionParameters()
                .addAddress("0x0000000000000000000000000000000000000169")
                .addBytes(data);
```

**Expected result**: The secondary call against the precompile address should fail with `INVALID_FEE_SUBMITTED`
**Actual result**: The secondary call is successful. 

Note: This is also reproducible with ExchangeRate precompile address

### Hedera network

other

### Version

0.46.0

### Operating system

None

0xivanov:
Note: `testTransferToNonExistingSystemAccounts` in test-clients

PR #13053, opened 2024-04-30
fix: Exchange rate precompile return error when called with value #13052

stefan-stefanooov:
**Description**:
Executing an internal call with value against PRNG/Exchange rate precompile addresses returns success. Only the contract that acts as middleman is credited with hbars.

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/pull/12793

Fixes #https://github.com/hashgraph/hedera-services/issues/11218


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/ExchangeRateSystemContractTest.java
```diff
@@ -100,6 +103,7 @@ void convertsPositiveNumberToTinycentsAsExpected() {
 
     @Test
     void convertsZeroToTinybarsAsExpected() {
+        given(frame.getValue()).willReturn(Wei.ZERO);
         givenRate(someRate);
 
         final var zeroInput = tinycentsInput(0);
```
lukelee-sl, line 109: Should there be a unit test that tests the new exception case?

lukelee-sl (commented):
Seems reasonable but should there not be a HAPI test to show that the transaction reverts as outlined in the original issue?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/ExchangeRateSystemContractTest.java
stefan-stefanooov: Added.

lukelee-sl (dismissed):
LGTM

stefan-stefanooov:
> Seems reasonable but should there not be a HAPI test to show that the transaction reverts as outlined in the original issue?

Hey @lukelee-sl 
Thank you for the review. 
I have added integration test as well.
Please have another look.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/ExchangeRateSystemContractTest.java
```diff
@@ -118,6 +122,17 @@ void inputCannotUnderflow() {
...
 
+    @Test
+    void valueShouldNotBeSentToThePrecompile() {
+        given(frame.getValue()).willReturn(Wei.MAX_WEI);
+
+        final var zeroInput = tinycentsInput(0);
+        final var result = subject.computeFully(zeroInput, frame);
+
+        assertThat(result.output()).isEqualTo(Bytes.EMPTY);
+        assertThat(result.result().getHaltReason().get()).isEqualTo(ExceptionalHaltReason.INVALID_OPERATION);
```
petreze, line 133: Shouldn't we test for the new `INVALID_FEE_SUBMITTED` case?

stefan-stefanooov: A new version available. Thank you.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/PrngSystemContract.java
```diff
@@ -102,7 +102,7 @@ public FullResult computeFully(@NonNull final Bytes input, @NonNull final Messag
                     PrecompiledContract.PrecompileContractResult.halt(Bytes.EMPTY, Optional.of(INVALID_OPERATION)),
                     gasRequirement,
                     null);
-        } catch (NullPointerException e) {
+        } catch (Exception e) {
```
stefan-stefanooov, line 105: Not in the scope, but in the current flow we handle only two types of exceptions. If we have something throw another  exception type, we are going to leak it out. 
Putting the most common type at the end of the `catch` chain, we ensure that all exceptions are handled.

Discussed with the team. 
Any comments?

lukelee-sl:
> > Seems reasonable but should there not be a HAPI test to show that the transaction reverts as outlined in the original issue?
> 
> Hey @lukelee-sl Thank you for the review. I have added integration test as well. Please have another look.

Thanks for adding the test. Looks good.

lukelee-sl (approved):
LGTM

petreze (approved):

PR #13055, opened 2024-04-30
fix: storage link mgmt with multiple inserts

### LINKED ISSUE #13054: Storage links can be broken
### Description

Storage links can be broken in certain conditions.

### Steps to reproduce

Run [this spec](https://github.com/hashgraph/hedera-services/blob/13054-fix-storage-links/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractStateSuite.java).

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Fixes #13054 
 - If an account's `firstContractStorageKey` has already been updated in the transaction, use its latest value.


Neeharika-Sompalli (approved):
Makes sense to me! Thanks @tinker-michaelj

lukelee-sl (approved):
LGTM

PR #13070, opened 2024-05-05
feat: add javadoc and diagram, delete dead code

tinker-michaelj:
**Description**:
 - Add javadoc to `ServicesMain#main()` and a few other places.
 - Add a [diagram](https://github.com/hashgraph/hedera-services/blob/1496274e9d766823e66049546a1bd22d47553d51/hedera-node/hedera-app/src/main/java/com/hedera/node/app/startup-phase-lifecycle.png) to try to help visualize the central role of the `Hedera` singleton in the startup phase of the application.
 - Delete some dead code.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/startup-phase-lifecycle.png
mhess-swl, line 1: This is great! 🤩 

My only thought is that this might be even more friendly as a sequence diagram, but I'm still able to follow it 👏

mhess-swl (approved):

thomas-swirlds-labs (approved):
Looks great! Thanks @tinker-michaelj  🥳 🙌

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -117,9 +121,37 @@ public void run() {
...
+     *     validation and register notification listeners on the platform.</li>
+     *     <li>Invoke {@link Platform#start()}.</li>
+     * </ol>
      *
-     * @param args First arg, if specified, will be the node ID
+     * <p>Please see the <i>startup-phase-lifecycle.png</i> in this directory to visualize
+     * the sequence of events in the startup phase and the centrality of the {@link Hedera}
+     * singleton.
+     *
+     * @param args optionally, what node id to run; required if the address book is ambiguous
```
Neeharika-Sompalli, line 154: This is an amazing description 🙌

Neeharika-Sompalli (approved):
Looks Great @tinker-michaelj . Thanks for putting the diagram and very detailed javadocs 🙌

derektriley:
From my perspective just having browsed the code repository briefly, I think it would be useful to add class/function names for some additional context throughout the diagram. I found myself diving into various classes looking for what this diagram is depicting. I annotated some things below as just an example (could be incorporated in a better format). Also to Matt's point a sequence diagram with class/function names would be great in addition to this diagram.

image

Issue #12853, opened 2024-04-17
Memory leak from MerkleDbDataSource.copyStatisticsFrom()


OlegMazurov:
### Description

`MerkleDbDataSource.copyStatisticsFrom()` is called from `VirtualRootNode.setupWithOriginalNode()` when a node enters reconnect as a learner and creates a new `dataSource`. It copies a reference to `statisticsUpdater` from the old `dataSource` into the new one. `statisticsUpdater`, however, keeps a reference to the old `dataSource` which will not be garbage collected after it is closed.

Another issue with the copy is that  the new `dataSource` created its own `statisticsUpdater` and got it linked in various lambdas. As `statisticsUpdater` is lazily initialized (via `registerMetrics()`) those lambdas will be using uninitialized `statisticsUpdater`, not the one copied from the old `dataSource`. As a result, metrics reported via those lambdas will be missing.

### Steps to reproduce

Memory dump histogram shows twice the number of `MerkleDbDataSources`, `MerkleDbStatisticsUpdaters` compared to a healthy node.
Metrics like `hashesStoreFileCount` stop being reported after reconnect.

### Hedera network

other

### Version

v0.49

### Operating system

Linux

PR #13097, opened 2024-05-06
fix: 12853: Memory leak from MerkleDbDataSource.copyStatisticsFrom()

artemananiev:
Fix summary: a reference to `MerkleDbDataSource` is removed from `MerkleDbStatisticsUpdater` class. Instead, a data source instance (to pull stats from) is passed as an arg to all statistics updater methods.

Fixes: https://github.com/hashgraph/hedera-services/issues/12853

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatisticsUpdater.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatisticsUpdater.java
@@ -21,23 +21,17 @@ import static com.swirlds.base.units.UnitConstants.BYTES_TO_MEBIBYTES;
 import com.swirlds.common.metrics.FunctionGauge;
 import com.swirlds.merkledb.collections.LongList;
 import com.swirlds.merkledb.collections.OffHeapUser;
+import com.swirlds.merkledb.config.MerkleDbConfig;
 import com.swirlds.merkledb.files.DataFileReader;
 import com.swirlds.metrics.api.Metrics;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.LongSummaryStatistics;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.IntConsumer;
 
 /**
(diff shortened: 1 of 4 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbStatisticsUpdater.java
```diff
@@ -48,22 +42,18 @@ public class MerkleDbStatisticsUpdater {
             .withFormat("%d");
 
     private final MerkleDbStatistics statistics;
-    private final MerkleDbDataSource<?, ?> dataSource;
```
artemananiev, line 51: This is the core of the fix. Since `MerkleDbStatisticsUpdater` may be reused across different `MerkleDbDataSource` instances, they can't hold a reference to a data source, otherwise the described memory leak would be observed. Instead, a data source instance is passed to all updater methods, so it knows where to pull stats from.

platform-sdk/swirlds-jasperdb/src/testFixtures/java/com/swirlds/merkledb/test/fixtures/TestType.java
```diff
@@ -79,8 +81,28 @@ public <K extends VirtualKey, V extends VirtualValue> DataTypeConfig<K, V> dataT
         return new DataTypeConfig<>(this);
     }
 
+    public Metrics getMetrics() {
+        if (metrics == null) {
```
anthony-swirldslabs, line 85: I hope tests that use this never run in parallel (now or in the future.) If that's a possibility even in theory, making the method `synchronized` might be a good idea.

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/testFixtures/java/com/swirlds/merkledb/test/fixtures/TestType.java
artemananiev: Yes, tests are single threaded

imalygin (approved):

PR #13104, opened 2024-05-07
fix: ignore token expiry status in `TokenDissociate`

### LINKED ISSUE #13102: Token expiration should not affect `TokenDissociate` semantics
### Description

A `TokenDissociate` with an expired token succeeded [here](https://hashscan.io/mainnet/transaction/0.0.928499@1714301115.696195061) despite the dissociated relationship still having a non-zero balance.

### Steps to reproduce

As above; in fact, [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/leaky/LeakyCryptoTestsSuite.java#L511) `HapiSpec` is an existing reproduction. (This was expected behavior with rent/expiry enabled.)

### Additional context

_No response_

### Hedera network

mainnet

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #13102 
 - Remove behavior of allowing dissociation and doing treasury return when a token is expired.
     * This is expected when expiry and rent are enabled, but incorrect without them.
 - Rename and enable spec `cannotDissociateFromExpiredTokenWithNonZeroBalance()` as a `@HapiTest` to validate new behavior.

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java
@@ -50,12 +50,12 @@ import com.hedera.node.app.spi.fees.FeeContext;
 import com.hedera.node.app.spi.fees.Fees;
 import com.hedera.node.app.spi.validation.ExpiryValidator;
 import com.hedera.node.app.spi.workflows.HandleContext;
+import com.hedera.node.app.spi.workflows.HandleException;
 import com.hedera.node.app.spi.workflows.PreCheckException;
 import com.hedera.node.app.spi.workflows.PreHandleContext;
 import com.hedera.node.app.spi.workflows.TransactionHandler;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import java.time.Instant;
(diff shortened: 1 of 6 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenDissociateFromAccountHandler.java
```diff
@@ -147,20 +147,10 @@ public void handle(@NonNull final HandleContext context) {
...
-                    if (treasuryTokenRel != null) {
-                        final var updatedTreasuryBalanceTokenRel = treasuryTokenRel.balance() + tokenRelBalance;
-                        treasuryBalancesToUpdate.add(treasuryTokenRel
-                                .copyBuilder()
-                                .balance(updatedTreasuryBalanceTokenRel)
-                                .build());
+                    if (token.tokenType() == NON_FUNGIBLE_UNIQUE) {
+                        throw new HandleException(ACCOUNT_STILL_OWNS_NFTS);
+                    } else {
+                        throw new HandleException(TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES);
```
Neeharika-Sompalli, line 153: We could probably add `FUTURE` comment, to add this same functionality when we enable expiration.

Neeharika-Sompalli (approved):
Makes sense to me. Except for a comment. Thanks @tinker-michaelj

PR #13107, opened 2024-05-07
refactor: refactor and rename hts classes for reuse

lukelee-sl:
**Description**:
This pr has a large change set but almost all changes are the renaming of base interfaces and classes used in the HTS system contracts in order to promote reuse when implementing the Hedera Account Service (HAS) system contract.

Changes include:
- create a common package under `com.hedera.node.app.service.contract.impl.exec.systemcontracts` to house common system contract interfaces and classes.
- Remove references to `Hts` in names in common classes.
- Create a common `AbstractCallAttempt` base class so that the generic type passed into `CallTranslator` and `AbstractCallTranslator` can be bounded.  This base class is empty for now but will contain common functionality in future prs.


tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

PR #12813, opened 2024-04-15
chore: update quickstart documentation

jjohannes:
**Description**:
- Update documentation (quickstart), mainly on Gradle and IntelliJ
- Add a `runTestClient` to allow you to run one of the "test clients" for a _services node_ as a Java process (not as test). This was described in the docs but not easily possible anymore with the current setup. (Do we want this?)

You can browse all the updated "quickstart" docs here:
https://github.com/hashgraph/hedera-services/tree/11568-update-dev-docs/docs

**Related issue(s)**:

#11568

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #13108, opened 2024-05-07
build: require a fixed 'major.minor.patch' Java version to be installed

### LINKED ISSUE #12798: Gradle Build: Exact JDK version (major.minor.patch) provisioning
The JDK we currently use on CI is `21.0.1`. The exact version is important as it is written into `module-info.class` files.  That's why we add it as input. Gradle's toolchain mechanism currently only tracks the major version. That's why we need the full version as a separate input:

https://github.com/hashgraph/hedera-services/blob/develop/build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts#L126-L130

(Note: the code is not 100% correct atm. It uses the version of the JDK running Gradle. On local machines, this may be a different version than the one provisioned by toolchains if Gradle is started with a different "major" JDK.)

Right now, users starting Gradle with a different JDK than `21.0.1` will not get any remote cache hits for compilation!

If we can't get this working reliably with toolchains (see https://github.com/gradle/gradle/issues/16628#issuecomment-2056923926) we should probably make sure the version we want is exactly the one Gradle was started with and let Gradle use that one. For that we could do a more precise version of the check proposed in: https://github.com/hashgraph/hedera-services/pull/12780

jjohannes:
**Description**:

This PR effectively removes using the [Gradle Java Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) feature, as it cannot be used to auto-download a precise JDK version (see https://github.com/gradle/gradle/issues/16628#issuecomment-2097540617). After experimenting with different things, I came to the conclusion that this is the best solution with current Gradle features.

This requires users to have the exact JDK required installed. This is documented in the "quickstart" guides once #12813 is merged.

I'll advocate for improving Gradle in this area and keep #12798 open for a better solution in the future. More details in that issue.

**Related issue(s)**:

A first solution for #12798

**Notes for reviewer**:

**Checklist**


build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
```diff
@@ -38,33 +38,18 @@ plugins {
...
-            " or above."
+// Fail the build if Gradle is started with a Java version that does not exactly match
+val javaVersion = "21.0.1"
+val currentJavaVersion = providers.systemProperty("java.version").get()
+
+if (currentJavaVersion != javaVersion) {
+    throw RuntimeException(
+        "Gradle runs with Java $currentJavaVersion. This project requires Gradle to run with Java $javaVersion. " +
+            "\n - From commandline: change JAVA_HOME and/or PATH to point at Java $javaVersion installation." +
+            "\n - From IntelliJ: change 'Gradle JVM' in 'Gradle Settings' to point at Java $javaVersion installation."
```
jjohannes, line 49: Same check as before, but for the **exact** version a a bit of rewording.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
```diff
@@ -38,33 +38,18 @@ plugins {
...
-java {
-    sourceCompatibility = MIN_JAVA_VERSION
-    targetCompatibility = MIN_JAVA_VERSION
-
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(21)
-        vendor = JvmVendorSpec.ADOPTIUM
-    }
-}
-
```
jjohannes, line 67: This does not add anything anymore, as Gradle will use the JDK it is running with by default. And we now check before that that is the one we need for everything.

build-logic/settings-plugins/src/main/kotlin/com.hedera.hashgraph.settings.settings.gradle.kts
```diff
@@ -24,7 +24,6 @@ pluginManagement {
 
 plugins {
     id("com.gradle.enterprise")
-    id("org.gradle.toolchains.foojay-resolver-convention")
```
jjohannes, line 27: We cannot use the auto-provisioning, there is no point in adding this plugin for it.

build-logic/project-plugins/src/main/kotlin/com.hedera.hashgraph.java.gradle.kts
anthony-swirldslabs: Personally, I find it limiting to fix a particular version of JDK for the build as this creates an extra friction when trying to experiment with newer JDKs, or may create real problems when a JDK requires an urgent update (e.g. due to a security vulnerability or similar.) So I cannot say I like this fix.

However, I do understand the intention of this change. And personally, I'd prefer if a minor (or even major) version mismatch was treated as a warning rather than an error (with a proper minimum version check, of course - that should still result in an error.)

But as far as the original idea is concerned, from functional perspective the fix looks correct to me. So I'm approving it.

anthony-swirldslabs (approved):

jjohannes:
Thank you for the feedback @anthony-swirldslabs. I am also not super happy.
I changed this to only fail in cases where the build will not work at all (as it was before).
In other cases, we print a warning if the version does not match exactly. This allows for experimentation and does not block you if you don't have the right minor version handy. The build works then, but wont't pull compilation results things from the remote cache.

anthony-swirldslabs (approved):

anthony-swirldslabs:
Thanks, @jjohannes . The new version of the fix looks great. Approved.

jeromy-cannon (approved):

PR #13115, opened 2024-05-07
chore: move towards making `Hash` immutable

### LINKED ISSUE #12456: `Hash` should become immutable
Having immutability of an instance is a good feature by itself. But right now, the driving force behind making `Hash` immutable is the switch protobuf. Events will contain hashes that are immutable. Ideally, we would wrap these immutable objects with a `Hash` instance instead of copying it. Because `Hash` allows access to the underlying array, this is not possible.

### Tasks
- Deprecate `Hash.getValue()`
- Refactor usage of `Hash.getValue()`. This will have to be done on a case by case basis. Most of the time, helper methods should be able to provide the same functionality without access to the underlying array. In the case where this is not possible, we can create a copy if its not cost prohibitive.
- Remove `Hash.getValue()`
- Hashes from the PBJ records should simply wrap the `Bytes` provided

### Sub tickets
- [x] #13427
- [x] #13510
- [ ] #13511

lpetrovic05:
part of #12456


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
```diff
@@ -97,6 +98,24 @@ default Hash digestSync(final SerializableHashable serializableHashable, final D
...
+    /**
+     * Computes a cryptographic hash (message digest) for the given message.
+     *
+     * @param message    the message contents to be hashed
+     * @param digestType the type of digest used to compute the hash
+     * @return the cryptographic hash for the given message contents
+     * @throws CryptographyException if an unrecoverable error occurs while computing the digest
+     */
+    @NonNull
+    byte[] digestBytesSync(@NonNull final byte[] message, @NonNull final DigestType digestType);
```
cody-littley, line 117: Are you intentionally returning a `byte[]` here instead of a `Hash`?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -79,7 +79,22 @@ public Hash(final byte[] value) {
...
      */
     public Hash(final byte[] value, final DigestType digestType) {
-        this(value, digestType, false, false);
+        if (value == null) {
+            throw new IllegalArgumentException("value");
+        }
+
+        if (digestType == null) {
+            throw new IllegalArgumentException("digestType");
+        }
```
cody-littley, line 88: Could we instead use `Objects.reqireNonNull()`?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -156,6 +135,13 @@ public byte[] getValue() {
         return value;
     }
 
+    /**
+     * @return a copy of the hash value as a byte array
+     */
+    public byte[] copyToByteArray() {
+        return Bytes.EMPTY.toByteArray();
+    }
```
cody-littley, line 143: Is this a place holder? It seems like this is always just returning an empty array.

```suggestion
    @NonNull
    public byte[] copyToByteArray() {
        return Bytes.EMPTY.toByteArray();
    }
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -79,7 +79,22 @@ public Hash(final byte[] value) {
      * 		the digest type
      */
     public Hash(final byte[] value, final DigestType digestType) {
-        this(value, digestType, false, false);
+        if (value == null) {
```
cody-littley, line 82: Maybe a good time to add nullity annotations to the constructor arguments.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -289,10 +276,10 @@ public String toString() {
     }
 
     /**
-     * Create a short string representation of this hash.
+     * Create a hexadecimal string representation of this hash.
      */
-    public String toShortString() {
-        return toShortString(SHORT_STRING_BYTES);
+    public String toHex() {
```
cody-littley, line 281:
```suggestion
    @NonNull
    public String toHex() {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -301,7 +288,7 @@ public String toShortString() {
      * @param length
      * 		the number of characters to include in the short string
      */
-    public String toShortString(final int length) {
+    public String toHex(final int length) {
```
cody-littley, line 291:
```suggestion
    @NonNull
    public String toHex(final int length) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/CryptoEngine.java
```diff
@@ -344,7 +350,8 @@ protected synchronized void applySettings() {
      * @param provider   the underlying provider to be used
      * @return the cryptographic hash for the given message contents
      */
-    private Hash digestSyncInternal(final byte[] message, final DigestType digestType, final DigestProvider provider) {
+    private byte[] digestSyncInternal(
```
cody-littley, line 353:
```suggestion
    @NonNull
    private byte[] digestSyncInternal(
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/CryptoEngine.java
```diff
@@ -344,7 +350,8 @@ protected synchronized void applySettings() {
...
+            final byte[] message, final DigestType digestType, final DigestProvider provider) {
```
cody-littley, line 354:
```suggestion
            @NonNull final byte[] message, @NonNull final DigestType digestType, @NonNull final DigestProvider provider) {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/DigestProvider.java
```diff
@@ -45,7 +44,7 @@ public DigestProvider() {
      * @throws NoSuchAlgorithmException
      * 		if an implementation of the required algorithm cannot be located or loaded
      */
-    protected Hash compute(final byte[] msg) throws NoSuchAlgorithmException {
+    protected byte[] compute(final byte[] msg) throws NoSuchAlgorithmException {
```
cody-littley, line 47:
```suggestion
    @NonNull
    protected byte[] compute(@NonNull final byte[] msg) throws NoSuchAlgorithmException {
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/DigestProvider.java
```diff
@@ -60,32 +59,14 @@ protected Hash compute(final byte[] msg) throws NoSuchAlgorithmException {
      * @throws NoSuchAlgorithmException
      * 		if an implementation of the required algorithm cannot be located or loaded
      */
-    protected Hash compute(final byte[] msg, final DigestType algorithmType) throws NoSuchAlgorithmException {
+    protected byte[] compute(final byte[] msg, final DigestType algorithmType) throws NoSuchAlgorithmException {
```
cody-littley, line 62: nullity annotations

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/DigestProvider.java
```diff
@@ -120,12 +101,12 @@ protected MessageDigest handleAlgorithmRequired(final DigestType algorithmType)
      * {@inheritDoc}
      */
     @Override
-    protected Hash handleItem(
+    protected byte[] handleItem(
```
cody-littley, line 104:
```suggestion
    @NonNull
    protected byte[] handleItem(
```

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
anthony-swirldslabs: Yeah, I believe we want to return `this.bytes.toByteArray()` here.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/internal/MerkleInternalDigestProvider.java
```diff
@@ -65,7 +65,7 @@ protected Hash handleItem(
         for (int index = 0; index < childHashes.size(); index++) {
             final Hash childHash = childHashes.get(index);
 
-            if (childHash == null || childHash.getValue() == null) {
+            if (childHash == null || childHash.getBytes() == null) {
```
anthony-swirldslabs, line 68: Can the hash bytes (or the former Hash value) actually be `null`? We could ensure their non-null status in the Hash class itself and avoid this check here.

mxtartaglia-sl (dismissed):
Nothing to add to Cody's and Antony's point

hendrikebbers:
My general thoughts on `Hash`:
If we refactor `Hash` to an interface `ImmutableHash` could be a record. That would make it way easier to define it as a public api in future.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: this is a big mistake, no idea how it happened. fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/DigestProvider.java
lpetrovic05: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/crypto/internal/MerkleInternalDigestProvider.java
lpetrovic05: It can be for now, until we get self serializable. When deserializing, we first construct an empty hash, then read the value. when the hash is constructed, its value is null. once we stop serializing hashes we can ensure that.

lpetrovic05:
> My general thoughts on `Hash`: If we refactor `Hash` to an interface `ImmutableHash` could be a record. That would make it way easier to define it as a public api in future.

We can make it a record in the future. Right now, my goal is to remove the access to the underlying `byte[]`. We also need to stop making it self serializable as well.

anthony-swirldslabs (approved):

cody-littley (approved):

jeromy-cannon (approved):

mxtartaglia-sl (approved):

hendrikebbers (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -26,21 +26,23 @@
...
 /**
  * A cryptographic hash of some data.
  */
 public class Hash implements Comparable<Hash>, SerializableWithKnownLength, Serializable {
-    private static final int SHORT_STRING_BYTES = 4;
     public static final long CLASS_ID = 0xf422da83a251741eL;
     public static final int CLASS_VERSION = 1;
 
     private byte[] value;
+    private Bytes bytes;
```
rbair23, line 45: Are we storing the same bytes twice?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -101,74 +112,45 @@ public Hash(final ByteBuffer byteBuffer, final DigestType digestType) {
...
         return value;
     }
 
+    /**
+     * @return a copy of the hash value as a byte array
+     */
+    public @NonNull byte[] copyToByteArray() {
+        return bytes.toByteArray();
+    }
+
```
rbair23, line 140: Why do both of these methods exist?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
david-bakin-sl: No, the second field is a `com.hedera.pbj.runtime.io.buffer.Bytes` which is a tiny thing that holds a ~pointer~reference to the `byte[] value` + offset + length.  So it's sharing the `value` array.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -262,7 +245,7 @@ public int hashCode() {
      * {@inheritDoc}
      */
     @Override
-    public int compareTo(final Hash that) {
+    public int compareTo(@NonNull final Hash that) {
```
david-bakin-sl, line 248: `@NonNull` not necessary as tested explicitly at L253-255.

Perhaps this is an IntelliJ catch-22?  It claims you _need_ the `@NonNull` annotation because the overridden method in interface `java.lang.Comparable` is annotated `@NonNull` - but that annotation is IntelliJ's own doing!  The _real_ method isn't annotated that way! See https://github.com/openjdk/jdk/blob/0bb5ae645165b97527ecccf02308df6072c363d8/src/java.base/share/classes/java/lang/Comparable.java#L141 - and it is in fact declared to throw `NullPointerException` when the argument is `null` - https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Comparable.html#compareTo(T)

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/ImmutableHash.java
```diff
@@ -53,7 +53,6 @@ public ImmutableHash(final Hash mutable) {
      */
     @Override
     public byte[] getValue() {
-        final byte[] value = super.getValue();
-        return Arrays.copyOf(value, value.length);
+        return copyToByteArray();
```
david-bakin-sl, line 56: Not sure this copy is necessary.  Given the two constructors above, and the rest of the constructors at the parent class `Hash`, I think _all_ constructors do the copy.  ???

platform-sdk/swirlds-fcqueue/src/main/java/com/swirlds/fcqueue/FCQueue.java
```diff
@@ -88,8 +88,10 @@ private static class ClassVersion {
 
     private static final long HASH_RADIX = 3;
 
+    /** The bytes of a NULL_HASH */
+    private static final byte[] NULL_HASH_BYTES = new byte[DIGEST_TYPE.digestLength()];
     /** A hash value representing a null element or a destroyed queue */
-    private static final ImmutableHash NULL_HASH = new ImmutableHash(new byte[DIGEST_TYPE.digestLength()]);
+    private static final ImmutableHash NULL_HASH = new ImmutableHash(NULL_HASH_BYTES);
```
david-bakin-sl, line 94: `NULL_HASH` also supports the idea of a `Hash.ZERO`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/shadowgraph/SyncLogging.java
```diff
@@ -46,9 +45,7 @@ public static String toShortHashes(List<Hash> hashes) {
         if (hashes == null) {
             return "null";
         }
-        return hashes.stream()
-                .map(h -> CommonUtils.hex(h.getValue(), BRIEF_HASH_LENGTH))
-                .collect(Collectors.joining(","));
+        return hashes.stream().map(h -> h.toHex(BRIEF_HASH_LENGTH)).collect(Collectors.joining(","));
```
david-bakin-sl, line 48: Not sure how often any of the methods here in `SyncLogging` are called, but if they are called reasonably often might be worth (in the future) rewriting these to _not_ use Java Streams.  

Though really convenient and usually nicely expressive they aren't high performant and generate lots of short-term intermediate objects.  Use for logging you try to minimize that, and if it was noticable a bespoke implementation that did it "brute force" using string builders in a loop would probably be better.

david-bakin-sl (approved):
Lots of comments for your consideration.

I don't think I've done so many code reviews for you.  Perhaps not any!  Anyway, so you know:  My style is to comment on anything I find interesting, and to let you use your own judgment on what you think of them.  If I find a bug then I "request changes", otherwise I "approve" and your actions are up to you.  I'm doing this in the hopes that when you do code reviews for _me_ you'll let me know what _you_ think and that way I'll learn some good stuff.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: The idea is that `value` will be removed as soon as possible, but it was too much to put into 1 PR

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: `getValue` is supposed to be removed. I should have marked it as deprecated, I would rather not do that now since I need approvals from 8 teams for this PR

lpetrovic05: `getValue` is supposed to be removed because we want to prevent access to the byte array. the new method explicitly says that a copy of the array is made, so that it is clear to anyone calling it that it has a performance impact

lpetrovic05:
> Lots of comments for your consideration.
> 
> I don't think I've done so many code reviews for you. Perhaps not any! Anyway, so you know: My style is to comment on anything I find interesting, and to let you use your own judgment on what you think of them. If I find a bug then I "request changes", otherwise I "approve" and your actions are up to you. I'm doing this in the hopes that when you do code reviews for _me_ you'll let me know what _you_ think and that way I'll learn some good stuff.
> 
> P.S. please fix the PR description by copying relevant bits from the issue's description. This will help you create a proper commit message for the merge, later. Also, this is a "feat" not a "chore" IMO.

Feat as opposed to chore makes sense. As for copying the ticket into the description, I don't see the point in that, I don't like duplicating data/code/text 😄

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -101,74 +112,45 @@ public Hash(final ByteBuffer byteBuffer, final DigestType digestType) {
...
+        this.bytes = other.bytes;
     }
 
     /**
      * Get the byte array representing the value of the hash.
      *
      * @return the hash value
      */
-    public byte[] getValue() {
+    public @NonNull byte[] getValue() {
```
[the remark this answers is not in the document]
lpetrovic05: It should have been deprecated, but I forgot to do it. I could easily add it now, but this PR needs approval from 8 teams, I am missing just 1 now. If I make any change that will mean another 5 days of waiting for approvals.

hedera-node/hedera-app/src/jmh/java/com/hedera/node/app/signature/impl/DoNothingCryptoEngine.java
```diff
@@ -42,6 +42,11 @@ public Hash digestSync(SerializableHashable serializableHashable, DigestType dig
         return null;
     }
 
+    @Override
+    public byte[] digestBytesSync(final byte[] message, final DigestType digestType) {
```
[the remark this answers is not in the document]
lpetrovic05: all the other methods return null, so I made this one do the same

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/BlockMetaSource.java
```diff
@@ -16,15 +16,13 @@
...
-import com.swirlds.common.crypto.DigestType;
-import com.swirlds.common.crypto.ImmutableHash;
 import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.evm.frame.BlockValues;
 
 /** Provides block information to a {@link HederaEvmTxProcessor}. */
 public interface BlockMetaSource {
-    Hash UNAVAILABLE_BLOCK_HASH = ethHashFrom(new ImmutableHash(new byte[DigestType.SHA_384.digestLength()]));
+    Hash UNAVAILABLE_BLOCK_HASH = org.hyperledger.besu.datatypes.Hash.wrap(Bytes32.wrap(new byte[32]));
```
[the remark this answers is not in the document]
lpetrovic05: I did not realize `Bytes32.ZERO` existed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
```diff
@@ -97,6 +98,24 @@ default Hash digestSync(final SerializableHashable serializableHashable, final D
      */
     Hash digestSync(final SerializableHashable serializableHashable, final DigestType digestType, boolean setHash);
 
+    /**
+     * Same as {@link #digestBytesSync(byte[], DigestType)} with DigestType set to {@link DigestType#SHA_384}
+     */
+    default byte[] digestBytesSync(@NonNull final byte[] message) {
```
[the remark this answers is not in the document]
lpetrovic05: synchronous, it means its done on the caller thread. this whole subsystem is something we are hoping to get rid of entirely

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmBlocks.java
```diff
@@ -29,7 +26,7 @@
  * Provides block information as context for a {@link HederaEvmTxProcessor}.
  */
 public interface HederaEvmBlocks {
-    Hash UNAVAILABLE_BLOCK_HASH = besuHashFrom(new ImmutableHash(new byte[DigestType.SHA_384.digestLength()]));
+    Hash UNAVAILABLE_BLOCK_HASH = org.hyperledger.besu.datatypes.Hash.wrap(Bytes32.wrap(new byte[32]));
```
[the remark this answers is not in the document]
lpetrovic05: `Hash.ZERO` does not make sense as a concept to me, a hash where all the bits are 0 seems just like a very unlikely value. I don't know why its used here, I did not try to figure out, I just wanted to remove the access to the underlying byte array

platform-sdk/swirlds-merkle/src/testFixtures/java/com/swirlds/merkle/test/fixtures/map/lifecycle/SaveExpectedMapHandler.java
```diff
@@ -117,20 +128,17 @@ public static String serialize(
      * 		the JSON file
      * @return ExpectedMap de-serialized from JSON file serialized to disk
      */
-    public static Map<MapKey, ExpectedValue> deserialize(final File sourceFile) {
+    public static Map<MapKey, ExpectedValue> deserialize(final File sourceFile) throws IOException {
```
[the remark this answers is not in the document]
lpetrovic05: Its a method used in a unit test. The test was failing, but the exception was not being propagated, so it took me a while to figure out why the test was failing. So now it just fails loudly if it fails.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
```diff
@@ -95,34 +97,34 @@ public static long[][] deepClone(long[][] original) {
...
      * @return 1 if first is bigger, -1 if second, 0 otherwise
      */
-    public static int arrayCompare(byte[] sig1, byte[] sig2) {
-        if (sig1 == null && sig2 == null) {
+    public static int arrayCompare(@Nullable final Bytes b1, @Nullable final Bytes b2) {
+        if (b1 == null && b2 == null) {
             return 0;
         }
-        if (sig1 == null && sig2 != null) {
+        if (b1 == null && b2 != null) {
```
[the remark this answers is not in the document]
lpetrovic05: I tried not the change this method at all, its beyond the scope of this PR

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
```diff
@@ -95,34 +97,34 @@ public static long[][] deepClone(long[][] original) {
...
         }
-        if (sig1.length < sig2.length) {
+        if (b1.length() < b2.length()) {
             return -1;
         }
-        if (sig1.length > sig2.length) {
+        if (b1.length() > b2.length()) {
             return 1;
         }
         return 0;
```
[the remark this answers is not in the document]
lpetrovic05: I tried not the change this method at all, its beyond the scope of this PR. I will open a separate ticket for this

david-bakin-sl:
> Feat as opposed to chore makes sense. As for copying the ticket into the description, I don't see the point in that, I don't like duplicating data/code/text 😄

Up to you of course, but 1) I hope the final merge commit message makes sense for a person just looking at a cloned repo and 2) the advantage would be for the _reader_ of the PR to have it right there.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/ImmutableHash.java
lpetrovic05: I did not change the logic, just call a different method. Once `Hash` is immutable, there is no reason for this class to exist.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/shadowgraph/SyncLogging.java
lpetrovic05: this is provided as a lambda to log4j, so they are executed only when a particular marker is turned on, which is only on for debugging

PR #13132, opened 2024-05-07
build: switch to Dependabot friendly dependency version notation

### LINKED ISSUE #11570: Process for ensuring the Gradle build and application dependencies are updated regularly
## Description

We need ensure existing processes/tools like Github dependabot can identify/recommend/plan/execute dependency updates for both the Gradle build and application dependencies. If the Gradle build cannot be configured to support existing tools like Github dependabot, then we need to build custom processes and/or tools to automate the dependency version management operations.

jjohannes:
**Description**:

Dependabot cannot understand **Module Names**. Thus we have to use the Maven coordinates when we define versions for it to pick them up correctly. Which is something we wanted to avoid initially. As a compromise, we add the **Module Name** of each dependency in the **because** statement. This way, you see both notations to avoid confusion. The updated `java-module-dependencies` plugin **checks if the Module Names are correct** (https://github.com/gradlex-org/java-module-dependencies/issues/105).

**Related issue(s)**:

Fixes #11570

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,12 @@
 version: 2
 updates:
   - package-ecosystem: "gradle"
-    directory: "/"
+    directory: "/hedera-dependency-versions"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "gradle"
+    directory: "/gradle/plugins"
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-benchmarks/build.gradle.kts
```diff
@@ -31,7 +31,7 @@ jmhModuleInfo {
     requires("com.swirlds.fchashmap")
     requires("com.swirlds.merkledb")
     requires("com.swirlds.virtualmap")
-    requires("com.swirlds.platform")
+    requires("com.swirlds.platform.core")
```
jjohannes, line 34: This was wrong, but somehow only detected now after the plugin update.

JeffreyDallas (approved):

jeromy-cannon (approved):

PR #13147, opened 2024-05-08
chore: PlatformContext API cleanup

hendrikebbers:
PlatformContext API cleanup
PlatformContextFactory introduced

Part of https://github.com/hashgraph/hedera-services/issues/13266


cody-littley (changes requested):
I think we should have a high level discussion on the end goal design for the platform context

hendrikebbers:
Meeting scheduled

PR #13155, opened 2024-05-08
fix: (cherry pick) Enhance purechecks for CryptoGetAccountBalanceHandler (#12839) into 50

thomas-swirlds-labs:
**Description**:

Cherry Pick of #12839 from 0.49 into 0.50


netopyr (changes requested):
This PR does not have release/0.50 as its target.

thomas-swirlds-labs:
Good catch @netopyr 👍

Issue #11159, opened 2024-01-25
Executing an internal call with amount against system accounts results in `PRECOMPILE_ERROR`


yiliev0:
### Description

Executing a call function with specifying an amount, directed at a contract that itself contains a call function, leads to a precompile error. This error specifically occurs when the target of the contract call is a system account within the range of 0.0.10 to 0.0.357.
EOA ---call---> Contract(call with amount) ---call[value]---> 0.0.357

This behaviour is observed in version 0.46, but also in previous versions.
https://hashscan.io/previewnet/transaction/1706173001.987222003 

### Steps to reproduce

1. Create an contract with the following function:
```
    function makeCallWithAmount(address _to, bytes memory _data) external payable returns (bool success, bytes memory returnData) {
        (success, returnData) = _to.call{value: msg.value}(_data);
    }
```
2. Deploy the contract and grab the contract address
3. Execute a ContractCall(with amount) against the contract address(step2) with the function name("makeCallWithAmount") with function parameter of the  system account address(0.0.357).

**Expected result**: The transaction should result with `INVALID_FEE_SUBMITTED`
**Actual result**: The transaction fails with `PRECOMPILE_ERROR`

### Hedera network

other

### Version

0.46.0

### Operating system

None

0xivanov:
Is the data parameter empty?

PR #13175, opened 2024-05-09
fix: internal calls to system accounts

### LINKED ISSUE #11158: Executing an internal call without amount against system accounts results in `PRECOMPILE_ERROR`
### Description

Executing a call function without specifying an amount, directed at a contract that itself contains a call function, leads to a precompile error. This error specifically occurs when the target of the contract call is a system account within the range of 0.0.10 to 0.0.357. 
EOA ---call---> Contract(call) ---call---> 0.0.357

This behaviour is observed in version 0.46, but also in previous versions. 
https://hashscan.io/previewnet/transaction/1706171808.147232330

### Steps to reproduce

1. Create an contract with the following function:
```
    function makeCallWithoutAmount(address _to, bytes memory _data) external returns (bool success, bytes memory returnData) {
        (success, returnData) = _to.call(_data);
    }
```
2. Deploy the contract and grab the contract address
3. Execute a ContractCall against the contract address(step2) with the function name("makeCallWithoutAmount") with function parameter of the  system account address(0.0.357)

**Expected result**: The transaction should be successful with no op
**Actual result**: The transaction fails with `PRECOMPILE_ERROR`

### Additional context

_No response_

### Hedera network

other

### Version

0.46.0

### Operating system

None

petreze:
**Description**:
This PR fixes the behaviour whenever we make internal calls to a system account both with and without sending value

**Related issue(s)**:

Fixes #11158, #11159 

**Notes for reviewer**:
The expected behaviour is:
- if we try to send value over the call - this should result in `INVALID_FEE_SUBMITTED`
- if we do not send value - the behaviour should be a no-op success

**Checklist**


netopyr (approved):
LGTM in regards to hedera-services.

jeromy-cannon (approved):

PR #13177, opened 2024-05-09
feat: real keys in fake address books

### LINKED ISSUE #13133: Real keys in RandomAddressBookGenerator
Currently, the keys used for address books generated by `RandomAddressBookGenerator` are not real keys. Useful for tests that need to be fast, not so useful when you want to use the address book to test signing.

Add the ability to generate real keys when generating an address book via `RandomAddressBookGenerator.

cody-littley:
closes #13133


lpetrovic05:
using real cryptography in unit tests can lead to very long tests, not sure if this is the best approach to take

cody-littley:
> using real cryptography in unit tests can lead to very long tests, not sure if this is the best approach to take

This was discussed in our group meeting, and the decision was that it is ok to have real keys generated by this utility given that it is optional and disabled by default.

edward-swirldslabs (approved):
Just reflecting on the work needed to be done when we switch away from address book to consensus roster...  the size of the PR...

PR #13178, opened 2024-05-09
feat(cancun): Finish HIP-866 non-support for Cancun blobs

### LINKED ISSUE #11700: feat(cancun): HIP-866 add EVM compatibility for non-supported Cancun blob features
Support [HIP-866](https://hips.hedera.com/hip/hip-866) - "Add EVM compatibility for non-supported Cancun blob features"

* Depends on `ContractsV_0_49Module` from #11698 
* Confirm (via testing) that the new Ethereum type 3 transactions (presented via `EthereumTransaction`) are rejected by modular (and mono) service
* Override two new EVM opcodes `VERSIONEDHASH` and `BLOBASFEE` to "behave as if there are no blobs and that there never have been blobs"
* Ensure that `versionedHashes` and `blobGasPrice` in a `TxValues` instance are set to safe values
* Provide tests for the above

david-bakin-sl:
**Description**:

Finish HIP-866 non-support for Cancun by testing rejection of type 3 (blob) transactions.

**Related issue(s)**:

Fixes #11700 

**Notes for reviewer**:

This will need to be cherry-picked into the release 0.50 branch.

**Checklist**


hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -78,6 +78,7 @@ public static EthTxData populateEthTxData(byte[] data) {
             return switch (rlpItem.asByte()) {
                 case 1 -> populateEip2390EthTxData(decoder.next(), data);
                 case 2 -> populateEip1559EthTxData(decoder.next(), data);
+                case 3 -> null; // We don't currently support Cancun "blob" transactions
```
lukelee-sl, line 81: There are callers of `populateEthTxData` that do not check for a null return value.  Example `MiscUtils` lines 778-781, `HandleWorkflow` lines 829-831, `ThrottleAccumulator` and other places.  Some of these are for mono service and some in modularized code.  I think the saving grace is a check in `IngestChecker` (at least for modularized) service that checks for a null and does not let the transaction proceed.  It seems like we might have a problem with NullPointerExceptions if a transaction ever escaped this check.  
Would it make sense to do null checks everywhere `populateEthTxData` is called or are we safe because of that check in `IngestChecker`?

david-bakin-sl: Hmm, good investigating, TY! I'll look more closely (as I had _assumed_ the `IngestChecker` was sufficient but didn't go further).

david-bakin-sl: OK, I fixed `MiscUtils`, `ThrottleAccumulator` has both uses guarded, `SignedTxnAccessor` is guarded, `HandleWorkflow` got the same fix as `MiscUtils`, and `IngestChecker` checks for nulls there (and throws).

All other uses are in unit tests where the test will end up failing (one way or another) if it gets a null.

lukelee-sl (approved):
LGTM

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
tinker-michaelj: Do we have any EETs that submit `EthereumTransaction` with invalid bytes? Seems likely, but not sure.

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
```diff
@@ -136,6 +138,7 @@ public HederaEvmTransactionProcessingResult execute(
                 .initialGas(gasAvailable)
                 .originator(senderEvmAddress)
                 .gasPrice(Wei.of(gasPrice))
+                .blobGasPrice((Wei.ONE))
```
tinker-michaelj, line 141: Will we ever use this value?

tinker-michaelj (approved):
LGTM, ty @david-bakin-sl !

hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/execution/HederaEvmTxProcessor.java
david-bakin-sl: It can be used if someone executes the `BLOBBASEFEE` opcode.  There won't be any blob transactions, but anyone can write a contract that uses the opcode.

It's set to `Wei.ZERO` by the `MessageFrame.Builder` constructor but our HIP says it should have the value `ONE`.

PR #13184, opened 2024-05-09
feat: misc cleanup tasks in SwirldsPlatform

### LINKED ISSUE #13182: Trim the fat on SwirldsPlatform.java
There are a bunch of small things that can be safely removed form `SwirldsPlatform.java`, the net effect being a much cleaner file.

cody-littley:
closes #13182


lpetrovic05 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/SwirldsPlatform.java
```diff
@@ -692,7 +660,9 @@ public void start() {
      * </ul>
      */
     public void performPcesRecovery() {
-        thingsToStart.start();
+        platformContext.getFileSystemManager().start();
```
hendrikebbers, line 663: Oh, I just see that. Looks like I totally missed that `FileSystemManager extends Startable, Stoppable`. That is wrong from my point of view since this ends in having start/stop in the public api and allows to stop parts of the PlatformContext within the platform. 

We should get rid of that and provide an api that allows us to start/stop the platform from outside.

hendrikebbers: Created https://github.com/hashgraph/hedera-services/issues/13230

cody-littley: I agree -- all of the base utilities should ideally not require platform input to control their lifecycle. Metrics is another example of something the platform is currently responsible for calling "start" on.

lpetrovic05 (approved):

PR #13206, opened 2024-05-10
docs: update token claim airdrop design doc

MiroslavGatsanoga:
**Description**:

- Following up on comments from https://github.com/hashgraph/hedera-services/pull/12838
- Remove approach for dispatching child CryptoTransfer
- Describe check if the airdrop sender has approved token allowances

**Checklist**


IvanKavaldzhiev (dismissed):
LGTM

hedera-node/docs/design/services/token-service/airdrops/token-claim-airdrop.md
```diff
@@ -120,21 +120,17 @@ An update into the `feeSchedule` file would be needed to specify that.
         - Verify that the pending airdrops list does not have any duplicate entries
     - Pre-handle:
         - The transaction must be signed by the account referenced by a `receiver_id` for each entry in the pending airdrops list
-        - Confirm that for the given pending airdrops ids in the transaction there are corresponding pending transfers existing in state
-        - Check if the sender has sufficient balance to fulfill the airdrop
     - Handle:
+        - Confirm that for the given pending airdrops ids in the transaction there are corresponding pending transfers existing in state
+        - Check if the sender has sufficient balance or has enough approved allowance to fulfill the airdrop
```
netopyr, line 125: I am unsure about this one. We need to double-check.

In my understanding, we charge the sender all fees when they submit the TokenAirdrop transaction. However, when the receiver claims the token, the sender should not be charged anything.

netopyr:
Please add that we also need to update `ApiPermissionConfig` and probably the different `throttles.json`-files to include the new transaction.

hedera-node/docs/design/services/token-service/airdrops/token-claim-airdrop.md
MiroslavGatsanoga: > In my understanding, we charge the sender all fees when they submit the TokenAirdrop transaction. However, when the receiver claims the token, the sender should not be charged anything.

That is my understanding too. I think the wording here can be improved, what I mean by _"balance"_ is the amount of tokens that are being claimed in the senders account.
Would something like _"Check if the sender has sufficient amount or has enough approved allowance of the tokens being claimed to fulfill the airdrop"_ be better?

netopyr: Ah, yes, thanks for the clarification. I misunderstood the original text. The new one is better. Thanks!

IvanKavaldzhiev (approved):
LGTM

netopyr (approved):
LGTM - thanks @MiroslavGatsanoga

PR #13225, opened 2024-05-12
feat(HIP796): Implement HIP-796 moveUnitsBetweenDifferentUserPartitions functionality

vvairava:
Description

Move the units from one partition to a different partition on a different user account, but requiring a signature from the user’s account being debited

Existing Files

Compile
./gradlew assemble is SUCCESS


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/crypto/HapiCryptoTransfer.java
```diff
@@ -483,6 +500,42 @@ protected Consumer<TransactionBody.Builder> opBodyDef(final HapiSpec spec) throw
...
+                .addTransfers(
+                        AccountAmount.newBuilder().setAccountID(fromAccount).setAmount(-amount));
+        b.addTokenTransfers(reduceUnits);
+        final var appendUnits = TokenTransferList.newBuilder()
+                .setToken(toPartitionToken)
+                .addTransfers(AccountAmount.newBuilder().setAccountID(toAccount).setAmount(+amount));
+        b.addTokenTransfers(appendUnits);
+    }
+
+    private void PartitionMoveWithoutUserSignature(final CryptoTransferTransactionBody.Builder b, final HapiSpec spec) {
```
thomas-swirlds-labs, line 522: Other than one line, this code is identical to `PartitionMoveWithUserSignature`. Could consider refactoring common code into a single method to avoid duplication.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/crypto/HapiCryptoTransfer.java
```diff
@@ -483,6 +500,31 @@ protected Consumer<TransactionBody.Builder> opBodyDef(final HapiSpec spec) throw
...
     }
 
+    private void FungibleTokenPartitionMove(final CryptoTransferTransactionBody.Builder b, final HapiSpec spec) {
+        final var extra = (this.PARTITION_TOKEN_OPERATION_MODE == "PartitionMoveWithoutUserSignature")
+                ? appendedPartitionFromTo.get()
+                : appendedPartitionFromTowithSign.get();
+        final var involved = extra.getLeft();
+        final var fromAccount = TxnUtils.asId(involved[0], spec);
+        final var fromPartitionToken = TxnUtils.asTokenId(involved[1], spec);
+        final var toAccount = (this.PARTITION_TOKEN_OPERATION_MODE != "PartitionMoveWithoutUserSignature")
```
thomas-swirlds-labs, line 510: It might be clearer to modify this line like so 

`final var toAccount = (this.PARTITION_TOKEN_OPERATION_MODE = "PartitionMoveWithUserSignature")`

To avoid the confusion of double negatives. Otherwise LGTM 👍

vvairava: @thomas-swirlds-labs, I have pushed the changes.. Can you please check it. Thanks

thomas-swirlds-labs: LGTM @vvairava. Thanks 👍

thomas-swirlds-labs (approved):
I think we need to address the naming conventions and do a little refactoring here in a later PR to avoid confusion and overloading around PartitionMoveWith/WithoutSignatures

PR #13231, opened 2024-05-13
chore: use `Bytes` for event signatures

### LINKED ISSUE #13139: Use Bytes for event signatures
In an effort to move events to protobuf, signatures should be used as `Bytes` instances instead of `byte[]`

lpetrovic05:
closes #13139


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Signature.java
```diff
@@ -78,6 +79,13 @@ public byte[] getSignatureBytes() {
         return signatureBytes;
     }
 
+    /**
+     * @return the bytes of this signature in an immutable instance
+     */
+    public Bytes getBytes() {
```
alittley, line 85: missing method nullity annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
```diff
@@ -144,7 +155,7 @@ public BaseEventHashedData getHashedData() {
     /**
      * @return the signature for the event
      */
-    public byte[] getSignature() {
+    public Bytes getSignature() {
```
alittley, line 158: missing nullity annotation

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Signature.java
lpetrovic05: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
lpetrovic05: added

mxtartaglia-sl (dismissed):

alittley (dismissed):

anthony-swirldslabs (dismissed):

jeromy-cannon (dismissed):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
```diff
@@ -133,45 +135,43 @@ public static int arrayCompare(byte[] sig1, byte[] sig2) {
...
+        if (a1 != null && a2 == null) {
             return 1;
         }
-        if (sig1 == null && sig2 != null) {
+        if (a1 == null && a2 != null) {
             return -1;
         }
-        maxLen = Math.max(sig1.length, sig2.length);
-        minLen = Math.min(sig1.length, sig2.length);
+        final int maxLen = (int) Math.max(a1.length(), a2.length());
```
hendrikebbers, line 156: I just ask myself if an if might be better for performance & memory since we do not need to downcast long to int. Never proofed it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/ConsensusUtils.java
```diff
@@ -72,9 +74,10 @@ public static boolean coin(@NonNull final EventImpl event) {
         // find whitening for round
         for (final EventImpl w : judges) { // calculate the whitening byte array
             if (w != null) {
-                final int mn = Math.min(whitening.length, w.getSignature().length);
+                final Bytes sig = w.getBaseEvent().getSignature();
+                final int mn = Math.min(whitening.length, (int) sig.length());
```
hendrikebbers, line 78: Do we know why the protobuff model uses long for length? If it is always int maybe we should change protobuff.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -254,22 +255,22 @@ static KeyStore createEmptyTrustStore() throws KeyStoreException {
...
             final Signature sig = Signature.getInstance(CryptoConstants.SIG_TYPE2, CryptoConstants.SIG_PROVIDER);
             sig.initVerify(publicKey);
-            sig.update(data);
-            return sig.verify(signature);
-        } catch (NoSuchAlgorithmException | NoSuchProviderException e) {
+            data.updateSignature(sig);
+            return signature.verifySignature(sig);
+        } catch (final NoSuchAlgorithmException | NoSuchProviderException e) {
             // should never happen
             throw new CryptographyException(e);
```
hendrikebbers, line 272: Can you add a message to that exception.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
```diff
@@ -254,22 +255,22 @@ static KeyStore createEmptyTrustStore() throws KeyStoreException {
...
-        } catch (InvalidKeyException | SignatureException e) {
+        } catch (final InvalidKeyException | SignatureException e) {
             logger.error(LogMarker.EXCEPTION.getMarker(), "", e);
```
hendrikebbers, line 274: Can you add a message

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
lpetrovic05: I think that would be considered premature optimization

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/ConsensusUtils.java
lpetrovic05: I have no idea. Maybe @anthony-swirldslabs knows

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
lpetrovic05: added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/CryptoStatic.java
lpetrovic05: added

mxtartaglia-sl (dismissed):

alittley (approved):

jeromy-cannon (approved):

alittley (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/ConsensusUtils.java
anthony-swirldslabs: There's not a specific answer.

From theoretical perspective, the protobuf as a protocol doesn't really need to have a limitation on the sizes/lengths. And hence, choosing long as the "longest" integer value for sizes/lengths makes sense.

From practical perspective, the spec seems to impose a hard limit of 2GB on protobuf messages, meaning that it's technically impossible to encode repeated or size-limited fields longer than 2G, which means that a 32-bit integer should suffice for size values.

From yet another perspective, some PBJ APIs were inspired by stream-related APIs where there's not really a limit on the length of the stream. So similar to the point 1 above, using a long makes sense.

I think that the points 1 and 3 played a role historically, and so we ended up with longs.

You may open a feature request against PBJ to try and switch to ints, but I'm unsure about its prioritization, or if there's any other legitimate reasons to keep using longs.

anthony-swirldslabs (dismissed):

jeromy-cannon (approved):

alittley (dismissed):

mxtartaglia-sl (approved):

edward-swirldslabs (approved):

jeromy-cannon (approved):

alittley (approved):

anthony-swirldslabs (approved):

mxtartaglia-sl (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/consensus/ConsensusUtils.java
hendrikebbers: Adding @jasperpotts for discussion. I seen the casting to `int` at several places.

PR #13256, opened 2024-05-13
chore: FileAppendHandler unit test coverage

### LINKED ISSUE #13188: FileAppendHandler [58 %]

kimbor:
**Description**:
Improve unit test coverage for FileAppendHandler class

**Related issue(s)**:

Fixes #13188

### DIFF
--- a/hedera-node/hedera-file-service-impl/src/test/java/com/hedera/node/app/service/file/impl/test/handlers/FileAppendHandlerTest.java
+++ b/hedera-node/hedera-file-service-impl/src/test/java/com/hedera/node/app/service/file/impl/test/handlers/FileAppendHandlerTest.java
@@ -18,12 +18,17 @@ package com.hedera.node.app.service.file.impl.test.handlers;
 
 import static com.hedera.hapi.node.base.ResponseCodeEnum.FILE_DELETED;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_FILE_ID;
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-file-service-impl/src/test/java/com/hedera/node/app/service/file/impl/test/handlers/FileAppendHandlerTest.java
```diff
@@ -143,7 +168,7 @@ void preHandleWorksAsExpected() throws PreCheckException {
 
         subject.preHandle(realPreContext);
 
-        assertTrue(realPreContext.requiredNonPayerKeys().size() > 0);
+        assertFalse(realPreContext.requiredNonPayerKeys().isEmpty());
```
mhess-swl, line 171: Since we're changing tests anyway it might be good to replace junit with assertj on the lines that are changed or added. Just a thought though

mhess-swl (approved):

hedera-node/hedera-file-service-impl/src/test/java/com/hedera/node/app/service/file/impl/test/handlers/FileAppendHandlerTest.java
povolev15: @mhess-swl I would ask during the call because it is all over the place

povolev15 (approved):

PR #13281, opened 2024-05-14
fix: stabilize exchange rate system contract spec

lukelee-sl:
**Description**:
A recent pr (#13053) made a modification to the ExchangeRateSystemContract to return INVALID_FEE_SUBMITTED response code if value is sent to the system contract. 

A new related test was added at that time but it tested for a more generic CONTRACT_REVERT_EXECUTED vs the more specific response code.

Unknown why the ci check failed in the above mentioned pr.


david-bakin-sl (approved):
LGTM

PR #13291, opened 2024-05-14
fix: Use mainnet `throttles.json` for testnet also

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/13296
Use mainnet `throttles.json` for testnet also


tinker-michaelj (approved):

Nana-EC (approved):

Issue #13296, opened 2024-05-14
Update testnet throttles in repo to be same as mainnet throttles 


Neeharika-Sompalli:
Update the `hedera-node/configuration/testnet/upgrade/throttles.json` to be same as `hedera-node/configuration/mainnet/upgrade/throttles.json`

PRs should be open to 

- [x] release/0.49
- [x] release/0.50
- [x] develop

PR #13298, opened 2024-05-14
fix: use mainnet `throttles.json` for testnet

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/13296
Use mainnet throttles.json for testnet also

PR to release/0.49 - https://github.com/hashgraph/hedera-services/pull/13291


tinker-michaelj (approved):

Nana-EC (approved):

Issue #13172, opened 2024-05-09
Repurpose BaseEventHashedData


lpetrovic05:
Initially, the idea was to remove `BaseEventHashedData` since it is part of the legacy event structure. But event creator does not sign events, and it cannot output a fully formed event. This means that its output must be all of the event parts except for the signature. This is exactly what `BaseEventHashedData` contains. So a better approach would be to repurpose it.

### Tasks
- Remove usages of `BaseEventHashedData` as part of gossip event (or any other event types)
- Remove reliance on the serialization and hashing of `BaseEventHashedData`
- Rename `BaseEventHashedData` to `UnsignedEvent`, which will be the output of the event creator

PR #13343, opened 2024-05-16
feat: remove `BaseEventHashedData` from `DetailedConsensusEvent`

lpetrovic05:
part of #13172

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
@@ -175,7 +175,7 @@ public class GossipEvent implements Event, SelfSerializable {
 
     @Override
     public Iterator<Transaction> transactionIterator() {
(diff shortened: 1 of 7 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
```diff
@@ -168,6 +173,28 @@
         return hashedData.getDescriptor();
     }
 
+    @Override
+    public Iterator<Transaction> transactionIterator() {
+        return Arrays.asList((Transaction[]) hashedData.getTransactions()).iterator();
```
alittley, line 178: Do you need to check if `hashedData.getTransactions()` returns `null`?

lpetrovic05: In practice, transaction should never be null, although the legacy code pathways allow it to happen in theory. This theoretical pathway should be closed soon.

alittley (approved):

Neeharika-Sompalli (approved):
LGTM from `hedera-base` changes

PR #13345, opened 2024-05-16
chore: Switch to Transaction.getApplicationPayload()

### LINKED ISSUE #12885: Hedera handle protobufs
As part of the block stream project, platform has now started exposing payloads (formerly transactions) as PBJ records.
Hedera should stop relying on `Transaction.getContents` and start consuming PBJ payloads from `Transaction.getPayload` or the convenience method `Transaction.getApplicationPayload`.

Once this transition is complete, `Transaction.getContents` should be removed.

In order to avoid extra effort, this change could be done only in the modular code if the expectation is that the old code will be deprecated in the near term.

netopyr:
**Description**:

Switch to `Transaction.getApplicationPayload()` instead of the deprecated method `Transaction.getContents()`

**Related issue(s)**:

Fixes #12885

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
@@ -610,7 +610,7 @@ public class HandleWorkflow {
         } catch (final Exception e) {
             logger.error("Possibly CATASTROPHIC failure while handling a user transaction", e);
             if (transactionInfo == null) {
-                final var baseData = extractTransactionBaseData(platformTxn.getContents());
+                final var baseData = extractTransactionBaseData(platformTxn.getApplicationPayload());
                 if (baseData.transaction() == null) {
                     // FUTURE: Charge node generic penalty, set values in record builder, and remove log statement
                     logger.error("Failed to parse transaction from creator: {}", creator);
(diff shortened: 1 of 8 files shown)


povolev15 (approved):
LGTM TY @netopyr

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -1034,15 +1034,15 @@ private record TransactionBaseData(
...
         // This method is only called if something fatal happened. We do a best effort approach to extract the
         // type of the transaction, the TransactionBody and the payer if not known.
-        if (contents == null || contents.length == 0) {
+        if (content.length() == 0) {
             return new TransactionBaseData(NONE, Bytes.EMPTY, null, null, null);
         }
 
         HederaFunctionality function = NONE;
-        Bytes transactionBytes = Bytes.wrap(contents);
+        Bytes transactionBytes = content;
```
Neeharika-Sompalli, line 1045: nit : We don't need this `Bytes transactionBytes` anymore right ?

netopyr: We do. It is non-final and may be reassigned in the code below. Technically, we could make the input parameter non-final and reuse it. But this would be so unusual that people will easily overlook it. Therefore, I think introducing this local variable is the better choice.

PR #13377, opened 2024-05-17
fix: 13335 Fixed incorrect classId in `OnDiskKeySerializer` and `OnDiskValueSerializer` classes

imalygin:
**Description**:

This PR fixes incorrect use of `classId`-s in  `OnDiskKeySerializer` and  `OnDiskValueSerializer`. Unfortunately, existing unit tests didn't catch that so I updated the test as well

**Related issue(s)**:

Fixes #13335

PR #13380, opened 2024-05-19
fix: Enable `@HapiTest` concurrency

### LINKED ISSUE #13310: Support concurrent `@HapiTest` execution
### Problem

Some `@HapiTest`s temporarily change the network configuration (e.g., change the fees, exchange rate, properties, permissions, or throttles).

This means they cannot run concurrently with other `@HapiTest`s, because those other tests' expectations may be sensitive to changes in network configuration.

For example, if test `A` temporarily overrides `staking.fees.stakingRewardPercentage=50` then another test `B` trying to validate that `0.0.800` receives the default 10% of fees will fail if run concurrently.

Call tests like `A` "**leaky**". Even though we must run leaky tests sequentially, we would like to run all other `@HapiTest`s concurrently.

### Solution

1️⃣ Migrate from a custom `TestEngine` to the JUnit Jupiter engine to enjoy declarative synchronization control via the [`@Execution(CONCURRENT)`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/parallel/ExecutionMode.html) and [`@ResourceLock({READ | READ_WRITE})`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/parallel/ResourceLock.html) annotations.
 - [x] Move initialization of the long-lived (shared) subprocess network into a `LauncherSessionListener`.
 - [x] Add a `NetworkTargetingExtension` implementing the lifecycle callbacks to:
    1. Set thread-local state about what network should be targeted by a `HapiSpec` dynamically registered in the thread---either the shared network or an `@Isolated` network.
    2. Create and teardown `@Isolated` networks as needed.
 - [x] Make `@HapiTest` and `@LeakyHapiTest` JUnit5 meta-annotations with `@TestFactory` and appropriate `@ResourceLock` variants. 

2️⃣ Update all methods returning a `HapiSpec` to instead return JUnit5 dynamic tests.
 - [x] Make all `HapiSpec` factories return a `Stream<DynamicTest>` that initially includes a single `DynamicTest` that executes the `HapiSpec` and succeeds iff the spec does.
     * These streams will later include variants of the `HapiSpec` such as replacing all `ContractCall`s and `ContractCreate`s with `EthereumTransaction`s.
- [x] Update `HapiSpec` factories to automatically target any network bound to the thread by the `NetworkTargetingExtension`.

3️⃣ Change our Gradle `Test` tasks to be vanilla `useJUnitPlatform()` tasks.
 - [x] Replace custom `@Tag` processing with the standard `includeTags()` DSL.

4‍⃣ Eliminate unneeded use of `extends HapiSuite`
 - [x] Remove this for all `@HapiTestSuite`-annotated classes, remove `@HapiTestSuite` annotation.

### Alternatives

_No response_

tinker-michaelj:
Closes #13310.

(_Note to reviewers_: also includes https://github.com/hashgraph/hedera-services/pull/13377 to fix an ISS that would otherwise break an improved ND Reconnect check)

**Description**:

1️⃣  Replaces `HapiTestEngine` with the JUnit Jupiter `TestEngine`.
  * Launches the shared test network before `TestPlan` execution with a [simple `LauncherSessionListener`](https://github.com/hashgraph/hedera-services/blob/13310-hapi-test-concurrency/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/SharedNetworkLauncherSessionListener.java#L31).
    - **Note:** Because Gradle's `JUnitPlatformTestExecutionListener` [can't handle](https://github.com/gradle/gradle/issues/6453) concurrent test class execution(❗), this PR doesn't immediately add support for running against multiple test networks simultaneously; as the main use case would be...running multiple test classes concurrently. 
 * Splices the `HapiSpec` and JUnit Jupiter lifecycles via a few tiny extensions [here](https://github.com/hashgraph/hedera-services/tree/13310-hapi-test-concurrency/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/extensions); a tweak to make `HapiSpec` implement `Executable`; and its factories return an implied `Stream<DynamicTest>` [here](https://github.com/hashgraph/hedera-services/blob/13310-hapi-test-concurrency/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpec.java#L1062).
  * Eliminates custom Gradle `Test` tasks in favor of the default task; uses JUnit Jupiter [tag expressions](https://github.com/hashgraph/hedera-services/blob/13310-hapi-test-concurrency/hedera-node/test-clients/build.gradle.kts#L66) to target different CI checks.
     - Updates GitHub Action workflows with the appropriate Gradle invocations.
     - Makes concurrent execution the default _within_ test classes (while still executing test classes sequentially due to the Gradle bug as above).

2️⃣ Refactors every test class running in CI.
  1. Eliminates all use of `extends HapiSuite` and the implied boilerplate.
     - `HapiSuite` still exists in the code base only because it used by yahcli and JRS.
  2. Ensures tests that cannot run concurrently are annotated with `@ResourceLock` via a custom meta-annotation `@LeakyHapiTest`.
  3. Re-writes some classes to demonstrate newly unlocked use of the JUnit Jupiter programming model; see [`SelfDestructSuite`](https://github.com/hashgraph/hedera-services/blob/13310-hapi-test-concurrency/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/opcodes/SelfDestructSuite.java#L68) for an example using `@Nested` classes, `@BeforeAll`/`@AfterAll` lifecycle methods, and declarative synchronization control using `@ResourceLock`.

3️⃣  Fixes some issues surfaced while stabilizing concurrent tests.
  1. In `FrameRunner`, when propagating a halt in `CustomMessageProcessor` to the calling frame (for backward compatibility), adds the appropriate call to `tracePostExecution()`.
      * Since `ContractCallSuite` ran sequentially, [`hscsEvm010ReceiverMustSignContractTx()`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractCallSuite.java#L2040) was not exposed to the sidecar validator.
  4. In `TokenUpdateHandler`, only updates to a blank token memo **if** the update is coming via HAPI, since a `tokenUpdate()` system contract ABI cannot encode a difference between leaving the memo untouched and making it blank.
      * [`TokenUpdatePrecompileSuite`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java) was not annotated with `@HapiTestSuite`, hence was not running in CI.
  5. Makes `UtilPrngHandler.handle()` a no-op if `utilPrng.isEnabled=false` for backward compatibility.
      * `FeatureFlagSuite` did not have a `.failOnErrors()` modifier on its parallel operations.
  6. Removes pointless `FeeManager` logging when `GET_ACCOUNT_DETAILS` is received. 
  7. Removes `WARN` log when a requested stake node id is missing from address book.
  8. Rationalizes `HevmTransactionFactory` to decide whether to allow calls to missing contracts based on the in-use `FeatureFlags`, since as written its behavior can be inconsistent with the in-use `TransactionProcessor`. (E.g. `v0.38` EVM is in use but `contracts.evm.allowCallsToNonContractAccounts=true`.)

4️⃣ Deletes many, many lines of dead code; both obsolete `HapiSuite`'s and `HapiSpec` rot.


nathanklick (changes requested):
@jjohannes Please review, this reverts many of the changes you had put in place. 

@tinker-michaelj I am not a fan of removing the test definitions from Gradle and instead making them tag driven only via CI. @jjohannes had previously codified the test suites as part of the Gradle build to make it easy to run locally of via CI and I am not sure we want to undo those changes.

tinker-michaelj:
> @jjohannes Please review, this reverts many of the changes you had put in place.
> 
> @tinker-michaelj I am not a fan of removing the test definitions from Gradle and instead making them tag driven only via CI. @jjohannes had previously codified the test suites as part of the Gradle build to make it easy to run locally or via CI and I am not sure we want to undo those changes.

Got it Nathan, I will ask around and see if people found those useful; I never used them and felt it was a bit distracting for day-to-day work. 🤔

nathanklick:
> Got it Nathan, I will ask around and see if people found those useful; I never used them and felt it was a bit distracting for day-to-day work. 🤔
> 
> It's a minor point for sure, but reasoning was:
> 
> * I never run suites by `@Tag` locally, so having a shortcut for that isn't helpful.
> * Having so many registered `Test` tasks requires an extra click in IDE each time you want to run an individual test class or method.
> * It's _possible_ to misclick in IDE and choose a `Test` task that doesn't include the class or method's `@Tag`; then you have to manually edit the run configuration.
> 
> So...veryy minor points. But I consolidated to the default task b/c I felt it was a small win.

@tinker-michaelj @jjohannes My biggest concern with this change is the fact we have now forced CI to be aware of the internal test structures and tags to a highly granular level. I felt the use of clearly defined tasks in Gradle provides a much better public API for CI to depend on and thereby relieving CI of the responsibility of understanding/managing test tags and the test internals.

Neeharika-Sompalli (dismissed):
As you described in the call. Everything makes sense. It is extremely useful to run tests concurrently and improve CI job times. 

tyvm @tinker-michaelj for all the effort you put into this ! 🙌

nathanklick (approved):
@tinker-michaelj Thanks for making the changes and sorting this out to use the custom tasks. LGTM at this point.

Neeharika-Sompalli (approved):
LGTM !

cody-littley (approved):

Neeharika-Sompalli (dismissed):
LGTM ! tyvm @tinker-michaelj

nathanklick (dismissed):

lukelee-sl (dismissed):
LGTM.  tyvm.

Neeharika-Sompalli (approved):

lukelee-sl (approved):
LGTM

cody-littley (approved):

nathanklick (approved):

PR #13378, opened 2024-05-17
Fix records for dissociating tokens

Neeharika-Sompalli:
WIP

PR #13407, opened 2024-05-20
fix: (Cherry-pick) Fix records for dissociating deleted tokens

### LINKED ISSUE #13379: Fix the error while finalizing record for dissociation of deleted token
Dissociation of a token when its deleted is allowed, even if the balance is no zero. 
While dissociating deleted non-fungible token, the `FinalizeParentRecordHandler` fails with error. 

Solution:
Iterate through the tokenRelation entries for Non Fungible Tokens also in the `FinalizeChildrecordHandler` , so we will be able to deduct all child changes from the parent record successfully. 
[This step](https://github.com/hashgraph/hedera-services/blob/block-streams/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeParentRecordHandler.java#L122) is missing in ChildRecordFinalizer. 

Remove the duplicated code.

PR should go into: 

- [ ] Release/0.49 - https://github.com/hashgraph/hedera-services/pull/13407
- [ ] Release/0.50 - https://github.com/hashgraph/hedera-services/pull/13408
- [ ] Develop - https://github.com/hashgraph/hedera-services/pull/13378

Neeharika-Sompalli:
Fixes #13379 
Cherry-pick https://github.com/hashgraph/hedera-services/pull/13378
Cherry-pick https://github.com/hashgraph/hedera-services/pull/13416

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/RecordFinalizerBase.java
@@ -255,7 +255,7 @@ public class RecordFinalizerBase {
                         originalToken.treasuryAccountId(),
                         modifiedToken.treasuryAccountId(),
                         nftChanges,
-                        null);
+                        tokenRelChanges);
             }
         }
         return nftChanges;
@@ -296,6 +296,7 @@ public class RecordFinalizerBase {
             @Nullable final Map<EntityIDPair, Long> tokenRelChanges) {
         final var isMint = senderAccountId.accountNum() == 0;
         final var isWipeOrBurn = receiverAccountId.accountNum() == 0;
+        final var isTreasuryChange = nftId.serialNumber() == -1;
         final var nftTransfer = NftTransfer.newBuilder()
                 .serialNumber(nftId.serialNumber())
                 .senderAccountID(senderAccountId)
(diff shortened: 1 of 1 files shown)


tinker-michaelj (approved):

mhess-swl:
Because there are a number of changes made in `RecordFinalizerBase`–invoked for essentially every transaction–it makes me nervous to put these changes in 0.49 without re-running a lot of diff, performance, and longevity tests. It seems less risky to ship this known bug in 0.49; even though we ship a bug, we don't lose the confidence we have in the tests that we've run up to this point. Conversely, if we do fix this bug but don't re-run all the tests (which there isn't time to do), it's possible we might introduce a much worse bug in the record handling.

Neeharika-Sompalli:
> Because there are a number of changes made in `RecordFinalizerBase`–invoked for essentially every transaction–it makes me nervous to put these changes in 0.49 without re-running a lot of diff, performance, and longevity tests. It seems much less risky to ship this known bug in 0.49; even though we ship a bug, we don't lose the confidence we have in the tests that we've run up to this point. Conversely, if we do fix this bug but don't re-run all the tests (which there isn't time to do), it's possible we might introduce a much worse bug in the record handling.



> Because there are a number of changes made in `RecordFinalizerBase`–invoked for essentially every transaction–it makes me nervous to put these changes in 0.49 without re-running a lot of diff, performance, and longevity tests. It seems much less risky to ship this known bug in 0.49; even though we ship a bug, we don't lose the confidence we have in the tests that we've run up to this point. Conversely, if we do fix this bug but don't re-run all the tests (which there isn't time to do), it's possible we might introduce a much worse bug in the record handling.

@mhess-swl yes, we are not planning to get a new tag with this fix.

PR #13416, opened 2024-05-20
fix: Fix record for treasury update

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/13379

- Adds a missing check for record finalizer


tinker-michaelj (approved):
🙏

PR #13417, opened 2024-05-20
fix: (cherry-pick) Fix record for treasury update

Neeharika-Sompalli:
Cherry-pick : https://github.com/hashgraph/hedera-services/pull/13416
Fixes https://github.com/hashgraph/hedera-services/issues/13379

Adds a missing check for record finalizer


tinker-michaelj (approved):
Ah great catch, tyvm @Neeharika-Sompalli !

PR #13421, opened 2024-05-20
fix: allow default `ContractID`s

tinker-michaelj:
**Description**:
 - Permits use of `ContractID.DEFAULT` as an implied no-op call to the zero address.


lukelee-sl (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/utils/ConversionUtilsTest.java
```diff
@@ -81,6 +82,11 @@ void outOfRangeBiValuesAreZero() {
                 0L, asExactLongValueOrZero(BigInteger.valueOf(Long.MIN_VALUE).subtract(BigInteger.ONE)));
     }
 
+    @Test
+    void besuAddressIsZeroForDefaultContractId() {
+        assertEquals(Address.ZERO, contractIDToBesuAddress(ContractID.DEFAULT));
```
david-bakin-sl, line 87: Always a good idea to test external assumptions ...

david-bakin-sl (approved):
LGTM

PR #13429, opened 2024-05-21
feat: HIP-904: adapt auto creation logic unlimited associations

### LINKED ISSUE #13374: HIP-904: Adapt auto/lazy creation logic to support the unlimited maxAutoAssociation and set correct initial values

kselveliev:
**Description**:
This pr adapts auto creation logic so that now if unlimitedAutoAssociationsEnabled is true we create the account 
with maxAssociations value of -1 ( -1 == INFINITY).

This PR modifies:
`AutoAccountCreator` added if statement to take into account the new flag and determine the maxAssocation value of the account
`BaseTokenHandler` now takes into account if the account maxAssociation is -1, does not throw NO_REMAINING_AUTOMATIC_ASSOCIATIONS
`AssociateTokenRecipientsStep` now takes into account if the account maxAssociation is -1, does not throw NO_REMAINING_AUTOMATIC_ASSOCIATIONS

Modified `AutoAccountCreationSuite`, since the new flag is enabled by default modified all current tests in it to override the flag to false so that they behave as before, added 2 new happy path tests.

**Related issue(s)**:

Fixes #13374 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -168,10 +168,12 @@ private TokenAssociation validateAndBuildAutoAssociation(
         final var tokenRel = tokenRelStore.get(accountId, tokenId);
         final var config = handleContext.configuration();
 
-        if (tokenRel == null && account.maxAutoAssociations() > 0) {
-            validateFalse(
-                    account.usedAutoAssociations() >= account.maxAutoAssociations(),
-                    NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
+        if (tokenRel == null && (account.maxAutoAssociations() > 0 || account.maxAutoAssociations() == -1)) {
```
kselveliev, line 171: This could probably be simplified

kselveliev:
From the auto creation docs we have a description 'In the case for which the auto-/ lazy-create is from a transfer of fungible token or an NFT, the fees should include the auto association fee and the first auto-renewal period rent for the token association, and should be paid by the sender.' Currently this pr is not modifying any fee logic so need to figure out if this is in the scope of this pr or not

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -168,10 +168,12 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
+            if (account.maxAutoAssociations() != -1) {
+                validateFalse(
+                        account.usedAutoAssociations() >= account.maxAutoAssociations(),
+                        NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
+            }
```
jsync-swirlds, line 176: This could be much simpler, and also has a subtle bug if `account.maxAutoAssociations() == 0`:
```suggestion
        if (tokenRel == null) {
            if (account.maxAutoAssociations() != -1) {
                validateFalse(
                        account.usedAutoAssociations() >= account.maxAutoAssociations(),
                        NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
            }
```
The `validateFalse` check handles all the other cases.

stoyanov-st (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/AutoAccountCreationUnlimitedAssociationsSuite.java
```diff
@@ -0,0 +1,513 @@
...
+                                .tokenType(TokenType.NON_FUNGIBLE_UNIQUE)
+                                .adminKey(MULTI_KEY)
+                                .supplyKey(MULTI_KEY)
+                                .supplyType(TokenSupplyType.INFINITE)
+                                .initialSupply(0)
+                                .treasury(TOKEN_TREASURY)
+                                .via(NFT_CREATE),
+                        mintToken(
+                                NFT_INFINITE_SUPPLY_TOKEN,
+                                List.of(ByteString.copyFromUtf8("a"), ByteString.copyFromUtf8("b"))),
```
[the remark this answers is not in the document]
kselveliev: Hmm this actually won't work. For nft the initial supply should always be 0 at the beginning and start minting from there. I just double checked this with the hedera docs.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -334,11 +334,12 @@ protected TokenRelation autoAssociate(
                 entitiesConfig.limitTokenAssociations() && numAssociations >= tokensConfig.maxPerAccount(),
                 TOKENS_PER_ACCOUNT_LIMIT_EXCEEDED);
 
-        // TODO: check for unlimited associations flag and include -1 as infinite associations
         final var maxAutoAssociations = account.maxAutoAssociations();
         final var usedAutoAssociations = account.usedAutoAssociations();
-        validateFalse(usedAutoAssociations >= maxAutoAssociations, NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
 
+        if (!entitiesConfig.unlimitedAutoAssociationsEnabled() || account.maxAutoAssociations() != -1) {
```
lukelee-sl, line 340: I understand what this if statement is trying to do but if you can leave a comment to clarify your intent, that would be helpful.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -168,10 +168,12 @@ private TokenAssociation validateAndBuildAutoAssociation(
         final var tokenRel = tokenRelStore.get(accountId, tokenId);
         final var config = handleContext.configuration();
 
-        if (tokenRel == null && account.maxAutoAssociations() > 0) {
-            validateFalse(
-                    account.usedAutoAssociations() >= account.maxAutoAssociations(),
-                    NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
+        if (tokenRel == null && (account.maxAutoAssociations() > 0 || account.maxAutoAssociations() == -1)) {
+            if (account.maxAutoAssociations() != -1) {
```
lukelee-sl, line 172: do you not care about `entitiesConfig.unlimitedAutoAssociationsEnabled()` here?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
jsync-swirlds: Even better, as this same check may be needed in some other places, create a method to call here:
```java
    // used in some handler steps, also, needs javadoc...
    protected hasUnlimitedAutoAssociations(@NonNull final Account account) {
        return !(entitiesConfig.unlimitedAutoAssociationsEnabled() && account.maxAutoAssociations() = -1);
    }
```
Then here we would have
```suggestion
        if (hasUnlimitedAutoAssociations(account)) {
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
jsync-swirlds: This collapses to the following if you simplify with the understanding that the value cannot be `< -1`:
```suggestion
        if (tokenRel == null && (account.maxAutoAssociations() != 0)) {
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
jsync-swirlds: This might be a good use for the method I suggested [in another comment](https://github.com/hashgraph/hedera-services/pull/13429/files#r1622973066)

anastasiya-kovaliova (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
kselveliev: I added a code comment and also applied Joseph's suggestion for creating a method.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
kselveliev: Yes It seems that here we also care about the feature flag. I'm now using the method call suggested by Joseph.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
kselveliev: I applied the above suggestion.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AutoAccountCreator.java
```diff
@@ -82,12 +83,14 @@ public AccountID create(@NonNull final Bytes alias, int maxAutoAssociations) {
         String memo;
 
         final var isAliasEVMAddress = EntityIdUtils.isOfEvmAddressSize(alias);
+        final var entitiesConfig = handleContext.configuration().getConfigData(EntitiesConfig.class);
```
netopyr, line 86: I think the method parameter `maxAutoAssociations` needs to be renamed and the Javadoc updated. With the changes, the parameter is merely a suggestion that may not be observed. Maybe `requiredAutoAssociations` or something similar.

kselveliev: Are you suggesting that we should rename the second parameter of 
`public AccountID create(@NonNull final Bytes alias, int maxAutoAssociations)` in AutoAccountCreator
because I got confused of the highlighted code above.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AutoAccountCreator.java
```diff
@@ -82,12 +83,14 @@ public AccountID create(@NonNull final Bytes alias, int maxAutoAssociations) {
...
+        final int autoAssociations = entitiesConfig.unlimitedAutoAssociationsEnabled() ? -1 : maxAutoAssociations;
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/BaseTokenHandler.java
```diff
@@ -463,4 +465,18 @@ public static TokenAssociation asTokenAssociation(final TokenID tokenId, final A
...
+     *
+     * @param account         the account to check, must not be null
+     * @param entitiesConfig  the configuration settings to check against, must not be null
+     * @return                {@code true} if unlimited auto-associations are enabled and the account's
+     * max auto-associations is set to -1 otherwise {@code false}
+     * @throws NullPointerException if the account or entitiesConfig is null
+     */
+    public static boolean hasUnlimitedAutoAssociations(
+            @NonNull final Account account, @NonNull EntitiesConfig entitiesConfig) {
+        return entitiesConfig.unlimitedAutoAssociationsEnabled() && account.maxAutoAssociations() == -1;
```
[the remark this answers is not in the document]
kselveliev: Comment addressed

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -167,11 +168,14 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
         final var tokenRel = tokenRelStore.get(accountId, tokenId);
         final var config = handleContext.configuration();
+        final var entitiesConfig = config.getConfigData(EntitiesConfig.class);
 
-        if (tokenRel == null && account.maxAutoAssociations() > 0) {
-            validateFalse(
-                    account.usedAutoAssociations() >= account.maxAutoAssociations(),
-                    NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
+        if (tokenRel == null && account.maxAutoAssociations() != 0) {
+            if (!hasUnlimitedAutoAssociations(account, entitiesConfig)) {
```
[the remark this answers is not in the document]
kselveliev: In the above if statements:
First if: We want to have maxAutoAssociations == -1 or maxAutoAssociations > 0
Nested if: We want to skip NO_REMAINING_AUTOMATIC_ASSOCIATIONS validation if we are have unlimited associations enabled but we want to not skip it when maxAutoAssociations > 0. 

These if statements differ in logic so I'm not sure if we will be able to combine them into one or maybe
I have misunderstood the suggestion.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AutoAccountCreator.java
netopyr: Yes, sorry. GitHub did not allow me to put the comment where it belonged.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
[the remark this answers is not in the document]
netopyr: Oh, I see. This is confusing.

What I meant are the conditions in this code section:
```
 if (!hasUnlimitedAutoAssociations(account, entitiesConfig)) {
    validateFalse(
            account.usedAutoAssociations() >= account.maxAutoAssociations(),
            NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
}
```

I think you should either put both in one if-statement or use them as a parameter for the validate-call.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AutoAccountCreator.java
kselveliev: Understood, comment addressed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenUtil.java
```diff
@@ -0,0 +1,29 @@
...
+package com.hedera.node.app.service.token.impl.util;
+
+public class TokenUtil {
+
+    /**
+     * The value for unlimited automatic associations
+     */
+    public static final int UNLIMITED_AUTOMATIC_ASSOCIATIONS = -1;
+
+    public TokenUtil() {
```
jsync-swirlds, line 26: Why make this public and throw?  Why not just make the constructor private and empty?

kselveliev: Right, I autogenereted it with intellij and then forgot to switch it to private.
I made it private.

kselveliev:
I'm merging this, the code can later be reviewed again when the PR is opened against develop.

PR #13446, opened 2024-05-21
chore: move scratchpad out of common

### LINKED ISSUE #13445: Move scratchpad out of swirlds common
The scratchpad utility currently lives in swirlds common. It should live in platform core.

cody-littley:
closes #13445


lpetrovic05:
Can you please add an explanation on why this needs to live in platform

cody-littley:
@lpetrovic05 

> Can you please add an explanation on why this needs to live in platform

The broad answer is that nothing should live in swirlds common, and this is a utility used only by the platform. I'm hoping we can delete common sooner rather than later.

The specific answer is that as I want to create some static utilities for organizing the platform file system. Scratchpads should use these utilities for deciding their file paths. Since common can't depend on platform, scratchpads must end up in platform.

alittley (approved):

edward-swirldslabs (approved):

PR #13449, opened 2024-05-21
fix: explicitly set `ledger.id=0x00` 

tinker-michaelj:
**Description**:
 - Explicitly set the mainnet ledger id in _application.properties_ to simplify transaction tooling that doesn't expect an empty `0.0.121` file.


tinker-michaelj:
Retargeted to `release/0.49.

PR #13450, opened 2024-05-21
fix: explicitly set ledger.id=0x00

tinker-michaelj:
**Description**:
 - Explicitly set the mainnet ledger id in _application.properties_ to simplify transaction tooling that doesn't expect an empty `0.0.121` file.


dalvizu (approved):

SimiHunjan (approved):

steven-sheehy (approved):

Nana-EC (approved):

nathanklick (approved):
Assuming this is the correct ledger ID value for mainnet, then I see no issues with this change.

poulok (approved):

rbair23 (approved):

PR #13479, opened 2024-05-23
feat: Add `redirectForAccount` proxy contract support

### LINKED ISSUE #13333: Implement redirectForAccount proxy contract

lukelee-sl:
**Description**:
- Adds support for to return the byte code for the `redirectForAccount` proxy contract if invoked by an EOA as a top level call.
- Added xTests
- Added HAPI tests.
- Turn on feature flag for HAS in previewnet and local node.

**Related issue(s)**:

Fixes #13333

** Notes for reviewers:

I added a state to `DispatchingEvmFrameState`.  This was necessary in order to return the correct bytecode and code hash when called from the evm.  The `redirectForAddress` proxy bytecode should only be returned on a top level call from an EOA as internal calls should use the `IHederaAccountService` and `HederaAccountService` interface/contract.

**Checklist**


jeromy-cannon (dismissed):

hedera-node/hedera-app/src/xtest/java/contract/HbarAllowanceXTest.java
```diff
@@ -0,0 +1,119 @@
...
+        runHasCallAndExpectOnSuccess(
+                APPROVED_BESU_ADDRESS,
+                Bytes.wrapByteBuffer(HbarAllowanceTranslator.HBAR_ALLOWANCE.encodeCall(
+                        Tuple.of(OWNER_HEADLONG_ADDRESS, APPROVED_HEADLONG_ADDRESS))),
+                output -> assertEquals(
+                        asBytesResult(HbarAllowanceTranslator.HBAR_ALLOWANCE_PROXY
+                                .getOutputs()
+                                .encodeElements((long) SUCCESS.getNumber(), BigInteger.valueOf(1_000L))),
+                        output),
+                "Successful execution of hbarAllowance for SENDER by OWNER.");
```
david-bakin-sl, line 74: Text here and at L64 should be different to distinguish tests?

hedera-node/hedera-app/src/xtest/java/contract/HbarApproveXTest.java
```diff
@@ -0,0 +1,119 @@
...
+                bytesForRedirectForAccount(
+                        HbarApproveTranslator.HBAR_APPROVE_PROXY.encodeCallWithArgs(
+                                UNAUTHORIZED_SPENDER_HEADLONG_ADDRESS, BigInteger.valueOf(100L)),
+                        OWNER_ID),
+                output -> assertEquals(
+                        asBytesResult(
+                                HbarApproveTranslator.HBAR_APPROVE.getOutputs().encodeElements((long)
+                                        SUCCESS.getNumber())),
+                        output),
+                "Successful execution of hbarAllowance for SENDER by OWNER.");
```
david-bakin-sl, line 65: Here and at L75 `hbarAllowance` -> `hbarApprove`, plus should be different text?

david-bakin-sl (dismissed):
All reviewed but the HAPI suite ... I'll have to get to that later.  

Some of these things you might consider better done later rather than in this PR (if you want them done at all) and if so leave some `// FUTURE` comments so we sweep over them shortly ...

hedera-node/test-clients/src/main/resource/contract/contracts/IHRC632/IHRC632.bin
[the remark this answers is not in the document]
lukelee-sl: It's probably empty because it is an interface.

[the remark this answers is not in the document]
david-bakin-sl: Oh!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
```diff
@@ -165,7 +165,7 @@ private MessageFrame finishedAsCall(
             if (account == null && contractMustBePresent(config, featureFlags, contractId)) {
                 validateTrue(transaction.permitsMissingContract(), INVALID_ETHEREUM_TRANSACTION);
             } else {
-                code = account.getEvmCode();
+                code = account.getEvmCode(Bytes.wrap(transaction.payload().toByteArray()));
```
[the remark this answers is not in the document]
lukelee-sl: I'll add a check.

tinker-michaelj (commented):
I'm pretty confused here I think...why can't we just:
1. Make the redirect bytecode a no-op for anything except the supported function selectors; and,
2. Make `ProxyEvmAccount` behave like `TokenEvmAccount` and always returns the code and code hash implied by that redirect bytecode?

Internal calls would also "just work" in that case afaict...this many touch points feels like massive overkill. 🤔

lukelee-sl:
> I'm pretty confused here I think...why can't we just:
> 
> 1. Make the redirect bytecode a no-op for anything except the supported function selectors; and,
> 2. Make `ProxyEvmAccount` behave like `TokenEvmAccount` and always returns the code and code hash implied by that redirect bytecode?
> 
> Internal calls would also "just work" in that case afaict...this many touch points feels like massive overkill. 🤔

The main problem is that I need to determine whether to do one of three things when asked for the bytecode on an account
1) return the `redirectForAddresses` bytecode if the function selector matches the system contract functions.
2) return the contract bytecode if the account is a contract. or 
3) return 0x otherwise.  
There is also the returning the correct code hash to consider.

Given the comments I received from you and David an alternative implementation to consider is to create a third descendant of `AbstractMutableEvmAccount` (currently there is `TokenEvmAccount` and `ProxyEvmAccount`).  

Where 
- `TokenEvmAcccount`  behaves as it does now.
- Rename the current `ProxyEvmAccount` to `ProxyEvmContract` and return this if the account address is a contract.  This behaves the same as the current `ProxyEvmAccount`.
- Create a new `ProxyEvmAccount` that behaves as you describe above - return 0x or the redirectForAccount bytecode if the function selector matches the system contract support functions.

Would this be a cleaner implementation?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java
```diff
@@ -85,14 +85,20 @@
 public class DispatchingEvmFrameState implements EvmFrameState {
     public static final Key HOLLOW_ACCOUNT_KEY =
             Key.newBuilder().keyList(KeyList.DEFAULT).build();
-    private static final String TOKEN_BYTECODE_PATTERN = "fefefefefefefefefefefefefefefefefefefefe";
+    private static final String ADDRESS_BYTECODE_PATTERN = "fefefefefefefefefefefefefefefefefefefefe";
 
     @SuppressWarnings("java:S6418")
```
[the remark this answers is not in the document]
lukelee-sl: Hard-coded secrets are security-sensitive.  It wrongly thinks the bytes are for a key.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/ProxyEvmAccount.java
```diff
@@ -49,6 +52,16 @@ public class ProxyEvmAccount extends AbstractMutableEvmAccount {
     protected final AccountID accountID;
     protected final EvmFrameState state;
 
+    /*
+     * Four byte function selectors for the two functions that are eligible for proxy redirection
+     * in the Hedera Account Service system contract
+     */
+    private static final Set<Integer> ACCOUNT_PROXY_FUNCTION_SELECTOR = Set.of(
+            // hbarAllowance(address spender)
+            0xbbee989e,
```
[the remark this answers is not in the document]
lukelee-sl: done

david-bakin-sl (approved):
I do like this better, TY!!  LGTM!

hedera-node/hedera-app/src/xtest/java/contract/HbarApproveXTest.java
david-bakin-sl: Still needs a fix if you want

jeromy-cannon (approved):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/HederaEvmAccount.java
```diff
@@ -67,8 +68,11 @@ public interface HederaEvmAccount extends MutableAccount {
...
      * {@link Account#getCode()}.
      *
+     * @param functionSelector the function selector to use when fetching the code.  If more than 4 bytes for the
+     *                         function selector is passed in, only the first 4 bytes will be used.
+     *                         Only relevant for regular accounts.
      * @return the EVM code for this account
      */
     @NonNull
-    Code getEvmCode();
+    Code getEvmCode(@NonNull final Bytes functionSelector);
```
tinker-michaelj, line 77: We don't need this argument any more, do we?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
```diff
@@ -159,18 +157,31 @@ private MessageFrame finishedAsCall(
             @NonNull final ContractsConfig config) {
         Code code = CodeV0.EMPTY_CODE;
         final var contractId = transaction.contractIdOrThrow();
+        final var contractMustBePresent = contractMustBePresent(config, featureFlags, contractId);
 
-        if (canLoadCodeFromAccount(transaction, worldUpdater, contractId, config, featureFlags)) {
-            final var account = worldUpdater.getHederaAccount(to);
-            if (account == null && contractMustBePresent(config, featureFlags, contractId)) {
-                validateTrue(transaction.permitsMissingContract(), INVALID_ETHEREUM_TRANSACTION);
+        // If the contract has been deleted, then always use empty byte code
```
david-bakin-sl, line 162: Comment at L162 doesn't agree with comment at L180/182.  Or maybe it's just in the wrong place.

david-bakin-sl: In fact I would rewrite with quick exit on deleted contract first, something like

```java
if (contractDeleted(worldUpdater, contractId)) {
    validateTrue(....);
   return buildMessageCall(to, transaction, CodeV0.EMPTY_CODE);
}
final var account ...
if (account == null) {
   if (contractMustBePresent) validateTrue(...);
   return buildMessageCall(to, transaction, CodeV0.EMPTY_CODE);
}
// Now we've got a valid contract, continue:
...
```

This unnests everything and leaves the remaining code as the happy path which will be easy to read.

david-bakin-sl (commented):
Comments.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/utils/FrameBuilder.java
lukelee-sl: Nice suggestion!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/HederaEvmAccount.java
lukelee-sl: we might if we want to restrict `TokenEvmAccount` and `ProxyEvmAccount` to return the proxy contract byte code only if the function selector matches the expected functions on tokens/accounts.

david-bakin-sl (approved):
LGTM

jeromy-cannon (approved):

jeromy-cannon (dismissed):

jeromy-cannon (dismissed):

jeromy-cannon (dismissed):

david-bakin-sl (approved):
LGTM

jeromy-cannon (approved):

tinker-michaelj (approved):
LGTM, tyvm @lukelee-sl !

Nana-EC (approved):

PR #13493, opened 2024-05-23
Remove all threads from Cryptography.

### LINKED ISSUE #13492: Remove threads from cryptography
Now that mono-services is no longer production code, there exists no remaining production use case of async cryptography. Remove all threads from the `Cryptography` utility.

cody-littley:
closes #13492


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
```diff
@@ -129,24 +129,6 @@ default Hash getNullHash() {
...
-     * Starting in version 0.43 and onwards, the {@link SignatureType#ECDSA_SECP256K1} signature algorithm requires the
-     * payload to be a KECCAK-256 hash of the original message. Verification will fail if the message is not 32 bytes in
-     * length and the output of 256-bit hashing function.
-     *
-     * @param signatures a list of signatures to be verified
-     * @deprecated use {@link #verifySync(List)} instead, asynchronous verification is a feature that is scheduled for
-     * removal
-     */
-    @Deprecated(forRemoval = true)
-    void verifyAsync(final List<TransactionSignature> signatures);
```
hendrikebbers, line 148: When we remove `verifyAsync(...)` it would make make sense to rename `verifySync(...)` to `verify(...)`.

hendrikebbers (approved):
Code looks good. I ask myself if performance tests are needed since you moved from a non blocking api to a blocking api at several code places.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Cryptography.java
lpetrovic05: I would be in favor of eventually removing the whole module. A lot of it is based on our legacy serialization which we are moving away from.

cody-littley: @hendrikebbers 

> Code looks good. I ask myself if performance tests are needed since you moved from a non blocking api to a blocking api at several code places.

There are no production use cases of async verification, only testing apps and mono-services (which is no longer production code).

alittley (approved):

edward-swirldslabs (approved):

PR #13507, opened 2024-05-24
feat: Improve logging for failure case when reading config file

### LINKED ISSUE #13506: Improve logging for failure case when reading config file.
### Problem

I think we can improve the logs for the failure case when reading the config file.

### Solution

- Add a log message with the ParseException.

### Alternatives

_No response_

matteriben:
## Description

- Improve logging for failure case when reading config file

### Related Issues

- Closes #13506

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
@@ -86,8 +86,10 @@ public final class LegacyConfigPropertiesLoader {
         try (final Scanner scanner = new Scanner(configPath, StandardCharsets.UTF_8)) {
             final AddressBook addressBook = new AddressBook();
             boolean nextNodeIdParsed = false;
+            long lineNumber = 0;
             while (scanner.hasNextLine()) {
                 final String line = readNextLine(scanner);
+                lineNumber++;
                 if (!line.isEmpty()) {
                     final String[] lineParameters = splitLine(line);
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
```diff
@@ -117,6 +120,7 @@ public static LegacyConfigProperties loadConfigFile(@NonNull final Path configPa
                                     addressBook.add(address);
                                 }
                             } catch (final ParseException ex) {
+                                logger.info(LogMarker.CONFIG.getMarker(), "{}:{}:{}: {}", configPath, lineNumber, ex.getErrorOffset(), ex.getMessage());
```
cody-littley, line 123: I suggest logging this at an error level and using the `EXCEPTION` marker. This is an error case, and we want parse exceptions to cause our tests to fail. Logging an info message is useful for debugging but will not cause the test validator to fail.
```suggestion
                                logger.error(EXCEPTION.getMarker(), "{}:{}:{}: {}", configPath, lineNumber, ex.getErrorOffset(), ex.getMessage());
```

cody-littley: Perhaps while this code is being modified, we could make the log message more informative. Perhaps something like this:

```
file {}, line {}, column {}: {}
```

cody-littley (commented):
It looks like spotless is failing. To fix, navigate to the root of the repo and run the command `./gradlew spotless apply`.

cody-littley (approved):

PR #13512, opened 2024-05-24
feat: Finally make Hash immutable

### LINKED ISSUE #13510: Remove remaining usage of `Hash.getValue()`
Part of this mini-epic #12456

- Remove the last usage of `getValue()`
- Remove the byte array variable from Hash
- Remove the method `getValue()`

lpetrovic05:
closes #13510


hedera-node/hedera-app/src/jmh/java/com/hedera/node/app/signature/impl/DoNothingCryptoEngine.java
```diff
@@ -24,6 +24,7 @@
 import com.swirlds.common.crypto.TransactionSignature;
 import com.swirlds.common.io.SelfSerializable;
 import java.util.List;
+import org.jetbrains.annotations.NotNull;
```
anthony-swirldslabs, line 27: Isn't it `edu.umd.cs.findbugs.annotations.NonNull` that we use mostly? While the extra dependency on the `org.jetbrains` exists today since we use this annotation in a few places (probably by mistake and unintentionally, and should've used the findbugs instead), I don't believe we want to have this dependency longer term.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -178,14 +147,17 @@ public int getVersion() {
     @Override
     public void serialize(@NonNull final SerializableDataOutputStream out) throws IOException {
         requireNonNull(digestType, "digestType");
-        requireNonNull(value, "value");
+        requireNonNull(bytes, "value");
```
anthony-swirldslabs, line 150: Nit: s/"value"/"bytes"/

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -221,26 +193,15 @@ public boolean equals(final Object obj) {
...
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public int hashCode() {
-        return ((value[0] << 24) + (value[1] << 16) + (value[2] << 8) + (value[3]));
+        return bytes.getInt(0);
     }
```
anthony-swirldslabs, line 205: Minor: I understand that the new implementation implicitly assumes that PBJ Bytes are big endian by default, and so it is equivalent to the old one. Two points:
1. It might make the code less cryptic if we avoid the implicit assumption and use the `getInt(offset, byteOrder)` version of the method explicitly. Or, if we want the maximum performance, perhaps a comment here in code suggesting that we're aware of this assumption might be helpful.
2. Out of curiosity: why does this hashCode() implementation ignore the rest of the bytes in the bytes array? Is this done for performance reasons only? Does it make sense to compute the hash using all the bytes instead? Note that Bytes.hashCode() does this already. Although I'm unsure if anything would break if this implementation changes - do we persist this hash value anywhere?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -89,21 +84,7 @@ public Hash(@NonNull final byte[] value, @NonNull final DigestType digestType) {
...
-     *
-     * @param byteBuffer
-     * 		a buffer that contains the data for this hash
-     * @param digestType
-     * 		the digest type of this hash
-     */
-    public Hash(@NonNull final ByteBuffer byteBuffer, @NonNull final DigestType digestType) {
-        this(digestType);
-        byteBuffer.get(this.value);
+        this.bytes = Bytes.wrap(value);
```
anthony-swirldslabs, line 87: Note that this relies on the calling code to be nice and never modify the bytes in the array. I'm not sure about the performance implications - perhaps they outweigh the concern - but I think that making a copy of the array here would be safer from immutability perspective. Otherwise, I suggest to update the javadoc for this constructor (and perhaps this class, too) to clarify the requirement for the callers to be nice.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/utility/CommonUtils.java
```diff
@@ -217,6 +218,28 @@ public static String hex(final byte[] bytes, final int length) {
...
     }
 
+    /**
+     * Converts Bytes to a lowercase hexadecimal string.
+     *
+     * @param bytes  the bytes to hexadecimal
+     * @param length the length of the array to convert to hex
+     * @return a {@link String} containing the lowercase hexadecimal representation of the byte array
+     */
+    public static String hex(final Bytes bytes, final int length) {
```
anthony-swirldslabs, line 228: BTW, couldn't we use Bytes.toHex() instead of this utility method?

hedera-node/hedera-app/src/jmh/java/com/hedera/node/app/signature/impl/DoNothingCryptoEngine.java
Neeharika-Sompalli: +1

hendrikebbers: +1

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
hendrikebbers: +1

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -118,21 +99,9 @@ public Hash(@NonNull final Hash other) {
         }
 
         this.digestType = other.digestType;
-        this.value = Arrays.copyOf(other.value, other.value.length);
         this.bytes = other.bytes;
```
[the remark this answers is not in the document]
hendrikebbers: +1

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
hendrikebbers: +1

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/CryptoEngine.java
```diff
@@ -214,7 +214,8 @@ public Hash digestSync(final SelfSerializable serializable, final DigestType dig
     public Hash digestSync(
             final SerializableHashable serializableHashable, final DigestType digestType, final boolean setHash) {
         try {
-            final Hash hash = serializationDigestProvider.compute(serializableHashable, digestType);
+            final Hash hash =
+                    new Hash(serializationDigestProvider.compute(serializableHashable, digestType), digestType);
```
hendrikebbers, line 218: can you split this in 2 lines by storing `serializationDigestProvider.compute(serializableHashable, digestType)` in a var?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/PlatformSigner.java
```diff
@@ -64,10 +64,24 @@ public PlatformSigner(@NonNull final KeysAndCerts keysAndCerts) {
...
+     * Same as {@link #sign(byte[])} but takes a {@link Bytes} object instead of a byte array.
+     */
+    private @NonNull com.swirlds.common.crypto.Signature signBytes(@NonNull final Bytes data) {
+        try {
+            data.updateSignature(signature);
+            return new com.swirlds.common.crypto.Signature(SignatureType.RSA, signature.sign());
+        } catch (final SignatureException e) {
+            // this can only occur if this signature object is not initialized properly, which we ensure is done in the
+            // constructor. so this can never happen
+            throw new CryptographyException(e);
```
hendrikebbers, line 77: please add a message to the exception

hedera-node/hedera-app/src/jmh/java/com/hedera/node/app/signature/impl/DoNothingCryptoEngine.java
lpetrovic05: inteliij sneaked this one in without me noticing, fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: This comment made me find [an issue](https://github.com/hashgraph/pbj/issues/256) in the `Bytes` class.

1. I don't think we care if its big endian or little, so long as its consistent. And not reversing the bytes is better for performance, which is important. I will add a comment.
2. Performance is very important for a few different use cases of this class. For gossip, we make huge hash sets of events, and when profiling a long time ago, this made a huge difference. Apart from performance, we are dealing with a cryptographic hash, so the first four bytes are just as optimal as anything else we can compute. I don't see why changing the implementation would affect anything, I think the only important thing is that this method is consistent at runtime.

I will add comments to the code with explanations for everything.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
lpetrovic05: I would not introduce the performance overhead of copying the array, I will add comments.

lpetrovic05: In the future, we can move towards not using `byte[]` at all

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/utility/CommonUtils.java
lpetrovic05: Bytes.toHex() does not allow us to specify the length. I could convert the whole thing to hex and take a substring if that's what you prefer

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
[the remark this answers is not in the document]
lpetrovic05: I changed the comment. There is no need to create a deep copy since a hash should be immutable

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/PlatformSigner.java
lpetrovic05: added

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/engine/CryptoEngine.java
lpetrovic05: done

platform-sdk/swirlds-merkle/src/testFixtures/java/com/swirlds/merkle/test/fixtures/map/lifecycle/SaveExpectedMapHandler.java
```diff
@@ -18,15 +18,22 @@
 
 import static com.swirlds.logging.legacy.LogMarker.EXCEPTION;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
```
lpetrovic05, line 21: Please ignore these PTT related changes. Maintaining this code is a waste of time because we want to delete it as soon as possible.

alittley (approved):

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
anthony-swirldslabs: Thanks for filing the issue, but there's no bugs - I've commented in that ticket and provided more details.

Also, thanks for the additional context and the comment in code. It makes sense to me.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
anthony-swirldslabs: Sounds good.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/utility/CommonUtils.java
anthony-swirldslabs: Alternatively, you could check if anyone uses toHex() already, and if not, perhaps you could move the length implementation into that method so that it finally becomes useful.

But I'm okay either way.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -221,26 +188,23 @@ public boolean equals(final Object obj) {
...
     public int hashCode() {
-        return ((value[0] << 24) + (value[1] << 16) + (value[2] << 8) + (value[3]));
+        // A few notes about this implementation:
+        // - it is very important that this method has very low overhead, some usages are very performance sensitive
+        // - since this is a cryptographic hash, it is safe to use the first 4 bytes of the hash as the hash code.
+        //   it fulfills the requirements of hashCode, it is consistent with equals, and it is very unlikely to conflict
+        //   with other hash codes. by using the first 4 bytes, we can avoid unnecessary overhead
+        // - it does not make a difference if the bytes are big endian or little endian, as long as they are consistent.
+        //   this is why we don't specify the order when calling getInt(), we use whatever order is the default to avoid
+        //   the overhead of reordering the bytes
```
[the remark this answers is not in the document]
anthony-swirldslabs: I'd prefer if we use the getInt(long, ByteOrder) still. But I'm okay as long as we rely on the default byte order consciously as this comment suggests.

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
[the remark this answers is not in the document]
lpetrovic05: maybe I didn't phrase it well, what I meant was consistent during runtime. if it changes after a restart, that is fine.

jeromy-cannon (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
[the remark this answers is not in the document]
cody-littley: > I'd prefer if we use the getInt(long, ByteOrder) 

`hashCode()` returns an int, so there is no benefit to getting a long. The higher order bits will just be chopped off when it is cast to an int.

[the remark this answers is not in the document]
cody-littley: > This appears to place a requirement that the default never change, and remain identical across system architectures.

I don't think we should make this a contract of this class, even if we happen to never change its implementation. Any time we are relying on `hashCode()` to return the same value value between JVM runs should be a considered a code smell. I recall there was an issue with virtual maps a while back where it broke because it assumed `hashCode()` would never change. This is a super dangerous assumption I think.

cody-littley (approved):
Missing some nullity annotations, but I'm intentionally not bringing up nits on this PR since it's one of those difficult 20-reviewer sorts of things.

Neeharika-Sompalli (approved):
LGTM! Thanks @lpetrovic05

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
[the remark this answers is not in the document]
jsync-swirlds: Definitely the wording is a bit confusing here.  The following _might_ be clearer, though it's not easy to express in a short comment. The comment confusion is not critical, and can be done separately.
```suggestion
        // - it does not make a difference if the bytes are big endian or little endian, as long as the order
        //   is reasonably stable. This is why we don't specify the byte order here; we can accept the current
        //   default byte order and avoid the overhead of reordering the bytes
```

PR #13519, opened 2024-05-24
fix: waive privileged signing requirements for file append

### LINKED ISSUE #13475: `FileAppendHandler#preHandle()` should waive privileged signing requirements
### Description

The fee schedules admin (`0.0.56`) should not be required to sign with the `0.0.111` admin key doing a `FILE_APPEND`.

### Steps to reproduce

1. Set the `0.0.56` key to a different key from `0.0.111`
2. Attempt a fee schedules update e.g. with `gcr.io/hedera-registry/yahcli:0.4.2`; the `FileAppend`'s will fail with `INVALID_SIGNATURE`.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.49.7

### Operating system

None

povolev15:
Waive privileged signing requirements for file append

**Related issue(s)**:

Fixes #13475

### DIFF
--- a/hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileAppendHandler.java
+++ b/hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileAppendHandler.java
@@ -101,7 +101,7 @@ public class FileAppendHandler implements TransactionHandler {
         final var fileStore = context.createStore(ReadableFileStore.class);
         final var transactionFileId = requireNonNull(op.fileID());
         preValidate(transactionFileId, fileStore, context);
-        final var areSignaturesWaived = fileSignatureWaivers.areFileUpdateSignaturesWaived(body, context.payer());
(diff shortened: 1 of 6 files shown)


thomas-swirlds-labs (dismissed):
LGTM. Thanks! @tinker-michaelj 💯

tinker-michaelj (changes requested):
Oops this doesn't actually work, will ping out-of-band; we need a `FILE_APPEND`-specific version of [this method](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileSignatureWaiversImpl.java#L49).

povolev15:
> Oops this doesn't actually work, will ping out-of-band; we need a `FILE_APPEND`-specific version of [this method](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileSignatureWaiversImpl.java#L49).

fixed it thank for the catch

tinker-michaelj (approved):
LGTM, ty @povolev15 !

Issue #13522, opened 2024-05-24
Don't include data source builder configuration into state snapshots


artemananiev:
When a virtual map / virtual root node is serialized to a stream (as a part of a state snapshot), it writes its data source builder configuration. Then this configuration is read back, when the snapshot is loaded.

Note, however, after a virtual map is restored, its data source builder isn't used to create new data sources, but only to copy/save/restore existing data sources. It doesn't need any config, like MerkleDbTableConfig for MerkleDb builders. There is no need to save this information to state snapshots.

Another thing to clean up is in-memory vs MerkleDb builders. In-memory data source builder is used in tests to create new data sources. Tests that serialize/deserialize virtual maps may be rewritten to always use MerkleDb. It would simplify VirtualRootNode code, so that it can always use MerkleDb data source builder, when deserialized.

PR #13524, opened 2024-05-24
fix: 13522: Don't include data source builder configuration into state snapshots

artemananiev:
Fix summary:
* `VirtualRootNode` serialization version is increased to 2, which doesn't store data source builder
* `VirtualDataSourceBuilder` is changed to a service interface to use with `ServiceLoader` mechanism
* When a root node is deserialized, it now always creates an instance of MerkleDb data source builder (using `ServiceLoader`)
* Unit tests that use (de)serialization are updated to use MerkleDb
* A new unit test is added to test virtual root node deserialization from version 2

Fixes: https://github.com/hashgraph/hedera-services/issues/13522


platform-sdk/swirlds-jasperdb/src/main/resources/META-INF/services/com.swirlds.virtualmap.datasource.VirtualDataSourceBuilder
```diff
@@ -0,0 +1 @@
+com.swirlds.merkledb.MerkleDbDataSourceBuilder
```
hendrikebbers, line 1: Instead of handling the services file on our own we started in base to move to the Google Auto annotations. See:

https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-config-impl/build.gradle.kts
`mainModuleInfo { annotationProcessor("com.google.auto.service.processor") }`

https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-config-impl/src/main/java/com/swirlds/config/impl/internal/ConfigurationBuilderFactoryImpl.java
`@AutoService(ConfigurationBuilderFactory.class)`

We should try to have a common pattern in the repo. Open for discussion

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/datasource/VirtualDataSourceBuilder.java
```diff
@@ -37,6 +38,20 @@
...
 
+    /**
+     * Creates a new data source builder to use, when a virtual root node is deserialized. Such builder
+     * can't be used to create new data sources, since it lacks virtual map configuration and types,
+     * but it can copy and restore existing data sources.
+     */
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    static <K extends VirtualKey, V extends VirtualValue> VirtualDataSourceBuilder<K, V> newDeserializedBuilder() {
+        final ServiceLoader<VirtualDataSourceBuilder> services = ServiceLoader.load(VirtualDataSourceBuilder.class);
+        for (final VirtualDataSourceBuilder builder : services) {
```
hendrikebbers, line 49: We should define some best practices how ServiceLoader should be used. Have it on my todo-list to create a doc. In this code you for example do not check if more than 1 impl is available.

artemananiev:
Withdrawing this PR for now. A different solution is proposed, I'll prototype that.

PR #13527, opened 2024-05-25
fix: Skip transactions older than the software version

### LINKED ISSUE #13487: Transactions from a previous software version should be skipped
### Description

Post-`v0.49.7` upgrade, records were created for expired transactions that were submitted before the upgrade and reached consensus after.

These records should not have been created; instead the expired transactions should have been completely skipped since they were created by an earlier software version. (C.f. [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/StandardProcessLogic.java#L112) in mono-service.) 

### Steps to reproduce

Do an analogous upgrade.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.49.7

### Operating system

None

mhess-swl:
Closes #13487


povolev15 (dismissed):
LGTM TY @mhess-swl

tinker-michaelj (approved):
LGTM, tyvm @mhess-swl !

PR #13529, opened 2024-05-25
fix: include retry information in timeout failures

tinker-michaelj:
**Description**:
 - Restructure status waits to surface timeout details in exception messages.
 - Clean up `LeakyContractTestsSuite`:
    1. Fix a bug that failed to preserve EVM version after overriding to `v0.34`.
    2. Update expectations for new `MAX_GAS_LIMIT_EXCEEDED` pure checks.
    3. Move some specs that actually use all default properties to run concurrently in `ContractCallSuite`.


lukelee-sl (approved):
LGTM. thank you

PR #13533, opened 2024-05-25
fix: update first key on insertion or removal; add log validation to CI

### LINKED ISSUE #13514: `firstContractKey` should update even on net-zero slot usage change
### Description

A contract call both added and removed `n` slots in the same transaction, but its first contract key was not updated. (This has no functional impact but would ultimately prevent full expiration of the contract if expired and deleted.)

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.49.7

### Operating system

None

### LINKED ISSUE #13534: Add log validation to CI checks
### Problem

If a _hgcaa.log_ in a CI check network contains an unexpected `WARN` or `ERROR`, that check will not automatically fail.

### Solution

Add a `UtilVerb` to validate the target network logs and include at the end of each CI check.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13514
     * Adds a `@HapiTest` that fails log validation without the fix in `IterableStorageManager`.
 - Closes #13534
     * Removes pointless `FeeManager` `WARN` logs surfaced during check execution on the PR.
     * Catches and ignores a `ConcurrentModificationException` if thrown while answering a `getAccountRecords()` query (this has always been a best-effort query and is unused in production environments).
     * Falls back to scanning the _hgcaa.log_ for a node status if its Prometheus endpoint doesn't bind.

**Required cherry-picks**:
- [ ] Node changes to `release/0.50`


lukelee-sl (approved):
LGTM.  thank you!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -324,14 +325,18 @@ public List<TransactionRecord> getRecords(@NonNull final AccountID accountID) {
...
         // While we still need to gather more records, collect them from the different histories.
         final var records = new ArrayList<TransactionRecord>(maxRemaining);
-        for (final var transactionID : transactionIDs) {
-            final var history = histories.get(transactionID);
-            if (history != null) {
-                final var recs = history.orderedRecords();
-                records.addAll(recs.size() > maxRemaining ? recs.subList(0, maxRemaining) : recs);
-                maxRemaining -= recs.size();
-                if (maxRemaining <= 0) break;
+        try {
```
david-bakin-sl, line 328: Or why not just make the local `records` be a `ConcurrentLinkedQueue` and pretty much be done with it.  (Except then L342 becomes `records.stream().sorted(...)`.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/RecordAssertions.java
```diff
@@ -85,12 +86,20 @@ public static void triggerAndCloseAtLeastOneFile(final HapiSpec spec) throws Int
...
+        });
+    }
+
+    public static void doIfNotInterrupted(@NonNull final InterruptibleRunnable runnable) {
+        requireNonNull(runnable);
+        try {
+            runnable.run();
+        } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
+            throw new IllegalStateException(e);
```
david-bakin-sl, line 102: interesting!  I've often wondered about that Java recommendation about always rethrowing `InterruptedException` ...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/InterruptibleRunnable.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.services.bdd.spec.utilops.streams;
+
+public interface InterruptibleRunnable {
+    void run() throws InterruptedException;
```
david-bakin-sl, line 20: Now that I've seen `RecordAssertions` why not provide a default. method here that encapsulates an abstract `run()` with the try/catch/rethrow something better?

david-bakin-sl (approved):
LGTM (a couple of comments for your consideration)

jeromy-cannon (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
tinker-michaelj: The CME arises from traversing the iterator constructed by this `for`-`each`,
```
            for (final var transactionID : transactionIDs) {
```
...which I'm not quite sure how to work around. But this query is literally unused afaik. 🤷

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/RecordAssertions.java
tinker-michaelj: Sonar likes it. 🤷

PR #13540, opened 2024-05-27
feat: extract `HederaNetwork` interface with initial `SubProcessNetwork` impl

### LINKED ISSUE #13539: Make `HederaNetwork` an interface with a gRPC-agnostic API for submitting transactions and sending queries to the network, with of course the option to target specific nodes. (These APIs should still be blocking, however, so support easy implementation via the existing `HapiClients` class.)

tinker-michaelj:
**Description**:
 - Closes #13539 
     1. Makes `HederaNetwork` an interface with `send()` and `submit()` methods for performing operations against the network.
     2. Provides implementation support via `AbstractNetwork` (for basic name and node management) and `AbstractGrpcNetwork` (for `HapiClients` usage).
     3. Implements both a `SubProcessNetwork` (for `@HapiTest`s run via default `Test` task); and a `RemoteNetwork` (for continuing to support `HapiSpec`'s run via `SuiteRunner` in JRS).
     4. Extracts `switch` statements to select appropriate gRPC endpoints into `GprcUtils`; eliminates all explicit gRPC usage in concrete `HapiSpecOperation` classes and performs all operations through the spec's `HederaNetwork`.
         * For `HapiTxnOp` subclasses this was trivial (simply replace `callToUse`); for `HapiQueryOp` it required a refactor to move some duplicated `submitWith()` behavior into the base class and introduce an `abstract queryFor()` method to isolate sending the query to just the base class.


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #13558, opened 2024-05-28
docs: Design doc for HIP-632 `isAuthorizedRaw`

david-bakin-sl:
**Description**:

Design doc for HIP-632's `isAuthorizedRaw`.

**Related issue(s)**:

One task of #13560 

**Notes for reviewer**:

Currently needs one further update to `isAuthorizedRaw` on my side, then I'll push that section.

**Checklist**

### DIFF
--- a/hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
+++ b/hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
@@ -52,11 +52,39 @@ match the address given as an argument to `isAuthorizedRaw`.
 * If an Hedera account then it gets the account, looks in it to find a key _which it must
 have_ and which must be a _single key_ (otherwise: failure).  Then given the key and the
 signature verifies that the signature matches the message hash and is attested by 
-the account. This is an operation provided by `HandleContext` (which can be accessed via
-the `HederaWorldUpdater.Enhancement`).
+the account. Looking up the accounts is an operation provided by `HandleContext` (which can be accessed via
+the `HederaWorldUpdater.Enhancement`).  Validating the signature is done by directly using methods
+  from the platform.
   * Some fee will be charged to account for the signature verification cost, but it
 will have to be hard-coded (at this time) because we have no access to that fine-grained
 fee now.
+    * Current proposal is to charge a fixed fee (in dollars) derived from the cost of a 
+      crypto transfer of 1 tinybar signed with an EC key.
+
+A feature flag will control enabling `isAuthorizedRaw`.
+
+## Testing
+
+E2E testing necessary:
+
+### Positive
+
+(All with sufficient gas:)
+
+1. EVM address, valid hash+signature -> SUCCESS
+1. EVM address, _invalid_ hash+signature -> FAILED
+1. EVM address, valid hash+signature but recovered address does not match given address -> FAILED
+1. Hedera account w/ single ED key, valid hash+signature -> SUCCESS
+1. Hedera account w/ single ED key, _invalid_ hash+signature -> FAILED
+
+### Negative
+
+1. EVM address, insufficient gas -> FAILED
+1. Hedera account, insufficient gas -> FAILED
+1. Hedera account w/ 1 key which is EC -> FAILED
+1. Hedera account w/ >1 key -> FAILED
+
+__
 
 ## TBD
 
@@ -75,10 +103,11 @@ system contract calls, not to mention varying the fee for individual system cont
 methods.  (All existing system contract methods are either "cheap" - as in too small to
 charge for - or involve child transactions that themselves do incur fees.)  We _can_
 charge hard-coded fees.
-  * TBD #1: A fee schedule
-mechanism to charge fees for individual system contract methods 
-  * TBD #2: How do we choose the fee for the Hedera signature verification part of
-`isAuthorized` and `isAuthorizedRaw`?
+    
+* Need to have a way to get to signature verification via the `HandleContext` (instead of using
+  platform classes directly).
+
+
 
 
 
--- a/hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
+++ b/hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
@@ -15,6 +15,7 @@ may be several/many per Hedera account, there is in fact only _one_: the _alias_
 
 * This mainly affects the method `getVirtualAddresses` which will return an array, but
 it will have _at most one_ EVM address in it.
+* An issue is present to edit HIP-632 to clarify that it is dealing with aliases, not virtual addresses.  
   
 ## Phased implementation
 
@@ -34,8 +35,6 @@ Following the existing pattern for HAS each method will be given a package under
   that does the work.
   
 ## Design for `isAuthorizedRaw`
-
-[TBD shortly as I fix up some problems with understanding HAS]
 
 `isAuthorizedRaw` has two flows: Either the argument address is a Hedera account, or it is
 an EVM address.
@@ -63,6 +62,17 @@ fee now.
 
 A feature flag will control enabling `isAuthorizedRaw`.
 
+## Future considerations
+
+`isAuthorizedRaw` uses the Platform's `Cryptography` crypto-engine, available through a
+deprecated-for-removal API (via `CryptographyHolder`).  Additionally, `isAuthorized` (unlike
+`isAuthorizedRaw`) needs to grok non-simple keys (i.e., threshold keys, key lists).  This
+functionality is present in Service's app-spi module but not exported to other modules (directly
+or indirectly from the `HandleContext`).  It needs to be exposed in that way.
+* Currently what is exported w.r.t. signatures is only that the signatures on the top-level 
+transaction have been verified.  Not the ability to verify an arbitrary signature given hash and
+  signature and an account (with keys).
+
 ## Testing
 
 E2E testing necessary:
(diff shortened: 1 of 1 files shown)


hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,62 @@
...
+This note addresses the implementation of `isAuthorized` and `isAuthorizedRaw` of 
+[HIP-632](https://hips.hedera.com/hip/hip-632) - "Hedera Account Service (HAS) 
+System Contract".
+
+## Preliminary notice w.r.t. "virtual addresses"
+
+HIP-632 as written refers a lot to 
+[HIP-631](https://hips.hedera.com/hip/hip-631) - "Account Virtual Addresses".
+However, that HIP, though `Accepted` will not be implemented.  (Or at least, not
+at any time soon.)  So -632 needs to be read with that in mind.  Specifically,
```
Nana-EC, line 12:
```suggestion
at any time soon.)  So HIP-632 needs to be read with that in mind.  Specifically, 
```
```suggestion
at any time soon.)  So -632 needs to be read with that in mind.  Specifically, 
```

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,62 @@
...
+at any time soon.)  So -632 needs to be read with that in mind.  Specifically, 
+anywhere that there's a provision for dealing with "virtual addresses", of which there
+may be several/many per Hedera account, there is in fact only _one_: the _alias_.
+
+* This mainly affects the method `getVirtualAddresses` which will return an array, but
+it will have _at most one_ EVM address in it.
+  
+## Phased implementation
+
+The first release of -631 will include only `isVirtualAddress` and `isAuthorizedRaw`.
```
Nana-EC, line 21: We shouldn't do anything related VirtualAddresses at this point
```suggestion
The first release of HIP-631 will include only `isAuthorizedRaw`.
```

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,62 @@
...
+
+Following the existing pattern for HAS each method will be given a package under 
+`...contract.impl.exec.systemcontracts.has` with two classes:
+* `[method]Translator` - which first confirms that the call is for the given method, and
+  then builds a `TransactionBody` for it.
+* `[method]Call` - which takes that `TransactionBody` and then has an `execute()` method
+  that does the work.
+
+
+## Design for `isVirtualAddress`
```
Nana-EC, line 37: Drop this section.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,88 @@
...
+KECCAK256 version.
+  
+* `isAuthorized{Raw}` does a relatively expensive (resource-wise) computation that should
+be charged for.  Currently we have no method for charging fees (gas or otherwise) for
+system contract calls, not to mention varying the fee for individual system contract
+methods.  (All existing system contract methods are either "cheap" - as in too small to
+charge for - or involve child transactions that themselves do incur fees.) 
+  * TBD #1: A
+mechanism to charge fees for individual system contract methods 
+  * TBD #2: How do we
```
Nana-EC, line 82: matching ecrecover is a good starting point. It works for the EC case but an additional buffer may be needed for the node logic.
We should charge at least the amount of an HCS transaction with a matching key structure that requires validation.


We may consider a hard coded hbar value based on the piping options available to us.

Nana-EC (commented):
Good start in capturing the logic.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: Ah, I wasn't clear.  I can charge a _hard coded_ amount, e.g., the 3000gas that `ECRECOVER` charges.  I can't charge some fee-schedule amount specific to this method.  I'll rewrite this.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,88 @@
...
+`org.hyperledger.besu.evm.precompile.ECRECPrecompiledContract` which will compute
+the `ECRECOVER` operation given the message hash + signature (encoded as a calldata for
+the precompile) (and a `MessageFrame` which it ignores).  This returns either the
+recovered address, if successful, or "empty", if not.  The recovered address must then
+match the address given as an argument to `isAuthorizedRaw`.
+  
+* If an Hedera account then it gets the account, looks in it to find a key _which it must
+have_ and which must be a _single key_ (otherwise: failure).  Then given the key and the
+signature verifies that the signature matches the message hash and is attested by 
+the account. This is an operation provided by `HandleContext` (which can be accessed via
```
lukelee-sl, line 55: A little more detail would be nice (which methods are needed from `HandleContext` and what information do they provide.

lukelee-sl (approved):
LGTM

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,84 @@
...
+system contract calls, not to mention varying the fee for individual system contract
+methods.  (All existing system contract methods are either "cheap" - as in too small to
+charge for - or involve child transactions that themselves do incur fees.)  We _can_
+charge hard-coded fees.
+  * TBD #1: A fee schedule
+mechanism to charge fees for individual system contract methods 
+  * TBD #2: How do we choose the fee for the Hedera signature verification part of
+`isAuthorized` and `isAuthorizedRaw`?
+
+
```
Nana-EC, line 83: Missing E2E test section.
Let's call out some positive and negative scenarios.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,84 @@
...
+  that does the work.
+  
+## Design for `isAuthorizedRaw`
+
+[TBD shortly as I fix up some problems with understanding HAS]
+
+`isAuthorizedRaw` has two flows: Either the argument address is a Hedera account, or it is
+an EVM address.
+
+* If an EVM address then this is simply the `ECRECOVER` precompile.  We have direct
```
Nana-EC, line 43: This may not be the case always right?
You could have an evm address that is a Hedera account id alias i.e. long zero.
In that case it wouldn't use `ecrecover` but would rather use the same flow as we could for an accountId

Nana-EC (commented):
Good improvements.
Left more comments.

Notably still a few TBD's in here.
Let's resolve them or remove them and capture tickets for follow up

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: Will do.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: As written HIP-632 specifically disallows long-zero addresses:

1) `isVirtualAddress` is defined to return `false` for a long-zero
2) When `isVirtualAddress` returns `false` then `isAuthorizedRaw` immediately returns `false`.

Maybe we could do better?

Nana-EC: Yeah, for the mentioned methods that makes sense but for `isAuthorized{Raw}()` given it just checks authorization I don't see any need for it not to work unless there's a concern unmentioned.

david-bakin-sl: Will add this suggestion (+ an update to the HIP) to the epic for the rest of -632.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Where does this go?
+
+A "hedera account service" system contract is being set up in
+`com.hedera.node.app.service.contract.impl.exec.systemcontracts.has`, and also at
+`...contract.impl.exec.processors.HasTranslatorsModule` (for the individual methods).
+
+Following the existing pattern for HAS each method will be given a package under 
+`...contract.impl.exec.systemcontracts.has` with two classes:
+* `[method]Translator` - which first confirms that the call is for the given method, and
```
lukelee-sl, line 32: Technically we are not actually creating a TransactionBody in this case.  We normally do not if the function is a query type that does not change state and thus does not need to dispatch a TransactionBody to another service.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,136 @@
...
+an EVM address.
+
+* If an EVM address then this is simply the `ECRECOVER` precompile.  We have direct
+access to the `ECRECOVER` precompile from BESU in the class
+`org.hyperledger.besu.evm.precompile.ECRECPrecompiledContract` which will compute
+the `ECRECOVER` operation given the message hash + signature (encoded as a calldata for
+the precompile) (and a `MessageFrame` which it ignores).  This returns either the
+recovered address, if successful, or "empty", if not.  The recovered address must then
+match the address given as an argument to `isAuthorizedRaw`.
+  * 3000gas will be charged to match `ECRECOVER` behavior
```
lukelee-sl, line 49: Are we not charging 1.5M gas for the function call?

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
```diff
@@ -0,0 +1,136 @@
...
+`isAuthorizedRaw` has two flows: Either the argument address is a Hedera account, or it is
+an EVM address.
+
+* If an EVM address then this is simply the `ECRECOVER` precompile.  We have direct
+access to the `ECRECOVER` precompile from BESU in the class
+`org.hyperledger.besu.evm.precompile.ECRECPrecompiledContract` which will compute
+the `ECRECOVER` operation given the message hash + signature (encoded as a calldata for
+the precompile) (and a `MessageFrame` which it ignores).  This returns either the
+recovered address, if successful, or "empty", if not.  The recovered address must then
+match the address given as an argument to `isAuthorizedRaw`.
```
lukelee-sl, line 48: It might be important to mention somewhere (perhaps the HIP is a better place for this) that the signature must be in the compressed form for `ecrecover` to succeed.  
I don't know if we want to relax this necessity in the future by implementing our own recover method verses what Besu does?

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: Hmm. Yes.  That's wrong.  I'll fix it ASAP.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: I'll clarify that.

hedera-node/docs/design/services/smart-contract-service/hedera-account-service-isAuthorized-design.md
david-bakin-sl: Wait - I thought that was only when you were doing the signing that in the test code you needed the 64 byte form?

lukelee-sl: Yes the 65 byte form but it should be constructed using the compressed form and attaching the `v` value.  I think if this is what I had to do to get the test to work, anyone trying to use the function will have to do the same?

david-bakin-sl: Not sure.  The (internal) doc that motivated this call showed examples from various sources including Uniswap v2 and v3 where their contract functions that actually called ECRECOVER had passed in `r`, s`, and `v` as separate arguments, plus another example from 0x protocol where they had a datatype `Signature` that had separate `r`, `s`, and `v` fields.  So, based on that data set, having the separate components of the signature available seems common, or, at least, familiar.

lukelee-sl (approved):
LGTM

mhess-swl (approved):
Makes sense from a services perspective 👍

PR #13570, opened 2024-05-29
feat: limit resubmission of transactions

### LINKED ISSUE #13569: Don't resubmit signature transactions forever
We want to resubmit state signature transactions if the host event becomes stale. But eventually if enough rounds pass, it's no longer worth doing so. We don't want an event stuck slightly behind the network to build up an unlimited number of transactions to resubmit.

cody-littley:
closes #13569


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/resubmitter/DefaultTransactionResubmitter.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    @Override
+    @NonNull
+    public List<ConsensusTransactionImpl> resubmitStaleTransactions(@NonNull final GossipEvent event) {
+        if (eventWindow == null) {
+            throw new IllegalStateException("Event window is not set");
+        }
+
+        final List<ConsensusTransactionImpl> transactionsToResubmit = new ArrayList<>();
+        for (final ConsensusTransactionImpl transaction : event.getHashedData().getTransactions()) {
```
kfa-aguda, line 64: This is worth the null check.

cody-littley: This is only showing up as `@Nullable` due to a quirk with how our serialization framework instantiates objects, in practice it will never be null.

As a general rule, we only require null checks for method arguments, and then only if failing to do a null check may result in an error thrown at a much later time. If the method will throw a nullpointer exception without a null check being performed, then the check doesn't buy us much. The important thing is to throw when there is a problem, as opposed to storing a null in a data structure and being surprised at a much later time when it finally causes a problem.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/resubmitter/DefaultTransactionResubmitter.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private EventWindow eventWindow;
+    private final long maxSignatureResubmitAge;
+
+    private final TransactionResubmitterMetrics metrics;
+
+    /**
+     * Constructor.
+     */
+    public DefaultTransactionResubmitter(@NonNull final PlatformContext platformContext) {
```
alittley, line 44: missing `platformContext` doc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/StateConfig.java
```diff
@@ -77,6 +77,9 @@
  * @param validateInitialState          If false then do not do ISS validation on the state loaded from disk at startup.
  *                                      This should always be enabled in production environments. Disabling initial
  *                                      state validation is intended to be a test-only feature.
+ * @param maxSignatureResubmitAge       The maximum age of a signature that can be resubmitted (if the host event
```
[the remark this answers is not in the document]
cody-littley: updated javadoc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/resubmitter/DefaultTransactionResubmitter.java
cody-littley: added

alittley (approved):

PR #13575, opened 2024-05-29
chore: add check for non extant system account

lukelee-sl:
**Description**:
Handle calls to non extant system account gracefully.  
This change should allow ContractCallSuite#idVariantsTreatedAsExpected to pass.


david-bakin-sl (approved):
LGTM but is this actually a "chore"? (not a "fix"?)

PR #13578, opened 2024-05-29
fix: skip storage link repair with more than 4M slots in state

tinker-michaelj:
**Description**:
 - The `V0500ContractSchema` migration is impractical with very large state sizes; skip it in those cases since it does not have any short-term functional implications.
     * Current mainnet states use ~2M slots, so selected 4M as a very conservative cutoff for the `v0.50` release.
 - Stabilizes `CongestionPricingSuite` by increasing gas offered for contract calls to guarantee congestion limits are reached.


tomzhenghedera (approved):

tomzhenghedera (approved):

david-bakin-sl (approved):
LGTM.  

Guess since we don't need it for mainnet, it's fine.  If we ever _do_ need to do something like this we can use external memory (e.g., a file or database (e.g., sqllite), if we plan for it.

mhess-swl (approved):

PR #13579, opened 2024-05-29
feat: remove context from platform builder

### LINKED ISSUE #13572: Remove platform context from platform builder API
We eventually want to remove the `PlatformContext` from the public API of the platform.

As an incremental step in that direction, remove `PlatformContext` from the public API of the platform builder. It should be possible to configure all base utilities without using a platform context to do so.

cody-littley:
closes #13572


alittley (approved):

tinker-michaelj (approved):
LGTM, ty @cody-littley !

PR #13535, opened 2024-05-26
fix: add v0.50 migrations; clean up schema mgmt

tinker-michaelj:
**Description**:
 - Closes #13437 by:
     1. Extracting all `Schema`s into top-level classes with internally controlled versions, deleting the `SemanticVersion` argument in `Service#registerSchemas()`. (ℹ️ This is ofc mechanical but touches many files.)
     2. Refactoring schema use analysis from `MerkleSchemaRegistry` into a separate `SchemaUseAnalysis`.
        * This refactor clarifies that a schema can be used for three purposes: `STATE_DEFINITIONS`, `MIGRATION`, and `RESTART`; and documents the conditions for each in javadoc.
 - Closes https://github.com/hashgraph/hedera-services/issues/13537
 - Closes #13538 by:
     1. Registering new `MIGRATION`-only `V0500ContractSchema` and `V0500TokenSchema` schemas to repair storage linked lists and first storage keys, respectively.
     2. Adding a `Map<String, Object> sharedValues` to the `MigrationContext` to facilitate passing information about the correct first keys from the `ContractService` to the `TokenService`.
     3. Ordering `Service` registrations by a new `Service#migrationOrder()` ordinal.

PR #13595, opened 2024-05-30
fix: add migrations to repair storage links

tinker-michaelj:
**Description**:
 - Cherry-pick necessary parts of #13535 
    * The extra cleanup to make all `Schema`s top-level classes, remove the `version` argument from `Service.registerSchemas()`, and extract mono migration from `Hedera`---while important tech debt to pay down---does not need to be in `release/0.50`.


Neeharika-Sompalli (dismissed):

lukelee-sl (approved):
LGTM

PR #13608, opened 2024-05-30
feat: no status nexus

### LINKED ISSUE #13607: Remove PlatformStatusNexus
There should not be a "nexus" object for the platform status. The platform status should be communicated via wires.

cody-littley:
closes #13607


platform-sdk/sdk/settings.txt
```diff
@@ -44,5 +44,4 @@ prometheus.endpointEnabled,               false
 
 loadKeysFromPfxFiles,                     false
 crypto.enableNewKeyStoreModel,            false
-crypto.cpuVerifierThreadRatio,            0.1
```
alittley, line 47: did you intend to include this change in this PR?

alittley (approved):

platform-sdk/sdk/settings.txt
cody-littley: Yes. It is not currently used. I removed it from the config record in another PR, but forgot to update the `settings.txt` we use for local testing.

PR #13619, opened 2024-05-31
Remove "proto." prefix from metric names

### LINKED ISSUE #13617: Remove proto prefix from metric names
Remove the proto part from metrics generated in the gRPC layer, e.g., `app_proto:CryptoService:createAccountHdl`. It does not add value and is confusing because it is easily seen as part of the category.

netopyr:
**Description**:

This PR removes the "proto." prefix from metric names generated for all gRPC endpoints. The prefix is part of the generated `servicesName` but can be confusing when appearing in the metric names.

**Related issue(s)**:

Fixes #13617


Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr

Issue #13375, opened 2024-05-17
HIP-904: Adapt CryptoTransferHandler's fee calculation to ensure that the sender is covering the fees for auto-associations


mustafauzunn:
- [x] release/0.52 - https://github.com/hashgraph/hedera-services/pull/14123
- [x] develop - https://github.com/hashgraph/hedera-services/pull/14107

PR #13620, opened 2024-05-31
feat: HIP-904: Sender should cover fees for auto associations

bilyana-gospodinova:
**Description**:
With this PR the fee calculations on crypto transfer are adapted according to HIP-904. Now, the sender covers the fees for the auto associations if the `unlimitedAutoAssociations` is enabled.

Fixes https://github.com/hashgraph/hedera-services/issues/13375

TODO: The tests need to be adapted.


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -616,10 +629,100 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+                .flatMap(tt -> Stream.concat(
+                        tt.transfers().stream().filter(aa -> aa.amount() > 0).map(AccountAmount::accountID),
+                        tt.nftTransfers().stream().map(NftTransfer::receiverAccountID)
+                ))
+                .forEach(recipientsSet::add);
+        return recipientsSet;
+    }
+
+    /**
+     * This method iterates through all token transfers (for fungible tokens an NFTs) and creates a map
```
kselveliev, line 710: nit: fungible tokens an**d** NFTs

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -616,10 +629,100 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+        final Set<AccountID> recipientsSet = getTransferRecipients(tokenTransfers);
+        final HashMap<AccountID, HashSet<TokenID>> associationsMap = getTokenTransferAssociations(tokenTransfers);
+
+        var associationsRbs = 0;
+        for (var accountId : recipientsSet) {
+            var account = accountStore.getAccountById(accountId);
+            // If this is a transfer to an alias and the account is null this is a hollow account creation case,
+            // and we need to set default values.
+            if (account == null && accountId.hasAlias()) {
+                var numAutoAssociations = associationsMap.get(accountId).size();
```
kselveliev, line 660: One suggestion here for better readability here might be:
```
if(isHollow(account)) {
     accociationsRbs += calculateHollowAccountRbs(params)
} else {
     accociationsRbs += calculateAccountRbs(params)
} 
```

and then in those 2 new methods extracts the calculation logics.
This could improve code readability.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -616,10 +629,100 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+        // If the token transfers collection is empty, then we have only hbars transfer and no token
+        // associations are needed, so we do not need to calculate the next part.
+        if (tokenTransfers.isEmpty()) {
+            return 0;
+        }
+
+        final var accountStore = feeContext.readableStore(ReadableAccountStore.class);
+        final var tokenRelStore = feeContext.readableStore(ReadableTokenRelationStore.class);
+
+        final Set<AccountID> recipientsSet = getTransferRecipients(tokenTransfers);
```
Neeharika-Sompalli, line 651: If all we are doing is finding if the token relation doesn't exist, we should be able to re-use some functionality from `validateAndBuildAutoAssociation` and count number of new associations that will be formed with this list. 

And then we need to do is add the rbs value in fees. 
```
numAutoAssociations * THREE_MONTHS_IN_SECONDS * CREATE_SLOT_MULTIPLIER;
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -616,10 +629,100 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+                    }
+                    // If there is no association, we need to compute the fee to create it.
+                    // Case 1: maxAutoAssociations == -1 -> unlimited auto associations, just increase newAutoAssociations counter
+                    // Case 2: maxAutoAssociations == 0 -> no auto associations can be made
+                    // Case 3: maxAutoAssociations > 0 -> we need to check if there are unused slots
+                    if (account.maxAutoAssociations() >= 0) {
+                        validateFalse(
+                                account.usedAutoAssociations() + 1 > account.maxAutoAssociations(),
+                                NO_REMAINING_AUTOMATIC_ASSOCIATIONS);
+                    }
```
Neeharika-Sompalli, line 679: Why should we check this ? If unlimitedAutoAssociationsIdEnabled, we should just charge for new associations right ?

Neeharika-Sompalli, line 679: We should not throw in `calculateFees`. We should compute fees and this validation should be in `handle`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
bilyana-gospodinova: Yes, we need to find the new associations that need to be created. I am not sure if we can reuse even parts of the implementation in `validateAndBuildAutoAssociation` because `calculateFees` is called before `handle` and in `handle` we already have the hollow accounts that are part of the transfers but in `calculateFees` they are not created yet. Because of that the accountId is still null at that point and `tokenRelStore.get(accountId, tokenId)` will throw an exception.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
bilyana-gospodinova: I will move the validation. Thank you.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
bilyana-gospodinova: Fixed

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
bilyana-gospodinova: Comment addressed partially. I did not use `isHollow` because if I pass just the account it can be null and I can't check for its alias. If I pass just the accountId then I need to fetch the account from the store twice and it seems redundant. To create an `isHollow` method I need to pass both the account and the accountId and this would seem strange to me in terms of readability.

hedera-node/hapi-fees/src/main/java/com/hedera/node/app/hapi/fees/usage/crypto/CryptoOpsUsage.java
```diff
@@ -86,6 +87,9 @@ public void cryptoTransferUsage(
         incRb += TOKEN_ENTITY_SIZES.bytesUsedToRecordTokenTransfers(
                 weightedTokensInvolved, weightedTokenXfers, xferMeta.getNumNftOwnershipChanges());
         accumulator.addRbs(incRb * USAGE_PROPERTIES.legacyReceiptStorageSecs());
+
+        // Add rbs for auto associations
+        accumulator.addRbs((xferMeta.getNumFungibleTokenTransfers() + xferMeta.getNumNftOwnershipChanges()) * THREE_MONTHS_IN_SECONDS * CREATE_SLOT_MULTIPLIER);
```
bilyana-gospodinova, line 92: I added some default values here as we don't have the actual number of new associations in the current transfer.

tinker-michaelj (approved):
This looks good for now; post merging [this PR](https://github.com/hashgraph/hedera-services/pull/13639) it will likely become simplest to dispatch the auto-association with the desired account as the "synthetic" payer. Tyvm @bilyana-gospodinova !

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -616,10 +628,110 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
         return feeContext
                 .feeCalculator(subType)
                 .addBytesPerTransaction(bpt)
+                .addRamByteSeconds(associationsRbs)
```
MrValioBg, line 631: Before, the logic for paying association fees was in CryptoCreateHandler the price per association slot, was about 0.0018 when this was paid during a CryptoCreate.

Now when moving it here and adding it to RamByteSeconds, it could introduce changes in the price per slot, as RamBytesSeconds are cheaper for transfers, as defined in the feeSchedules.

We estimated that instead of paying 0.0018, a slot could cost the users about 0.00005.
Is this change okay? Or should we think of some MULTIPLIER to get closer to the price per slot when it was paid during CryptoCreate.

What are your opinions @Neeharika-Sompalli @tinker-michaelj

PR #13637, opened 2024-06-01
feat: declarative `@HapiTest` entity creation

### LINKED ISSUE #13565: Add declarative entity creation to `@HapiTest`s
### Problem

Our existing `@HapiTest`s suffer from a huge amount of boilerplate devoted to creating entities.

This at minimum distracts from the main purpose of the test; and in many cases, obscures it almost entirely.

### Solution

Use the JUnit Jupiter extension APIs to support declarative entity creation in three ways:
- [ ] Via `static` field injection (`BeforeAllCallback` API)
- [ ] Via instance field injection (`BeforeEachCallback` API)
- [ ] Via `@Test` method parameter injection (`ParameterResolver` API)

This will require a few steps.
  - [ ] Defining Java types that represent entities that should be pre-created before a `@HapiTest` runs.
  - [ ] Using JUnit `AnnotationSupport` to inject instances of these types into test classes and instances.
  - [ ] Using a `ParameterResolver` to inject instances into `@Test` methods.
  - [ ] Having an appropriate extension that will bind a `List<SpecEntity>` of all declarative dependencies to the thread before invoking a `HapiSpec` factory.
  - [ ] Updating `HapiSpec.run()` to ensure all required entities are created on the target network and "linked" to its registry before executing the user's operations.
  - [ ] Adding `@HapiTest` DSL that works in terms of Java instances instead of `HapiSpecRegistry` names.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13565 
 - Please focus review on the differences between [`TokenUpdatePrecompileSuite`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenUpdatePrecompileSuite.java#L102) on `develop` and [`UpdateTokenPrecompileTest`](https://github.com/hashgraph/hedera-services/blob/13565-declarative-hapi-test-setup/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/token/UpdateTokenPrecompileTest.java#L71)  + [`GetTokenKeyPrecompileTest`](https://github.com/hashgraph/hedera-services/blob/13565-declarative-hapi-test-setup/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/token/GetTokenKeyPrecompileTest.java) on this branch.
     * Seven substantiative `@HapiTest`'s are split into 28 `@HapiTest`'s of no more than 7 operations (mostly 2 or 3).
     * Despite 4x increase in number of test and use of extra `@DisplayName` and `@Nested` annotations, line count drops from ~1500 lines to ~500 lines.


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

jeromy-cannon (approved):

lukelee-sl (approved):
LGTM

PR #13639, opened 2024-06-03
chore: `HandleWorkflow` refactor

### LINKED ISSUE #13638: Refactor and simplify HandleWorkflow and HandleContext to use Dagger components
- [x] Currently there is code from HandleWorkflowImpl that is duplicated in HandleContextImpl. Remove the duplication and re-use code
- [x] Refactor `HandleWorkflow` to be simple and create classes for different responsibilities

Neeharika-Sompalli:
Fixes #13638 

Sub PRs:
- Done in this PR https://github.com/hashgraph/hedera-services/pull/13771
- https://github.com/hashgraph/hedera-services/pull/13780
- https://github.com/hashgraph/hedera-services/pull/13804
- In this PR: Refactored to use `DispatchHandleContext` and deleted `HandleContextImpl`. Refactored `HandleWorkflow` 

**Two intentional Changes from mono-service behavior:**'

- The records for auto-creation, the child record will contain the auto creation fee. This will give much more clarity, see [HIP-993](https://github.com/hashgraph/hedera-improvement-proposal/pull/993).
- We don't fail the user transaction when there are too many hollow account finalizations. We only finalize how many ever possible in the transaction.


tinker-michaelj (approved):

rbair23:
It seems I always start my PR reviews a few minutes before the merge happens, doh. I'm providing a lot of comments on the PR, even though it has already been merged.

On the whole, I do not like this PR and I think we should revisit the aims and the effect. I see we only had a single confirmed review for such an essential part of the system. I'm lacking now an understanding of the system architecture, as there are many new classes introduced, and I don't see any kind of corresponding design documentation.

There is a tradeoff in system design between fewer components that do more, and more components that do less. With many smaller components, each individual component is very simple, but the overall system becomes harder to understand and begins to have surprising emergent properties, and becomes harder to get good performance. Abstraction should be _just enough_ but no more.

This is true when "components" are modules, or "components" are packages, or "components" are classes, or "components" are methods. In all cases, excessive abstraction is the enemy of clarity and performance.

There are tactical things wrong with this PR, but I question whether this effort was time well spent and whether we have made things more complicated and difficult to understand. I will provide more direct observations.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -545,12 +551,22 @@ public void onPreHandle(@NonNull final Event event, @NonNull final HederaState s
...
+        try {
+            if (shouldDump(daggerApp.initTrigger(), MOD_POST_EVENT_STREAM_REPLAY)) {
+                dumpModChildrenFrom(recoveredState, MOD_POST_EVENT_STREAM_REPLAY, MerkleStateChild.childrenToDump());
+            }
+        } catch (Exception e) {
+            logger.error("Error dumping state after migration at MOD_POST_EVENT_STREAM_REPLAY", e);
+        }
+        // Always close the block manager so replay will end with a complete record file
         daggerApp.blockRecordManager().close();
     }
```
rbair23, line 564: What is the tradeoff here? What does it mean to dump state? What is a DumpCheckpoint? What is the effect on this for restart, upgrade, etc? Where is this design documented? I always thought the state dumpers were temporary for just the mono->mod migration. Are they to be kept?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/ThrottleException.java
```diff
@@ -0,0 +1,42 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.hapi.node.base.ResponseCodeEnum;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * This class is used to throw a {@link ThrottleException} when a transaction is gas throttled.
+ */
+public class ThrottleException extends Exception {
```
rbair23, line 27: Exception throwing can be expensive (depending on whether stack traces are being filled in or not). Normally you would not use exceptions for control flow. In this particular case, I am not sure we can say that hitting a throttle is an exceptional situation -- it can happen quite easily on the backend because we cannot do gas filtering on ingest. I haven't looked into the call sites for this, but I am concerned over using an exception for "when a transaction is gas throttled".

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
```diff
@@ -1202,4 +1202,37 @@ public EthereumTransactionRecordBuilder feeChargedToPayer(@NonNull long amount)
     public List<AccountAmount> getPaidStakingRewards() {
         return paidStakingRewards;
     }
+
+    @Override
+    public String toString() {
```
rbair23, line 1207: Is there any risk of this being called in production? What is the use for this toString() method? We should be very careful with introducing toString methods, because we have found real performance problems if these get invoked (sometimes generating excessive logs). The problem is usually that someone adds a simple `"" + someObject` type line and then we find excessive time spent constructing strings.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -80,8 +78,21 @@ enum TransactionCategory {
...
+         */
         UNLIMITED_CHILD_RECORDS,
+        /**
+         * The number of preceding transactions is limited by the number of nanoseconds between the
+         * last-assigned consensus time and the current platform-assigned consensus time. (Or,
+         * before block streams, even further artificially limited to three records for
+         * backward compatibility.)
+         */
         LIMITED_CHILD_RECORDS
     }
```
rbair23, line 97: I love all the new comments that were added, here and elsewhere in the PR 👍🏻

rbair23:
> It seems I always start my PR reviews a few minutes before the merge happens, doh. I'm providing a lot of comments on the PR, even though it has already been merged.
> 
> On the whole, I do not like this PR and I think we should revisit the aims and the effect. I see we only had a single confirmed review for such an essential part of the system. I'm lacking now an understanding of the system architecture, as there are many new classes introduced, and I don't see any kind of corresponding design documentation.
> 
> There is a tradeoff in system design between fewer components that do more, and more components that do less. With many smaller components, each individual component is very simple, but the overall system becomes harder to understand and begins to have surprising emergent properties, and becomes harder to get good performance. Abstraction should be _just enough_ but no more.
> 
> This is true when "components" are modules, or "components" are packages, or "components" are classes, or "components" are methods. In all cases, excessive abstraction is the enemy of clarity and performance.
> 
> There are tactical things wrong with this PR, but I question whether this effort was time well spent and whether we have made things more complicated and difficult to understand. I will provide more direct observations.

Reviewing the PR, it seems that most of the additional complication did not come from this PR, but from some previous ones that happened between the time I was last involved in the code. I think it would be helpful to have a design doc that describes how all this works now. At the beginning of this project I had produced diagrams and design documentation, with the goal that the diagrams would be detailed enough so that you could easily go from the diagram to the code and back again. There is a lot of complexity that crept in over time, so it is probably worth some discussions on what the truth is now, and either get the code to match the diagrams or the diagrams to match the code or (most likely) some combination of things. I don't have an accurate mental model anymore of how this works.

Neeharika-Sompalli:
> > It seems I always start my PR reviews a few minutes before the merge happens, doh. I'm providing a lot of comments on the PR, even though it has already been merged.
> > On the whole, I do not like this PR and I think we should revisit the aims and the effect. I see we only had a single confirmed review for such an essential part of the system. I'm lacking now an understanding of the system architecture, as there are many new classes introduced, and I don't see any kind of corresponding design documentation.
> > There is a tradeoff in system design between fewer components that do more, and more components that do less. With many smaller components, each individual component is very simple, but the overall system becomes harder to understand and begins to have surprising emergent properties, and becomes harder to get good performance. Abstraction should be _just enough_ but no more.
> > This is true when "components" are modules, or "components" are packages, or "components" are classes, or "components" are methods. In all cases, excessive abstraction is the enemy of clarity and performance.
> > There are tactical things wrong with this PR, but I question whether this effort was time well spent and whether we have made things more complicated and difficult to understand. I will provide more direct observations.
> 
> Reviewing the PR, it seems that most of the additional complication did not come from this PR, but from some previous ones that happened between the time I was last involved in the code. I think it would be helpful to have a design doc that describes how all this works now. At the beginning of this project I had produced diagrams and design documentation, with the goal that the diagrams would be detailed enough so that you could easily go from the diagram to the code and back again. There is a lot of complexity that crept in over time, so it is probably worth some discussions on what the truth is now, and either get the code to match the diagrams or the diagrams to match the code or (most likely) some combination of things. I don't have an accurate mental model anymore of how this works.

I will be opening a PR with design document and diagram for just this change and will add you as reviewer. This PR did not change the flow or logic that was in `HandleWorkflow` except the two intentional changes mentioned [here](https://github.com/hashgraph/hedera-services/pull/13639#issuecomment-2173858502). @tinker-michaelj  opened HIP PR for those changes [here](https://github.com/hashgraph/hedera-improvement-proposal/pull/993).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
Neeharika-Sompalli: We wanted to still keep StateDumper code, so we can use them when needed. For example, with this change we validated the record streams and the state is the same for a four 30 min intervals and validated with `release/0.49` branch, to make sure nothing is broken from the initial mod-service code.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
Neeharika-Sompalli: yes, will delete this.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/ThrottleException.java
Neeharika-Sompalli: We had to distinguish the kind of `HandleException` that was happening [here](https://github.com/hashgraph/hedera-services/blob/release/0.51/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java#L594) to charge different fees. In the new code is caught [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java#L154) . Do you recommend any other approach?

tinker-michaelj:
> > It seems I always start my PR reviews a few minutes before the merge happens, doh. I'm providing a lot of comments on the PR, even though it has already been merged.
> > On the whole, I do not like this PR and I think we should revisit the aims and the effect. I see we only had a single confirmed review for such an essential part of the system. I'm lacking now an understanding of the system architecture, as there are many new classes introduced, and I don't see any kind of corresponding design documentation.
> > There is a tradeoff in system design between fewer components that do more, and more components that do less. With many smaller components, each individual component is very simple, but the overall system becomes harder to understand and begins to have surprising emergent properties, and becomes harder to get good performance. Abstraction should be _just enough_ but no more.
> > This is true when "components" are modules, or "components" are packages, or "components" are classes, or "components" are methods. In all cases, excessive abstraction is the enemy of clarity and performance.
> > There are tactical things wrong with this PR, but I question whether this effort was time well spent and whether we have made things more complicated and difficult to understand. I will provide more direct observations.
> 
> Reviewing the PR, it seems that most of the additional complication did not come from this PR, but from some previous ones that happened between the time I was last involved in the code. I think it would be helpful to have a design doc that describes how all this works now. At the beginning of this project I had produced diagrams and design documentation, with the goal that the diagrams would be detailed enough so that you could easily go from the diagram to the code and back again. There is a lot of complexity that crept in over time, so it is probably worth some discussions on what the truth is now, and either get the code to match the diagrams or the diagrams to match the code or (most likely) some combination of things. I don't have an accurate mental model anymore of how this works.

Note that despite adding 99% unit test coverage and isolating multiple responsibilities in standalone classes, this PR eliminated so much duplicate code between `HandleWorkflow` and `HandleContextImpl` that it still reduced line count.**(!)** 

Functional behavior is identical except for the auto-creation records documented in [HIP-993](https://github.com/hashgraph/hedera-improvement-proposal/pull/993); we done replay diff testing between `release/0.50` and this branch to validate this.

Neeharika-Sompalli:
@rbair23  The design document for the `HandleWorkflow` refactored is https://github.com/hashgraph/hedera-services/pull/13923. Thank you!

PR #13641, opened 2024-06-03
feat(HIP796): Implement HIP-796 moveNftsBetweenSameUserPartitions functionality

vvairava:
Description

User wants to  move non-fungible tokens from one partition to another partition on the same user account, without requiring a signature from the user

Existing Files

Compile
./gradlew assemble is SUCCESS


thomas-swirlds-labs (approved):
LGTM. Thanks @vvairava 👍

PR #13645, opened 2024-06-03
build: activate more Javadoc lint checks and fail on problems

jjohannes:
**Description**:

- Activates Javadoc lint checks (except _missing_) and fail on problems as part of `qualityGate`

Related:
- #14337 
- #14338
- #14339

**Related issue(s)**:

#11569

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
@@ -67,7 +67,8 @@ jvmDependencyConflicts.patch {
     module("com.google.guava:guava") {
         (annotationLibraries -
                 "com.google.code.findbugs:jsr305" -
-                "com.google.errorprone:error_prone_annotations")
+                "com.google.errorprone:error_prone_annotations" -
+                "org.checkerframework:checker-qual")
             .forEach { removeDependency(it) }
     }
     module("com.google.protobuf:protobuf-java-util") {
@@ -136,6 +137,7 @@ extraJavaModuleInfo {
     module("org.apache.commons:commons-math3", "org.apache.commons.math3")
     module("org.apache.commons:commons-collections4", "org.apache.commons.collections4")
     module("com.esaulpaugh:headlong", "headlong")
+    module("org.checkerframework:checker-qual", "org.checkerframework.checker.qual")
     module("org.connid:framework", "org.connid.framework")
     module("org.connid:framework-internal", "org.connid.framework.internal") {
         exportAllPackages()
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
@@ -214,6 +214,7 @@ tasks.withType<Javadoc>().configureEach {
         )
         options.windowTitle = "Hedera Consensus Node"
         options.memberLevel = JavadocMemberLevel.PACKAGE
+        addStringOption("Xdoclint:all,-missing", "-Xwerror")
     }
 }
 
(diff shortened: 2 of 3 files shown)


platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/io/OutputStreamFactory.java
jsync-swirlds, line 1: None of the changes in this file appear to be necessary or related to javadoc...

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/spi/WritableQueueStateBase.java
```diff
@@ -84,9 +84,7 @@ public final void commit() {
     }
 
     /**
-     * {@inheritDoc}
-     *
-     * <p>Clears the set of peeked values and added values.
+     * Clears the set of peeked values and added values.
```
jsync-swirlds, line 87: This does not appear to be necessary, and it changes the intended documentation.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/BaseEventHashedData.java
```diff
@@ -330,6 +330,7 @@ public long getSelfParentGen() {
      * @return the maximum generation of the other parents
      * @deprecated this method should be replaced since there can be multiple other parents.
      */
+    @Deprecated
```
jsync-swirlds, line 333: If this is required to pass, we may be a bit too strict enforcing recommendations.
There can be legitimate reasons to mark methods deprecated in javadoc without also annotating them...

jsync-swirlds (commented):
Just a few items.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/spi/WritableQueueStateBase.java
jjohannes: The method does not override another method, hence `{@inheritDoc}` is not allowed.

platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/io/OutputStreamFactory.java
jjohannes: These two methods are deprecated. It is a limitation of the `javadoc` tool. It complains about importing deprecated methods directly and there is no way to suppress that on an import. Since this is the only place we run into this, I figured it's a small price to pay to do this change in order to get rid of the warnings.

jsync-swirlds: That's unfortunate.  I suspect we'll encounter this in other places; a consequence of perhaps being a little too sensitive in the build process.

timo0 (dismissed):

cody-littley (dismissed):

gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
```diff
@@ -67,7 +67,8 @@ jvmDependencyConflicts.patch {
     module("com.google.guava:guava") {
         (annotationLibraries -
                 "com.google.code.findbugs:jsr305" -
-                "com.google.errorprone:error_prone_annotations")
+                "com.google.errorprone:error_prone_annotations" -
+                "org.checkerframework:checker-qual")
```
jjohannes, line 71: Turns out we need the annotations from `checker-qual` on the classpath for Javadoc. Therefore, we no longer remove them from the transitive dependencies of Guava.

rbarkerSL (approved):

PR #13652, opened 2024-06-04
feat: Reduce services-hedera's usage of PlatformContext

### LINKED ISSUE #13612: Reduce services-hedera's usage of PlatformContext
The goal is to eliminate this altogether, but this PR starts with minimizing it.

Many of the building blocks/base utilities classes used by `services-hedera` today are retrieved from the PlatformContext. 

1. We should provide these building blocks/base utilities provided by PlatformContext so service-hedera can access them without going to platform.getContext().
2. Make these utilities visible to Dagger so they can be easily injected by dependent classes

In this first PR, we don't have to physically eliminate the calls to platform.getContext() itself, but make the utilities in it visible and switch as many instances as realistically possible.

kfa-aguda:
This PR explicitly constructs some classes required in constructing the PlatformContext (from outside of it).

Many of the building blocks/base utilities classes used by services-hedera today are retrieved from the PlatformContext. Some of these are 'automatically' created. This could be problematic for a few reason, one being testing. It's desirable to have full control of as many platform building blocks as possible. Another is that services code uses Dagger for lifecycle management. It's useful to be able to pass these building block instances to dagger so subsequent classes that need them can simply ask for them (via DI). The goal is to begin providing these building blocks/base utilities to the PlatformBuilder so service-hedera can:

1. pass them on to Dagger for lifecycle management
2. Access them without going to platform.getContext() i.e. via dependency injection

This PR is the first cut of point 1 above.

In this first PR, we have not tried to physically eliminate the calls to platform.getContext() itself, but make the utilities in it visible and switch as many instances as realistically possible later.

Fixes #13612


Neeharika-Sompalli (approved):
LGTM Thanks @kfa-aguda ! But there are merge conflicts to be resolved

Neeharika-Sompalli (dismissed):
LGTM !

Neeharika-Sompalli (approved):
LGTM !

PR #13657, opened 2024-06-04
fix: Use the correct PRNG contract address

### LINKED ISSUE #10552: Use the correct PRNG contract address in child records
When we call a PrngContract, we save the record with the HTC contract address. This is done in mod so we can mirror mono, but we should fix it after the differential testing.

Full context: https://github.com/hashgraph/hedera-services/pull/10074#discussion_r1409208043

JivkoKelchev:
**Description**:

Set correct PRNG address

**Related issue(s)**:

Fixes #10552

**Notes for reviewer**:

**Checklist**


lukelee-sl (approved):
LGTM. thanks

PR #13685, opened 2024-06-05
feat: HIP-904 Implement TokenReject functionality

MrValioBg:
**Description**:
Implement 

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,380 @@
...
+        context.requireKey(key);
+    }
+
+    @SuppressWarnings("java:S2259")
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn, "Transaction body cannot be null");
+        final var op = txn.tokenRejectOrThrow();
+
+        validateFalsePreCheck(op.rejections().isEmpty(), INVALID_TRANSACTION_BODY);
```
Neeharika-Sompalli, line 144: We could probably add `EMPTY_TOKEN_REJECTIONS` response code

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,380 @@
...
+
+        final var hederaConfig = context.configuration().getConfigData(HederaConfig.class);
+        final var ledgerConfig = context.configuration().getConfigData(LedgerConfig.class);
+        final var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
+        validateTrue(tokensConfig.tokenRejectEnabled(), NOT_SUPPORTED);
+        validateTrue(rejections.size() <= ledgerConfig.tokenRejectsMaxLen(), TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED);
+
+        final var accountStore = context.writableStore(WritableAccountStore.class);
+        final var rejectingAccount = getIfUsableForAliasedId(
+                rejectingAccountID, accountStore, context.expiryValidator(), INVALID_ACCOUNT_ID);
```
Neeharika-Sompalli, line 184: @Nana-EC  Are we expecting to allow aliases in `TokenReject` transaction? If so, we should let mirror node know that aliases are allowed here.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,380 @@
...
+        requireNonNull(txn, "Transaction body cannot be null");
+        final var op = txn.tokenRejectOrThrow();
+
+        validateFalsePreCheck(op.rejections().isEmpty(), INVALID_TRANSACTION_BODY);
+        if (op.hasOwner()) {
+            validateAccountID(op.owner(), null);
+        }
+
+        final var uniqueTokenReferences = new HashSet<TokenReference>();
+        for (final var rejection : op.rejections()) {
```
Neeharika-Sompalli, line 150: Should we also have a check we reject if each rejection has more than 2 account amount transfers ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,380 @@
...
+        final var config = feeContext.configuration();
+        final var tokenMultiplier =
+                requireNonNull(config).getConfigData(FeesConfig.class).tokenTransferUsageMultiplier();
+
+        final int totalRejections = op.rejections().size();
+        int numOfFungibleTokenRejections = 0;
+        int numOfNFTRejections = 0;
+        for (final var rejection : op.rejections()) {
+            if (rejection.hasFungibleToken()) {
+                // Each fungible token rejection involves 2 AccountAmount transfers
```
Neeharika-Sompalli, line 350: Do we have a check anywhere to make sure each token. rejection has 2 AccountAmount transfers ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+
+        final int weightedTokensInvolved = tokenMultiplier * totalRejections;
+        final int weightedFungibleTokens = tokenMultiplier * numOfFungibleTokenRejections;
+        final long bpt =
+                calculateBytesPerTransaction(weightedTokensInvolved, weightedFungibleTokens, numOfNFTRejections);
+        final long rbs = USAGE_PROPERTIES.legacyReceiptStorageSecs() * bpt;
+
+        return feeContext
```
Neeharika-Sompalli, line 364: Makes sense to me. 💯  Seems exactly same as `CryptoTransfer`

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenRejectSuite.java
```diff
@@ -0,0 +1,556 @@
...
+                                .hasNoSpender(),
+                        getTokenNftInfo(NON_FUNGIBLE_TOKEN_B, 1L)
+                                .hasAccountID(TOKEN_TREASURY)
+                                .hasNoSpender(),
+                        getAccountBalance(TOKEN_TREASURY).logged().hasTokenBalance(FUNGIBLE_TOKEN_A, 0L),
+                        getAccountBalance(TOKEN_TREASURY).logged().hasTokenBalance(FUNGIBLE_TOKEN_B, 750L));
+    }
+
+    @HapiTest
+    final Stream<DynamicTest> tokenRejectFailsWithInvalidBodyInputsScenarios() {
```
Neeharika-Sompalli, line 488: We should probably add a test that validates fees

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Hey! 
Our only use case is:
Bob rejects token A he owns and the treasury account Carol receives the whole token A balance of Bob.

And since:
1 TokenTransfer = 1 TokenReference.
1 Token Transfer body contains List<AccountAmounts>

We always have exactly 2 AccountAmounts, because we have 2 different accounts that exchange tokens.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Hey! 
Our only use case is:
Bob rejects token A he owns and the treasury account Carol receives the whole token A balance of Bob.

And since:
1 TokenTransfer = 1 TokenReference.
1 Token Transfer body contains List<AccountAmounts>

We always have exactly 2 AccountAmounts, because we have 2 different accounts that exchange tokens.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/TokenRejectSuite.java
MrValioBg: Done. I've created test that validates the fees and compares them to cryptoTransfer fees.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Okay. Will add it to the protobufs and include it :)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
Neeharika-Sompalli: @MrValioBg Since user can submit any invalid transactions, do we reject any transaction that is not like the valid use case ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
Neeharika-Sompalli: @MrValioBg Since user can submit any invalid transactions, do we reject any transaction that is not like the valid use case ?

MrValioBg: Users can only specify the token they want to reject as a TokenReference. We have checks if they actually own it. If they don't have a balance of it, the TokenReject fails.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Users can only specify the token they want to reject as a TokenReference. ( They don't specify treasury accounts, we retrieve that)
We have checks if they actually own it. If they don't have a balance of it, the TokenReject fails.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: ( They don't specify treasury accounts, we find which is the treasury account)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
Nana-EC: yes we should.
It's always been a long term vision that aliases be used anywhere accountIds could be within reasons.
If there is a reason to hold back on aliases we should note it, or if we need to state this support we should also capture it.
Otherwise yeah we should inform mirror explicitly of this possibility.

MrValioBg: The TokenReject functionality supports aliases, indeed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
MrValioBg: Done, I've added it. :)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
edwin-greene: Thanks, the mirror node will then also support aliases.

Neeharika-Sompalli (dismissed):
LGTM! Thanks @MrValioBg

Neeharika-Sompalli (approved):
LGTM! Thanks @MrValioBg

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
```diff
@@ -251,6 +255,21 @@ public enum TokenValidations {
...
 
+    /**
+     * Enum to determine the type of validations to be performed on the token account relationship. If the token account
+     * relationship is allowed to be frozen or not
+     */
+    public enum TokenRelValidations {
+        /**
+         * Token account relationship should not be frozen
+         */
+        REQUIRE_NOT_FROZEN,
```
Nana-EC, line 266: This should be Freeze or Pause, or is pause logic somewhere else?

Nana-EC (approved):
LG, with the notion that the exception should be freeze and a pause, not just freeze.
Maybe can be addressed in immediate follow up PR

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenRejectHandler.java
```diff
@@ -0,0 +1,381 @@
...
+     */
+    private TokenTransferList processFungibleTokenRejection(
+            @NonNull final TokenReference rejection,
+            @NonNull final Account rejectingAccount,
+            @NonNull final WritableTokenStore tokenStore,
+            @NonNull final WritableTokenRelationStore relStore) {
+        final var accountID = rejectingAccount.accountIdOrThrow();
+        final var token = getIfUsable(rejection.fungibleTokenOrThrow(), tokenStore);
+        final var tokenId = token.tokenIdOrThrow();
+        final var tokenRelation = getIfUsable(accountID, tokenId, relStore);
```
Neeharika-Sompalli, line 285: Wouldn't this fail if the token is paused or frozen ?

MrValioBg: It would indeed fail.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
MrValioBg: Yes, and also by default it uses REQUIRE_NOT_FROZEN.
However, I think since we currently don't have a usecase to support this case, we could probably remove this enum and leave it required by default. ( Just so we don't add code thats not neccesary needed).

I could remove in the follow up PR, which will also change the protobuf branch

Nana-EC: Okay thanks

nathanklick (approved):

rbarkerSL (approved):
LGTM

Issue #11160, opened 2024-01-25
Work package: investigate, retest and/or fix issues around Ethereum equivalency


stoqnkpL:
### Abstract/Motivation
In the last several months, a lot of Hedera<>Ethereum in-equivalence issues were addressed. 
In testing various items, some bugs were identified that should either be already resolved, or are a side-effect of these efforts.

### Goal
Test and reproduce the following scenarios, automate the test cases, investigate the issues and fix the still active bugs. 
More information is available in each item's description:
```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-services/issues/8373
- [ ] https://github.com/hashgraph/hedera-services/issues/9120
- [ ] https://github.com/hashgraph/hedera-services/issues/11033
- [ ] https://github.com/hashgraph/hedera-services/issues/11034
- [ ] https://github.com/hashgraph/hedera-services/issues/11158
- [ ] https://github.com/hashgraph/hedera-services/issues/11159
```

georgi-l95:
All issues tagged with `blocker` tag are left as future items, because changes there require extensive testing. Due to 0.48 being so close, we prefer not to take additional risks.

PR #13686, opened 2024-06-05
test: Add hapi tests for call operations on system accounts

### LINKED ISSUE #11137: System accounts disambiguation: Create hapi tests for Call operations for system accounts
### Problem

### Solution

Add tests for **Call** operations(**CallOp**, **DelegateCallOp**, **CallCodeOp**, **StaticCallOp**) involving the system accounts and ensure that these tests conform to the following expected behavior.
image

### Alternatives

_No response_

kselveliev:
**Description**:
This PR adds hapi tests for call operations(call, callCode, staticcall, delegateCall) against system account.

After the introduction of `system-accounts-operations.md` doc all call to system accounts should result in success.

This PR add new tests in `Evm46ValidationSuite` for all call types against different combinations of system accounts and expects success result.

**Related issue(s)**:

Fixes #11137 

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
```diff
@@ -126,6 +126,7 @@ public class Evm46ValidationSuite {
     public static final List<Long> existingSystemAccounts = List.of(999L, 1000L);
     public static final List<Long> systemAccounts =
             List.of(0L, 1L, 9L, 10L, 358L, 359L, 360L, 361L, 750L, 751L, 999L, 1000L);
+    public static final List<Long> callOperationsSuccessSystemAccounts = List.of(0L, 1L, 358L, 750L, 751L, 999L, 1000L);
```
kselveliev, line 129: Notice here I'm not using the `public static final List<Long> systemAccounts` because the call with will revert with PRECOMPILE_ERROR for accounts 9, 10, 359, 360, 361. 
According to our current doc `system-accounts-operations.md` all calls to system accounts should result in success.
There were some bug fixes that were addressing those precompile errors `[here](https://github.com/hashgraph/hedera-services/issues/11160)` but it looks like there are still some corner cases where the calls fail.
Should I open a ticket for those failing cases. @Neeharika-Sompalli what do you think.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
```diff
@@ -1474,4 +1475,92 @@ final Stream<DynamicTest> directCallToSystemAccountResultsInSuccessfulNoOp() {
...
+            int finalI = i;
+            opsArray[i] = withOpContext((spec, opLog) -> allRunFor(
+                    spec,
+                    contractCall(
+                                    contract,
+                                    functionName,
+                                    mirrorAddrWith(callOperationsSuccessSystemAccounts.get(finalI)))
+                            .hasKnownStatus(SUCCESS)
+                            .via("callTest" + finalI)));
+        }
```
MrValioBg, line 1494: Maybe we could extract the common logic into a separate method?

Also do we need the ".via("callTest" + finalI)));"  part? Seems like we are not doing any verifications with it.

kselveliev: Good suggestion. I extracted the common looping logic in a new method. And also removed the .via as we don't need it.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
```diff
@@ -1474,4 +1475,63 @@ final Stream<DynamicTest> directCallToSystemAccountResultsInSuccessfulNoOp() {
...
+        final var contract = "CallOperationsCheckerSuccess";
+        final var functionName = "staticcall";
+        final HapiSpecOperation[] opsArray = getCallOperationsOnSystemAccounts(contract, functionName);
+        return defaultHapiSpec("testStaticCallOperationsForSystemAccounts")
+                .given(uploadInitCode(contract), contractCreate(contract))
+                .when()
+                .then(opsArray);
+    }
+
+    private HapiSpecOperation[] getCallOperationsOnSystemAccounts(String contract, String functionName) {
```
MrValioBg, line 1523:
```suggestion
    private HapiSpecOperation[] getCallOperationsOnSystemAccounts(final String contract, final String functionName) {
```
nit:

MrValioBg (approved):
LGTM!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
kselveliev: Comment addressed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/evm/Evm46ValidationSuite.java
Neeharika-Sompalli: Yes, I will defer to @tinker-michaelj or @david-bakin-sl or @lukelee-sl .

kselveliev: @Neeharika-Sompalli, today I spent some time debugging and finding the issues in my first comment.
Cases 9,10 are evm precompiles so they are expected to fail with improper input, no issues here.
Cases 359, 360, 361 are our hts, exchange rate and prng system contract, we are obviously passing incorrect abi input so they revert. This is expected but was not noted in system-accounts-operations.md. I've just added a commit explaining that. 
So all in all there are no issues, we can move forward with this PR.

bilyana-gospodinova (approved):
LGTM

Neeharika-Sompalli (approved):

Neeharika-Sompalli (approved):

PR #13730, opened 2024-06-06
docs: 13690 Added a design doc for `Ledger State API`

### LINKED ISSUE #13690: Create a design document for the Public Merkle API project
In accordance with a new process, there has to be a design document representing #10988

Core Part of this Design:
- Get all of the state represented in the States API

imalygin:
**Description**:

This design doc is a compilation of the following Notion pages:

[State Refactor for State Proofs](https://www.notion.so/swirldslabs/State-Refactor-for-State-Proofs-0c1310cb7c8442409335431f53ec22db)

[Public Specified Merkle Tree & State](https://www.notion.so/swirldslabs/Public-Specified-Merkle-Tree-State-ab28e518d1f94ed2a84fcea9513cbc58#4eaae3323d3c454984b1e0ab162cb014)

[Public Merkle Tree API](https://www.notion.so/swirldslabs/Public-Merkle-Tree-API-698ae50389b843cba04976ec0714fbd9)

[Merkle tree to protobuf](https://www.notion.so/swirldslabs/Merkle-tree-to-protobuf-616ee86f676d4bf2acc3679cfe828ce6)

[Proposed Platform State Structure](https://www.notion.so/swirldslabs/Proposed-Platform-State-Structure-d37ae2c018f94b81ac884745ca4ee022)

Fixes #13690 

**Notes for reviewer**:

**Checklist**


platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+* String label for State Name
+* Protobuf message for child data
+
+When writing to disk it should have a SingletonState protobuf message with 3 fields:
+
+* Hash of children Hashes
+* String label for State Name
+* Protobuf message for child data. This field is problematic as it can be one of many kinds. It seems the best solution is probably using protobuf `any` type. The other option is `oneof` with all the singleton content types. See:
+  * [https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto) ## Saved State Snapshot Changes
```
alittley, line 151: is `## Saved State Snapshot Changes` intended to be tacked on to this line? Or is it supposed to be on its own line?

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+Currently [PlatformState](/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformState.java) is a special case node of the merkle tree. As a part of the simplification effort **it should be refacored to a set of singleton states**. Also, some of the `PlatformState code` is no longer relevant, and needs to be removed. There is a prerequsite for the `PlatformState` to become a singleton - Platform State split
```
alittley, line 153: `refacored` -> `refactored`

alittley, line 153: You have the following in code ticks:`PlatformState code`

but I assume you intended to just have `PlatformState` in ticks

alittley, line 153: prerequsite -> prerequisite

alittley, line 153: You have `- Platform State split` tacked on to the end of this line, but it doesn't seem to belong

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+* epoch hash
+  * we no longer support emergency reconnect
+* next epoch hash
+  * we no longer support emergency reconnect
+* uptime data
+  * this is a half baked feature and not currently utilized for anything. We should reconsider how we are going to store this data when we have time to revisit the concept (risk of choosing bad protobuf definitions is too high to make it worth preserving existing functionality)
+
+### Refactoring
+
+Once `PlatformState` is cleaned-up and `RosterMap` and `TssData` are extracted, these objects should be turned into **singleton state objects**. This requires protobuf defintions for these objects in `HAPI` module.
```
alittley, line 197: `defintions` -> `definitions`

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
+# State Refactor for State Proofs
```
alittley, line 1: This document is missing test related considerations, which are meant to be included here.

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+# New State Merkle Tree Design
+
+So the new merkle tree design will be simplified to a single root node which contains a list of named states. Each state is one of 3 types: Map, Queue or Singleton.
+
+![state.png](assets/state.png)
+
+> ***Simplify, Simplify & Simplify some more***
+
+Needs to be the philosophy here, we have to make the merkle tree as simple as possible. Remove as much legacy code as possible and all unused code. The reason we need it simple is what ever we expose today will be backed into API and very hard to change in the future. For example what goes into a internal node hash. Is it just the hash of the two 48 byte hashes or is it the hash of two complex serialized objects wrapped in a parent serialized object each with lots of fluff.
```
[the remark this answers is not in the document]
imalygin: I tried to reformat the latest version to make it more readable.

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+ * All files written outside of the directory will be discared.
+ */
+void protoSerialize(WriteableSequentialData out, Path artifactsDir);
+```
+
+Artifacts dir is where a node may store additional files beyond the main state file `SignedState.pbj` For example, MerkleDb will copy (hard-link) its files to that dir rather than writing all data to the state file.
+
+`AbstractMerkleInternal` may provide default implementation for this method by writing all child nodes (as protobuf messages, with tags and lengths) to the stream. Each child will be serialized along with its index in the parent to avoid relying on any order in the stream.
+
+If an internal node needs to store any data beyond its children (do we really want to support it?), the corresponding class may have custom implementation of the method above, delegate to super to write children, then write its own data.
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+A new method is to be added to `MerkleNode`:
+
+```java
+/**
+ * Serialize this node and all its child nodes to the specified output
+ * buffer in protobuf format. If a node needs to store anything aside
+ * the buffer, like large binary data, it should be written to the
+ * specified artifacts directory. This directory will be preserved, so
+ * when a node is deserialized, directory content will be available.
+ * All files written outside of the directory will be discared.
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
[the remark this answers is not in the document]
imalygin: I removed this part.

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+## Merkle Tree to Protobuf
+
+### Writing to protobuf
+
+A new method is to be added to `MerkleNode`:
+
+```java
+/**
+ * Serialize this node and all its child nodes to the specified output
+ * buffer in protobuf format. If a node needs to store anything aside
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+To write children, another helper method is needed in `MerkleNode`:
+
+```java
+/**
+ * Returns the size, in bytes, of this object, when it is serialized to
+ * protobuf format using {@link #protoSerialize()} method, including
+ * all children data.
+ */
+int getProtoSizeInBytes();
```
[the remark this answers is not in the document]
imalygin: Removed.

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+MerkleNode root = merkleDataInputStream.readMerkleTree(dir)
+```
+
+will be replaced with something like
+
+```java
+MerkleNode root = new MerkleHederaState(protoInputStream, dir)
+```
+
+Since applications may have different node classes used as the root of the tree, such calls will be delegated to applications, similar to how genesis states are currently constructed. The example above is valid for `Hedera`, which will use `MerkleHederaState` as the root.
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+Problem:
+
+* `ConstructableRegistry` must go away
+* Class IDs must go away
+* Without the registry and class IDs, when a state is loaded from a protobuf stream, what key/value codecs to use for the state?
+
+Proposal:
+
+* `Hedera` object contains a services registry (of type `ServicesRegistryImpl`). This registry can be used to store all known key and value codecs by service name + state key
+* `PlatformBuilder` already have a mechanism to delegate genesis state creation to the app. A similar mechanism can be provided to delegate loading existing state from a stream to the app.
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+Problem:
+
+* `ConstructableRegistry` must go away
+* Class IDs must go away
+* Without the registry and class IDs, when a state is loaded from a protobuf stream, what key/value codecs to use for the state?
+
+Proposal:
+
+* `Hedera` object contains a services registry (of type `ServicesRegistryImpl`). This registry can be used to store all known key and value codecs by service name + state key
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+* `PlatformBuilder` already have a mechanism to delegate genesis state creation to the app. A similar mechanism can be provided to delegate loading existing state from a stream to the app.
+* `Hedera` will implement this mechanism by creating a `MerkleHederaState` object from a protobuf stream - see "Reading from protobuf" section above - and providing it access to all registered codecs using the services registry
+* When `MerkleHederaState` reads its state children from the stream, it will first read state metadata (service name and state key from the label, state type may be hard-coded?), then find the codecs, and provide them to the state children
```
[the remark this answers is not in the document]
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
imalygin: Agreed. I removed the link

imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
imalygin: Removed

platform-sdk/docs/proposals/state-refactor/state-refactor.md
imalygin: Added test related considerations

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,76 @@
...
+
+![state-modules](assets/state-modules.png)
+
+Note that the `swirlds-state-api` is already created and has a set of classes by [this PR](https://github.com/hashgraph/hedera-services/pull/12570). There is also one [more PR](https://github.com/hashgraph/hedera-services/pull/13331) related to this change.
+
+There is one more change that can be relevant in the context of public APIs. This is the proposed change to the Merkle tree structure.
+
+![new-merkle-tree-structure.png](assets/new-merkle-tree-structure.png)
+
+``
```
alittley, line 46: misplaced ticks ``

Can be fixed if you end up needing to submit an updated version of this proposal

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,76 @@
...
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.
+
+The goal of this design proposal is to ensure that one can interact with all the objects in the Merkle tree using the newly introduced State API.
+
+This refactoring will enable optimized state proof, as it will be possible to represent an n-ary root node as a virtual binary tree.
+Therefore, the number of hashes in the state proof will be minimized.
+
+## Requirements
+
+- `PlatformState` to a singleton object as defined by the State API.
```
alittley, line 21: grammar: this sentence seems to be missing a verb. Maybe intention was 

"`PlatformState` should be migrated to..."

can be fixed if you end up needing to submit an updated version of the proposal

alittley (approved):
+1

anthony-swirldslabs (dismissed):
+1

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Fixed

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed the ticks

anthony-swirldslabs (approved):
+1

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+The goal of this design proposal is to ensure that one can interact with all the objects in the Merkle tree using the newly introduced State API.
+
+This refactoring will enable optimized state proof, as it will be possible to represent an n-ary root node as a virtual binary tree.
+Therefore, the number of hashes in the state proof will be minimized.
+
+## Requirements
+
+- `PlatformState` should be refactored to a singleton object as defined by the State API.
+- the Merkle tree should be reshaped so that `MerkleHederaState` becomes the root of the tree and `PlatformState` becomes its child.
```
poulok, line 22: These two items are implementation details, not requirements. I think the requirement is for every part of the state, including `PlatformState` to be accessible through the State API.

The second bullet is unclear. Why is this being specified? What does this change mean to the users of the State API?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hedera state.
+  This module should have a minimal set of dependencies. Ideally, the Block Node should not have a compile-time dependency
+  on any other modules but this one to interact with the state.
```
poulok, line 25: The use of the word "ideally" in a requirement makes me fidget a little. What would stop this from being a proper requirement and removing the "ideally"? What stops us from saying that we will definitely ensure that the block node does not need any compile time dependency other that on this module in order to interact with the state?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
+# Merkle State API
+
+## Summary
+
+Refactor the Merkle tree so that it can be fully represented by the State API objects. Create a module that will provide simple and easy-to-use API for the Merkle tree.
```
poulok, line 5: The summary should include a statement explaining WHY this proposal is needed. Why does the tree need to be fully represented by the State API?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Purpose and Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.
+
+The goal of this design proposal is to ensure that one can interact with all the objects in the Merkle tree using the newly introduced State API.
```
poulok, line 14: This statement does not match with the title of the proposal PR. I understand that the two concepts are interrelated, but it is not made clear in this document. Nowhere in the document are state proofs discussed. There are several purposes for this proposal:

1. Create an easy to use API for the entire state so that Block Nodes can interact with the state without having to understand the intricacies of merkle trees and merkleDB
2. Define the state in public format so that state proofs are publicly consumable
3. Reshape the tree into binary format to enable small state proofs

The title of the proposal indicates the purpose of all this is for state proofs. If that is true, all this should be described here.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hedera state.
+  This module should have a minimal set of dependencies. Ideally, the Block Node should not have a compile-time dependency
+  on any other modules but this one to interact with the state.
+- there should be a new module - `swirlds-state-impl` module to host the implementation of these interfaces and helper classes.
+
+## Changes
+
+### Public API
+
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change.
```
poulok, line 32: If there are no changes being made to the public API in this design proposal, that statement should appear explicitly. i.e. "There are not changes to public APIs in this proposal".

But this is not really true. The first requirement says that the platform must be refactored to a singleton in the State API, which means it must be defined as a protobuf object, which is a public API. So I am confused.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+There is one more change that can be relevant in the context of public APIs. This is the proposed change to the Merkle tree structure.
+
+![new-merkle-tree-structure.png](assets/new-merkle-tree-structure.png)
+
+## Components and Architecture
+
+### Platform State
+
+Currently `PlatformState` is a special case node of the merkle tree. As a part of the simplification effort **it should be refactored to a singleton state**.
```
edward-swirldslabs, line 50: Clarification Requested.    If the platform state is a singleton does every update of its substructure require rehashing of the entire singleton?   I'm asking because the TSS protocol is going to be noisy and cause a lot of state changes within the platform state.  Or is the plan that TSS related state is stored outside of the platform state?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Modules and Repos
+
+Two new modules are added as a part of this proposal:
+
+![state-modules](assets/state-modules.png)
+
+Note that the `swirlds-state-api` is already created and has a set of classes by [this PR](https://github.com/hashgraph/hedera-services/pull/12570). There is also one [more PR](https://github.com/hashgraph/hedera-services/pull/13331) related to this change.
+
+There is one more change that can be relevant in the context of public APIs. This is the proposed change to the Merkle tree structure.
```
poulok, line 42: This is a good diagram, but all the context and reasoning for this change is missing from the proposal.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
poulok, line 50: What is meant by "a special case node of the merkle tree"?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Components and Architecture
+
+### Platform State
+
+Currently `PlatformState` is a special case node of the merkle tree. As a part of the simplification effort **it should be refactored to a singleton state**.
+
+### KV Node
+
+We need to have a class that represents key-value node in the tree, as we have it for singleton and queue nodes. Currently, this class is missing.
```
poulok, line 54: I don't understand this part. What does this mean and why is it needed? What is the difference between the one you say already exists and the one that does not yet exist?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: > If the platform state is a singleton does every update of its substructure require rehashing of the entire singleton?

Yes. We'll need to rehash the entire singleton whenever its state changes. 

>Or is the plan that TSS related state is stored outside of the platform state?

The plan is to extract TSS from the platform state. TSS is going to be a separate leaf node (likely, a `VirtualMap` to make constant rehashing easier)

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+### KV Node
+
+We need to have a class that represents key-value node in the tree, as we have it for singleton and queue nodes. Currently, this class is missing. 
+
+### Merkle state refactoring
+
+- Move `PlatformState` from the root to `MerkleHederaState`
+- Replace usage of existing root of `State` type with `MerkleHederaState`
+  - It would be much easier to do, if they implemented the same interface (e.g. `RootNode`). Currently, there is now direct relation between these classes.
```
poulok, line 60: I'm confused by this statement's wording. "Currently, there is _now_...." I think should be "Currently, there is _no_...."

I think this should be in the implementation plan instead of here.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+- Make sure that the state migration happens properly during the version update
+
+## Test plan
+
+This design doesn't add any new functionality. Therefore, the assumption is that as long as all existing tests are passing (unit tests, integration tests, and performance tests), the change is considered to be tested.
```
poulok, line 65: Migration testing definitely needs to be part of the plan here. The state is changing in 2 ways: 1) PlatformState is moving into the new API and into protobuf definition, and 2) the location of the platform state is moving.

These need to be tested prior to the feature being enabled in `develop`. There is no Hedera state migration test that is run as part of the PR checks, so there needs to be a plan for this. Maybe you run the migration locally, or you give a branch to devops and see if they can run a migration using a real testnet state.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Implementation and delivery plan
+
+- Move State API implementation classes from `swirlds-platform-core` to `swirlds-state-impl`.
+- Add common interface for `MerkleHederaState` and `State` to make the migration easier.
```
poulok, line 70: Should there be a step to remove this interface once the migration is done?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+- Make `MerkleHederaState` a root of the Merkle tree.
+- Implement a class that represents KV node.
+- Refactor `PlatformState` into a singleton. This depends on [migrating PlatformState to protobuf](https://github.com/hashgraph/hedera-services/pull/13694). Migrate the data.
```
poulok, line 73: The PR link seems to be wrong. From the text, I expected a PR that defines the `PlatformState` in protobuf.

Also, the format of `PlatformState` as protobuf is a public API. I think this information needs to be included as part of this proposal or at least linked and explained how it relates to this one since it is a dependency.

poulok (changes requested):
Great start for this proposal, but there are holes that need to be filled in. I suspect @Ivan has all the information in his head, it just needs to be written down :-) 

The most important hole to fill is the test plan, and including the new public API of the `PlatformState`.


UPDATE: I initially vetoed this since i think it needs several additions, but after reminding myself what the numbers mean, I want to change my vote to a +0.5. I don't know how to remove the "Changes Requested" without also "Approving". Obviously this work needs to be done and I think everything presented here is on track. If more information is provided as I requested in my comments, I would change my vote to a +1.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: It means that unlike other top-level nodes, `PlatformState` is not a singleton, queue or KV state.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Summary
+
+Refactor the Merkle tree so that it can be fully represented by the State API objects. Create a module that will provide simple and easy-to-use API for the Merkle tree.
+
+## Purpose and Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
```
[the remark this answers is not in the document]
imalygin: As you recommended, I added explanation what "easy to use" API means in this context.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed this part out of the proposal

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed the world "ideally".

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I added "Why" section to the document

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed this sentence.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I added clarification about the `PlatformState`

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed it as it should be a part of a different design proposal

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed it, as it should be a part of the proposal about State API object serialization

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I meant to say "there is _no_ direct relation between these classes" but I removed this part as it should be a part of a different design proposal.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Eventually we should remove `State` class as it's no longer needed. I will make it a part of a different proposal.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Fair enough. I updated `Test plan` section.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+### Platform State
+
+Currently `PlatformState` is a special case node of the merkle tree. That is, it doesn't conform to State API. As a part of the simplification effort **it should be refactored to a singleton state**.
+
+## Test plan
+
+Migration testing:
+
+- we need to make sure that the data of the current platform state migrates properly to a platform state represented by a singleton node.
+- we need to test that `PlatformState` can be properly migrated to protobuf format[](https://)
```
poulok, line 59: empty trailing hyperlink

poulok (commented):
+0.9 vote

Looks great, just a few more points of clarification needed.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Public API
+
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change. `PlatformState` class will become a part of the public API. [](https://)
+
+### Modules and Repos
+
+Two new modules are added as a part of this proposal:
+
+![state-modules](assets/state-modules.png)
```
hendrikebbers, line 44: I can not open any of the illustrations with draw.io As far as I know, we defined to use that for documentation. We should do it here, too.

hendrikebbers (commented):
Vote -0.1 based on the current state (several points unclear or missing from my point of view)

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -2,34 +2,40 @@
...
+to a separate module as a part of this proposal.
 
-This refactoring will enable optimized state proof, as it will be possible to represent an n-ary root node as a virtual binary tree.
-Therefore, the number of hashes in the state proof will be minimized.
+### Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.[](https://)
```
alittley, line 24: There is an empty hyperlink at the end of this line

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -2,34 +2,40 @@
...
+  This module should have a minimal set of dependencies. The Block Node should not have a compile-time dependency
   on any other modules but this one to interact with the state.
 - there should be a new module - `swirlds-state-impl` module to host the implementation of these interfaces and helper classes.
 
 ## Changes
 
 ### Public API
 
-State API classes and interfaces will migrate to a designated module. The classes themselves will not change.
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change. `PlatformState` class will become a part of the public API. [](https://)
```
alittley, line 38: another trailing empty hyperlink here

edward-swirldslabs:
> I don't know how to remove the "Changes Requested" without also "Approving". 

You can clear your block by hitting the recycle button next to the blocked symbol at the list of reviewers.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.[](https://)
+
+## Requirements
+
+- `PlatformState` should be refactored to a singleton object as defined by the State API.
+- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hedera state.
+  This module should have a minimal set of dependencies. The Block Node should not have a compile-time dependency
+  on any other modules but this one to interact with the state.
+- there should be a new module - `swirlds-state-impl` module to host the implementation of these interfaces and helper classes.
```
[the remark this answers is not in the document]
imalygin: We don't plan any other implementation. My initial thinking was that API and implementation in one module will be confusing, and I followed the pattern that we have for other modules - `swirlds-config-api` / `swirlds-config-impl` or `swirlds-metrics-api`/ `swirlds-metrics-impl`. However, I don't have a strong opinion about this and, in fact, that would be easier for me to implement as well, if we didn't have to have another module.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Changes
+
+### Public API
+
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change. `PlatformState` class will become a part of the public API. [](https://)
```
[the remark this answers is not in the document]
imalygin: We want to provide public API for the Merkle tree. `PlatformState` is a part of the Merkle tree. I just don't see any reason why we should hide `PlatformState`. Am I missing something?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+- `PlatformState` should be refactored to a singleton object as defined by the State API.
+- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hedera state.
+  This module should have a minimal set of dependencies. The Block Node should not have a compile-time dependency
+  on any other modules but this one to interact with the state.
+- there should be a new module - `swirlds-state-impl` module to host the implementation of these interfaces and helper classes.
+
+## Changes
+
+### Public API
```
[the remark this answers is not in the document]
imalygin: I'll add an example

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
hendrikebbers: For metrics and config we did it because it must work with a different config / metrics system. If you want to use then platform in a spring boot application you want to have the spring metrics system be the implementation of our APIs. That was a wish since the beginning. If we do not assume to have a second implementation than I would put both in 1 module to minimize the module count.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
hendrikebbers: PlatformState sounds to me like "this is the state of the platfom instance at this moment". That sounds wrong for block nodes for example. But maybe I'm wrong.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Both images are viewable in a browser and in IDE. Please, use one of these.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: >That sounds wrong for block nodes for example

Why does it sound wrong? That is, I'm not _adding_ the `PlatformState` to the Merkle tree, it's already there, always has been. I'm just covering it with a common interface.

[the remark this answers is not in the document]
imalygin: The usage part is not clear for me. I assume that it may be useful for the Block Node, but I don't know how exactly it's going to be useful. 

For the internal use (in platform-sdk or services) there is an alternative interface that provides direct access to the `PlatformState` instance. 

To sum up - this platform state refactoring is driven by the fact that we need to be the Merkle tree uniformed and hide its complexity behind the scene, not by necessity of exposing `PlatofrmState` as a part of API

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed the PR link and added proto definition to the proposal

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: okay, fair enough. I'll remove the notion of `swirlds-state-impl`

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+### Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.[](https://)
+
+## Requirements
+
+- `PlatformState` should be refactored to a singleton object as defined by the State API.
```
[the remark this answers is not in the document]
imalygin: I added more context about singletons to the document.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -2,34 +2,72 @@
...
 ## Changes
 
 ### Public API
 
-State API classes and interfaces will migrate to a designated module. The classes themselves will not change.
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change. `PlatformState` class will become a part of the public API.
+
+Usage example:
+```java
+private PlatformState findPlatformState(HederanState state) {
```
anthony-swirldslabs, line 64: Nit: s/Hederan/Hedera/

anthony-swirldslabs (dismissed):
+0.(9)

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,300 @@
...
+
+- string label for State Name
+- protobuf message for child data
+
+In Java code it's represented by the following classes:
+```java
+public interface ReadableSingletonState<T> {
+  @NonNull
+  String getStateKey();
+  T get();
```
hendrikebbers, line 38: can you add a newline bethween the methods

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,300 @@
...
+  void put(@Nullable T value);
+
+  boolean isModified();
+}
+```
+
+## Requirements
+
+- `PlatformState` should be refactored to a singleton object as defined by the State API.
+- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hedera state.
```
hendrikebbers, line 52:
```suggestion
- there should be a new module - `swirlds-state-api` - to host a set of interfaces, records, and abstract classes that represent the Hashgraph state.
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
hendrikebbers: Better use hashgraph

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Fixed

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Removed

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
+# Merkle State API
+
+## Summary
+
+Refactor the Merkle tree so that it can be fully represented by the State API objects.
+Create a module that will provide a simple and easy-to-use API for the Merkle tree.
```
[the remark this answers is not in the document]
imalygin: I updated PR title

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Fixed

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
rbair23: I honestly don't know what to call it. In the open source world `Swirlds` isn't right, and `Hedera` isn't right, and `Hashgraph` might be right but also might be misleading because it makes me think that the hashgraph itself is stored in state, but it is not.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: @hendrikebbers would you recommend renaming interface `com.swirlds.state.HederaState` into `com.swirlds.state.HashgraphState`?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
rbair23: Another thought, I'm not sure `Public Merkle API` is quite right. The `Merkle`-ness is actually not exposed in the states API, it is an implementation detail.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Done

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: @rbair23 would `ApplicationState` or `AppState` be too generic?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
jasperpotts: We should be careful of noisy state changes as they will all get written to block stream for ever. We might be able to summarize once per round if that helps. But something to keep in mind.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: As long as people agree, I can go with `MerkleState`. I don't have a strong opinion

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: Would "Public Application State API" be better?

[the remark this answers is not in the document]
anthony-swirldslabs: Personally, I like "Public Application State API".

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
poulok: Diagrams must be provided in DrawIO format for the reason @hendrikebbers gave.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
poulok: Where was this explanation added? I'm having trouble finding it.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
poulok: > The usage part is not clear for me. I assume that it may be useful for the Block Node, but I don't know how exactly it's going to be useful.

Block nodes will read the block stream that contains state changes. It will then apply these state changes to its local state so that it has an up to date state from which to answer queries and produce state proofs.

> For the internal use (in platform-sdk or services) there is an alternative interface that provides direct access to the PlatformState instance.
I'm confused here. I thought platform will have to access `PlatformState` through the State API as well?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: It's in "Why?" section

```
The Merkle tree is a rather complex data structure, and to make interaction easier, we need to hide this complexity behind the API.
Another reason to have this API is to provide additional flexibility in changing the implementation details of the Merkle tree.
As long as they are hidden, we are free to change them.
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,65 @@
...
+As long as they are hidden, we are free to change them.
+
+Note that this API is already implemented as a part of the Service Modularization project, but these classes will be moved
+to a separate module as a part of this proposal.
+
+### Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
```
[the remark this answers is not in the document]
imalygin: I mentioned importance of this API for the Block node.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I agree but this should be a part of a different design proposal

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: okay, I'll remove it as it's already implemented anyway.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: Sure, I reworded it.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: > I thought platform will have to access PlatformState through the State API as well?

Even though it's possible to make it so, there is no benefit in a such a change. Why bother looking up a singleton by the service name and then get a value from there, if we can use the following interface instead? 

```
public interface MerkleRoot extends MerkleInternal {
    [skipped]
    /**
     * Get the platform state.
     *
     * @return the platform state
     */
    @NonNull
    PlatformState getPlatformState();
    /**
    [skipped]
}
````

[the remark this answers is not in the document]
Comment (hendrikebbers) [reply platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md]: I assume it is only me :)

File: platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md (hunk 37 above)
Comment (hendrikebbers) [reply platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md]: If it is a new one, yes

File: platform-sdk/docs/proposals/001-application-state-api/001-application-state-api.md [hunk 42]
```diff
@@ -0,0 +1,294 @@
...
+     */
+    SoftwareVersion first_version_in_birth_round_mode = 8;
+
+    /**
+     * A consensus round.<br/>
+     * The last round before the birth round mode was enabled.
+     * <p>
+     * This SHALL be `MAX_UNSIGNED` if birth round mode has not yet been enabled.
+     */
+    uint64 last_round_before_birth_round_mode = 9;
```
cody-littley, line 165: The birth round fields should have high index values. They are needed for now, but we will remove them after the birth round migration.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
cody-littley: Most of the platform state is rewritten exactly once per round. There is nothing particularly inefficient about using the state API in order to handle the platform state - i.e. I think there is minimal benefit to breaking this into multiple singleton objects.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
cody-littley: I prefer the term "ledger state", aka the "L" from DLT. It's not just the application's state, it's the platform and the application's state.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
cody-littley: "Ledger State" is also an option. I don't oppose "Merkle State" though.

alittley (approved):

platform-sdk/docs/proposals/001-application-state-api/001-application-state-api.md
imalygin: Sounds good. I will update the definition.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: It's not a new one. We have a `HederaState` interface now.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: The majority agrees on `MerkleState`. So, let it be it.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
imalygin: `MerkleState` is winning so far.

hendrikebbers:
Question: How can I see if the last changes happened before the PR moved to voting? Looks like a change in the board (https://github.com/orgs/hashgraph/projects/73/views/1) does not add any entry to the history of a PR. Adding @poulok since this is important for #13870

Or do we say that everything happened after somebody commented a number must have happened after the status of the PR moved to 'voting'?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
+# Merkle State API
+
+## Summary
+
+Refactor the Merkle tree so that it can be fully represented by the State API objects.
```
hendrikebbers, line 5:
```suggestion
Refactor the Merkle tree so that it can be fully represented by the State API.
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
+# Merkle State API
+
+## Summary
```
hendrikebbers, line 3: Your summary sounds more like a todo list to get the proposal implemented.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+
+Refactor the Merkle tree so that it can be fully represented by the State API objects.
+Create a module that will provide a simple and easy-to-use API for the Merkle tree.
+
+## Purpose and Context
+
+### Why?
+
+The Merkle tree is a rather complex data structure, and to make interaction easier, we need to hide this complexity behind the API.
```
hendrikebbers, line 12:
```suggestion
The Merkle tree is a rather complex data structure, and to make interaction easier, we need to hide this complexity behind an API.
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+Another reason to have this API is to provide additional flexibility in changing the implementation details of the Merkle tree.
+As long as they are hidden, we are free to change them. This API will be critically important for the Block Node to 
+interact with the state without knowing the details of the Merkle tree.
+
+Note that this API is already implemented as a part of the Service Modularization project, but these classes will be moved
```
hendrikebbers, line 17: For me, it is important to see the dependencies of the API & Implementation. What additional modules/libs are needed?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+to a separate module as a part of this proposal.
+
+### Context
+
+There are two issues that need to be addressed:
+
+* We need to have a public API for the Merkle tree that is simple and easy to use, residing in a separate module with a minimal set of dependencies.
+* The Merkle tree has to be fully expressed in terms of State API. Currently, `PlatformState` stands out and needs to be refactored.
```
hendrikebbers, line 25: `PlatformState` will be described later in the doc. Please refer to that or move the description more to the top.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### Singleton
+
+A singleton is a node with 2 children:
+
+- string label for State Name
```
hendrikebbers, line 31:
```suggestion
- label of type `String` for the state name
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+- protobuf message for child data
+
+In Java code it's represented by the following classes:
```
hendrikebbers, line 34: Does `String getStateKey()` references the name and `T get()` the protobuf message? Otherwise I do not see how this is related the description.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,255 @@
...
+Migration testing:
+
+- we need to make sure that the data of the current platform state migrates properly to a platform state represented by a singleton node.
+- we need to test that `PlatformState` can be properly migrated to protobuf format
+
+## Implementation and delivery plan
+
+- Move State API implementation classes from `swirlds-platform-core` to `swirlds-state-api`.
+- Refactor `PlatformState` into a singleton. Migrate the data.
+- Move `MerlkeHederaState` to `swirlds-state-api`
```
hendrikebbers, line 255:
```suggestion
- Move `MerkleHederaState` to `swirlds-state-api`
```

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: How you you propose to reword it?

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I added the javadoc for these interfaces to make it clearer.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I added a reference

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Sure. I specified all the dependencies

poulok:
>Question: How can I see if the last changes happened before the PR moved to voting? Looks like a change in the board (https://github.com/orgs/hashgraph/projects/73/views/1) does not add any entry to the history of a PR. Adding @poulok since this is important for https://github.com/hashgraph/hedera-services/pull/13870
>Or do we say that everything happened after somebody commented a number must have happened after the status of the PR moved to 'voting'?

I think we have to go by the history line that says the PR was marked as "Ready for Review". That starts the 3 day window.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,311 @@
...
+ */
+message PlatformState {
+    /**
+     * A consensus round.<br/>
+     * The round represented by this state.
+     * <p>
+     * This message SHALL represent the network state after handling of all
+     * transactions reaching consensus in all previous rounds.<br/>
+     * The first state (genesis state) SHALL have a round of 0.<br/>
+     * The first round to handle transactions SHALL be round `1`.
```
poulok, line 158: I think this comment has an off-by-one error. A state with round 10 means all transactions in rounds 0 through 10 have been applied. The statement `This message SHALL represent the network state after handling of all
     transactions reaching consensus in all previous rounds` misses the current round.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
```diff
@@ -0,0 +1,311 @@
...
+     */
+    uint64 round = 1;
+
+    /**
+     * A consensus timestamp for this state.
+     * <p>
+     * This SHALL be the consensus timestamp of the first transaction in
+     * the current round.
```
poulok, line 166: Not all rounds will have transactions, so I don't think this is right. There could even be rounds with no events.

The javadoc in the `ConsensusRound` class says:

```
The timestamp of the end of a round. Is equal to the consensus timestamp of the last transaction in the last event in the round, or the consensus timestamp of the last event in the round if there are no transactions in the last event. The consensus timestamp of a round is guaranteed to be strictly greater than the consensus timestamp of the previous round.
```
but this does not cover empty rounds. We need to figure out how to handle that case. I'm not sure what the answer should be.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
edward-swirldslabs: I prefer simply `State` but if qualification is warranted then I am in agreement with `LedgerState` in hiding the Merkle implementation and covering all use of state.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: Fair enough. Lazar recommended removing this field in favor of the one that is in `ConsensusSnapshot`, so this comment is no longer there.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
imalygin: I removed this field as well.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
cody-littley: Yeah, when we handle round 7, we've handled all transactions up to and including the transactions within round 7.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
cody-littley: - if there are transactions, the timestamp is equal to the timestamp of the last transaction
- if there are no transactions, the timestamp is equal to the timestamp of the last event
- if there are no events, the timestamp is equal to the timestamp of the previous round plus a small constant (1ns, 100ns, whatever. Should be a setting that services configures.)

poulok: @cody-littley this is great. I think this should be specified in the javadoc so external consumers know exactly how to verify these situations.

imalygin: Done. I updated javadoc for `ConsensusSnapshot.consensus_timestamp`

poulok (dismissed):

poulok (dismissed):

alittley (dismissed):
+1

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
kfa-aguda: `MerkleState` is an implementation detail. +1 on `LedgerState` - that's industry standard terminology and it's intuitive.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
hendrikebbers: Just an idea:

This proposal aims to define a public api for the ledger state. The api should be used to interact with the Merkle tree that will be an implementation detail in future. Since we already have a state api in the services layer a goal of the proposal is to move. that api to the platform layer and use it as central api to access the ledger state in all our code & products.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
hendrikebbers: +1 on `LedgerState`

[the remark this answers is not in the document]
rbair23: +1 on `LedgerState1, great name.

anthony-swirldslabs (approved):
+1

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
alittley: > label of type String for the state name

`getStateKey()`

> protobuf message for child data

`get()`


I will modify the description to hopefully add some more clarity

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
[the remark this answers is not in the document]
alittley: `LedgerState` wins.

platform-sdk/docs/proposals/001-merkle-state-api/001-merkle-state-api.md
alittley: I will modify the summary

kfa-aguda (dismissed):
+1

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Changes
+
+### Public API
+
+State API classes and interfaces will migrate to a designated module. The classes themselves will not change. `PlatformState` class will become a part of the public API.
+
+Usage example:
+```java
+private PlatformState findPlatformState(HederaState state) {
```
rbair23, line 125:
```suggestion
private PlatformState findPlatformState(LedgerState state) {
```

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
```diff
@@ -0,0 +1,315 @@
...
+ * restart and reconnect.
+ */
+message PlatformState {
+
+    /**
+     * A version describing the current version of application software.
+     * <p>
+     * This SHALL be the software version that created this state.
+     */
+    proto.SemanticVersion creation_software_version = 1;
```
rbair23, line 159: Does this get updated each release? Or only the genesis release? The name `creation_software_version` is not strange to me, I am not sure what it is exactly.

rbair23 (dismissed):
+1

anthony-swirldslabs (dismissed):
+1

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
alittley: Every time a new state object is constructed, this field will be populated with the current software version. Which is to say: yes, this will be updated each release, not just at genesis.

`This SHALL be the software version that created this state.` Do you think that this statement needs to be expanded upon? I can't think of any alternate interpretations that make it ambiguous, but i I'm open to suggestions

anthony-swirldslabs (approved):
+1

poulok (approved):

edward-swirldslabs (commented):
+0.99  - Do these changes need to be reflected in HIP anywhere?

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
rbair23: I think I don't understand the lifecycle of a PlatformState, or what is meant by "every time a new state object is constructed". Does this mean every fast copy? Or does it mean at restart? Or at upgrade? Why is it important to store this version information with the platform state, as the version is available in other ways (and is in another place in state)?

kfa-aguda (approved):

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
alittley: > Does this mean every fast copy? Or does it mean at restart? Or at upgrade?

Practically, the value will only be changed at genesis, and at upgrades (since in all other cases, the version of the prior state will just be carried forward.

> and is in another place in state

I wasn't aware that this was stored elsewhere in the state. Bearing that in mind, I can't think of any reason that the duplicate data would need to be in the platform state. Since @imalygin wrote this proposal, I'll hold off making this change until he can weigh in.

poulok: @alittley i don't think we need to wait for @imalygin to return to decide this. The PlatformState structure was based on the current structure and input from the hashgraph team. If the version is somewhere else in state and is now accessible to the platform, we can use that. I will point out though, that removing it from the platform state will require every single app to include the version in its state. Sure, it's already in the Hedera App state, but is that a requirement we want to add? Maybe the version should stay in the platform and be removed from the app state.

@rbair23 what do you think?

hendrikebbers:
+0.9 the proposal is great but I do not have enough knowledge of the technical topic to do +1

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
rbair23: Until block streams, we have leniency in modifying this. I'm fine if this goes in. But we should double-check this sometime before it becomes baked in stone forever.

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
alittley: done

platform-sdk/docs/proposals/001-ledger-state-api/001-ledger-state-api.md
alittley: I will merge as-is, and we can reconsider this later

edward-swirldslabs (approved):
+1 the hip I asked about would come through block-streams at the very latest.

Issue #11900, opened 2024-03-05
Gradle Build: Automate sorting entries in module-info files


jjohannes:
Current situation is annoying as it requires a lot of manual sorting, in particular when you work on multiple modules. An argument to turn the check off is that we also do not check the order for test/jmh dependencies in "build.gradle.kts" files. Or, we show automate the sorting as part of `spotlessApply` (if that is possible).

jjohannes:
We do not want to turn off the step, but rather automate it through a custom Spotless `FormatterStep`.

jjohannes:
We should/could do a similar auto sorting for the "requires" entries in "build.gradle.kts" files.

PR #13734, opened 2024-06-07
build: automate sorting of 'module-info' entries with spotless

jjohannes:
**Description**:

With this, the sorting is now done automatically and consistently when you run `qualityGate` or `spotlessApply`.

The PR also re-formats all `module-info` files according to the stricter rules, which follow what is already done in most _platform_ projects. The changes are:
- Strict order of blocks: 
  1. `requires transitive`
  2. `requires`
  3. `requires static transitive`
  4. `requires static`
- No blank lines or comments in between
- All our own modules from namespaces – "com.swirlds", "com.hedera.node", "com.hedera.storage" – are now on top. This was not the case before, because the old check did not support to configure multiple namespaces for this.

Let me know, if you think the rules should be different.

**Related issue(s)**:

#11900

**Notes for reviewer**:

**Checklist**


timo0 (approved):
🥳

jeromy-cannon (approved):

PR #13761, opened 2024-06-10
feat: remove gossip event.get hashed data part 2

### LINKED ISSUE #13760: Remove usage of GossipEvent.getHashedData Part-2
Since the serialization and hashing of events will be based on protobuf, `BaseEventHashedData` will no longer be an internal part of an event. To complete the migration, all usages of `GossipEvent.getHashedData` should be removed.

Note: because this task is too big for one PR, it will be split into multiple tickets and multiple PRs

lpetrovic05:
closes #13760


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
```diff
@@ -97,6 +97,11 @@ private static final class ClassVersion {
      * This latch counts down when prehandle has been called on all application transactions contained in this event.
      */
     private final CountDownLatch prehandleCompleted = new CountDownLatch(1);
+    /**
+     * The actual birth round to return. May not be the original birth round if this event was created in the software
+     * version right before the birth round migration.
+     */
+    private long birthRoundOverride;
```
alittley, line 104: Is `birthRoundOverride` the correct name for this? (I know this is how it was named in the old class, but I think it was probably misnamed there, too)

This member variable simply represents the `birthRound`, and it just so happens that is possible to override the `birthRound`

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/main/java/com/swirlds/platform/test/consensus/framework/validation/ConsensusRoundValidation.java
```diff
@@ -99,10 +99,7 @@ public static void validateIterableRounds(
      * @param e2 the second event
      */
     private static void assertConsensusEvents(final String description, final EventImpl e1, final EventImpl e2) {
-        final boolean equal = Objects.equals(e1.getHashedData(), e2.getHashedData())
-                && Objects.equals(
-                        e1.getBaseEvent().getConsensusData(), e2.getBaseEvent().getConsensusData())
-                && e1.isWitness() == e2.isWitness();
+        final boolean equal = Objects.equals(e1.getBaseEvent(), e2.getBaseEvent()) && e1.isWitness() == e2.isWitness();
```
alittley, line 102: with this change `GossipEvent::getConsensusData` no longer has any usages

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
lpetrovic05: I think it makes sense. `birthRound` is part of the core event data, and its part of the hash, this variable overrides that and returns whatever is set.

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/main/java/com/swirlds/platform/test/consensus/framework/validation/ConsensusRoundValidation.java
lpetrovic05: for the time being, yes. but it will be used in upcoming PRs, this is a PBJ record that will be part of the block stream

jeromy-cannon (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
```diff
@@ -243,7 +250,7 @@ public long getGeneration() {
      * @return the birth round of the event
      */
     public long getBirthRound() {
```
hendrikebbers, line 252: I agree with @alittley regaridng the naming. Now we have a `setBirthRoundOverride` but a `getBirthRound`. That is not really consinstent.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/resubmitter/DefaultTransactionResubmitter.java
```diff
@@ -63,20 +66,18 @@ public List<ConsensusTransactionImpl> resubmitStaleTransactions(@NonNull final G
         }
 
         final List<ConsensusTransactionImpl> transactionsToResubmit = new ArrayList<>();
-        for (final ConsensusTransactionImpl transaction : event.getHashedData().getTransactions()) {
-            if (transaction.isSystem()) {
-                if (transaction instanceof final StateSignatureTransaction signatureTransaction) {
+        final Iterator<Transaction> iterator = event.transactionIterator();
+        while (iterator.hasNext()) {
+            final Transaction transaction = iterator.next();
+            if (transaction.getPayload().kind().equals(PayloadOneOfType.STATE_SIGNATURE_PAYLOAD)) {
```
hendrikebbers, line 72:
```suggestion
            if (Objects.equal(transaction.getPayload().kind(),PayloadOneOfType.STATE_SIGNATURE_PAYLOAD) {
```
or
```suggestion
            if (PayloadOneOfType.STATE_SIGNATURE_PAYLOAD.equals(transaction.getPayload().kind())) {
```
By doing so a NPE can never be thrown if we run in the error that kind is null. No idea if that can happen :D

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/stale/StaleEventDetectorMetrics.java
```diff
@@ -70,7 +71,9 @@ public void reportStaleEvent(@NonNull final GossipEvent event) {
         int systemTransactions = 0;
         int appTransactions = 0;
 
-        for (final ConsensusTransactionImpl transaction : event.getHashedData().getTransactions()) {
+        final Iterator<Transaction> iterator = event.transactionIterator();
```
hendrikebbers, line 74: Why do you switched to exliziet iterator usage instead of forEach?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/validation/DefaultInternalEventValidator.java
```diff
@@ -174,8 +175,9 @@ private boolean areRequiredFieldsNonNull(@NonNull final GossipEvent event) {
      */
     private boolean isTransactionByteCountValid(@NonNull final GossipEvent event) {
         int totalTransactionBytes = 0;
-        for (final ConsensusTransaction transaction : event.getHashedData().getTransactions()) {
-            totalTransactionBytes += transaction.getSerializedLength();
+        final Iterator<Transaction> iterator = event.transactionIterator();
```
hendrikebbers, line 178:
```
event.transactionStream()
    .mapToInt(t -> t.getSerializedLength())
    .sum();
```
That would look way better from my point of view.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/internal/EventImpl.java
```diff
@@ -304,10 +302,21 @@ public GossipEvent getBaseEvent() {
...
+    }
+
+    /**
+     * Check if the event has other parents.
+     *
+     * @return true if the event has other parents
      */
-    public BaseEventHashedData getHashedData() {
-        return baseEvent.getHashedData();
+    public boolean hasOtherParent() {
```
hendrikebbers, line 318:
```suggestion
    public boolean hasOtherParents() {
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
lpetrovic05: I am tagging @cody-littley since I think he originally wrote the code so he can comment.

But I think `setBirthRound` is more misleading. `birthRound` is an immutable part of an event, it cannot be set after an event is created. So even if the override is changed, the birth round remains the same. The override seems to be some kind of migration trick (don't know the exact details) that will be used only once when migrating and will be deleted afterwards.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/stale/StaleEventDetectorMetrics.java
lpetrovic05: the event interface only provides and `Iterator`, not an `Iterable`, so I cannot use forEach. This will change once transactions are no longer stored in arrays

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/validation/DefaultInternalEventValidator.java
lpetrovic05: see comment https://github.com/hashgraph/hedera-services/pull/13761#discussion_r1636234187

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/internal/EventImpl.java
lpetrovic05: This is an issue with the fact that we only half-way support multiple other parents. We added support in some parts of the code, but not in others. Consensus, which uses this method, does not support multiple other parents. So I think it would be confusing if only this method implied there could be multiple parents, but not others.

jeromy-cannon (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
alittley: Bearing this in mind, it seems that the member should be called `birthRound`, and the method should be called `overrideBirthRound`

hendrikebbers (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
cody-littley: This variable is planned to be used only during the migration from a generation based system to a birth round based system. Events created prior to the migration will be interpreted with as having a "fake" birth round. Lots of technical details, but this is necessary to prevent some bad corner cases (zombie events). Once the migration is finished, this variable will be deleted.

This variable is not serialized, it's set in the intake pipeline.

Whatever we name this variable, I think it's important to capture the fact that it's not the same as an event's actual birth round.

What about something like `birthRoundToReturn`? This captures the fact that it might be the real (serialized) birth round, or possibly something different.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
cody-littley: I'm a fan of the naming schema

- `getBirthRound()`
- `overrideBirthRound(long birth round)`

This getter method is intentionally named like a regular getter. After the birth round migration, this method will actually be just a regular getter (since the override is a migration-only construct). All parts of the code should use this getter the same way they would if there was no such thing as an overriding birth round.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
alittley: I [had suggested](https://github.com/hashgraph/hedera-services/pull/13761#discussion_r1636454454) member name `birthRound`, and method name `overrideBirthRound`, which I think gets the same thing across.

I'm fine with your suggestion though

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/GossipEvent.java
lpetrovic05: renamed

alittley (approved):

cody-littley (approved):

jeromy-cannon (approved):

PR #13766, opened 2024-06-10
chore: Delete `XTest` module

Neeharika-Sompalli:
Deletes XTest module, since most of the tests are no more working. We will have integration tests that validate state with Embedded Hedera in the future


tinker-michaelj (approved):

kimbor (approved):

jeromy-cannon (approved):

PR #13771, opened 2024-06-10
chore: Some miscellaneous changes needed for `HandleWorkflow` refactor

Neeharika-Sompalli:
Related to #13638 

- Removed unused `HandleContext` methods
- Since there are no SystemContracts, we don't need privileges to be there for `CONTRACT_CREATE` and `CONTRACT_UPDATE` transactions. There was a review comment [on this PR](https://github.com/hashgraph/hedera-services/pull/9151#discussion_r1367152481) to remove these 
- Moved `PrecedingTransactionCategory` into `HandleContext` interface
- Added `Dispatch` interface that will have all the context needed for processing transactions in service logic

**Dagger Changes:**
- Added three scopes under `@Singleton` scope to clarify the scope for each object created while handling transactions
    - [UserTxnScope](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTxnScope.java) - This is annotated on objects that last complete  platform transaction
    - [UserDispatchScope](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/user/UserDispatchScope.java) - This is annotated on objects that last only the user transaction dispatch
    - [ChildDispatchScope](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/child/ChildDispatchScope.java) - This is annotated on objects that last only the child transaction dispatch
 - Added three Components for the above three scopes. [UserTxnComponent](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTransactionComponent.java), [UserDispatchComponent](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/user/UserDispatchComponent.java), [ChildDispatchComponent](https://github.com/hashgraph/hedera-services/blob/7be45485b45c18e9b7f87082eed4d77e130eb973/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/child/ChildDispatchComponent.java) The components `create()` method is injected with the dependencies that are are basic and needed to handle the transaction.
 - Each Component has modules that provide needed dependencies in the particular scope. They are annotated with `@Module`
 - Added `ChildQualifier` to differentiate any dependencies injected in multiple scopes. This qualifier is used to differentiate dependencies for the child transaction scope


tinker-michaelj (approved):

tinker-michaelj (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
```diff
@@ -86,22 +83,6 @@ public SystemPrivilege hasPrivileges(
                     payerId, txBody.fileUpdateOrThrow().fileIDOrThrow().fileNum());
             case FILE_APPEND -> checkFileChange(
                     payerId, txBody.fileAppendOrThrow().fileIDOrThrow().fileNum());
-            case CONTRACT_UPDATE -> {
```
mhess-swl, line 89: Why are `CONTRACT_UPDATE` and `CONTRACT_DELETE` removed?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTxnScope.java
```diff
@@ -0,0 +1,45 @@
...
+ * This scope contains the following bindings:
+ * <ol>
+ *     <li>HederaState</li>
+ *     <li>configuration</li>
+ *     <li>TokenContext</li>
+ *     <li>RecordListBuilder</li>
+ *     <li>PreHandleResult</li>
+ *     <li></li>
+ *  </ol>
+ */
```
mhess-swl, line 40: I'm confused how this scope is different than `@UserDispatchScope`?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/modules/LastHandledTime.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Qualifier to distinguish between two Instant values provided by Dagger.
+ * We annotate this qualifier for the last handled time.
+ */
+@Target({METHOD, PARAMETER, TYPE})
+@Retention(RUNTIME)
+@Documented
+@Qualifier
+public @interface LastHandledTime {}
```
mhess-swl, line 37: I get that this annotation is for distinguishing Instants, but shortening it to `LastHandledTime` – I suspect that people will ask what "last handled time" is a lot in the future

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
Neeharika-Sompalli: There are no system contracts right now. This is confusing to think we need privileges for `CONTRACT_UPDATE` and `CONTRACT_DELETE`. @tinker-michaelj  had a comment[ on the PR](https://github.com/hashgraph/hedera-services/pull/9151#discussion_r1367152481) , but it was not removed.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTxnScope.java
Neeharika-Sompalli: UserTxnScope includes anything that happens only once for platform transaction. Like staking period hook, genesis record hook. Since we consider user and child as dispatches now to re-use the code for both, `UserDispatchScope` includes all objects that are shared in `DispatchProcessr` and `HandleLogic` only for User transaction.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/modules/LastHandledTime.java
Neeharika-Sompalli: This is just provided by `BlockRecordManagerImpl.lastHandledTime`. I will add a comment. Thanks @mhess-swl

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTxnScope.java
mhess-swl: 👍 Not necessary in this PR, but we may want to point that out in docs somewhere

hedera-node/hedera-app/src/main/java/com/hedera/node/app/authorization/PrivilegesVerifier.java
```diff
@@ -234,7 +197,7 @@ private SystemPrivilege checkCryptoUpdate(
         final var targetId = op.accountIDToUpdateOrElse(AccountID.DEFAULT);
         final long targetNum = targetId.accountNumOrElse(0L);
         final var treasury = accountsConfig.treasury();
-        final var payerNum = payerId.accountNumOrThrow();
+        final var payerNum = payerId.accountNumOrElse(0L);
```
netopyr, line 200: We should not do this IMO, unless we have a check somewhere that `treasury` and `accountsConfig.systemAdmin()` can never be 0. We should return `UNAUTHORIZED` if the `accountNum` is not set.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/Dispatch.java
```diff
@@ -0,0 +1,150 @@
...
+import com.swirlds.platform.state.PlatformState;
+import com.swirlds.state.spi.info.NodeInfo;
+import java.time.Instant;
+import java.util.Set;
+
+/**
+ * The context needed for executing business logic of a service. This has two implementations - one for user transactions scope
+ * and one for dispatched child transactions scope.
+ */
+public interface Dispatch {
```
netopyr, line 42: I think we should rename this interface. Maybe `DispatchContext`?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/Dispatch.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * The builder for the transaction record in the scope
+     * @return the builder
+     */
+    SingleTransactionRecordBuilderImpl recordBuilder();
```
netopyr, line 47: This looks wrong. Shouldn't this be `SingleTransactionRecordBuilder`?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/Dispatch.java
```diff
@@ -0,0 +1,150 @@
...
+     * The transaction info for the transaction
+     * @return the transaction info
+     */
+    TransactionInfo txnInfo();
+
+    /**
+     * The payer of the transaction. This will be the synthetic payer for child transactions.
+     * @return the payer
+     */
+    AccountID syntheticPayer();
```
netopyr, line 65: I think the name is misleading as it is only a synthetic payer for child transactions. Maybe just `payer` or `transactionPayer`?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/Dispatch.java
```diff
@@ -0,0 +1,150 @@
...
+     * The consensus time of the user transaction
+     * @return the consensus time
+     */
+    Instant consensusNow();
+
+    /**
+     * The required keys needed to sign the transaction
+     * @return the required keys
+     */
+    Set<Key> requiredKeys();
```
netopyr, line 101: I assume these sets will be immutable. This should be mentioned in the JavaDoc.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/Dispatch.java
```diff
@@ -0,0 +1,150 @@
...
+     * The handle context for the transaction scope
+     * @return the handle context
+     */
+    HandleContext handleContext();
+
+    /**
+     * The savepoint stack for the transaction scope
+     * @return the savepoint stack
+     */
+    SavepointStackImpl stack();
```
netopyr, line 119: Again, I wonder why this is not referencing the interface `SavepointStack`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTransactionComponent.java
```diff
@@ -0,0 +1,154 @@
...
+     * The token context for the user transaction
+     * @return the token context
+     */
+    TokenContext tokenContext();
+
+    /**
+     * The savepoint stack used for the user transaction
+     * @return the savepoint stack
+     */
+    SavepointStackImpl stack();
```
netopyr, line 130: Can we use the interface `SavepointStack` instead?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/modules/StateModule.java
```diff
@@ -0,0 +1,50 @@
...
+public interface StateModule {
+    @Provides
+    @UserTxnScope
+    static HederaState provideHederaState(WorkingStateAccessor workingStateAccessor) {
+        return workingStateAccessor.getHederaState();
+    }
+
+    @Provides
+    @UserTxnScope
+    static SavepointStackImpl provideSavepointStackImpl(@NonNull final HederaState state) {
```
netopyr, line 41: Same as above. It would be better if we could use `SavepointStack`.

PR #13778, opened 2024-06-10
chore: design proposal for self event buffer

### LINKED ISSUE #13777: Write proposal for elimination of accidental self branches
It's possible for an honest node to create a branch in the hashgraph if it creates a self event, gossips that event, and then crashes prior to that event being written to a PCES file.

Propose a solution that makes it impossible for a node to self-branch in this way.

cody-littley:
closes #13777


platform-sdk/docs/proposals/noAccidentalBranches.md
```diff
@@ -0,0 +1,100 @@
...
+# Risks
+
+This new component risks increasing the time it takes for an event to reach consensus, since we are adding a delay
+between the creation of a self event and the time when the event becomes eligible to be gossiped.
+
+# Future Work
+
+An idea has been pitched where we make the PcesWriter super, super fast (possibly by memory-mapped files). In such a
+scenario, we might be able to remove the self event buffer entirely, and feed gossip from the events that come
+out of the writer.
```
anthony-swirldslabs, line 74: With this future in mind, why not simplify the design and nuke the `SelfEventBuffer` altogether? And I suppose the `PcesWindowManager` could be nuked too, although I'm not entirely sure.

The idea is to feed the `Gossip` from both the `PcesSqequencer` and the `PcesWriter` directly. The first input drops self events in the `Gossip` component because we only want to gossip them once they're persisted. The second input drops non-self events because they've already been gossiped through the first input.

Wouldn't this solve the problem statement with minimum changes to the existing architecture, or there's any flaws that the proposed more complex solution addresses?

cody-littley: The main problem here is that we aren't 100% sure that it's possible to take the alternate approach, and the work to test its plausibility might end up being quite large.

anthony-swirldslabs: From what I see in this design PR and the related GitHub ticket, the proposed design hasn't been tested for its plausibility yet, too. If that is true, then I'd propose to start the test with the simplest design possible - e.g. the one I described above - and only introduce more complexity in the form of new components/wiring if the simple idea doesn't work.

I won't block your design if you strongly believe that it's superior to the simple solution. However, I'd really like to hear what other voters think, and ultimately I'd like to see at least some theoretical arguments that would show the advantage of the more complex solution and/or any flaws of my simple idea that I may have missed.

cody-littley:
Leemon signaled a -1 for this this approach, closing

poulok:
> Leemon signaled a -1 for this this approach, closing

@cody-littley Please include the reasoning for this. This step is very important. We need to document WHY this approach was withdrawn so we do not have to rehash or find conversations in slack.

cody-littley:
Leemon's primary concern was the potential impact this could have on C2C.

PR #13780, opened 2024-06-10
chore: Add logic for `ChildDispatchScope`,  `UserDispatchScope` and `UserTxnScope`

Neeharika-Sompalli:
Related to #13638 

Add logic for `ChildDispatchScope`,  `UserDispatchScope` and `UserTxnScope`

**ChildDispatchScope:** 
- Added `ChildDispatchFactory` that creates child dispatches based on the child category and reversing behavior
- Added `ChildRecordBuilderFactory` to initialize child records
- Added `ChildTxnInfoFactory` to provide `TransactionInfo` for child transactions

**UserDispatchScope:**
- Added `PrehandleResultManager` to provide updated prehandle results whenever called 
- Added `UserRecordInitializer` to initialize user record

**UserTxnScope:** 
- Added `HollowACcountCompleter` to finalize hollow accounts
- Added `SchedulePurger` to purge schedules


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/logic/SchedulePurger.java
```diff
@@ -0,0 +1,77 @@
...
+     * time.
+     * @param userTxnContext the user transaction component
+     */
+    public void expireSchedules(@NonNull UserTransactionComponent userTxnContext) {
+        final var lastHandledTxnTime = userTxnContext.lastHandledConsensusTime();
+        if (lastHandledTxnTime == Instant.EPOCH) {
+            return;
+        }
+        if (userTxnContext.consensusNow().getEpochSecond() > lastHandledTxnTime.getEpochSecond()) {
+            final var firstSecondToExpire = lastHandledTxnTime.getEpochSecond();
```
netopyr, line 63: These calculations look different than the original code. Does it calculate the same values?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/logic/SchedulePurger.java
```diff
@@ -0,0 +1,77 @@
...
+            final var lastSecondToExpire = userTxnContext.consensusNow().getEpochSecond() - 1;
+            final var scheduleStore = new WritableStoreFactory(
+                            userTxnContext.stack(),
+                            ScheduleService.NAME,
+                            userTxnContext.tokenContext().configuration(),
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Yes that makes sense. Thanks @netopyr

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/logic/SchedulePurger.java
Neeharika-Sompalli: The original code was looking at first consensus time of last block and current time. But we can accomplish the same by using the last handled time and current time. So, we purge expired schedules for each transaction.

tinker-michaelj (dismissed):

tinker-michaelj (approved):

tinker-michaelj (approved):

netopyr (approved):
LGTM - tyvm @Neeharika-Sompalli

PR #13785, opened 2024-06-11
feat: add `EmbeddedHedera`

### LINKED ISSUE #13425: Replace `XTest` by using `@HapiTest` with an embedded `HederaNetwork`
### Problem

As a developer I often want to validate pure application logic, especially workflow logic that integrates many components. 

Running a `@HapiTest` against a subprocess network does, of course, accomplish this.

However, 
  1. The time spent starting a subprocess network dominates the time of running a single `@HapiTest`.
  2. Even the very short time-to-finality of such a network adds up to hours when running thousands of `@HapiTest`s.
  3. Debugging a node running in a separate JVM is much more cumbersome than debugging a JUnit test process.
  4. It is impossible to generate deterministic record streams from a live network; so regression tests need fuzzy-matching.

The `xtest` module tried to address these pain points, but has the massive shortcoming that we cannot reuse `HapiSpec` DSL and our existing library of `@HapiTest`s. 

This issue is to take the basic ideas from `xtest`, discard that module, and instead allow `@HapiTest`'s to be run against a `Hedera` network "embedded" in the JUnit test process.

### Solution

:one: **Encapsulate `@HapiTest` network operations**
- [x] #13539
- [x] Extend the existing subprocess-based `HederaNetwork` to implement the new APIs via gRPC stubs managed by `HapiClients`.
- [x] Add a `RemoteHederaNetwork` implementation that can target tests at a remote network (i.e. not managed subprocesses).
- [x] Instantiate all `HapiSpec`'s with a target `HederaNetwork`; eliminate any use of the `nodes` spec property that does not consist of instantiating a `SubProcessHederaNetwork` or `RemoteHederaNetwork`.
- [x] Change all `HapiTxnOp#callToUse()` and `HapiQueryOp#submitWith()` implementations to invoke the appropriate `HederaNetwork` API on the spec's target network.

:two: **Implement the `EmbeddedHederaNetwork`**
- [x] #13537
- [x] #13543
- [x] #13542
- [x] Create an `EmbeddedHederaNetwork` that instantiates a `FakePlatform`, `FakeServicesRegistry`, and `FakeHederaState`; then uses then to instantiate and initialize a single `Hedera` that will represent every node in the network.
- [x] Implement the `HederaNetwork` APIs to directly call `{Ingest/PreHandle/Handle/Query}Workflow`s as appropriate with synchronous responses.
- [x] Register a `testEmbedded` task that configures the `SharedNetworkSessionListener` to set the global `SHARED_NETWORK` to be an `EmbeddedHederaNetwork` instead of a `SubProcessHederaNetwork`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13425 ; please see issue for primary work items.

Also,
 - Adds comments and does minor cleanup in `Hedera`.
 - Simplifies current synthetic account creation code as it is temporary pre-Block Streams, and should assume it is starting from an empty state at genesis.
 - Deletes `GenesisRecordBuilder` as this is obsolete with Block Streams and was questionable to have in library code anyways.
 - Adds `ServicesMigrator` API to allow `EmbeddedHedera` to pass a `FakeServicesMigrator`.
 - Passes `Configuration` to `getStatesToCreate()` for Services whose state definitions depend on configuration.


edward-swirldslabs (dismissed):

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/Schema.java
```diff
@@ -71,6 +72,18 @@ public Set<StateDefinition> statesToCreate() {
...
 
+    /**
+     * A variant of {@link #statesToCreate()} that allows for the {@link Configuration} to be
+     * consulted when determining which states to create. This is useful if the schema needs to
+     * create different states based on the configuration.
+     *
+     * @param configuration The configuration to consult
+     * @return A map of all states to be created. Possibly empty.
+     */
+    public Set<StateDefinition> statesToCreate(@NonNull final Configuration configuration) {
```
hendrikebbers, line 83: NonNull / NUllable annotation for method is missing

hendrikebbers (commented):
Only reviews platform code

jeromy-cannon (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -367,82 +305,53 @@ public void onStateInitialized(
...
-            deserializedVersion = new HederaSoftwareVersion(
-                    toPbj(((SerializableSemVers) previousVersion).getProto()),
-                    toPbj(((SerializableSemVers) previousVersion).getServices()),
-                    0);
         } else {
-            deserializedVersion = new HederaSoftwareVersion(SemanticVersion.DEFAULT, SemanticVersion.DEFAULT, 0);
+            if (previousVersion != null) {
+                logger.fatal("Deserialized state not created with Hedera software");
+                throw new IllegalStateException("Deserialized state not created with Hedera software");
+            }
```
Neeharika-Sompalli, line 347: So, we don't allow migrating from mono-state now?

tinker-michaelj: True Neeha, this PR begins the mono removal that I am continuing in [this](https://github.com/hashgraph/hedera-services/pull/13828) draft PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -532,101 +415,44 @@ private void unmarkMigrationRecordsStreamed(HederaState state) {
...
-            // TBD: notifications.register(NewSignedStateListener.class, daggerApp.newSignedStateListener());
-            // com.hedera.node.app.service.mono.state.exports.NewSignedStateListener
-            // Has some relationship to freeze/upgrade, but also with balance exports. This was the trigger that
-            // caused us to export balance files on a certain schedule.
-        } catch (final Throwable th) {
-            logger.error("Fatal precondition violation in HederaNode#{}", daggerApp.nodeId(), th);
-            daggerApp.systemExits().fail(1); // TBD: Better exit code?
-        }
+                case REPLAYING_EVENTS, STARTING_UP, OBSERVING, RECONNECT_COMPLETE, CHECKING, FREEZING, BEHIND -> {
+                    // Nothing to do here, just enumerate for completeness
```
Neeharika-Sompalli, line 447: Should we log here, so we know the platform status

imalygin (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/config/ConfigProviderBase.java
```diff
@@ -86,8 +86,11 @@ protected void addFileSource(
         };
 
         try {
-            final Path propertiesPath =
-                    Optional.ofNullable(System.getenv(envName)).map(Path::of).orElseGet(() -> Path.of(defaultPath));
+            final Path propertiesPath = Optional.ofNullable(System.getenv(envName))
+                    .or(() -> Optional.ofNullable(
+                            System.getProperty(envName.toLowerCase().replace("_", "."))))
+                    .map(Path::of)
+                    .orElseGet(() -> Path.of(defaultPath));
```
Neeharika-Sompalli, line 93: Why do we need this ?

Neeharika-Sompalli (approved):
LGTM!

PR #13786, opened 2024-06-11
chore: enable HIP-206 and HIP-906

lukelee-sl:
**Description**:
Enable HIP-206 and HIP-906 by default in all environments.


Nana-EC (approved):

tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

PR #13809, opened 2024-06-12
Metric label proposal

hendrikebbers:
Metric label proposal

### DIFF
--- a/platform-sdk/docs/proposals/metric-labels/README.md
+++ b/platform-sdk/docs/proposals/metric-labels/README.md
@@ -153,7 +153,7 @@ Next to that all that ideas will end in having a hardcoded `nodeId` label in the
 ## Future Work
 
 As a next step we can think about labels next to `nodeId` that makes sense to have in our metrics.
-For example we can add a label called `type` that defines if a metric is a counter, gauge, or histogram.
+For example, we can add a label called `type` that defines if a metric is a counter, gauge, or histogram.
 Another option is to add labels for layers (like `base`, `consensus`, `services`, ...).
 By doing so we can add more information to the metrics that can be used by the monitoring system.
 
--- a/platform-sdk/docs/proposals/metric-labels/README.md
+++ b/platform-sdk/docs/proposals/metric-labels/README.md
@@ -46,7 +46,7 @@ The metrics module will be extended by a new data type called `Label`.
 A label is a key-value pair that can be defined as a `record`:
 
 ```java
-record Label(String key, String value) {}
+record Label(@NonNull String key, @Nullable String value) {}
 ``` 
 
 The `Metric` interface will be extended by a new method called `getLabels()` that returns a list of labels:
@@ -56,6 +56,7 @@ interface Metric {
     
     //...
     
+    @NonNull
     List<Label> getLabels();
 }
 ```
@@ -71,7 +72,11 @@ abstract class MetricConfig {
     
     //...
     
-    MetricConfig withLabels(Label... labels) {...}
+    @NonNull
+    MetricConfig withLabels(@NonNull Label... labels) {...}
+    
+    @NonNull
+    MetricConfig withLabels(@NonNull Collection<Label> labels) {...}
 }
 ```
 
@@ -83,12 +88,13 @@ Based on that we need to modify the `Metrics` interface to use the labels as the
 interface Metrics {
     
     // old methods:
-    // Object getValue(final @NonNull String category, final @NonNull String name);
-    // void remove(final @NonNull String category, final @NonNull String name);
+    // Object getValue(@NonNull String category, @NonNull String name);
+    // void remove(@NonNull String category, @NonNull String name);
     
-    Object getValue(final @NonNull String category, final @NonNull String name, final @NonNull Set<Label> labels);
+    @Nullable
+    Object getValue(@NonNull String category, @NonNull String name, @NonNull Set<Label> labels);
     
-    void remove(final @NonNull String category, final @NonNull String name, final @NonNull Set<Label> labels);
+    void remove(@NonNull String category, @NonNull String name, @NonNull Set<Label> labels);
 }
 ```
 
@@ -140,6 +146,7 @@ Here we need to check that a platform metric (that has a `nodeId` label) is stil
 
 - Should the `getLabels()` method return `Set`, `Collection` or `List`?
 - Should we add a new methods to the `Metrics` interface to get for example all metrics with a specific label?
+- How will a label with a `null` value be handled?
 
 ## Alternatives
 
--- a/platform-sdk/docs/proposals/metric-labels/README.md
+++ b/platform-sdk/docs/proposals/metric-labels/README.md
@@ -1,7 +1,9 @@
 # Metric labels
 
-+author: @hendrikEbbers
-+status: draft
+| Metadata           | Entities                                   | 
+ |--------------------|--------------------------------------------|
+ | Designers          | [@hendrikebbers](https://github.com/hendrikebbers), [@mxtartaglia-sl](https://github.com/mxtartaglia-sl) |
+
 
 ## Summary
 
--- a/platform-sdk/docs/proposals/metric-labels/README.md
+++ b/platform-sdk/docs/proposals/metric-labels/README.md
@@ -1,0 +1,166 @@
+# Metric labels
+
+author: @hendrikEbbers
+status: draft
+
+## Summary
+
+This proposal describes a new feature for the metrics module.
+The feature is called "metric labels" and allows to add labels to metrics.
+Labels are key-value pairs that can be used to filter and group metrics.
+Why this proposal uses the name "metric labels" (like prometheus does) other libraries (like micrometer) use the term tags.
+
+## Motivation
+
+Today the metrics module only supports the definition of metrics without adding custom labels.
+While each metric has a  hardcoded category, name, and nodeId no other custom information can be added.
+This proposal suggests adding a new concept called "metric labels" to the metrics module and
+migration the `category`, `name`, and `nodeId` information to labels.
+While the category and name are defined as String values the nodeId is a class called `NodeId`.
+By doing so the metrics module becomes more flexible and can be used 100% independent of the
+platform module since it does not depend on the platform module's `nodeId` anymore.
+
+## Goals
+
+- Add a new concept called "metric labels" to the metrics module
+- Migrate the `category`, `name`, and `nodeId` information to labels
+- Make the metrics module independent of the platform module
+- Provide a public API to define labels for metrics
+- Add labels to the snapshot API
+- The output of the snapshot API should contain the labels
+- The output of the snapshot API must be compatible with the current output (no breaking changes)
+
+## Non-Goals
+
+Even with the new concept the metrics api should be only used in our internal code.
+The metrics module should not be used by the end user.
+Here the preferred solution is still that we forward our metrics to a monitoring system that the custom application code uses.
+
+## Design & Architecture
+
+This chapter describes the changes that are needed to add labels to the metrics module.
+
+### Public API
+
+The metrics module will be extended by a new data type called `Label`.
+A label is a key-value pair that can be defined as a `record`:
+
+```java
+record Label(String key, String value) {}
+``` 
+
+The `Metric` interface will be extended by a new method called `getLabels()` that returns a list of labels:
+
+```java
+interface Metric {
+    
+    //...
+    
+    List<Label> getLabels();
+}
+```
+
+> [!IMPORTANT]  
+> I currently ask myself if the method should return `Set`, `Collection` or `List`. Help is appreciated.
+
+Like the name or category, the labels are defined when a metric is created and cannot be changed later.
+Based on that the MetricConfig class will be extended by a new method called `withLabels`:
+
+```java
+abstract class MetricConfig {
+    
+    //...
+    
+    MetricConfig withLabels(Label... labels) {...}
+}
+```
+
+By adding labels the unique identifier of a metric is the key and the labels.
+Based on that we need to modify the `Metrics` interface to use the labels as the unique identifier of a metric:
+
+```java
+
+interface Metrics {
+    
+    // old methods:
+    // Object getValue(final @NonNull String category, final @NonNull String name);
+    // void remove(final @NonNull String category, final @NonNull String name);
+    
+    Object getValue(final @NonNull String category, final @NonNull String name, final @NonNull Set<Label> labels);
+    
+    void remove(final @NonNull String category, final @NonNull String name, final @NonNull Set<Label> labels);
+}
+```
+
+Since the old methods are only used in tests or demos today we can do that change easily.
+
+#### Snapshot API
+
+The snapshot API will be extended to return the labels of a metric for a snapshot.
+By doing so any monitoring system that uses the snapshot API can use the labels.
+Today we only have 2 system that uses the snapshot API: the prometheus endpoint and the CSV exporter.
+Since no class of the snapshot API is public today (based on the internal usage of `nodeId`) we plan to move parts
+of the snapshot API to the public API.
+
+### Migration
+
+Next to extending the public API we need to migrate the current metrics to the new concept.
+
+#### Migration of the `category`, `name`, and `nodeId`
+
+The `category`, `name`, and `nodeId` information of a metric will be migrated to labels.
+The key of a metric is the combination of the `category` and `name`.
+That will not be changed but next to that both values will be added as labels to the metric.
+By doing so a metric can be identified by its key and labels.
+
+Let's assume a metric with the category `myCategory`, the name `myName`.
+As today the key of the metric is `myCategory.myName`.
+Next to that the metric will have 2 labels: `category=myCategory` and `name=myName`.
+The unique identifier of the metric is the key and the labels.
+That is a breaking change since today the key is the unique identifier of a metric.
+Since we do not have api today to add labels the migration should be minimalistic.
+
+#### Migration of the prometheus endpoint
+
+The `nodeId` of a platform is already added as a label today when using the prometheus endpoint.
+For that endpoint we just need to remove the hard coded `nodeId` label and add a general algorithm to add labels.
+
+#### Migration of the CSV exporter
+
+The CSV exporter use the `nodeId` today to create the file name.
+Since the CSV exporter is only used by the platform module we can continue with that approach.
+
+## Testing
+
+The metrics module is tested by unit tests today. We need to add tests for the new concept of labels.
+Next to that we need to add tests of the CSV exporter and the prometheus endpoint to ensure that the labels are added correctly.
+Here we need to check that a platform metric (that has a `nodeId` label) is still working correctly and ends in the exact same output as today.
+
+## Open Questions
+
+- Should the `getLabels()` method return `Set`, `Collection` or `List`?
+- Should we add a new methods to the `Metrics` interface to get for example all metrics with a specific label?
+
+## Alternatives
+
+Since the idea of this proposal is to make the metrics module independent of the platform module we need to remove
+the usage of the `NodeId` class or extract the class from the platform and define it as general API.
+Since the class does not make any sense outside the consensus node (like in block node for example) this is not a good idea.
+Another option would be to change the type of the `nodeId` to a `String` but that would end in an ugly API for any
+system next to the platform that uses the metrics module since `null` or any other default must always be used as the `nodeId`.
+Next to that all that ideas will end in having a hardcoded `nodeId` label in the metrics that will be forwarded to the monitoring system.
+
+## Future Work
+
+As a next step we can think about labels next to `nodeId` that makes sense to have in our metrics.
+For example we can add a label called `type` that defines if a metric is a counter, gauge, or histogram.
+Another option is to add labels for layers (like `base`, `consensus`, `services`, ...).
+By doing so we can add more information to the metrics that can be used by the monitoring system.
+
+## References
+
+- https://prometheus.io/docs/practices/naming/#labels
+- https://prometheus.io/docs/practices/instrumentation/#use-labels
+- https://www.javadoc.io/doc/io.micrometer/micrometer-core/1.1.0/io/micrometer/core/instrument/Tag.html
+- https://github.com/vert-x3/vertx-micrometer-metrics/blob/master/src/main/java/io/vertx/micrometer/Label.java
+- https://geode.apache.org/docs/guide/115/tools_modules/micrometer/micrometer-meters.html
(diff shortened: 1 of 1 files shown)


platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,166 @@
+# Metric labels
+
+author: @hendrikEbbers  
+status: draft
```
timo0, line 4: Not sure if this makes sense, since as soon it's accepted you need to push a change and need reviews again

hendrikebbers: No. The idea is: as long as it is a draft we should discuss without voting ;) Than it will be moved to `voting` and people can vote. We can even add `implementation in progress`, `implemented` and `declined`. From my point of view we need to store the state in the proposal. How will you otherwise know if a proposal that is merged is already implemented?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,166 @@
...
+- Make the metrics module independent of the platform module
+- Provide a public API to define labels for metrics
+- Add labels to the snapshot API
+- The output of the snapshot API should contain the labels
+- The output of the snapshot API must be compatible with the current output (no breaking changes)
+
+## Non-Goals
+
+Even with the new concept the metrics api should be only used in our internal code.
+The metrics module should not be used by the end user.
```
timo0, line 36: Would be the application (like services) also considered end user?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,166 @@
...
+interface Metric {
+    
+    //...
+    
+    List<Label> getLabels();
+}
+```
+
+> [!IMPORTANT]  
+> I currently ask myself if the method should return `Set`, `Collection` or `List`. Help is appreciated.
```
timo0, line 65: Favor of `Set` to avoid duplications

platform-sdk/docs/proposals/metric-labels/README.md
timo0: For this we have the board I guess

hendrikebbers: Once it is commited you want to have a state in the file

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: no, that is what I inlcude in our (hashgraph ecosystem) code

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: That is my current idea

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+    @NonNull
+    MetricConfig withLabels(@NonNull Label... labels) {...}
+    
+    @NonNull
+    MetricConfig withLabels(@NonNull Collection<Label> labels) {...}
+}
+```
+
+By adding labels the unique identifier of a metric is the key and the labels.
+Based on that we need to modify the `Metrics` interface to use the labels as the unique identifier of a metric:
```
mxtartaglia-sl, line 84: As far as I understand other metric frameworks, labels are used to count specific observations of a general event,
E.G:  While counting httpRequestsTotal, I would like to classify them into POST, PUT, GET, or by returned status code, or by device generating the request, or by the invoked path.
So each time I observe an event I am tracking with the metric, (assuming a counter) I would use 
httpRequestsTotal.increase(method,deviceType,statusCode, invokedPath).

later I can use promql to get how many get request i received: http_requests_total{method="GET"} or how many requests resulted in an error: http_requests_total{statusCode="500"} 

If I got your idea right, In your proposal, labels are part of the name of the metric, which would make:
a) metric names in the code and in Grafana harder to map.
b) more difficult to query as I would have to know all the labels in advance to create a meaningful query

I also think that category and unit are necessary Prometheus values that one needs to include when mapping our information to Prometheus.

I think that the proposed approach is not compatible with what other frameworks call `labels` or `tags` and including it in the present form in ours would create confusion and eventual incompatibility when we need to map information from existing libraries to swirlds metric framework.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+As a next step we can think about labels next to `nodeId` that makes sense to have in our metrics.
+For example, we can add a label called `type` that defines if a metric is a counter, gauge, or histogram.
+Another option is to add labels for layers (like `base`, `consensus`, `services`, ...).
+By doing so we can add more information to the metrics that can be used by the monitoring system.
+
+## References
+
+- https://prometheus.io/docs/practices/naming/#labels
+- https://prometheus.io/docs/practices/instrumentation/#use-labels
```
mxtartaglia-sl, line 172: From this article:
```
Use labels
Few monitoring systems have the notion of labels and an expression language to take advantage of them, so it takes a bit of getting used to.

When you have multiple metrics that you want to add/average/sum, they should usually be one metric with labels rather than multiple metrics.

For example, rather than http_responses_500_total and http_responses_403_total, create a single metric called http_responses_total with a code label for the HTTP response code. You can then process the entire metric as one in rules and graphs.
```

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: Oh, you got me wrong ;) labels **should not** be part of the name. But 1 metric will be identified by its name and its label. Today, a metric is unique only based on its name. With the proposal that will change and you need to give name + labels to get a unique request. Only query for the name can end in a collection of metrics

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
++status: draft
+
+## Summary
+
+This proposal describes a new feature for the metrics module.
+The feature is called "metric labels" and allows to add labels to metrics.
+Labels are key-value pairs that can be used to filter and group metrics.
+Why this proposal uses the name "metric labels" (like prometheus does) other libraries (like micrometer) use the term tags.
+
+## Motivation
```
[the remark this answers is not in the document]
rbair23: I started using tags with this Helidon module I was writing. I actually don't know how it manifests itself in Grafana. I am interested to find out. I think for example I can create a metric "transactions-per-second" and then have tags for the actual transaction type, and this will somehow give me overall tps plus per-type tps?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+Today the metrics module only supports the definition of metrics without adding custom labels.
+While each metric has a  hardcoded category, name, and nodeId no other custom information can be added.
```
rbair23, line 16:
```suggestion
While each metric has a hardcoded category, name, and nodeId no other custom information can be added.
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+This proposal suggests adding a new concept called "metric labels" to the metrics module and
+migration the `category`, `name`, and `nodeId` information to labels.
```
rbair23, line 18:
```suggestion
migrate the `category`, `name`, and `nodeId` information to labels.
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+- Migrate the `category`, `name`, and `nodeId` information to labels
+- Make the metrics module independent of the platform module
+- Provide a public API to define labels for metrics
+- Add labels to the snapshot API
+- The output of the snapshot API should contain the labels
+- The output of the snapshot API must be compatible with the current output (no breaking changes)
+
+## Non-Goals
+
+Even with the new concept the metrics api should be only used in our (hashgraph) code.
```
rbair23, line 35: I assume "(hashgraph)" also includes services.

platform-sdk/docs/proposals/metric-labels/README.md
mxtartaglia-sl: I would certainly prefer the second option; labels can be appended, and combined in different ways and sometimes omitted. Also, metrics instances are sometimes used as instance fields:

```
class SomeClass{
  private Metric metric; 

  SomeClass(){
    metric =metrics.getOrCreate("http_metric"); 
  }


  int myHttpHandler(String methodName, String deviceType){

     int status = handle();
     List<Label> labels = 
       List.of(Label.of(methodName), Label.of(status), Label.of(deviceType)):
   
     metric.increase(labels);
     //or metric.increase();
     //or metric.increase(Label.of(methodName));
     //or metric.increase(Label.of(methodName), Label.of(deviceType));
    //or metric.increase(Label.of(status), Label.of(deviceType));
     return status;
  }
}
```

The first option has the drawback that:
a) The user must define a label for every possible `int` that `handle()` would return when creating the metric. This seems plausible with the status code example, but how about other situations where it could return anything?
b) you must change all instances where the metric is stored in a field.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+By doing so any monitoring system that uses the snapshot API can use the labels.
+Today we only have 2 system that uses the snapshot API: the prometheus endpoint and the CSV exporter.
+Since no class of the snapshot API is public today (based on the internal usage of `nodeId`) we plan to move parts
+of the snapshot API to the public API.
+
+### Migration
+
+Next to extending the public API we need to migrate the current metrics to the new concept.
+
+#### Migration of the `category`, `name`, and `nodeId`
```
mxtartaglia-sl, line 115: We want to add NodeId as a label instead of a value, and I agree with the need there.
But I need to understand why names and categories should be handled as labels. This is not evident in the proposal.

As I said before, there is a difference in that the labels should be optional. You can track values without labels, but you can't without name and category.

On the other hand, one existing problem is tracking the metrics exposed in Grafana to the actual code that modifies their value.  If we complicate the mapping of a metric's name with the code that interacts with it ,

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+Next to that we need to add tests of the CSV exporter and the prometheus endpoint to ensure that the labels are added correctly.
+Here we need to check that a platform metric (that has a `nodeId` label) is still working correctly and ends in the exact same output as today.
+
+## Open Questions
+
+Please help me to answer the following questions before we can start with the implementation:
+
+- Should the `getLabels()` method return `Set`, `Collection` or `List`?
+- Should we add a new methods to the `Metrics` interface to get for example all metrics with a specific label?
+- How will a label with a `null` value be handled?
```
mxtartaglia-sl, line 151: >- Should the `getLabels()` method return `Set`, `Collection` or `List`?

Set. Order is not important. Repeated labels are not desired. 

> - Should we add a new methods to the `Metrics` interface to get for example all metrics with a specific label?

As I don't agree that it should be part of metric creation, I don't agree with this functionality either. 

> - How will a label with a `null` value be handled?

Labels should be optional; a label with a null value should be similar to not being indicated.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+migration the `category`, `name`, and `nodeId` information to labels.
+While the category and name are defined as String values the nodeId is a class called `NodeId`.
+By doing so the metrics module becomes more flexible and can be used 100% independent of the
+platform module since it does not depend on the platform module's `nodeId` anymore.
+
+## Goals
+
+- Add a new concept called "metric labels" to the metrics module
+- Migrate the `category`, `name`, and `nodeId` information to labels
+- Make the metrics module independent of the platform module
```
mxtartaglia-sl, line 27: This is IMO the main benefit of adding labels as a generally available feature in the framework.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,175 @@
...
+As today the key of the metric is `myCategory.myName`.
+Next to that the metric will have 2 labels: `category=myCategory` and `name=myName`.
+The unique identifier of the metric is the key and the labels.
+That is a breaking change since today the key is the unique identifier of a metric.
+Since we do not have api today to add labels the migration should be minimalistic.
+
+#### Migration of the prometheus endpoint
+
+The `nodeId` of a platform is already added as a label today when using the prometheus endpoint.
+For that endpoint we just need to remove the hard coded `nodeId` label and add a general algorithm to add labels.
```
mxtartaglia-sl, line 132: What does that algorithm look like? Where should it live? Is it part of this proposal?

hendrikebbers:
@mxtartaglia-sl and I discussed about the API and it looks like we found enough use cases to say that both ways that we discussed make sense:

```
Metric metric = metricFactory.create("http-response-time");
metric.measure(300);

metricFactory.setGlobalLabel("Platform", "123");

Metric metricEndpoint = metricFactory.getOrCreate("http-response-time", Label.of("endpoint", "/id"));

Metric metricGet = metricFactory.getOrCreate("http-response-time", Label.of("method", "GET"));
Metric metricPost = metricFactory.getOrCreate("http-response-time", Label.of("method", "POST"));
metricGet.measure(300);
metricPost.measure(200);

Metric metricFoo = metricFactory.getOrCreate("http-response-time");
metricFoo.measure(300, Label.of("method", "GET"));
metricFoo.measure(200, Label.of("method", "POST"));
metricFoo.measure(300);

MetricSupplier metricSupplier = metricFactory.getOrCreate("http-response-time");
metricSupplier.getOrCreate(Label.of("method", "GET")).measure(300);
metricSupplier.getOrCreate(Label.of("method", "POST")).measure(200);
```

hendrikebbers:
Copied from https://prometheus.io/docs/concepts/data_model/ as an example for a metric:

`api_http_requests_total{method="POST", handler="/messages"}`

Here the `handler` should be defined at metric creation but the `method` should be defined at measurement time.

hendrikebbers:
@mxtartaglia-sl and I had a discussion how timeseries with same name and different labels are stored and how you can request a value ignoring the labels. I found this one: https://stackoverflow.com/questions/45154993/how-can-i-group-labels-in-a-prometheus-query and https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches and https://prometheus.io/docs/prometheus/latest/querying/examples/

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+counter.increment();
+```
+
+Let's see how that metric can be used in a filter that filter transactions and counts the number of transactions of type `fileUpload`:
+
+```java
+class TransactionFilter {
+    
+    private final static Counter.Config config = new Counter.Config("transactionsCategory", "transactionCount");
```
mxtartaglia-sl, line 240: remove final and static from the example as the code would not compile when handle method overrides the value

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+    private final Counter counter;
+    
+    TransactionFilter(Metrics metrics) {
+        counter = metrics.getOrCreate(config);
+    }
+
+    void handle(Transaction transaction) {
+        if(transaction.getType().equals("fileUploadTransaction")) {
```
mxtartaglia-sl, line 280: no if needed, as they can use new Label("transactionType", transaction.getType())

mxtartaglia-sl (commented):
Majors concerns addressed, I am ok in splitting this into 2 separates functionalities, so long it is still compatible with the underlying metric backend semantics.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,322 @@
...
+the usage of the `NodeId` class or extract the class from the platform and define it as general API.
+Since the class does not make any sense outside the consensus node (like in block node for example) this is not a good idea.
+Another option would be to change the type of the `nodeId` to a `String` but that would end in an ugly API for any
+system next to the platform that uses the metrics module since `null` or any other default must always be used as the `nodeId`.
+Next to that all that ideas will end in having a hardcoded `nodeId` label in the metrics that will be forwarded to the monitoring system.
+
+### Definition of Label at measurement time
+
+In the given concept the labels are defined when a metric is created.
+This is how several other libraries (like micrometer) handle labels.
```
mxtartaglia-sl, line 217: This is partially true, as only the keys of micrometer tags have to be predeclared ( even though the constructor really takes pairs of key/values; the values don't matter ).
https://stackoverflow.com/a/49171202
 In mircrometer, the values also have to be specified when doing the measurement.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Motivation
+
+Today the metrics module only supports the definition of metrics without adding custom labels.
+While each metric has a hardcoded category, name, and nodeId no other custom information can be added.
+This proposal suggests adding a new concept called "metric labels" to the metrics module and
+migrate the `category`, `name`, and `nodeId` information to labels.
+While the category and name are defined as String values the nodeId is a class called `NodeId`.
+By doing so the metrics module becomes more flexible and can be used 100% independent of the
+platform module since it does not depend on the platform module's `nodeId` anymore.
```
lpetrovic05, line 21: maybe add that metrics will probably be needed by block nodes, which makes this proposal more important

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Goals
+
+- Add a new concept called "metric labels" to the metrics module
+- Migrate the `category`, `name`, and `nodeId` information to labels
```
lpetrovic05, line 26: Can a metric exist without a name? If `name` is a label, maybe it should be a special kind of mandatory label

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+## Design & Architecture
+
+This chapter describes the changes that are needed to add labels to the metrics module.
+
+### About labels
+
+Before we start with the implementation we need to define what labels are.
+Labels are key-value pairs that can be used to filter and group metrics.
+Labels are used by monitoring systems like prometheus to create more dynamic dashboards.
+For example, a metric that counts the number of requests can have a label called `method` that defines if the request was a `GET` or `POST` request.
```
lpetrovic05, line 49: so this is one metric or two? intuitively, I would say these are 2 metrics, one with a label `GET` and another with a label `POST`. this is conceptually confusing to me. can you add more detail explaining this

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+By doing so the monitoring system can create a dashboard that shows the number of requests per method.
+
+In prometheus a concrete metric is defined like this (see https://prometheus.io/docs/concepts/data_model/#notation):
+
+```
+api_http_requests_total{method="POST", handler="/messages"}
+```
+
+Here a metric is defined by a name (`api_http_requests_total`) and a set of labels (`method="POST"` and `handler="/messages"`).
```
lpetrovic05, line 58: so a set of labels defines a unique ID of a metric?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+The metrics module will be extended by a new data type called `Label`.
+A label is a key-value pair that can be defined as a `record`:
+
+```java
+record Label(@NonNull String key, @Nullable String value) {}
+``` 
+
+> [!NOTE]  
+The value of a label can be `null`.
+For our prometheus endpoint a label with a `null` value will be handled as if the label would not exist.
```
lpetrovic05, line 76: seems simpler if we don't allow null, so either there is a label or not. any reason why we would support null?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+    
+    @NonNull
+    MetricConfig withLabels(@NonNull Label... labels) {...}
+    
+    @NonNull
+    MetricConfig withLabels(@NonNull Set<Label> labels) {...}
+}
+```
+
+By adding labels the unique identifier of a metric is the key and the labels.
```
lpetrovic05, line 106: key and labels? which key?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+interface Metrics {
+    
+    // old methods:
+    // Object getValue(@NonNull String category, @NonNull String name);
+    // void remove(@NonNull String category, @NonNull String name);
+    
+    @Nullable
+    Object getValue(@NonNull String category, @NonNull String name, @NonNull Set<Label> labels);
+    
+    void remove(@NonNull String category, @NonNull String name, @NonNull Set<Label> labels);
```
lpetrovic05, line 120: if category and name are labels now, don't we need to simply supply the set of labels?

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+Since most of the filter methods that the `Metrics` interface defines today are not used we will not introduce any additional filter methods in this proposal.
+If we see that additional methods are needed we can add them later.
+
+We will have labels that are global for a `Metrics` instance.
+The best example is the `nodeId` label.
+We do not want to add the `nodeId` label to the creation of each metric.
+Instead of that we want to add the `nodeId` label to the `Metrics` instance.
+That should happen when a `Metrics` instance is created.
+The creation of a `Metrics` instance is not public api today.
+With the implementation of this proposal a `MetricsFactory` interface should be introduced that allows to create a `Metrics` instance.
```
lpetrovic05, line 136: can we define a name for these labels that are set at the metrics instance? like default labels or something

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+The definition of the api should follow our rules regarding services as defined at https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/base/service-architecture/service-architecture.md
+
+##### Examples
+
+Here are some examples how to define a metric with labels:
+
+```java
+final Label transactionTypeLabel = new Label("transactionType", "fileUpload");
```
lpetrovic05, line 154: a labels builder would be useful, if you want to specify a bunch of them, you can do it fluently

PR #13812, opened 2024-06-12
feat: Make Events accept PBJ payloads

### LINKED ISSUE #13709: Submit PBJ payloads
### Intro
Before transaction/payloads get put into an event that is being created, we currently create and self-serializable instances that contain the PBJ version of the payload. These self-serializable payloads are submitted to the EventCreator. This ticket is to change any code that interacts with the payload prior to and up to being put into an event to use the PBJ objects instead of self-serializable. The EventCreator will accept PBJ payloads and wrap each in a self-serializable instance.

### Tasks
- [ ] identify the lifecycle of transactions before they get put into events
- [ ] replace the usage of SS classes with PBJ records
- [ ] same for unit tests

### Notes
- Transactions get put into events in the constructor for `BaseEventHashedData`
- `ConsensusTransactionImpl` will eventually become the payload wrapper that is not SS
- A PBJ record of a payload is `OneOf<PayloadOneOfType>`

timo0:
Fix #13709


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
```diff
@@ -148,19 +152,35 @@ public synchronized boolean submitApplicationTransaction(@NonNull final SwirldTr
...
     }
 
+    /**
+     * Convert a transaction to a {@link ConsensusTransactionImpl}. This is a hack because {@code OneOf<>} interface
+     * can not return a size of the transaction.
+     *
+     * @param transaction the transaction to convert
+     * @return the converted transaction
+     */
+    private ConsensusTransactionImpl transactionConversion(final OneOf<PayloadOneOfType> transaction) {
```
hendrikebbers, line 176: Question: If I understand it correctly, the OneOf Contains an ENUM type to define the transaction type as key and the raw bytes of the transaction as value. If that is correct: why do you wrap the bytes in a new object and return it to get it's lenght later? Wouldn't it be better to create getSize(OneOf...) method that extracts the value of the OneOf and returns the size?

hendrikebbers: ... and now I see that you use it at other places, to to get the serialized length. Based on that it makes sense ;)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
```diff
@@ -148,19 +152,35 @@ public synchronized boolean submitApplicationTransaction(@NonNull final SwirldTr
...
+        return switch (transaction.kind()) {
+            case STATE_SIGNATURE_PAYLOAD -> new StateSignatureTransaction(transaction.as());
+            case APPLICATION_PAYLOAD -> new SwirldTransaction(transaction.as());
+            default -> throw new IllegalArgumentException("Unexpected value: " + transaction.kind());
```
hendrikebbers, line 180:
```suggestion
            default -> throw new IllegalArgumentException("Unexpected transaction type: " + transaction.kind());
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
hendrikebbers: But: If we need to transform the OneOf anyway at this position to a concrete transaction class (and as it looks this method might be called more than 1 time per transaction) doesn't it make sense to do that directly at the start and work with the concrete transaction instead of OneOf? The benefit of OneOf is less  memory + object creation at runtime. Since we now create a new object we miss all benefits. Or do I miss something?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/BaseEventHashedData.java
```diff
@@ -154,7 +157,14 @@ public BaseEventHashedData(
         this.birthRound = birthRound;
         this.birthRoundOverride = birthRound;
         this.timeCreated = Objects.requireNonNull(timeCreated, "The timeCreated must not be null");
-        this.transactions = transactions;
+        this.transactions = transactions.stream()
+                .map(t -> switch (t.kind()) {
+                    case STATE_SIGNATURE_PAYLOAD -> new StateSignatureTransaction(t.as());
+                    case APPLICATION_PAYLOAD -> new SwirldTransaction(t.as());
+                    default -> throw new IllegalArgumentException("Unexpected value: " + t.kind());
```
hendrikebbers, line 164:
```suggestion
                    default -> throw new IllegalArgumentException("Unexpected transaction type: " + t.kind());
```

hendrikebbers (commented):
I'm not completely done but before I continue I need some input:

I assume that we use `OneOf<PayloadOneOfType>` to be more compatible with services (have a better common API). Internally in platform that has zero benefit for us. Is that correct?

Since you now create `ConsensusTransaction` instances out of the `OneOf<PayloadOneOfType>` I do not see any benefit to use the more complex api `OneOf<PayloadOneOfType>` internally instead of working with `ConsensusTransaction`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
timo0: The grater picture is that all self serializable objects will go away and the "raw' OneOfs will be used.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/DefaultTransactionPool.java
```diff
@@ -42,7 +43,7 @@ public DefaultTransactionPool(@NonNull final TransactionPoolNexus transactionPoo
      * {@inheritDoc}
      */
     @Override
-    public void submitSystemTransaction(@NonNull final ConsensusTransactionImpl transaction) {
+    public void submitSystemTransaction(@NonNull final OneOf<PayloadOneOfType> transaction) {
```
lpetrovic05, line 46: maybe we should check if its actually a system transaction. Or maybe its even better if we just allow the class `StateSignaturePayload` to be submitted

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
```diff
@@ -132,12 +136,12 @@ public TransactionPoolNexus(@NonNull final PlatformContext platformContext) {
 
     /**
      * Attempt to submit an application transaction. Similar to
-     * {@link #submitTransaction(ConsensusTransactionImpl, boolean)} but with extra safeguards.
+     * {@link #submitTransaction} but with extra safeguards.
      *
      * @param transaction the transaction to submit
      * @return true if the transaction passed all validity checks and was accepted by the consumer
      */
-    public synchronized boolean submitApplicationTransaction(@NonNull final SwirldTransaction transaction) {
```
lpetrovic05, line 140: becuase of the size check, its better if we accept `Bytes` and `StateSignaturePayload` instead of `OneOf`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
lpetrovic05, line 176: for this I think we need a standalone utility that returns the transaction size. we will need it elsewhere. also, the utility should not use the wrappers, it should calculate the size the same way the wrapper would. it would also be good to add a unit test to see if the size always matches the wrapped size.

platform-sdk/swirlds-platform-core/src/testFixtures/java/com/swirlds/platform/test/fixtures/event/RandomEventUtils.java
```diff
@@ -87,14 +93,21 @@ public static BaseEventHashedData randomEventHashedDataWithTimestamp(
                         otherParent.getGeneration(),
                         otherParent.getBaseEvent().getHashedData().getBirthRound());
 
+        final List<OneOf<PayloadOneOfType>> hapiTransactions = new ArrayList<>();
```
lpetrovic05, line 96: we should not call them HAPI transactions. HAPI is a hedera concept, the platform should be independent of it

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/pool/TransactionPoolTests.java
```diff
@@ -35,18 +39,20 @@ class TransactionPoolTests {
...
             assertTrue(isPriority);
             transactionList.add(transaction);
             return true;
         });
 
         final TransactionPool transactionPool = new DefaultTransactionPool(transactionPoolNexus);
-        final ConsensusTransactionImpl transaction = mock(ConsensusTransactionImpl.class);
+        final ByteBuffer buffer = ByteBuffer.allocate(Long.BYTES);
+        buffer.putLong(RandomUtils.nextLong());
+        final OneOf<PayloadOneOfType> transaction = new OneOf<>(APPLICATION_PAYLOAD, buffer.array());
```
lpetrovic05, line 55: should be `Bytes`, not `byte[]`

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/tipset/TipsetEventCreatorTests.java
```diff
@@ -209,10 +213,20 @@ private void validateNewEvent(
             simulatedNode.tipsetWeightCalculator.addEventAndGetAdvancementWeight(descriptor);
         }
 
+        final List<OneOf<PayloadOneOfType>> hapiTransactions = Stream.of(newEvent.getTransactions())
```
lpetrovic05, line 216: should not call them hapi

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/tipset/TipsetEventCreatorTests.java
```diff
@@ -209,10 +213,20 @@ private void validateNewEvent(
...
         // We should see the expected transactions
-        assertArrayEquals(expectedTransactions, newEvent.getTransactions());
-
-        assertDoesNotThrow(() -> simulatedNode.eventCreator.toString());
+        IntStream.range(0, expectedTransactions.size()).forEach(i -> {
+            final OneOf<PayloadOneOfType> expected = expectedTransactions.get(i);
+            final OneOf<PayloadOneOfType> actual = hapiTransactions.get(i);
+            assertEquals(expected.kind(), actual.kind(), "Transaction kind " + i + " mismatch");
+            assertArrayEquals(
+                    (byte[]) expected.value(), (byte[]) actual.value(), "Transaction payload " + i + " mismatch");
```
lpetrovic05, line 226: why is it being casted to a byte[] here?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
```diff
@@ -172,22 +177,21 @@ public synchronized boolean submitApplicationTransaction(@NonNull final SwirldTr
      * @return true if successful
      */
     public synchronized boolean submitTransaction(
-            @NonNull final ConsensusTransactionImpl transaction, final boolean priority) {
+            @NonNull final OneOf<PayloadOneOfType> transaction, final boolean priority) {
 
         Objects.requireNonNull(transaction);
+        final boolean isSystem = STATE_SIGNATURE_PAYLOAD.equals(transaction.kind());
```
lpetrovic05, line 183: create util for checking isSystem, also make `Transaction.isSystem` use this

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/SwirldTransaction.java
```diff
@@ -66,6 +66,21 @@ public SwirldTransaction(final byte[] contents) {
...
 
+    /**
+     * Constructs a new transaction with no associated signatures.
+     *
+     * @param contents
+     * 		the binary content/payload of the Swirld transaction
+     * @throws IllegalArgumentException
+     * 		if the {@code contents} parameter is null or a zero-length array
+     */
+    public SwirldTransaction(final Bytes contents) {
```
lpetrovic05, line 77: deprecate other constructor, make is call this one for the time being

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/EventImplTests.java
```diff
@@ -79,29 +82,30 @@ void testTransactionIterator() {
...
     }
 
     private void verifyTransactionIterator(
-            final TransactionData data, final Set<? extends Transaction> transactionSet) {
-        for (int i = 0; i < data.transactions.length; i++) {
-            final Transaction t = data.transactions[i];
+            final TransactionData data, final Set<OneOf<PayloadOneOfType>> transactionSet) {
+        for (int i = 0; i < data.transactions.size(); i++) {
+            final OneOf<PayloadOneOfType> t = data.transactions.get(i);
+            final boolean isSystem = STATE_SIGNATURE_PAYLOAD.equals(t.kind());
```
lpetrovic05, line 101: another issystem

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/TransactionSizeUtilsTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class TransactionSizeUtilsTest {
+    private static final SplittableRandom RANDOM = new SplittableRandom();
+
+    @ParameterizedTest
+    @MethodSource("buildArguments")
+    void testSizeComparisons(final OneOf<PayloadOneOfType> payload, final SwirldTransaction swirldTransaction) {
```
lpetrovic05, line 39: also for state sigs

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/TransactionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import com.hedera.hapi.platform.event.EventPayload.PayloadOneOfType;
+import com.hedera.hapi.platform.event.StateSignaturePayload;
+import com.hedera.pbj.runtime.OneOf;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Utility class for handling PJB transactions.
+ */
+public final class TransactionUtils {
```
lpetrovic05, line 28: probably better to go with the name payloads immediately, since this will be the new term we want to use

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/TransactionUtilsTest.java
```diff
@@ -0,0 +1,91 @@
...
+                    .build();
+            arguments.add(Arguments.of(
+                    new OneOf<>(PayloadOneOfType.STATE_SIGNATURE_PAYLOAD, payload),
+                    new StateSignatureTransaction(payload)));
+        });
+
+        return arguments.stream();
+    }
+
+    private static Bytes randomBytes() {
```
lpetrovic05, line 86: there is a class called `Randotron` that we aim to use for these purposes. It has additional methods for generating various random things and it always prints the seed so that we know which seed fails. its better to use this class for all our random needs

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/pool/TransactionPoolNexus.java
timo0: done

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/EventImplTests.java
timo0: done

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/TransactionSizeUtilsTest.java
timo0: done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/TransactionUtils.java
timo0: done

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/TransactionUtilsTest.java
timo0: done

lpetrovic05 (approved):

lpetrovic05 (dismissed):

lpetrovic05 (approved):

PR #13832, opened 2024-06-13
fix: Fix hollow account completion when completing with nft transfer

### LINKED ISSUE #13831: Hollow account completion with NFT transfers fails with INVALID_ACCOUNT_ID
### Description

Hollow account completion with nft transfer fails with INVALID_ACCOUNT_ID

### Steps to reproduce

Create hollow account.
Now try to complete it by nft transfer from the hollow account signing with the hollow account will result in INVALID_ACCOUNT_ID.

### Additional context

_No response_

### Hedera network

other

### Version

51

### Operating system

None

kselveliev:
This PR fixes hollow account completion when done with nft transfer

This PR modifies ... in order to support ...
`CryptoTransferHandler` now checkSender check if account is hollow and require signature for it
`CryptoTransferSuite` - added hapi test that verifies hollow account completion
-->

**Related issue(s)**:

Fixes #13831 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
@@ -510,9 +510,13 @@ public class CryptoTransferHandler implements TransactionHandler {
         // If the sender account is immutable, then we throw an exception.
         final var key = senderAccount.key();
         if (key == null || !isValid(key)) {
-            // If the sender account has no key, then fail with INVALID_ACCOUNT_ID.
-            // NOTE: should change to ACCOUNT_IS_IMMUTABLE
-            throw new PreCheckException(INVALID_ACCOUNT_ID);
+            if (isHollow(senderAccount)) {
+                meta.requireSignatureForHollowAccount(senderAccount);
+            } else {
+                // If the sender account has no key, then fail with INVALID_ACCOUNT_ID.
+                // NOTE: should change to ACCOUNT_IS_IMMUTABLE
+                throw new PreCheckException(INVALID_ACCOUNT_ID);
+            }
         } else if (!nftTransfer.isApproval()) {
             meta.requireKey(key);
         }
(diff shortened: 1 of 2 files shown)


stoyanov-st (approved):
LGTM 💯

bilyana-gospodinova (approved):
LGTM

MiroslavGatsanoga (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoTransferHandler.java
```diff
@@ -510,9 +510,13 @@ private void checkSender(
         // If the sender account is immutable, then we throw an exception.
         final var key = senderAccount.key();
         if (key == null || !isValid(key)) {
```
Neeharika-Sompalli, line 512: This condition if `!isValid(` can be inside else for clarity

kselveliev: With this suggestion do we expected the end version of the code to look like this? ->
```
final var key = senderAccount.key();
if (isHollow(senderAccount)) {
    meta.requireSignatureForHollowAccount(senderAccount);
} else {
    if (key == null || !isValid(key)) {
        // If the sender account has no key, then fail with INVALID_ACCOUNT_ID.
        // NOTE: should change to ACCOUNT_IS_IMMUTABLE
    } else if (!nftTransfer.isApproval()) {
        meta.requireKey(key);
    }
}

```

Neeharika-Sompalli: yes Agreed @kselveliev

Neeharika-Sompalli: Oh actually. I take it back. This makes sense also.

Neeharika-Sompalli (approved):
LGTM! Thanks @kselveliev

PR #13833, opened 2024-06-13
chore: Upgrade Netty to Latest Stable Release Version 4.1.110

### LINKED ISSUE #13679: Upgrade Netty
Upgrade Netty to version 4.1.110

thomas-swirlds-labs:
**Description**:

- [x] Update Netty Version from 4.1.87 to 4.1.110
- [x] Update dependent grpc versions due to breaking change after release 101: https://github.com/grpc/grpc-java/issues/10665 
- [x] Remove requirement to exchange dependencies from `io.grpc.*` to `io.helidon.grpc:io.grprc`
- [x] Fixes Hapi Test failures

**Related issue(s)**:

Fixes #13679


nathanklick (approved):
Needs review and approval by @jjohannes

nathanklick (changes requested):
Also needs review and approval by @jsync-swirlds

jjohannes (dismissed):

jsync-swirlds:
The changes, as such, appear reasonable.  The more significant question is whether this will encounter the same stability issues we encountered previously.
Two questions:
1. Have you checked the final output package produced with this change to ensure it doesn't include duplicate (and obsolete) versions of some of our core packages?
   That was one of the more concerning problems we encountered with the Google dependencies previously, the io.grpc libraries referred to obsolete versions of some libraries to be compatible with Java 7 and Java 8, and this conflicted with the versions we need to use, which created instability in the software.
   If you've unpacked these jars and verified that we don't have any unintended new dependencies that may create problems for us, then I don't see a major concern with this change _for now_.
1. There is [some work](https://github.com/hashgraph/pbj/pull/259) @rbair23 has done to create a gRPC implementation in PBJ _that will plug into Helidon_. That work, if continued, _may_ require that we continue using the Helidon server modules at a minimum.  Have we discussed this change broadly enough to determine if we need to make adjustments to avoid conflicts with that work?

netopyr:
> 2. There is [some work](https://github.com/hashgraph/pbj/pull/259) @rbair23 has done to create a gRPC implementation in PBJ _that will plug into Helidon_. That work, if continued, _may_ require that we continue using the Helidon server modules at a minimum.  Have we discussed this change broadly enough to determine if we need to make adjustments to avoid conflicts with that work?

I have talked about this with @rbair23. We do not plan to switch to Helidon in the near term, but we will consider it if the block node team's experience is positive.

jsync-swirlds (dismissed):
Discussed dependencies separately, and the result looks good.

nathanklick (dismissed):
Needs reapprovals from @jsync-swirlds and @jasperpotts at a minimum.

hedera-node/hedera-mono-service/src/main/java/module-info.java
jsync-swirlds, line 1: Probably a conflict leftover; this file should be removed.

jsync-swirlds (dismissed):
Looks good, just one small question about the empty module-info file.

jsync-swirlds (approved):

netopyr (approved):
LGTM - thanks @thomas-swirlds-labs

JeffreyDallas (approved):

Issue #13848, opened 2024-06-14
Analyze account property "inheritance" when finalizing a hollow account


david-bakin-sl:
When finalizing a hollow account by doing a contract create from an EOA the question arises:  What are the newly-finalized account's properties?  Things like auto renew account, staking, etc., but especially the presence/absence of an admin key, and if present, what is it to be?

Need to review this as the answers aren't clear at this point.

Consider:
a) who's doing it?  e.g., what's the POV of an ethereum developer?
b) what did we do in mono, and was it correct anyway?
c) what flexibility (or lack of it) do various settings mean for the contract account going forward?  (is there lock in to things that can't be changed?)

david-bakin-sl:
More discussion on internal slack: https://swirldslabs.slack.com/archives/C056T02B6UT/p1718327321632719

stoyanov-st:
We are not setting staking data or auto renew account. The admin key is zeroed which makes the contract immutable, so no flexibility whatsoever.

PR #13849, opened 2024-06-14
fix: address null pointer exception in hollow contract flow

lukelee-sl:
**Description**:
Due to mono service fidelity, the parent contracts attributes were copied for externalizing the record in the hollow creation flow for contracts.  If the contract frame does not have a contract parent then handle with default values.

**Checklist**


tinker-michaelj (approved):
Makes sense to me, tyvm @lukelee-sl !

david-bakin-sl (approved):
Looks clean.  (Just want to reference #13848 here which asks us to look again at the inheritance issue.)

LGTM

PR #13853, opened 2024-06-14
feat: HIP-904: Sender should cover fees for auto associations V2

bilyana-gospodinova:
This PR adds the needed changes in the calculations needed for HIP-904. Now the sender of a crypto transfer is charged for the auto association slots if the `unlimitedAutoAssociations` flag is enabled.


kselveliev:
Closing this one, there is a new pr with the changes in this one which is opened against develop -> https://github.com/hashgraph/hedera-services/pull/13907

PR #13866, opened 2024-06-14
docs: Glossary for Technical Terms

### LINKED ISSUE #13077: Glossary needed for technical terms
### Problem

I'm always frustrated when ... 

I run across a new Hedera-specific technical term (or a term which obviously has some kind of Hedera-specific "twist") and I _have no idea what it means_ nor do I have any clue _where to look for the definition_.

And frequently these terms have important nuances.

For example, today I ran across "civilian accounts".  WTH?  We distinguish between military and non-military accounts?  Of course not.  It's something else that we use only in test contexts, but there's no definition anywhere.

And in fact, in the code, we have a _lot_ of different "kinds" of accounts based on properties and or usage-at-the-moment.  It'll take you only a minute to think of at least 5 different terms for accounts, each with subtle but important (frequently security related!) differences, and there are more than 5 I can assure you.

And "accounts" is only one aspect of the much bigger issue.

### Solution

We need a glossary of our technical terms in the repo, sufficient to answer all of these questions.

### Alternatives

Keep bugging more knowledgable developers on our internal slack.  (This works for me, but doesn't help our open source community, nor is it scalable.)

thomas-swirlds-labs:
**Description**:

This is an initial attempt at capturing and documenting some of the most commonly occurring Hedera-specific technical
 terms that can be confusing to our open source dev community. 

This is a living document and will be updated, modified and added to over time. 

**Related issue(s)**:

Fixes #13077


david-bakin-sl:
Do you want to put this at top-level in the repo or should it go under `hedera-node/docs` with all the other documentation?

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Address Book**: The data structure that contains data for each consensus node in the network, including node ID, alias, IP address, and weight.
+
+## Account Aliases
+**Account Aliases**: A user-friendly identifier that can be classified as either a public key or an Ethereum Virtual Machine (EVM) address. Instead of inputting a long string of characters, users can use simpler and more memorable aliases.
+
+## Application Accounts
+**Application Accounts:** Used by developers and users interacting with applications deployed on Hedera. These accounts enable users to interact with the functionality provided by dApps, such as gaming, finance, or other decentralized services.
+
+## Auto Associations
+**Auto Associations**: The ability of accounts to automatically associate with tokens. An auto association slot is one or more slots you approve that allow tokens to be sent to your contract without explicit authorization for each token type. If this property is not set, you must approve each token before it is transferred to the contract for the transfer to be successful via the `TokenAssociateTransaction` in the SDKs.
```
david-bakin-sl, line 63: This is [HIP-23](https://hips.hedera.com/hip/hip-23) but there have been subsequent modifications, don't have the full set of HIPs for those.  But [HIP-904](https://hips.hedera.com/hip/hip-904) - currently in progress, not deployed - is possibly the most significant change.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Auto-Renew Accounts
+**Auto-Renew Accounts**: A feature designed to automatically renew the lifecycle of certain entities (like accounts, files, smart contracts, topics, or tokens) by funding their renewal fees. This feature is particularly useful for entities that require continuous operation over long periods, as it automates the process of keeping these entities active by periodically charging a linked auto-renew account.
```
david-bakin-sl, line 66: [HIP-16](https://hips.hedera.com/hip/hip-16) and [HIP-672](https://hips.hedera.com/hip/hip-23) for this, but should be noted that we don't currently charge rent for anything nor do we expire anything!  Nevertheless, the facility is there waiting to be re-tested and turned on so we must always be cognizant of in design/code/test.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+---
+
+## BLS
+**Boneh-Lynn-Shacham (BLS)**: A paired-elliptic-curve based signature algorithm with attractive characteristics for aggregate signatures (signatures produced by aggregating many individual signatures, and verifiable with a single public key).  BLS comes from the names of the authors of the initial research paper (Dan Boneh, Ben Lynn, Hovav Shacham).
```
david-bakin-sl, line 71: As long as you name the authors (great!) link the paper at [the author's site](https://hovav.net/ucsd/papers/bls04.html)?

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Block Signature**: Aggregated (BLS) signature of the hash of Block stream merkle root. Appended to the end of a block stream it is used to verify the block was produced by the majority weight of the network. These signatures will also be used for Invalid State Detection.
+
+## Block Stream
+**Block Stream**: A fully verifiable stream of data consisting of individual blocks corresponding to consensus rounds. Each block contains all events, transactions, and state changes that occurred within one consensus round. A block is transmitted from consensus node to block nodes as a stream of individual block items, which form a merkle tree that is verified by a Block Signature. Each block is connected to the previous block by a Block Hash which, together form a block chain.
+
+---
+
+
+## Civilian Accounts
+**Civilian Accounts**:  User accounts that are not associated with any special roles or permissions beyond standard user functionality. By contrast, System Accounts and Node Accounts are specialized accounts that have specific roles and responsibilities within the infrastructure and governance of the Hedera network.
```
david-bakin-sl, line 86: Note, maybe, that term seems to be confined to tests?  I've certainly never heard it used in any other context.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Config.txt
+**Config.txt**: A text file loaded by the platform on startup if no valid state is found on disk. This file contains the address book to use when starting from genesis.
+
+## Congestion Pricing
```
david-bakin-sl, line 91: Can refer to `docs/fees/automated-congestion-pricing.md`, also maybe `docs/design/app/fees.md`

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Congestion Pricing**: A mechanism designed to manage network congestion by dynamically adjusting transaction fees based on network demand. The primary goal of congestion pricing is to discourage excessive network usage during peak times.
+
+## Consensus Time
```
david-bakin-sl, line 94: Maybe refer to `docs/design/services/consensus-service/consensus-service.md`?

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Consensus Time**: The timestamp assigned to a transaction once it has reached consensus across the network. This timestamp indicates the moment when the network agrees on the order and validity of the transaction, ensuring that all participants have a consistent view of the transaction sequence.
+
+---
+
+## Delegate Call
```
david-bakin-sl, line 99: Could potentially refer to any of a number of articles, how about maybe [this one](https://medium.com/@solidity101/understanding-call-delegatecall-and-staticcall-primitives-in-ethereum-smart-contracts-dfff21caa727)?

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Delegate Call**: Used to call a function in another contract, but with the context of the calling contract. This allows for the execution of functions in one contract as if they were part of another but retains the original caller’s context (msg.sender and msg.value).
+
+---
+
+## EOA
+**EOA (Externally Owned Accounts)**: Accounts that are controlled by private keys and are not associated with smart contracts. These accounts are used to send and receive transactions on the network and interact with smart contracts. EOAs are typically owned by individual users and are used to manage their assets and participate in the network.
```
david-bakin-sl, line 105: Can contrast to "Token Account" and "Contract Account".

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+---
+
+## Address Book
+**Address Book**: The data structure that contains data for each consensus node in the network, including node ID, alias, IP address, and weight.
+
+## Account Aliases
+**Account Aliases**: A user-friendly identifier that can be classified as either a public key or an Ethereum Virtual Machine (EVM) address. Instead of inputting a long string of characters, users can use simpler and more memorable aliases.
+
+## Application Accounts
```
david-bakin-sl, line 59: This is a subset of "EOA" I think?  It's not a "technical term" in the same way ... (not that it shouldn't be defined here ...)

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Delegate Call**: Used to call a function in another contract, but with the context of the calling contract. This allows for the execution of functions in one contract as if they were part of another but retains the original caller’s context (msg.sender and msg.value).
+
+---
+
+## EOA
+**EOA (Externally Owned Accounts)**: Accounts that are controlled by private keys and are not associated with smart contracts. These accounts are used to send and receive transactions on the network and interact with smart contracts. EOAs are typically owned by individual users and are used to manage their assets and participate in the network.
+
+---
+
+## Hollow Accounts
```
david-bakin-sl, line 109: Refer to your definition of "Account Alias", above.  And maybe reference HIP-32 and HIP-583 here too.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+---
+
+## Leaky Tests
+**Leaky Tests**: Are unit or other tests that should not be run in parallel because they can “leak” some of their state or data out into the global state.
+
+---
+
+## Precompiles
+**Precompiles**: Contracts that are compiled and deployed on the blockchain before they are executed. They are used to optimize the execution of certain operations that are computationally expensive or require a large amount of gas to execute. Precompiled contracts are typically used to perform cryptographic operations, such as hashing, encryption, and decryption, that are required for the operation of the blockchain.
```
david-bakin-sl, line 120: This is not quite it.  In Ethereum chains they are things that _behave as if they were contracts_ deployed at specific addresses (in Ethereum, they are very low addresses, `0x00` .. `0x0F` or so).  They aren't _implemented_ as contracts because that's far too slow - they're built-in to the code of the Ethereum EVM, which checks to see if the address you're calling is one of the precompiles and then just runs the related code.
* With one exception: The EIP-4788 "beacon roots contract" which _is_ a real contract deployed at a specific address ...

We call our equivalent "system contracts".  But that's a relatively recent decision we made and thus you'll find, in our codebase, a bunch of references to "precompiles" - including in names of packages/classes/methods - that are still there.  Plus in older documentation.

Anyway, maybe refer to "System Contracts" you defined below in order to provide the contrast.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+---
+
+## Security Model V2
```
david-bakin-sl, line 124: Please definitely refer to this [official blog post](https://hedera.com/blog/hedera-smart-contract-service-security-model-update).

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+**Security Model v2**: Hedera introduced the HSCS Security Model v2 to enhance the security of its network building with improved features to better safeguard against potential vulnerabilities and attacks.  By incorporating enhanced cryptographic techniques, robust consensus mechanisms, decentralized governance, and continuous monitoring.
+Key features include - Restrictions on Smart contracts Storage; Prohibition on Delegate Calls to System Smart Contracts; Limited Access to EOAs' Storage; Token Allowance Requirement for Balance Changes
+
+## Sidecars
+**Sidecars**: In Services, a sidecar refers to additional records that are created alongside the main transaction records. These sidecar records provide more detailed information about the transactions, making it easier to debug and understand the state changes that occur as a result of the transactions.
```
david-bakin-sl, line 129: Might note that these are related to the current _record streams_ (which themselves could use a glossary entry) - and that both record stream files and their related sidecars will be obsoleted when block streams replace them.

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Synthetic Transactions
```
david-bakin-sl, line 131: An example or two might be useful.  E.g., it's a couple of synthetic transactions that causes the network to freeze at a specific time so it can be upgraded.

Also, confusing to me, I see uses of "synthetic transaction" in the docs where I would have expected to see "child transaction".  I might have to learn more about what's intended by this.

david-bakin-sl (dismissed):
First cut at review.  Will come back later for last entries.

I highly applaud this effort!  Thank you!  My comments are in no way _criticism_!  Or even "required changes" - hence the approval checkmark.  I'm just hoping you'll find them useful suggestions for improvement of this new glossary.

jeromy-cannon (approved):

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+## T
+- [Token Allowance](#token-allowance)
+
+## W
+- [Weibar](#weibar)
+
+---
+
+## Address Book
+**Address Book**: The data structure that contains data for each consensus node in the network, including node ID, alias, IP address, and weight.
```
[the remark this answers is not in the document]
jsync-swirlds: The address book is not stored as a system file (it's stored as a file on disk at the moment, that extends to data in state that's written to a file on disk during prepare upgrade in DAB phase 2, and purely in state in DAB phase 3).

There is an "address book" file (two actually) in files 0.0.101 and 0.0.102.  Those two files contain _more_ information than what's in the "network address book", and are manually managed (as I understand it) by devops.

In my opinion, the network address book _should_ be defined as the in-state structure that contains information for _all nodes_ in the network, not just consensus, but also block nodes, future validator nodes, perhaps even mirror nodes (although that might not be valuable, hard to be certain this early).

Glossary.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Account Aliases
+**Account Aliases**: A user-friendly identifier that can be classified as either a public key or an Ethereum Virtual Machine (EVM) address. Instead of inputting a long string of characters, users can use simpler and more memorable aliases.
```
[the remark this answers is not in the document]
jsync-swirlds: I think this is misunderstanding aliases.  Aliases are _always_ public key values (as currently defined), and are mostly 33 byte EVM address values.  The "triplet" form (e.g. 0.0.18273892) is the simpler and easier to "type".
An alias is _not_ intended to be user-friendly, but to offer a _smart-contract friendly_ identifier.

Glossary.md
jsync-swirlds: Also, not just contracts, but all accounts, must "associate" (whether automatically or manually) to a token before that token can be received, held, or sent by that account.

HIP-904 allows the number of "automatic association" slots to be `-1` which represents _infinite_ slots.

Glossary.md
jsync-swirlds: There are _two_ BLS, one for the signature algorithm, and another for the paired curve discovered to use with that algorithm.  Interestingly the authors are _not_ the same (except for Ben Lynn).

Here we should specify `BLS Algorithm` to differentiate from the `BLS Curve Family`.

Glossary.md
jsync-swirlds: Also, the Node Account or System Account are not "specialized", the accounts are the same as any other account, we just _treat_ them differently based on configuration values (System Accounts) and address book information (Node Accounts).

Note also that a ContractID works like an account identifier (and _is_ an account ID) because a smart contract is an account with certain fields set to contract-specific values.

Glossary.md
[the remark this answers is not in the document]
thomas-swirlds-labs: Incorporated feedback to glossary. Thanks!

Glossary.md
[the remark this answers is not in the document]
thomas-swirlds-labs: Incorporated feedback 👍

thomas-swirlds-labs:
@david-bakin-sl and @jsync-swirlds I've the PR with your suggestions. Much appreciated!

thomas-swirlds-labs:
> Do you want to put this at top-level in the repo or should it go under `hedera-node/docs` with all the other documentation?

For now I have put it into `/docs` at @netopyr suggestion and opened another issue to address some inconsistencies and duplication between files living under `/docs` and `hedera-node/docs` https://github.com/hashgraph/hedera-services/issues/13917

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+## Address Book
+**Address Book**: The _**network**_ address book is data structure that contains information (such as node ID, alias, IP address, and weight) for all nodes in the network, 
+not just consensus, but also block nodes, future validator nodes and more.
+The address book is stored as a file on disk at the moment but will be stored purely in state as part of Dynamic Address Book phase 3.	Disambiguation: There is also an "address book" in files 0.0.101 and 0.0.102 which is different to the "network address book" which are manually managed by devops.
+
+
+## Account Aliases
+**Account Aliases**: These are identifiers which are public key values, and are mostly 33 byte EVM address values. The "triplet" form (e.g. 0.0.18273892) is the simpler and easier to "type".
```
Neeharika-Sompalli, line 61: The aliases can be protobuf serialized keys of ED25519 or ECDSA Keys or EVM address

jsync-swirlds: That's correct, there are more options than just a public key.

An EVM address is a 20 byte keccak-256 of the ECDSA account key.
The "synthetic" EVM address (built from the account ID) is a bit weird as it's not stored as an alias, but can be used in the alias field of the `AccountID` message.
That bit of semantic gymnastics is courtesy of wanting EVM addresses but not wanting to implement the multiple-EVM-address features in HIP-631.

Probably way too much for a "glossary" (and not yet final or published), but there is good detail [here](https://github.com/hashgraph/hedera-protobufs/blob/pbj-storage-spec-review/documents/api/services/basic_types.md#proto-AccountID)

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Auto Associations
+**Auto Associations**: The ability of accounts to automatically associate with tokens. An auto association is one or more slots you approve that allow tokens to be sent to your contract or account without explicit authorization for each token type. If this property is not set, you must approve each token manually (via the `TokenAssociateTransaction` in the SDKs) before it can be received, held, or sent by that account.
+https://hips.hedera.com/hip/hip-904 - currently in progress, will represent a significant change which allows for the number of "automatic association" slots to be set to -1 which represents infinite slots.
+
+
+## Auto-Renew Accounts
+**Auto-Renew Accounts**: A feature designed to automatically renew the lifecycle of certain entities (like accounts, files, smart contracts, topics, or tokens) by funding their renewal fees. This feature is particularly useful for entities that require continuous operation over long periods, as it automates the process of keeping these entities active by periodically charging a linked auto-renew account.
```
Neeharika-Sompalli, line 78: > A feature designed to automatically renew the lifecycle of certain entities
If we are describing feature we should probably just say `AutoRenewal`

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+## Block Signature
+**Block Signature**: Aggregated (BLS) signature of the hash of Block stream merkle root. Appended to the end of a block stream it is used to verify the block was produced by the majority weight of the network. These signatures will also be used for Invalid State Detection.
+
+## Block Stream
+**Block Stream**: A fully verifiable stream of data consisting of individual blocks corresponding to consensus rounds. Each block contains all events, transactions, and state changes that occurred within one consensus round. A block is transmitted from consensus node to block nodes as a stream of individual block items, which form a merkle tree that is verified by a Block Signature. Each block is connected to the previous block by a Block Hash which, together form a block chain.
+
+---
+
+## Child Transaction
+**Child Transaction**: A [Synthetic Transaction](#synthetic-transactions) that the EVM or Services initiates as a result of fulfilling a user-initiated transaction (e.g. [Hollow Accounts](#hollow-accounts) creation, or a scheduled execution).
```
Neeharika-Sompalli, line 98: nit: We can add auto-account creation also as example.

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+## Config.txt
+**Config.txt**: A text file loaded by the platform on startup if no valid state is found on disk. This file contains the address book to use when starting from genesis.
+
+## Congestion Pricing
+**Congestion Pricing**: A mechanism designed to manage network congestion by dynamically adjusting transaction fees based on network demand. The primary goal of congestion pricing is to discourage excessive network usage during peak times. Refer to [Congestion Pricing](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/fees/automated-congestion-pricing.md) and [Fees](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/app/fees.md).
+
+## Consensus Time
+**Consensus Time**: The timestamp assigned to a transaction once it has reached consensus across the network. This timestamp indicates the moment when the network agrees on the order and validity of the transaction, ensuring that all participants have a consistent view of the transaction sequence. See [Consensus Service](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/services/consensus-service/consensus-service.md) for more information.
```
Neeharika-Sompalli, line 110: Consensus Service is not related to Consensus timestamp. Consensus timestamp is assigned by platform

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+---
+
+## Delegate Call
+**Delegate Call**: Used to call a function in another contract, but with the context of the calling contract. This allows for the execution of functions in one contract as if they were part of another but retains the original caller’s context (msg.sender and msg.value). See [Delegate Call](https://medium.com/@solidity101/understanding-call-delegatecall-and-staticcall-primitives-in-ethereum-smart-contracts-dfff21caa727) for more information.
+ 
+---
+
+## EOA
+**EOA (Externally Owned Accounts)**: Accounts that are controlled by private keys and are not associated with smart contracts. These accounts are used to send and receive transactions on the network and interact with smart contracts. In contrast to contract accounts or token accounts, EOAs are typically owned by individual users and are used to manage their assets and participate in the network.
```
Neeharika-Sompalli, line 120: You can mention they are #civilian-accounts or #Application-accounts previously mentioned

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+---
+
+## Hollow Accounts
+**Hollow Accounts**: An account that has been created with an account number and alias but lacks an account key. This term is specifically used to describe accounts generated through the Auto Account Creation feature, which allows applications to create user accounts instantly, even without an internet connection, by assigning an account alias. See also [Account Aliases](#account-aliases), https://hips.hedera.com/hip/hip-32 and https://hips.hedera.com/hip/hip-583.
```
Neeharika-Sompalli, line 125:
```suggestion
**Hollow Accounts**: An account that has been created with an account number and alias but lacks an account key. This term is specifically used to describe accounts generated through the Auto Account Creation feature when the alias is a EVM address, which allows applications to create user accounts instantly, even without an internet connection, by assigning an account alias. See also [Account Aliases](#account-aliases), https://hips.hedera.com/hip/hip-32 and https://hips.hedera.com/hip/hip-583. 
```

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+**Delegate Call**: Used to call a function in another contract, but with the context of the calling contract. This allows for the execution of functions in one contract as if they were part of another but retains the original caller’s context (msg.sender and msg.value). See [Delegate Call](https://medium.com/@solidity101/understanding-call-delegatecall-and-staticcall-primitives-in-ethereum-smart-contracts-dfff21caa727) for more information.
+ 
+---
+
+## EOA
+**EOA (Externally Owned Accounts)**: Accounts that are controlled by private keys and are not associated with smart contracts. These accounts are used to send and receive transactions on the network and interact with smart contracts. In contrast to contract accounts or token accounts, EOAs are typically owned by individual users and are used to manage their assets and participate in the network.
+
+---
+
+## Hollow Accounts
```
Neeharika-Sompalli, line 124: We can add Auto-Creation as a section also, since we mentioned it 🙂 . Need not be in the same PR.

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+## System Contracts
+**System Contracts**:  Smart contracts that have a permanent address in the network, meaning they will always be available at the same contract address. One example is the Token Service System Contract which allows smart contracts on Hedera to interact with the Hedera Token Service, offering functionalities like token creation, burning, and minting through the EVM
+
+## System Transaction
+**System Transaction**: A transaction created by the platform (not services) software. An example of which is a node's signature on a state. Inside an event, every transaction has a flag saying whether it is a system transaction or not.
+admin transaction-a HAPI transaction signed by the council. Such as freeze.
```
Neeharika-Sompalli, line 158: We can delete admin transaction here. I think we mentioned already in this file

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+## Token Allowance
+**Token Allowance**: A feature that allows an account (the owner) to authorize another account (the spender) to transfer tokens on its behalf up to a specified limit. This mechanism is similar to the ERC-20 allowance mechanism used in Ethereum.
+The amount of tokens that a user has authorized a smart contract to spend on their behalf. This allowance is set by the user and can be modified or revoked at any time. It allows smart contracts to spend tokens on behalf of the user without requiring explicit approval for each transaction.
```
Neeharika-Sompalli, line 163:
```suggestion
The amount of tokens that a user has authorized another user or smart contract to spend on their behalf. This allowance is set by the user and can be modified or revoked at any time. It allows the spender to spend tokens on behalf of the owner without requiring explicit approval for each transaction.
```

Neeharika-Sompalli (dismissed):
Looks like a great start. Thanks @thomas-swirlds-labs. 🎉  I had few comments, they can be addressed in separate PR if needed 🙂

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+## W
+- [Weibar](#weibar)
+
+---
+
+## Address Book
+**Address Book**: The _**network**_ address book is data structure that contains information (such as node ID, alias, IP address, and weight) for all nodes in the network, 
+not just consensus, but also block nodes, future validator nodes and more.
+The address book is stored as a file on disk at the moment but will be stored purely in state as part of Dynamic Address Book phase 3.	Disambiguation: There is also an "address book" in files 0.0.101 and 0.0.102 which is different to the "network address book" which are manually managed by devops.
```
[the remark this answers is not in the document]
SimiHunjan: I am not sure the community understands what "dynamic address book phase 3" means. I don't think we have a HIP yet for that. I guess maybe when it becomes available we could link it for reference. Maybe we should define what a "Dynamic Address Book" is and link to that. 

Suggestion:

**Dynamic Address Book**: The envisioned state where the system can create, update, or delete nodes in the address book through Hedera transactions. In this future state, rather than manually updating nodes during predefined upgrade periods, the address book can be managed in real-time with the following transaction types:

[NodeCreate](https://github.com/hashgraph/hedera-protobufs/blob/main/services/node_create.proto): Adds a new node to the address book.
[NodeUpdate](https://github.com/hashgraph/hedera-protobufs/blob/main/services/node_update.proto): Modifies the details of an existing node.
[NodeDelete](https://github.com/hashgraph/hedera-protobufs/blob/main/services/node_delete.proto): Removes a node from the address book.

The actions of the transactions will still take effect at the node upgrade window [if I understood correctly].

docs/glossary.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+## Account Aliases
+**Account Aliases**: These are identifiers which are public key values, and are mostly 33 byte EVM address values. The "triplet" form (e.g. 0.0.18273892) is the simpler and easier to "type".
+An alias is not intended to be user-friendly, but instead to offer a smart-contract friendly identifier. See https://hips.hedera.com/hip/hip-32 and https://hips.hedera.com/hip/hip-583 for more information.
+
+
+## Admin Transaction 
+**Admin Transaction**: A HAPI transaction which has been signed by the council, for example a Freeze Transaction
+
```
SimiHunjan, line 67: Suggestion:

A HAPI transaction requires the transaction fee payer to be a privileged system account authorized to handle admin transactions. A privileged system account is defined as an account with an entity ID of 0.0.1000 or lower. Each type of admin transaction specifies which accounts can serve as authorized transaction fee payers. Regular users in the community are not permitted to authorize these types of transactions. Some privileged accounts, like account 0.0.2 or 0.0.58, are owned by the Hedera Council. This means their keys are on the account and their signatures would be required if the account is being modified or used to pay for admin transactions. 

Examples of Admin Transactions include the following:
- Freeze Transactions
- System Delete
- System Undelete 
- Address Book File 0.0.101/102 updates
- Exchange Rate File 0.0.111 updates
- Dynamic Properties File 0.0.121 updates

Not sure if this is too much. Feel free to edit as you see fit.

jeromy-cannon (dismissed):

docs/glossary.md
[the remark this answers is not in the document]
thomas-swirlds-labs: > These might work better as definition lists, by the way. Here's an example (the spacing was off also, so it's a good example).

Thanks @jsync-swirlds. I will implement this formatting in a separate PR 👍 

> I am not sure the community understands what "dynamic address book phase 3" means. I don't think we have a HIP yet for that. I guess maybe when it becomes available we could link it for reference. Maybe we should define what a "Dynamic Address Book" is and link to that.

Good suggestion, have added. Thanks @SimiHunjan.

docs/glossary.md
```diff
@@ -0,0 +1,175 @@
...
+## W
+- [Weibar](#weibar)
+
+---
+
+## Address Book
+**Address Book**: The _**network**_ address book is data structure that contains information (such as node ID, alias, IP address, and weight) for all nodes in the network, 
+not just consensus, but also block nodes, future validator nodes and more.
+The address book is stored as a file on disk at the moment but will be stored purely in state as part of Dynamic Address Book phase 3.	Disambiguation: There is also an "address book" in files 0.0.101 and 0.0.102 which is different to the "network address book" which are manually managed by devops.
```
jsync-swirlds, line 57: Just a note, your spacing is still off here (there's a tab in place of a space after the first sentence).
Tip: setting your tab width in github to 16 makes tabs super easy to spot.

docs/glossary.md
thomas-swirlds-labs: Thanks @Neeharika-Sompalli 🎉 Fixed it

docs/glossary.md
thomas-swirlds-labs: Thanks @jsync-swirlds. I have opened a separate issue https://github.com/hashgraph/hedera-services/issues/14016 to address

docs/glossary.md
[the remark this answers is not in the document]
thomas-swirlds-labs: Will address styling, formatting and definition lists in a future PR

Neeharika-Sompalli (approved):

JeffreyDallas (approved):

docs/glossary.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Sidecars
+
+**Sidecars**: In Services, a sidecar refers to additional records that are created alongside the
+main transaction records. These sidecar records provide more detailed information about the
+transactions, making it easier to debug and understand the state changes that occur as a result of
+the transactions. Note that these are related to the current record stream implementation and will
+be replaced by block streams in the future.
+
+## Synthetic Transaction
```
david-bakin-sl, line 274: Unfortunately my previous comment to you was wrong.  This definition is true but is partial, it turns out that all "Child Transactions" are considered "Synthetic Transactions" too.

david-bakin-sl (approved):
Gosh I like this!

Now that I've had _my_ 2¢ I suggest you quickly address any other comments and then merge this in!  No need to delay, plus the sooner it's in the sooner other people can party on it (add their own technical terms that they've needed to know and think others would).  This is a great resource you started!

david-bakin-sl:
oops! already merged!  excellent!  I'll start adding to it shortly!  THANKYOU!

PR #13875, opened 2024-06-15
feat: remove attributes when externalizing finalized hollow contract

lukelee-sl:
**Description**:
When finalizing a contract create onto a hollow account, for externalization purposes, only set the adminKey to be the contract key.


tinker-michaelj (approved):
LGTM, tyvm @lukelee-sl !

PR #13881, opened 2024-06-17
build: Make sure also 'reports' has all expected lifecycle tasks

jjohannes:
Fixes this publishing error:

>Could not determine the dependencies of task ':closeSonatypeStagingRepository'.
> Task with path ':reports:releaseMavenCentral' not found in root project 'hedera-services'.

**Description**:

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.reports.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.reports.gradle.kts
@@ -17,6 +17,7 @@
 plugins {
     id("jvm-ecosystem")
     id("jacoco-report-aggregation")
+    id("com.hedera.gradle.lifecycle")
     id("com.hedera.gradle.repositories")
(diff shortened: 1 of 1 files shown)


gradle/plugins/src/main/kotlin/com.hedera.gradle.reports.gradle.kts
```diff
@@ -17,6 +17,7 @@
 plugins {
     id("jvm-ecosystem")
     id("jacoco-report-aggregation")
+    id("com.hedera.gradle.lifecycle")
```
rbarkerSL, line 20: This adds in the `releaseMavenCentral` task which the `closeSonatypeStagingRepository` task depends on.

For this reports subproject `releaseMavenCentral` is a no-op.

rbarkerSL (approved):
LGTM pending workflow completion

Issue #13867, opened 2024-06-14
Reconnect learner memory leak


lpetrovic05:
A memory leak can occur when a node acts as a reconnect learner and receives a state that lacks sufficient signatures. At this point, an exception is thrown and the state that was received does not get released.

```
2024-06-13 23:16:30.113 381      ERROR EXCEPTION        <<reconnect: reconnect-controller>> ReconnectLearnerThrottle: Error while receiving a signed state! {"causeOfFailure":"ERROR"} [com.swirlds.logging.legacy.payload.ReconnectFailurePayload]
com.swirlds.platform.reconnect.ReconnectException: com.swirlds.platform.state.signed.SignedStateInvalidException: Signed state lacks sufficient valid signatures. This state has 0 valid signatures representing 0/481 weight
	at com.swirlds.platform.reconnect.ReconnectLearner.execute(ReconnectLearner.java:161) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.ReconnectHelper.reconnectLearner(ReconnectHelper.java:133) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.ReconnectHelper.receiveSignedState(ReconnectHelper.java:111) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.ReconnectController.executeReconnect(ReconnectController.java:114) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.ReconnectController.run(ReconnectController.java:95) ~[swirlds-platform-core-0.49.7.jar:?]
	at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]
Caused by: com.swirlds.platform.state.signed.SignedStateInvalidException: Signed state lacks sufficient valid signatures. This state has 0 valid signatures representing 0/481 weight
	at com.swirlds.platform.state.signed.SignedState.throwIfNotVerifiable(SignedState.java:590) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.DefaultSignedStateValidator.validate(DefaultSignedStateValidator.java:54) ~[swirlds-platform-core-0.49.7.jar:?]
	at com.swirlds.platform.reconnect.ReconnectLearner.execute(ReconnectLearner.java:157) ~[swirlds-platform-core-0.49.7.jar:?]
	... 5 more
```

PR #13882, opened 2024-06-17
fix: reconnect learner memory leak

lpetrovic05:
backport to 51 of #13867

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/reconnect/ReconnectLearner.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/reconnect/ReconnectLearner.java
@@ -152,15 +152,21 @@ public class ReconnectLearner {
     @NonNull
     public ReservedSignedState execute(@NonNull final SignedStateValidator validator) throws ReconnectException {
         increaseSocketTimeout();
+        ReservedSignedState reservedSignedState = null;
         try {
             receiveSignatures();
-            final ReservedSignedState reservedSignedState = reconnect();
+            reservedSignedState = reconnect();
             validator.validate(reservedSignedState.get(), addressBook, stateValidationData);
             ReconnectUtils.endReconnectHandshake(connection);
             return reservedSignedState;
         } catch (final IOException | SignedStateInvalidException e) {
+            if (reservedSignedState != null) {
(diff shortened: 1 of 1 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/reconnect/ReconnectLearner.java
```diff
@@ -152,15 +152,21 @@ private void resetSocketTimeout() throws ReconnectException {
...
             ReconnectUtils.endReconnectHandshake(connection);
             return reservedSignedState;
         } catch (final IOException | SignedStateInvalidException e) {
+            if (reservedSignedState != null) {
+                // if the state was received, we need to release it or it will be leaked
+                reservedSignedState.close();
+            }
             throw new ReconnectException(e);
         } catch (final InterruptedException e) {
+            // an interrupt can only occur in the reconnect() method, so we don't need to close the reservedSignedState
```
anthony-swirldslabs, line 169: Minor: you never know what the code above is going to be doing in the future. Since the closing is guarded by a `!= null` check anyway, I really don't see harm in closing the state here as well.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/reconnect/ReconnectLearner.java
cody-littley: Good point, but since this is a cherry pick, probably better to make this change in develop rather than in this particular PR.

cody-littley (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/reconnect/ReconnectLearner.java
anthony-swirldslabs: Agree. Please feel free to resolve the conversation then (preferably, after filing a ticket to implement the suggested improvement.)

timo0 (approved):

Nana-EC (changes requested):
Blocking inclusion into 0.51

lpetrovic05:
stale, closing

PR #13687, opened 2024-06-05
feat: HIP-904 Support unlimited max auto associations and adjust fees

stoyanov-st:
**Description**:
For the needs of HIP-904 we want to extend the current logic regarding `maxAutoAssociations` to support -1 as unlimited value.
We've added a feature flag for easy configuration.
We've extended current handlers for create, update, lazy/auto-create to support the unlimited value and adjusted the fees correspondingly.
We also adjusted the fees for transfers resulting in auto associations.

**Related issue(s)**:

Fixes #12868 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #13907, opened 2024-06-18
feat: HIP-904 Add crypto transfer fee calculations

### LINKED ISSUE #13375: HIP-904: Adapt CryptoTransferHandler's fee calculation to ensure that the sender is covering the fees for auto-associations

kselveliev:
**Description**:
This pr is a clone of https://github.com/hashgraph/hedera-services/pull/13853.
This pr aims to be opened against develop without needing all the logic from https://github.com/hashgraph/hedera-services/pull/13687 only needing the feature flag.

As we only want this pr to contain the crypto transfer fee calculations and be independent from the other changes in https://github.com/hashgraph/hedera-services/pull/13687. This makes it easier for review. This pr will have duplicated logic from 13687 only the logic that adds the maxAutoAssocation feature flag.

This PR adds the needed changes in the calculations needed for HIP-904. Now the sender of a crypto transfer is charged for the auto association slots if the unlimitedAutoAssociations flag is enabled.

This PR modifies 
`FeeContext` - adds needed classes for TokenAssociateToAccountHandler.calculateFees()
`AssociateTokenRecipientsStep` add fee logic based on the maxAutoAssociations flag.

**Related issue(s)**:

Fixes #13375

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/fees/FeeCalculator.java
+++ b/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/fees/FeeCalculator.java
@@ -102,4 +102,10 @@ public interface FeeCalculator {
      */
     @NonNull
     FeeCalculator resetUsage();
+
+    /**
+     * Gets the price for signature verifications only.
+     * @return the vpt price
+     */
+    long getVptPrice();
 }
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
@@ -205,9 +205,6 @@ public class TokenAssociateToAccountHandler extends BaseTokenHandler implements 
         final var body = feeContext.body();
         final var op = body.tokenAssociateOrThrow();
         final var calculator = feeContext.feeCalculator(DEFAULT);
-        final var accountId = op.accountOrThrow();
-        final var readableAccountStore = feeContext.readableStore(ReadableAccountStore.class);
-        final var account = readableAccountStore.getAccountById(accountId);
         final var unlimitedAutoAssociations =
                 feeContext.configuration().getConfigData(EntitiesConfig.class).unlimitedAutoAssociationsEnabled();
 
@@ -215,7 +212,7 @@ public class TokenAssociateToAccountHandler extends BaseTokenHandler implements 
             final var exchangeRateInfo = feeContext.exchangeRateInfo();
 
             calculator.resetUsage();
-            final var price = getTinybarsFromTinyCents(
+            final var singleSigPrice = getTinybarsFromTinyCents(
                     calculator.getVptPrice(), exchangeRateInfo.activeRate(feeContext.consensusNow()));
             final var newAssociationsCount = op.tokens().size();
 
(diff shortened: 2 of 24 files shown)


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/fees/FeeCalculator.java
```diff
@@ -102,4 +102,10 @@ public interface FeeCalculator {
      */
     @NonNull
     FeeCalculator resetUsage();
+
+    /**
+     * Gets the price for signature verifications only.
+     * @return the vpt price
+     */
+    long getVptPrice();
```
netopyr, line 110: This looks wrong. `FeeCalculator` should hide these nasty details.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
```diff
@@ -192,14 +200,65 @@ private record Validated(@NonNull Account account, @NonNull List<Token> tokens)
...
+
+            var totalSigPrice = 0L;
+            // If the transaction category is USER then this is the main transaction. If it is CHILD
+            // then this is a transaction that was internally dispatched and the signature verifications
+            // were already performed and charged in the main transaction so we don't need to
+            // calculate them again here.
+            if (feeContext.transactionCategory() == TransactionCategory.USER) {
+                totalSigPrice = Math.max(0, feeContext.numTxnSignatures() - 1) * singleSigPrice;
+            }
+            return new Fees(0L, totalSigPrice + newAssociationsCount * associateInTinybars, 0L);
```
netopyr, line 231: This turns the fee calculation upside-down. You should take a `FeeCalculator`, provide the used resources (e.g. number of signatures), and call `calculate()` to get the value.

netopyr (changes requested):
The details of fee calculation should be hidden from the handlers. Have you seen how the fee calculation is done for example in the handlers of the consensus service?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
bilyana-gospodinova: Since the token associate price is now fixed, in the handler we needed to have the canonical price for an association and the price for any additional signatures apart from the first one that is free. Any other calculations in the calculator and the `calculate()` method needed to be reset as they increased additionally the result price. We have discussed this with @tinker-michaelj and @Neeharika-Sompalli and we used the `TokenMintHandler` as reference. This is why needed to expose the `getVptPrice()` method in `FeeCalculator`.

tinker-michaelj: Yes @netopyr discussions converged on wanting fees to be a very exact `$0.05`-per-association; but with an increased fee at current prices for each additional signature.

Hacking resource prices in _feeSchedules.json_ to get this behavior would not be a **huge** amount of work, but would:
  1. Make it less obvious what is going on; and,
  2. Require a council-signed update to file `0.0.111`.

We can however, do this, and up somewhere that is more like the standard pattern.

netopyr: Yes, please do this.

Fee calculation is the core responsibility of the `FeeCalculator`. We have to strictly enforce this to avoid a mess. If the current functionality is insufficient, we should discuss whether the `FeeCalculator` can be extended. This will result in a cleaner design than implementing exceptional cases directly in the handlers.

netopyr: In general, keeping the design clean is always a much higher priority than implementing a temporary feature flag.

Why do you think the `feeSchedules.json` approach is problematic? We can upload it anytime, as we would toggle a feature flag.

netopyr: We can have different `feeSchedules.json` files depending on the environment.

bilyana-gospodinova: @netopyr, we opened a separate PR with the refactored version of the code - https://github.com/hashgraph/hedera-services/pull/13942. Can you please review it and let us know if the new approach is better?

PR #13911, opened 2024-06-18
Refactoring of proposal template

hendrikebbers:


platform-sdk/docs/proposals/template.md
```diff
@@ -12,6 +12,7 @@ Give a 1-3 sentence summary of the capability in the design proposal.
 | Functional Impacts | Services, DevOps, Mirror Node, etc. |
 | Related Proposals  | Proposal-1, Proposal-2              |
 | HIPS               | HIP-1, HIP-2,                       |
+| State              | Withdrawn by template-2.md          |
```
poulok, line 15: Two things:
1. I am confused by this wording. A proposal is "withdrawn by" anything
2. I disagree with adding this state. The state is tracked in the Design Proposal Project and it is not good to track it in multiple places.

platform-sdk/docs/proposals/template-2.md
```diff
@@ -0,0 +1,184 @@
+# Proposal Title Goes Here
+
+| Metadata           | Entities                            | 
+|--------------------|-------------------------------------|
+| Designers          | John, Mary, alpha-team              |
+| Functional Impacts | Services, DevOps, Mirror Node, etc. |
+| Related Proposals  | Proposal-1, Proposal-2              |
+| HIPS               | HIP-1, HIP-2,                       |
+| State              | DRAFT                               |
```
poulok, line 9: Same and comment below - this is tracked in the board and should not be duplicated here.

platform-sdk/docs/proposals/template-2.md
```diff
@@ -0,0 +1,184 @@
...
+
+This section should give an answer to the question: "Why is this proposal necessary?"
+
+All subtitled sections are optional and can be removed if not applicable.
+A minimal proposal must have a purpose and context section without any sub-sections.
+In any case the section must give a detailed answer to the question: "Why is this proposal necessary?"
+
+### Motivation
+
+Describe the motivation for the proposal. What problem does it solve? What use cases does it support?
```
poulok, line 27: This seems redundant with the `Purpose and Context` section. The question of "Why is this proposal necessary?" is the the same as "What problem does this solve?"

platform-sdk/docs/proposals/template-2.md
```diff
@@ -0,0 +1,184 @@
...
+Remove this section if not applicable.
+
+### Migration
+
+Describe the migration path for existing usage.
+What changes are needed to upgrade from the current state to the new state?
+
+Remove this section if not applicable.
+
+## Testing
```
poulok, line 127: I actually prefer "Test Plan", because it indicates a purposeful set of actions that have been set out ahead of time. But the exact wording is not super important.

platform-sdk/docs/proposals/template-2.md
```diff
@@ -0,0 +1,184 @@
+# Proposal Title Goes Here
```
poulok, line 1: Is the idea to have two different templates? I'm confused by a `template-2.md` rather than changing the existing one.

hendrikebbers: I did that because it was not clear to me what to do with the result of an "old" proposal. Well, the template is not really a proposal, so here it makes sense to just change it. Will do that.

platform-sdk/docs/proposals/template-2.md
hendrikebbers: Fine to change it. For me "Test Plan" is way more concrete than what I would expect in that section.
Like "This is the exact plan/schedule of tests that must be executed"

platform-sdk/docs/proposals/template-2.md
hendrikebbers: ok, will remove

platform-sdk/docs/proposals/template.md
hendrikebbers: ok, will remove

platform-sdk/docs/proposals/template-2.md
hendrikebbers: Agree as long as you have text under the `Purpose and Context` header instead of just having it as a wrapper around all the h3 headers. My idea was more like that:

```
## Purpose and Context

### Motivation
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren.

### Goals
Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.

### Non-Goals
At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua.
```

platform-sdk/docs/proposals/template-2.md
edward-swirldslabs: I agree with editing the template in the original file.  It makes it clear as to what has changed through github's automated diff.  Right now I would need to read both and pay close attention to determine if anything has changed.

platform-sdk/docs/proposals/template-2.md
edward-swirldslabs: I like the goals and non-goals as explicit substructure to purpose and context.   I feel that the motivation section is redundant with the content I would expect to see at the top-level of `Purpose and Context`.

hendrikebbers:
-0.5 test

edward-swirldslabs:
Is this PR still being worked on?

PR #13923, opened 2024-06-18
chore: HandleWorkflow Design Document

### LINKED ISSUE #13713: Add HandleWorkflow refactor design document
Add design document for the Handle Workflow refactor happening in https://github.com/hashgraph/hedera-services/issues/13638

Neeharika-Sompalli:
Fixes #13713 

- Adds `HandleWorkflow` design document merged in the [PR](https://github.com/hashgraph/hedera-services/pull/13639)
- Removed `SingleTransactionRecordBuilderImpl.toString()`  methods addressing review comment on merged HandleWorkflow refactor PR
- Changed the order of checks in `DispatchProcessor` for clarity. This is not changing any logic


thomas-swirlds-labs (commented):
The png resources could be moved under the app/image or design/images directories

Neeharika-Sompalli:
> The png resources could be moved under the app/image or design/images directories

Thanks @thomas-swirlds-labs ! Fixed

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,104 @@ The query workflow consists of the following steps:
...
+   - If there is a `ThorttleException` thrown, stack is rolled back and payer is charged without serviceFee. 
+   - If there is any unhandled exception, stack is rolled back and payer is charged. The record is set to 
+    `FAIL_INVALID` status. 
+6. **Track Utilization:** Tracks network utilization for the transaction handled
+7. **Record Finalization:** Finalizes the record from state changes
+8. **Commit:** Commits all the changes to the state
+9. **Metrics Update:** Updates any metrics that need to be updated
+10. **Record Stream:** Adds all the records to the record cache
+
+![dispatch_processor.png](images/dispatch_processor.pngsor.png)
```
jsync-swirlds, line 189: typo?
```suggestion
![dispatch_processor.png](images/dispatch_processor.png)
```

Neeharika-Sompalli: Yes @jsync-swirlds . Thank you! Fixed

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,114 @@ The query workflow consists of the following steps:
...
+3. Externalizes the record stream items
+4. Update metrics for the handled user transaction
+
+
+#### UserTxnWorkflow overview:
+
+![User Transaction Workflow](images/user_txn_workflow.png)
+
+1. **SkipHandleWorkflow**: If the transaction is from older software, the transaction will be skipped handling by calling `SkipHandleWorkflow`. 
+This writes a record with `BUSY` status and adds to record cache
```
jsync-swirlds, line 144: Is `BUSY` correct for obsolete transactions?
That would be very confusing to me if I were the client developer.  Something like `TRANSACTION_EXPIRED` would make a lot more sense to me.

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,114 @@ The query workflow consists of the following steps:
...
+- **UserTxnScope** - Objects that are created once for platform transaction. 
+Examples include the `Configuration`, `RecordListBuilder` and `TokenContext`.
+Dagger provides all the objects that can be constructed in this scope [here](https://github.com/hashgraph/hedera-services/tree/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/modules) 
+and [UserTxnComponent](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTransactionComponent.java)
+takes all the inputs that are needed to execute the user transaction.
+- **UserDispatchScope** - Objects that are created once for each user transaction that will be dispatched. 
+Examples include the `SingleTransactionRecordBuilder` for user transaction and `FeeContext`.
+Dagger provides all the objects that can be constructed in this scope in [UserDispatchModule](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/user/modules/UserDispatchModule.java) and `UserDispatchComponent`.
```
iwsimon, line 115: Since these are new components introduced from HandleWorkflow refactoring, it would be good to add either more javedoc on each methods of the classes or add some more descriptions here.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
jsync-swirlds, line 1: Is this change needed in a design document update?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
jsync-swirlds, line 1: Is this change needed in a design document update?

jsync-swirlds (commented):
A couple questions and some mistyped or leftover text.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
Neeharika-Sompalli: This was something Richard added as review comment on previous PR. I can update PR description

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
Neeharika-Sompalli: This is not changing anything. Just adding more clarity. I can update PR description.

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,104 @@ The query workflow consists of the following steps:
...
+   - Dispatches the transaction to the appropriate handler with appropriate handleContext constructed
+     based on the transaction scope. The handler will execute the business logic of the transaction.
+   - Updates the record to SUCCESS status
+   - If it is a user transaction, notifies if system file is updated and platform state is modified to
+     the appropriate facilities
+5. **Exception Handling:** When any exception is thrown in the steps under 4, the following steps are executed:
+   - If any HandleException is thrown, rolls back the complete stack and 
+     charges the payer. The payer is charged again because when stack is rolled back the previous
+     charges to the payer are rolled back as well. 
+   - If there is a `ThorttleException` thrown, stack is rolled back and payer is charged without serviceFee.
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Thanks @jsync-swirlds

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,114 @@ The query workflow consists of the following steps:
...
+
+#### HandleWorkflow overview:
+
+![Basic Overview](images/handle_basic_overview.png)
+
+The `HandleWorkflow` class is responsible for handling the platform transaction and providing the record stream.
+The overall high level steps are as follows:
+
+1. Calls `BlockRecordManager` to update the new consensus time for the user transaction, puts the lastBlockInfo in state if needed.
+when blocks
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Removed. Thanks @jsync-swirlds

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +86,114 @@ The query workflow consists of the following steps:
...
+2. `UserTxnWorkflow` is called to handle the transaction and provide record stream
+3. Externalizes the record stream items
+4. Update metrics for the handled user transaction
+
+
+#### UserTxnWorkflow overview:
+
+![User Transaction Workflow](images/user_txn_workflow.png)
+
+1. **SkipHandleWorkflow**: If the transaction is from older software, the transaction will be skipped handling by calling `SkipHandleWorkflow`.
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed it. Thanks @jsync-swirlds

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: The transaction is not handled. Its not same as `TRANSACTION_EXPIRED`. We can add a new response code. But mono-service was adding `BUSY` status, which is almost same as it is being throttled and not handled.

hedera-node/docs/design/app/workflows.md
```diff
@@ -1,11 +1,11 @@
-### Workflow Onset
+## Workflow Onset
 
 At the beginning of each workflow, a `Transaction` has to be parsed and validated.
 As these steps are the same for all workflows, this functionality has been extracted and made available independently.
 All related classes can be found in the package `com.hedera.node.app.workflows.onset`.
 Details about the required pre-checks can be found [here](transaction-prechecks.md).
```
david-bakin-sl, line 6: Neither this document nor `transaction-prechecks.md` uses the word "pure" which is part of the name of the method `pureChecks` implemented by each service.  It would be helpful to relate "pure" to "pre-checks" in both documents.

(Didn't we recently have a confusion where `pureChecks` is  a) run during ingest and b) at least one service (I forget which) did a check which in fact wasn't supposed to happen during ingest (where the tx is rejected without being paid for by the erroring user?  Having "pure" documented here might help prevent that in future.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
jsync-swirlds: Fair, just a little unexpected to see code changes in document PRs.

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: Which classes do you mean @iwsimon? 
You would like to have javadoc on all Dagger provides and binds methods in `UserDispatchModule` ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
david-bakin-sl: yes, and it is a (minor) semantic change that makes it not a "chore", IMO.

hedera-node/docs/design/app/workflows.md
iwsimon: I saw only javadoc on the top of all those Dagger classes. If we could add javadoc on each method, it will be helpful.

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: If the transaction is submitted by using older software version. This happened as an edge case during one of the upgrades and we added this check. Sure, will discuss what response will be better with team.

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: Okay, will do. Most of them should be self explanatory. Thanks!

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: Thanks @david-bakin-sl . `pureChecks` is supposed to be called only in pre-handle and handle. They are different from `transaction-prechecks` that happen in Ingest.  `pureChecks` are checks that are independent of State or Configuration, so they remain the same when called in pre-handle or handle. Call to `PureChecks` in Ingest workflow will be removed in the future, since it was added only as safety measure before calling `calculateFees`. I will update Ingest Workflow part with that.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
```diff
@@ -220,17 +220,17 @@ private void chargeCreator(@NonNull final Dispatch dispatch, @NonNull final Erro
...
                 dispatch.payerId(),
                 dispatch.txnInfo().functionality(),
                 dispatch.txnInfo().txBody());
         if (hasWaivedFees) {
             return;
         }
+        final var fees = dispatch.fees();
+        if (fees.nothingToCharge()) {
+            return;
+        }
```
rbair23, line 233: This change looks like something that should be in a separate PR and not part of updating a design document

Neeharika-Sompalli: Reverted. Thanks @rbair23

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java
Neeharika-Sompalli: Reverted @jsync-swirlds

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionRecordBuilderImpl.java
Neeharika-Sompalli: Reverted as well. Thanks @jsync-swirlds

hedera-node/docs/design/app/workflows.md
david-bakin-sl: TY, but also please mention "pure" somewhere so readers know that "pure" = "pre" when they actually look at the code.  I suggest?

Neeharika-Sompalli: Yes, added the `pureChecks` step, in Ingest Workflow and Pre-handle. Could you please check if that helps?

jsync-swirlds (approved):

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +95,113 @@ The query workflow consists of the following steps:
...
+
+All the objects used while handling the transaction belong to one of the following Dagger scopes. 
+- **Singleton** - Objects that are created once and used for the entire lifecycle of the application.
+Examples include the `NodeInfo` and `WorkingStateAccessor`.
+- **UserTxnScope** - Objects that are created once for platform transaction. 
+Examples include the `Configuration`, `RecordListBuilder` and `TokenContext`.
+Dagger provides all the objects that can be constructed in this scope [here](https://github.com/hashgraph/hedera-services/tree/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/modules) 
+and [UserTxnComponent](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/txn/UserTransactionComponent.java)
+takes all the inputs that are needed to execute the user transaction.
+- **UserDispatchScope** - Objects that are created once for each user transaction that will be dispatched.
```
rbair23, line 122: I don't understand the difference between these two scopes. I looked tat the linked `UserDispatchComponent`, but it was empty.

hedera-node/docs/design/app/workflows.md
```diff
@@ -86,4 +95,113 @@ The query workflow consists of the following steps:
...
+2. `UserTxnWorkflow` is called to handle the transaction and provide record stream
+3. Externalizes the record stream items
+4. Update metrics for the handled user transaction
+
+
+#### UserTxnWorkflow overview:
+
+![User Transaction Workflow](images/user_txn_workflow.png)
+
+1. **SkipHandleWorkflow**: If the transaction is from older software, the transaction not be handled.
```
rbair23, line 151: I don't see the value of `SkipHandleWorkflow`. First, it uses the name "workflow" which, up until this point, has meant one of the larger flows -- handle, pre-handle, ingest, query. It really isn't much of a workflow, as it has a single method.

Second, why have the additional class? It is called from exactly one place, a private mtethod in `UserTxnWorkflow`. To me, this is just an implementation detail of the (misnamed) `sideEffectsOfExecution` method. I could see it as a separate private method in `UserTxnWorkflow`, but why is it a completely separate class?

In the end, there is a matter of personal preference. Maybe calling it a workflow is too much, and it is a "step" in the workflow. If I were writing the code, I'd eliminate the class and fold it into `UserTxnWorkflow`. For me, when understanding a system:
- When I approach a code base, every class must be visited to understand what it is for.
- If I encounter a class that is public, I assume it must be used from something outside its current package. I use "find usages" in IntelliJ to discover who is using it. If it is being used in exactly one place, I wonder why it is public, or whether it is in the wrong package. Sometimes it is, sometimes it isn't. But I have to spend time and mental energy to determine whether this is true. If I get to the end of that work and discover it could have been package private, then I want to make it such, so that it has only that visibility that it truly requires. This makes it easier for people in the future to understand it.
- Every class, and every public method, adds to the conceptual "weight" of the codebase. Sometimes this is worth it, because the conceptual clarity compensates for the weight. Sometimes it isn't, where things actually got less clear (because it isn't clear why a class exists on its own)

Some people like to break a class down into many tiny pieces in an attempt to tame complexity. Rather than one big file with thousands of lines and hundreds of functions, they prefer hundreds of classes with 10's of lines each. Each individual class is extremely easy to understand, so this seems like an improvement. But the overall complexity did not reduce -- the code is still doing all the same things it did before, but now you have to read hundreds of classes to understand the relationship between them. And you have hundreds of classes for the JVM to class-load, and hundreds of instances to create and collaborate to create a final result.

In my judgement, I liked the original code where we had a small handful of classes and a few methods and each "step" in our diagram could be a class, function, or commented block of code in a single method, depending on how "meaty" and complex, or reusable, each step was.

When I look at the handle package now, I see:
- 17 packages
- 66 Classes (.java files, not counting inner classes)

I think we have gone wrong.

rbair23: Maybe I am reacting to excessive Dagger "magic"? Or maybe when I see things like `FlowUtils` it sets of alarm bells (any time "utils" show up, I want to dig into why we felt like it was necessary. In this case, `FlowUtils` looks very suspicious because it has a public hashset `CONTRACT_OPERATIONS` and a public static method `isContractOperation`, and I'm not sure why we have both. And it also shows that we have "Possibly CATASTROPHIC failure" thrown from two places, which seems odd. I wonder if this file only exists because when we refactored we had two similar classes that both needed the same thing but didn't have a convenient parent class so we created the utils class as an admission that the structure wasn't quite right and we needed *somewhere* to put it?).

Also the fact that we have `logic` packages. I don't know what that means, exactly. What goes into a `logic` package vs. the normal package? For example, why do we have some classes in `com.hedera.node.app.workflows.handle.flow.txn.logic` and some in `com.hedera.node.app.workflows.handle.flow.txn`? `logic` is such a vague term. When I see vagueness in naming, it usually means the concepts are not clearly understood, and it makes it very hard to understand a codebase, because the only way to learn it is to open and read every source file, since the names don't have a clear meaning.

Which brings me back to `UserTxnWorkflow`. These methods are defined (and I removed the *excellent* documentation -- well done!! -- to point out a couple things that could be improved):

```
    public Stream<SingleTransactionRecord> execute() {
        try {
            return sideEffectsOfExecution();
        } catch (final Exception e) {
            logger.error("{} - exception thrown while handling user transaction", ALERT_MESSAGE, e);
            return failInvalidRecordStream();
        }
    }

    private Stream<SingleTransactionRecord> sideEffectsOfExecution() {
        if (isOlderSoftwareEvent()) {
            skipHandleWorkflow.execute(userTxn);
        } else {
            if (userTxn.isGenesisTxn()) {
                genesisWorkflow.executeIn(userTxn.tokenContext());
            }
            defaultHandleWorkflow.execute(userTxn);
            updateWorkflowMetrics();
        }
        return allSideEffectsFrom(userTxn.recordListBuilder());
    }
```

The `execute` method is well named, and the documentation for it was wonderful. It is very short, and delegates to a private method `sideEffectsOfExecution`. The problem:
- `sideEffectsOfExecution` is a poor name. It is actually `execute`! It has the exact same method signature!
- It is only called from `execute`, so why is it a separate method?

It seems clear that the body of sideEffectsOfExecution should be moved into `execute` and `sideEffectsOfExecution` eliminated. There appears to be no reason for the misnamed method to even exist.

Neeharika-Sompalli: > I don't see the value of SkipHandleWorkflow

@rbair23 Agreed this is more like a step. If we were to rename this as `SkipHandleStep` and 
I can make it as a private method inside `UserTxnWorkflow`, since will still be easy to unit test.

The original `HandleWorkflow` had nested try catches and was difficult to unit test. Also there was code duplicated between user and child transactions. So, this refactor tried to split it into multiple classes that limit a class to one responsibility and re-use the same code for user and child transactions.

I agree, we could probably delete some classes, as long as it is easier to unit test.
Some classes like `ErrorReporter,  are only used by `DispatchProcessor` and could be moved inside the class. But, that becomes harder to unit test. 

We can have a call next week when you are available and discuss which classes can be merged, if that works.

rbair23: Another example I would cite would be `ErrorReporter`. The function of the `ErrorReporter` is great, and the actual flow of the code seems very reasonable. And actually it works exactly as I would have expected -- except the name. I came across it looking at `DispatchProcessor#processDispatch` which on the first line (after the null check) it calls `errorReporter.errorReportFor(dispatch)`. Now, I assumed we must not be erroring out every request, so it must be that this is not actually generating an error report, but that it is doing some kind of validation result and the the result could be an error, or it could be valid. But the class name and method name suggest this is creating an error report -- only context allowed me to consider that there was such a thing as a healthy error report. The terms are contradictory, which makes it strange.

We probably would be better off with something like Validator and ValidationResult, or something that means "a piece of code that checks whether this dispatch is good" and returns something that means "the result of the check", where the result of the check could be "good" or "bad, with information".

I think it is very important to get names just right, to give an intuitive understanding of what the system is doing and how it is going to do it.

rbair23: Another example that I don't quite understand, and is related to the `FlowUtils`. The `DispatchUsageManager` has a `screenForCapacity` method that only performs work if `isContractOperation`, and it checks to see if the gas has been exhausted.

Fundamentally, this looks to be totally in the wrong place. The basic design of modularity was that module-specific logic lived in the modules, and generic logic lived in the base. If we put module-specific logic into the base, then we've broken modularity.

Fees, throttles, etc, should really be part of each particular module. We didn't do that yet because we need to rewrite fees and throttles, but we haven't got there yet. I'm not sure if it is necessary for us to break this design for contracts?

In addition, the name `screenForCapacity` does imply it is somehow filtering based on capacity, but it has a nice generic name. Whereas the actual implementation only screens for capacity of contracts, and even more interesting it only calls the `networkUtilizationManager` to track the transaction if it is a contract operation. That seems wrong? Shouldn't congestion pricing / network utilization affect many other kinds of things? Every thing? The method name doesn't imply it is only for contracts, and the comment "Tracks usage of the given dispatch before it is sent to a handler" appears to disagree with the implementation. And `trackUsage` is where tracking of usage is supposed to go, so why is it here?

I assume the answer is because it just so happens that our implementation calls `screenForCapacity` before it calls `trackUsage`, so for smart contracts (which are the only case we have today where we use a back-end throttle), we found that without this line of code we weren't tracking usage properly. But that seems like a bug in the calling logic -- it should call trackUsage before it calls `screenForCapacity`? The name of `screenForCapacity` doesn't carry any meaning to suggest that it also is tracking capacity.

rbair23: >@rbair23 Agreed this is more like a step. If we were to rename this as SkipHandleStep and
I can make it as a private method inside UserTxnWorkflow, since will still be easy to unit test.

Ah, and now we get I think to the point :-). I do not believe in making code more visible than necessary for the sake of testing, except in rare circumstances (usually related to threading, or static utility functions).

This is a testing philosophy -- black-box vs. white-box. Many people would say that unit testing is inherently white-box testing, and therefore having access to every method within the class is needed to test it properly. Early in my career, I fully subscribed to this. Clearly a lot of people agree, which is why `@VisibleForTesting` exists. But I have come to very much dislike this form of testing.

The problem for me is three-fold:
1. If you make a method (or class) visible for testing, you make it harder to understand the actual codebase, at the same time you make it easier to white-box test. You might also impact performance (number of classes, methods, interfaces, etc)
2. White-box testing is inherently brittle. Small changes in the implementation require large changes in the test code because the test code is not taking advantage of data hiding! Not only does this make it hard to update code (since many tests break and need to be fixed), and make code reviews much harder (because many additional files are touched and many tests), but it also wipes out a large part of the value of tests, which is to catch regressions when the implementation changes! Instead, the tests are only as good as the most recent implementation change made them, since they have to be rewritten for each change to the internal implementation details.
3. You often find code that the IDE says is in use, only to find that the *only* code using it is test code. By keeping the code with minimal access, and testing the public APIs, you are able to find and eliminate dead code.

Of these reasons, #2 is the most compelling to me. My approach to unit testing is to test the API of a class, but not to open-up access to the internal details. It sometimes requires a more thoughtful approach to API and testing, but it results in something that is more stable across implementation changes, and provides a safety net between implementation changes.

> Also there was code duplicated between user and child transactions [...] and re-use the same code for user and child transactions

I think this is great and definitely a good idea! Big improvement!

> So, this refactor tried to split it into multiple classes that limit a class to one responsibility

The definition of the "one responsibility" can be broad ("responsible to pre-handle a transaction") or narrow ("throw an exception if the value is out of bounds"). If it is too broad, you get big classes with lots going on, if too narrow you get lots of tiny classes and cannot see the forrest from the trees. I agree with the idea that every class should have a clearly defined scope and responsibility, but we have to be careful not to let it be too narrow.

david-bakin-sl: If you have private methods that are really large enough to need a unit test then there are valid software engineering ways to make them _visible_ for testing - typically by putting them - with other related code! - in a separate class with a well-defined purpose.
* Which has an additional advantage in that putting them in a separate class and exposing them properly puts more urgency on you to make sure you've
    1. given it a proper _name_ that says what it is really doing and
    2. that you properly refactor _it_ into smaller units (that is, making new private methods) that you can understand.  
    
    And the advantage of that is it helps you understand what this newly extracted class was doing when embedded in the original class and _that_ understanding can help you clarify things there.

Would you agree to some/most of that, BTW, Richard?

hedera-node/docs/design/app/workflows.md
```diff
@@ -18,22 +18,28 @@ The ingest-workflow is single-threaded, but multiple calls can run in parallel.
...
-2. **Deduplicate.** The transaction is checked to ensure it has not been processed before.
-3. **Check throttles.** Throttling must be observed and checked as early as possible.
-4. **Get payer account.** The account data of the payer is read from the latest immutable state.
-5. Account Balance
+1. **Check timeBox** Checks whether the transaction duration is valid as per the configuration for valid durations 
+for the network, and whether the current node wall-clock time falls between the transaction start and the transaction end (transaction start + duration)
+2. **Parse transaction.** The transaction arrives as a byte-array. The required parts are parsed and the structure and syntax are validated.
+3. **Deduplicate.** The transaction is checked to ensure it has not been processed before.
+4. **Check throttles** Throttling must be observed and checked as early as possible.
+    1. **Pure Checks** Calls `pureChecks` method that does validations based on the respective handler's transaction body
```
david-bakin-sl, line 26: I don't think this is the "nesting" you intended.

david-bakin-sl (commented):
TY, `pureChecks` is better now, I'm happy with that.

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: > I agree with the idea that every class should have a clearly defined scope and responsibility, but we have to be careful not to let it be too narrow.

Makes sense. 

Thanks for the great feedback @rbair23 . I will open a PR with the changes as you suggested and also look for classes that can be moved into parent class.

Neeharika-Sompalli: > I agree with the idea that every class should have a clearly defined scope and responsibility, but we have to be careful not to let it be too narrow.

Makes sense. 

> If we put module-specific logic into the base, then we've broken modularity.

I agree this is in the wrong place. But, we didn't try to move logic to its own modules. Just teased out the one big `HandleWorkflow` class. We need to discuss how/if this should be moved into contract-service. 

> Whereas the actual implementation only screens for capacity of contracts, and even more interesting it only calls the networkUtilizationManager to track the transaction if it is a contract operation. That seems wrong? Shouldn't congestion pricing / network utilization affect many other kinds of things? Every thing?

Yes, it is weird. We usually track utilization after the transaction is handled for every transaction [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java#L114) for every transaction. It is only for contract operations, we check before handling if we exhaust gas once we handle the transaction.  I didn't change the way it was implemented, since it is a bigger change in throttles. It should be changed in the future

Thanks for the great feedback @rbair23 . I will open a PR with the changes as you suggested.

Neeharika-Sompalli: > I agree with the idea that every class should have a clearly defined scope and responsibility, but we have to be careful not to let it be too narrow.

Makes sense. 

> If we put module-specific logic into the base, then we've broken modularity.

I agree this is in the wrong place. But, we didn't try to move logic to its own modules. Just teased out the one big `HandleWorkflow` class. We need to discuss how/if this should be moved into contract-service. 

> Whereas the actual implementation only screens for capacity of contracts, and even more interesting it only calls the networkUtilizationManager to track the transaction if it is a contract operation. That seems wrong? Shouldn't congestion pricing / network utilization affect many other kinds of things? Every thing?

Yes, it is weird. We usually track utilization after the transaction is handled for every transaction [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/flow/dispatch/logic/DispatchProcessor.java#L114). It is only for contract operations, we check before handling if we exhaust gas once we handle the transaction and fail before-hand. I didn't change the way it was implemented originally, since I wasn't sure. May be because it is needed for ContractCalls. @tinker-michaelj  do you know? But it seems natural that fees and throttles should be maintained for each dispatch separately, instead of handling it in one place in HandleWorkflow.  It could be changed in the future

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: - `UserTxnScope` includes extra objects that deal with side effects of advancing consensus time (for example, staking updates, genesis records externalization) and verifying top level signatures to finalize hollow accounts
- `UserDispatchScope` includes only objects needed to validate and dispatch the user transaction to its handler. `UserDispatchComponent` is an implementation of `Dispatch` fully provided by Dagger.
- `ChildDispatchScope` includes only objects needed to validate and dispatch the child transaction to its handler.

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: I went through all the classes with @tinker-michaelj . 

Out of 63 .java files in the `workflows.handle` package, we have:

**Workflow API and implementation (1)**
 - HandleWorkflow

**Used in workflow implementation and extracted for testability (10)**
 - ScheduleExpirationHook, GenesisWorkflow, StakingPeriodTimeHook, HollowAccountCompleter, SchedulePurger, UserTxnWorkflow, SkipHandleWorkflow, DefaultHandleWorkflow, FlowUtils, PreHandleResultManager

**Misc support classes (3)**
 - WarmupContextImpl, CacheWarmer, HandleWorkflowMetrics

**Dispatch API and implementation (3)**
 - Dispatch, DispatchProcessor, DispatchHandleContext

**Used in Dispatch implementation and extracted for testability (4)**
 - ErrorReport, ErrorReporter, ChildTxnInfoFactory, ChildDispatchFactory

**HandleContext implementation (7)**
 - WritableQueueStateStack, WritableSingletonStateStack, WritableKVStateStack, SavepointStackImpl, WritableStatesStack, AttributeValidatorImpl, ExpiryValidatorImpl

**Belong in record-facility module (8)**
 - TriggeredFinalizeContext, RecordListBuilder, SingleTransactionRecordBuilderImpl, TokenContextImpl, ChildFinalizeContextImpl, RecordFinalizer, UserRecordInitializer, ChildRecordBuilderFactory

**Belong in other misc facility (config-, throttle-, fee-) module (4)**
 - SystemFileUpdateFacility, PlatformStateUpdateFacility, DispatchUsageManager, ThrottleException

**Belong in other package entirely (1)**
 - MigrationContextImpl

**Scope "container classes" for Dagger support (3)**
 - UserTransactionComponent, UserDispatchComponent, ChildDispatchComponent

**DI only (14)**
 - HandlersInjectionModule, HandleWorkflowInjectionModule, DispatchSubcomponentsModule, PreHandleResultModule, ContextModule, StateModule, LastHandledTime, ActiveConfigModule, UserTxnScope, UserDispatchScope, UserDispatchModule, ChildDispatchScope, ChildDispatchModule, ChildQualifier

**Enums added for clarity (5)**
 - WorkDone, WorkflowCheck, OfferedFeeCheck, DuplicateStatus, ServiceFeeStatus

After "collapsing" the classes extracted for testability and moving the facility classes into their proper homes, we can have just 14 classes in `workflows.handle` that are not purely for DI or syntactic sugar (e.g. enums to improve readability).  

@rbair23 Do you suggest we do this PR now or later when we become clear on facilities?

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: Thanks @david-bakin-sl . Fixed it.

hedera-node/docs/design/app/workflows.md
rbair23: >If you have private methods that are really large enough to need a unit test then there are valid software engineering ways to make them visible for testing - typically by putting them - with other related code! - in a separate class with a well-defined purpose.
>
>Which has an additional advantage in that putting them in a separate class and exposing them properly puts more urgency on you to make sure you've given it a proper name that says what it is really doing and
that you properly refactor it into smaller units (that is, making new private methods) that you can understand.
And the advantage of that is it helps you understand what this newly extracted class was doing when embedded in the original class and that understanding can help you clarify things there.
>
>Would you agree to some/most of that, BTW, Richard?

A class may be complex because it is trying to do too much that it shouldn't be trying to do. For example, I have a class that handles customer account creation and purchase order creation. That class is clearly doing too much, and we should split it apart so that each class has a very well defined task.

On the other hand, I might have a class that handles customer account creation, and as part of that process there are 10 things it needs to do. I may be able to just do those 10 things in a single method. Or maybe in 10 methods. Or I could break them out into 10 different classes. If the code is simple enough, then a single method is the least complex and most obvious thing you could do. If they're a little more complex, then you can break it out into 10 methods. You get a sense as to whether or not that is reasonable based on the number and type of arguments you have to pass to those methods. If you find you're passing a ton of arguments into the method, then it isn't loosely coupled, and you therefore are probably doing the wrong thing by moving that logic into methods. Or if you have to turn what were local variables into class variables just so you can share state, then you've probably done the wrong thing. Or if you have to create a new complex object that is returned from the methods to contain all the things that have changed within the method so you can update the local variables, then you have probably done the wrong thing. You'd have been better off with one really long method rather than lots of little ones.

And all that is compounded if you find you have the same problem is breaking it out into separate classes.

On the other hand, if the coupling is really loose between the main code and these helper functions or classes, then it could be a good thing (modulo performance issues or number of classes you have to class load or anything else like that).

If you find you have methods or classes that are loosely coupled both at a code level and a conceptual level (for example, Math.abs and Math.min or cryptographic utilities, etc., or big chunks like record builders vs. record writers), then breaking them out is totally sensible, two emphatic thumbs up.

I think the principle is, make every class and every function fight for its survival. Contrary to conventional wisdom, don't fear a big method. Use information hiding to your advantage (including hiding information _within_ a class -- fear global variables, but fear instance variables too when a local variable is sufficient). Favor black-box unit testing and don't weaken information hiding for the sake of testing without feeling **very** guilty about it. Instead, consider flaws in your system design that force you to weaken information hiding for the sake of adequate testing (there is very likely something wrong with your design), or think of better ways of testing that exercise the private method without having to expose it. And of course, if you take what was a private method or private field and make it into a non-private class, then it weakens information hiding (and thus exposes your implementation). Only if the concept truly was essential to the design, and the design was incorrect before, should it be made public (i.e. it has fought for its survival and won).

And when we do code reviews, we should be brutal about killing off as many of these classes as we can, question their reason for being, and force them to fight for their survival.

If I can fully test a private method through the public API, then I have done myself a huge favor should I refactor that class in the future. All my tests continue to work unchanged! If I have to test that private method directly, then my tests are at risk of being invalidated when I refactor that method. This makes them less useful. If the method really should be separate and distinct from the class in question, then that is a worthy change. Otherwise, probably the only real value you got by moving it out, was that it forced you (the person making the change) to think through the problem and load it into your head so you understood it, but the next person who comes along will be confused, because the inherent complexity wasn't changed, just moved around.

rbair23: >After "collapsing" the classes extracted for testability and moving the facility classes into their proper homes, we can have just 14 classes in workflows.handle that are not purely for DI or syntactic sugar (e.g. enums to improve readability).
>
>@rbair23 Do you suggest we do this PR now or later when we become clear on facilities?

I would like a clear design I can reason about. I saw in `develop` we now have performance regression due to this refactoring:

>If you make a method (or class) visible for testing, you make it harder to understand the actual codebase, at the same time you make it easier to white-box test. You might also impact performance (number of classes, methods, interfaces, etc)

I'm not surprised that we had an unexpected performance regression, because we made it harder to understand the actual codebase, and these things happen. As an industry we focus so much on DRY and abstraction. You've no doubt heard the old saw ["We can solve any problem by introducing an extra level of indirection"](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering). Except it isn't true. Every extra level of indirection makes individual parts of the system easy to understand, but makes the overall system incredibly difficult to understand, and makes the performance characteristics impossible to understand.

Neeharika-Sompalli: Opened a clean up PR to address naming changes and removing few classes [here](https://github.com/hashgraph/hedera-services/pull/13996).

jsync-swirlds (dismissed):

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: > I saw in develop we now have performance regression due to this refactoring

The reason for a performance regression is using Dagger for providing preHandleResult where needed. Dagger computes it once and is re-used when it is requested in that scope later. Unfortunately we were using it in logger for the first time. So, we had a performance regression as far as I understood. 

I have fixed that in PR [here](https://github.com/hashgraph/hedera-services/pull/13996.). We will remove the dagger for provisioning if that helps for clarity.

hedera-node/docs/design/app/workflows.md
```diff
@@ -17,23 +17,29 @@ The ingest-workflow is single-threaded, but multiple calls can run in parallel.
...
+         and `Transaction Prechecks` found [here](transaction-prechecks.md) are different.
+7. **Get payer account.** The account data of the payer is read from the latest immutable state.
+8. Account Balance
    1. **Check account balance.** The account of the payer is checked to ensure it is able to pay the fee.
    2. **Estimate fee.** Compute the fee that is required to pay for the transaction.
-6. **Verify payer's signature.** The signature of the payer is checked. (Please note: other signatures are not checked here, but in later stages)
-7. **Submit to platform.** The transaction is submitted to the platform for further processing.
-8. **TransactionResponse.** Return `TransactionResponse`  with result-code.
+9. **Verify payer's signature.** The signature of the payer is checked. (Please note: other signatures are not checked here, but in later stages)
+10. **Submit to platform.** The transaction is submitted to the platform for further processing.
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Ah yes, didn't look at all steps in Ingest workflow. Will fix. Thanks @netopyr

hedera-node/docs/design/app/workflows.md
```diff
@@ -44,7 +50,10 @@ It iterates through each transaction and initiates the pre-handle workflow in a
 The workflow consists of the following steps:
 
 1. **Parse Transaction.** The transaction arrives as a byte-array. The required parts are parsed and the common information is validated.
-2. **Call PreTransactionHandler.** Depending on the type of transaction, a specific `PreTransactionHandler` is called. It validates the transaction-specific parts and pre-loads data into the cache. It also creates a `TransactionMetadata` and sets the required keys.
+2. **Call PreTransactionHandler.** Depending on the type of transaction, a specific `PreTransactionHandler` is called. 
+   It validates the transaction-specific parts and pre-loads data into the cache. It also creates a `TransactionMetadata` and sets the required keys.
+   It also calls `pureChecks` method that does validations based on the respective handler's transaction body.
```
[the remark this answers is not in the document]
Neeharika-Sompalli: yes, fixed. Thank you!

hedera-node/docs/design/app/workflows.md
```diff
@@ -17,23 +17,29 @@ The ingest-workflow is single-threaded, but multiple calls can run in parallel.
...
 The ingest workflow consists of the following steps:
 
-0. **Check node.** The node is checked to ensure it is not in a state that prevents it from processing transactions.
-1. **Parse transaction.** The transaction arrives as a byte-array. The required parts are parsed and the structure and syntax are validated.
-2. **Deduplicate.** The transaction is checked to ensure it has not been processed before.
-3. **Check throttles.** Throttling must be observed and checked as early as possible.
-4. **Get payer account.** The account data of the payer is read from the latest immutable state.
-5. Account Balance
+1. **Check node.** The node is checked to ensure it is not in a state that prevents it from processing transactions.
+2. **Check timeBox** Checks whether the transaction duration is valid as per the configuration for valid durations
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Thanks @netopyr

jsync-swirlds (approved):

tinker-michaelj (dismissed):

tinker-michaelj (approved):

jsync-swirlds (approved):

hedera-node/docs/design/app/workflows.md
Neeharika-Sompalli: There are more changes merged to develop. I will update this document in next PR.

Neeharika-Sompalli (commented):
I will open PR for updating this document with latest changes in develop that removed Dagger subcomponents. Thanks

PR #13937, opened 2024-06-19
chore: Extract key verification from HandleContext

### LINKED ISSUE #13934: Extract key verification functionality

netopyr:
**Description**:

This PR extracts key verification from HandleContext

**Related issue(s)**:

Fixes #13934 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/key/KeyVerifier.java
```diff
@@ -53,26 +49,15 @@ public interface KeyVerifier {
...
      */
     @NonNull
     SignatureVerification verificationFor(@NonNull Key key, @NonNull VerificationAssistant callback);
-
-    /**
-     * Look for a {@link SignatureVerification} that applies to the given hollow account.
-     * @param evmAlias The evm alias to lookup verification for.
-     * @return The {@link SignatureVerification} for the given hollow account.
-     */
-    @NonNull
```
Neeharika-Sompalli, line 69: Why should these be removed from here and add to `AppKeyVerifier` ? I must be missing something

netopyr: This functionality is not used by any of the services and therefore not needed in the SPI (unless we plan to use it soon). 

`AppKeyVerifier` defines additional functionality that is only used internally by `hedera-app`. An interface is necessary because there are at least two implementations.

Neeharika-Sompalli (approved):
LGTM!

Neeharika-Sompalli (approved):
Makes sense to me. Thanks @netopyr

lukelee-sl (approved):
LGTM

PR #13948, opened 2024-06-20
chore: Extract record builder functionality from HandleContext

netopyr:
**Description**:

This PR extracts the record builder functionality from HandleContext

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
```diff
@@ -0,0 +1,77 @@
...
+     * Returns a record builder for the given record builder subtype.
+     *
+     * @param recordBuilderClass the record type
+     * @param <T> the record type
+     * @return a builder for the given record type
+     * @throws NullPointerException if {@code recordBuilderClass} is {@code null}
+     * @throws IllegalArgumentException if the record builder type is unknown to the app
+     */
+    @NonNull
+    <T> T current(@NonNull Class<T> recordBuilderClass);
```
Neeharika-Sompalli, line 35: nit : May be can be named as `getOrCreate()`. So it looks like `context.recordBuilders().getOrCreate(TokenAccountWipeRecordBuilder.class)`?

Neeharika-Sompalli (approved):
LGTM other than one comment. Thanks @netopyr

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
netopyr: Great idea! I will change the name.

Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr

Neeharika-Sompalli (approved):

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
```diff
@@ -0,0 +1,77 @@
...
+     * Returns a record builder for the given record builder subtype.
+     *
+     * @param recordBuilderClass the record type
+     * @param <T> the record type
+     * @return a builder for the given record type
+     * @throws NullPointerException if {@code recordBuilderClass} is {@code null}
+     * @throws IllegalArgumentException if the record builder type is unknown to the app
+     */
+    @NonNull
+    <T> T getOrCreate(@NonNull Class<T> recordBuilderClass);
```
david-bakin-sl, line 35: Should these arguments be `final`? (all methods)

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return the new child record builder
+     * @param <T> the record type
+     * @throws NullPointerException if {@code recordBuilderClass} is {@code null}
+     * @throws IllegalArgumentException if the record builder type is unknown to the app
+     */
+    @NonNull
+    <T> T addChildRecordBuilder(@NonNull Class<T> recordBuilderClass);
+
+    /**
+     * Adds a removable child record builder to the list of record builders. Unlike a regular child record builder,
```
david-bakin-sl, line 51: Reading this comment and the comment above at `addChildRecordBuilder` I don't get the difference between them, unless somehow it is that one is "removed" and the other "reverted", but in that case I don't understand the terminology.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
netopyr: No, `RecordBuilders` is an interface.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
netopyr: In one case, the child record builder is removed from the list if the parent transaction fails, meaning the child record will not appear in the stream. In the other case, it is reverted if the parent transaction fails, meaning there will be a record in the record stream, but it has the status `REVERTED_SUCCESS`.

david-bakin-sl: TY for the explanation.  But ... can it be memorialized in the comment?

netopyr: Sure, I can do it in a follow-up PR.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/RecordBuilders.java
david-bakin-sl: ok.

david-bakin-sl (approved):
LGTM!

Nana-EC (approved):

PR #13959, opened 2024-06-20
chore: add delegate call tests

lukelee-sl:
**Description**:
Add tests to ensure that calling `cryptoTransfer()` and `hbarApprove()` via `delegate call()` results in a `CONTRACT_REVERT`
**Related issue(s)**:

**Checklist**


hedera-node/test-clients/src/main/resource/contract/contracts/HRC632Contract/HRC632Contract.sol
```diff
@@ -19,4 +19,12 @@ contract HRC632Contract is HederaAccountService {
         responseCode = HederaAccountService.hbarApprove(owner, spender, amount);
         require(responseCode == HederaResponseCodes.SUCCESS, "Hbar approve failed");
     }
+
+    function hbarApproveDelegateCall(address owner, address spender, int256 amount) external {
+        (bool success, bytes memory result) = precompileAddress.delegatecall(abi.encodeWithSignature("hbarApproveCall(address,address,int256)", owner, spender, amount));
+        if (!success) {
+            revert ("hbarApprove() Failed");
```
david-bakin-sl, line 26: "Failed" -> "Failed As Expected" perhaps?

hedera-node/test-clients/src/main/resource/contract/contracts/AtomicCryptoTransfer/AtomicCryptoTransfer.sol
```diff
@@ -13,4 +13,13 @@ contract CryptoTransferV2 is HederaTokenService {
             revert ("Crypto Transfer Failed");
         }
     }
+
+    function transferMultipleTokensDelegateCall(IHederaTokenService.TransferList memory transferList,
+        IHederaTokenService.TokenTransferList[] memory tokenTransfers) external {
+        (bool success, bytes memory result) = precompileAddress.delegatecall(abi.encodeWithSignature("cryptoTransfer(((address,int64,bool)[]),(address,(address,int64,bool)[],(address,address,int64,bool)[])[])", transferList, tokenTransfers));
+        if (!success) {
+            revert ("Crypto Transfer Failed");
```
david-bakin-sl, line 21: "Failed" "Failed As Expected" perhaps?

david-bakin-sl (approved):
Unfortunate that in the contracts here the `revert` means that the _test_ will succeed.  Yet it makes sense that the tests are written that way, because you also want to test the record stream.  

I suggest below changing the revert message in each method to make it clear the failure is _expected_.  Alternatively the _method names_ could be changed to indicate that the method is expected to always revert.  Anyway, I'd prefer _something_ to make it extra clear to the reader.

david-bakin-sl (approved):
LGTM!

PR #13969, opened 2024-06-21
chore: refactor to use `ThrottleUsageSnapshot`

### LINKED ISSUE #13878: Refactor `DeterministicThrottle` to use PBJ `ThrottleUsageSnapshot` directly
### Problem

`DeterministicThrottle` defines [its own `UsageSnapshot` record](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/throttles/DeterministicThrottle.java#L253) instead of using the equivalent [PBJ type](https://github.com/hashgraph/hedera-protobufs/blob/e19bb9758a3c22b2b6abe5427a58f3a787a2d245/services/state/throttles/throttle_usage_snapshots.proto#L43). This is not only redundant, it incurs overhead when resetting the throttle state at the beginning of a transaction.

### Solution

Replace `DeterministicThrottle.UsageSnapshot` with `ThrottleUsageSnapshot`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13878


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

Neeharika-Sompalli (approved):

PR #13974, opened 2024-06-21
chore: Extract throttle functionality from HandleContext

### LINKED ISSUE #13972: Extract throttle functionality

netopyr:
**Description**:

This PR extracts the throttle functionality from HandleContext.

**Related issue(s)**:

Fixes #13972 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/throttle/ThrottleAdviser.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.spi.throttle;
+
+import com.hedera.hapi.node.base.HederaFunctionality;
+
```
david-bakin-sl, line 20: Perhaps this deserves a javadoc at the interface level?

Perhaps especially, since, at the interface level, the comment at L36-37 makes no sense since it isn't at all clear _what the child transactions are it is talking about_.  Those are injected into the implementation's constructor, and don't appear here at all!
* (Made more sense when this was part of the `HandleContext` and so much more was going on over there.)

david-bakin-sl (commented):
One comment here, asking for a clarifying comment in the code. Otherwise LGTM.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/AppThrottleAdviser.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private final NetworkUtilizationManager networkUtilizationManager;
+    private final Instant consensusNow;
+    private final RecordListBuilder recordListBuilder;
+    private final SavepointStackImpl stack;
+
+    @Inject
+    public AppThrottleAdviser(
+            @NonNull final NetworkUtilizationManager networkUtilizationManager,
+            @NonNull final Instant consensusNow,
```
mhess-swl, line 53: Some documented context on this timestamp would probably be helpful

mhess-swl (approved):

tinker-michaelj (approved):
LGTM, ty @netopyr !

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/AppThrottleAdviser.java
netopyr: It contains the consensus time of the current transaction. We use it all over the place. Do you look for documentation that explains consensus time in general and how we use it?

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/throttle/ThrottleAdviser.java
netopyr: Oh, an oversight. I will add interface level documentation in a follow-up PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/AppThrottleAdviser.java
mhess-swl: Yes, the definition of `consensusNow` is clear and ubiquitous. What isn't clear is the scope of `consensusNow`, especially in relation to the other parameters here. With all the dagger refactoring I'm quite confused about all the scopes, but perhaps I just need to get accustomed to it

PR #13975, opened 2024-06-21
Metric label proposal

hendrikebbers:
Proposal for the metric label api.

PR of draft with discussion history: https://github.com/hashgraph/hedera-services/pull/13809


rbair23 (dismissed):
+1

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,341 @@
...
+Since no class of the snapshot API is public today (based on the internal usage of `nodeId`) we plan to move parts
+of the snapshot API to the public API.
+
+### Migration
+
+Next to extending the public API we need to migrate the current metrics to the new concept.
+
+#### Migration of the `category`, `name`, and `nodeId`
+
+The `nodeId` information of a metric will be migrated to labels.
```
anthony-swirldslabs, line 204: The proposal looks great overall. However, the only part that I cannot find here is how exactly it is proposed to add the new `nodeId` labels given that the `Label`'s value is a `String`. Would the code using this API (or the `MetricsFactory` implementation) just use the `NodeId.toString()`, or there will be a different mechanism to format a `NodeId` as a `String` for the purpose of putting it into a `Label`'s value ? Or perhaps the `Label` could be a generic type like `record Label<T>(String key, T value){}` and just use the `value.toString()` to get the actual string that you use as the value of the label? 

These are just some random options, and personally I don't have any preferences. But I'd expect this document to add some details about the `NodeId` label migration because currently it's unclear.

anthony-swirldslabs (approved):
+0.9

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: Good point. Today the nodeID is already a label for prometheus. In our prometheus endpoint we therefore migrate a NodeId to a string. The same code will be used in future. The main difference is that this won't happen hidden in the prometheus endpoint code but when creating the `Metrics` instance for a platform.

hendrikebbers: @anthony-swirldslabs do you want me to update the proposal with that information and create a new PR?

anthony-swirldslabs: @hendrikebbers Thanks for clarifying. Given that this update wouldn't be material, I'd be okay with pushing it into this same PR. Another option would be to get this one merged, and then publish a tiny update PR. Or you can leave things as they are, given that you confirmed that you have a clear plan for this particular aspect. I'm good either way. I'll leave it up to you and other reviewers to decide.

Please feel free to resolve this conversation if you have enough approvals and would like to merge the current version of the proposal. Thanks!

hendrikebbers: adding @poulok for feedback to get the process right

poulok: If you only need to add clarifying comments, i consider that immaterial and can be updated on this PR

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,341 @@
...
+By adding labels to the metrics module a single metric is now uniquely defined by its name and the labels.
+For example, metrics that counts the number of requests can have a label called `method` that defines if the request was a `GET` or `POST` request.
+That metrics are defined by the name "api_http_requests_total" and the label `method`.
+By doing so the monitoring system can create a dashboard that shows the number of requests per method.
+In the given example we would have 2 metric instances: one for `GET` and one for `POST`:
+
+- Metric 1: `api_http_requests_total{method="GET"}`
+- Metric 2: `api_http_requests_total{method="POST"}`
+
+While today the name of a metric is the unique identifier of a metric the name and the labels are the unique identifier of a metric when using labels.
```
poulok, line 60: I had to read this sentence a few times to get it. I suggest a minor rewording (immaterial):

```suggestion
While today the name of a metric is the unique identifier of a metric, the unique identifier will become the name and the labels once labels are added.
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,341 @@
...
+Like the name or category, the labels are defined when a metric is created and cannot be changed later.
+Based on that the MetricConfig class will be extended by a new method called `withLabels`:
+
+```java
+abstract class MetricConfig {
+    
+    //...
+    
+    @NonNull
+    MetricConfig withLabels(@NonNull Label... labels) {...}
```
poulok, line 115:
```suggestion
    MetricConfig withLabels(@NonNull Label... label) {...}
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,341 @@
...
+For example, metrics that counts the number of requests can have a label called `method` that defines if the request was a `GET` or `POST` request.
+That metrics are defined by the name "api_http_requests_total" and the label `method`.
+By doing so the monitoring system can create a dashboard that shows the number of requests per method.
+In the given example we would have 2 metric instances: one for `GET` and one for `POST`:
+
+- Metric 1: `api_http_requests_total{method="GET"}`
+- Metric 2: `api_http_requests_total{method="POST"}`
+
+While today the name of a metric is the unique identifier of a metric the name and the labels are the unique identifier of a metric when using labels.
+By doing so both metrics can use the same name but are still unique as long as the labels are different.
```
poulok, line 61: Is this desirable? It seems quite confusing to me. Why not keep the unique identifier the same as before?

For example, it seems confusing to allow two metrics with the name `api_http_requests_total' with labels `method` and `handler` but allow another metric to be called `api_http_requests_total` with only the label `method`. To me, this should not be allowed.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,341 @@
...
+```java
+
+interface Metrics {
+    
+    // old methods:
+    // Object getValue(@NonNull String category, @NonNull String name);
+    // void remove(@NonNull String category, @NonNull String name);
+    
+    @Nullable
+    Object getValue(@NonNull String category, @NonNull String name, @NonNull Set<Label> labels);
```
poulok, line 134: If we are going to include labels as part of the unique id, then it makes sense to me to have a`MetricId` record with these values in it. Then it is very clear what is part of a metric's unique identifier.

poulok (approved):
+1

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: @anthony-swirldslabs please check my small change here: https://github.com/hashgraph/hedera-services/pull/13975/commits/37df54713e5572047b8647a22cd59e845a4a62c9

anthony-swirldslabs (approved):
+1

platform-sdk/docs/proposals/metric-labels/README.md
anthony-swirldslabs: Looks great. Thank you!

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+```
+api_http_requests_total{method="POST", handler="/messages"}
+```
+
+Here a metric is defined by a name (`api_http_requests_total`) and a set of labels (`method="POST"` and `handler="/messages"`).
+In prometheus that defines the unique identifier of a metric.
+
+When using labels to create dashboards in Grafana prometheus provides a rich query language to filter and group metrics: https://prometheus.io/docs/prometheus/latest/querying/examples/
+
+We will use the same practices and definition as in prometheus regarding labels.
```
Nana-EC, line 77: Might be good to either link to those practice or call out some important ones to avoid.

Nana-EC (commented):
Nice work
Looking forward to being able to use this in HSCS

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: @poulok can I add the links without restart voting?

- https://prometheus.io/docs/practices/naming/#labels
 - https://prometheus.io/docs/practices/instrumentation/#use-labels

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+    void handle(Transaction transaction) {
+        if(transaction.getType().equals("fileUploadTransaction")) {
+            Counter.Config configWithLabel = config.withLabels(new Label("transactionType", "fileUpload"));
+            Counter counter = metrics.getOrCreate(config);
+            counter.increment();
+        } else if(transaction.getType().equals("fileDeleteTransaction")) {
+            Counter.Config configWithLabel = config.withLabels(new Label("transactionType", "fileDelete"));
+            Counter counter = metrics.getOrCreate(config);
+            counter.increment();
+        }
```
netopyr, line 280: This is not a realistic example. We have ~50 transaction types, and the number is increasing. A method with 50 if-else branches will hopefully never make it through code review. 🙂

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The definition of the api should follow our rules regarding services as defined at https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/base/service-architecture/service-architecture.md
+
+##### Examples
+
+Here are some examples how to define a metric with labels:
+
+```java
+final Label transactionTypeLabel = new Label("transactionType", "fileUpload");
+final Counter.Config config = new Counter.Config("transactionsCategory", "transactionCount").withLabels(transactionTypeLabel);
```
netopyr, line 176: I would prefer a solution that does not require me to create `Label` instances, e.g.:
```
final Counter.Config config = new Counter.Config("transactionsCategory", "transactionCount")
    .addLabel("transactionType", "fileUpload");
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+#### Migration of the prometheus endpoint
+
+The `nodeId` of a platform is already added as a label today when using the prometheus endpoint.
+For that endpoint we just need to remove the hard coded `nodeId` label and add a general algorithm to add labels.
+
+#### Migration of the CSV exporter
+
+The CSV exporter use the `nodeId` today to create the file name.
+Since the CSV exporter is only used by the platform module we can continue with that approach.
```
netopyr, line 220: How are labels going to be used in the CSV-file?

netopyr (commented):
I find the proposal problematic. 

In Prometheus, the usage of labels is limited. You define a metric with a fixed **list** of label names. For instance you can define a  metric `Foo` with a label names `bar` and access the actual metrics with Foo{bar="a"} or Foo{bar="b"}.

The API suggests that the label names are flexible. For instance, you can define a metric `Foo` with the label `bar="a"` and another metric `Foo` with the label `baz="1"`. Does Prometheus allow this? Are the two "Foo"s connected or are these two independent metrics? What happens if I create a third `Foo` with `bar="b", baz="2"`?

I think we should follow the Prometheus approach more closely and allow a a fixed list of label names only. If they are defined only once, this also helps to prevent user errors.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+The feature is called "metric labels" and allows to add labels to metrics.
+Labels are key-value pairs that can be used to filter and group metrics.
+Why this proposal uses the name "metric labels" (like prometheus does) other libraries (like micrometer or open telemetry) use the term tags or attributes.
+
+## Motivation
+
+Today the metrics module only supports the definition of metrics without adding custom labels.
+While each metric has a hardcoded category, name, and nodeId no other custom information can be added.
+This proposal suggests adding a new concept called "metric labels" to the metrics module and
+migrate the `category`, `name`, and `nodeId` information to labels.
```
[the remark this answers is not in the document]
hendrikebbers: nothing. that sentence should have removed.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+- Make the metrics module independent of the platform module
+- Provide a public API to define labels for metrics
+- Add labels to the snapshot API
+- The output of the snapshot API should contain the labels
+- The output of the snapshot API must be compatible with the current output (no breaking changes)
+- Create better and more dynamic dashboards in Grafana by using labels
+
+## Non-Goals
+
+Even with the new concept the metrics api should be only used in our (hashgraph) code.
```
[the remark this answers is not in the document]
hendrikebbers: hashgraph is for me everything that we create in the Hedera ecosystem (inlcuding mirror node, ...). Maybe I need to find a better name.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The `Metric` interface will be extended by a new method called `getLabels()` that returns a list of labels:
+
+```java
+interface Metric {
+    
+    //...
+    
+    @NonNull
+    Set<Label> getLabels();
```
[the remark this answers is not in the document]
hendrikebbers: makes sense. Will add that

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: ok, do you have a better one? I just wanted to have any example

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: Again both might be usefull. Will add that

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: The nodeID label will be ignored and all other labels will be added to the name.

hendrikebbers: But I assume it makes sense to make that more clear

hendrikebbers:
@netopyr regarding your general comment I would say that you are wrong. Look for example at the following part of the prometheus documentation: https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors

The show how to do query with the following metric definitions:

```
http_requests_total
http_requests_total{replica="rep-a"}
http_requests_total{replica="rep-b"}
http_requests_total{environment="development"}
```

As you can see labels are `not defined as a fixed list`.

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: done

platform-sdk/docs/proposals/metric-labels/README.md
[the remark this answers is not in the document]
hendrikebbers: done

platform-sdk/docs/proposals/metric-labels/README.md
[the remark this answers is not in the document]
hendrikebbers: done

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+```java
+record Label(@NonNull String key, @Nullable String value) {}
+``` 
+
+> [!NOTE]  
+The value of a label can be `null`.
+For our prometheus endpoint a label with a `null` value will be handled as if the label would not exist.
+This would allow us to define a label without a value and with a possible extension (see "Future Work") we define a concrete value for a label later.
+Today best practice is to not use labels with a `null` value.
```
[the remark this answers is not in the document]
hendrikebbers: you are right. Maybe it is nonnull for now.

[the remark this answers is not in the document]
hendrikebbers: done

platform-sdk/docs/proposals/metric-labels/README.md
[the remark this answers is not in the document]
hendrikebbers: done

hendrikebbers:
Withdrawn by #14090

PR #13992, opened 2024-06-22
fix: passing upgrade `@HapiTest` 

### LINKED ISSUE #13953: Replace Service JRS CI tests with `@HapiTest`s
### Problem

JRS tests, although closer to production environments, are much slower, more opaque, and brittle than `@HapiTest`s that can be run locally against a subprocess network.

### Solution

Replace all tests that run in the "Hedera Node JRS Panel" test with equivalent `@HapiTest`s.

- [ ] Upgrade Software Version - https://github.com/hashgraph/hedera-services/issues/13993
- [ ]  Create @HapiTest that adds and removes Nodes in same upgrade Test
- [x] #14051

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13953


lpetrovic05 (approved):

Neeharika-Sompalli (approved):

Neeharika-Sompalli (dismissed):

Neeharika-Sompalli (dismissed):

Neeharika-Sompalli (approved):

jeromy-cannon (approved):

edward-swirldslabs (approved):
Normal platform level code would not allow use of VAR and have javadoc on class.  Since this is in the merkle API and that was recently migrated, I don' t know if the use of VAR was eliminated or if the code was brought into alignment with the platform coding standard.   I am approving with the assumption that this code would have been grandfathered in the move.

rbarkerSL (approved):

PR #13995, opened 2024-06-22
feat: repeatable `@HapiTest`

### LINKED ISSUE #13994: Repeatable `@HapiTest`s
### Problem

`@HapiTest`s would be stronger as a regression suite if they could be run in a completely deterministic mode.

A developer could then take a snapshot of records produced by, say, running all `ScheduleService` tests; do a refactor; and then verify no unintentional side effects of the change by repeating those tests and looking at the record running hash.

But there are several non-repeatable ingredients in the current `@HapiTest`s:
  1. They choose `TransactionID` valid start times based on wall clock time. (This is, of course, necessary when running against a `SubProcessNetwork`; but not for an `EmbeddedNetwork`.)
  2. They always generate new cryptographic keys.
  3. _(Occasionally)_ They use additional concurrency not controlled by JUnit Jupiter execution modes.

### Solution

1️⃣ Use synthetic time
- [x] Switch `EmbeddedNetwork` to assign "consensus times"  based on a synthetic clock.
- [x] Inject the synthetic `InstantSource` to Hedera; provide to all ingest components via Dagger.
- [x] Define an interface for `HapiSpec` operations to use when choosing `TransactionID` valid start times.
- [x] Update the `NetworkTargetingExtension` to further customize `HapiSpec` to use time backed by the same synthetic clock advanced by `EmbeddedNetwork` when appropriate.
- [x] Change `HapiSpecSleep` and `HapiSpecWaitUntil` behavior to simply "push forward" the global synthetic clock instead of actually sleeping in embedded mode.

2️⃣ Use well-known keys
- [x] Add a `KeyFactory` variant that cycles through a few dozen primitive keys (i.e., plenty to avoid repetition within a single `@HapiTest`) instead of using key generators.
- [x] Again use `NetworkTargetingExtension` to further customize `HapiSpec` to make this the default key generator when targeting an embedded network.

3️⃣ Switch `inParallel` to blocking order in embedded mode
- [x] Unless some spec will actually lose its intended effect, switch `inParallel()` to submit all sub-operations on the main thread instead of using an executor in embedded mode.

4️⃣ Add a `testRepeatable` task that is analogous to `testEmbedded` but disables all JUnit Jupiter parallelism.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #13994 
 - There is one required change on the consensus node, because we called `Instant.now()` in several places; these are all replaced by use of an `InstantSource` injected into Hedera.
 - There is _also_ one small fix to avoid a NPE when `getTxnReceipt()` is called with a missing `TransactionID#accountID`.
 - Please see the issue for detailed work items on _test-clients/_ changes.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -168,8 +168,22 @@ public final class Hedera implements SwirldMain {
...
+     * The source of time the node should use for screening transactions at ingest.
+     */
+    private final InstantSource instantSource;
+
+    /**
+     * The contract service singleton, kept as a field here to avoid constructing twice
+     * (once in constructor to register schemas, again inside Dagger component).
+     */
+    private final ContractServiceImpl contractServiceImpl;
+
```
Neeharika-Sompalli, line 186: Is this correct, why do we need ContractServiceImpl in app ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/staking/StakeRewardCalculatorImpl.java
```diff
@@ -36,15 +38,19 @@
...
     /**
      * Default constructor for injection.
      * @param stakePeriodManager the stake period manager
      */
     @Inject
-    public StakeRewardCalculatorImpl(@NonNull final StakePeriodManager stakePeriodManager) {
+    public StakeRewardCalculatorImpl(
+            @NonNull final StakePeriodManager stakePeriodManager, @NonNull final InstantSource instantSource) {
         this.stakePeriodManager = stakePeriodManager;
+        this.instantSource = Objects.requireNonNull(instantSource);
```
Neeharika-Sompalli, line 53: Do we need this here ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/EmbeddedHedera.java
```diff
@@ -16,551 +16,74 @@
...
-    private final NodeId defaultNodeId;
-    private final AccountID defaultNodeAccountId;
-    private final AddressBook addressBook;
-    private final ToyPlatform platform;
-    private final PlatformState platformState = new PlatformState();
-    private final Map<AccountID, NodeId> nodeIds;
-    private final Map<NodeId, com.hedera.hapi.node.base.AccountID> accountIds;
-    private final FakeHederaState state = new FakeHederaState();
-    private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+public interface EmbeddedHedera {
```
Neeharika-Sompalli, line 31: We should probably add more documentation when this is used and how it is different from regular tests. Need not be done in this PR. But will be helpful.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/RepeatableEmbeddedHedera.java
```diff
@@ -0,0 +1,128 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * An embedded Hedera node that handles transactions synchronously on ingest and thus
+ * cannot be used in concurrent tests.
+ */
```
Neeharika-Sompalli, line 48: We should add some more documentation that this uses synthetic time and is used for comparing recordstreams ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/ConcurrentEmbeddedHedera.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.stream.IntStream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * An embedded Hedera node that can be used in concurrent tests.
+ */
```
Neeharika-Sompalli, line 48: Should be add `CONCURRENT_EMBEDDED` tag for Leaky tests ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/EmbeddedHedera.java
Neeharika-Sompalli: We could also specify the tests that cannot use Embedded mode like the ones using ECDSA keys

Neeharika-Sompalli (approved):
Looks Great ! 🙌 tyvm @tinker-michaelj. Had a few minor comments

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/EmbeddedHedera.java
tinker-michaelj: True Neeha, created [this issue](https://github.com/hashgraph/hedera-services/issues/14046) to add a detailed _package-info.java_ that covers the expected purpose for each `HederaNetwork` implementation.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/RepeatableEmbeddedHedera.java
tinker-michaelj: Yes, will include in the _package-info.java_. 💯

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/ConcurrentEmbeddedHedera.java
tinker-michaelj: I don't think embedded mode affects which tests are leaky or not; so I believe the existing use of `@LeakyHapiTest` should be unchanged. 🤔

nathanklick (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/staking/StakeRewardCalculatorImpl.java
tinker-michaelj: Ah no Neeha, will remove in future PR. 🙏

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/ContractServiceImpl.java
```diff
@@ -16,28 +16,28 @@
...
 
 /**
  * Implementation of the {@link ContractService}.
  */
-public enum ContractServiceImpl implements ContractService {
-    CONTRACT_SERVICE;
-
+public class ContractServiceImpl implements ContractService {
     public static final long INTRINSIC_GAS_LOWER_BOUND = 21_000L;
-    public static final long HBARS_TO_TINYBARS = 100_000_000L;
```
lukelee-sl, line 33: were these no longer used?

lukelee-sl (approved):
Smart contract changes LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/ContractServiceImpl.java
tinker-michaelj: Not used...not sure why they were left here.

PR #14028, opened 2024-06-25
feat: HIP-904 Add crypto transfer fee calculations - FeeSchedules version

bilyana-gospodinova:
**Description**:
This pr is a clone of https://github.com/hashgraph/hedera-services/pull/13942. It avoids the addition of new methods in the FeeCalculator and relies on a future PR with updated fee schedules.

Fixes https://github.com/hashgraph/hedera-services/issues/13375


kselveliev (dismissed):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/UtilVerbs.java
```diff
@@ -1082,21 +1082,26 @@ private static void reduceFeeComponentsFor(
...
+    private static ByteString customFeeSchedules(String feeSchedules) {
         SysFileSerde<String> serde = new FeesJsonToGrpcBytes();
         var baos = new ByteArrayOutputStream();
         try {
-            var schedulesIn = HapiFileCreate.class.getClassLoader().getResourceAsStream("FeeSchedule.json");
+            var schedulesIn = HapiFileCreate.class.getClassLoader().getResourceAsStream(feeSchedules);
             if (schedulesIn == null) {
-                throw new IllegalStateException("No FeeSchedule.json resource available!");
+                throw new IllegalStateException(
+                        String.format("No feeSchedule.json resource available for: %s", feeSchedules));
```
zhpetkov, line 1104: Does the file name start with the lowercase letter?

bilyana-gospodinova: Yes, the original file in genesis, testnet and mainnet is named "feeSchedule.json".

netopyr: Nit: I believe concatenating the Strings is faster than using `String.format()` in this particular case.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java
```diff
@@ -200,14 +200,25 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
+        final var unlimitedAutoAssociations =
+                feeContext.configuration().getConfigData(EntitiesConfig.class).unlimitedAutoAssociationsEnabled();
+
+        if (unlimitedAutoAssociations) {
+            calculator.resetUsage();
+            calculator.addVerificationsPerTransaction(Math.max(0, feeContext.numTxnSignatures() - 1));
+
+            final var newAssociationsCount = op.tokens().size();
+            calculator.addBytesPerTransaction(newAssociationsCount);
+            return calculator.calculate();
```
netopyr, line 213: I like this! Very clear and concise.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -177,6 +182,25 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
+
+            final var unlimitedAutoAssociations =
+                    config.getConfigData(EntitiesConfig.class).unlimitedAutoAssociationsEnabled();
+            if (unlimitedAutoAssociations) {
+                final var topLevelPayer = handleContext.payer();
+                final var syntheticCreation = TransactionBody.newBuilder()
+                        .tokenAssociate(
+                                new Builder().account(account.accountId()).tokens(token.tokenId()));
+                final var fees = handleContext.dispatchComputeFees(
+                        syntheticCreation.build(), topLevelPayer, ComputeDispatchFeesAsTopLevel.NO);
```
netopyr, line 194: I believe this has to be updated because we do not want to dispatch transactions in such cases.

bilyana-gospodinova: @netopyr, Could you please elaborate on that? We call only the `computeFees` method so that the calculation will be performed the same way as in a regular token associate transaction. This way we get the same 0.05$ per association and we add them to the crypto transfer fees. Also, there is no code duplication with this approach.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/UtilVerbs.java
zhpetkov: In our case, it works faster with `String. format()`. I tested with modified `customFeeSchedules()`. The changes are:  ```
long startTime, endTime;

            // Benchmark concatenation
            startTime = System.nanoTime();
            if (true) { // Replace with actual condition
                String info = "No feeSchedule.json resource available for: " + feeSchedules;
                System.out.println(info);
                endTime = System.nanoTime();
                System.out.println("Concatenation: " + (endTime - startTime) + " ns");
            } 
```. 
When I run `createHollowAccountWithMultipleSendersAndCompleteIt()` the result are:

- Concatenation: 131542 ns
- Concatenation: 146166 ns
- Concatenation: 94000 ns
- Concatenation: 107334 ns
- Concatenation: 136042 ns
 ~ 123 017 ns
 
- String.format(): 92292 ns
- String.format(): 69333 ns
- String.format(): 81250 ns
- String.format(): 72958 ns
- String.format(): 79292 ns
~ 79 025 ns

**Review by kselveliev — APPROVED** [review]

File: hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAssociateToAccountHandler.java [hunk 4]
```diff
@@ -200,14 +200,25 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
...
-                .legacyCalculate(sigValueObj -> usageGiven(CommonPbjConverters.fromPbj(body), sigValueObj, account));
+        final var calculator = feeContext.feeCalculatorFactory().feeCalculator(DEFAULT);
+        final var unlimitedAutoAssociations =
+                feeContext.configuration().getConfigData(EntitiesConfig.class).unlimitedAutoAssociationsEnabled();
+
+        if (unlimitedAutoAssociations) {
+            calculator.resetUsage();
+            calculator.addVerificationsPerTransaction(Math.max(0, feeContext.numTxnSignatures() - 1));
+
+            final var newAssociationsCount = op.tokens().size();
```
Neeharika-Sompalli, line 211: nit : This is just `numTokens` . Need not always be `newAssociationsCount`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -177,6 +182,25 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
+
+                final var storeFactory = handleContext.storeFactory();
+                final var tokenApi = storeFactory.serviceApi(TokenServiceApi.class);
+                tokenApi.chargeNetworkFee(
```
Neeharika-Sompalli, line 198: I don't like adding `chargeNetworkFee` in multiple places inside the handler logic. It should be a part of app. But will defer to @netopyr @tinker-michaelj , what they think about this.

netopyr: I agree that this is not the right place. All fees should be calculated in the implementations of `TransactionHandler.calculateFees()`. This value is used during checks and also charged automatically in the workflow.

I am actually surprised that the `TokenServiceApi` provides this method. We should remove it IMO. Currently, it is only used in `FeeAccumulator` and we should be able to access this functionality without using `TokenServiceApi`. Same is true for the other methods that charge fees. But this is outside the scope of this PR.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
netopyr: We should not dispatch transactions within other transactions because this creates undesired dependencies. Instead, common functionality should be extracted into methods that can be called directly.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/UtilVerbs.java
netopyr: Interesting result. I would have never expected this. I stand corrected. 🙂

We should not run microbenchmarks like this, though. The outcome can be flawed in too many ways. Instead, we should always use JMH.

zhpetkov: I conducted tests with JMH and found that string concatenation is faster with minimal values than string format.

PR #14034, opened 2024-06-25
chore: esign proposal for the health monitor

### LINKED ISSUE #13957: Write design proposal for soft backpresure
Create a design proposal for the new soft backpressure mechanism.

cody-littley:
closes #13957

### DIFF
--- a/platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
+++ b/platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
@@ -1,0 +1,127 @@
+# Proposal Title Goes Here
+
+---
+
+## Summary
+
+Give a 1-3 sentence summary of the capability in the design proposal.
+
+| Metadata           | Entities                            | 
+|--------------------|-------------------------------------|
+| Designers          | Cody Littley                        |
+| Functional Impacts | Consensus node performance          |
+| Related Proposals  | n/a                                 |
+| HIPS               | n/a                                 |
+
+---
+
+## Purpose and Context
+
+The platform's current backpressure mechanism has very bad performance characteristics.
+We need to replace the current backpressure implementation with something more performant.
+
+### Requirements
+
+- The backpressure mechanism must prevent an OOM in the event of any arbitrary component
+  becoming arbitrarily slow.
+- The backpressure mechanism should not slow down critical processes when enganged.
+  Specifically, transaction handling should not be impeded when backpressure is engaged.
+  This includes slowdowns due to the transaction handling thread being starved of work.
+- The backpressure mechanism should not be overly aggressive. That is, we shouldn't see
+  it engaging under normal circumstances on nodes that are not in a state of distress.
+- The implementation of the backpressure mechanism should be sufficiently clean and
+  maintainable. A performant solution that is too complex is not a solution we should
+  accept as a solution.
+
+### Design Decisions
+
+This change can be split into two distinct sub-problems: the detection of a system in duress, and the action that
+is taken to apply backpressure on a distressed system.
+
+#### Detection
+
+The current platform is built from a number of "components". A component is usually a unit of logic that runs on its
+own thread (or logical thread). Each component has a queue (or logical queue) of data it is ingesting. If a component
+produces output, that output is enqueued within the queues of other components.
+
+For each component with a queue, define a preferred size. The queue will not apply hard backpressure even if the
+size of the queue exceeds the preferred size. That is, insertion into the queue will never block. However, a queue
+whose size exceeds its preferred size will be considered to be "unhealthy".
+
+In the background, a new component called the "health monitor" will periodically check the health of each queue.
+Checking the health of a queue is cheap (i.e. an atomic long read). The frequency of polling is configurable, with the
+default being 10hz.
+
+The health monitor records the length of time that each queue has been in an unhealthy state. If a queue has been
+unhealthy for one minute, that is a much stronger signal than if a queue has been unhealthy for a few hundred
+milliseconds. If a previously unhealthy queue is observed as being healthy, the timer for that queue is immediately
+reset to 0.
+
+When the health monitor reports the total health of the system, it reports that value as a duration. Specifically,
+it reports the amount of time that the most unhealthy queue has been unhealthy. If there is one queue that has been
+unhealthy for 10s and another that has been unhealthy for 100s, it will report that the system unhealthiness factor
+is 100s. If the second queue clears its backlog and becomes healthy, it will then report that the system unhealthiness
+factor is 10s.
+
+The health monitor publishes health reports over a wire. If it sends a duration of 0s, that means that the system is in
+a healthy state. As the system becomes more and more unhealthy, it will periodically publish higher and higher
+durations.
+
+#### Reaction
+
+Various subsystems react to the health monitor's reports in different ways.
+
+##### Event Creation
+
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the creation of self
+events is disabled. Once the system reports an unhealthy duration below X seconds, event creation may resume.
+
+##### Gossip
+
+The sync gossip algorithm has the concept of a "sync permit". In order to initiate a sync operation, a permit must
+be acquired. When the sync is completed, the permit is released. If there are 5 permits available (configurable),
+then at any point in time there may be at most 5 sync operations in progress.
+
+When the health monitor reports that the system is becoming healthy, the gossip system takes no action until the
+unhealthy duration exceeds the grace period (a configurable duration). This is designed to prevent the gossip
+subsystem from reacting too aggressively to little blips.
+
+When the system is unhealthy for a duration that exceeds the grace period, the gossip system will begin revoking
+permits. Revoking a permit does not halt a sync operation that is currently in progress, but it can prevent a new
+sync operation from starting. For each second that the system is unhealthy beyond the grace period, permits are revoked
+at the rate specified by configuration.
+
+When the system eventually becomes healthy again, permits are gradually returned according to the rate specified in
+configuration.
+
+There is a special edge case when permits are being returned. Suppose the system has been unhealthy for a while, and
+then suddenly the backlog of work is cleared and the system becomes healthy. If all permits have been revoked, then
+it may take several seconds for the first permit to be returned. This means that a healthy system may have zero permits
+for a time after becoming healthy. To combat this problem, there is a configurable "minimum healthy permits" value.
+When the system is healthy, this is the minimum number of un-revoked permits that will be available. Note that this
+value is not respected in an unhealthy system, which should eventually have 0 permits if the problem is not resolved.
+
+##### Acceptance of application transactions
+
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the acceptance of
+application transactions is disabled. Once the system reports an unhealthy duration below X seconds,
+application transactions are accepted again.
+
+Although the acceptance of transactions and the creation of events can be configured with different thresholds,
+the default configuration disables both at the same threshold.
+
+#### Alternatives Considered
+
+Although various families of algorithms have been considered, there have been no viable
+alternate proposals brought to my attention.
+
+## Test Plan
+
+### Unit Tests
+
+All new code should be completely covered by unit tests.
+
+### Performance Tests
+
+We should run this code in full sized networks under load. We should observe the long term behavior, and ensure
+that the net result of the changes is improved performance of the system as a whole.


platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+durations.
+
+#### Reaction
+
+Various subsystems react to the health monitor's reports in different ways.
+
+##### Event Creation
+
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the creation of self
+events is disabled. Once the system reports an unhealthy duration below X seconds, event creation may resume.
```
anthony-swirldslabs, line 77: The proposal looks good overall. However, there's quite a number of configurable parameters that can change the performance profile of this implementation significantly. 

They will need tuning initially. And they will need re-tuning when the code in hedera-services changes significantly enough to make the old parameter values inefficient, which I'm afraid may happen more often than we'd like.

Apart from this concern, the proposal looks good to me.

anthony-swirldslabs (approved):
+0.75

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Detection
+
+The current platform is built from a number of "components". A component is usually a unit of logic that runs on its
+own thread (or logical thread). Each component has a queue (or logical queue) of data it is ingesting. If a component
+produces output, that output is enqueued within the queues of other components.
+
+For each component with a queue, define a preferred size. The queue will not apply hard backpressure even if the
+size of the queue exceeds the preferred size. That is, insertion into the queue will never block. However, a queue
+whose size exceeds its preferred size will be considered to be "unhealthy".
```
rbair23, line 49: The fundamental problem I have this proposal is this line: "define a preferred size". It may be that any solution we devise has some metric that must be defined. The fewer of these we have, the better. We should strive for a design with the fewest knobs possible. The more knobs, the more likely we get one wrong, leading to aberrant behavior.

I think we should be thinking about larger "chunks". The current breakdown of components and health checking between components is too fine-grained. Jasper also opened my mind to another problem we really need to think about: each task added to the fork-join pool may take an unknown amount of time to execute. Or a single component might put lots and lots of thing son the FJP. In both cases, we have the potential for a FJP DOS. Even if we went to the trouble of adopting this design, we are likely to find it doesn't actually help, because the FJP is mismatched on # worker threads and type and shape of tasks added to the FJP.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+In the background, a new component called the "health monitor" will periodically check the health of each queue.
+Checking the health of a queue is cheap (i.e. an atomic long read). The frequency of polling is configurable, with the
+default being 10hz.
+
+The health monitor records the length of time that each queue has been in an unhealthy state. If a queue has been
+unhealthy for one minute, that is a much stronger signal than if a queue has been unhealthy for a few hundred
+milliseconds. If a previously unhealthy queue is observed as being healthy, the timer for that queue is immediately
+reset to 0.
```
rbair23, line 58: What is being used as a timer and what frequency does that "clock" operate at?

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+When the health monitor reports the total health of the system, it reports that value as a duration. Specifically,
+it reports the amount of time that the most unhealthy queue has been unhealthy. If there is one queue that has been
+unhealthy for 10s and another that has been unhealthy for 100s, it will report that the system unhealthiness factor
+is 100s. If the second queue clears its backlog and becomes healthy, it will then report that the system unhealthiness
+factor is 10s.
```
rbair23, line 64: What is the argument for choosing 100s instead of 110s?

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+The health monitor publishes health reports over a wire. If it sends a duration of 0s, that means that the system is in
+a healthy state. As the system becomes more and more unhealthy, it will periodically publish higher and higher
+durations.
```
rbair23, line 68: "over a wire" -- does this mean a network connection, or a wire in the wiring framework? I'm assuming the latter, although initially I thought it was the former.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+it may take several seconds for the first permit to be returned. This means that a healthy system may have zero permits
+for a time after becoming healthy. To combat this problem, there is a configurable "minimum healthy permits" value.
+When the system is healthy, this is the minimum number of un-revoked permits that will be available. Note that this
+value is not respected in an unhealthy system, which should eventually have 0 permits if the problem is not resolved.
+
+##### Acceptance of application transactions
+
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the acceptance of
+application transactions is disabled. Once the system reports an unhealthy duration below X seconds,
+application transactions are accepted again.
```
rbair23, line 108: Maybe this could be changed to just say, if the node is unhealthy, then it won't accept new transactions. We should elsewhere define unhealthy as a boolean toggled based on some criteria. It could be just duration exceeding X seconds, or could be based on exponential backoff, etc, but this part of the specification doesn't need to define how we compute whether the node is healthy or not, we just need to say, when it is unhealthy, we don't accept new transactions. We should also describe why this is the case. I think the answer is, because we don't want to create a self-event, or because we won't be creating self-events and this will cause transactions to timeout.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+##### Gossip
+
+The sync gossip algorithm has the concept of a "sync permit". In order to initiate a sync operation, a permit must
+be acquired. When the sync is completed, the permit is released. If there are 5 permits available (configurable),
+then at any point in time there may be at most 5 sync operations in progress.
+
+When the health monitor reports that the system is becoming healthy, the gossip system takes no action until the
+unhealthy duration exceeds the grace period (a configurable duration). This is designed to prevent the gossip
+subsystem from reacting too aggressively to little blips.
```
rbair23, line 87: Similar to my comment in "Acceptance of application transactions", should the gossip system even be aware of any of this nuance? Or is it enough to just say "The node is unhealthy, so I'm backing off" or "The node is unhealthy so I'm shutting things down" and let some other code determine what constitutes healthy or unhealthy based on the monitor data?

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
cody-littley: > we are likely to find it doesn't actually help

Initial testing shows significant improvement in the health of the system.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+The platform's current backpressure mechanism has very bad performance characteristics.
+We need to replace the current backpressure implementation with something more performant.
+
+### Requirements
+
+- The backpressure mechanism must prevent an OOM in the event of any arbitrary component
+  becoming arbitrarily slow.
+- The backpressure mechanism should not slow down critical processes when enganged.
+  Specifically, transaction handling should not be impeded when backpressure is engaged.
```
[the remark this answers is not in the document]
cody-littley: I agree. The new backpressure mechanism doesn't treat transaction handling differently. But since transaction handling is currently the system bottleneck, I think it's important to call it out specifically in the acceptance criteria for this change.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
cody-littley: One important goal of the health monitor is to reduce sensitivity to short lived blips. For example, if we get a huge burst of work from gossip, that burst of work shouldn't engage backpressure as long as we clear the backlog in a timely manner.

The idea here was to only pay attention to the longest unhealthy period, with the idea being that shorter unhealthy durations are more likely to just be blips.

In practice, there probably isn't a lot of difference between taking the sum and taking the max. In general, I'd expect for one subsystem to be more of a bottleneck than others, and for that bottleneck to trigger backpressure before other subsystems fall behind. At the very least, that is the current pattern we see today with transaction handling being the bottleneck.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
cody-littley: Wiring framework. We call the connections between components "wires".

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
cody-littley: This entire paradigm is made significantly simpler if we can make the following paradigm shift for event/creation & transaction acceptance. My goal for the health monitor change was to shift the current paradigm as little as possible, with the idea being that we should address this issue as a separate change.

# Current Algorithm

There is a method the app can call to send transactions. This puts the transactions into a queue. When the platform creates an event, it takes transactions from the queue. If the queue is full or if the platform thinks now is a bad time, it rejects transactions (the method returns a boolean to indicate if it was accepted).

# Proposed Algorithm

App registers a provider with the platform. When the platform is about to create an event, it requests up to X bytes of transactions from the app. If the app has transactions available, it provides those transactions.

The platform exposes various health factors through an API to the app. It is up to the app to determine the policy for when a user transaction should be accepted or not.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the creation of self
+events is disabled. Once the system reports an unhealthy duration below X seconds, event creation may resume.
+
+##### Gossip
+
+The sync gossip algorithm has the concept of a "sync permit". In order to initiate a sync operation, a permit must
+be acquired. When the sync is completed, the permit is released. If there are 5 permits available (configurable),
+then at any point in time there may be at most 5 sync operations in progress.
+
+When the health monitor reports that the system is becoming healthy, the gossip system takes no action until the
```
poulok, line 85:
```suggestion
When the health monitor reports that the system is becoming unhealthy, the gossip system takes no action until the
```

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+There is a special edge case when permits are being returned. Suppose the system has been unhealthy for a while, and
+then suddenly the backlog of work is cleared and the system becomes healthy. If all permits have been revoked, then
+it may take several seconds for the first permit to be returned. This means that a healthy system may have zero permits
+for a time after becoming healthy. To combat this problem, there is a configurable "minimum healthy permits" value.
+When the system is healthy, this is the minimum number of un-revoked permits that will be available. Note that this
+value is not respected in an unhealthy system, which should eventually have 0 permits if the problem is not resolved.
+
+##### Acceptance of application transactions
+
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the acceptance of
```
poulok, line 106: How does this fit in with platform status and CHECKING? If the monitor determines that a node is unhealthy, will it put the node into CHECKING? Can an unhealthy node initiate a reconnect or teach a reconnect?

poulok (commented):
+0.5. I think this is worth a try with testing, but I'm not totally sold due to the extra complexity and number of knobs that need to be tuned per environment.

A couple of things:

1. IMO, there are some relevant sections from the template that are missing:

  - Architecture and/or Components
  - Configuration (this is spread out through the description, but having a consolidated list is useful, including the feature flag for the entire HM)
  - Metrics

2. I'd like to see a description of all the performance problems that have been observed so far with the current backpressure mechanism. Having a clear record of the problems will help us determine if this new approach is better, even if it comes with its own problems.
3. Here is another idea. Instead of monitoring queues, what if we determined the health of the node based on how far behind it is in consensus and left queues unbounded? If a node is behind other by a certain amount, something is causing the node to not keep up. Sure, this would allow a node to OOM but that's not the worst thing because it will just start up again. We would still need the platform statuses to determine if consensus is not being reached, but perhaps it is a simpler way to get most of what we want with many fewer knobs. Curious your thoughts on this.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+---
+
+## Purpose and Context
+
+The platform's current backpressure mechanism has very bad performance characteristics.
+We need to replace the current backpressure implementation with something more performant.
+
+### Requirements
+
+- The backpressure mechanism must prevent an OOM in the event of any arbitrary component
```
hendrikebbers, line 25:
```suggestion
- The backpressure mechanism must prevent an out of memory (OOM) error in the event of any arbitrary component
```

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
+# Health Monitor Proposal
+
+---
+
+## Summary
+
+A replacement for the current hard backpressure mechanism using a health monitor.
```
hendrikebbers, line 7: This reads to me that backpressure should be completely replaced by something different (called a health monitor). Later in the propasal I learned that this is just a name for more dynamic backpressure. Based on that the summary is misleading from my point of view,

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+  This includes slowdowns due to the transaction handling thread being starved of work.
+- The backpressure mechanism should not be overly aggressive. That is, we shouldn't see
+  it engaging under normal circumstances on nodes that are not in a state of distress.
+- The implementation of the backpressure mechanism should be sufficiently clean and
+  maintainable. A performant solution that is too complex is not a solution we should
+  accept as a solution.
+
+### Design Decisions
+
+This change can be split into two distinct sub-problems: the detection of a system in duress, and the action that
```
hendrikebbers, line 38: I do not understand this chapter? Does that mean that you will create 2 independent apis? 2 proposals? Please be more concreate.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+is taken to apply backpressure on a distressed system.
+
+#### Detection
+
+The current platform is built from a number of "components". A component is usually a unit of logic that runs on its
+own thread (or logical thread). Each component has a queue (or logical queue) of data it is ingesting. If a component
+produces output, that output is enqueued within the queues of other components.
```
hendrikebbers, line 45: That is a very good introduction even for people not aware of that,

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+
+For each component with a queue, define a preferred size. The queue will not apply hard backpressure even if the
```
hendrikebbers, line 47:
```suggestion
For each component with a queue, a preferred queue size must be defined. The queue will not apply hard backpressure even if the
```

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+size of the queue exceeds the preferred size. That is, insertion into the queue will never block. However, a queue
+whose size exceeds its preferred size will be considered to be "unhealthy".
+
+In the background, a new component called the "health monitor" will periodically check the health of each queue.
```
hendrikebbers, line 51: "health monitor" is a too generic name from my point of view.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+Checking the health of a queue is cheap (i.e. an atomic long read). The frequency of polling is configurable, with the
+default being 10hz.
```
hendrikebbers, line 53:
```suggestion
default being 10hz (10 polls per second).
```

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+The current platform is built from a number of "components". A component is usually a unit of logic that runs on its
+own thread (or logical thread). Each component has a queue (or logical queue) of data it is ingesting. If a component
+produces output, that output is enqueued within the queues of other components.
+
+For each component with a queue, define a preferred size. The queue will not apply hard backpressure even if the
+size of the queue exceeds the preferred size. That is, insertion into the queue will never block. However, a queue
+whose size exceeds its preferred size will be considered to be "unhealthy".
+
+In the background, a new component called the "health monitor" will periodically check the health of each queue.
+Checking the health of a queue is cheap (i.e. an atomic long read). The frequency of polling is configurable, with the
```
hendrikebbers, line 52: atomic integer read, right?

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
hendrikebbers: I want to decouple the monitoring thread from that. Sounds like a bad idea to have monitoring and execution using the same thread pool (if I understand it correctly)

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
hendrikebbers: that should be made clear (I had the same question). Next to that: Is it exported as a metric or how can somebody see the monitoring results?

hendrikebbers: ah, next section describes that.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
hendrikebbers: + I would like to see an overview of all needed config params (including a short description, param type and in best case default value)

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+is 100s. If the second queue clears its backlog and becomes healthy, it will then report that the system unhealthiness
+factor is 10s.
+
+The health monitor publishes health reports over a wire. If it sends a duration of 0s, that means that the system is in
+a healthy state. As the system becomes more and more unhealthy, it will periodically publish higher and higher
+durations.
+
+#### Reaction
+
+Various subsystems react to the health monitor's reports in different ways.
```
hendrikebbers, line 72: Here I would like to have an api definiton to "observe/react" on the health monitor. Since this is used at several components (as described in this chapter) a good api makes sense. I might even be good to have an "internal public api" that should be used that hides the impl detail of the health monitor.

hendrikebbers: Question: will the monitor notify (call) components or will components poll the monitor?

hendrikebbers: Question: While you describe how several internal components make use of the information of the monitor I miss an information about any metrics or general human based obersation. Is that not planned?

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+#### Alternatives Considered
+
+Although various families of algorithms have been considered, there have been no viable
+alternate proposals brought to my attention.
+
+## Test Plan
+
+### Unit Tests
+
+All new code should be completely covered by unit tests.
```
hendrikebbers, line 122: That is exatcly the point why the section should be optinal :D I would remove that since it is obvious

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+If the health monitor reports an unhealthy duration exceeding X seconds (configurable), then the acceptance of
+application transactions is disabled. Once the system reports an unhealthy duration below X seconds,
+application transactions are accepted again.
+
+Although the acceptance of transactions and the creation of events can be configured with different thresholds,
+the default configuration disables both at the same threshold.
+
+#### Alternatives Considered
+
+Although various families of algorithms have been considered, there have been no viable
```
hendrikebbers, line 115: various families of algorithms <- Here more detail is needed since otherwise nobody could add a possible solution that might not have been considered.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
```diff
@@ -0,0 +1,127 @@
...
+alternate proposals brought to my attention.
```
hendrikebbers, line 116:
```suggestion
alternate proposals brought to our attention.
```
From my point of view it would be wrong to use "I/me/my/..." in a proposal.

platform-sdk/docs/proposals/healthMonitor/healthMonitorProposal.md
edward-swirldslabs: I would be interested to know the mechanism for testing back pressure in unit tests.  That's not at all obvious when testing the whole pipeline at a unit test level.  What methodology should be used to cause back pressure to build up?  Is that covering realistic scenarios that we might see back pressure in the wild?

Issue #14042, opened 2024-06-25
Configure dev to use genesis/throttles-dev.json and preprod to use genesis/throttles.json


ElijahLynn:
### Description

We noticed that preprod environments were not loading throttles with the error:

```
2024-06-25 21:50:45.146 FATAL 570  V0490FileSchema - Throttle definitions could not be loaded from disk or from classpath
```

@tomzhenghedera mentioned he had seen this issue on our performance networks and we needed to change the path to prepend genesis. So @nirbosl and I tested this on one of the integration nodes (PR https://github.com/hashgraph/hedera-services/pull/14044) and it did load the throttles. 

### Steps to reproduce

gensis.properties

bootstrap.throttleDefsJson.resource=throttles.json (should be genesis/throttles*.json)

### Additional context

Also, how can we ensure that these files are kept up to date when there are changes made to this page or other values?

### Hedera network

other

### Version

build-develop-b43861c8

### Operating system

None

PR #14044, opened 2024-06-25
fix: update bootstrap.throttleDefsJson.resource path for bootstrap.properties for dev and preprod

ElijahLynn:
for:
* #14042

- **fix throttle path for dev bootstrap.properties**
- **fix throttle path for preprod bootstrap.properties**

Do we also need to update the compose boostrap.properties too?

https://github.com/hashgraph/hedera-services/blob/c958c4de5a29715e5732863b190bb96915138bb2/hedera-node/configuration/compose/bootstrap.properties#L1

### DIFF
--- a/hedera-node/configuration/dev/bootstrap.properties
+++ b/hedera-node/configuration/dev/bootstrap.properties
@@ -1,4 +1,4 @@
-bootstrap.throttleDefsJson.resource=throttles-dev.json
+bootstrap.throttleDefsJson.resource=genesis/throttles-dev.json
 accounts.storeOnDisk=true
 accounts.releaseAliasAfterDeletion=true
 tokens.storeRelsOnDisk=true
--- a/hedera-node/configuration/preprod/bootstrap.properties
+++ b/hedera-node/configuration/preprod/bootstrap.properties
@@ -1,5 +1,5 @@
 ledger.id=0x04
-bootstrap.throttleDefsJson.resource=throttles.json
+bootstrap.throttleDefsJson.resource=genesis/throttles.json
 accounts.blocklist.enabled=false
 accounts.blocklist.path=
 


dalvizu (approved):

Nana-EC (approved):

steven-sheehy (approved):

tinker-michaelj (approved):

poulok (approved):

rbair23 (approved):

hedera-node/configuration/dev/bootstrap.properties
```diff
@@ -1,4 +1,4 @@
-bootstrap.throttleDefsJson.resource=throttles-dev.json
+bootstrap.throttleDefsJson.resource=genesis/throttles-dev.json
```
Neeharika-Sompalli, line 1: Is this file still used? Modular code should be using `application.properties`

ElijahLynn: So, we copy the bootstrap properties from this repo into genesis.properties. We have to get this file renamed to genesis.properties in the repo I think. Until then we just copy it to the new name.

ElijahLynn: Should throttle defs be in genesis.properties or application.properties?

ElijahLynn: Would you be okay with merging as is and improve with a followup based on the outcome of this discussion?

Neeharika-Sompalli: Yes @ElijahLynn . Just was curious why this is still used. We can remove usage of this file later.

Neeharika-Sompalli (approved):

PR #14056, opened 2024-06-26
chore: inline pces proposal 2.0

### LINKED ISSUE #13777: Write proposal for elimination of accidental self branches

cody-littley:
closes #13777


rbair23 (approved):
+1

edward-swirldslabs (approved):

edward-swirldslabs:
+1

alittley (approved):
+1

platform-sdk/docs/proposals/inlinePces/inlinePcesProposal.md
```diff
@@ -0,0 +1,158 @@
...
+other events (i.e. events not created by ourselves) to gossip before they pass through the orphan buffer.
+
+### Impact on Wiring
+
+#### PCES Writer
+
+PCES Writer will be changed to use `FileChannel`. All writes will be synchronous. The PCES writer will accept an event
+as input, and will return that same event once it has been written to disk. The `flush` API will be removed
+(or, at least, no longer used by the consensus engine. It may be required for clean shutdown,
+which will be handled by another issue).
```
poulok, line 111: Does this mean the PCES Writer will be a direct scheduler component?

poulok (approved):
+1

I love that this will result in a reduction of components and wires....and complexity!

platform-sdk/docs/proposals/inlinePces/inlinePcesProposal.md
cody-littley: No, it will still be sequential. It will receive events as intake, write those events on its thread, and then pass events to its output after they are durable. What I mean by "synchronous" in this context is that the system will wait until the PCES writer has finished writing prior to using the events, as opposed to using them as they are written asynchronously.

PR #14068, opened 2024-06-26
perf: Create platform ForkJoinPool with asyncMode=true

### LINKED ISSUE #13967: Platform ForJoinPool should use async mode
### Problem

**ForkJoinPool.java:**
> `asyncMode` if true, establishes local first-in-first-out
     scheduling mode for forked tasks that are never joined. This
     mode may be more appropriate than default locally stack-based
     mode in applications in which worker threads only process
     event-style asynchronous tasks.

The platform pool is currently created by `DefaultExecutorFactory.createForkJoinPool()` with `asyncMode = false`. Newly forked tasks are executed by pool threads first and execution of already scheduled tasks may get delayed. This maybe the reason why some components with light tasks create backlog large enough to trigger back-pressure.

We should try switching to `asyncMode = true`, which fits platform's mode of execution (FIFO instead of LIFO), and evaluate its effect.

### Solution

Create the platform ForkJoinPool with `asyncMode = true`.

### Alternatives

_No response_

OlegMazurov:
**Description**:
Port #13968 to `release/0.51`.

**Related issue(s)**:

Fixes #13967

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultExecutorFactory.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/concurrent/internal/DefaultExecutorFactory.java
@@ -89,7 +89,7 @@ public class DefaultExecutorFactory implements ExecutorFactory {
         if (parallelism <= 0) {
             throw new IllegalArgumentException("parallelism must be greater than 0");
         }
-        return new ForkJoinPool(parallelism, forkJoinWorkerThreadFactory, handler, false);
+        return new ForkJoinPool(parallelism, forkJoinWorkerThreadFactory, handler, true);
     }
 
     @Override


alittley (approved):

OlegMazurov:
> Does not have a design proposal. Also, as far as I am aware, this did not show a significant improvement in testing at scale (if I'm wrong on this, please post evidence).
> 
> In general, I think we should avoid merging changes into release branches unless they are bug fixes or verified stability enhancements.

There is no design to propose. The rationale for the change was provided in #13967.
The purpose of the `ForJoinPool` parameter change is not to improve performance but provide more stability against the throbbing behavior (alternating between `ACTIVE` and `CHECKING`) of the broken back-pressure implementation.
That there is no significant change in performance from this change, either improvement or regression, is not a bad thing in that context. We have not observed the problematic behavior in any testing environment with either 40M or 20M state with the change so there is a good reason to believe that it does help.
Where there is no possibility to fix the root cause any improvement that reduces the risk should be desirable. Whether this change is appropriate for the target release should be evaluated on that basis.

cody-littley (approved):

hendrikebbers (approved):

Issue #13748, opened 2024-06-07
Use the new Statistics class in LearnerThread/LearningSynchronizer


poulok:
The `ReconnectMapStats` class already exists which tracks the necessary metrics, but it is not invoked yet so it does not do anything. This ticket is for adding code that need to determine which nodes, and how many, need to be sent to `ReconnectMapStats` so that the metrics are accurate.

The metrics must be reset at the end of the reconnect so that a subsequent reconnect starts from zero.

anthony-swirldslabs:
1st part: https://github.com/hashgraph/hedera-services/pull/14071

anthony-swirldslabs:
The second part is migrated to https://github.com/hashgraph/hedera-services/issues/14198

PR #14071, opened 2024-06-26
feat(reconnect): record metrics using ReconnectMapStats

anthony-swirldslabs:
**Description**:
Adding calls to `ReconnectMapStats` to record the new metrics related to reconnects. Please note that this is the only purpose of this PR. In this fix, the metrics are "recorded" in a default `ReconnectMapStats` interface implementation which is a no-op.

A future subsequent PR will inject a proper instance of the `ReconnectMapStats` implementation into the `LearnerPushTask` object so that the metrics can actually be emitted. However, this follow-up task is outside of scope of this particular PR.

**Related issue(s)**:

Contributes to #13748

**Notes for reviewer**:
The code builds. These changes will be properly tested once the new metrics are actually emitted.

**Checklist**


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/LearnerPushVirtualTreeView.java
```diff
@@ -338,4 +340,25 @@ public void setChild(final Long parent, final int childIndex, final Long child)
...
+
+    private boolean isLeaf(final long path) {
+        return path >= reconnectState.getFirstLeafPath() && path <= reconnectState.getLastLeafPath();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void recordHashStats(
```
artemananiev, line 352: What about other `LearnerTreeView` implementations? There are at least two more: `LearnerPushVirtualTreeView` (push model, virtual trees) and `LearnerPullVirtualTreeView` (pull model, virtual trees)

anthony-swirldslabs: The current implementation is targeting VirtualMaps specifically because regular merkle trees are too small currently to make the metrics any useful. For this reason, the `LearnerPushMerkleTreeView` is omitted.

As for the pull model, it's currently disabled under a feature flag. Additionally, I don't have sufficient expertise in this implementation. I'd suggest you to update the pull-model implementation as appropriate in a separate PR.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerTreeView.java
```diff
@@ -160,4 +161,23 @@ default void abort() {}
...
+     * whether a given child is an internal node or a leaf may depend on the LearnerTreeView implementation,
+     * and this method allows the implementation to define this logic as appropriate.
+     *
+     * @param mapStats a metrics recorder
+     * @param parent a parent
+     * @param childIndex a child index, same as a lesson query index
+     * @param nodeAlreadyPresent true if the learner tree has the query node already
+     */
+    default void recordHashStats(
+            final ReconnectMapStats mapStats, final T parent, final int childIndex, final boolean nodeAlreadyPresent) {
```
alittley, line 180: missing nullity annotations

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/LearnerPushVirtualTreeView.java
alittley, line 352: args missing nullity annotations

anthony-swirldslabs: In the latest commit, I in fact implemented the method in `LearnerPushMerkleTreeView` in order to cover all the branches that we currently use in production.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerTreeView.java
anthony-swirldslabs: Added.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/LearnerPushVirtualTreeView.java
anthony-swirldslabs: Added.

alittley (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/task/LearnerPushTask.java
```diff
@@ -55,6 +56,8 @@ public class LearnerPushTask<T> {
     private final LearnerTreeView<T> view;
     private final ReconnectNodeCount nodeCount;
 
+    private final ReconnectMapStats mapStats = new ReconnectMapStats() {};
```
hendrikebbers, line 59: I do not get this. By doing so, all methods are just empty. Based on that no call to the instance has any effect. Why do you create it at all?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
```diff
@@ -273,4 +275,26 @@ public void releaseNode(final MerkleNode node) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void recordHashStats(
+            @NonNull final ReconnectMapStats mapStats,
+            @NonNull final MerkleNode parent,
+            final int childIndex,
+            final boolean nodeAlreadyPresent) {
+        assert parent.isInternal();
```
hendrikebbers, line 288: Do we use `assert` in product code? I do not think so.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
```diff
@@ -273,4 +275,26 @@ public void releaseNode(final MerkleNode node) {
...
+        final MerkleNode child = parent.asInternal().getChild(childIndex);
+        // The child may be missing per the getChild() specification,
+        // and this method cannot reason about a `null`, so just bail.
+        if (child == null) return;
```
hendrikebbers, line 292: is that allowed in our format guideline?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/task/LearnerPushTask.java
anthony-swirldslabs: This is by design in this particular PR. This is to support an iterative approach to roll out a feature with small(er), self-sufficient PRs. Please take a look at the PR description where I explain this.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
anthony-swirldslabs: Yes, we do. At least we do in the former Platform Data code. And we do this a lot. This is done for performance reasons.

In this particular case, this is legit because the caller would only ever call this method for parents with actual children. Also, please see the javadoc for this method in the interface where it's defined.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
anthony-swirldslabs: I believe you either have a one-liner, or use a block. In this case, this is a one-liner, so it should be allowed.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/LearnerPushVirtualTreeView.java
artemananiev: The pull classes aren't used by default in `develop`, but they are fully functional and should be treated as fully supported. It would be surprising for anybody to enable a pull model and find that it doesn't support new metrics... If it's tracked in a separate ticket, it should be worked immediately after this one, not some time later

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
kfa-aguda: Just read the javadoc and it's not clear what in it mandates the use of assert. I don't think using assert is worth it here. I strongly recommend handling this as an exception instead of throwing a runtime error.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/reconnect/LearnerPushVirtualTreeView.java
anthony-swirldslabs: Okay, apparently it had an implementation for the `isLeaf` already, so I overrode the method in the `LearnerPullVirtualTreeView` as well.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
anthony-swirldslabs: I totally agree with that, and my preference has always been to use exceptions. But I've been told by @jasperpotts and @artemananiev that they use `assert` in Data code for the reasons I described above.

Since there's two reviewers who request using exceptions, I've updated the code accordingly.

PS. The javadoc for the method says `By the definition of this method, it is obvious that the parent is an internal node in the tree`. The `definition` is the first sentence of that javadoc which says `... about children of a given parent...`. And the method signature explicitly accepts a `childIndex` as the method operates on the children of the parent (again, per the definition of the method.)

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/views/LearnerPushMerkleTreeView.java
anthony-swirldslabs: Please see https://www.notion.so/swirldslabs/007-Java-Code-Style-a0130a7059b044f68447b9ca3a5242b3?pvs=4#490693eca99841ed9071158baa8a4175

litt3 (approved):

imalygin (approved):

PR #14078, opened 2024-06-27
chore: remove unneeded code

MrValioBg:
**Description**:
Remove unneeded code, since we probably won't have a use-case for transfers to override PAUSED status.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hapi/build.gradle.kts
+++ b/hapi/build.gradle.kts
@@ -31,9 +31,9 @@ tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-
 tasks.cloneHederaProtobufs {
     // uncomment below to use a specific tag
     //    tag = "v0.51.0"
+    branch = "main"
     // uncomment below to use a specific branch
     // branch = "main"
-    branch = "hip-904-proto-updates"
 }
 
 sourceSets {
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
@@ -256,21 +256,6 @@ public class TokenHandlerHelper {
     }
 
     /**
-     * Enum to determine the type of validations to be performed on the token account relationship. If the token account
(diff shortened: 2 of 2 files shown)


hapi/build.gradle.kts
```diff
@@ -32,8 +32,7 @@ tasks.cloneHederaProtobufs {
     // uncomment below to use a specific tag
     //    tag = "v0.51.0"
     // uncomment below to use a specific branch
-    //     branch = "main"
-    branch = "add_admin_key"
+    branch = "main"
```
Neeharika-Sompalli, line 35: This is already changed back to main. Please merge back latest develop.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/util/TokenHandlerHelper.java
```diff
@@ -321,25 +306,14 @@ public static TokenRelation getIfUsable(
...
         requireNonNull(tokenId);
         requireNonNull(tokenRelStore);
 
         final var tokenRel = tokenRelStore.get(accountId, tokenId);
 
         validateTrue(tokenRel != null, TOKEN_NOT_ASSOCIATED_TO_ACCOUNT);
-        if (tokenRelValidations == TokenRelValidations.REQUIRE_NOT_FROZEN) {
-            validateTrue(!tokenRel.frozen(), ACCOUNT_FROZEN_FOR_TOKEN);
-        }
+        validateTrue(!tokenRel.frozen(), ACCOUNT_FROZEN_FOR_TOKEN);
```
Neeharika-Sompalli, line 316: Son't we need to not check this for tokenReject ?

MrValioBg: The default case is that it checks if the account is frozen. This logic is redundant since it was decided that for TokenReject we won't have to bypass the frozen status

hapi/build.gradle.kts
MrValioBg: Done.

Neeharika-Sompalli (approved):

PR #14090, opened 2024-07-01
chore: Metrics label proposal

hendrikebbers:
Proposal for the metric label api.
Fixes https://github.com/hashgraph/hedera-services/issues/13885

PR of draft with discussion history: https://github.com/hashgraph/hedera-services/pull/13809
PR of first voting with discussion history: https://github.com/hashgraph/hedera-services/pull/13975


platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,369 @@
...
+
+```
+,,private_category,private_category,private_category,private_category,public_category,
+,,myCounter1,myCounter2,myCounter3,myCounter4,myDoubleGauge,
+,,1,2,3,4,1.230,4.560,
+```
+
+Here the first row defines the category of the metric and the second row the name. The 3rd row contains the values.
+An additional row between the categories and values will be added to include the labels of the metric.
+Since a metric can contain multiple labels the cell content must be quoted:
```
mxtartaglia-sl, line 331: According to the example, the label row will have a specific format that includes the value.

- what would happen with the total value of the metric (The one aggregating all values without lables)? 
- Is there any tool that parses this file automatically that depends on the format? Is this change still compatible? Who would modify that if not? 

I don't think the file is a well-structured CSV. However, after this modification, parsing the file would become more difficult, as quoting content with a comma character is insufficient to differentiate whether the comma is part of a value or an actual separator.


I wish we could move away from csv files altogether, but from what I know, we need this file for tests,

Have you considered the approach of denormalizing the metric labels? Like:
```
,,private_category,private_category,private_category, private_category
,,metric_with_label,metric_with_label,metric_with_label,metric_no_label
,,"",label_a,label_b,label_c,"", 
10,5,3,2,28
```

the quotes ("") only indicate that the first column of the label represents the total and would be the content for non-labeled metrics.
It seems more consistent with the current format.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,369 @@
...
+### Public API
+
+The metrics module will be extended by a new data type called `Label`.
+A label is a key-value pair that can be defined as a `record`:
+
+```java
+record Label(@NonNull String key, @NonNull String value) {}
+``` 
+
+The `Metric` interface will be extended by several methpds that allow to define label values to a metric:
```
mxtartaglia-sl, line 87:
```suggestion
The `Metric` interface will be extended by several methods that allow to define label values to a metric:
```

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,369 @@
...
+myDoubleGauge:,myDoubleGauge,
+myDoubleGauge:,myDoubleGauge,
+
+,,private_category,private_category,private_category,private_category,public_category,private_category,
+,,myCounter1,myCounter2,myCounter3,myCounter4,myDoubleGauge,myDoubleGauge,
+,,1,2,3,4,1.230,4.560,
+```
+
+The first section of the list is metadata that defines the file name (that includes the `nodeId`) and a list of all metric names.
+The metric category is ignored here. As you can see the 2 gauges with same name but different category are ends in 2 times the same line.
```
mxtartaglia-sl, line 318:
```suggestion
The metric category is ignored here. As you can see, the two gauges with the same name but different categories are printed twice in the same line.
```

mxtartaglia-sl (commented):
Leaving some comments,
Let me know if you want to discuss in a call.

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,369 @@
...
+    
+    /**
+    * Removes all metrics that have the same category and name as the given metric. Labels will be ignored.
+    */ 
+    void remove(@NonNull Metric metric);
+    
+    /**
+    * Removes all metrics that have the same category and name as the given metric configs. Labels will be ignored.
+    */ 
+    void remove(final @NonNull MetricConfig<?, ?> config);
```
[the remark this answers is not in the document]
hendrikebbers: I will check how often it is used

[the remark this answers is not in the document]
hendrikebbers: makes sense

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: > what would happen with the total value of the metric (The one aggregating all values without lables)? 

That is never stored. You can only get that by using a prometheus query

> Is there any tool that parses this file automatically that depends on the format?

As far as I know it is only for manual usage. Adding @netopyr and @lpetrovic05 for clarification

> I wish we could move away from csv files altogether, but from what I know, we need this file for tests

I do not think so. As far as I know it is only used by some developers.

> I don't think the file is a well-structured CSV.

I agree on that. Based on that I want to change only what is really needed and hope that we do not use/need it in future anymore. I assume my current idea is "good enough" and whoever uses the file today will ignore the new line anyway

netopyr: At least in the past, the CSV files were used everywhere, and the parsers were not very tolerant. The slightest change broke processes downstream.

hendrikebbers: lol, I always assumed that is the basic out for devs when you do not have grafana access. Let's see if @lpetrovic05 has more knowledge

lpetrovic05: CSV metrics are used to produce PDF graphs when running JRS. It would be great if JRS metrics went to prometheus.

I think the CSVs are also used by some JRS validators, @JeffreyDallas can comment on that

hendrikebbers: ok, so no open source code use them. Based on that I should not add anything to the proposal

platform-sdk/docs/proposals/metric-labels/README.md
```diff
@@ -0,0 +1,363 @@
...
+Next to extending the public API we need to migrate the current metrics to the new concept.
+
+#### Migration of the `nodeId`
+
+The `nodeId` information of a metric will be migrated to labels.
+Today we create a prometheus label out of the `nodeId` internally in our prometheus endpoint implementation.
+The same conversion can be used to create a label for the `nodeId` in future but should be moved from the prometheus endpoint to the platform code.
+The key of a metric won't change and still be defined as the combination of the `category` and `name`.
+The unique identifier of the metric is the combination key and the labels.
+That is a breaking change since today the key is the unique identifier of a metric.
```
poulok, line 271: I'm a little confused by this. How can it be that the key and the unique identifier can be different? Will `Metrics` not support getting multiple metrics from a single category/name combination?

poulok (approved):
+1

I'm a little confused on how the key and unique identifier can be different, but it's not a blocker for me.

platform-sdk/docs/proposals/metric-labels/README.md
mxtartaglia-sl: I don't know why the impact analysis should be limited only to "open source" code. 

It seems that we know:
1)  We have an impact on the CVS file,
2) The parser is used in the execution of JRS tests
3) The parser might not have a lot of tolerance for changes. 

JRS test panels did run on PRs, and I think the nightly test runs with that and uses the metrics' results to compare things. So, they seem pretty important.

If we operate with the premise that the parser can be broken with this change: 
* whose team is responsible for fixing it? 
   If it is platform responsibility, then I think it should be part of this proposal and work at the same time so people can still use the test.
   If it is not the platform's team's responsibility, we should note the impact on the responsible team and get their review, too, as they will have to fix it.

We can also move forward and see what happens along the way.

lpetrovic05 (approved):

platform-sdk/docs/proposals/metric-labels/README.md
hendrikebbers: if the labels are different, yes.

Assume you have a `HTTP_TIME` metric identifier. Now you can have a `method` label with the possible values of`GET` or `POST`. This will end in 2 instances.

poulok: Edited my original comment from "not support" to "now support".

In your example of `HTTP_TIME` with a `method` label and values `GET` and `POST`, the call to `Metrics.getOrCreate(myHttpTimeMetricKey)` will return 2 metrics. How will the caller distinguish between the two and ensure the correct one gets updated?

hendrikebbers: ahhh, got it. Here is some pseudo code:

```
MetricConfig config = new MetricConfig("category", "name").withLabels("HTTP_TYPE");

Metric metric1 = metrics.getOrCreate(config);
metric1.set(123);

final Label getLabel = new Label("HTTP_TYPE", "GET");
Metric metric2 = metrics.getOrCreate(config).withLabel(getLabel);
metric2. set(456);

final Label postLabel = new Label("HTTP_TYPE", "POST");
Metric metric3 = metrics.getOrCreate(config).withLabel(postLabel);
metric3. set(456);
```

As you can see there can be a metric with a specific label value (metric2, metric3) or even a metric without a definition for the label (metric3).

In prometheus you can than filter the 3 different metric like that:

```
category_name
category_name{http_type="GET"}
category_name{http_type="POST"}
```

How you can do more complexe queries (like get all or only specific labels) please have a look at https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors

timo0 (approved):

platform-sdk/docs/proposals/metric-labels/README.md
mxtartaglia-sl: After speaking with Hendrik, he explained why leaving this out of the proposal makes sense. I am resolving this comment.

mxtartaglia-sl (approved):
+1

PR #14105, opened 2024-07-01
chore: enable update `HapiTest`s

### LINKED ISSUE #14051: Create @hapitest that  add and removes nodes in subsequent upgrades, with non-contiguous node Ids. Do Restart and Reconnect

tinker-michaelj:
**Description**:
 - Closes #14051 by finishing and enabling `DabEnabledUpgradeTest` in CI.
 - Adds support for manipulating the `SoftwareVersion` of the events used to submit transactions to non-default nodes against an `EmbeddedNetwork`; see usage [here](https://github.com/hashgraph/hedera-services/blob/14051-enable-update-hapitests/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/records/DuplicateManagementTest.java#L162).
 
Includes required fixes for node software,
  1. Captures a node's `AddressBook` self-name in the `Node#description` field.
  2. Correctly parses and translatesa [`ServiceEndpoint`](https://github.com/hashgraph/hedera-protobufs/blob/54b258dfa27866d2654bb036112298b5a6693505/services/basic_types.proto#L1371) into the related `Node` fields (notably, the `ipAddressV4` string is the 4 octets of the IP address, not a UTF-8 string).
 
👉 *Note to reviewer:* The node changes are covered by new `@HapiTest`'s and will thus have unit test coverage when `test-clients` is moved under `hedera-app` test sources.

Also makes smaller improvements,
 - Standardizes how `SemanticVersion`s are logged to a single format that shows any non-zero `hedera.config.version` as the semver `build` part.
 - Only warns on receiving an event from a missing creator `NodeId` if the event has the current software version.
 - Removes `IsEmbeddedTest` since this is redundant; such tests are exactly those that don't use the system `InstantSource`.
 - Removes a potentially noisy log when declining to pay rewards to an account staking to a deleted node.


hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/validators/AddressBookValidator.java
```diff
@@ -151,31 +150,20 @@ public void validateAdminKey(@Nullable Key key) throws PreCheckException {
      */
     public void validateAccountId(@Nullable AccountID accountId) throws PreCheckException {
         validateAccountID(accountId, INVALID_NODE_ACCOUNT_ID);
-        validateFalsePreCheck(!accountId.hasAccountNum() && accountId.hasAlias(), INVALID_NODE_ACCOUNT_ID);
+        validateFalsePreCheck(
+                !requireNonNull(accountId).hasAccountNum() && accountId.hasAlias(), INVALID_NODE_ACCOUNT_ID);
```
iwsimon, line 154: line 152, validateAccountID() checked accountId is null, we don't need to add requireNonNull check here.

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/validators/AddressBookValidator.java
```diff
@@ -151,31 +150,20 @@ public void validateAdminKey(@Nullable Key key) throws PreCheckException {
...
-        validateFalse(endpoint.ipAddressV4().length() != 0 && !isIPv4(endpoint.ipAddressV4()), INVALID_IPV4_ADDRESS);
-    }
-
-    private boolean isIPv4(@NonNull final Bytes ip) {
-        requireNonNull(ip);
-
-        final var part = "(\\d{1,2}|(0|1)\\d{2}|2[0-4]\\d|25[0-5])";
-        final var regex = part + "\\." + part + "\\." + part + "\\." + part;
-        return ip.asUtf8String().matches(regex);
+        validateFalse(addressLen != 0 && addressLen != 4, INVALID_IPV4_ADDRESS);
```
iwsimon, line 167: This assumes the user will send exactly four bytes for IPV4. I have unit tests for Bytes.wrap("127.0.0.1") as IPV4 and will fail with addressLen !=4. I think we should allow user send any valid IPV4 as this as Bytes.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -202,44 +231,57 @@ private void extractAndReplaceArtifacts(
         }
     }
 
-    private void generateConfigPem(@NonNull final Path artifactsLoc) {
+    private void generateConfigPem(@NonNull final Path artifactsLoc, @NonNull final List<ActiveNode> activeNodes) {
```
iwsimon, line 234: We need to generate config.txt for all nodes in nodeStore, not just ActiveNode. NodeStore and StakingInforStore are not synced in this phase. We only get weight info from StakingInfoStore for active nodes.

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/validators/AddressBookValidator.java
tinker-michaelj: I fixed all these unit tests.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
tinker-michaelj: I could rename `ActiveNode` to `NonDeletedNode`; the only nodes not in this list have `deleted=true`.

What do you think? 🤔

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/validators/AddressBookValidator.java
tinker-michaelj: I just like to get rid of the IntelliJ warnings. 😳

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/validators/AddressBookValidator.java
iwsimon: Yes, I saw your change on the unit test.  You still assume the user will send exactly four bytes for IPV4.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
iwsimon: I had Streams.stream(nodeIds)
                    .map(nodeId -> nodeStore.get(nodeId.number()))
                    .filter(node -> node != null && !node.deleted())
                    .sorted(Comparator.comparing(Node::nodeId))
                    .forEach(node -> writeConfigLineAndPem(node, bw, artifactsLoc)); before which took care of the delete=true.

tinker-michaelj: Ah then the semantics haven't changed. 💯

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -221,11 +222,11 @@ private void extractAndReplaceArtifacts(
                 log.info("Finished generating config.txt and pem files into {}", artifactsLoc);
             }
             writeSecondMarker(marker, now);
-        } catch (final IOException e) {
+        } catch (final Throwable t) {
```
iwsimon, line 225: What other exception beside IOException could be here? We should add those instead of Throwable?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
```diff
@@ -0,0 +1,188 @@
...
+ * for the associated HIP-869 user stories.
+ * <p>
+ * Since this test upgrades the software version, it must run after any other test that does a restart assuming
+ * the config version is still zero.
+ */
+@Tag(UPGRADE)
+@Order(MAX_VALUE)
+@DisplayName("Upgrading with DAB enabled")
+@HapiTestLifecycle
+@OrderedInIsolation
```
iwsimon, line 88: If you could explain more of those annotations in https://github.com/hashgraph/hedera-services/blob/9c7c438fafee8eb55305f3f259acf5e97ae19a01/hedera-node/test-clients/README.md, it would be great.

iwsimon (approved):
LGTM. A couple small comments. Thanks @tinker-michaelj

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
tinker-michaelj: @JeffreyDallas found a case in the Solo environment where something other than `IOException` could be thrown...I don't remember what it was offhand. 🤔

JeffreyDallas: It was directory `data/upgrade/current` does not exist causes IllegalArgumentException from FileUtils

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
tinker-michaelj: These two are mentioned [here](https://github.com/hashgraph/hedera-services/blob/14046-spec-networks-doc/hedera-node/test-clients/README.md#meta-annotations) in the README...they are basically "syntactic sugar" for standard JUnit5 annotations.

Neeharika-Sompalli (dismissed):
LGTM! Thanks @tinker-michaelj

dalvizu (dismissed):

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
iwsimon: Thanks @JeffreyDallas. I will change to that check later.

iwsimon (approved):

JeffreyDallas (dismissed):

JeffreyDallas (approved):

Neeharika-Sompalli (approved):

iwsimon (approved):

JeffreyDallas (approved):

rbarkerSL (approved):

imalygin (approved):

PR #14108, opened 2024-07-02
feat: Implement `isAuthorizedRaw()`

### LINKED ISSUE #13560: feat: Implement HIP-632 system contract methods
Implement HIP-632 system contract methods

* [ ] design doc - #13558 
* [x] `isAuthorizedRaw` - #14108 (cherry pick to 0.52 is #14131)
* [ ] `isVirtualAddress`, `getVirtualAddresses` and `getHederaAddress`
* [ ] `isAuthorized`

lukelee-sl:
**Description**:
Implement the `isAuthorizedRaw()` function into the `HederaAccountService` system contract.

**Related issue(s)**:

Fixes #13560 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/HasCallFactory.java
```diff
@@ -72,7 +72,8 @@ public HasCallFactory(
             @NonNull final MessageFrame frame) {
         requireNonNull(input);
         requireNonNull(frame);
-        final var enhancement = proxyUpdaterFor(frame).enhancement();
+        final var proxyUpdater = proxyUpdaterFor(frame);
```
lukelee-sl, line 75: this change seems unnecessary?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/v051/Version051FeatureFlags.java
```diff
@@ -37,4 +37,9 @@ public boolean isHederaAccountServiceEnabled(@NonNull Configuration config) {
         requireNonNull(config);
         return config.getConfigData(ContractsConfig.class).systemContractAccountServiceEnabled();
     }
+
+    public boolean isHASIsAuthorizedRawMethodEnabled(@NonNull Configuration config) {
```
lukelee-sl, line 41: The name of this method could be something like `isAuthorizedRawMethodEnabled`?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if (authorized) {
+            authorized = switch (signatureType) {
+                case EC -> validateEcSignature(account.get());
+                case ED -> validateEdSignature(account.get(), key.get());
+                default -> false;};
+        }
+
+        final var gasRequirement = gasCalculator.gasCostInTinybars(HARDCODED_GAS_REQUIREMENT_GAS);
```
lukelee-sl, line 150: Should we move this up and use this amount when reporting an error in line 100?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,203 @@
...
+
+        final var result = authorized
+                ? gasOnly(successResult(encodedAuthorizationOutput(authorized), gasRequirement), SUCCESS, false)
```
lukelee-sl, line 153: we might want to set the last parameter - `isViewCall` to be true?  As this function does not change state.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/HasCallFactory.java
david-bakin-sl: yes, leftover from something I reverted when I merged your update

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: good idea

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+        final boolean authorized =
+                switch (signatureType) {
+                    case EC -> validateEcSignature(address, frame);
+                    case ED -> validateEdSignature(account, key.get());
+                    default -> false;
+                };
+
+        final var result = gasOnly(
+                successResult(encodedAuthorizationOutput(authorized), HARDCODED_GAS_REQUIREMENT_GAS), SUCCESS, true);
+        return result;
```
MrValioBg, line 154:
```suggestion
        return gasOnly(
                successResult(encodedAuthorizationOutput(authorized), HARDCODED_GAS_REQUIREMENT_GAS), SUCCESS, true);
```

MrValioBg: We could simplify this :)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /** Validate (native Hedera) ED25519 signature */
+    public boolean validateEdSignature(@NonNull final Account account, @NonNull final Key key) {
+        requireNonNull(account, "account");
+        requireNonNull(key, "key");
+
+        // Use of `com.swirlds.common.crypto.CryptographyHolder` straight from the Platform is deprecated:
+        // FUTURE: Get the `Cryptography` engine from the services app via the context (needs to be invented)
+
+        return CryptographyHolder.get()
```
MrValioBg, line 189: Validation if the account matches the one from the decoded signature is not permed, basically the `account` input is ignored. 
Since in Hedera multiple accounts can have the same key. This could lead to bug where if any address that has the same key and is passed as a parameter isAuthorizedRaw would also return true!

With @mustafauzunn we've added a case scenario that covers that, and currently it fails, due to the missing validation.
Check out this test case: `callWithDifferentAccountsWithSameKey`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+        result[63] = ov.get(); //                           v
+        System.arraycopy(signature, 0, result, 64, 32); //  r
+        System.arraycopy(signature, 32, result, 96, 32); // s
+
+        return Optional.of(result);
+    }
+
+    /** Make sure v ∈ {27, 28} - but after EIP-155 it might come in with a chain id ... */
+    @NonNull
+    public Optional<Byte> reverseV(@NonNull final byte v) {
```
MrValioBg, line 231:
```suggestion
    public Optional<Byte> reverseV(final byte v) {
```

Primitive types cannot be null

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+        requireNonNull(v, "v");
```
MrValioBg, line 232:
```suggestion
```

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+    // Besu's ECRECOVER precompile class will be used _directly_ (no need to execute bytecodes):
+    private static final ECRECPrecompiledContract ecPrecompile =
+            new ECRECPrecompiledContract(noCalculationGasCalculator);
+
+    // FUTURE: Gas for system contract method calls needs to be a) determined by measurement of
+    // resources consumed, and b) incorporated into the fee schedule
+    private static final long HARDCODED_GAS_REQUIREMENT_GAS = 1_500_000L;
+
+    public enum SignatureType {
+        Invalid,
```
MrValioBg, line 68:
```suggestion
        INVALID,
```

MrValioBg (changes requested):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+        // ECRECOVER's output always has status SUCCESS but the result Bytes are either empty or the recovered address.
+        if (ecrecoverResult.getOutput().isEmpty()) return false;
+        final var recoveredAddress =
+                ecrecoverResult.getOutput().slice(12).toArray(); // LAST 20 bytes are where the EVM address is
+
+        // ECRECOVER produced an address:  Must match our account's alias address
+        final var givenAddress = explicitFromHeadlong(address);
+        final var addressesMatch = 0 == Arrays.compare(recoveredAddress, givenAddress);
+
+        return addressesMatch;
```
MrValioBg, line 178:
```suggestion
        return 0 == Arrays.compare(recoveredAddress, givenAddress);
```

MrValioBg: If you prefer you could also keep it on separate line, but then I would probably suggest that we could use `boolean` instead of `var`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: I know the simplification but generally I prefer to do it this way - via an intermediate named local variable before returning the result - to make debugging now (and in the future!) easier.  You can see the result to be returned _before_ you actually return from the function, when in the debugger.

I've found this to be important many times when debugging.  And I like to leave it in the code because when you're debugging against a release (for example) you don't have the luxury of being able to change the code at that time.  So ... I'll leave it like this.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        // Key must match signature type
+        if (key.isPresent()) {
+            if (!switch (signatureType) {
+                case ED -> key.get().hasEd25519();
+                case EC -> false;
+                default -> false;
+            }) return bail.apply(INVALID_SIGNATURE_TYPE_MISMATCHING_KEY);
+        }
```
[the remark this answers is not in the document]
david-bakin-sl: I did this deliberately throughout this method.  There's a bunch of argument fetching and validation - and a little bit of it is specific to either EC or ED.  I felt using the enum plus having explicit switches would make it clearer to the reader at each step whether there was something going on only for EC, or only for ED, or for both.  I believe that changing some of these switches to the more concise `if` on only the particular case needed would somewhat obscure this (and increase the cognitive load of the reader).  IOW, I think the parallelism of all the similar-looking switches helps.  In the same way, in general I evaluate whether merging multiple if-conditions into one conditional expression hooked together with `&&` (or sometimes `||` helps readability or hinders it.  In short - IMO there's a readability/understandability trade-off between conciseness and "logical clarity".  Sometimes it goes one way, sometimes the other.

Do you agree (in general, and in this case specifically)?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: Eek.  Should have seen the IntelliJ warning.  TY

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: will fix, with above

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: Oh yeah, TY.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: WIll not change (see comment above).

(Unless you disagree with comment above - let me know.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/IsAuthorizedSuite.java
```diff
@@ -0,0 +1,572 @@
...
+                                CONTRACT_REVERT_EXECUTED,
+                                recordWith().status(FAIL_INVALID)));
+    }
+
+    @HapiTest
+    final Stream<DynamicTest> isAuthorizedRawAccountAndSignatureKeysCombinations() {
+        final AtomicReference<Address> accountNum = new AtomicReference<>();
+
+        return propertyPreservingHapiSpec("isAuthorizedRawAccountAndSignatureKeysCombinations")
+                .preserving(CONTRACTS_SYSTEM_CONTRACT_ACCOUNT_SERVICE_IS_AUTHORIZED_ENABLED)
```
tinker-michaelj, line 500: You could use the `@BeforeAll`/`@AfterAll` pattern like [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/opcodes/SelfDestructSuite.java#L76) to only override this property once for the entire test class and remove override at the end.

But that can be a separate PR, the tests all make sense.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
```diff
@@ -0,0 +1,266 @@
...
+
+        // Figure out what kind of signature we're dealing with thus what kind of account+key we need
+        final var signatureType =
+                switch (signature.length) {
+                    case 65 -> SignatureType.EC;
+                    case 64 -> SignatureType.ED;
+                    default -> SignatureType.Invalid;
+                };
+        if (signatureType == SignatureType.Invalid)
+            return bail.apply(FAIL_INVALID /* should be: "invalid argument to precompile" */);
```
tinker-michaelj, line 107: ⚠️ `FAIL_INVALID` should be reserved for internal system errors, not user errors.

david-bakin-sl: Will use `INVALID_TRANSACTION_BODY` as you suggested to me elsewhere.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
MrValioBg: Oh, got it. Yeah, I agree. 
Just a small suggestion then: What do you think about changing it to boolean, as I feel like var is a little bit of an "encoding" for primitive types?

MrValioBg: Oh, yes. I agree :)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java
david-bakin-sl: Good idea, I believe that too.

david-bakin-sl:
Closing, replaced by #14130 (due to problems rebasing/merging against develop).

PR #14110, opened 2024-07-02
chore: Remove `allowAutoAssociations` feature flag

### LINKED ISSUE #13954: Remove `contracts.allowAutoAssociations` flag in HIP-904
### Problem

Before HIP-904, adding auto-association slots to a contract or account required "pre-paying" for the resources that would be used by the associations.

There was no clear way to make an internal `CONTRACT_CREATE` cover this pre-payment while also respecting the precedent of child contracts "inheriting" all the Hedera-native properties of their parent contract.

Therefore we added a `contracts.allowAutoAssociations=false` property to block all use of auto-association slots with contracts.

But with HIP-904, resources used by associations are always charged at point-of-use to the payer account of the transaction that is creating the association.

So there is need to have this `contracts.allowAutoAssociations` toggle any more.

### Solution

Remove this property and all its uses.

### Alternatives

_No response_

stoyanov-st:
**Description**:

With HIP-904 we remove the pre-paying for auto association slots. This change makes the `contracts.allowAutoAssociations` flag obsolete as it was blocking the setting of `maxAutoAssociations` for `ContractCreate` and `ContractUpdate` transactions.
It is no longer the expected behaviour and we are charging for associations at point-of-use.
So with this PR we remove the feature flag and adjust or remove related tests.

**Related issue(s)**:

Fixes #13954 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java
@@ -24,7 +24,6 @@ import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_AUTORENEW_ACCOU
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CONTRACT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_MAX_AUTO_ASSOCIATIONS;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.MODIFYING_IMMUTABLE_CONTRACT;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.NOT_SUPPORTED;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT;
 import static com.hedera.hapi.util.HapiUtils.EMPTY_KEY_LIST;
 import static com.hedera.node.app.hapi.utils.CommonPbjConverters.fromPbj;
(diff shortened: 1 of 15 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java
```diff
@@ -191,8 +188,6 @@ private void validateSemantics(
                 validateFalse(
                         entitiesConfig.limitTokenAssociations() && newMaxAssociations > tokensConfig.maxPerAccount(),
                         REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT);
```
lukelee-sl, line 190: Is there any possibility that `newMaxAssociations` is negative in this block because `entitiesConfig.unlimitedAutoAssociationsEnabled()` is false?  Or do the other validations in this block preclude this from happening?

stoyanov-st: We actually check for negative here: https://github.com/hashgraph/hedera-services/blob/ceb6020b1511031ab6fa9cdab9918396c05c0619/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java#L179

And test it with HapiTest
https://github.com/hashgraph/hedera-services/blob/ceb6020b1511031ab6fa9cdab9918396c05c0619/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/hapi/ContractUpdateSuite.java#L503

lukelee-sl: Right I saw that I was wondering what would happen if the first clause was false and the `newMaxAssociations < 0` is true?  Not a big deal this is really just a nit.

lukelee-sl (dismissed):
LGTM.  Just a small nit question.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java
stoyanov-st: I see what you meant. Thanks! I added additional check for negative.

lukelee-sl (approved):
LGTM.  tyvm.

tinker-michaelj (approved):
LGTM, ty @stoyanov-st !

Issue #13683, opened 2024-06-05
HIP-904: Implement TokenCancelAirdrop transaction


MiroslavGatsanoga:
### Problem
We want to provide the option for an airdrop sender to cancel an airdrop that have been sent previously by them. That would be useful if the airdrop sender has made a mistake or the receiver hasn't claimed the airdrop for too long. This would be a way to remove ongoing fees for auto renewal that would be charged from the airdrop sender account.

### Solution
We should define this new operation in our Hedera grpc service and all of the needed components for it.
More details and information about the transaction type can be found in [this design doc](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/services/token-service/airdrops/token-cancel-airdrop.md).

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-services/issues/13997
- [ ] https://github.com/hashgraph/hedera-services/issues/14027
- [ ] https://github.com/hashgraph/hedera-services/issues/14054
- [ ] https://github.com/hashgraph/hedera-services/issues/14093
- [ ] https://github.com/hashgraph/hedera-services/issues/13949
- [ ] https://github.com/hashgraph/hedera-services/issues/13684
```

```[tasklist]
### Definition of done
- [x] Add all needed protobuf changes to the protobuf repo
- [x] Update `TokenServiceDefinition` to include `TokenCancelAirdrop` grpc operation
- [x] Define `TokenCancelAirdropHandler` with the following logic - `preHandle()`, `handle()`, `pureChecks()` and all needed methods following the business logic described in the design doc
- [x] Provide unit tests for all new components
```

vtronkov:
All stories related to cancel airdrop are merged

PR #14115, opened 2024-07-02
feat: HIP-904 Implement `TokenCancelAirdrop` transaction

stoyanov-st:
**Description**:

With HIP-904 we introduce Token Airdrops.
We want to provide the option for an airdrop sender to cancel an airdrop that have been sent previously by them. That would be useful if the airdrop sender has made a mistake or the receiver hasn't claimed the airdrop for too long. This would be a way to remove ongoing fees for auto renewal that would be charged from the airdrop sender account.

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/13683

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
```diff
@@ -0,0 +1,145 @@
...
+            @NonNull final ReadableAccountStore accountStore)
+            throws PreCheckException {
+        final var senderAccount = accountStore.getAliasedAccountById(senderId);
+        validateTruePreCheck(senderAccount != null, INVALID_ACCOUNT_ID);
+
+        // If the sender account is immutable, then we throw an exception.
+        final var key = senderAccount.key();
+        if (key == null || !isValid(key)) {
+            throw new PreCheckException(ACCOUNT_IS_IMMUTABLE);
+        }
```
Neeharika-Sompalli, line 102: This will fail when we do `requireKey`. We don't need extra check here

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn, "Transaction body cannot be null");
+        final var op = txn.tokenCancelAirdropOrThrow();
+
+        validateFalsePreCheck(op.pendingAirdrops().isEmpty(), EMPTY_PENDING_AIRDROP_ID_LIST);
+        validateFalsePreCheck(
+                op.pendingAirdrops().size() > MAX_ALLOWED_PENDING_AIRDROPS_TO_CANCEL, MAX_PENDING_AIRDROP_ID_EXCEEDED);
```
Neeharika-Sompalli, line 113: Can this be `MAX_PENDING_AIRDROPS_EXCEEDED`

vtronkov:
Created a new PR branched from develop - https://github.com/hashgraph/hedera-services/pull/14379

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
vtronkov: Fixed in https://github.com/hashgraph/hedera-services/pull/14379

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
vtronkov: Now the cancel is separated from the airdrop and we don't have the `MAX_PENDING_AIRDROPS_EXCEEDED` constant.

Issue #14147, opened 2024-07-03
register empty gradle tasks at root (releaseAdhocCommit, releaseDevelopCommit, releaseDevelopDailySnapshot, releaseDevelopSnapshot, releasePrereleaseChannel)


ElijahLynn:
We are experiencing a failure deploying the https://github.com/hashgraph/hedera-services/tree/v0.52.0-alpha.3 tag on this job: https://github.com/hashgraph/hedera-services/actions/runs/9784030137 with the error:

```
FAILURE: Build failed with an exception.

* What went wrong:
Cannot locate tasks that match ':releasePrereleaseChannel' as task 'releasePrereleaseChannel' not found in root project 'hedera-services'.
```

This was refactored to a loop here https://github.com/hashgraph/hedera-services/pull/11731/files#diff-849b9f52420974149e79598fd5f00a2a7abd48b6aeb9fea279821598c3199876R89-R95. @nathanklick suggested that the reason it isn't working is that we we need to register empty tasks on the root project like so: https://github.com/hashgraph/hedera-services/blob/63641ffdab4ec13759b901a58682f95b64bd4651/gradle/plugins/src/main/kotlin/com.hedera.gradle.lifecycle.gradle.kts#L39

PRs:
* https://github.com/hashgraph/hedera-services/pull/14148 (develop)
* https://github.com/hashgraph/hedera-services/pull/14150 (release/0.52)

Slack Discussion: https://swirldslabs.slack.com/archives/C03DN7L1T1A/p1720037447403079

rbarkerSL:
This will probably be resolved by the hiero-gradle-conventions changes

andrewb1269hg:
@rbarkerSL do we think this has been resolved?

rbarkerSL:
@ElijahLynn is this still an issue for you?

ElijahLynn:
let's close, I don't have any current issues.

PR #14150, opened 2024-07-03
fix: [release/0.52] register empty gradle tasks at root (releaseAdhocCommit, releaseDevelopCommit, releaseDevelopDailySnapshot, releaseDevelopSnapshot, releasePrereleaseChannel)

ElijahLynn:
fixes: 
* https://github.com/hashgraph/hedera-services/issues/14147

does:
- **register missing gradle tasks**
- **remove releaseMavenCentral task, duplicate in \*.lifecycle.grade.kts**

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.platform-publish.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.platform-publish.gradle.kts
@@ -88,8 +88,19 @@ publishing.repositories {
 // Register one 'release*' task for each publishing repository
 publishing.repositories.all {
     val ucName = name.replaceFirstChar { it.titlecase() }
-    tasks.register("release$ucName") {
-        group = "release"
-        dependsOn(tasks.named("publishMavenPublicationTo${ucName}Repository"))
+    val taskName = "release$ucName"
+
+    // Register the task if it is not already registered
+    if (!tasks.names.contains(taskName)) {
+        tasks.register(taskName) {
+            group = "release"
+            dependsOn(tasks.named("publishMavenPublicationTo${ucName}Repository"))
+        }
+    } else {
+        // If the task is already registered, we just add the group & dependency
+        tasks.named(taskName) {
+            group = "release"
+            dependsOn(tasks.named("publishMavenPublicationTo${ucName}Repository"))
+        }
     }
 }
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
@@ -24,5 +24,3 @@ plugins {
 group = "com.hedera.hashgraph"
 
 javaPlatform { allowDependencies() }
-
-tasks.register("releaseMavenCentral")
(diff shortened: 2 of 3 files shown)


nathanklick (approved):

ElijahLynn:
> An exception occurred applying plugin request [id: 'com.hedera.gradle.platform-publish']
> Failed to apply plugin 'com.hedera.gradle.platform-publish'.
   > Cannot add task 'releasePrereleaseChannel' as a task with that name already exists.

ElijahLynn:
> > An exception occurred applying plugin request [id: 'com.hedera.gradle.platform-publish']
> > Failed to apply plugin 'com.hedera.gradle.platform-publish'.
> > Cannot add task 'releasePrereleaseChannel' as a task with that name already exists.

I just pushed a commit to register them if they are not registered yet, not sure that is the best way, but just trying to get it working for now.

gradle/plugins/src/main/kotlin/com.hedera.gradle.platform-publish.gradle.kts
```diff
@@ -88,8 +88,19 @@ publishing.repositories {
 // Register one 'release*' task for each publishing repository
 publishing.repositories.all {
     val ucName = name.replaceFirstChar { it.titlecase() }
-    tasks.register("release$ucName") {
-        group = "release"
-        dependsOn(tasks.named("publishMavenPublicationTo${ucName}Repository"))
+    val taskName = "release$ucName"
+
+    // Register the task if it is not already registered
```
ElijahLynn, line 93: Not actually sure this is needed, we may just need the conditional below to modify the existing empty task that was defined at the root, in the lifecycle file. 

I was getting the error: `Cannot add task 'releasePrereleaseChannel' as a task with that name already exists.` without these conditionals because of the empty task definitions in the lifecycle file. Do we have two execution paths here?

gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
```diff
@@ -24,5 +24,3 @@ plugins {
 group = "com.hedera.hashgraph"
 
 javaPlatform { allowDependencies() }
-
-tasks.register("releaseMavenCentral")
```
rbarkerSL, line 28: Why is this registration being removed here?

rbarkerSL (commented):
Questions in com.hedera.gradle.versions

gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
rbarkerSL: Only need to registry at root level.

rbarkerSL (approved):
LGTM

gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
ElijahLynn: We found it was being duplicated in com.hedera.gradle.lifecycle.gradle.kts

ElijahLynn:
static code failure says, but I can't restart it, says it isn't required though so maybe it is okay:

image
image

PR #14148, opened 2024-07-03
fix: [develop] register empty gradle tasks at root (releaseAdhocCommit, releaseDevelopCommit, releaseDevelopDailySnapshot, releaseDevelopSnapshot, releasePrereleaseChannel)

ElijahLynn:
fixes: 
* https://github.com/hashgraph/hedera-services/pull/14147

does:
- **register missing gradle tasks**
- **remove releaseMavenCentral task, duplicate in \*.lifecycle.grade.kts**

PR #14160, opened 2024-07-04
fix: filter publishing by deactivating tasks

jjohannes:
**Description**:
Filter publishing by deactivating tasks rather than changing task dependencies.

With this change, we can still call ALL tasks of a certain publishing profile. E.g, we can do `releaseMavenCentral` or `releaseDevelopCommit` (without leading `:`). This is less risky of breaking if tasks are called in different ways (with or without `:`). And, more importantly, it allows us to again call all tasks that publish to CGP repositories (not Nexus) and only exist in the _platform_ projects. This broke with #14015.

This partially reverts #14015 (solving the original issue differently) and completely reverts #14148

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


thomas-swirlds-labs (approved):
LGTM. Thanks @jjohannes

gradle/plugins/src/main/kotlin/com.hedera.gradle.versions.gradle.kts
```diff
@@ -24,3 +24,5 @@ plugins {
 group = "com.hedera.hashgraph"
 
 javaPlatform { allowDependencies() }
+
+tasks.register("releaseMavenCentral")
```
ElijahLynn, line 28: Ahh, so this was needed after all, even though it is also in *.lifecycle.grade.kts. GTK.

.github/workflows/node-zxc-build-release-artifact.yaml
```diff
@@ -841,7 +841,7 @@ jobs:
           NEXUS_PASSWORD: ${{ secrets.sdk-ossrh-password }}
         with:
           gradle-version: ${{ inputs.gradle-version }}
-          arguments: ":release${{ inputs.release-profile }} -PpublishingPackageGroup=com.swirlds --scan -PpublishSigningEnabled=true --no-configuration-cache"
+          arguments: "release${{ inputs.release-profile }} -PpublishingPackageGroup=com.swirlds --scan -PpublishSigningEnabled=true --no-configuration-cache"
```
ElijahLynn, line 844: For my own learning, what does the : do?

nathanklick (approved):

thomas-swirlds-labs (approved):

rbarkerSL (approved):

ElijahLynn:
I updated the description with a link to the other PR and added which branch they target. I think we need a PR to develop as well. 

> completely reverts https://github.com/hashgraph/hedera-services/pull/14148 (develop) and https://github.com/hashgraph/hedera-services/pull/14150 (release/0.52).

ElijahLynn (approved):

.github/workflows/node-zxc-build-release-artifact.yaml
jjohannes: `:releaseSomething` invokes the task `releaseSomething` **only**  in the root project and fails if it is not there. That's the problem we had before this PR for tasks that do not exist in the root.

`releaseSomething` invokes multiple tasks called `releaseSomething` – in all projects where it exists. It skips projects that do not have it (e.g. the root) without error. 

I have a  section in a video that explains this a bit more: https://www.youtube.com/watch?v=sOo0p4Gpjcc&list=PLWQK2ZdV4Yl2k2OmC_gsjDpdIBTN0qqkE

With the older solution, we did `:releaseSomething` and then controlled which tasks are started in addition through dependency from that task.
With the new solution, we allow `releaseSomething` again for all cases. It might lead to more tasks starting than we need, but will skip the ones we do not want via the deactivation logic.

Usually I prefer the `:` solution because it is exact. But in this case, it is complex to get it right with all the different publishing cases.

PR #14163, opened 2024-07-06
chore: externalize genesis system file creations in record stream

### LINKED ISSUE #13594: Initial system files should be externalized via synthetic `FileCreate` at network genesis
### Description

Operators for test environments are forced to manually upload system file contents to externalize the initial exchange rates, fee schedules, throttles, etc. to mirror nodes in the environment.

### Steps to reproduce

Start a network from genesis, observe mirror nodes don't have active fee schedules.

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #13594


hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/schemas/V0490FileSchema.java
```diff
@@ -128,36 +135,67 @@ public Set<StateDefinition> statesToCreate(@NonNull final Configuration config)
...
-            @NonNull final NetworkInfo networkInfo) {
+    public void createGenesisAddressBookAndNodeDetails(@NonNull final GenesisContext genesisContext) {
+        requireNonNull(genesisContext);
+        final var networkInfo = genesisContext.networkInfo();
+        final var filesConfig = genesisContext.configuration().getConfigData(FilesConfig.class);
+        final var bootstrapConfig = genesisContext.configuration().getConfigData(BootstrapConfig.class);
 
-        logger.debug("Creating genesis address book and node details files");
+        // Create the master key that will own both of these special files
+        final var masterKey = KeyList.newBuilder()
```
Neeharika-Sompalli, line 151: This is genesis Key right?

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/schemas/V0490FileSchema.java
```diff
@@ -128,36 +135,67 @@ public Set<StateDefinition> statesToCreate(@NonNull final Configuration config)
...
+        for (final var nodeInfo : networkInfo.addressBook()) {
+            nodeDetail.add(NodeAddress.newBuilder()
+                    .stake(nodeInfo.stake())
+                    .nodeAccountId(nodeInfo.accountId())
+                    .nodeId(nodeInfo.nodeId())
+                    .rsaPubKey(nodeInfo.hexEncodedPublicKey())
+                    // we really don't have grpc proxy name and port for now.Temporary values are set.
+                    // After Dynamic Address Book Phase 2 release, we will have the correct values. Then update here.
+                    .serviceEndpoint(ServiceEndpoint.newBuilder()
+                            .ipAddressV4(Bytes.wrap("1.0.0.0"))
```
Neeharika-Sompalli, line 180: Probably have a constant that explains this ? Need not be done in same PR

hedera-node/test-clients/build.gradle.kts
```diff
@@ -62,9 +62,19 @@ val ciCheckTagExpressions =
...
+val prCheckInitialPorts =
+    mapOf(
+        "hapiTestCrypto" to "30000",
+        "hapiTestToken" to "31000",
+        "hapiTestRestart" to "32000",
+        "hapiTestSmartContract" to "33000",
+        "hapiTestNDReconnect" to "34000",
+        "hapiTestTimeConsuming" to "35000",
+        "hapiTestMisc" to "36000"
+    )
```
Neeharika-Sompalli, line 74: Is this hardcoding each PR check to use different ports?

Neeharika-Sompalli (approved):
Everything makes sense to me. A lot of work 😅 Thanks @tinker-michaelj

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/schemas/V0490FileSchema.java
tinker-michaelj: Correct!

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/schemas/V0490FileSchema.java
tinker-michaelj: Need to ask DAB team myself. 🤔

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: Yes I can't explain why it can matter since we'd imagine each check already running in isolation...but this seems to reduce port `BindException`'s. 🤷

jsync-swirlds: Ports can be bound for a lot of reasons in a K8S container, including by other containers on the same host in some setups.
The safest option for ensuring ports are available is to tell Kubernetes, in the pod definition, that the software within the container uses that port.  Kubernetes will then ensure the container has that port available on its localhost address (which may be shared with other containers on the same host) and bind it to the external network interfaces, possibly shifted via NAT, if requested.

Example:
```yaml
    spec:
      containers:
      - name: HapiTestReconnect
        image: hapi-test-ci-image
        ports:
        - containerPort: 34000
```

We probably need to work with devops to ensure our CI definitions include correct container port declarations.

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileGetContentsHandler.java
```diff
@@ -124,15 +139,26 @@ public void validate(@NonNull final QueryContext context) throws PreCheckExcepti
...
     private @Nullable FileContents contentFile(
-            @NonNull final FileID fileID, @NonNull final ReadableFileStore fileStore) {
+            @NonNull final FileID fileID,
+            @NonNull final ReadableFileStore fileStore,
+            @NonNull final Configuration config) {
         final var meta = fileStore.getFileMetadata(fileID);
         if (meta == null) {
-            return null;
+            if (notGenesisCreation(fileID, config)) {
+                return null;
```
Neeharika-Sompalli, line 155: Why do we need this ? Will the `fieStore` not contain genesis creations ?

Neeharika-Sompalli: We would have committed all genesis changes to state right ?

Neeharika-Sompalli (dismissed):
LGTM other then few comments. Thanks @tinker-michaelj

rbarkerSL:
Looks good from the devops-ci perspective.

Neeharika-Sompalli (approved):

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileGetContentsHandler.java
tinker-michaelj: This is to answer queries before handling the genesis transaction.

hedera-node/test-clients/build.gradle.kts
tinker-michaelj: I suspect you are right @jsync-swirlds 💯

Neeharika-Sompalli (approved):

edward-swirldslabs (approved):

rbarkerSL (approved):

PR #14186, opened 2024-07-09
feat: enhance tests for evm equivalence

### LINKED ISSUE #14114: Enhance tests around evm equivalence.
### Problem

The current suite of tests cover the majority of test cases for validating evm equivalence.  There are however a few use cases that are missing.

### Solution

Add coverage for certain missing use cases for evm equivalence to round out the testing. 

### Alternatives

_No response_

lukelee-sl:
**Description**:
- Add several tests that cover several evm equivalence scenarios.
- Fix issue with deploying contracts less than 4096 bytes.
- Change the response code returned when trying to transfer value to a system account from `INVALID_FEE_SUBMITTED` to `INVALID_CONTRACT_ID`
- Add test to ensure that 0 byte contract code cannot be created
 
**Related issue(s)**:

Fixes #14114

**Checklist**


Nana-EC (dismissed):

david-bakin-sl (dismissed):
LGTM! (new-style BDD test is so much easier to read - certainly the organization thereof)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
```diff
@@ -173,8 +173,8 @@ private long effectiveGasUsed(final long gasLimit, @NonNull final MessageFrame f
     private void propagateHaltException(MessageFrame frame, ExceptionalHaltReason haltReason) {
         if (haltReason.equals(INSUFFICIENT_CHILD_RECORDS)) {
             setPropagatedCallFailure(frame, HevmPropagatedCallFailure.RESULT_CANNOT_BE_EXTERNALIZED);
-        } else if (haltReason.equals(INVALID_FEE_SUBMITTED)) {
```
david-bakin-sl, line 176: `INVALID_FEE_SUBMITTED` is still used - e.g. `ExchangeRateSystemContract` and `PrngSystemContract`.  Should it be removed from here?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -265,7 +265,7 @@ private void doTransferValueOrHalt(
     private void doHaltIfInvalidSystemCall(
```
david-bakin-sl, line 265: This method name seems too generic.  It's only called in one place on the specific condition that the address is invalid, and it only returns one specific status code.  Shouldn't the name reflect that?

david-bakin-sl (approved):
For some reason a couple of things caught my eye that didn't last time.

Regardless: LGTM. (IOW, accept comments or not, as you wish, as usual.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/FrameRunner.java
lukelee-sl: I imagine that this response code still needs to be propagated.

PR #14187, opened 2024-07-09
build: filter publishing by deactivating tasks (develop)

jjohannes:
...rather than changing task dependencies.

With this change, we can still call ALL tasks of a certain publishing profile. E.g, we can do 'releaseMavenCentral' or 'releaseDevelopCommit' (without leading ':'). This is less risky of breaking if Gradle is called wrongly. And, more importantly, it allows us to again call all tasks that publish to CGP repositories (not Nexus) and only exist in the _platform_ projects.

**Description**:

**Related issue(s)**:

Same as #14160 but for **develop**

**Notes for reviewer**:

**Checklist**


rbarkerSL (approved):
LGTM

PR #14193, opened 2024-07-09
fix: cherry-pick to 0.52 fix for charging correct gas for new isAuthorizedRaw system contract method

### LINKED ISSUE #14191: fix: cherry-pick to 0.52 fix for charging correct gas for new `isAuthorizedRaw` system contract method
`isAuthorizedRaw` in 0.52 does not have the fix from develop where _correct_ gas is charged.  Currently it charges 1.5M gas for both EC and ED signatures.  EC signatures should cost only 3000 gas.  This discrepancy is large and would affect users of this method in 0.52.

david-bakin-sl:
**Description**:

isAuthorizedRaw in 0.52 does not have the fix from develop where correct gas is charged. Currently it charges 1.5M gas for both EC and ED signatures. EC signatures should cost only 3000 gas. This discrepancy is large and would affect users of this method in 0.52.

**Related issue(s)**:

Fixes #14191

**Notes for reviewer**:

Cherry picks _two_ commits together - 162b700 and 0d3c042 - the second is the fix, but its test suite _depends_ on the first (which moved a couple of methods from production code to test-only code).

**Checklist**


lukelee-sl (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM!

Issue #14209, opened 2024-07-10
fix: `isAuthorizedRaw` should take all gas available when it fails due to bad input


david-bakin-sl:
`isAuthorizedRaw` should take all gas available when it fails due to bad input.  (Currently it takes only the gas it is specified to take for success.)

PR #14210, opened 2024-07-10
fix: `isAuthorizedRaw` to take all gas on bad input

david-bakin-sl:
**Description:**:

`isAuthorizedRaw`, a system contract method, should take all offered gas when it fails due to bad input.

(Note that give a valid account, message (hash), and signature it is _not_ a failure for the signature to not be valid for that message + account.  But if the account doesn't _exist_ that's bad input and that is a failure.)

**Related issue(s)**:

Fixes: #14209

**Checklist**


Nana-EC (approved):

lukelee-sl (approved):
LGTM

PR #14214, opened 2024-07-11
chore: Generate root hash on an input block

### LINKED ISSUE #14137: Implement algorithm for generating a hash on an input block
Implement the algorithm for hashing a block as described [here](https://www.notion.so/swirldslabs/Block-Signing-10acfed51cb04e64946d524fc19802db?pvs=4#805164b9018b435cabc3b49ab5898891). Return the root hash of the tree

thomas-swirlds-labs:
**Description**:

Initial implementation of the algorithm for hashing a block as described [here](https://www.notion.so/swirldslabs/Block-Signing-10acfed51cb04e64946d524fc19802db?pvs=4#805164b9018b435cabc3b49ab5898891). Returns the root hash of the Input/Output Merkle tree

**Related issue(s)**:

Fixes #14137 

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+package com.hedera.node.app.util;
+
+import com.hedera.pbj.runtime.Codec;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.util.ArrayList;
+import java.util.List;
+
+public class HashTreeManager<T> {
```
mhess-swl, line 26: Javadoc for this class and its methods would be especially helpful, I think

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+            Bytes elementHash = hash(codec.toBytes(element));
+            hashList.add(elementHash);
+
+            if ((elements.indexOf(element) & 1) == 1) {
+                Bytes yHash = hashList.remove(hashList.size() - 1);
+                Bytes xHash = hashList.remove(hashList.size() - 1);
+                Bytes combinedHash = hash(yHash.append(xHash));
+                hashList.add(combinedHash);
+            }
+        }
```
mhess-swl, line 55: This logic exists in various separate places in this class. Intentionally not consolidated?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    public Bytes calculateMerkleRootHash() {
```
mhess-swl, line 58: If we're going to keep this class generic, I suspect we don't want to use the word 'merkle' anywhere

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+        Bytes merkleRootHash = hashList.get(hashList.size() - 1);
+        for (int i = hashList.size() - 2; i >= 0; i--) {
+            merkleRootHash = hash(hashList.get(i).append(merkleRootHash));
+        }
+        return merkleRootHash;
+    }
+
+    public String calculateMerkleRootOnLeafNodes(List<String> elements) {
```
mhess-swl, line 66: `List<T> elements`?
i'm really confused by this method. Why is it needed, and what is it doing?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+        List<String> hashList = new ArrayList<>();
+
+        // Initial hashing of elements
+        for (String element : elements) {
+            System.out.println(
+                    "Writing element " + element + " to the stream, with Hash: " + sha384(element.toString()));
+            hashList.add(sha384(element).toString());
```
mhess-swl, line 73: why are we adding a `.toString()` to the hash list here?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/utils/HashTreeManagerTest.java
```diff
@@ -0,0 +1,196 @@
...
+                String hash2 = hashTreeManager.sha384("input" + j).toString();
+                if (hash1.equals(hash2)) {
+                    foundCollision = true;
+                    break;
+                }
+            }
+            if (foundCollision) break;
+        }
+        assertThat(foundCollision).isTrue().withFailMessage("A collision was found.");
+    }
```
mhess-swl, line 119: This feels like testing the sha384 algorithm, which seems unnecessary. These two tests also require the `sha394()` method to be part of `HashTreeManager`'s public API, which also seems unnecessary. What benefits do these tests provide? We're already assuming sha384 won't collide, otherwise we'd use a different algorithm. Maybe I'm missing something?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,183 @@
...
+    private List<Bytes> hashList = new ArrayList<>();
+
+    private enum Alternator {
+        LEFT,
+        RIGHT
+    }
+
+    Alternator currentAlternate = Alternator.LEFT;
+
+    public void processNodes(List<T> elements) {
```
derektriley, line 44: Perhaps we should have another method which accepts as an argument a single element since I am not sure how we will be calling this (one block item at a time most likely).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
thomas-swirlds-labs: I've reworked the class to make it more generic and renamed methods to reflect the generic nature 👍

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
thomas-swirlds-labs: Refactored now 👍

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
thomas-swirlds-labs: Agreed and added. Thanks @derektriley

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
thomas-swirlds-labs: Removed as redundant now

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
```diff
@@ -0,0 +1,168 @@
...
+                new OutputMerkleTreeData(outputRoot, transactionOutput, transactionResult, stateChanges);
+        List<OutputMerkleTreeData> outputMerkleTreeDataList = new ArrayList<>();
+        outputMerkleTreeDataList.add(outputMerkleTreeData);
+
+        var rootHash = tree.getTreeRootAsString();
+        return rootHash;
+    }
+}
+
+class SimpleStringCodec implements Codec<String> {
```
mhess-swl, line 141: This class looks like a bunch of stubbed methods. What does it actually do?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        assertThat(tree.getTreeRoot()).isNotNull(); // A tree with three leaves still computes a root hash
+
+        tree.addElement("leaf4");
+        String newRootHash = tree.getTreeRootAsString();
+        // Root hash should change as new leaves are added
+        assertThat(rootHash).as("Checking that hashes are not equal").isNotEqualTo(newRootHash);
+    }
+
+    @Test
+    public void testHashBlockItem() throws NoSuchAlgorithmException {
```
mhess-swl, line 90: It may be useful to test the block item-related stuff in a separate class. Or maybe not. Just an idea, not a requirement

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        // A Merkle tree with a single leaf does have a root hash.
+        assertThat(tree.getTreeRoot()).isNotNull();
+
+        tree.addElement("leaf2");
+        String rootHash = tree.getTreeRootAsString();
+        assertThat(rootHash)
+                .isEqualTo(tree.getTreeRootAsString()); // Root hash should remain consistent after balancing
+
+        tree.addElement("leaf3");
+        assertThat(tree.getTreeRoot()).isNotNull(); // A tree with three leaves still computes a root hash
```
mhess-swl, line 81: I like the idea of this test, but wonder if we need more in the tree's public API in order to verify correct ordering. As it stands, really we're just testing that there's a root node. We're not checking that the hash is what we expect it to be, which means we could be adding them in any order and this test would only verify that `getTreeRootAsString()` returns not null. (Especially since the implementation isn't a tree at all, just an ordered collection of hashes!) The ordering is crucial if a block node wants to verify, so I think the order should be tested. 

Part of this ordering is testing the exact value of the hashes instead of just checking for non-null; the hashes are computed by concatenating child hashes together, so if they're in the wrong order or the tree isn't balanced correctly, the hash will be different from the expected. 

The other part of ordering is perhaps verifying the hashes of the "intermediate" parent nodes, though I suppose one could argue that those will also be reflected in the hash values if they're incorrect. Thoughts?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
mhess-swl: Definitely. I dunno that we'll ever want to pass in a collection at all, since we want to be hashing elements as quickly as possible, independent of the handle workflow. Passing in a collection with size > 1 means that at least one element is sitting in that list waiting for the construction of a separate element, which seems to violate (even if only slightly) the goal of hashing as soon as the block item becomes available.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
thomas-swirlds-labs: It's only used for testing purposes by supplying a most basic Codec

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
thomas-swirlds-labs: Makes sense. Have removed the addElements method as well to reflect 👍

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,125 @@
...
+    private final MessageDigest digest; // The message digest
+
+    /**
+     * Constructor for the HashTreeManager class.
+     * @param codec The codec used to encode and decode the elements in the tree
+     * @throws NoSuchAlgorithmException If the hashing algorithm is not found
+     */
+    public HashTreeManager(Codec<T> codec) throws NoSuchAlgorithmException {
+        this.codec = codec;
+        this.digest = MessageDigest.getInstance(HASHING_ALGO);
```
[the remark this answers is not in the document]
thomas-swirlds-labs: Agreed and added

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
thomas-swirlds-labs: Yep we can definitely look at that in a follow up PR

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/OutputMerkleTreeData.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class OutputMerkleTreeData {
+    private MerkleNode outputRoot;
+    private TransactionOutput transactionOutput;
+    private TransactionResult transactionResult;
+    private StateChanges stateChanges;
+
+    // Constructor
+    public OutputMerkleTreeData(
+            MerkleNode outputRoot,
```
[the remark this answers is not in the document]
thomas-swirlds-labs: Yeah, strictly speaking we probably don't need this utility class at all any more. Have removed ✅

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
mhess-swl: Ah, there's a default implementation in `Codec` 👍

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
thomas-swirlds-labs: I've added a new method `getHashList() `to `HashTreeManager` and used that to get and verify the hashes in the test at each point after adding leaves to the tree. 

This gives us coverage to validate the correctness of the ordering without exposing too much of the tree's implementation. 

As discussed in slack we may want to revisit once we have integrated more into `HandleWorkflow` however and add the ability to verify for example by passing in a path to the node and receiving the subroot/node, e.g. `root -> left -> right -> left -> left`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/util/HashTreeManager.java
```diff
@@ -0,0 +1,114 @@
...
+                        hash(x.append(y), digest)); // For each pair of nodes, appends y to x and hashes the result.
+            }
+            hashList = newHashList;
+        }
+    }
+
+    /** Get the list of hashes in the tree
+     * @return The list of hashes in the tree
+     */
+    public List<Bytes> getHashList() {
```
mhess-swl, line 78: Nit: `@VisibleForTesting`

hedera-node/hedera-app/src/test/java/com/hedera/node/app/util/HashTreeManagerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        actualOrder.add(tree.getHashList().toString());
+        String newRootHash = tree.getTreeRootAsString();
+        assertThat(newRootHash).isNotNull();
+        assertThat(newRootHash).isNotEqualTo(rootHashAfterAddingLeaf3); // Should change
+
+        List<String> expectedOrder = Arrays.asList(
+                "[38b060a751ac96384cd9327eb1b1e36a21fdb71114be07434c0cc7bf63f6e1da274edebfe76f65fbd51ad2f14898b95b]",
+                "[82a82763156fb2d6421ef934dd593921ebfebeee2b6626494715dab6664a5a4f1b64c7b83de6e9c93ae5199d898fc0b6]",
+                "[347180f5e69054571b00bea816e5937b166dbcb51983882f76af907c6802956294f1b5ea1b317bba15fd14b6d0c73a1f]",
+                "[550df2684cb630cbaed45d58fa7b538d0eb099dc31b3164f8468046da8ff03620b593b7058b82186eae060eab292feb9]");
```
mhess-swl, line 101: Nit: it would be nice to have a comment specifying either a visualization of the tree, or comments by each hash signifying which node it represents

mhess-swl (approved):
Not a ton of code, but definitely a ton of work. Well done, thanks @thomas-swirlds-labs! 🙌

Issue #11771, opened 2024-02-27
Refactor `PlatformState` into a singleton


imalygin:
Currently, `PlatformState` is an instance of `MerkleLeaf` and it deviates from the idea of having only three types in the Merkle tree (KV storage, queue, and singleton). It should be refactored to a singleton.

When this ticket is complete, `PlatformState` will be migrated to the protobuf format and accessed through the State API only.

`PlatformState` contains several fields to be deleted:

`epochHash` - we no longer support emergency reconnect
`nextEpochHash` - same
`uptimeData` - currently not utilized for anything. This is a part of a feature that was not fully implemented, and it's implementation should be reconsidered separately

The unit tests associated with this change are NOT part of this ticket. This ticket will NOT result in a mergable PR. Criteria of Done for this ticket is getting informal signoffs on the draft PR so that unit tests can be updated (another ticket) without rework.

imalygin:
Blocked by #13084

imalygin:
Blocked by #14220 
Blocked by #14230

imalygin:
Blocked by https://github.com/hashgraph/hedera-services/issues/11773

imalygin:
It makes sense to implement https://github.com/hashgraph/hedera-services/issues/11773 because it will be much easier to migrate from `PlatformState` to a singleton once it's done. 

For example, there is `buildGenesisState` method:

```
    public static ReservedSignedState buildGenesisState(
            @NonNull final PlatformContext platformContext,
            @NonNull final AddressBook addressBook,
            @NonNull final SoftwareVersion appVersion,
            @NonNull final MerkleRoot stateRoot) 
```

This is where we create initial platform state. Currently, `MerkleRoot` is implemented by two classes - `State` (current merkle root) and `MerkleHederaState` (to become a merkle root once #11773 is complete). If `MerkleHederaState` was a root, the migration would be more natural as `MerkleHederaState` implements `HederaState` interface which is a main interface of State API

imalygin:
Blocked by #11773

PR #14231, opened 2024-07-12
refactor: 14230 Removed `uptimeData` feature.

### LINKED ISSUE #14230: Remove `uptimeData` from `PlatformState`
We're not planning to store `uptimeData` in the `PlatformState`. Therefore, the code should be removed.

While we do plan on using uptime data in the future to more accurately calculate staking rewards for node operators, we are not certain of the format yet and therefore will remove it before creating the protobuf definition for the Platform State. It will be added back later when we are certain what it will contain.

imalygin:
**Description**:

This PR removes a set of classes related to `Update data` feature as it's used and shouldn't be a part of the platform state.

This PR is required to reduce the scope of https://github.com/hashgraph/hedera-services/issues/11771

**Related issue(s)**:

Fixes #14230 

**Notes for reviewer**:

**Checklist**


jeromy-cannon (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
```diff
@@ -127,10 +144,26 @@ public void setInitialState(@NonNull final MerkleRoot state) {
     public void handleConsensusRound(final ConsensusRound round) {
         final MerkleRoot state = stateRef.get();
 
-        uptimeTracker.handleRound(
-                round,
-                state.getPlatformState().getUptimeData(),
-                state.getPlatformState().getAddressBook());
+        if (!round.isEmpty()) {
```
litt3, line 147: I don't think this should be in the `SwirldStateManager`. Instead, I think this actually should consume self events filtered from the consensus events coming out of the `ConsensusEngine`. Unfortunately, making this change would require a design proposal and thus probably won't be permitted at this time.

What do you think about this interim step instead?

- Bundle this functionality into a java class named something like `SelfEventMonitor`
    - There can also be some simplifications made to this code, since we don't actually need the most recent time any events came to consensus apart from self events. So the hash map isn't needed anymore
- Leave the call to `SelfEventMonitor` here in the `SwirldStateManager` for now
- Maybe put a `future work` comment, describing where this functionality ought to live

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
```diff
@@ -89,14 +107,13 @@ public SwirldStateManager(
 
         Objects.requireNonNull(platformContext);
         Objects.requireNonNull(addressBook);
-        Objects.requireNonNull(selfId);
+        this.selfId = Objects.requireNonNull(selfId);
         this.stats = new SwirldStateMetrics(platformContext.getMetrics());
-        Objects.requireNonNull(statusActionSubmitter);
+        this.statusActionSubmitter = Objects.requireNonNull(statusActionSubmitter);
         this.softwareVersion = Objects.requireNonNull(softwareVersion);
+        this.time = Time.getCurrent();
```
litt3, line 114: We shouldn't use `Time.getCurrent()` in components - instead you can get the time object stored inside of `PlaformContext`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
imalygin: I limited the scope of the change to removal `uptimeData` from the `PlatformState`. Now `UptimeTracker` has transient state of uptime. That is, it's in-memory and it gets lost with the node restart.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformState.java
```diff
@@ -224,12 +222,26 @@
         }
     }
 
+    private void skipUptimeData(final SerializableDataInputStream in) throws IOException {
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeTracker.java
```diff
@@ -212,7 +218,7 @@ private void scanRound(
      */
     private void updateState(
```
litt3, line 219: the method name `updateState` is no longer suitable

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeTracker.java
```diff
@@ -62,6 +63,11 @@ public class UptimeTracker {
      */
     private final StatusActionSubmitter statusActionSubmitter;
 
+    /**
+     * The uptime data for all the nodes.
+     */
+    final UptimeData uptimeData;
```
litt3, line 69: member should be `private`

litt3 (commented):
I confirmed that the uptime metrics look good in the PR JRS run

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/uptime/UptimeTracker.java
imalygin: It's package-private because it's used in the test. Would you prefer package-private getter?

imalygin: I added a comment to javadoc about it

litt3 (approved):

PR #14232, opened 2024-07-12
chore: add proposed `IsAssociatedSystemContractTest` test refactor

tinker-michaelj:
**Description**:
 - Extends `SpecContract` and `SpecToken` DSL to support a cleaner implementation of `IsAssociatedSystemContractTest`, see [here](https://github.com/hashgraph/hedera-services/blob/hapi-test-refinements/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/IsAssociatedSystemContractTest.java#L49).
    * In particular, extends `SpecToken` object to support directly "calling" a token redirect contract, see [here](https://github.com/hashgraph/hedera-services/blob/hapi-test-refinements/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/IsAssociatedSystemContractTest.java#L92).
- Fixes `ContractUpdateSuite#updateMaxAutomaticAssociationsAndRequireKey()` after standardization of max entity lifetime.
-Adds convenience support for a `ContractID{evm_address=0x00...0T}` reference to a token contract (note the added `@HapiTest`s cover this).


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/dsl/contracts/TokenRedirectContract.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.services.bdd.spec.dsl.contracts;
+
+/**
+ * Enumerates the contracts that a Hedera native token "implements" by redirecting to a well-known
+ * system contract address.
+ */
+public enum TokenRedirectContract {
```
lukelee-sl, line 23: I guess we can add the `AccountRedirectContract` enum when it is needed later.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/dsl/operations/transactions/AssociateTokensOperation.java
```diff
@@ -32,7 +35,12 @@
  */
 public class AssociateTokensOperation extends AbstractSpecTransaction<AssociateTokensOperation, HapiTokenAssociate>
         implements SpecOperation {
+    @Nullable
     private final SpecAccount account;
+
+    @Nullable
+    private final SpecContract contract;
```
lukelee-sl, line 42: Is the idea that we will be adding operations like this in the future to keep the tests as clean as possible?  For example, I need an `approveAllowance` operation for the current set of tests that I am writing.  Hopefully the set of operations will remain somewhat manageable.  I really like the way the ultimate test turned out.

lukelee-sl (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM from test-clients! Will defer to @lukelee-sl  on smart contracts changes

PR #14233, opened 2024-07-13
chore:Disallow backpressure for direct schedulers

### LINKED ISSUE #10870: Disallow backpressure for direct schedulers

HarshSawarkar:
**Description**:
Added logic in TaskSchedulerBuilder.withType(), TaskSchedulerBuilder.withUnhandledTaskCapacity(), and TaskSchedulerBuilder.build() to disallow backpressure for direct schedulers.

**Related issue(s)**:

Fixes #10870 

**Notes for reviewer**:
The condition throws an IllegalArgumentException when the type is DIRECT or DIRECT_THREADSAFE and if unhandledTaskCapacity is set.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/deterministic/DeterministicTaskSchedulerBuilder.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/deterministic/DeterministicTaskSchedulerBuilder.java
@@ -16,6 +16,7 @@
 
 package com.swirlds.common.wiring.model.internal.deterministic;
 
+import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.DIRECT;
 import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.DIRECT_THREADSAFE;
 import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.NO_OP;
 
@@ -66,6 +67,10 @@ public class DeterministicTaskSchedulerBuilder<OUT> extends AbstractTaskSchedule
     @NonNull
     @Override
     public TaskScheduler<OUT> build() {
+        // Check to ensure unhandled task capacity is not set for direct schedulers
+        if ((type == DIRECT || type == DIRECT_THREADSAFE) && unhandledTaskCapacity != 1) {
+            throw new IllegalArgumentException("Direct schedulers cannot have an unhandled task capacity.");
+        }
 
         final boolean insertionIsBlocking = unhandledTaskCapacity != UNLIMITED_CAPACITY || externalBackPressure;
 
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
@@ -136,6 +136,9 @@ public abstract class AbstractTaskSchedulerBuilder<OUT> implements TaskScheduler
     @Override
     @NonNull
     public AbstractTaskSchedulerBuilder<OUT> withType(@NonNull final TaskSchedulerType type) {
+        if ((type == DIRECT || type == DIRECT_THREADSAFE) && unhandledTaskCapacity != 1) {
+            throw new IllegalArgumentException("Direct schedulers cannot have an unhandled task capacity.");
+        }
         this.type = Objects.requireNonNull(type);
         return this;
     }
@@ -146,6 +149,9 @@ public abstract class AbstractTaskSchedulerBuilder<OUT> implements TaskScheduler
     @Override
     @NonNull
     public AbstractTaskSchedulerBuilder<OUT> withUnhandledTaskCapacity(final long unhandledTaskCapacity) {
+        if (type == DIRECT || type == DIRECT_THREADSAFE) {
+            throw new IllegalArgumentException("Cannot set unhandled task capacity for direct schedulers.");
+        }
         this.unhandledTaskCapacity = unhandledTaskCapacity;
         return this;
     }
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/StandardTaskSchedulerBuilder.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/StandardTaskSchedulerBuilder.java
@@ -16,6 +16,7 @@
 
 package com.swirlds.common.wiring.schedulers.builders.internal;
 
+import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.DIRECT;
 import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.DIRECT_THREADSAFE;
 import static com.swirlds.common.wiring.schedulers.builders.TaskSchedulerType.NO_OP;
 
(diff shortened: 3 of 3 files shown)


hendrikebbers:
Hey @HarshSawarkar can you add some basic unit tests to check that funcionality of the builder please.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/model/internal/deterministic/DeterministicTaskSchedulerBuilder.java
```diff
@@ -66,6 +67,10 @@ public DeterministicTaskSchedulerBuilder(
     @NonNull
     @Override
     public TaskScheduler<OUT> build() {
+        // Check to ensure unhandled task capacity is not set for direct schedulers
+        if ((type == DIRECT || type == DIRECT_THREADSAFE) && unhandledTaskCapacity != 1) {
```
lpetrovic05, line 71: this check is repeated, it makes sense to put it in a static method

HarshSawarkar: Fixed it.

HarshSawarkar:
> Hey @HarshSawarkar can you add some basic unit tests to check that funcionality of the builder please.

Updated.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
```diff
@@ -146,6 +164,9 @@ public AbstractTaskSchedulerBuilder<OUT> withType(@NonNull final TaskSchedulerTy
     @Override
     @NonNull
     public AbstractTaskSchedulerBuilder<OUT> withUnhandledTaskCapacity(final long unhandledTaskCapacity) {
+        if (type == DIRECT || type == DIRECT_THREADSAFE) {
```
lpetrovic05, line 167: I think it makes sense to call the method here as well, just to avoid duplication

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
```diff
@@ -130,12 +130,30 @@ public AbstractTaskSchedulerBuilder<OUT> configure(@NonNull final TaskSchedulerC
...
+     * Ensures that direct schedulers do not have an unhandled task capacity set.
+     *
+     * <p>If the scheduler type is {@link TaskSchedulerType#DIRECT} or {@link TaskSchedulerType#DIRECT_THREADSAFE}
+     * and the unhandled task capacity is not 1, an {@link IllegalArgumentException} is thrown.
+     *
+     * @param type the type of the task scheduler
+     * @param unhandledTaskCapacity the unhandled task capacity
+     * @throws IllegalArgumentException if the type is direct or direct threadsafe and the unhandled task capacity is not 1
+     */
+    public static void validateDirectScheduler(TaskSchedulerType type, long unhandledTaskCapacity) {
```
lpetrovic05, line 143: generally, the convention we follow is to add `final` and `@NonNull`/`@Nullable` to parameters

HarshSawarkar: Updated

hendrikebbers (dismissed):

lpetrovic05 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
```diff
@@ -130,12 +130,31 @@ public AbstractTaskSchedulerBuilder<OUT> configure(@NonNull final TaskSchedulerC
...
+        }
+    }
+
     /**
      * {@inheritDoc}
      */
     @Override
     @NonNull
     public AbstractTaskSchedulerBuilder<OUT> withType(@NonNull final TaskSchedulerType type) {
+        validateDirectScheduler(type, this.unhandledTaskCapacity);
```
cody-littley, line 157: I'd suggest you only do this check when the scheduler is built. There could be some weirdness where it is possible to get an error while entering a legal combination of configurations but in the wrong order.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
```diff
@@ -146,6 +165,7 @@ public AbstractTaskSchedulerBuilder<OUT> withType(@NonNull final TaskSchedulerTy
     @Override
     @NonNull
     public AbstractTaskSchedulerBuilder<OUT> withUnhandledTaskCapacity(final long unhandledTaskCapacity) {
+        validateDirectScheduler(this.type, unhandledTaskCapacity);
```
cody-littley, line 168: I'd suggest you only do this check when the scheduler is built. There could be some weirdness where it is possible to get an error while entering a legal combination of configurations but in the wrong order.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/wiring/schedulers/builders/internal/AbstractTaskSchedulerBuilder.java
```diff
@@ -130,12 +130,31 @@ public AbstractTaskSchedulerBuilder<OUT> configure(@NonNull final TaskSchedulerC
...
+     * @param unhandledTaskCapacity the unhandled task capacity
+     * @throws IllegalArgumentException if the type is direct or direct threadsafe and the unhandled task capacity is not 1
+     */
+    public static void validateDirectScheduler(
+            @NonNull final TaskSchedulerType type, final long unhandledTaskCapacity) {
+        if ((type == TaskSchedulerType.DIRECT || type == TaskSchedulerType.DIRECT_THREADSAFE)
+                && unhandledTaskCapacity != 1) {
+            throw new IllegalArgumentException("Direct schedulers cannot have an unhandled task capacity.");
+        }
+    }
```
cody-littley, line 149: It seems a bit odd to make this method public. Outside potential unit tests, I don't think there will ever be use cases for somebody to call into this directly.

One way to do this differently might be to make this non-static and private. This would allow you to get rid of the arguments. If you do this, I'd rename it to something like `validateConfiguration()`, since it would be a decent place to do other checks if they ever get added.

HarshSawarkar: The method **validateDirectScheduler()** is called here: 
image
and here: 
image
Making it private would render it inaccessible in this build logic. Should this logic be inlined within the build() method for the classes mentioned above? Would this be a correct implementation?

HarshSawarkar: > It seems a bit odd to make this method public. Outside potential unit tests, I don't think there will ever be use cases for somebody to call into this directly.
> 
> One way to do this differently might be to make this non-static and private. This would allow you to get rid of the arguments. If you do this, I'd rename it to something like `validateConfiguration()`, since it would be a decent place to do other checks if they ever get added.

The method **validateDirectScheduler()** is called here: 
image
and here: 
image
Making it private would render it inaccessible in this build logic. Should this logic be inlined within the build() method for the classes mentioned above? Would this be a correct implementation?

cody-littley: Perhaps this could be a private static method that lives inside `StandardTaskSchedulerBuilder.java`.

HarshSawarkar: > Perhaps this could be a private static method that lives inside `StandardTaskSchedulerBuilder.java`.

Making the method static would require the variables **type** and **unhandledTaskCapacity** in **_AbstractTaskSchedulerBuilder_** to be made static. Is this feasible?

poulok:
Hi @HarshSawarkar, I came across this PR while doing some cleanup. I apologize that we let this one slip through the cracks. Are you interested in updating the PR and moving toward merging? If not, I would like to close it.

PR #14243, opened 2024-07-15
test: add tests to verify calculateFees() stability in smart-contracts handlers

### LINKED ISSUE #13681: Harden `calculateFees()` in SmartContractService
Please revisit all implementations of TransactionHandler.calculateFees() in the SmartContractService and ensure they handle invalid input data correctly. We decided that we will simply fail fast for now if an error occurs while calculating the fees, i.e., we throw a PreCheckException.

(Once all handers have been hardened this way, we can remove the pureChecks() call from the ingest workflow.)

MrValioBg:
**Description**:
This PR adds tests to verify calculateFees() stability in smart-contracts handlers.

**Related issue(s)**:

Fixes #13681 

**Notes for reviewer**:
> [!IMPORTANT]  
> Please read this [comment](https://github.com/hashgraph/hedera-services/pull/14243#issuecomment-2230244346)

**Checklist**


MrValioBg:
The unit tests that I've added are for all the handlers in the smart contract service module that had the implementation of the calculateFees() method.

All of the unit tests verify that if the body for the specific transaction type is missing, this won't result in a `NullPointerException`, thus verifying that the calculateFees() methods are not using any part of those bodies for calculation.

The exception is only for the `ContractUpdateHandler`, where we actually use some fields of the specific body for the transaction type, i.e., `ContractUpdateTransactionBody`. There, we will throw a NullPointerException if the `ContractUpdateTransactionBody` is null. However, we have a test verifying that if the contractID is null, an unexpected `NullPointerException` won't be thrown. This is because, in its [calculateFees() method](https://github.com/hashgraph/hedera-services/blob/b53c465315b7ab8b3c0a292c5adc243abb01b43c/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/ContractUpdateHandler.java#L317), we are properly handling the null case.

I've not added another "masking" exception for the `NullPointerException` that could be thrown, as if it was null and it happened to reach the pureChecks, it would also throw `NullPointerException`, so it won't be an "unexpected exception", thus it does not need "hardening."

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/fee/SmartContractFeeBuilderTest.java
```diff
@@ -50,4 +52,15 @@ void assertGetFileContentQueryFeeMatrices() {
         assertEquals(0, result.getServicedata().getSbh());
         assertEquals(0, result.getServicedata().getTv());
     }
+
+    @Test
```
david-bakin-sl, line 56: 1. Why not test the two query ones as well (`getContractCallLocalFeeMatrices`, `getContractByteCodeQueryFeeMatrices`)?

1. Needs a comment explaining why these `xxxFeeMatrices` methods need this specific test, it's not really obvious.

david-bakin-sl (commented):
I have a question below plus why don't the read-only/query handlers also need this protection?  I know they don't have a `pureChecks` but can't it come up during `validate` or somewhere else?

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/fee/SmartContractFeeBuilderTest.java
MrValioBg: 2. I've added the clarifying comment

1. I haven't done tests for them because the issue describes that we should "harden" only calculateFees() methods. (However, I could add some tests if you think it would be useful.)

I think that the removal of pureChecks from IngestFlow won't affect the queries and computeFees() methods, so it's "hardening" the computeFees() is probably not related to the issue. Maybe @netopyr could give some clarification about this.

david-bakin-sl (approved):
Generally LGTM though I'm still interested in why the query fee matrices methods don't need to be tested (question from previously).

Neeharika-Sompalli (approved):
LGTM from `hedera-base`. Will defer to David if his comment on test is resolved.

PR #14250, opened 2024-07-15
feat(reconnect): log ReconnectMapStats

### LINKED ISSUE #13749: Update ReconnectNodeCount interface to log the new metrics
The existing `ReconnectNodeCount` interface logs the old reconnect metrics. It needs to be updated to log the new metrics tracked in `ReconnectMapStats`. We want them in the logs as well as the metrics so that the information is available when running locally or in unit tests.

anthony-swirldslabs:
**Description**:
Log the stats collected in `ReconnectMapStats`.

While an implementation of the `ReconnectMapStats` interface, such as the `ReconnectMapMetrics`, emits actual metrics already, it is still useful to also log the accumulated stat values once the reconnect is finished. This simplifies analyzing the metric values in a situation when the metrics are not published to Grafana directly, for example, when running a test on a developer branch locally or in a custom test network.

**Related issue(s)**:

Fixes #13749

**Notes for reviewer**:
Trivial changes simply adding extra logging.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
@@ -29,6 +29,10 @@ public class ReconnectMapMetrics implements ReconnectMapStats {
 
     private static final String RECONNECT_MAP_CATEGORY = "reconnect_vmap";
 
+    /** A map label as passed to the constructor, w/o any normalization. */
+    @Nullable
+    private final String originalLabel;
+
     private final ReconnectMapStats aggregateStats;
 
     private final Counter transfersFromTeacher;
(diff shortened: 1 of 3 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
```diff
@@ -29,6 +29,10 @@ public class ReconnectMapMetrics implements ReconnectMapStats {
 
     private static final String RECONNECT_MAP_CATEGORY = "reconnect_vmap";
 
+    /** A map label as passed to the constructor, w/o any normalization. */
```
kfa-aguda, line 32: curious - what does `normalization` mean here? If you have any round of changes to add, do you mind explaining it? Otherwise it's a super minor thing so it can be added in a later PR. Changes look good to me

kfa-aguda (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
anthony-swirldslabs: This refers to https://github.com/hashgraph/hedera-services/blob/245a322d4b856075d96fee6957045707dec262a3/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java#L66-L67

just below in this same class.

PR #14256, opened 2024-07-16
feat: use platform event post consensus

### LINKED ISSUE #13864: Use only the PlatformEvent class post-consensus
Atm, `ConsensusRound` holds a list of consensus events which are instances of `EventImpl`. Due to the change in the event structure, `ConsensusRound` should return a list of `PlatformEvent`s.

### LINKED ISSUE #13999: Provide system transactions to app
Atm, the `Transaction` interface exposes system transactions to apps. But at runtime, we filter out system transactions and never provide them to the app. This should be changed so that the app always receives all transactions.

lpetrovic05:
closes #13864
closes #13999


jeromy-cannon (approved):

rbarkerSL (approved):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/uptime/UptimeTests.java
```diff
@@ -82,31 +81,31 @@ private static List<EventImpl> generateEvents(
                     .setCreatorId(nodeId)
                     .setConsensusTimestamp(time.now())
                     .build();
-            final EventImpl event = new EventImpl(platformEvent, null, null);
 
             if (!noJudges.contains(nodeId) && firstEventCreated.add(nodeId)) {
-                event.setFamous(true);
-                event.setJudgeTrue();
+                // event.setFamous(true);
```
litt3, line 86: were these commented lines intentionally left here?

litt3 (approved):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/uptime/UptimeTests.java
lpetrovic05: not intentionally. but there is a PR open that removes this file: https://github.com/hashgraph/hedera-services/pull/14231
to avoid fixing something that is about to be deleted and needing reviews again, I will just leave this as is.

PR #14267, opened 2024-07-16
test: add tests that shows the difference of transferToken and transferFrom from contracts

### LINKED ISSUE #6220: Validate HTS transferToken and transferFrom functionality
### Problem

We are missing bdd tests that exercise the transferToken and transferFrom functionality for HTS tokens from contracts where the contract owns the tokens.

### Solution

Write tests to fill this gap in coverage

### Alternatives

_No response_

lukelee-sl:
**Description**:

**Related issue(s)**:

Fixes #6220 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/dsl/operations/transactions/TransferTokenOperation.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/dsl/operations/transactions/TransferTokenOperation.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.services.bdd.spec.dsl.operations.transactions;
+
+import static com.hedera.services.bdd.spec.transactions.TxnVerbs.cryptoTransfer;
+import static com.hedera.services.bdd.spec.transactions.token.TokenMovement.moving;
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.services.bdd.spec.HapiSpec;
+import com.hedera.services.bdd.spec.SpecOperation;
+import com.hedera.services.bdd.spec.dsl.entities.SpecAccount;
+import com.hedera.services.bdd.spec.dsl.entities.SpecContract;
+import com.hedera.services.bdd.spec.dsl.entities.SpecToken;
+import com.hedera.services.bdd.spec.transactions.token.HapiTokenAssociate;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+
+/**
+ * Associates an account with one or more tokens.
+ */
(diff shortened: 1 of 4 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/token/TransferTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+                // Transfer fungible tokens to the TokenTransferContract contract
+                cryptoTransfer(moving(20, FUNGIBLE_TOKEN).between(ACCOUNT, TOKEN_TRANSFER_CONTRACT)));
+    }
+
+    @Nested
+    @DisplayName("successful when")
+    @Order(1)
+    class SuccessfulTransferTokenTest {
+        @HapiTest
```
david-bakin-sl, line 98: Would this test be easier to read if some of this syntax noise was abstracted away with useful names?  E.g.,

* give a name to `asHeadlongAddress(Bytes.ofUnsignedLong(transferContractId.get()).toArray())` - perhaps `transferContractIdAsAddress` (or even the shorter `transferContractAddr`) and then that name was used as the argument to `contractCall(...)`?  (Same for `receiverContractId` too of course.)
* Similar for `fungibleTokenNum.get()`?  Might not be worth if you think of it as saving 6 characters but `.get()` is just noise, maybe it reads better without it (even though you have to invent a name).

Though it is effort to invent the names (I know!) and each such name requires a field declaration line, the result would, I think, be that the actual _test_ which you're trying to understand would "pop out" at you (since it would be much shorter - # of lines - and much less crufty) and then would be more comprehensible.

In the future - suppose we had Solidity in the build for these two contracts, perhaps we could then generate Java wrappers for each contract we have!  So you could write `TOKEN_TRANSFER_CONTRACT.approvePublic(fungibleToken, transferContractAddr, 2L).gas(gas_1M)` instead of the 8 lines it takes now?  And it would be syntax checked too!)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/token/TransferTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+        @DisplayName("transferring owner's tokens using transferToken function")
```
david-bakin-sl, line 99: For clarity: `using transferToken function` -> `using transferToken function which doesn't need an allowance` or `using transferToken function without explicit allowance`.

david-bakin-sl (commented):
Left a couple of comments on what I think could be done (in general!) to improve understandability of the tests for anyone who comes along later to try to read it.  Knowing that we're still feeling our way around this new test style in the first place, and have long been looking for more understandable tests in this codebase in the second place.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/dsl/operations/transactions/TransferTokenOperation.java
```diff
@@ -0,0 +1,96 @@
...
+
+    // non-standard ArrayList initializer
+    @SuppressWarnings({"java:S3599", "java:S1171"})
+    public TransferTokenOperation(
+            final long amount,
+            @NonNull final SpecToken token,
+            @NonNull final SpecAccount sender,
+            @NonNull final SpecAccount receiver) {
+        super(new ArrayList<>() {
+            {
```
david-bakin-sl, line 56: Would `List.of(...)` work instead of the anon class "initializer"?

david-bakin-sl (approved):
Wow everything is so much simpler with that operation, and with the use of `SpecContract`.  Everything!  The `@BeforeAll` and now in each test it's really visible what's going on.  TY!  

(We should add to our "cookbook" a list of "good example" test classes - maybe it's already there - this could be one.)

MrValioBg (approved):
LGTM! Thank you @lukelee-sl !

PR #13830, opened 2024-06-13
13658 implement token airdrop transaction

vtronkov:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #14278, opened 2024-07-17
feat: refactor CryptoTransferHandler

JivkoKelchev:
**Note**: This PR is a separated part of https://github.com/hashgraph/hedera-services/pull/13830

**Description**:

* Extract the transfer logic into "service" class static methods.
This way, we allow other handlers to execute crypto transfers without dispatching child transactions and build child records.
* Adds the possibility to execute custom fee assessment steps separately. This was needed in `TokenAirdropHandler` to prepay the customs fees. Also, in the airdrop claim transaction, we will need to transfer the tokens free of charge.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CryptoTransferExecutor.java
```diff
@@ -0,0 +1,226 @@
...
+import com.hedera.node.app.service.token.records.CryptoTransferRecordBuilder;
+import com.hedera.node.app.spi.workflows.HandleContext;
+import com.hedera.node.app.spi.workflows.HandleException;
+import com.hedera.node.app.spi.workflows.PreCheckException;
+import com.hedera.node.config.data.LazyCreationConfig;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CryptoTransferExecutor {
```
Neeharika-Sompalli, line 35: We can make this a Singleton and a base class for both `CryptoTransfer` and `AirdropHandler`. That way the methods need not be static

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CryptoTransferExecutor.java
```diff
@@ -0,0 +1,226 @@
...
+            CryptoTransferRecordBuilder recordBuilder) {
+        executeCryptoTransfer(txn, transferContext, context, validator, recordBuilder, false);
+    }
+
+    public static void chargeCustomFee(TransactionBody txn, TransferContextImpl transferContext) {
+        final var customFeeStep = new CustomFeeAssessmentStep(txn.cryptoTransferOrThrow());
+        var transferBodies = customFeeStep.assessCustomFees(transferContext);
+        var topLevelPayer = transferContext.getHandleContext().payer();
+        for (var customFeeTransactionBody : transferBodies) {
+            if (customFeeTransactionBody
```
tinker-michaelj, line 60: Won't this always be equivalent to skipping the first body,
```
for (int i = 1, n = transferBodies.size(); i < n; i++) {
   ...
```
? I think that will be clearer.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CryptoTransferExecutor.java
JivkoKelchev: I changed it to be a singleton, and got rid of the static methods. But about the base class, we had a discussion a while ago when we decided not to use the dispatch. 
, Michael had some concerns about the hierarchy and handlers having common parents.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/CryptoTransferExecutor.java
JivkoKelchev: Yes, it is the first one.
I was thinking that, if we just skip the first one was not clear enough, why we skip it.
Anyway, I did it like you suggest, and added a comment about it.

Neeharika-Sompalli (dismissed):
LGTM! Seems like just moving around code

hedera-node/configuration/mainnet/upgrade/feeSchedules.json
```diff
@@ -433,6 +433,52 @@
           ]
         }
       },
+      {
+        "transactionFeeSchedule": {
```
Neeharika-Sompalli, line 437: Need to merge back develop

vtronkov: Done 👍 The diff is okay now.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/CryptoTransferValidator.java
```diff
@@ -85,49 +90,18 @@ public void pureChecks(@NonNull final CryptoTransferTransactionBody op) throws P
...
-            validateFalsePreCheck(
-                    uniqueTokenAcctIds.isEmpty() && nftIds.isEmpty(), EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS);
-        }
-        validateFalsePreCheck(tokenIds.size() < tokenTransfers.size(), TOKEN_ID_REPEATED_IN_TOKEN_LIST);
+    /**
+     * Performs pure checks that validates basic fields in the token airdrop transaction.
+     * @param op the token airdrop transaction body
+     * @throws PreCheckException if any of the checks fail
+     */
+    public void airdropsPureChecks(@NonNull final TokenAirdropTransactionBody op) throws PreCheckException {
```
Neeharika-Sompalli, line 101: These should be in `airdropValidator`. Or have a base class for both. But having airdropPureChecks in `CryptoTransferValidator` name looks different

ibankov: This is fixed in the handler PR, as there was similar comment there.

Neeharika-Sompalli: Ah got it. Thanks

Neeharika-Sompalli (dismissed):
LGTM!

Neeharika-Sompalli (approved):

Issue #13889, opened 2024-06-17
Change the lifecycle of the pending records to be in `SavePoint`


Neeharika-Sompalli:
Related to https://hips.hedera.com/hip/hip-993
- [x] #14129
- [x] Create a `RecordBuilder` interface that shows the `TransactionCategory` and `ReversingBehavior`
- [x] Create an API on `SavePointStack` to return List<RecordBuilder> given TransactionCategory and ReversingBehavior
- [x] Replace all the places `RecordListBuilder` is used with a method in `RecordBuilder` that uses SavePointStackImpl
- [x] #14171

PR #14280, opened 2024-07-17
Generify record builder interface to allow for block stream items

mhess-swl:
I've attempted to add a layer of abstraction, namely `SingleTransactionBuilder`, on top of `SingleTransactionRECORDBuilder` in order to move towards creating blocks/block items in place of records. The change set looks large but a lot of it is renaming only. Notably, there are only minimal changes to the handle workflow itself. The most interesting classes are probably (not in order):

* StreamProducer
* SingleTransaction[Record]Builder
* AbstractBlockManagerImpl
* BlockRecordManagerImpl
* BlockStreamManager
* BlockStreamProducer
* BlockRecordStreamProducer

This implementation isn't finished, but I would like some feedback before moving further. After our block streams meeting yesterday, do these changes align with the vision discussed?

Note: I don't intend to merge this PR, but wanted a place to gather feedback


hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionBlockItems.java
```diff
@@ -0,0 +1,52 @@
...
+        ) implements SingleTransaction {
+    public SingleTransactionBlockItems {
+        requireNonNull(transaction, "transaction must not be null");
+//        requireNonNull(transactionResult, "transactionResult must not be null");
+        requireNonNull(transactionOutputs, "transactionOutputs must not be null");
+//        requireNonNull(stateChanges, "stateChanges must not be null");
+    }
+
+    // This is used by BlockStream, and is not serialized.
+    public record TransactionOutputs(@Nullable TokenType tokenType) {}
```
Neeharika-Sompalli, line 47: Is this not a protobuf message ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -172,28 +174,32 @@ public void rebuild() {
     public void add(
             final long nodeId,
             @NonNull final AccountID payerAccountId,
-            @NonNull final List<SingleTransactionRecord> transactionRecords) {
+            @NonNull final List<SingleTransaction> singleTxns) {
```
Neeharika-Sompalli, line 177: Why do we need the wrapper with transaction and consensusTime ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionBlockItemsBuilderImpl.java
```diff
@@ -0,0 +1,1189 @@
...
+    // fields needed for TransactionRecord
+    // Mutable because the provisional consensus timestamp assigned on dispatch could
+    // change when removable records appear "between" this record and the parent record
+    private Instant consensusNow;
+    private Instant parentConsensus;
+    private TransactionID transactionID;
+    private List<TokenTransferList> tokenTransferLists = new LinkedList<>();
+    private List<AssessedCustomFee> assessedCustomFees = new LinkedList<>();
+    private List<TokenAssociation> automaticTokenAssociations = new LinkedList<>();
+
```
Neeharika-Sompalli, line 152: Is this same as `SingleTransactionRecordBuilderImpl` ?

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/records/OngoingBlockInfo.java
```diff
@@ -23,18 +23,19 @@
 import java.time.Instant;
 
 /**
- * Interface for reading running hashes and block information managed by BlockRecordManager
+ * Interface for reading recent block information.
  */
-public interface BlockRecordInfo {
+public interface OngoingBlockInfo {
```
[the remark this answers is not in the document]
mhess-swl: Sure :) Maybe we don't even need to change it. If it didn't have 'record' in it... 😅

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionBlockItems.java
mhess-swl: No it's not. I'm not aware of any protobuf that groups these block items. There's `Block`, of course, but that's for a round not a single transaction. Did I miss something that has these already?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SingleTransactionBlockItemsBuilderImpl.java
mhess-swl: I copy-pasted this initially. I've taken a lot more out of this class so they aren't the same now

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/SingleTransactionBuilder.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.List;
+import java.util.Set;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Defines API for constructing stream items of a single transaction dispatch.
+ * The implementation may produce only records or could produce block items
+ */
+public interface SingleTransactionBuilder extends FeeBuilder {
```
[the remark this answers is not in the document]
mhess-swl: Sure :)

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/record/SingleTransaction.java
```diff
@@ -0,0 +1,34 @@
...
+import com.hedera.hapi.node.base.Timestamp;
+import com.hedera.hapi.node.base.Transaction;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+/**
+ * A representation of a single transaction
+ */
+public interface SingleTransaction {
```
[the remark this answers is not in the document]
mhess-swl: So we can have both `SingleRecordTransaction` and `BlockStreamTransaction`. If both of these classes implement an interface, then we can just reference the interface in the handle workflow. 

Also, by using an interface instead of `SingleRecordTransaction`, we don't have to change every single place that uses `SingleRecordTransaction` yet. It's a way that a PR can be split so there aren't so many files changed, but without actually changing the functionality

mhess-swl:
Feedback gathered. Moving ahead now that #13889 is merged

PR #14288, opened 2024-07-17
chore: Initial BlockStreamManagerImpl wip

### LINKED ISSUE #14247: Create an initial version of BlockStreamManager to produce Blocks

derektriley:
**Work In Progress**

This PR will introduce a new implementation of BlockRecordManager. This new implementation is in BlockStreamManagerImpl and is incomplete in many ways but will serve as a starting point for further development for the team. Currently this implementation is serial in execution and block the handle thread while writing to the block stream.  Only writing to files is currently implemented. If we want to write both the previous Record stream and the Block stream, some additional consideration to the design will be necessary. Perhaps a Set of BlockRecordManager could be invoked upon during HandleWorkflow to produce both streams with the appropriate logic.

A Block begins upon calling `blockRecordManager.startRound();` during `HandleWorkflow::handleRound`. Currently a Block will end in the call to `blockRecordManager.endRound(state);` if the rounds until the next block have passed according to the setting defined in `BlockStreamConfig.java` (numRoundsInBlock). Note the Block needs to actually be closed through some asynchronous process once the Node has gathered enough signatures of the Merkle root hash of the BlockProof.

The `RunningHash` is still maintained in state but is now the running hash of every BlockItem being written to the block stream. I'm not sure if we still need this? We we will need to implement keeping track of the currently computed Merkle root hashes of the transaction input and outputs trees. Perhaps this could be managed through `BlockStreamProducer` impl.

The `BlockStreamWriter` interface defines methods for initializing, writing items and closing a writer. There is an implementation which writes to Files. The `BlockStreamWriterFactoryImpl` provides an implementation which currently is non-concurrent. The `BlockStreamFileWriterV1` can be wrapped by `ConcurrentBlockStreamWriter` to allow for concurrent writes. The writes would not block the caller.

Similarly for `BlockStreamProducer` there is a single threaded implementation that could be wrapped by a concurrent implementation of the interface to make it thread-safe and non-blocking for the caller.

Fixes #14247 

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
@@ -1,0 +1,400 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.node.app.records.impl;
+
+import static com.hedera.node.app.records.BlockRecordService.EPOCH;
+import static com.hedera.node.app.records.impl.BlockRecordInfoUtils.HASH_SIZE;
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.hapi.node.base.Timestamp;
+import com.hedera.hapi.node.state.blockrecords.BlockInfo;
+import com.hedera.hapi.node.state.blockrecords.RunningHashes;
+import com.hedera.node.app.records.BlockRecordManager;
+import com.hedera.node.app.records.BlockRecordService;
+import com.hedera.node.app.records.schemas.V0490BlockRecordSchema;
+import com.hedera.node.app.state.SingleTransactionStreamRecord;
+import com.hedera.node.config.ConfigProvider;
+import com.hedera.node.config.data.BlockStreamConfig;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.common.crypto.DigestType;
+import com.swirlds.common.crypto.Hash;
+import com.swirlds.platform.state.PlatformState;
+import com.swirlds.platform.state.spi.WritableSingletonStateBase;
+import com.swirlds.platform.system.transaction.ConsensusTransaction;
+import com.swirlds.state.HederaState;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.time.Instant;
+import java.util.concurrent.ExecutorService;
+import java.util.stream.Stream;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ *
+ */
+@Singleton
+public class BlockStreamManagerImpl implements BlockRecordManager {
+    private static final Logger logger = LogManager.getLogger(BlockStreamManagerImpl.class);
+    private final ExecutorService executor;
+
+    /**
+     * The number of rounds in the current provisional block. "provisional" because the block is not yet complete.
+     */
+    // Should this be stored in BlockInfo so that it's safe for restarts?
+    private int roundsUntilNextBlock;
+
+    /**
+     * Keeps track if this block is currently open.
+     */
+    // Should this be stored in BlockInfo so that it's safe for restarts?
+    private boolean blockOpen;
+
+    private final int numBlockHashesToKeepBytes;
+    private final int numRoundsInBlock;
+    private final BlockStreamProducer blockStreamProducer;
+    /** True when we have completed event recovery. This is not yet implemented properly. */
+    private boolean eventRecoveryCompleted = false;
+
+    /**
+     * A {@link BlockInfo} of the most recently completed block. This is actually available in state, but there
+     * is no reason for us to read it from state every time we need it, we can just recompute and cache this every
+     * time we finish a provisional block.
+     */
+    private BlockInfo lastBlockInfo;
+
+    @Inject
+    public BlockStreamManagerImpl(
+            @NonNull final ExecutorService executor,
+            @NonNull final ConfigProvider configProvider,
+            @NonNull final HederaState state,
+            @NonNull final BlockStreamProducer blockStreamProducer) {
+
+        this.executor = requireNonNull(executor);
+        requireNonNull(state);
+        requireNonNull(configProvider);
+        this.blockStreamProducer = requireNonNull(blockStreamProducer);
+
+        // Get static configuration that is assumed not to change while the node is running
+        final var blockStreamConfig = configProvider.getConfiguration().getConfigData(BlockStreamConfig.class);
+        this.numRoundsInBlock = blockStreamConfig.numRoundsInBlock();
+        if (this.numRoundsInBlock <= 0) {
+            throw new IllegalArgumentException("numRoundsInBlock must be greater than 0");
+        }
+        this.numBlockHashesToKeepBytes = blockStreamConfig.numOfBlockHashesInState() * HASH_SIZE;
+
+        // Initialize the last block info and provisional block info.
+        // NOTE: State migration happens BEFORE dagger initialization, and this object is managed by dagger. So we are
+        // guaranteed that the state exists PRIOR to this call.
+        final var states = state.getReadableStates(BlockRecordService.NAME);
+        final var blockInfoState = states.<BlockInfo>getSingleton(V0490BlockRecordSchema.BLOCK_INFO_STATE_KEY);
+        this.lastBlockInfo = blockInfoState.get();
+        assert this.lastBlockInfo != null : "Cannot be null, because this state is created at genesis";
+        this.blockOpen = false;
+        this.roundsUntilNextBlock = numRoundsInBlock;
+
+        // Initialize the stream file producer. NOTE, if the producer cannot be initialized, and a random exception is
+        // thrown here, then startup of the node will fail. This is the intended behavior. We MUST be able to produce
+        // block streams, or there really is no point to running the node!
+        final var runningHashState =
+                states.<RunningHashes>getSingleton(V0490BlockRecordSchema.RUNNING_HASHES_STATE_KEY);
+        final var lastRunningHashes = runningHashState.get();
+        assert lastRunningHashes != null : "Cannot be null, because this state is created at genesis";
+        this.blockStreamProducer.initFromLastBlock(lastRunningHashes, this.lastBlockInfo.lastBlockNumber());
+    }
+
+    @Override
+    public boolean startUserTransaction(
+            @NonNull Instant consensusTime, @NonNull HederaState state, @NonNull PlatformState platformState) {
+        if (EPOCH.equals(lastBlockInfo.firstConsTimeOfCurrentBlock())) {
+            // This is the first transaction of the first block, so set both the firstConsTimeOfCurrentBlock
+            // and the current consensus time to now
+            final var now = new Timestamp(consensusTime.getEpochSecond(), consensusTime.getNano());
+            lastBlockInfo = lastBlockInfo
+                    .copyBuilder()
+                    .consTimeOfLastHandledTxn(now)
+                    .firstConsTimeOfCurrentBlock(now)
+                    .build();
+            putLastBlockInfo(state);
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public void advanceConsensusClock(@NonNull Instant consensusTime, @NonNull HederaState state) {
+        final var builder = this.lastBlockInfo
+                .copyBuilder()
+                .consTimeOfLastHandledTxn(Timestamp.newBuilder()
+                        .seconds(consensusTime.getEpochSecond())
+                        .nanos(consensusTime.getNano()));
+        if (!this.lastBlockInfo.migrationRecordsStreamed()) {
+            // Any records created during migration should have been published already. Now we shut off the flag to
+            // disallow further publishing
+            builder.migrationRecordsStreamed(true);
+        }
+        final var newBlockInfo = builder.build();
+
+        // Update the latest block info in state
+        final var states = state.getWritableStates(BlockRecordService.NAME);
+        final var blockInfoState = states.<BlockInfo>getSingleton(V0490BlockRecordSchema.BLOCK_INFO_STATE_KEY);
+        blockInfoState.put(newBlockInfo);
+        // Commit the changes. We don't ever want to roll back when advancing the consensus clock
+        ((WritableSingletonStateBase<BlockInfo>) blockInfoState).commit();
+
+        // Cache the updated block info
+        this.lastBlockInfo = newBlockInfo;
+    }
+
+    @Override
+    public void endUserTransaction(
+            @NonNull SingleTransactionStreamRecord transactionRecord, @NonNull HederaState state) {
+        // check if we need to run event recovery before we can write any new records to stream
+        if (!this.eventRecoveryCompleted) {
+            // FUTURE create event recovery class and call it here. Should this be in startUserTransaction()?
+            this.eventRecoveryCompleted = true;
+        }
+
+        // Write the user transaction records to the block stream.
+        blockStreamProducer.writeBlockItems(
+                transactionRecord.blockItemStream() == null
+                        ? Stream.empty()
+                        : requireNonNull(transactionRecord.blockItemStream()));
+    }
+
+    @Override
+    public void startRound() {
+        roundsUntilNextBlock--;
+
+        // We do not have an open block so create a new one.
+        if (!isBlockOpen()) beginBlock();
+    }
+
+    @Override
+    public void endRound(@NonNull HederaState state) {
+        if (roundsUntilNextBlock <= 0) {
+            updateBlockInfoForEndRound(state);
+            resetRoundsUntilNextBlock();
+        }
+
+        updateRunningHashesInState(state);
+
+        blockStreamProducer.endBlock();
+    }
+
+    @Override
+    public void close() {
+        try {
+            blockStreamProducer.close();
+        } catch (final Exception e) {
+            // This is a fairly serious warning. This basically means we cannot guarantee that some records were
+            // produced. However, since the {@link BlockRecordManager} is a singleton, this close method is only called
+            // when the node is being shutdown anyway.
+            logger.warn("Failed to close blockStreamProducer properly", e);
+        }
+    }
+
+    @Override
+    public void markMigrationRecordsStreamed() {
+        lastBlockInfo =
+                lastBlockInfo.copyBuilder().migrationRecordsStreamed(true).build();
+    }
+
+    @NonNull
+    @Override
+    public Instant consTimeOfLastHandledTxn() {
+        final var lastHandledTxn = lastBlockInfo.consTimeOfLastHandledTxn();
+        return lastHandledTxn != null
+                ? Instant.ofEpochSecond(lastHandledTxn.seconds(), lastHandledTxn.nanos())
+                : Instant.EPOCH;
+    }
+
+    @Override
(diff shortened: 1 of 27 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamFormat.java
```diff
@@ -38,13 +39,37 @@
  */
 public interface BlockStreamFormat {
 
+    /**
+     * Given the starting running hash and a stream of serialized {@link BlockItem}s, compute the new running hash by
+     * adding each {@link BlockItem}'s hash to the running hash one at a time.
```
mhess-swl, line 44: I'm still unsure if we're keeping this running hash or not. I had the impression the running hash is going away, and instead we'll use the hash inside of each block's `BlockProof`. But, if that is the case, are we maintaining evm equivalence only by keeping the hashes of the last 256 blocks? Or, do we need a per-transaction running hash to report in addition to the running block hash?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamProducerSingleThreaded.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private void writeSerializedBlockItem(@NonNull final Bytes serializedItem) {
+        try {
+            // Depending on the configuration, this writeItem may be an asynchronous or synchronous operation. The
+            // BlockStreamWriterFactory instantiated by dagger will determine this.
+            writer.writeItem(serializedItem, runningHash);
+        } catch (final Exception e) {
+            // This **may** prove fatal. The node should be able to carry on, but then fail when it comes to
+            // actually producing a valid record stream file. We need to have some way of letting all nodesknow
```
mhess-swl, line 94:
```suggestion
            // actually producing a valid block file. We need to have some way of letting all nodes know
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamProducerSingleThreaded.java
```diff
@@ -0,0 +1,248 @@
...
+        final var lastRunningHash = getRunningHashObject();
+        closeWriter(lastRunningHash, this.currentBlockNumber);
+    }
+
+    @Override
+    public void writeSystemTransaction(ConsensusTransaction platformTxn) {
+        final var serializedBlockItem = format.serializeSystemTransaction(platformTxn);
+        updateRunningHashes(serializedBlockItem);
+        writeSerializedBlockItem(serializedBlockItem);
+    }
```
mhess-swl, line 151: I'm wondering if this method should do any checking to make sure it's not writing a system transaction block item anywhere except immediately following a header. Not sure if that logic should live here or if it should live in the block manager.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamProducerSingleThreaded.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    public void close() throws Exception {
+        // FUTURE: close() should wait until the block is completed, which cannot happen until the block proof is
+        // produced, and that cannot happen until the signatures have been gossiped. Today, this will close in
+        // unpredictable ways, and will most likely result in the block being written incompletely without a block
+        // proof.
```
mhess-swl, line 158: I suspect we'll end up collecting signatures for the block proof in the block manager, and only call this `close()` when we have sufficient signatures. What do you think, though?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamProducerSingleThreaded.java
```diff
@@ -0,0 +1,248 @@
...
+import javax.inject.Inject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ *  A single threaded implementation of {@link BlockStreamProducer} where operations may block the single calling thread.
+ *  This simple implementation is also useful for testing and for the "Hedera Local Node", which is a version of a
+ *  consensus node that may be run as a standalone application, useful for development. This class is not thread-safe.
+ */
+public class BlockStreamProducerSingleThreaded implements BlockStreamProducer {
```
mhess-swl, line 46: Looks great overall!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
```diff
@@ -0,0 +1,400 @@
...
+
+    /**
+     * A {@link BlockInfo} of the most recently completed block. This is actually available in state, but there
+     * is no reason for us to read it from state every time we need it, we can just recompute and cache this every
+     * time we finish a provisional block.
+     */
+    private BlockInfo lastBlockInfo;
+
+    @Inject
+    public BlockStreamManagerImpl(
```
mhess-swl, line 83: Does this class need to be thread safe? I don't think so since we should only be calling it from the handle workflow, but it's worth asking the question

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
```diff
@@ -0,0 +1,400 @@
...
+                    .firstConsTimeOfCurrentBlock(now)
+                    .build();
+            putLastBlockInfo(state);
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public void advanceConsensusClock(@NonNull Instant consensusTime, @NonNull HederaState state) {
```
mhess-swl, line 141: This and a few other methods in this class are identical to `BlockRecordManagerImpl`. We may want to consider relocating the common code

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
```diff
@@ -0,0 +1,400 @@
...
+        if (!this.eventRecoveryCompleted) {
+            // FUTURE create event recovery class and call it here. Should this be in startUserTransaction()?
+            this.eventRecoveryCompleted = true;
+        }
+
+        // Write the user transaction records to the block stream.
+        blockStreamProducer.writeBlockItems(
+                transactionRecord.blockItemStream() == null
+                        ? Stream.empty()
+                        : requireNonNull(transactionRecord.blockItemStream()));
```
mhess-swl, line 178: Does `TransactionRecord` have a `blockItemStream()` property now?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamManagerImpl.java
```diff
@@ -0,0 +1,400 @@
...
+            hashOfJustFinishedBlock.getBytes(0, newBlockHashesBytes, newBlockHashesBytes.length - HASH_SIZE, HASH_SIZE);
+        }
+
+        return new BlockInfo(
+                justFinishedBlockNumber,
+                lastBlockInfo.firstConsTimeOfCurrentBlock(),
+                Bytes.wrap(newBlockHashesBytes),
+                lastBlockInfo.consTimeOfLastHandledTxn(),
+                lastBlockInfo.migrationRecordsStreamed(),
+                null);
```
mhess-swl, line 370:
```suggestion
        return lastBlockInfo.copyBuilder().
                .blockNumber(justFinishedBlockNumber)
                .blockHashes(Bytes.wrap(newBlockHashesBytes))
                .firstConsTimeOfCurrentBlock(null);
```

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
```diff
@@ -0,0 +1,46 @@
...
+import com.swirlds.config.api.ConfigProperty;
+import com.swirlds.config.api.validation.annotation.Max;
+import com.swirlds.config.api.validation.annotation.Min;
+
+/**
+ * Configuration for block streams. This configuration is enabled when BlockRecordStreamConfig is set to
+ * recordFileVersion=8. It is an error for recordFileVersion != blockVersion.
+ *
+ * @param logDir directory for writing block files
+ * @param compressFilesOnCreation when true record and sidecar files are compressed with GZip when created
```
mhess-swl, line 31:
```suggestion
 * @param compressFilesOnCreation when true, block files are compressed with GZip when created
```

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
```diff
@@ -0,0 +1,46 @@
...
+ * @param numOfBlockHashesInState the number of block hashes to keep in state for block history
+ * @param streamFileProducer the type of stream file producer to use. Currently only "concurrent" and "sequential" are supported
```
mhess-swl, line 33: "SERIAL" instead of "sequential"?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamFormat.java
```diff
@@ -38,13 +39,37 @@
...
+     * @return the new running hash, or startRunningHash if there were no items to add
+     */
+    @NonNull
+    Bytes computeNewRunningHash(
+            @NonNull final MessageDigest messageDigest,
+            @NonNull final Bytes startRunningHash,
+            @NonNull final Bytes serializedItem);
+
+    /**
+     * Returns an instance of a non-thread safe MessageDigest that can be used to compute the hash of block items.
```
mhess-swl, line 59: We want to be very careful with a non thread safe object if we're using `ConcurrentBlockStreamWriter`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/BlockStreamWriterFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.hedera.node.app.records.impl.producers;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Creates a new {@link BlockStreamWriter} instance on demand, based on configuration. During processing of
+ * transactions, when we determine it is time to write a new block to file, then we need to create a new
+ * {@link BlockStreamWriter} instance. This factory is used to create that instance. Creation of the instance may fail,
```
mhess-swl, line 24: > then we need to create a new BlockStreamWriter instance

We may want to do this differently with the grpc writer, e.g. to transmit multiple blocks without closing a socket in between

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v1/BlockStreamFormatV1.java
```diff
@@ -130,80 +196,6 @@ private TransactionResult extractTransactionResult(@NonNull final SingleTransact
...
-        //        if (body == null && item.transactionRecord().memo().equals("End of staking period calculation
-        // record")) {
-        //            // TODO: Skip this one for now till I figure out how we want to handle it.
-        //            return builder.build();
-        //        }
-        // One case where it appears to be empty is when we have a transactionRecord
-
-        // item.transactionRecord().body().kind()
-        //        var body = item.transaction().bodyOrThrow();
-        switch (item.transactionOutputs().transactionBodyType()) {
```
mhess-swl, line 154: Was this `switch` extracted to somewhere else?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v1/BlockStreamFormatV1.java
```diff
@@ -69,8 +111,32 @@ public Bytes serializeEventMetadata(@NonNull final EventMetadata eventMetadata)
 
     /** {@inheritDoc} */
     @Override
-    public Bytes serializeSystemTransaction(@NonNull final SystemTransaction systemTransaction) {
-        return null;
+    public Bytes serializeSystemTransaction(@NonNull final ConsensusTransaction systemTransaction) {
+        var systemTxnBuilder = SystemTransaction.newBuilder();
+
+        switch (systemTransaction) {
+            case StateSignatureTransaction stateSignatureTransaction -> {
```
mhess-swl, line 118: This code seems to suggest the platform will gather the signatures and generate a single `StateSignatureTransaction`. I was under the impression that signatures for each node come in different system transactions; do we not need to collect those somewhere?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v8/BlockStreamFileWriterV1.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    public BlockStreamFileWriterV1(
+            @NonNull final BlockStreamConfig config,
+            @NonNull final NodeInfo nodeInfo,
+            @NonNull final FileSystem fileSystem,
+            @NonNull final SemanticVersion hapiProtoVersion,
+            @NonNull final Signer signer) {
+
+        if (config.blockVersion() != BlockStreamFormatV1.VERSION_8) {
```
mhess-swl, line 108:
```suggestion
        if (config.blockVersion() >= BlockStreamFormatV1.VERSION_8) {
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v8/BlockStreamFileWriterV1.java
```diff
@@ -0,0 +1,244 @@
...
+            logger.fatal(
+                    "Bad configuration: BlockStreamFileWriterV1 used with block version {}", config.blockVersion());
+            throw new IllegalArgumentException("Configuration block version must be >= 7 for BlockStreamFileWriterV1!");
+        }
+
+        this.state = State.UNINITIALIZED;
+        this.hapiProtoVersion = requireNonNull(hapiProtoVersion);
+        this.compressFiles = config.compressFilesOnCreation();
+
+        // Compute directories for record and sidecar files.
```
mhess-swl, line 118:
```suggestion
        // Compute directory for block files
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v8/BlockStreamFileWriterV1.java
```diff
@@ -0,0 +1,244 @@
...
+        final Path blockDir = fileSystem.getPath(config.logDir());
+        nodeScopedBlockDir = blockDir.resolve("block-" + nodeInfo.memo());
+
+        // Create parent directories if needed for the record file itself.
```
mhess-swl, line 122:
```suggestion
        // Create parent directories if needed for the block file itself.
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v8/BlockStreamFileWriterV1.java
```diff
@@ -0,0 +1,244 @@
...
+    private Bytes endRunningHash;
+    /**
+     * The block number for the file we are writing. Each file corresponds to one, and only one, block. Once it is
+     * set in {@link #init}, it is never changed.
+     */
+    private long blockNumber;
+
+    private enum State {
+        UNINITIALIZED,
+        OPEN_WITHOUT_HEADER,
```
mhess-swl, line 96: I don't see `OPEN_WITHOUT_HEADER` used anywhere. Needed?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/producers/formats/v8/BlockStreamWriterFactoryImpl.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param nodeInfo the node info
+     * @param fileSystem the file system to use, also used in testing to be able to use a non-standard file system
+     */
+    @Inject
+    public BlockStreamWriterFactoryImpl(
+            @NonNull final ExecutorService executor,
+            @NonNull final ConfigProvider configProvider,
+            @NonNull final SelfNodeInfo nodeInfo,
+            @NonNull final Signer signer,
+            @NonNull final FileSystem fileSystem) {
```
mhess-swl, line 56: We may want to make this `@Nullable` when we support the grpc writer

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamProducer.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Produces a stream of blocks. This is used by the {@link BlockStreamManagerImpl}
+ */
+public interface BlockStreamProducer extends AutoCloseable {
+
+    /**
+     * Write a stream of Block Items to the block stream.
+     *
+     * @param blockItems The record stream items to write.
```
mhess-swl, line 35:
```suggestion
     * @param blockItems The block stream items to write.
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamProducer.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    void writeBlockItems(@NonNull final Stream<BlockItem> blockItems);
+
+    /**
+     * Initialize the current running hash of record stream items. This is called only once to initialize the running
```
mhess-swl, line 40:
```suggestion
     * Initialize the current running hash of block stream items. This is called only once to initialize the running
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamProducer.java
```diff
@@ -0,0 +1,83 @@
...
+     * hash on startup. This is called on the handle transaction thread. It is loaded from a saved state. At genesis,
+     * there is no saved state, so this will be called with null.
```
mhess-swl, line 42:
```suggestion
     * there is no saved state, so this will be called with a null running hash and a block number of X
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockStreamProducer.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * <p>Begins a new block. It is an error to begin a new block when the previous block has not been closed.
+     */
+    void beginBlock();
+
+    /**
+     * Called to close the current open block.
+     */
+    void endBlock();
+
```
mhess-swl, line 81: Missing javadoc

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/SingleTransactionStreamRecord.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.stream.Stream;
+
+/**
+ * This class is a record of a stream of single transaction records and a stream of block items.
+ * It is used to store the streams of single transaction records and block items in a single object.
+ */
+public record SingleTransactionStreamRecord(
+        @Nullable Stream<SingleTransactionRecord> singleTransactionRecordStream,
+        @Nullable Stream<BlockItem> blockItemStream) {}
```
mhess-swl, line 29: This is the same model that `SingleTransactionRecordBuilderImpl` follows, specifically that there's no indication of which properties are set until runtime (`SingleTr...` simply implements a superset of all the properties). While it's generally good to be consistent, I've found reasoning about the transaction builder to be very difficult. I think we would benefit significantly from making two separate (java) record classes with an interface for any properties common to both types.

mhess-swl (commented):
This is lots of work, well done!!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
derektriley, line 1: Reading the Block Stream draft HIP I am wondering if we need to end a block when a round is over vs only ending if the round contains a user transaction.

mhess-swl: Definitely a valid concern. I inadvertently left the poc running for a number of hours, and it almost killed my machine. I had accumulated 647 MB of mostly-empty .blk files. It took `rm -rf` ~22 min to remove them all

PR #14289, opened 2024-07-17
chore: cherry-pick Dont modify allowances on token reject

Neeharika-Sompalli:
Cherry-pick https://github.com/hashgraph/hedera-services/pull/14282


tinker-michaelj (approved):

PR #14291, opened 2024-07-17
feat: use "inline" auto-association fee payments

tinker-michaelj:
**Description**:
 - Switches from doing auto-associations via dispatch to doing them inline in the `CryptoTransferHandler`.
     * This implies the handler must charge fees whenever the context is not an internal dispatch from the `ContractService` (in that case the cost comes out of the `gas` budget for the EVM transaction).
     * Hence exposes `chargeNetworkFee()` in `HandleContext`.
 - Since there is no longer a synthetic `TokenAssociate` child, adds throttle check based on the number of automatic token associations in the children that are there.
 - Updates unit and `@HapiTest` expectations accordingly.


hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
```diff
@@ -119,6 +119,13 @@ enum PrecedingTransactionCategory {
     @NonNull
     AccountID payer();
 
+    /**
+     * Attempts to charge the payer in this context the given amount.
+     * @param amount the amount to charge
+     * @throws IllegalArgumentException if the payer cannot afford the fee
```
david-bakin-sl, line 125: `IllegalArgumentException` is sort of a stretch, that's usually for "static" constraints, like, must be positive.  Where this is "dynamic", if you see what I mean.  Sticking with JDK types, perhaps `RangeException`?

Actually a better option would be to have this method return something.  Even just a `boolean`.  Skip the exception (which already is not a pattern used in this `HandleContext` class except for a couple of places).

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -184,62 +197,27 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
-        accountStore.put(accountModified);
-        // We need to set this as auto-association
-        final var newTokenRel = requireNonNull(tokenRelStore.get(accountId, tokenId))
-                .copyBuilder()
-                .automaticAssociation(true)
-                .build();
-        tokenRelStore.put(newTokenRel);
+    private long associationFeeFor(@NonNull final HandleContext context, @NonNull final TransactionBody txnBody) {
+        final var topLevelPayer = context.payer();
+        // If we get here the payer must exist
```
david-bakin-sl, line 218: What does this comment mean?  `HandleContext.payer()` is declared to return `@NonNull` and is simply a getter.  So it isn't referring to L217 immediately above.

Should this be a comment on the method itself (like documenting a precondition)?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -184,62 +197,27 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
+        final var fees = context.dispatchComputeFees(txnBody, topLevelPayer, ComputeDispatchFeesAsTopLevel.NO);
+        return fees.serviceFee() + fees.networkFee() + fees.nodeFee();
```
david-bakin-sl, line 220: isn't this `fees.totalFee()`?

david-bakin-sl (dismissed):
Some comments for your consideration.

(Also, is this a `chore` and not a `fix` or maybe `feat`?  Also a reminder the PR description is TBS.)

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/HandleContext.java
tinker-michaelj: Yes I forgot to mark the initial PR "Draft"...was veryy half-baked. 😳 Changed this.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
tinker-michaelj: Fixed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
tinker-michaelj: Fixed.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -181,65 +195,33 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
-                return asTokenAssociation(tokenId, accountId);
-            } else {
-                // Once the unlimitedAssociationsEnabled is enabled, this block of code can be removed and
-                // all auto-associations will be done through the synthetic transaction and charged.
-                final var newRelation = autoAssociate(account, token, accountStore, tokenRelStore, config);
-                return asTokenAssociation(newRelation.tokenId(), newRelation.accountId());
+            // We only charge auto-association fees inline if this is a user disaptch; in that case the
+            // contract service will take the auto-association costs from the remaining EVM gas
+            if (context.recordBuilders()
+                    .getOrCreate(SingleTransactionRecordBuilder.class)
```
david-bakin-sl, line 201: (Gotta say, isn't obvious at this line how a method called `getOrCreate` works given an _interface_ as the target class - had to trace this one down ...)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java
```diff
@@ -181,65 +195,33 @@ private TokenAssociation validateAndBuildAutoAssociation(
...
+                    .isUserDispatch()) {
+                final var unlimitedAssociationsEnabled =
+                        config.getConfigData(EntitiesConfig.class).unlimitedAutoAssociationsEnabled();
+                // And the "sender pays" fee model only applies when using unlimited auto-associations
+                if (unlimitedAssociationsEnabled) {
+                    final var autoAssociationFee = associationFeeFor(context, PLACEHOLDER_SYNTHETIC_ASSOCIATION);
+                    if (!context.tryToChargePayer(autoAssociationFee)) {
+                        throw new HandleException(INSUFFICIENT_PAYER_BALANCE);
```
david-bakin-sl, line 209: `INSUFFICIENT_PAYER_BALANCE` seems right - but given that this PR is to make sure the _gas_ is charged for it ... do we have a rule we follow about using this response code vs. `INSUFFICIENT_GAS`? I would expect something like if it is directly EVM related then `INSUFFICIENT_GAS` but if it is more "hapi" (Hedera semantics) related then `INSUFFICIENT_PAYER_BALANCE`, but I've never heard it discussed.

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/transfer/AdjustFungibleTokenChangesStepTest.java
```diff
@@ -146,7 +152,7 @@ void doesTokenBalanceChangesWithAllowances() {
         assertThat(receiverAccountBefore.numberPositiveBalances()).isEqualTo(2);
         assertThat(senderRelBefore.balance()).isEqualTo(1000L);
         // There is an association happening during the transfer for auto creation
-        assertThat(receiverRelBefore.balance()).isEqualTo(1);
+        assertThat(receiverRelBefore.balance()).isEqualTo(0);
```
david-bakin-sl, line 155: Why wasn't this assert needed (or useful) also at L106 above? (Only instance in these tests modified in this PR where it wasn't.)

david-bakin-sl (commented):
LGTM but this is a comment not an approval because I want to come back to it.

david-bakin-sl (approved):
DOne looking. LGTM

Neeharika-Sompalli (approved):
LGTM! We should probably target this to develop

tinker-michaelj:
Will retarget to `develop`.

Issue #14234, opened 2024-07-13
(HIP-904) system contract `transfer*()` variants need updated gas calcs


tinker-michaelj:
### Description

Although the sender-pays model for charging auto-associations can make a `CryptoTransfer` considerably more expensive when auto-associations happen, the `gas` cost of HTS transfer variants [has not changed](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/ClassicTransfersCall.java#L166).

### Steps to reproduce

Trigger an auto-association through a `transfer*()` system contract call; note the gas price is unchanged vs the same transfer without an auto-association.

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
Done in https://github.com/hashgraph/hedera-services/pull/14299

PR #14292, opened 2024-07-17
fix: include auto-association work in `gas` cost of transfer system contracts

tinker-michaelj:
**Description**:
 - C.f. #14234 
 - Increases `gas` cost of a transfer system contract by how many auto-associations are reported in the stream builder returned from dispatch.
 - Covered in `@HapiTest` that does NFT transfer with and without auto-association and asserts the USD price of the increased gas cost is $0.06 (HAPI cost with 20% surcharge through EVM).

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/extensions/SpecEntityExtension.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/extensions/SpecEntityExtension.java
@@ -155,7 +155,9 @@ public class SpecEntityExtension implements ParameterResolver, BeforeAllCallback
                 .map(AccountSpec::name)
                 .filter(n -> !n.isBlank())
                 .orElse(defaultName);
-        return new SpecAccount(name);
+        final var account = new SpecAccount(name);
+        account.builder().maxAutoAssociations(annotation == null ? 0 : annotation.maxAutoAssociations());
+        return account;
     }
 
     private SpecKey keyFrom(@Nullable final KeySpec annotation, @NonNull final String defaultName) {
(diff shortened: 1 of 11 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/extensions/SpecEntityExtension.java
```diff
@@ -155,7 +155,9 @@ private SpecAccount accountFrom(@Nullable final AccountSpec annotation, @NonNull
                 .map(AccountSpec::name)
                 .filter(n -> !n.isBlank())
                 .orElse(defaultName);
-        return new SpecAccount(name);
+        final var account = new SpecAccount(name);
+        account.builder().maxAutoAssociations(annotation == null ? 0 : annotation.maxAutoAssociations());
```
david-bakin-sl, line 159: Isn't this statement a noop?

david-bakin-sl (commented):
Have a question.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/extensions/SpecEntityExtension.java
david-bakin-sl: NVM, I get it.  Atypical use of "builder" pattern here.

lukelee-sl (approved):
LGTM.

lukelee-sl (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM  from `hedera-base`

tinker-michaelj:
Not needed in `v0.52` since unlimited auto-associations were disabled.

PR #14310, opened 2024-07-18
chore: use in-repo protobufs

### LINKED ISSUE #14301: Demonstrate in-repo protobufs
### Problem

The logistical challenges of always fetching protobuf sources from a separate repo are prolifically documented and lived.

### Solution

Per @jjohannes -
- [ ] Remove the `GitClone` task [here](https://github.com/hashgraph/hedera-services/blob/develop/gradle/plugins/src/main/kotlin/com.hedera.gradle.protobuf.gradle.kts#L25).
- [ ] And its implementation [here](https://github.com/hashgraph/hedera-services/blob/develop/gradle/plugins/src/main/kotlin/com/hedera/gradle/tasks/GitClone.kt).
- [ ] Update the `pbj` and `proto` task source sets [here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/build.gradle.kts#L40-L49) to use sources from the project itself instead of the `GitClone` outputs.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14301 
 - Removes `GitClone` task and instead sources protobufs from _hapi/hedera-protobufs/_.
 - Initial contents are head of https://github.com/hashgraph/hedera-protobufs `main`.


tinker-michaelj:
Will be merged via https://github.com/hashgraph/hedera-services/pull/14579

PR #14316, opened 2024-07-18
fix: fixed a few issues on Dynamic Address Book

### LINKED ISSUE #14242: Fixed a few issues on Dynamic Address Book
Updated NodeUpdateTest, NodeCreateTest
Fixed HapiSpecRegistry.saveNodeMeta()
Fixed V053AddressBookSchema.migrate() adminKey default by checking account 55 key first.
Organized code so some common methods can be used.
Changed name to "node"+(nodeid+1) from node.description.

### LINKED ISSUE #14329: Assume node self-name is `node<NODE_ID+1>`, decouple from `Node#description` field
### Problem

We temporarily used the [`Node#description`](https://github.com/hashgraph/hedera-protobufs/blob/0deeafce2451b620aaddac291896e8d5cb257fcd/services/state/addressbook/node.proto#L62) protobuf type to allow test clients to inform the network of non-standard "self names" used in the test _config.txt_.

This was not a permanent change, as we want to standardize all self-names to `node<NODE_ID+1>`.

### Solution

Remove the `NodeInfo#selfName` [field](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/info/NodeInfo.java#L109) and decouple `Node#description` from _config.txt_ self names.

### Alternatives

_No response_

iwsimon:
**Description**:

Updated NodeUpdateTest, NodeCreateTest
Fixed HapiSpecRegistry.saveNodeMeta()
Fixed V053AddressBookSchema.migrate() adminKey default by checking account 55 key first.
Organized code so some common methods can be used.
Changed name to "node"+(nodeid+1) from node.description.

**Related issue(s)**:

Fixes #14242 
Fixes #14329

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip869/NodeCreateTest.java
```diff
@@ -263,13 +270,12 @@ final Stream<DynamicTest> failOnTooManyServiceEndpoints() {
...
         return hapiTest(
                 newKeyNamed(ED_25519_KEY).shape(KeyShape.ED25519),
-                overriding("nodes.gossipFqdnRestricted", "false"),
                 nodeCreate("nodeCreate")
                         .description("hello")
                         .gossipCaCertificate("gossip".getBytes())
                         .grpcCertificateHash("hash".getBytes())
                         .accountId(asAccount("0.0.100"))
-                        .gossipEndpoint(GOSSIP_ENDPOINTS)
+                        .gossipEndpoint(GOSSIP_ENDPOINTS_IPS)
```
kimbor, line 278: Why did you change this? With this change, the test becomes identical to allFieldsSetHappyCaseForIps() case defined below.

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
```diff
@@ -65,13 +70,30 @@ public void migrate(@NonNull final MigrationContext ctx) {
...
-                Key.newBuilder().ed25519(bootstrapConfig.genesisPublicKey()).build();
+        final var accountConfig = ctx.configuration().getConfigData(AccountsConfig.class);
+        var adminKey = Key.DEFAULT;
+        if (readableAccounts != null) {
+            final var adminAccount = readableAccounts.get(AccountID.newBuilder()
+                    .accountNum(accountConfig.addressBookAdmin())
+                    .build());
+            if (adminAccount != null) {
+                adminKey = adminAccount.keyOrElse(Key.DEFAULT);
+            }
```
Neeharika-Sompalli, line 90: If this is null it is error right ?

hedera-node/hedera-addressbook-service/src/main/java/com/hedera/node/app/service/addressbook/AddressBookHelper.java
```diff
@@ -0,0 +1,44 @@
...
+package com.hedera.node.app.service.addressbook;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.Spliterator.DISTINCT;
+
+import com.hedera.hapi.node.state.common.EntityNumber;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Spliterators;
+import java.util.stream.StreamSupport;
+
```
Neeharika-Sompalli, line 26: nit: javadoc for class

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -220,13 +219,12 @@ private void extractAndReplaceArtifacts(
             FileUtils.cleanDirectory(artifactsDir);
             UnzipUtility.unzip(archiveData.toByteArray(), artifactsLoc);
             log.info("Finished unzipping {} bytes for {} update into {}", size, desc, artifactsLoc);
-            if (desc.equals(PREPARE_UPGRADE_DESC)) {
-                requireNonNull(nodes, "Cannot generate config.txt without a valid list of active nodes");
+            if (nodes != null) {
```
Neeharika-Sompalli, line 222: Why is this condition changed?

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -235,6 +233,13 @@ private void extractAndReplaceArtifacts(
         }
     }
 
+    private long getNextNodeID(@NonNull List<ActiveNode> nodes) {
+        requireNonNull(nodes);
+        final long maxNodeId =
+                nodes.stream().mapToLong(a -> a.node.nodeId()).max().orElse(-1L);
+        return maxNodeId + 1;
+    }
+
```
Neeharika-Sompalli, line 242: Can we move this also to `AddressBookHelper` ? Can we re-use one version of this method?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip869/NodeCreateTest.java
```diff
@@ -443,6 +450,33 @@ final Stream<DynamicTest> failsAtIngestForUnAuthorizedTxns() {
                         .via("nodeCreation"));
     }
 
+    @LeakyHapiTest(overrides = {"nodes.maxNumber"})
```
tinker-michaelj, line 453: These tests are so (admirably!) focused that it barely matters, but the [checklist](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/README.md#the-hapitest-checklist) advises using an explicit `@DisplayName`.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip869/NodeCreateTest.java
iwsimon: They not the same. I wanted to test the default nodes.gossipFqdnRestricted setting without overriding, gossipEndpoint can only accept ip address, not domain name. You had nodes.gossipFqdnRestricted overrides in both allFieldsSetHappyCaseForDomains and allFieldsSetHappyCaseForIps.

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
iwsimon: readableAccounts.get(AccountID) is @Nullable. So I checked the null here.

hedera-node/hedera-addressbook-service/src/main/java/com/hedera/node/app/service/addressbook/AddressBookHelper.java
iwsimon: Added.

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
iwsimon: Because we have this check earlier:
 final var activeNodes = desc.equals(PREPARE_UPGRADE_DESC) ? allActiveNodes() : null;

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
iwsimon: The reason I created AddressBookHelper initially is tried to re-use getNextNodeID(@NonNull final ReadableNodeStore nodeStore) here. But I kept on getting failed immutable exception during DabEnabledUpgradeTest by using it, I was going to check with you and @tinker-michaelj later. This small function get me to pass that error for now.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip869/NodeCreateTest.java
iwsimon: Added @DisplayName.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip869/NodeCreateTest.java
kimbor: Ok, but now there is no place that domain names are being tested. There needs to be a test that overrides gossipFqdnRestricted and then uses domain names.

iwsimon: Updated the test with domain names.

kimbor: Looks good, thanks!

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
iwsimon: Since this is called in ExecutorService thread, we can't use ReadableStore directly.

kimbor (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @iwsimon

PR #14318, opened 2024-07-18
feat: log socket exception stack trace periodically

### LINKED ISSUE #14304: Add periodic stack trace printing for socket exception
Add a `RateLimiter` and `RateLimiter::requestAndTrigger()` check to periodically kick the error into the code path that reports the stack trace.  https://github.com/hashgraph/hedera-services/blob/e386407169bf21f81a8d29b2a1cce88b5782d627/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java#L92

```
if (SOCKET_EXCEPTIONS.getMarker().equals(marker) && !stackTraceRateLimiter.requestAndTrigger()) { 
```

anthony-swirldslabs:
**Description**:
Add a RateLimiter and log the stack trace of socket exceptions once a minute.

**Related issue(s)**:

Fixes #14304

**Notes for reviewer**:
Trivial changes to improve logging and diagnostics.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
@@ -21,6 +21,7 @@ import static com.swirlds.logging.legacy.LogMarker.SOCKET_EXCEPTIONS;
 
 import com.swirlds.common.crypto.config.CryptoConfig;
 import com.swirlds.common.platform.NodeId;
+import com.swirlds.common.utility.throttle.RateLimiter;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.platform.Utilities;
 import com.swirlds.platform.crypto.KeysAndCerts;
@@ -70,10 +71,12 @@ public final class NetworkUtils {
      *
      * @param e          the exception that was thrown
      * @param connection the connection used when the exception was thrown
+     * @param socketExceptionRateLimiter a rate limiter for reporting full stack traces for socket exceptions
      * @throws InterruptedException if the provided exception is an {@link InterruptedException}, it will be rethrown
      *                              once the connection is closed
      */
-    public static void handleNetworkException(final Exception e, final Connection connection)
+    public static void handleNetworkException(
+            final Exception e, final Connection connection, final RateLimiter socketExceptionRateLimiter)
             throws InterruptedException {
         final String description;
         // always disconnect when an exception gets thrown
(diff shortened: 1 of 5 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
```diff
@@ -90,8 +93,12 @@ public static void handleNetworkException(final Exception e, final Connection co
         // we use a different marker depending on what the root cause is
         final Marker marker = NetworkUtils.determineExceptionMarker(e);
         if (SOCKET_EXCEPTIONS.getMarker().equals(marker)) {
-            final String formattedException = NetworkUtils.formatException(e);
-            logger.warn(marker, "Connection broken: {} {}", description, formattedException);
+            if (socketExceptionRateLimiter.requestAndTrigger()) {
+                logger.warn(marker, "Connection broken: {}", description, e);
```
kfa-aguda, line 97: not in a compiler so can't check but I think this should be `Connection broken: {} {}` so the placeholders match the args?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
```diff
@@ -90,8 +93,12 @@ public static void handleNetworkException(final Exception e, final Connection co
...
+            } else {
+                final String formattedException = NetworkUtils.formatException(e);
+                logger.warn(marker, "Connection broken: {} {}", description, formattedException);
+            }
         } else {
             logger.error(EXCEPTION.getMarker(), "Connection broken: {}", description, e);
```
kfa-aguda, line 103: same thing here though this isn't part of your change. If it is, could you pls fix?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
anthony-swirldslabs: No. This is a feature of the Logger interface - it accepts one extra argument of type `Throwable`, and when present, it prints its entire stack trace. If there was an explicit `{}` in the formatting string, then the logger would only print the results of the `throwable.toString()` which wouldn't include the stack trace and which would defeat the purpose of this very fix.

This is how you usually want to log all your exceptions, unless they are too noisy like this one, where we have to do it only occasionally using the RateLimiter.

So this code is written by design and there's not a bug w.r.t. the `e`/`{}`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkUtils.java
anthony-swirldslabs: Please see above.

kfa-aguda (approved):

Issue #14309, opened 2024-07-18
Drop KeyIndexType.SEQUENTIAL_INCREMENTING_LONGS support


artemananiev:
MerkleDb supports two key modes: long keys and generic keys. The corresponding values are captured in the `KeyIndexType` enum. Long mode provides significant performance boost, but unfortunately its use cases are very limited. Today this mode is only used in tests and benchmarks, while all Hedera virtual maps use the generic mode.

Here is a related comment in `OnDiskKeySerializer`:

```java
    public int getSerializedSize() {
        // We're going to use variable size keys, always. MerkleDB was designed with
        // fast paths if you knew you were using a Long as the key -- but we really
        // cannot use that. The problem manifests itself with state proofs. We wanted
        // to be able to say that we only store a long for the account ID, and implicitly
        // know the shard and realm as they could be system properties when the JVM
        // is started up. The problem is that the shard and realm MUST be part of the
        // serialized bytes, because they MUST be part of the state proof. Otherwise, you
        // would only have a proof of the accountNum, but not the shard and realm (well,
        // you may be able to reconstruct the shard and realm via the signature on the
        // state proof, if each shard has its own public key / ledger id, but that is
        // very cumbersome!).
        //
        // With protobuf serialization, we can encode shard/realm/num as a single long
        // in many cases, since "0" is the default for the fields, and therefore can
        // be skipped entirely. And anyway they are varint encoded, so for low values
        // of shard and realm (likely to be true for a very long time), they will
        // at most consume 7 bits each of the value.
        //
        // While that may be true, the KeySerializer requires either fixed size or variable
        // size, and protobuf would either be fixed > 8 bytes, or variable sized, and being
        // fixed but greater than 8 bytes doesn't help us in performance, so we will
        // have to use variable size always.
        return VARIABLE_DATA_SIZE;
    }
```

This ticket is to drop the long key mode support and leave the generic mode only.

PR #14323, opened 2024-07-19
fix: 14309: Drop KeyIndexType.SEQUENTIAL_INCREMENTING_LONGS support

artemananiev:
Fix summary:

* Removed `KeyIndexType` enum and all related code. All keys are handled as generic now
* Removed `getIndexType()` method from `KeySerializer`
* Removed all code related to long keys from `MerkleDbDataSource`
* Removed all metrics related to long keys
* Removed `VirtualLongKey` interface
* No migration code is needed: existing apps (Hedera and MTT) don't use long keys

Fixes: https://github.com/hashgraph/hedera-services/issues/14309


imalygin (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -245,6 +245,8 @@ public HalfDiskHashMap(
             numOfBuckets = Integer.highestOneBit(minimumBuckets) * 2;
             // we are new so no need for a loadedDataCallback
             loadedDataCallback = null;
+            // write metadata
+            writeMetadata(storeDir);
```
anthony-swirldslabs, line 249: This looks like a change compared to the previous behavior that is not explicitly listed in the PR description. Is this change intentional? If so, it would be nice to expand the comment at line 248 and indicate a reason why we want to (re-)write the metadata here.

It's a bit scary to see a write() method in a ctor that generally is responsible for loading/reading only. I'm not sure how well this would work in general, or, say, in unit and other tests if they happen to call this ctor.

artemananiev: Yes, this change is intentional and related to the fix. I would say, it's been a bug in HDHM from the very beginning, just not discovered until the current PR. In the tests, an empty data source is created. Previously, it used long keys, and therefore there was a LongList index, but now with generic keys a new HDHM is created instead. When such a data source is loaded back from disk, LongList was ready to handle missing files, while HDHM complained about missing metadata file, which resulted in test failures. The fix is to force writing the metadata file, when a new HDHM is created.

anthony-swirldslabs: Got it. Sounds good. Thanks!

anthony-swirldslabs (approved):

imalygin (approved):

litt3 (approved):

OlegMazurov (approved):
+1 for benchmarking code

PR #14324, opened 2024-07-19
feat: Add feature flag cookbook item

tinker-michaelj:
**Description**:
- Adds a cookbook item on minimizing the effects of feature flag rollouts on `@HapiTest` classes.

### DIFF
--- a/hedera-node/test-clients/SpecCookbook.md
+++ b/hedera-node/test-clients/SpecCookbook.md
@@ -50,7 +50,7 @@ static void beforeAll(@NonNull final TestLifecycle testLifecycle) {
   testLifecycle.overrideInClass(Map.of("feature.isEnabled", "true"));
 }
 ```
-(For a large feature there will multiple such test classes, of course, each with its own `@BeforeAll` method.)
+(For a large feature there will be multiple such test classes, of course, each with its own `@BeforeAll` method.)
 
 Second, if there are any new tests that require `feature.isEnabled=false`, place them in a test class with a
 `@BeforeAll` method that disables the feature and then restores the default after the test class finishes:
--- a/hedera-node/test-clients/SpecCookbook.md
+++ b/hedera-node/test-clients/SpecCookbook.md
@@ -15,6 +15,7 @@ The cookbook contains mostly concrete examples; but it is good to keep the princ
 
 ## Table of contents
 - [Patterns](#patterns)
+  * [DO minimize the impact of feature flag rollouts](#do-minimize-the-impact-of-feature-flag-rollouts)
   * [DO create template objects to enumerate families of related tests](#do-create-template-objects-to-enumerate-families-of-related-tests)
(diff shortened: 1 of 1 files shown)


vtronkov (approved):

hedera-node/test-clients/SpecCookbook.md
```diff
@@ -28,6 +29,44 @@ The cookbook contains mostly concrete examples; but it is good to keep the princ
...
+```java
+@BeforeAll
+static void beforeAll(@NonNull final TestLifecycle testLifecycle) {
+  testLifecycle.overrideInClass(Map.of("feature.isEnabled", "false"));
+}
+```
+
+Third, consider existing tests that require `feature.isEnabled=false` to pass. By default, move all such tests into an
+appropriate `@HapiTestLifecycle` test class as above. The only exception is if there is just one such test; in that
+case, use `@LeakyHapiTest(overrides = {"feature.isEnabled"})` and start the test by overriding the flag to `false`. No
```
david-bakin-sl, line 66: Is the reason for this exception simply that it is lighter weight in terms of required syntax?  If so, consider whether we should have this exception _or not_?  Having only _one_ pattern might be easier overall, even if the occasional unitary test has to have a few extra lines of code.

hedera-node/test-clients/SpecCookbook.md
```diff
@@ -28,6 +29,44 @@ The cookbook contains mostly concrete examples; but it is good to keep the princ
...
+@BeforeAll
+static void beforeAll(@NonNull final TestLifecycle testLifecycle) {
+  testLifecycle.overrideInClass(Map.of("feature.isEnabled", "false"));
+}
+```
+
+Third, consider existing tests that require `feature.isEnabled=false` to pass. By default, move all such tests into an
+appropriate `@HapiTestLifecycle` test class as above. The only exception is if there is just one such test; in that
+case, use `@LeakyHapiTest(overrides = {"feature.isEnabled"})` and start the test by overriding the flag to `false`. No 
+matter how many such tests exist, carefully evaluate for each one why it fails with `feature.isEnabled=true`. If there
```
david-bakin-sl, line 67: Is it worth mentioning _how_ to find out which tests fail this way?  Or is it obvious that you just set the config value to `true` and see what falls out?

david-bakin-sl (dismissed):
This test cookbook _is getting better and better_ and it's so **great** we have it.  TYVM for keeping it going!

LGTM!

hedera-node/test-clients/SpecCookbook.md
tinker-michaelj: Agreed, removed this exception.

hedera-node/test-clients/SpecCookbook.md
tinker-michaelj: Good point, noted the easiest thing is to set default to `true` and open a draft PR, see what fails.

Neeharika-Sompalli (approved):
🙏

david-bakin-sl (approved):
LGTM!

PR #14326, opened 2024-07-19
docs: add evm-transaction-response-codes.md

### LINKED ISSUE #14207: list of potential response codes that an EVM transaction could have in receipt
Is there a potential list of on-chain errors that could be returned by the “receipts”?
For example, 
we already know “Insufficient_tx_fee” can be returned by the receipt. Please let us know if any of these errors or more could be returned through receipt other than the contract revert.

MrValioBg:
**Description**:
This PR adds a doc with possible response codes for failed smart-contract EVM transactions that could be propagated to the top-level status.

**Related issue(s)**:

Fixes #14207 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
+++ b/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
@@ -1,12 +1,16 @@
 # EVM Transaction Response Codes in Hedera
 
-When executing an EVM transaction to a smart contract using `contractCall` in Hedera, different response codes may be present for the top-level status after the transaction is finished. These responses are found in the transaction receipt.
+When executing an EVM transaction to a smart contract using `contractCall` in Hedera and it fails, a response code is present in the transaction receipt to indicate the top-level status.
 
-The errors listed below could occur for a `contractCall` which creates child transactions, and the revertReason or haltReason which is set when operation fails is propagated to the top-level call.
-There are other errors that could occur in the childs, but the ones listed below are the specific ones that are propagated to the top-level call, which currently happens [here](https://github.com/hashgraph/hedera-services/blob/774ed309600e4b4acd9e1ca72fcd87d354c8b9ff/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java#L141-L169), in the `HederaEvmTransactionResult` in the `finalStatus()` method.
-Other revert/halt reasons from the child should default to CONTRACT_REVERT_EXECUTED for the top-level status.
+The default status for a failing `contractCall` is CONTRACT_REVERT_EXECUTED, which is included in the receipt. Child transactions might have more specific errors.
+In certain cases, the top-level status may differ from CONTRACT_REVERT_EXECUTED. The errors listed below are those special cases that can occur during the execution of a transaction and will be propagated as the top-level status.
+
+The handling of those cases currently happens [here](https://github.com/hashgraph/hedera-services/blob/774ed309600e4b4acd9e1ca72fcd87d354c8b9ff/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java#L141-L169), in the `HederaEvmTransactionResult` in the `finalStatus()` method.
+
 
 ## Possible Propagated Response Codes
+
+Note: The description for the statuses is copied from [protobuf repo](https://github.com/hashgraph/hedera-protobufs/blob/main/services/response_code.proto)
 
 - **INVALID_SOLIDITY_ADDRESS**
     -  The system is not able to find the user’s Solidity address.
--- a/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
+++ b/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
@@ -1,7 +1,7 @@
 # EVM Transaction Response Codes in Hedera
 
 When executing a transaction processed by the EVM (i.e., EVM Transaction) using `ContractCreate`, `ContractCall`, `EthereumTransaction`, and `ContractCallLocal` in Hedera, different response codes may be present for the top-level status after the transaction is finished. 
-These responses are found in the transaction receipt. This document's goal is to list the possible EVM transaction type-specific status codes for failure.
+These responses are found in the transaction receipt. This document's goal is to list the possible EVM transaction type-specific status codes for failure (i.e. different from SUCCESS).
 
 ## Propagated EVM Transaction Response Codes
 These response codes can appear for `ContractCall`, `ContractCreate`, and `EthereumTransaction`.
--- a/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
+++ b/hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
@@ -1,59 +1,118 @@
 # EVM Transaction Response Codes in Hedera
 
-When executing an EVM transaction to a smart contract using `contractCall` in Hedera and it fails, a response code is present in the transaction receipt to indicate the top-level status.
+When executing a transaction processed by the EVM (i.e., EVM Transaction) using `ContractCreate`, `ContractCall`, `EthereumTransaction`, and `ContractCallLocal` in Hedera, different response codes may be present for the top-level status after the transaction is finished. 
+These responses are found in the transaction receipt. This document's goal is to list the possible EVM transaction type-specific status codes for failure.
 
-The default status for a failing `contractCall` is CONTRACT_REVERT_EXECUTED, which is included in the receipt. Child transactions might have more specific errors.
-In certain cases, the top-level status may differ from CONTRACT_REVERT_EXECUTED. The errors listed below are those special cases that can occur during the execution of a transaction and will be propagated as the top-level status.
-
-The handling of those cases currently happens [here](https://github.com/hashgraph/hedera-services/blob/774ed309600e4b4acd9e1ca72fcd87d354c8b9ff/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java#L141-L169), in the `HederaEvmTransactionResult` in the `finalStatus()` method.
-
-
-## Possible Propagated Response Codes
-
-Note: The description for the statuses is copied from [protobuf repo](https://github.com/hashgraph/hedera-protobufs/blob/main/services/response_code.proto)
-
-- **INVALID_SOLIDITY_ADDRESS**
-    -  The system is not able to find the user’s Solidity address.
-
-- **INVALID_ALIAS_KEY**
-    -  An alias used in a CryptoTransfer transaction is not the serialization of a primitive Key message. This means a Key with a single Ed25519 or ECDSA(secp256k1) public key and no unknown protobuf fields.
-
-- **INVALID_SIGNATURE**
-    -  The transaction signature is not valid.
-
-- **MAX_ENTITIES_IN_PRICE_REGIME_HAVE_BEEN_CREATED**
-    -  The maximum number of entities allowed in the current price regime have been created.
-
-- **INSUFFICIENT_GAS**
-    -  Not enough gas was supplied to execute the transaction.
-
-- **LOCAL_CALL_MODIFICATION_EXCEPTION**
-    -  Local execution (query) is requested for a function which changes state.
-
-- **MAX_CHILD_RECORDS_EXCEEDED**
-    -  A contract transaction tried to use more than the allowed number of child records, either through system contract records or internal contract creations.
-
-- **INVALID_CONTRACT_ID**
-    -  The contract ID is invalid or does not exist.
-
-- **INVALID_FEE_SUBMITTED**
-    -  The fee submitted is invalid.
-
-- **MAX_CONTRACT_STORAGE_EXCEEDED**
-    -  Contract permanent storage exceeded the currently allowable limit.
-
-- **MAX_STORAGE_IN_PRICE_REGIME_HAS_BEEN_USED**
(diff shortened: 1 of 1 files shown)


MrValioBg:
Note, the specification is only for contractCall operation. 
For contractCreate operation, there could be more cases like ERROR_DECODING_BYTESTRING or CONTRACT_BYTECODE_EMPTY for example.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,59 @@
+# EVM Transaction Response Codes in Hedera
+
+When executing an EVM transaction to a smart contract using `contractCall` in Hedera and it fails, a response code is present in the transaction receipt to indicate the top-level status.
```
MrValioBg, line 3: Note, the specification is only for contractCall operation. 
For contractCreate operation, there could be more cases like ERROR_DECODING_BYTESTRING or CONTRACT_BYTECODE_EMPTY for example.

The [issue](https://github.com/hashgraph/hedera-services/issues/14207) mentions "EVM Transaction", but It seems that it probably refers just to contractCall. Any thoughts?

david-bakin-sl: I think it would be valuable - and meet the intent of the issue too - to include contract create.

It would be desirable but probably not necessary to distinguish codes that could be returned by one or the other or both.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,59 @@
...
+- **INVALID_SOLIDITY_ADDRESS**
+    -  The system is not able to find the user’s Solidity address.
+
+- **INVALID_ALIAS_KEY**
+    -  An alias used in a CryptoTransfer transaction is not the serialization of a primitive Key message. This means a Key with a single Ed25519 or ECDSA(secp256k1) public key and no unknown protobuf fields.
+
+- **INVALID_SIGNATURE**
+    -  The transaction signature is not valid.
+
+- **MAX_ENTITIES_IN_PRICE_REGIME_HAVE_BEEN_CREATED**
```
david-bakin-sl, line 24: Yikes!  Might (maybe) be worth segregating - or marking with an asterisk - a couple of these which are real anomalies - nothing you'd actually expect to get.  This particular one indicates we've got too many total contracts in the entire system (I think).  If someone hits this, well, there's nothing they can do about it except wait until the network is fixed.

Similar for `MAX_STORAGE_IN_PRICE_REGIME_HAS_BEEN_USED`.  

`MAX_CONTRACT_STORAGE_EXCEEDED` is almost but not quite the same - that does apply to an individual contract, not the entire system. The configured limit is huge and no contract is likely to reach it in normal operation anytime soon, but you can hit it you really apply yourself.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **INSUFFICIENT_TX_FEE**
+    -  The fee provided in the transaction is insufficient for this type of transaction.
+
+- **INSUFFICIENT_PAYER_BALANCE**
+    -  The transaction payer could not afford a custom fee.
+
+- **CONTRACT_EXECUTION_EXCEPTION**
+    -  This code is used for any contract execution-related error not handled by specific error codes listed above.
+    - *Example:* An error in Besu due to differences in supported Solidity versions. For instance, if Besu supports up to 0.8.23 and the contract is compiled with a version higher than 0.8.23 containing opcode changes, this error would be thrown.
```
david-bakin-sl, line 56: This would be better expressed in terms of EVM versions, e.g., Cancun vs Shanghai.

david-bakin-sl (commented):
Some comments but this is definitely going in the right direction, TY

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
Nana-EC: EVM Transaction is any transaction that is processed by the EVM on Hedera.
So that's `ContractCreate`, `ContractCall`, `EthereumTransaction` and even `ContractCallLocal`.
We should indeed distinguish since the request was to support clients to customize their code for support.
As such a client that always uses EVM tools that end up submitting an `EthereumTransaction` should only be concerned with codes applicable to that flow.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,59 @@
+# EVM Transaction Response Codes in Hedera
+
+When executing an EVM transaction to a smart contract using `contractCall` in Hedera and it fails, a response code is present in the transaction receipt to indicate the top-level status.
+
+The default status for a failing `contractCall` is CONTRACT_REVERT_EXECUTED, which is included in the receipt. Child transactions might have more specific errors.
+In certain cases, the top-level status may differ from CONTRACT_REVERT_EXECUTED. The errors listed below are those special cases that can occur during the execution of a transaction and will be propagated as the top-level status.
+
+The handling of those cases currently happens [here](https://github.com/hashgraph/hedera-services/blob/774ed309600e4b4acd9e1ca72fcd87d354c8b9ff/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransactionResult.java#L141-L169), in the `HederaEvmTransactionResult` in the `finalStatus()` method.
```
Nana-EC, line 8: Code lines will change so I don't think it's necessary to note them. At the most you can do the `Class.function()` you have with a summarized mention of the cause.

Nana-EC (commented):
good stuff, some suggestions

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: I agree. Will edit that.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: I've factored that in the latest update. Please re-check :)

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Yes, agreed. I've included those in a "separate" section and added some clarifying comments.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Done, I've adjusted the example

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,120 @@
...
+- **WRONG_NONCE**
+  - This transaction specified an `ethereumNonce` that is not the current `ethereumNonce` of the account.
+
+- **WRONG_CHAIN_ID**
+  - `EthereumTransaction` was signed against a chain ID that this network does not support.
+
+- **INVALID_ETHEREUM_TRANSACTION**
+  - The Ethereum transaction either failed parsing or failed signature validation, or some other `EthereumTransaction` error not covered by another response code.
+
+Note: The description for the statuses is copied from [protobuf repo](https://github.com/hashgraph/hedera-protobufs/blob/main/services/response_code.proto)
```
david-bakin-sl, line 120: I would move this line up in the file - near the top really.  (Say, after L10)

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,120 @@
+# EVM Transaction Response Codes in Hedera
+
+When executing a transaction processed by the EVM (i.e., EVM Transaction) using `ContractCreate`, `ContractCall`, `EthereumTransaction`, and `ContractCallLocal` in Hedera, different response codes may be present for the top-level status after the transaction is finished. 
+These responses are found in the transaction receipt. This document's goal is to list the possible EVM transaction type-specific status codes for failure (i.e. different from SUCCESS).
+
+## Propagated EVM Transaction Response Codes
+These response codes can appear for `ContractCall`, `ContractCreate`, and `EthereumTransaction`.
+It is possible for `ContractCreate` and `ContractCall` to create child transactions, and if one of them fails with a special revert or halt reason, the status is propagated to the top-level status of the transaction.
+`ContractCall` creates child transactions by executing logic that results in another call to a function/contract.
+`ContractCreate` creates child transactions by having logic containing calls in the constructor.
```
david-bakin-sl, line 10: Do you want to make these two lines part of a bulleted list?  And add a line for `EthereumTransaction`.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,120 @@
...
+
+- **CONTRACT_REVERT_EXECUTED**
+  - The default error for contract execution being reverted. This also happens when the REVERT opcode is executed in the contract.
+
+- **INVALID_SOLIDITY_ADDRESS**
```
david-bakin-sl, line 15: This is an example of where the description in the response code's enum protobuf is not as good as it could be.  In this case even the response code _name_ is not as good as it could be.

I suggest finishing this doc "as is" and (getting it merged) and maybe later if there's interest coming back to it and fixing up some of these explanations with what's really going on.  (And fixing up the protobuf comments as well.)

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,120 @@
...
+
+Those are some possible, but very unlikely to happen, top-level status codes. 
+As for MAX_STORAGE_IN_PRICE_REGIME_HAS_BEEN_USED and MAX_ENTITIES_IN_PRICE_REGIME_HAVE_BEEN_CREATED, they indicate too many contracts/entities in the entire system, so the caller can't do anything except wait until the network is fixed.
+MAX_CONTRACT_STORAGE_EXCEEDED applies to an individual contract but still is very hard to reach under normal operation.
+
+- **MAX_CONTRACT_STORAGE_EXCEEDED**
+  - Contract permanent storage exceeded the currently allowable limit.
+
+- **MAX_STORAGE_IN_PRICE_REGIME_HAS_BEEN_USED**
+  - All contract storage allocated to the current price regime has been consumed.
```
david-bakin-sl, line 57: (This one and the next one are examples of a response code explanation that doesn't really explain much.  What's a price regime anyway?  That's why you needed the sentences above to explain them.  This is in reference to my previous comment about fixing up the protobuf enum decl later.)

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
```diff
@@ -0,0 +1,120 @@
...
+- **AUTORENEW_DURATION_NOT_IN_RANGE**
+  - The duration is not a subset of `[MINIMUM_AUTORENEW_DURATION, MAXIMUM_AUTORENEW_DURATION]`.
+
+- **CONTRACT_NEGATIVE_GAS**
+  - Negative gas was offered in a smart contract call.
+
+- **MAX_GAS_LIMIT_EXCEEDED**
+  - Gas exceeded the currently allowable gas limit per transaction.
+
+- **INVALID_MAX_AUTO_ASSOCIATIONS**
```
david-bakin-sl, line 84: But now there's a well-defined upper limit, right?  Is it worth putting that upper bound here (if that's right)?

david-bakin-sl (approved):
Very slight changes suggested, but basically: LGTM.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Yeah, I agree. Some of the other descriptions are also slightly refined for better readability, so in the future, it could be worth revisiting the descriptions.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Done :)

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Yep, I think that would make it more clear. Done.

hedera-node/docs/design/services/smart-contract-service/evm-transaction-response-codes.md
MrValioBg: Currently, INVALID_MAX_AUTO_ASSOCIATIONS also handles the case where we have maxAutomaticTokenAssociations as less than -1.

But there is also REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT which defines the upper limit of 5000 (if you want to go above that, you must set maxAutoAssociations to -1).

I think since we put INVALID_MAX_AUTO_ASSOCIATIONS, it would indeed be useful to also add the other one, for clarity.

I also added NEGATIVE_ALLOWANCE_AMOUNT (which is a little weird because it represents two different contexts; in this case, for Ethereum transactions, it is used for the maxGasAllowance field, but it is also used for the CryptoApproveAllowance operation).

david-bakin-sl (approved):
LGTM

stoyanov-st (approved):
Good job!
LGTM.

Neeharika-Sompalli (approved):
LGTM! Thanks @MrValioBg

PR #14279, opened 2024-07-17
feat: Add TokenAirdropHandler

JivkoKelchev:
**Note**: This PR is a separated part of https://github.com/hashgraph/hedera-services/pull/13830

**Description**:
Add implementation of the TokenAirdropHandler.

**Related issue(s)**:

Fixes #13699
Fixes #13700
Fixes #13701
Fixes #13703

**Notes for reviewer**:
- Adding preHandle, pureCheck, handle and calculateFees logic

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #14335, opened 2024-07-19
feat: implement token claim airdrop pre handle pure checks

### LINKED ISSUE #14325: Implement `preHandle()` and `pureChecks()` methods in `TokenClaimAirdropHandler`

povolev15:
implement token claim airdrop pre handle pure checks

**Related issue(s)**:

Fixes #14325

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
@@ -23,6 +23,7 @@ import static java.util.Objects.requireNonNull;
 
 import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.HederaFunctionality;
+import com.hedera.hapi.node.base.Key;
 import com.hedera.hapi.node.base.PendingAirdropId;
 import com.hedera.hapi.node.state.token.Account;
 import com.hedera.hapi.node.transaction.TransactionBody;
@@ -56,18 +57,18 @@ public class TokenClaimAirdropHandler implements TransactionHandler {
     public void preHandle(@NonNull PreHandleContext context) throws PreCheckException {
         requireNonNull(context);
         final var op = context.body().tokenClaimAirdrop();
+        requireNonNull(op);
         final var pendingAirdrops = op.pendingAirdrops();
 
-        if (pendingAirdrops.isEmpty()) {
-            throw new PreCheckException(EMPTY_PENDING_AIRDROP_ID_LIST);
-        }
         final var accountStore = context.createStore(ReadableAccountStore.class);
 
         for (final var pendingAirdrop : pendingAirdrops) {
             AccountID receiverId = pendingAirdrop.receiverIdOrThrow();
             Account account = accountStore.getAccountById(receiverId);
             requireNonNull(account);
-            context.requireKey(account.key());
+            Key key = account.key();
+            requireNonNull(key);
+            context.requireKey(key);
         }
     }
 
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
@@ -16,8 +16,17 @@
 
 package com.hedera.node.app.service.token.impl.handlers;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.EMPTY_PENDING_AIRDROP_ID_LIST;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.MAX_PENDING_AIRDROP_ID_EXCEEDED;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.PENDING_NFT_AIRDROP_ALREADY_EXISTS;
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.HederaFunctionality;
+import com.hedera.hapi.node.base.PendingAirdropId;
+import com.hedera.hapi.node.state.token.Account;
 import com.hedera.hapi.node.transaction.TransactionBody;
+import com.hedera.node.app.service.token.ReadableAccountStore;
 import com.hedera.node.app.spi.fees.FeeContext;
 import com.hedera.node.app.spi.fees.Fees;
 import com.hedera.node.app.spi.workflows.HandleContext;
@@ -26,6 +35,8 @@ import com.hedera.node.app.spi.workflows.PreCheckException;
 import com.hedera.node.app.spi.workflows.PreHandleContext;
 import com.hedera.node.app.spi.workflows.TransactionHandler;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.List;
+import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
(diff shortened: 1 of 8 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +53,46 @@ public TokenClaimAirdropHandler() {
     }
 
     @Override
-    public void preHandle(@NonNull PreHandleContext context) throws PreCheckException {}
+    public void preHandle(@NonNull PreHandleContext context) throws PreCheckException {
+        requireNonNull(context);
+        final var op = context.body().tokenClaimAirdrop();
```
david-bakin-sl, line 58: I know this got here via a dispatch so `op` is gonna be good but still put the `requireNonNull` in.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +53,46 @@ public TokenClaimAirdropHandler() {
...
+        if (pendingAirdrops.isEmpty()) {
+            throw new PreCheckException(EMPTY_PENDING_AIRDROP_ID_LIST);
+        }
+        final var accountStore = context.createStore(ReadableAccountStore.class);
+
+        for (final var pendingAirdrop : pendingAirdrops) {
+            AccountID receiverId = pendingAirdrop.receiverIdOrThrow();
+            Account account = accountStore.getAccountById(receiverId);
+            requireNonNull(account);
+            context.requireKey(account.key());
```
david-bakin-sl, line 70: IntelliJ tells me `account.key()` might return `null`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +53,46 @@ public TokenClaimAirdropHandler() {
...
+        }
+
+        if (pendingAirdrops.size() > 10) {
+            throw new PreCheckException(MAX_PENDING_AIRDROP_ID_EXCEEDED);
+        }
+
+        final int numAirdrops = pendingAirdrops.size();
+        final Set<PendingAirdropId> uniqueAirdrops = Set.copyOf(pendingAirdrops);
+        if (numAirdrops != uniqueAirdrops.size()) {
+            throw new PreCheckException(PENDING_NFT_AIRDROP_ALREADY_EXISTS);
```
david-bakin-sl, line 93: The protobuf decl of this response code doesn't have a Javadoc, unfortunately.  If it did it would be nice if it said that this response code can also mean "duplicate nft airdrops specified in same tx".

david-bakin-sl (changes requested):
Needs one fix - _code_ looks ok, I'll come back later to look at the tests.

hedera-node/hedera-token-service-impl/src/main/java/module-info.java
```diff
@@ -12,6 +12,7 @@
     requires transitive com.swirlds.state.api;
     requires transitive com.hedera.pbj.runtime;
     requires transitive dagger;
+    requires transitive java.annotation;
```
david-bakin-sl, line 15: This line is _wrong_ - should be omitted!  The actual problem is that `CryptoTransferValidationHelper.java` imports `javax.annotation. Nullable` instead of what it should: `edu.umd.cs.findbugs.annotations.Nullable`.  Please fix _that_ instead!  (It's in the branch you're merging into, not from this PR.)

david-bakin-sl (changes requested):
This is the required fix I mentioned above - I goofed and didn't save it before finishing that review up there, sorry.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +53,46 @@ public TokenClaimAirdropHandler() {
...
     @Override
-    public void preHandle(@NonNull PreHandleContext context) throws PreCheckException {}
+    public void preHandle(@NonNull PreHandleContext context) throws PreCheckException {
+        requireNonNull(context);
+        final var op = context.body().tokenClaimAirdrop();
+        final var pendingAirdrops = op.pendingAirdrops();
+
+        if (pendingAirdrops.isEmpty()) {
+            throw new PreCheckException(EMPTY_PENDING_AIRDROP_ID_LIST);
+        }
```
vtronkov, line 63: Because we want only computing signing requirements in the `preHadle()`(per https://github.com/hashgraph/hedera-services/pull/14279#discussion_r1685837369) and we have this check in `pureCheck()`, should we remove this check from here?

david-bakin-sl (approved):
LGTM, TY!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
vtronkov: @povolev15 I wonder if this check is needed at all. On the `TokenAirdropTransaction` we are throwing `PENDING_NFT_AIRDROP_ALREADY_EXISTS` when we try to create a new airdrop but it already exists in the store/state not if we have duplicated airdrops in a single transaction.

In the case of a claim, if we try to **claim** an airdrop twice in a single transaction I would think that the first **claim** will be successful and the second one will throw that the airdrop doesn't exist - which sounds logical to me. What do you think?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +54,46 @@ public TokenClaimAirdropHandler() {
...
+            requireNonNull(key);
+            context.requireKey(key);
+        }
+    }
 
     @Override
-    public void pureChecks(@NonNull TransactionBody txn) throws PreCheckException {}
+    public void pureChecks(@NonNull TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn);
+
```
kimbor, line 78: Should also check that sender_id is not null. I'm not sure if it's necessary to check that an account exists for sender_id. What happens if someone sends an airdrop and then their account goes away? Is the airdrop still able to be claimed?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -42,10 +54,46 @@ public TokenClaimAirdropHandler() {
...
+        requireNonNull(context);
+        final var op = context.body().tokenClaimAirdrop();
+        requireNonNull(op);
+        final var pendingAirdrops = op.pendingAirdrops();
+
+        final var accountStore = context.createStore(ReadableAccountStore.class);
+
+        for (final var pendingAirdrop : pendingAirdrops) {
+            AccountID receiverId = pendingAirdrop.receiverIdOrThrow();
+            Account account = accountStore.getAccountById(receiverId);
```
kimbor, line 67: Should also check that sender_id is not null. I'm not sure if it's necessary to check that an account exists for sender_id. What happens if someone sends an airdrop and then their account goes away? Is the airdrop still able to be claimed?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
vtronkov: For the `AirdropTransaction` we have the same check(https://github.com/hashgraph/hedera-services/pull/14279). You can see it in `CryptoTransferValidator.validateNftTransfers`:
```
validateFalsePreCheck(
        !nftIds.isEmpty() && nftIds.contains(nftTransfer.serialNumber()), INVALID_ACCOUNT_AMOUNTS);
```
And we are throwing `INVALID_ACCOUNT_AMOUNTS`. In my opinion, this is not the most precise error code but we are reusing the validations from CryptoTransfer. So I think we should be consistent with that and update the response code here from `PENDING_NFT_AIRDROP_ALREADY_EXISTS` to `INVALID_ACCOUNT_AMOUNTS`. What do you think?

PR #14337, opened 2024-07-22
chore: fix Javadoc warnings in 'test-clients'

jjohannes:
**Description**:
Fixes warnings given by the `javadoc` tool. 

**Related issue(s)**:

In preparation for #13645

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/AbstractEmbeddedHedera.java
```diff
@@ -19,7 +19,6 @@
 import static com.hedera.hapi.util.HapiUtils.parseAccount;
 import static com.hedera.node.app.hapi.utils.CommonPbjConverters.fromPbj;
 import static com.hedera.services.bdd.junit.hedera.ExternalPath.ADDRESS_BOOK;
-import static com.swirlds.platform.config.legacy.LegacyConfigPropertiesLoader.loadConfigFile;
```
jjohannes, line 22: This is a limitation of the `javadoc` tool. It complains about importing deprecated methods directly and there is no way to suppress that on an import. Hence, this is replaced by a normal import.

derektriley (approved):
LGTM.

PR #14358, opened 2024-07-23
chore: remove compiler lint exceptions and fix (addressbook)

jjohannes:
**Description**:

Saw that these are mostly unnecessary while testing something else. Might as well clean this up right aways as these are rather new modules. They should have a clean setup.

**Related issue(s)**:

#11569

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-addressbook-service-impl/src/main/java/module-info.java
```diff
@@ -12,8 +12,8 @@
     requires com.hedera.node.app.service.token;
     requires com.hedera.pbj.runtime;
     requires org.apache.logging.log4j;
+    requires static transitive java.compiler;
```
jjohannes, line 15: JavaCompile's `export` check reveals that this should be transitive, because the `@Generated` annotation (located in the `java.compiler` module) is used to annotate public API.

rbarkerSL (approved):

PR #14361, opened 2024-07-23
build: use the 'org.web3j.solidity' Gradle plugin to compile contracts

### LINKED ISSUE #11657: Gradle Build: Solidity support
Breakdown as discussed with @david-bakin-sl.

1. Generate `bin` and `json` from `sol` in https://github.com/hashgraph/hedera-services/tree/develop/hedera-node/test-clients/src/main/resource/contract/contracts - `bin` and `json` are no longer checked in; `sol` do not need to be in resources (not needed at runtime); may use `org.web3j.solidity` Gradle plugin.
2. Get **System Contracts** `sol` from from https://github.com/hashgraph/hedera-smart-contracts instead of maintaining a copy in hedera-services.
3. Only **System Contracts**: generate Java code based on them in `hedera-smart-contract-service-impl`

jjohannes:
**Description**:

See #11657

**Related issue(s)**:

Resolves #11657

**Notes for reviewer**:

**Checklist**


Nana-EC:
@jjohannes any progress on this as 0.55 is complete, will this be in 0.56?
Thanks

PR #14374, opened 2024-07-24
feat: HIP-904: Implement TokenClaimAirdrop transaction

vtronkov:
**Description**:
Add TokenClaimAirdrop basic implementation

**Notes for reviewer**:
- Add configuration
- Add throttles.json values
- Add TokenClaimAirdrop empty shell - we always throw a unsupported exception until we merge the actual implementation of the feature flag


MrValioBg (approved):
LGTM! Thank you very much!

Just a nit - consider re-checking function params and making them final / @NonNull (We could do it in a following PR with the implementation.)

Neeharika-Sompalli (dismissed):
LGTM!

hedera-node/test-clients/src/main/resource/testSystemFiles/throttles-dev.json
```diff
@@ -69,7 +69,8 @@
 			"TokenFreezeAccount",
 			"TokenUnfreezeAccount",
 			"TokenGrantKycToAccount",
-			"TokenRevokeKycFromAccount"
+			"TokenRevokeKycFromAccount", 
+      "TokenClaimAirdrop"
```
stoyanov-st, line 73: Could you please adjust the indentation here?

hedera-node/hedera-file-service-impl/src/main/resources/genesis/throttles-dev.json
```diff
@@ -10,6 +10,7 @@
 			"CryptoCreate",
 			"CryptoTransfer",
             "TokenAirdrop",
+				"TokenClaimAirdrop",
```
stoyanov-st, line 13: Also adjust here.

stoyanov-st (approved):
LGTM 🚀 , just a small nit for the json indentations

hedera-node/hedera-file-service-impl/src/main/resources/genesis/throttles-dev.json
vtronkov: Fixed 👍

hedera-node/test-clients/src/main/resource/testSystemFiles/throttles-dev.json
vtronkov: Not sure what's going on but in Github it looks weird but when I open it in IntelliJ it looks fine 🤷

stoyanov-st: Is uses spaces instead of tabs

vtronkov: Now it should be ok 😄

Neeharika-Sompalli (approved):
LGTM! Thanks @vtronkov

nathanklick (approved):
Approved as to form only. No assurances have been made as the substance of these changes.

PR #14376, opened 2024-07-24
feat: implement token claim airdrop pre handle pure checks

### LINKED ISSUE #14325: Implement `preHandle()` and `pureChecks()` methods in `TokenClaimAirdropHandler`

vtronkov:
implement token claim airdrop pre handle pure checks

**Related issue(s)**:

Fixes #14325


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -44,10 +55,46 @@ public TokenClaimAirdropHandler() {
...
+        final var pendingAirdrops = op.pendingAirdrops();
+
+        final var accountStore = context.createStore(ReadableAccountStore.class);
+
+        for (final var pendingAirdrop : pendingAirdrops) {
+            AccountID receiverId = pendingAirdrop.receiverIdOrThrow();
+            Account account = accountStore.getAccountById(receiverId);
+            requireNonNull(account);
+            Key key = account.key();
+            requireNonNull(key);
```
Neeharika-Sompalli, line 71: Instead of `requireNonNull` , we should use `validateTruePrecheck`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -44,10 +55,46 @@ public TokenClaimAirdropHandler() {
...
+
+        final var op = txn.tokenClaimAirdrop();
+        requireNonNull(op);
+
+        final List<PendingAirdropId> pendingAirdrops = op.pendingAirdrops();
+        if (pendingAirdrops.isEmpty()) {
+            throw new PreCheckException(EMPTY_PENDING_AIRDROP_ID_LIST);
+        }
+
+        if (pendingAirdrops.size() > 10) {
```
Neeharika-Sompalli, line 88: This 10 should be taken from config

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
vtronkov: Changed 👍 
But why do we need this change? Is it because we want to throw `PreCheckException` instead of `NullPointerException`? If so - do we want to change the rest of the `requireNonNull` from the same method?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
vtronkov: As with the cancel discussion we had - we are going to move that into the handle PR(https://github.com/hashgraph/hedera-services/pull/14563). I will remove it for now

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -44,10 +56,42 @@ public TokenClaimAirdropHandler() {
...
+        final List<PendingAirdropId> pendingAirdrops = op.pendingAirdrops();
+        if (pendingAirdrops.isEmpty()) {
+            throw new PreCheckException(EMPTY_PENDING_AIRDROP_ID_LIST);
+        }
+
+        final int numAirdrops = pendingAirdrops.size();
+        final Set<PendingAirdropId> uniqueAirdrops = Set.copyOf(pendingAirdrops);
+        if (numAirdrops != uniqueAirdrops.size()) {
+            throw new PreCheckException(PENDING_NFT_AIRDROP_ALREADY_EXISTS);
+        }
```
ibankov, line 93: Here we are checking if there are duplicate `PendingAirdropId`s in the transaction body, I think we should use another error message, as this one specifies NFTs and also we use that error for airdrops in sate.  I think we should use either `PENDING_AIRDROP_ID_REPEATED` or `INVALID_TRANSACTION_BODY` here.

vtronkov: @povolev15 since you worked on that can you check Ivan's comment

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
vtronkov: Here is the commit adding it to the handle PR: https://github.com/hashgraph/hedera-services/pull/14563/commits/07baa9bdad6206cc30bd520b43b7176068d298dc

PR #14379, opened 2024-07-24
feat: HIP-904 Implement TokenCancelAirdrop transaction

vtronkov:
**Description**:

With HIP-904 we introduce Token Airdrops.
We want to provide the option for an airdrop sender to cancel an airdrop that have been sent previously by them. That would be useful if the airdrop sender has made a mistake or the receiver hasn't claimed the airdrop for too long. This would be a way to remove ongoing fees for auto-renewal that would be charged from the airdrop sender account.

**Related issue(s)**:
Fixes https://github.com/hashgraph/hedera-services/issues/14027


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
+        requireNonNull(txn, "Transaction body cannot be null");
+        final var op = txn.tokenCancelAirdropOrThrow();
+
+        validateFalsePreCheck(op.pendingAirdrops().isEmpty(), EMPTY_PENDING_AIRDROP_ID_LIST);
+        validateFalsePreCheck(
+                op.pendingAirdrops().size() > MAX_ALLOWED_PENDING_AIRDROPS_TO_CANCEL, MAX_PENDING_AIRDROP_ID_EXCEEDED);
```
Neeharika-Sompalli, line 108: Max allowed pending airdrops should be a config

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public void handle(@NonNull final HandleContext context) throws HandleException {
+        var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
+        validateTrue(tokensConfig.cancelTokenAirdropEnabled(), NOT_SUPPORTED);
+    }
+
+    @NonNull
+    @Override
+    public Fees calculateFees(@NonNull final FeeContext feeContext) {
+        return Fees.FREE;
```
Neeharika-Sompalli, line 138: nit: We should probably throw UnsupportedOperationException when something is yet to be implemented. It is possible we forget implementing this, if we leave as Fees.FREE. Need  not be changed here, just for future PRs 🙂

Neeharika-Sompalli (approved):
LGTM! except few comments. Those comments can be fixed in next PR. Thanks @vtronkov

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
vtronkov: This is deleted and the `feeCalculation` is implemented in the related PR: https://github.com/hashgraph/hedera-services/pull/14381

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenCancelAirdropHandler.java
vtronkov: If we want it in the config we should move the check in the handle, right? Since in the pure-check, we don't have access to the config. I can update the related handle story with this(https://github.com/hashgraph/hedera-services/pull/14380)

Neeharika-Sompalli: yes, we should do it in handle and remove this check

vtronkov: Added in the handle - https://github.com/hashgraph/hedera-services/pull/14380/commits/b0071b437b3e8261ebe739d9fa241468184300f2 👍

PR #14399, opened 2024-07-25
feat: inline auto associations

### LINKED ISSUE #14284: Charge auto-association fees in `CryptoTransferHandler` pending lite dispatch
### Problem

Several key performance benchmarks rely on high-TPS auto-associations, which makes using dispatch [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/transfer/AssociateTokenRecipientsStep.java#L187) too expensive an implementation despite the other advantages of taking that approach.

### Solution

For now, switch to charging fees and (when they can be paid) directly creating an auto-association in the `CryptoTransferHandler`. 

For future release, see if a "light" form of dispatch could let us return to the dispatch implementation without changing the benchmarks.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14284


Neeharika-Sompalli (dismissed):
LGTM! Thanks @tinker-michaelj

Neeharika-Sompalli (approved):
LGTM!

david-bakin-sl (approved):
LGTM (also I reviewed previous version)

PR #14402, opened 2024-07-25
fix: consensus performance issue

### LINKED ISSUE #14378: Investigate consensus performance issue
The commit `9a52d674d0251fa7a21b590cf6dcfd9b1e62d449` that fixes an edge case in the node removal scenario seems to have introduced a performance issue in consensus when running on a large network.

This commit has been reverted until it has been triaged.

Investigate the cause of the slowdown, and re-apply the commit with the performance fixed.

lpetrovic05:
closes #14378


timo0 (dismissed):

litt3 (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
```diff
@@ -306,8 +306,23 @@ public List<ConsensusRound> addEvent(@NonNull final EventImpl event) {
...
             event.setRoundCreated(ConsensusConstants.ROUND_UNDEFINED);
-            checkInitJudges(event);
-            ConsensusRound consensusRound = calculateAndVote(event);
+            ConsensusRound consensusRound;
+
+            final boolean lastJudgeFound = checkInitJudges(event);
+
+            if (!noInitJudgesMissing()) {
+                // we should not do any calculations or voting until we have found all the init judges
+                return rounds;
```
edward-swirldslabs, line 315: It would be clearer to return `Collections.emptyList()` instead of `rounds`.  I was confused at first that rounds might not be empty.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ConsensusImpl.java
lpetrovic05: makes sense, done. I also moved the declaration of rounds further down

timo0 (approved):

litt3 (approved):

edward-swirldslabs (approved):

rbarkerSL (approved):

PR #14432, opened 2024-07-26
refactor: Use PBJ EventDescriptor 

### LINKED ISSUE #14007: Make EventDescriptor a wrapper
The current class `com.swirlds.platform.system.events.EventDescriptor` should become `EventDescriptorWrapper`. Since the event descriptor information is now stored in the `com.hedera.hapi.platform.event.EventDescriptor` PBJ record, the legacy class should wrap the PBJ class and provide wrappings for `Hash` and `NodeId`. It also should not be self-serializable anymore.

timo0:
Fix #14007


lpetrovic05 (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @timo0

PR #14455, opened 2024-07-29
feat: Airdrops with treasury and custom fee receivers

### LINKED ISSUE #14423: (AIRDROP - 07) Sender airdrops a token with a custom fee to be paid by the receiver to the fee collector account for the token fee

### LINKED ISSUE #14443: (AIRDROP - 09) Sender airdrops a token with a custom fee to be paid by the receiver to the token treasury account

### LINKED ISSUE #14479: (AIRDROP - 08) Sender airdrops a token with a custom fee to be paid by the receiver to the fee collector account of another token fee when all fee collector accounts are exempt

### LINKED ISSUE #14485: (AIRDROP - 29) Given a token with a custom fee that should be paid by the receiver when the airdrop is sent to a fee collector account of another token with the allCollectorsAreExempt=true then the receiver should not pay any custom fees

vtronkov:
**Description**:
- When the receiver is a treasury account or custom fee collector we want to skip the custom fee validation because these users are exempt from paying them.
- We also exempt users from paying if the allCollectorsExempt flag is true and the receiver is a fee collector to a token from the same type

**Related issue(s)**:

Fixes #14423 
Fixes #14443 
Fixes #14479
Fixes #14485


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -144,13 +147,24 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
             final var tokenId = xfers.tokenOrThrow();
-            validateTrue(tokenHasNoCustomFeesPaidByReceiver(tokenId, tokenStore), INVALID_TRANSACTION);
-
+            final var token = getIfUsable(tokenId, tokenStore);
             boolean shouldExecuteCryptoTransfer = false;
             final var transferListBuilder = TokenTransferList.newBuilder().token(tokenId);
 
             // process fungible token transfers if any
             if (!xfers.transfers().isEmpty()) {
+                for (var transfer : xfers.transfers()) {
```
david-bakin-sl, line 156: Consider moving this entire loop into a method - or the body of the loop (or both).  Would make it possible to _unit test_.  

Same at L222-227 for nfts.

This method `handle` is incredibly complex - far too complex for reasonable testing.  THis code here is a loop in a conditional in a loop.  140 lines in one method ... not a good idea.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -216,7 +237,7 @@ final var record =
                         tokenId,
                         xfers.nftTransfers(),
                         tokenRelStore,
-                        tokenStore,
+                    token,
```
david-bakin-sl: indenting is wrong

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -255,17 +276,49 @@ final var record =
...
     }
 
+    /**
+     * When we do an airdrop we need to check if there are custom fees that needs to be paid by the receiver. 
+     * If there are, an error is returned.
+     * However, there is an exception to this rule - if the receiver is the fee collector or the treasury account 
+     * they are exempt from paying the custom fees thus we don't need to check if there are custom fees.
+     * This method returns if the receiver is the fee collector or the treasury account.
+     */
+    private static boolean skipCustomFeeValidation(
```
david-bakin-sl, line 280: This method and the next make sense.  Good that they were broken out.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -255,17 +276,49 @@ final var record =
...
+        List<CustomFee> customFees,
+        AccountID receiverId,
+        AccountID treasuryId) {
+        var theReceiverIsTheTreasury = receiverId.equals(treasuryId);
+        return thereIsASingleFeeCollectorAndItIsTheReceiver(customFees, receiverId) || theReceiverIsTheTreasury;
```
david-bakin-sl, line 283: as a micro-optimization consider reversing the two sides of the `||`

vtronkov:
> please rerun spotless

Oops. Done :)

david-bakin-sl (commented):
Should have been a unit test that covered this aspect of `handle` that, now, would fail unless fixed.  Sorry I missed that earlier. Not too late to _add_ unit tests to `TokenAirdropHandlerTest`.  (Especially if refactored according to comments below.)

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
vtronkov: It turns out that we have a method that already does the same thing. I should've known - it's not a new functionality...
Anyway, I've deleted my methods and I'm calling `CustomFeeExemptions.isPayerExempt` instead.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -255,17 +269,32 @@ final var record =
...
+     */
+    private static boolean skipCustomFeeValidation(Token token, AccountID receiverId) {
+        for (var customFee : token.customFees()) {
+            if (CustomFeeExemptions.isPayerExempt(customFeeMetaFrom(token), customFee, receiverId)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
```
MrValioBg, line 287: Nit, but I think it would be easier to read as one-liner in this case.
```suggestion
private static boolean skipCustomFeeValidation(@NonNull final Token token, @NonNull final AccountID receiverId) {
    return token.customFees().stream()
        .anyMatch(customFee -> CustomFeeExemptions.isPayerExempt(customFeeMetaFrom(token), customFee, receiverId));
}
```

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
vtronkov: I agree, but since this method is implemented in https://github.com/hashgraph/hedera-services/pull/14279 - we will refactor it there 👍

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
vtronkov: Much better! Thanks

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
david-bakin-sl: ah, ok! thank you!

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -256,17 +270,29 @@ final var record =
         }
     }
 
+    /**
+     * When we do an airdrop we need to check if there are custom fees that needs to be paid by the receiver.
+     * If there are, an error is returned.
```
david-bakin-sl, line 275: "is returned _from the HAPI call_" - not from this method

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -256,17 +270,29 @@ final var record =
...
+     * However, there is an exception to this rule - if the receiver is the fee collector or the treasury account
+     * they are exempt from paying the custom fees thus we don't need to check if there are custom fees.
+     * This method returns if the receiver is the fee collector or the treasury account.
+     */
+    private static boolean skipCustomFeeValidation(Token token, AccountID receiverId) {
```
david-bakin-sl, line 280: or it could be called `isExemptFromCustomFees`?

david-bakin-sl (approved):
I'd still like to see unit tests for `handle()` - but I'll assume you'll do that when you refactor that method in the base feature branch.

MrValioBg (approved):
LGTM!

PR #14465, opened 2024-07-29
feat: Limit rate that PCES events are replayed

### LINKED ISSUE #14397: Limit rate of PCES replay
## Problem

- In testing with the health monitor enabled, we have observed instances where PCES replay reads events from disk too rapidly, causing an OOM
- The health monitor can't respond quickly enough to reliably prevent this from happening

## Proposed solution

- rather than only relying on the health monitor to limit the rate of PCES replay, we should put a rate limit on how quickly we are willing to read events from disk
- we need to configure the rate limit such that events are read from disk quickly enough to keep the transaction handle thread fully busy throughout replay, but not so quickly that the initial event influx can cause problems with node memory

litt3:
- closes #14397


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
```diff
@@ -67,19 +71,21 @@ public class PcesReplayer {
...
-            final @NonNull Time time,
+            final @NonNull PlatformContext context,
             final @NonNull StandardOutputWire<PlatformEvent> eventOutputWire,
             final @NonNull Runnable flushIntake,
             final @NonNull Runnable flushTransactionHandling,
             final @NonNull Supplier<ReservedSignedState> latestImmutableState,
             final @NonNull Supplier<Boolean> isSystemHealthy) {
 
-        this.time = Objects.requireNonNull(time);
+        this.time = context.getTime();
```
mxtartaglia-sl, line 81: Should we check the context is not null?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
```diff
@@ -67,19 +71,21 @@ public class PcesReplayer {
...
         this.eventOutputWire = Objects.requireNonNull(eventOutputWire);
         this.flushIntake = Objects.requireNonNull(flushIntake);
         this.flushTransactionHandling = Objects.requireNonNull(flushTransactionHandling);
         this.latestImmutableState = Objects.requireNonNull(latestImmutableState);
         this.isSystemHealthy = Objects.requireNonNull(isSystemHealthy);
+
+        this.config = context.getConfiguration().getConfigData(PcesConfig.class);
```
mxtartaglia-sl, line 88: This approach hides the coupling between the two pieces.
It's been used in many places and I know that this pattern simplifies changes to the code, but I found it challenging in other scenarios, e.g., (when working with the FileSystemManager) to test without knowing objects' direct dependencies and relationships.

mxtartaglia-sl (approved):
Approved with a Nit, and a personal opinion

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
litt3: the `requireNonNull` check simply throws a null pointer exception if the check fails. Since that's the exact behavior of calling a method on `null`, there isn't any explicit check required

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
litt3: Is there a different pattern you're advocating for?

I'm not particularly attached to this way of doing things, but we should probably reach internal team agreement on a preferred strategy if we want to change it.

mxtartaglia-sl: IMO, the higher the layer we can push the usage of the configs object, the better. And after a certain point, the config framework should no longer be used, and all values should have been retrieved. 

One problem I faced was that when testing something else and needing these configuration objects, It was not enough to mock the constructor parameters; I needed to go to the implementation and see how the config framework was being used to mock the config framework, too. That defeats the black box testing strategy. 

 This would also help to provide fast failures, as any necessary value that is not retrieved after a point would produce a failure instead of waiting for it to be used to fail. Stretching the argument a little more, the more the configuration framework is used inside the inner layers, the more difficult it would be to replace it in the future if needed.
 
 But I am stating this from a personal point of view, given that I don't know if there is a place that can become the boundary where all the configs are retrieved.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
mxtartaglia-sl: That is why we should not call `Objects.requireNonNull` without a proper message; it is hard to know if it is a validation or an uncaught error. 
In any case, as you said, both are unexpected scenarios with similar stacktraces.

lpetrovic05:
I feel like this is a problem with the health monitor that should be addressed, we could theoretically have a similar problem with events from gossip

litt3:
> I feel like this is a problem with the health monitor that should be addressed, we could theoretically have a similar problem with events from gossip

The problem here is that an influx of hundreds of thousands of events can enter the system before the health monitor notices full queues. There's no way to *guarantee* that this can't happen from gossip, but it's not really conceivable.

I don't think the lack of a strong guarantee is a *problem* with the health monitor that can be fixed- it's inherent to it. If it's not tolerable, then an entirely different solution is required.

lpetrovic05:
> > I feel like this is a problem with the health monitor that should be addressed, we could theoretically have a similar problem with events from gossip
> 
> The problem here is that an influx of hundreds of thousands of events can enter the system before the health monitor notices full queues. There's no way to _guarantee_ that this can't happen from gossip, but it's not really conceivable.
> 
> I don't think the lack of a strong guarantee is a _problem_ with the health monitor that can be fixed- it's inherent to it. If it's not tolerable, then an entirely different solution is required.
> 
> Or do you mean there should be a rate limit imposed on events coming in from gossip?

I just think that this is hiding an inherent problem with the health monitor, which does not seem like a good path forward to me

litt3:
> > > I feel like this is a problem with the health monitor that should be addressed, we could theoretically have a similar problem with events from gossip
> > 
> > 
> > The problem here is that an influx of hundreds of thousands of events can enter the system before the health monitor notices full queues. There's no way to _guarantee_ that this can't happen from gossip, but it's not really conceivable.
> > I don't think the lack of a strong guarantee is a _problem_ with the health monitor that can be fixed- it's inherent to it. If it's not tolerable, then an entirely different solution is required.
> > Or do you mean there should be a rate limit imposed on events coming in from gossip?
> 
> I just think that this is hiding an inherent problem with the health monitor, which does not seem like a good path forward to me

Fair enough - I'm open to alternate suggestions

lpetrovic05:
> > > > I feel like this is a problem with the health monitor that should be addressed, we could theoretically have a similar problem with events from gossip
> > > 
> > > 
> > > The problem here is that an influx of hundreds of thousands of events can enter the system before the health monitor notices full queues. There's no way to _guarantee_ that this can't happen from gossip, but it's not really conceivable.
> > > I don't think the lack of a strong guarantee is a _problem_ with the health monitor that can be fixed- it's inherent to it. If it's not tolerable, then an entirely different solution is required.
> > > Or do you mean there should be a rate limit imposed on events coming in from gossip?
> > 
> > 
> > I just think that this is hiding an inherent problem with the health monitor, which does not seem like a good path forward to me
> 
> Fair enough - I'm open to alternate suggestions

I feel like some kind of blocking mechanism as a fall back for the event intake is needed. And gossip would probably never trigger this mechanism, but its there if the health monitor fails, and it should address the PCES issue

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
litt3: I agree, it is often painful to conclude which config values affect testing.

Perhaps you could start a conversation with the rest of the platform team, to decide if we need to shift our general strategy for using config values?

edward-swirldslabs (approved):

PR #14468, opened 2024-07-29
feat: instantiate `ContractService` with `AppContext` signature verifier

### LINKED ISSUE #14248: feat: Expose signature verification from app spi module to other services (e.g., smart contracts)
There needs to be a proper way for smart contracts service to get access to the app's ability to verify an arbitrary set of signatures, given an account, the data, and the signatures.  Currently no service has the ability to do that - the only thing exposed is the list of verified signatures (that signed the transaction).  But `isAuthorized{Raw}` requires it.  (`IsAuthorized` requires it for accounts with keys of arbitrary complexity and cryptography type, `isAuthorizedRaw` only for accounts with a single "simple" key.)

`isAuthorized` takes a protobuf containing (arbitrarily many) signature pairs (public key "prefix" and the signature bytes), as well as an account and the message (hash). ([HIP-632 link](https://hips.hedera.com/hip/hip-632#isauthorizedaddress-messagehash-signatureblob-function-usage) for 'isAuthorized`)
  * It doesn't need to pass the protobuf itself to services - it can transcode its contents (the signature pairs) to a regular Java type.

Currently `isAuthorizedRaw` does it by directly using a platform `Cryptography` crypto engine, but:
    * that API is deprecated-for-removal, and
    * for `isAuthorized` there's the need to grok complex keys
   
A fee calculator (or something) needs to be part of the API because the signature verification - encapsulated in the app service - has a wide range of cost (keys can be very complex) that the system contract will need to charge.

(Additionally, before a new fee schedule is available we should get a much closer estimate of gas costs for signature verification - but perhaps this should be a separate ticket.)

This is a prerequisite for implementing `isAuthorized` in the smart contracts service.

tinker-michaelj:
**Description**:
 - Closes #14248


hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -259,7 +263,14 @@ public Hedera(
                 () -> version.readableServicesVersion(),
                 () -> HapiUtils.toString(version.getHapiVersion()));
         fileServiceImpl = new FileServiceImpl();
-        contractServiceImpl = new ContractServiceImpl(instantSource);
+
+        final var appContext = new AppContextImpl(
+                instantSource,
+                new SpiSignatureVerifier(
```
Neeharika-Sompalli, line 269: Should this be called `AppSignatureVerifier` to keep similar naming ?

Neeharika-Sompalli (dismissed):
Makes sense to me, except for a naming comment. Thanks @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
tinker-michaelj: Makes sense @Neeharika-Sompalli , renamed this.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/signature/AppSignatureVerifier.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public boolean verifySignature(
+            @NonNull final Key key,
+            @NonNull final Bytes bytes,
+            @NonNull final SignatureMap signatureMap,
+            @Nullable final Function<Key, SimpleKeyVerification> simpleKeyVerifier) {
+        final Set<ExpandedSignaturePair> sigPairs = new HashSet<>();
+        signatureExpander.expand(key, signatureMap.sigPair(), sigPairs);
+        final var results = signatureVerifier.verify(bytes, sigPairs);
```
david-bakin-sl, line 67: If I start here at the call to `verify` it takes me to the implementation of `super`'s `verify` which is `SignatureVerifierImpl.verify` and there at L73 for an EC signature it calls `createPreparerForEC` which at L110 does a `keccak256DigestOf` the incoming bytes.  

But up above at `SignatureVerifier` the class comment says that for EC the incoming bytes is already the keccak hash of the payload.  (And that's the way the HAS system contract methods are defined, too: `isAuthorizedRaw` and `isAuthorized` are defined in HIP-632 to take a `messageHash`.

So is that 2nd keccak hash one too many?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/signature/AppSignatureVerifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.concurrent.TimeoutException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class AppSignatureVerifierTest {
+    private static final Key A_KEY = Key.newBuilder()
```
david-bakin-sl, line 52: Why not `A_KEY` -> `ED_KEY`, `B_KEY` -> `ED_KEY`, and `C_KEY` -> `CID_KEY`?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/signature/AppSignatureVerifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+                    final Set<ExpandedSignaturePair> pairs = invocationOnMock.getArgument(2);
+                    pairs.addAll(expandedPairs);
+                    return null;
+                })
+                .given(signatureExpander)
+                .expand(eq(TEST_KEY), eq(SignatureMap.DEFAULT.sigPair()), any(Set.class));
+
+        final var hederaConfig = DEFAULT_CONFIG.getConfigData(HederaConfig.class);
+        given(future.get(hederaConfig.workflowVerificationTimeoutMS(), TimeUnit.MILLISECONDS))
+                .willReturn(verification);
```
david-bakin-sl, line 99: Seems appropriate for the `@BeforeEach`?

Possibly also L87-95 - in this set of tests the same expanded map is used in each test but in the future maybe not?

david-bakin-sl (commented):
A comment for now - everything looks _very_ good except for my question about double-hashing, which I'm confused on.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/signatures/SignatureVerifier.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param key the key whose signature should be verified
+     * @param bytes the payload to verify
+     * @param signatureMap the cryptographic signatures to verify against
+     * @param simpleKeyVerifier an optional override strategy for verifying simple key signatures
+     * @return true if the signature is valid; false otherwise
+     */
+    boolean verifySignature(
+            @NonNull Key key,
+            @NonNull Bytes bytes,
+            @NonNull SignatureMap signatureMap,
```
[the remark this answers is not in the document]
tinker-michaelj: You mean if the EC public key isn't 65 bytes? Added javadoc to clarify it must be in 33-byte compressed form, same as used in the `Key` structure for HAPI.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/signatures/SignatureVerifier.java
```diff
@@ -0,0 +1,65 @@
...
+            @Nullable Function<Key, SimpleKeyVerification> simpleKeyVerifier);
+
+    /**
+     * Convenience method for getting the number of Ed25519 and ECDSA(secp256k1) keys in a {@link Key} structure,
+     * useful for estimating the maximum amount of work that will be done in verifying the signature of the key.
+     * @param key the key structure to count simple keys in
+     * @return the number of Ed25519 and ECDSA(secp256k1) keys in the key
+     */
+    SimpleKeyCount countSimpleKeys(@NonNull Key key);
```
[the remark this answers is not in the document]
tinker-michaelj: Moved into interface.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/signature/AppSignatureVerifier.java
tinker-michaelj: Excellent question...we have to sort out what bytes people will be passing to this system contract...sounds like based on `isAuthorizedRaw` we want the bytes here to be a 32-byte hash only?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/signature/AppSignatureVerifierTest.java
tinker-michaelj: Good point, created a `@Nested` class where the universal setup is in a `@BeforeEach`.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/signatures/SignatureVerifier.java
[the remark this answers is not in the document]
david-bakin-sl: That's a good clarification for the Javadoc - but still - you could stick 100 bytes into that protobuf field - it would still be a valid protobuf (it would deserialize fine).  But you couldn't use the value: it's not a 33-byte compressed signature.  So when that happens, does this method throw an exception or return `false`?  With one caveat I don't much care which, I just want to know how to code against this API.  (The caveat is that the system contract could return more specific information to the caller via a response code if clearly bad input could be distinguished from `false`.)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/signature/AppSignatureVerifier.java
david-bakin-sl: Yes I think that's ok.  Well, for EC.  I think for ED, based on what I found on the intertubes, signatures are most often done with either SHA-384 or SHA-512 which of course have different lengths.

hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/signatures/SignatureVerifier.java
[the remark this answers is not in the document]
tinker-michaelj: Ah good clarification, adding to javadoc also that `SignaturePair`s with overly long public keys are completely ignored.

david-bakin-sl (approved):
LGTM - the API is fine! - (initial tests favorable - if anything turns up can fix it then)

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #14478, opened 2024-07-30
fix: Temporarily disable throttle state reads for auto associations

### LINKED ISSUE #14472: Disable auto-association throttle
Disable the auto-association throttle. It is crucial that we do not do any file reads in `ThrottleAccumulator.getAutoAssociationsCount()`.

mhess-swl:
In throttle calculations, auto association counting currently takes way too long due to almost-certain I/O ops happening as we read state from disk. This PR includes a quick performance fix to avoid these ops for now.

Closes #14472

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
@@ -658,6 +658,7 @@ public class ThrottleAccumulator {
         return implicitCreationsCount;
     }
 
+    @SuppressWarnings("ReplaceInefficientStreamCount")
     public int getAutoAssociationsCount(
             @NonNull final TransactionBody txnBody, @NonNull final ReadableTokenRelationStore relationStore) {
         int autoAssociationsCount = 0;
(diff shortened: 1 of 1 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -666,15 +667,18 @@ public int getAutoAssociationsCount(
             return 0;
         }
         for (var transfer : cryptoTransferBody.tokenTransfers()) {
-            final var tokenID = transfer.token();
             autoAssociationsCount += (int) transfer.transfers().stream()
                     .filter(accountAmount -> accountAmount.amount() > 0)
-                    .map(AccountAmount::accountID)
-                    .filter(accountID -> hasNoRelation(relationStore, accountID, tokenID))
+                    // FUTURE: This is a temporary performance fix to avoid state reads for auto-associations
```
netopyr, line 672: I think this makes the throttles more severe instead of disabling them because it accounts for tokens that already are associated. Maybe it is sufficient to return 0 from this method.

mhess-swl:
Going with #14498 instead. Closing

Issue #14466, opened 2024-07-29
fix: docker image needs to include registry mirrors


rbarkerSL:
## Background

Need to update docker image creation in github actions workflows to use the docker registry mirrors. This should enable release flows to move forward.

PR #14488, opened 2024-07-30
Fix title check

dalvizu:
**Description**:
Updates GHA to use docker registry mirror so that title check does not fail due to rate limits

**Related issue(s)**:
#14466

### DIFF
--- a/.github/workflows/flow-pull-request-formatting.yaml
+++ b/.github/workflows/flow-pull-request-formatting.yaml
@@ -50,7 +50,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          version: v25.0.5
           daemon-config: |
             {
               "registry-mirrors": [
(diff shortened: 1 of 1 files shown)


.github/workflows/flow-pull-request-formatting.yaml
```diff
@@ -49,6 +49,14 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          daemon-config: |
+            {
+              "registry-mirrors": [
+                "https://artifacts.swirldslabs.io/central-docker-external/"
+              ]
+            }
+
```
rbarkerSL, line 59: We shouldn't add this to this action. The daemon-config is for docker engine setup.

.github/workflows/flow-pull-request-formatting.yaml
```diff
@@ -61,7 +69,7 @@ jobs:
             exit 1
 
   assignee-check:
-    name: Assignee Check
+    name: Assigneed Check
```
rbarkerSL, line 72: Why are we renaming this check with a typo?

rbarkerSL (changes requested):

nathanklick (changes requested):
Invalid PR, `daemon-config` is not a valid option here. Will break all PRs when merged.

PR #14493, opened 2024-07-30
fix: add `event.migrateEventHashing=false` override to upgrade `@HapiTest`

tinker-michaelj:
**Description**:
 - Until changes in event hashing are enabled, an upgrade with `event.migrateEventHashing=true` will expect the PCES events to be in a `v0.52` format and fail otherwise.
 - This PR updates dev configuration _settings.txt_ to set `event.migrateEventHashing=false`.


timo0 (approved):

thomas-swirlds-labs (approved):
LGTM @tinker-michaelj 👍

iwsimon (approved):

PR #14497, opened 2024-07-30
fix: use gagues for ReconnectMapMetrics and reset them to zero on new…

### LINKED ISSUE #14440: Use gauges instead of counters, and reset to zero for new reconnects

anthony-swirldslabs:
… reconnects

**Description**:
We want to measure ReconnectMapMetrics for each reconnect run, and a node can reconnect multiple times while it's running. However, Counters cannot be reset to zero and can only grow. So in this fix I'm replacing Counters with LongGauges and reset them when a new reconnect starts (aka when a new instance of ReconnectMapMetrics is created.)

**Related issue(s)**:

Fixes #14440

**Notes for reviewer**:
Unit tests pass in swirlds-common, swirlds-merkle, and swirlds-virtualmap.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
@@ -16,14 +16,14 @@
 
 package com.swirlds.common.merkle.synchronization.stats;
 
-import com.swirlds.metrics.api.Counter;
+import com.swirlds.metrics.api.LongGauge;
 import com.swirlds.metrics.api.Metrics;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.Objects;
 
 /**
- * An implementation of ReconnectMapStats that emits all the stats as Counter metrics.
+ * An implementation of ReconnectMapStats that emits all the stats as LongGauge metrics.
  */
 public class ReconnectMapMetrics implements ReconnectMapStats {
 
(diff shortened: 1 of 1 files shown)


artemananiev (approved):

imalygin (approved):

timo0 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/merkle/synchronization/stats/ReconnectMapMetrics.java
```diff
@@ -67,48 +67,77 @@ public ReconnectMapMetrics(
...
+        // Reset metric values to zeros on reconnect start
+        resetMetrics();
     }
 
     private static String formatName(final String name, final String label) {
         return (label == null || label.isBlank() ? name : (name + "_" + label + "_")) + "Total";
     }
 
+    private static void add(final LongGauge metric, final long value) {
+        metric.set(metric.get() + value);
```
lpetrovic05, line 110: this is not thread safe, 2 threads updating at the same time might increment it only once. I don't know if there will be any contention on this, but its worth noting.

@hendrikebbers this PR points to the fact that there is a requirement that we cannot meet atm. is there any way to support it in the short term?

anthony-swirldslabs: Yes, that's correct, this method is not thread-safe. Currently, the reconnect is a single-threaded process and its protocol is not designed to handle reconnect of multiple sub-trees in parallel today. While there are some ideas to parallelize the process in the future, there's not any concrete plans for doing that in the short term.

So as far as the current usages of the ReconnectMapMetrics are concerned, it's okay for this method to not be thread-safe at this time.

If in the future a parallel reconnect algorithm is implemented, then this metrics collection code would need changes to synchronize access to the metric values. But doing so is out of scope of the current fix for the reasons mentioned above.

PR #14498, opened 2024-07-30
chore: Add config to disable frontend throttles

### LINKED ISSUE #14389: Lock frontend-throttles later
Frontend throttles are called from many threads during the ingest workflow and must, therefore, be thread-safe. Backend throttles, on the other hand, are only called from the handle thread. Currently, the throttle functionality is not thread-safe. Frontend throttles are implemented by wrapping the regular throttles and guarding the functionality via synchronized methods. Unfortunately, this causes performance issues because the throttle functionality can be slow in some instances (e.g., it may involve several disk reads).

We should refactor the functionality so that only reading and updating the bucket data are guarded while preparing the required information (e.g., the estimated number of auto-associations) can be done concurrently.

Neeharika-Sompalli:
Fixes #14389 

Adds config to disable frontend throttles for testing


mhess-swl (approved):

tinker-michaelj (approved):

MrValioBg (approved):
LGTM! Thank you

PR #14503, opened 2024-07-31
chore: remove implicit record stream assumption from builder names

tinker-michaelj:
**Description**:
 - Rename record stream builders to just `StreamBuilder`'s to reduce confusion in migration to block stream.

### DIFF
--- a/hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenBaseRecordBuilder.java
+++ b/hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenBaseRecordBuilder.java
@@ -17,13 +17,13 @@
 package com.hedera.node.app.service.token.records;
 
 import com.hedera.hapi.node.base.TokenType;
-import com.hedera.node.app.spi.workflows.record.SingleTransactionRecordBuilder;
+import com.hedera.node.app.spi.workflows.record.StreamBuilder;
 import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
  * The base interface for Token Service record builders that record operations on Tokens.
  */
-public interface TokenBaseRecordBuilder extends SingleTransactionRecordBuilder {
+public interface TokenBaseRecordBuilder extends StreamBuilder {
     /**
      * Sets the {@link TokenType} of the token the recorded transaction created or modified.
      * @param tokenType the token type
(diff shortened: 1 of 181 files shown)


david-bakin-sl (approved):
LGTM, for smart contract service only (totally bland, a true "chore").

Surprising to me to see in one place at one time how the usage of `ContractCallRecordBuilder` is all over smart contract service.  I guess it's unavoidable - it's a key point of communication between the smart contract service and the outside world.  But still, a key dependency.  Wonder if there's any alternative that would decouple things a bit more, that would be worth it.  Sometimes you just need the dependency/coupling.  (Same of course for token service.)

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenBaseRecordBuilder.java
```diff
@@ -17,13 +17,13 @@
...
 import com.hedera.hapi.node.base.TokenType;
-import com.hedera.node.app.spi.workflows.record.SingleTransactionRecordBuilder;
+import com.hedera.node.app.spi.workflows.record.StreamBuilder;
 import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
  * The base interface for Token Service record builders that record operations on Tokens.
  */
-public interface TokenBaseRecordBuilder extends SingleTransactionRecordBuilder {
+public interface TokenBaseRecordBuilder extends StreamBuilder {
```
MrValioBg, line 26: Shouldn't this also be TokenBaseStreamBuilder?

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenBurnStreamBuilder.java
```diff
@@ -23,7 +23,7 @@
  * A {@code RecordBuilder} specialization for tracking the side effects of a {@code TokenBurn}
```
MrValioBg, line 23: Maybe we could also improve the JavaDocs for clarity, as here it's stated as "RecordBuilder."

Neeharika-Sompalli (approved):
LGTM!

hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/records/TokenBurnStreamBuilder.java
tinker-michaelj: Definitely...did a global replace of `{@code RecordBuilder}` with `{@code StreamBuilder}`...realized there are also a huge number of occurrences of the string "record builder"...will change those more opportunistically. 😓

MrValioBg (approved):
LGTM! Thank you @tinker-michaelj

david-bakin-sl (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #14511, opened 2024-07-31
feat: limit PCES replay frequency

litt3:
- cherry pick of https://github.com/hashgraph/hedera-services/pull/14465

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
@@ -22,8 +22,10 @@ import static com.swirlds.logging.legacy.LogMarker.STARTUP;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import com.swirlds.base.time.Time;
+import com.swirlds.common.context.PlatformContext;
 import com.swirlds.common.formatting.UnitFormatter;
 import com.swirlds.common.io.IOIterator;
+import com.swirlds.common.utility.throttle.RateLimiter;
 import com.swirlds.common.wiring.wires.output.StandardOutputWire;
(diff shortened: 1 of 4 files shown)


mxtartaglia-sl (approved):

edward-swirldslabs (approved):

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
```diff
@@ -170,6 +178,10 @@ public NoInput replayPces(@NonNull final IOIterator<PlatformEvent> eventIterator
                 // until it catches up before we can continue.
                 waitUntilHealthy();
 
+                if (config.limitReplayFrequency() && !rateLimiter.requestAndTrigger()) {
+                    continue;
+                }
```
anthony-swirldslabs, line 183: Minor: this creates a tight loop, potentially consuming more CPU than it could've. It would be nice to have a new API in the RateLimiter class that would wait/sleep until a token is available.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesReplayer.java
litt3: I think it would be worth using such a feature if it existed, but probably not worth implementing for this use case. The slight extra CPU usage isn't a big concern at this point of startup, with the bottleneck being a single transaction handling thread.

anthony-swirldslabs: Sounds good.

kfa-aguda (approved):
lgtm

PR #14514, opened 2024-07-31
feat: Translate block transactions to `SingleTransactionRecord`s

### LINKED ISSUE #14438: Translate `BlockItem`s with same `TransactionID` to equivalent `SingleTransactionRecord`s
To start, we must ensure all information in the current stream of `SingleTransactionRecord`s is preserved in the block stream. (This includes not only `TransactionRecord`s but also sidecars.)

The most comprehensive way to validate we have done this is to implement a translation **FROM**,
  - A list of `BlockItem` tuples `(Transaction, TransactionResult[, TransactionOutput][, StateChanges][, ContractStateChanges][, ContractActions][, ContractBytecodes])` whose `TransactionResult`s have `TransactionID` with the same `accountID` and `transactionValidStart`.
  
**TO**
 - A list of `SingleTransactionRecord`s.

We need this to work for all valid combinations of such tuples, including those generated by e.g., a `ScheduleSign` that triggers a `ContractCall` that has a child `CryptoTransfer` that triggers an auto-creation.

mhess-swl:
There are a few items left to address, each noted with a comment in the code. We can implement these after the `TransactionRecordTranslator` interface is merged so we don't keep anyone from implementing transaction type-specific translators.

Closes #14438


hedera-node/hedera-app/src/main/java/module-info.java
```diff
@@ -48,10 +48,10 @@
     requires io.netty.transport;
     requires org.apache.commons.lang3;
     requires org.apache.logging.log4j;
-    requires org.jetbrains.annotations;
     requires static com.github.spotbugs.annotations;
     requires static com.google.auto.service;
     requires static java.compiler;
+    requires static org.jetbrains.annotations;
```
david-bakin-sl, line 54: Why are we using `org.jetbrains.annotations` here instead of `edu.umd.cs.findbugs.annotations` that we use everywhere else in services?

The only three files that use it are `RoundStateChangeListener`, `ConcurrentBlockItemWriter`, and `GrpcBlockItemWriter` - all in the package `com.hedera.node.blocks.impl`, and the annotation that's used is `@NotNull` which exists in the usual `edu.umd.cs.findbugs.annotations` module (though it is called `@NonNull` there of course).

(Strangely I don't see `edu.umd.cs.findbugs.annotations` in any `module-info.java` in the repo!  Is it somehow part of the build system?)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -0,0 +1,290 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.services.bdd.junit.support.translators;
+
+import static com.hedera.hapi.block.stream.output.UtilPrngOutput.EntropyOneOfType.PRNG_BYTES;
+import static com.hedera.hapi.block.stream.output.UtilPrngOutput.EntropyOneOfType.PRNG_NUMBER;
+import static com.hedera.services.bdd.junit.support.translators.BlockStreamTransactionTranslator.*;
```
david-bakin-sl, line 21: This can be a static import of `BlockTransaction` instead of `*`?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -0,0 +1,290 @@
...
+        public List<BlockItem> asItems() {
+            final var blockItems = new ArrayList<BlockItem>();
+            blockItems.add(txnInput);
+            blockItems.add(transactionResult);
+            if (transactionOutput != null) {
+                blockItems.add(transactionOutput);
+            }
+            blockItems.add(stateChanges);
+
+            return blockItems;
```
david-bakin-sl, line 60: Ok, here's an alternative for a laugh:

```java
return Stream.of(txnInput, transactionResult, transactionOutput, stateChanges)
        .filter(Objects::nonNull)
        .toList();
```

I think.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<TransactionRecord> translateAll(final List<BlockTransaction> transactions) {
+        throw new NotImplementedException();
+    }
+
+    private void validateBlockItems(final List<BlockItem> blockItems) {
```
david-bakin-sl, line 132: I don't get why this doesn't just take a `BlockTransaction` instead of turning it into a `List` first.  If a `BlockTranslation` already then you don't have to worry about index numbers here, information which seems properly encapsulated in `BlockTransaction.asBlockTransaction` factory method already.

(Only caller (at this time) already has the `BlockTransaction`.  If the issue is that maybe a different caller might someday only have the list of things then that argues for this `validateBlockItems` method being a member of the record `BlockTransaction` itself and called inside of `asBlockTransaction` first thing.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -0,0 +1,290 @@
...
+            throw new IllegalArgumentException("Expected second block item to be a txnInput result");
+        }
+
+        final var stateChangesItem = blockItems.getLast();
+        if (!stateChangesItem.hasStateChanges()) {
+            throw new IllegalArgumentException("Expected last block item to be state changes");
+        }
+    }
+
+    private TransactionRecord.Builder parseTransaction(
```
david-bakin-sl, line 153: (From here on down I'm skipping the `parseXXX` routines because I'm not sure what fields are required to be copied over to the record builders.  For example, here in `parseTransaction`, why only the txid and the memo and not any of the other stuff? Dunno, beyond my scope.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -0,0 +1,290 @@
...
+
+        // TODO: assign `newPendingAirdrops` (if applicable)
+
+        trb.receipt(rb.build());
+
+        return trb;
+    }
+
+    private void maybeAssignEvmAddress(TransactionRecord.Builder trb, TransactionOutput txnOutput) {
+        // Are these the only two places where default EVM addresses are assigned?
```
david-bakin-sl, line 282: "default" -> "EVM address alias"?  I think that's what you mean.

david-bakin-sl: HIP-583 says you can get an EVM address alias via a CrytoTransfer and EthereumTransaction too, if that's relevant here.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/TransactionRecordTranslator.java
```diff
@@ -0,0 +1,48 @@
...
+    TransactionRecord translate(@NonNull T transaction);
+
+    /**
+     * Much like the {@link #translate(Object)} method, but for translating a collection of transactions,
+     * or for translating a single transaction to multiple {@link TransactionRecord} outputs.
+     *
+     * @param transactions a collection of transactions to translate
+     * @return the equivalent txnInput records
+     */
+    List<TransactionRecord> translateAll(final List<T> transactions);
```
david-bakin-sl, line 47: I think you can implement this here via a default method.

david-bakin-sl (approved):
Looks nice! Mostly optional comments here, except about the annotations, which to my understanding of how we're using annotations should be fixed (but maybe as a separate PR, though I see no reason why you can just smush that fix in here).

tinker-michaelj (commented):
I like the idea of the `BlockTransaction` wrapper that groups the input/output `BlockItem`s...but note the `StateChanges` item can actually be needed to "fill in the gaps" to complete translation for **multiple** transactions.

So I would lean toward,
```
record SingleTransactionBlockItems(
            @NonNull Transaction txn,
            @NonNull TransactionResult result,
            @Nullable TransactionOutput output) {}
```

with a translator like,
```
List<SingleTransactionRecord> translateToRecords(
  @NonNull List<SingleTransactionBlockItems> items,
  @NonNull StateChanges stateChanges);
```

```

File: hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java [hunk 8]
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        //            if (txnOutput.hasTopicCreate()) {
+        //                rb.topicID(txnOutput.topicCreate().topicID());
+        //            }
+
+        if (txnOutput.hasSubmitMessage()) {
+            //                rb.topicSequenceNumber(txnOutput.submitMessage().topicSequenceNumber());
+            rb.topicRunningHashVersion(
+                    txnOutput.submitMessage().topicRunningHashVersion().protoOrdinal());
```
MiroslavGatsanoga, line 221: It looks like the `protoOrdinal()` has values in reverse order i.e. 0 is the most recent version, while services handler uses `RUNNING_HASH_VERSION = 3L`, is that correct?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: The reason I wrote it this way is because this method came before `asBlockTransaction`, and I didn't think to change it :) good call, will do

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: There may be more optimal ways to parse these things. I matched whatever I could find, but probably missed stuff

hedera-node/hedera-app/src/main/java/module-info.java
mhess-swl: I was surprised to see this in here too. I only added the `static` part because gradle mandated it

david-bakin-sl: I think it was just a mistake in the feature branch - that could be _fixed_ (hint hint)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: From the comments in the `RunningHashVersion` protobuf, the `protoOrdinal()`, it's correct that 0 is the most recent version, and it does appear that the submit message handler uses 3L.

This code is only taking the output and setting it on the record builder, so it should be correct in either case.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: Done

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: Done

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
mhess-swl: Added a commented version since there doesn't seem to be an alias available on those transaction types

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/TransactionRecordTranslator.java
mhess-swl: Done

hedera-node/hedera-app/src/main/java/module-info.java
mhess-swl: Looks like there are other instances of Jetbrains `@NotNull` in `develop`. I'd prefer to fix this there in a separate PR

david-bakin-sl: ok! but before `stream-manager-spike` gets merged to `develop` please!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
david-bakin-sl: (not really for a laugh, I was feeling shy)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
```diff
@@ -0,0 +1,103 @@
...
+        return builder.build();
+    }
+
+    public static class Builder {
+        private Transaction txn;
+        private TransactionResult result;
+        private TransactionOutput output;
+
+        public Builder txn(@NonNull final Transaction txn) {
+            this.txn = txn;
```
david-bakin-sl, line 85: These two methods in the builder with argument annotated `@NonNull` - do you want to check that?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
```diff
@@ -0,0 +1,103 @@
...
+            this.result = result;
+            return this;
+        }
+
+        public Builder output(@Nullable TransactionOutput output) {
+            this.output = output;
+            return this;
+        }
+
+        public SingleTransactionBlockItems build() {
```
david-bakin-sl, line 99: I've sometimes wondered why the `build` method of builders doesn't do the verification of all fields.  E.g., non-null.  It just isn't generally done, that I've seen, but why not?

david-bakin-sl (approved):
LGTM (quite different now!)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
mhess-swl: Good point!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
mhess-swl: I don't understand, do we not want that?

david-bakin-sl: no i mean the usual thing is to have a `requireNonNull()` right away

mhess-swl: Verification done in `build()` as suggested 👍

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
mhess-swl: Done

david-bakin-sl (approved):
LGTM!

hedera-node/hedera-app/src/main/java/module-info.java
mhess-swl: https://github.com/hashgraph/hedera-services/pull/14548

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
MiroslavGatsanoga: Yeah, actually there's a [TODO for block stream writing](https://github.com/hashgraph/hedera-services/blob/stream-manager-spike/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/IoBlockItemsBuilder.java#L702) for to be confirmed for the value of the field, so if there is going to be some change it should be addressed there.

MiroslavGatsanoga (approved):
LTGM

PR #14518, opened 2024-07-31
chore: Refactor AirdropHandler

Neeharika-Sompalli:
Refactoring `TokenAirdropHandler` based on review comments


vtronkov (approved):

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -252,148 +194,210 @@ final var record =
...
+     * Gets the airdrop fee for the token airdrop transaction.
+     * @param feeContext the fee context
+     * @return the airdrop fee
+     */
+    private long airdropFeeFor(@NonNull final FeeContext feeContext) {
+        return feeContext
+                .feeCalculatorFactory()
+                .feeCalculator(SubType.DEFAULT)
+                .calculate()
+                .totalFee();
```
vtronkov, line 377: With this code wouldn't we always charge $0.1? Shouldn't we check for associations? If there are we should charge - $0.05, if there aren't - $0.1.

JivkoKelchev (approved):
LGTM, thank you @Neeharika-Sompalli

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
Neeharika-Sompalli: Discussed with Ali. It seems these are the fees. Will update the implementation

> Here are 3 possible scenarios, let's assume you are airdropping token A to me:
> - If I explicitly associated with token A, we only charge the CryptoTransfer fee of $0.001, 
> - If I have open slots or unlimited association (maxAutomaticAssociations: -1), meaning I don't have explicit association to token A but can receive airdrops, then TokenAirdrop would cost $0.1
> - If I have no open slots (maxAutomaticAssociations: 0), which means the airdrop will go to pending state, then TokenAirdrop would be $0.1 still.

PR #14527, opened 2024-08-01
refactor: Provide EventTransaction to the app

### LINKED ISSUE #14426: Provide EventPayload to the app instead of OneOf
`EventPayload` (soon to be `EventTransaction`) is a simple wrapper that holds a `OneOf`. Since the block stream will contain the wrapper, it makes more sense to provide it to the application instead of a simple `OneOf`.

timo0:
Fix #14426

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/Transaction.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/Transaction.java
@@ -42,7 +42,7 @@ public sealed interface Transaction permits ConsensusTransaction {
      * A convenience method for retrieving the application transaction {@link Bytes} object. Before calling this method,
      * ensure that the transaction is not a system transaction by calling {@link #isSystem()}.
      *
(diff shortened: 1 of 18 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/Transaction.java
```diff
@@ -46,7 +45,7 @@ public sealed interface Transaction permits ConsensusTransaction {
      * @return the application transaction Bytes or null if the transaction is a system transaction
```
lpetrovic05, line 45: seems this javadoc is innacurate, it says it will return null, but it wont

lpetrovic05 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/transaction/Transaction.java
timo0: fixed

lpetrovic05 (approved):

PR #14538, opened 2024-08-01
fix: Fixes issue with newVersion not being specified in versionAsSpecified Task

### LINKED ISSUE #14537: Gradle versionAsSpecified task is breaking when providing newVersion
When `VersionAsSpecified` task is called the exception is always thrown:

FAILURE: Build failed with an exception.

```
* What went wrong:
Configuration cache state could not be cached: input property `newVersion` of task `:versionAsSpecified` of type `org.gradle.api.DefaultTask`: error writing value of type 'org.gradle.api.internal.provider.OrElseProvider'
> No newVersion property provided! Please add the parameter -PnewVersion=<version> when running this task.

```

Need to fix this

rbarkerSL:
**Description**:

Fixes issue in versionAsSpecified task where the exception is always thrown in the property definition.

**Related issue(s)**:

Fixes #14537

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.root.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.root.gradle.kts
@@ -117,7 +117,7 @@ tasks.register("versionAsSpecified") {
 
     inputs.property("newVersion", providers.gradleProperty("newVersion").orElse(""))
 
-    if(inputs.properties["newVersion"] as String == "") {
+    if (inputs.properties["newVersion"] as String == "") {
         throw IllegalArgumentException(
             "No newVersion property provided! " +
                 "Please add the parameter -PnewVersion=<version> when running this task."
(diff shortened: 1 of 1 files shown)


gradle/plugins/src/main/kotlin/com.hedera.gradle.root.gradle.kts
```diff
@@ -115,19 +115,14 @@ tasks.register("versionAsSnapshot") {
...
-            .orElse(
-                provider {
-                    throw IllegalArgumentException(
-                        "No newVersion property provided! " +
-                            "Please add the parameter -PnewVersion=<version> when running this task."
-                    )
-                }
-            )
-    )
+    inputs.property("newVersion", providers.gradleProperty("newVersion").orElse("INVALID"))
```
nathanklick, line 118: This should leverage an `orElseThrow()` if available. If `orElseThrow()` is not available then `null` should be returned and the `if` should check for a `null` value instead of a magic string.

nathanklick (changes requested):

gradle/plugins/src/main/kotlin/com.hedera.gradle.root.gradle.kts
rbarkerSL: I'll do the check on `null`

rbarkerSL: Well... the orElse requires either a provider<String!> or a <String!> parameter. So I can use the empty string.

derektriley (dismissed):

nathanklick (approved):

PR #14549, opened 2024-08-01
fix: move `VirtualMap#registerMetrics()` call to a node initializer

### LINKED ISSUE #14520: Many VirtualMap / MerkleDb metrics are missing in grafana
A few examples of missing / zero metrics:

* virtual_map_vmap_size_TokenService_ACCOUNTS
* virtual_map_vmap_lifecycle_flushDurationMs_TokenService_ACCOUNTS
* merkle_db_ds_files_totalSizeMb_TokenService_ACCOUNTS
* merkle_db_ds_flushes_hashesStoreFileSizeMb_TokenService_ACCOUNTS

These metrics are not available, or not set (or sometimes set to zero) in both production and pre-production grafana instances, for all networks. The only exception is `engnet1`, which was started to run my branch about an hour ago. It shows all metrics just fine. It means, there are no issues with VirtualMap / MerkleDb, the metrics are produced correctly, but there are issues with exporting them from the process or importing to prometheus / grafana.

The branch was synchronized with develop a few hours ago, but there are no metrics specific changes in the branch. The test was started from a genesis state. When the same branch was running in the same network two days ago (but from an existing state), the metriss were not available.

tinker-michaelj:
**Description**:
 - Closes #14520 
 - Instead of calling `VirtualMap#registerMetrics()` only at the moment of creating an _absent_ node in `MerkleStateRoot#putServiceStateIfAbsent()`, does this work in an initialization callback that receives the node for the state being defined, no matter if present or absent at beginning of the call.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistry.java
@@ -298,29 +298,30 @@ public class MerkleSchemaRegistry implements SchemaRegistry {
                             return map;
                         });
                     } else {
-                        stateRoot.putServiceStateIfAbsent(md, () -> {
-                            // MAX_IN_MEMORY_HASHES (ramToDiskThreshold) = 8388608
-                            // PREFER_DISK_BASED_INDICES = false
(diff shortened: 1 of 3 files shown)


david-bakin-sl (approved):
LGTM

edward-swirldslabs (approved):

Neeharika-Sompalli (approved):
LGTM!

PR #14551, opened 2024-08-02
refactor: 11771 Converted `PlatformState` into a Singleton

### LINKED ISSUE #11771: Refactor `PlatformState` into a singleton

imalygin:
**Description**:

This PR refactors access to the platform state. It introduces `PlatformStateAccessor` interface that has all the access methods of `PlatformState`. This interface two implementations - `PlatformState` and a combination of  State API classes `ReadablePlatformStateStore` and `WriteablePlatformStateStore`. These classes are backed by `ReadableSingletonState` and 
`WritableSingletonState` respectively. 

PR highlights:

- `V0540PlatformStateSchema` is a class that holds meta information about the platform state such as the service name and the state key. Also, it's responsible for initializing genesis state if the platform state is absent.
- The platform state has a different lifecycle from the other services, and it has to be initialized first. `Hedera.initPlatformState` addresses this issue. For the same reason the migration of the platform state happens not in the schema but in `MerkleStateRoot#migrate`.
- `com.hedera.node.app.state.PlatformStateAccessor` is no longer needed and, therefore, its removed in favor of accessing the platform state via instance of `ReadablePlatformStateStore` retrieved from `ReadableStoreFactory`
- `PlatformState` class is deprecated and should be removed in the next version

**Related issue(s)**:

Fixes #11771 

**Notes for reviewer**:

**Checklist**


david-bakin-sl (commented):
Looking only at `cli-clients/.../SignedStateHolder.java` looks good to me.  builds too.  so i'm good.  Let me know when this goes out of draft and you need a signoff.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -421,9 +444,9 @@ public void onStateInitialized(
      * <p>If the {@code deserializedVersion} is {@code null}, then this is the first time the node has been started,
      * and thus all schemas will be executed.
      *
-     * @param state current state
+     * @param state               current state
```
artemananiev, line 447: Extra (redundant?) whitespaces

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -250,14 +255,16 @@ public Hedera(
                         """,
                 HEDERA);
         bootstrapConfigProvider = new BootstrapConfigProviderImpl();
-        bootstrapConfig = bootstrapConfigProvider.getConfiguration();
+        final Configuration bootstrapConfig = bootstrapConfigProvider.getConfiguration();
         version = getNodeStartupVersion(bootstrapConfig);
         servicesRegistry = registryFactory.create(constructableRegistry, bootstrapConfig);
+        registerPlatformStateSchema();
```
artemananiev, line 261: Why is the platform service registered differently from all other services? Why is it an anonymous inner class in `Hedera` rather than a regular class similar to other services?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -286,13 +293,29 @@ public Hedera(
...
+                    MerkleStateRoot.class,
+                    () -> new MerkleStateRoot(new MerkleStateLifecyclesImpl(this), softwareVersionSupplier)));
         } catch (final ConstructableRegistryException e) {
             logger.error("Failed to register " + MerkleStateRoot.class + " factory with ConstructableRegistry", e);
             throw new IllegalStateException(e);
         }
     }
 
+    private void registerPlatformStateSchema() {
+        servicesRegistry.register(new Service() {
```
artemananiev, line 305: Shouldn't it be extracted to a separate class?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/PrecompileMintThrottlingCheck.java
```diff
@@ -67,7 +67,7 @@ public class PrecompileMintThrottlingCheck extends HapiSuite {
     private final AtomicReference<TimeUnit> unit = new AtomicReference<>(SECONDS);
     private final AtomicInteger maxOpsPerSec = new AtomicInteger(50);
     private static final int EXPECTED_MAX_MINTS_PER_SEC = 50;
-    private static final double ALLOWED_THROTTLE_NOISE_TOLERANCE = 0.05;
+    private static final double ALLOWED_THROTTLE_NOISE_TOLERANCE = 0.10;
```
artemananiev, line 70: Is it related to this PR?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
imalygin: >Why is the platform service registered differently from all other services? 

It has to be registered first. Other services may need access to platform state accessor for their purposes. 

>Why is it an anonymous inner class in Hedera rather than a regular class similar to other services?

It doesn't have to be anonymous but it's doing nothing but tying the schema and the service registry together. So, I thought it's not pulling its weight as a full-blown class.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
imalygin: I don't see a reason why it should be. It's not going be used anywhere else and it plays a role of connecting tissue.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/PrecompileMintThrottlingCheck.java
imalygin: Yes. It's related to this PR. @tinker-michaelj recommended updating this parameter to stabilize one of the HAPI tests.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/GenesisStateBuilder.java
```diff
@@ -68,11 +64,18 @@ public static ReservedSignedState buildGenesisState(
             @NonNull final MerkleRoot stateRoot) {
 
         final BasicConfig basicConfig = platformContext.getConfiguration().getConfigData(BasicConfig.class);
-        stateRoot.setPlatformState(buildGenesisPlatformState(addressBook, appVersion));
+        final PlatformStateAccessor platformState;
+        if (stateRoot instanceof MerkleStateRoot root) {
+            platformState = new PlatformStateAccessorSingleton(root);
+        } else {
+            platformState = new PlatformState();
+        }
```
artemananiev, line 72: What about adding a method to `MerkleRoot` like `initPlatformState()` or `createPlatformState()` to avoid instanceof checks here?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -38,13 +38,13 @@ public interface MerkleRoot extends MerkleInternal {
      * @return the platform state
      */
     @NonNull
-    PlatformState getPlatformState();
+    PlatformStateAccessor getPlatformStateAccessor();
```
artemananiev, line 41: Keeping method name as `getPlatformState()`, just changing the return type, would reduce the number of changes in this PR.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/GenesisStateBuilder.java
```diff
@@ -68,11 +64,18 @@ public static ReservedSignedState buildGenesisState(
             @NonNull final MerkleRoot stateRoot) {
 
         final BasicConfig basicConfig = platformContext.getConfiguration().getConfigData(BasicConfig.class);
-        stateRoot.setPlatformState(buildGenesisPlatformState(addressBook, appVersion));
+        final PlatformStateAccessor platformState;
```
artemananiev, line 67: Where is this platform state object used? It's created below, then initialized in `initGenesisPlatformState()`, and then ...?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -200,7 +196,7 @@ public MerkleStateRoot() {
     @Override
     public void init(
             final Platform platform,
-            final PlatformState platformState,
+            final PlatformStateAccessor platformState,
```
artemananiev, line 199: How is this `platformState` related to the field with the same name (created in the constructor)?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -257,6 +254,8 @@ private MerkleStateRoot(@NonNull final MerkleStateRoot from) {
                 setChild(childIndex, childToCopy.copy());
             }
         }
+
+        this.platformState = new PlatformStateAccessorSingleton(this);
```
artemananiev, line 258: Is a new object required here? One of the children in the loop right above was this platform state singleton. A copy of it has been already created in the loop. My assumption would be that copy should be used as `this.platformState`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -914,32 +922,18 @@ public SwirldState getSwirldState() {
...
-        // Shift all existing children to the right
-        if (getNumberOfChildren() != 0) {
-            for (int i = getNumberOfChildren(); i > 0; i--) {
-                MerkleNode child = getChild(i - 1);
-                setChild(i, child.copy());
-                setChild(i - 1, null);
-            }
-        }
-        setChild(PLATFORM_STATE_INDEX, platformState);
+    public void updatePlatformStateAccessor(@NonNull final PlatformStateAccessor source) {
```
artemananiev, line 935: Method name looks confusing to me. It doesn't update the accessor, it updates accessor state, did I get it right? So a better name would be `updatePlatformState()`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformState.java
```diff
@@ -34,16 +34,13 @@
 
 /**
  * State managed and used by the platform.
+ * @deprecated Use {@link PlatformStateAccessorSingleton}. This object should be accessed via the State API.
  */
-public class PlatformState extends PartialMerkleLeaf implements MerkleLeaf {
+@Deprecated(since = "0.54.0", forRemoval = true)
```
artemananiev, line 39: Are there plans to change existing non-Hedera apps to not use this class? If not, then deprecation doesn't make sense.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
```diff
@@ -0,0 +1,443 @@
...
+                schema,
+                schema.statesToCreate().iterator().next());
+        platformSingletonSupplier = () -> new SingletonNode<>(
+                md.serviceName(),
+                md.stateDefinition().stateKey(),
+                md.singletonClassId(),
+                md.stateDefinition().valueCodec(),
+                PlatformState.newBuilder().build());
+
+        root.putServiceStateIfAbsent(md, platformSingletonSupplier);
```
artemananiev, line 82: I can't say for sure, but it doesn't look right. Any schema should be registered just once. In particular, this new `V0540PlatformStateSchema` is already registered in `Hedera`, and schema registration should have triggered platform state creation. Why is it needed here?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
```diff
@@ -0,0 +1,443 @@
...
+     *
+     * <p>
+     * When running in {@link AncientMode#GENERATION_THRESHOLD}, this value is the minimum generation non-ancient. When
+     * running in {@link AncientMode#BIRTH_ROUND_THRESHOLD}, this value is the minimum birth round non-ancient.
+     * </p>
+     *
+     * @return the ancient threshold after this round has reached consensus
+     */
+    public long getAncientThreshold() {
+        com.hedera.hapi.platform.state.ConsensusSnapshot consensusSnapshot =
```
artemananiev, line 248: Why fully qualified class name here?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/GenesisStateBuilder.java
imalygin: Actually, I realized that this code is not even needed. I can remove it entirely.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Agreed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
imalygin: Removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -240,9 +243,6 @@ public String getInfoString(final int hashDepth) {
      */
     @Override
     public String toString() {
-        return new ToStringBuilder(this)
-                .append("platformState", getPlatformState())
-                .append("swirldState", getSwirldState())
-                .toString();
+        return new ToStringBuilder(this).append("swirldState", getSwirldState()).toString();
```
litt3, line 246: As far as I can tell, `PlatformState` is still part of this class, right? Why was it removed from the `toString` method?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/SwirldState.java
```diff
@@ -44,15 +44,15 @@ public interface SwirldState extends MerkleNode {
      * </p>
      *
      * @param platform                the Platform that instantiated this state
-     * @param platformState           the platform state
+     * @param platformState   the platform state
```
litt3, line 47: wonky javadoc alignment

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/merkle/StateUtils.java
```diff
@@ -30,9 +30,14 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Objects;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 /** Utility class for working with states. */
 public final class StateUtils {
+
+    private static final Logger logger = LogManager.getLogger();
```
litt3, line 39: `logger` has no usages

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
litt3: > Other services may need access to platform state accessor for their purposes.

Do other services *currently* need access to platform state?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
```diff
@@ -0,0 +1,444 @@
...
+                return ((PlatformStateAccessorSingleton) source)
+                        .getPlatformState()
+                        .copyBuilder();
+            } else {
+                return convert(source).copyBuilder();
+            }
+        });
+    }
+
+    private void updatePlatformState(BiFunction<PlatformState, PlatformState.Builder, PlatformState.Builder> updater) {
```
litt3, line 423: The values in `PlatformState` are always set as a group, like this.

```
        platformState.setRound(round.getRoundNum());
        platformState.setConsensusTimestamp(round.getConsensusTimestamp());
        platformState.setCreationSoftwareVersion(softwareVersion);
        platformState.setRoundsNonAncient(roundsNonAncient);
        platformState.setSnapshot(round.getSnapshot());
```

It seems odd to be making all these copies and doing work every time a value is set, rather than just creating a single `update` method, that accepts all the `PlatformState` values as arguments

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateConverter.java
```diff
@@ -0,0 +1,246 @@
...
+import java.security.cert.X509Certificate;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * This class handles conversion from PBJ objects related to the platform state to the corresponding Java objects, and vice versa.
+ */
+public class PlatformStateConverter {
```
litt3, line 48: There aren't any unit tests for this class.

I think each conversion should be tested, to verify that you can go to and from the related types, and maintain equality

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
```diff
@@ -0,0 +1,444 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * This class is a drop-in replacement for {@link PlatformState} that uses singleton object under the hood.`
+ * It assumes that the instance of {@link MerkleRoot} passed in the constructor is a {@link MerkleStateRoot} instance.
+ * It registers the singleton object for the platform state in the {@link MerkleStateRoot} if it is not already present.
+ */
+public class PlatformStateAccessorSingleton implements PlatformStateAccessor {
```
litt3, line 54: given that the `get`s and `set`s on this class aren't simple `=` assignments, it might make sense to verify for each one whether a `get` after a `set` yields the expected value

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: For no good reason. I returned it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateLifecycles.java
```diff
@@ -51,21 +51,22 @@ public interface MerkleStateLifecycles {
...
      *
-     * @param state the working state of the network to be initialized
-     * @param platform the platform used by this node
+     * @param state                 the working state of the network to be initialized
+     * @param platform              the platform used by this node
      * @param platformState the working state of the platform
-     * @param trigger the reason for the initialization
-     * @param previousVersion if non-null, the network version that was previously in use
+     * @param trigger               the reason for the initialization
+     * @param previousVersion       if non-null, the network version that was previously in use
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessor.java
```diff
@@ -0,0 +1,244 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.time.Instant;
+
+/**
+ * This interface represents the platform state and provide access to the state's properties.
+ */
+public interface PlatformStateAccessor {
+
+    String PLATFORM_NAME = "PlatformState";
+    String PLATFORM_STATE_KEY = "PLATFORM_STATE";
```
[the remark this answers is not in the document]
imalygin: I'm not sure if there any particular reason for that but we have this pattern established.

tinker-michaelj (approved):
LGTM, ty @imalygin !

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeRecordHandler.java
```diff
@@ -183,7 +183,7 @@ private void logHbarFinalizationFailInvalid(
                         .map(accountId -> String.format(
                                 "\tOriginal : %s%n\tModified : %s",
                                 accountStore.getOriginalValue(accountId), accountStore.get(accountId)))
-                        .collect(Collectors.joining("%n")));
+                        .collect(Collectors.joining("\n")));
```
anthony-swirldslabs, line 186: While the change makes sense, I hope nobody parses the old log message today. If someone does, then they need changing too.

platform-sdk/platform-apps/demos/HelloSwirldDemo/src/main/java/com/swirlds/demo/hello/HelloSwirldDemoState.java
```diff
@@ -99,7 +99,7 @@ private HelloSwirldDemoState(final HelloSwirldDemoState sourceState) {
     }
 
     @Override
-    public synchronized void handleConsensusRound(final Round round, final PlatformState platformState) {
+    public synchronized void handleConsensusRound(final Round round, final PlatformStateAccessor platformState) {
```
anthony-swirldslabs, line 102: Why do we keep the `PlatformStateAccessor` around? Why not use the States API directly here? (I understand it's an override method, I'm asking about the base method and all the implementations.)
Or otherwise, why not use the Accessor everywhere instead of dealing with States API directly?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -109,21 +116,17 @@
...
 
-    //    private static final long CLASS_ID = 0x2de3ead3caf06392L;
-    // Uncomment the following class ID to run a mono -> modular state migration
-    // NOTE: also change class ID of ServicesState
     private static final long CLASS_ID = 0x8e300b0dfdafbb1aL;
-    private static final int VERSION_1 = 30;
-    private static final int CURRENT_VERSION = VERSION_1;
+    // Migrates from `PlatformState` to State API singleton
+    public static final int VERSION_2 = 31;
+    private static final int CURRENT_VERSION = VERSION_2;
```
anthony-swirldslabs, line 129: Just checking: we don't need to update the `CLASS_ID` too, do we?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -365,7 +375,24 @@ public void preHandle(@NonNull final Event event) {
...
      */
     @Override
-    public MerkleNode migrate(final int ignored) {
+    public MerkleNode migrate(final int version) {
+        if (version < VERSION_2) {
+            final var zerothChild = getChild(0);
+            if (!(zerothChild instanceof PlatformState platformState)) {
+                throw new IllegalStateException("Expected a PlatformState as the first child");
+            }
+            preV054PlatformState = platformState;
```
anthony-swirldslabs, line 384: Which `platformState` is it at line 384? It cannot be the one declared at line 381 because the latter is already out of scope here. It appears that you're reading the value of the class field instead of the zerothChild here, which may not be what you want to read here.

See https://blogs.oracle.com/javamagazine/post/java-pattern-matching-instanceof-flow-scoping

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessor.java
```diff
@@ -0,0 +1,243 @@
...
+import com.swirlds.platform.system.SoftwareVersion;
+import com.swirlds.platform.system.address.AddressBook;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.time.Instant;
+
+/**
+ * This interface represents the platform state and provide access to the state's properties.
+ */
+public interface PlatformStateAccessor {
```
anthony-swirldslabs, line 31: In addition to my other comment above, it would be best not to have this interface (I didn't realize you're introducing it in this PR, I thought it was an existing one), and instead always use the States API approach everywhere.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/impl/PbjConverter.java
```diff
@@ -0,0 +1,249 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class handles conversion from PBJ objects related to the platform state to the corresponding Java objects, and vice versa.
+ */
+public class PbjConverter {
```
anthony-swirldslabs, line 52: I'm just wondering why is this class necessary? If the PlatformState is modeled in protobuf, then PBJ already generates the codecs. Why do we need this converter?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -316,16 +328,14 @@ public SoftwareVersion getSoftwareVersion() {
     /**
      * {@inheritDoc}
      *
-     * <p>Called by the platform to either build a genesis state, or to deserialize a Merkle node in a saved state
-     * with the stable class id of the Services Merkle tree root during genesis (c.f. our constructor, in which
-     * we register this method as the factory for the {@literal 0x8e300b0dfdafbb1a} class id).
+     * <p>Called by the platform to either build a genesis state.
```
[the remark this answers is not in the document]
imalygin: :) fixed

platform-sdk/platform-apps/demos/HelloSwirldDemo/src/main/java/com/swirlds/demo/hello/HelloSwirldDemoState.java
imalygin: The idea is that in some cases we may want to have an alternative and simple implementation of the platform state (`PlatformState`). When it comes to the services code, the strong preference is for States API. This API is more flexible but it's less straightforward than `PlatformStateAccessor`. So we keep it around for the platform code.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: `CLASS_ID` should remain intact.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: It's actually `platformState` for that `if` statement
Take a look at this example from the link you provided
image

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessor.java
imalygin: That would complicate things by a lot with no particular gains. Also, we'd have to support all the "service" infrastructure for the demo apps, which is excessive. We do need two implementation of `PlatformStateAccessor` - one is backed by the platform state (self-sufficient, easy to create and maintain) and another backed by State API (much more flexible, aligned with the other services, but more difficult to maintain).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/impl/PbjConverter.java
imalygin: This class can be avoided but in this case I'd have to migrate multiple domain objects to PBJ. That would significantly extend the scope of this PR. Besides, I'm not sure if  that's a direction we're taking. That is, is the goal to get rid of all the domain object if favor of the objects defined in `hedera-protobuf` library?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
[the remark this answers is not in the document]
imalygin: Fixed

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/FinalizeRecordHandler.java
imalygin: We'll use "scream test" approach.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
imalygin: It looks like it's no longer relevant with Michael's change.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: No longer relevant.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
imalygin: No longer relevant.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
imalygin: No longer relevant.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
imalygin: would you consider current implementation reasonable?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
anthony-swirldslabs: Oh, I missed this part. Interesting. It doesn't look super intuitive though, so myself, I'd prefer to avoid this particular version of the pattern. But it's good to know there's not a bug.

tinker-michaelj (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/impl/PbjConverter.java
anthony-swirldslabs: Consistency may be one argument in favor of having everything in protobuf. But I see your reasoning regarding the size of the PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
artemananiev: Just for consistency. For example, a person may need to find all services (or all schemas from all services). To me, it's worth extracting to an outer class, even if that class is just 20 lines long. It may also be a good place for JavaDoc

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
```diff
@@ -180,12 +180,18 @@ public boolean startUserTransaction(
...
         // start a new block.
-        final var isFirstTransactionAfterFreezeRestart = platformState.getFreezeTime() != null
-                && platformState.getFreezeTime().equals(platformState.getLastFrozenTime());
+        final var isFirstTransactionAfterFreezeRestart = platformState.freezeTime() != null
+                && platformState.freezeTimeOrThrow().equals(platformState.freezeTime());
         if (isFirstTransactionAfterFreezeRestart) {
-            platformState.setFreezeTime(null);
+            final var writableStore =
+                    new WritablePlatformStateStore(state.getWritableStates(PlatformStateService.NAME));
+            writableStore.setFreezeTime(null);
```
artemananiev, line 194: Why isn't it like
```java
state.getWritableStates(PlatformStateService.NAME).getSingleton(PLATFORM_STATE_KEY).setFreezeTime(null);
```
?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -898,6 +925,11 @@ public void mapDeleteChange(@NonNull K key) {
         }
     }
 
+    @NonNull
+    public Function<SemanticVersion, SoftwareVersion> getVersionFactory() {
```
edward-swirldslabs, line 929: javadoc

platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/platform/PlatformTestingToolState.java
```diff
@@ -1334,11 +1331,6 @@ public void init(
 
         // initialize data structures used for FCQueue transaction records expiration
         initializeExpirationQueueAndAccountsSet();
-
-        logger.info(LOGM_DEMO_INFO, "Dual state received in init function {}", () -> new ApplicationDualStatePayload(
```
kfa-aguda, line 1338: i see the platform state was only being used here for logging purposes but do we not need this log line anymore? why?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateLifecycles.java
```diff
@@ -34,6 +34,14 @@
  * interface; but in the future will be callbacks registered with a platform builder.
  */
 public interface MerkleStateLifecycles {
+    /**
+     * Called when a {@link MerkleStateRoot} needs to ensure its {@link PlatformStateAccessor} implementation
+     * is initialized.
+     *
+     * @param state the root of the state to be initialized
+     */
+    void initPlatformState(@NonNull State state);
```
kfa-aguda, line 43: I love how it's so much easier to reason about `platform state` now that it's just a `state`!

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -198,11 +210,7 @@ public MerkleStateRoot() {
      * on restart, on reconnect, or any other time indicated by the {@code trigger}.
      */
     @Override
-    public void init(
-            final Platform platform,
-            final PlatformState platformState,
-            final InitTrigger trigger,
-            final SoftwareVersion deserializedVersion) {
+    public void init(final Platform platform, final InitTrigger trigger, final SoftwareVersion deserializedVersion) {
```
kfa-aguda, line 213: I see in many places, previously multi-lined arg list are now inlined. Not asking to change it, but we probably want to keep it consistent. A 3-args list should probably remain multi-lined.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -365,7 +375,24 @@ public void preHandle(@NonNull final Event event) {
      * {@inheritDoc}
      */
     @Override
-    public MerkleNode migrate(final int ignored) {
+    public MerkleNode migrate(final int version) {
+        if (version < VERSION_2) {
+            final var zerothChild = getChild(0);
+            if (!(zerothChild instanceof PlatformState platformState)) {
+                throw new IllegalStateException("Expected a PlatformState as the first child");
```
kfa-aguda, line 382: it's worth having a unit test specifically for this method, which should include this condition. I see references in SerializationTest but i don't think they test the  unhappy paths like this or the case version isn't < VERSION_2 or if child == null on line 391.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -155,13 +151,20 @@ public PlatformState getPlatformState() {
...
+     * Updates the platform state.
      *
-     * @param platformState the platform state
+     * @param platformStateAccessor the platform state
      */
     @Override
-    public void setPlatformState(@NonNull final PlatformState platformState) {
-        setChild(ChildIndices.PLATFORM_STATE, platformState);
+    public void updatePlatformState(@NonNull final PlatformStateAccessor platformStateAccessor) {
+        if (platformStateAccessor instanceof PlatformState platformState) {
```
kfa-aguda, line 160: probably obvious but i fail to understand why we need this check. PlatformStateAccessor has 3 impls - PlatformState, R/WStateStores. Are we throwing an exception if it's one of R/WStateStores? The tests i could find all use `PlatformState` so not sure.

kfa-aguda (commented):
Very helpful PR - concise refresher on how state access works. 
It however took a while to get through. I think it could have been broken down into two or more PRs. For example, we could have had one PR introducing the new interfaces and deprecating the PlatformState, then a subsequent PR which just updates the references. That or some other mechanism, as long as it's broken down.

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -0,0 +1,565 @@
...
+                .copyBuilder()
+                .lowestJudgeGenerationBeforeBirthRoundMode(lowestJudgeGenerationBeforeBirthRoundMode));
+    }
+
+    @Override
+    public void bulkUpdate(Consumer<PlatformStateAccessor> updater) {
+        final var accumulator = new PlaformStateValueAccumulator();
+        updater.accept(accumulator);
+        setAllFrom(accumulator);
+    }
```
litt3, line 191: won't this overwrite any values that weren't set in the bulk update?

 I think this bulk update logic needs some testing. We should make sure that everything we want to set gets set, and everything not specified remains unchanged.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/GenesisStateBuilder.java
```diff
@@ -34,31 +34,37 @@ public final class GenesisStateBuilder {
...
-        platformState.setCreationSoftwareVersion(appVersion);
-        platformState.setRound(0);
-        platformState.setLegacyRunningEventHash(null);
-        platformState.setConsensusTimestamp(Instant.ofEpochSecond(0L));
-
-        return platformState;
+            final long genesisFreezeTime = basicConfig.genesisFreezeTime();
+            if (genesisFreezeTime > 0) {
+                platformState.setFreezeTime(Instant.ofEpochSecond(genesisFreezeTime));
+            }
```
litt3, line 57: shouldn't this be operating on `v`, rather than `platformState`?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessor.java
```diff
@@ -0,0 +1,250 @@
...
+     *                                                  enabled
+     */
+    void setLowestJudgeGenerationBeforeBirthRoundMode(long lowestJudgeGenerationBeforeBirthRoundMode);
+
+    /**
+     * This is a convenience method to update multiple fields in the platform state in a single operation.
+     * @param updater a consumer that updates the platform state
+     */
+    void bulkUpdate(Consumer<PlatformStateAccessor> updater);
+}
```
litt3, line 250: missing nullity annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformState.java
```diff
@@ -526,7 +551,16 @@ public long getLowestJudgeGenerationBeforeBirthRoundMode() {
...
     public void setLowestJudgeGenerationBeforeBirthRoundMode(final long lowestJudgeGenerationBeforeBirthRoundMode) {
         this.lowestJudgeGenerationBeforeBirthRoundMode = lowestJudgeGenerationBeforeBirthRoundMode;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void bulkUpdate(Consumer<PlatformStateAccessor> updater) {
+        updater.accept(this);
```
litt3, line 564: param missing `final` and nullity annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -0,0 +1,565 @@
...
+    public void setLowestJudgeGenerationBeforeBirthRoundMode(final long lowestJudgeGenerationBeforeBirthRoundMode) {
+        final var previousState = stateOrThrow();
+        update(previousState
+                .copyBuilder()
+                .lowestJudgeGenerationBeforeBirthRoundMode(lowestJudgeGenerationBeforeBirthRoundMode));
+    }
+
+    @Override
+    public void bulkUpdate(Consumer<PlatformStateAccessor> updater) {
+        final var accumulator = new PlaformStateValueAccumulator();
```
litt3, line 188: param `updater` missing nullity annotation and final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -0,0 +1,565 @@
...
+
+    private void update(@NonNull final PlatformState stateBuilder) {
+        this.state.put(stateBuilder);
+        if (writableStates instanceof CommittableWritableStates committableWritableStates) {
+            committableWritableStates.commit();
+        }
+    }
+
+    private static class PlaformStateValueAccumulator implements PlatformStateAccessor {
+
```
litt3, line 209: I think this would benefit from a class doc, describing the use

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    private void update(@NonNull final PlatformState stateBuilder) {
+        this.state.put(stateBuilder);
+        if (writableStates instanceof CommittableWritableStates committableWritableStates) {
+            committableWritableStates.commit();
+        }
+    }
+
+    private static class PlaformStateValueAccumulator implements PlatformStateAccessor {
```
litt3, line 208: misspelled "plaform" in "PlaformStateValueAccumulator"

IMO this would be better as a standalone class. It's pretty big to fit as internal class

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessorSingleton.java
imalygin: I added `PlatformStateAccessor#bulkUpdate` to address this issue.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/services/ServicesRegistry.java
```diff
@@ -77,5 +77,12 @@ public int compareTo(@NonNull final Registration that) {
...
      */
-    void register(@NonNull final Service service);
+    void register(@NonNull Service service);
+
+    /**
+     * Returns a sub-registry containing only the services with the given names.
+     * @param serviceNames the names of the services to include in the sub-registry
+     * @return the sub-registry
+     */
+    ServicesRegistry subRegistryFor(@NonNull String... serviceNames);
```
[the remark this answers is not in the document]
imalygin: I think vararg is justified here as the idea is to take just a few services of all the services collection, and we know these services during compile time. Currently, we have only one use case for that and it's Platform State initialization (see `Hedera#initPlatformState`)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ids/schemas/V0490EntityIdSchema.java
```diff
@@ -67,11 +67,8 @@ public Set<StateDefinition> statesToCreate() {
     @Override
     public void migrate(@NonNull MigrationContext ctx) {
         final var entityIdState = ctx.newStates().getSingleton(ENTITY_ID_STATE_KEY);
-        final var config = ctx.configuration().getConfigData(HederaConfig.class);
-
-        final var isGenesis = ctx.previousVersion() == null;
-        if (isGenesis) {
-            // Set the initial entity id to the first user entity minus one
+        if (entityIdState.get() == null) {
```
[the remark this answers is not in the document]
imalygin: Yes, the assumption is that there may be other reasons for `entityIdState` to be null.

platform-sdk/platform-apps/tests/PlatformTestingTool/src/main/java/com/swirlds/demo/platform/PlatformTestingToolState.java
imalygin: The only reason for the `platformState` parameter to exist in this method across all the implementation was this particular line of code. It used to be relevant but given the fact that it mentions the dual state, it's no longer the case.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/GenesisStateBuilder.java
imalygin: Good catch. Yes, it should

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: It's a valid observation. I'll fix it and add more tests.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: The point of this check is that this class is not supposed to be used with the other implementation because `State` class is not used in Hedera app. However, we still need this class for demo apps.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: I basically let Spottles to decide how it should look like but I see your point. Let me update it

hedera-node/hedera-app/src/main/java/com/hedera/node/app/services/ServicesRegistry.java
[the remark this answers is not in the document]
imalygin: Fixed

hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java
```diff
@@ -201,13 +201,13 @@ public CompletableFuture<Void> extractSoftwareUpgrade(@NonNull final Bytes archi
...
-     * @param platformState the platform state
+     * @param platformStateStore the platform state
      * @return true if a freeze is scheduled, false otherwise
      */
-    public boolean isFreezeScheduled(final PlatformState platformState) {
-        requireNonNull(platformState, "Cannot check freeze schedule without access to the dual state");
-        final var freezeTime = platformState.getFreezeTime();
-        return freezeTime != null && !freezeTime.equals(platformState.getLastFrozenTime());
+    public boolean isFreezeScheduled(@NonNull final ReadablePlatformStateStore platformStateStore) {
+        requireNonNull(platformStateStore, "Cannot check freeze schedule without access to the dual state");
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/schemas/V0540PlatformStateSchema.java
```diff
@@ -0,0 +1,53 @@
...
+import com.hedera.hapi.node.base.SemanticVersion;
+import com.hedera.hapi.platform.state.PlatformState;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.state.spi.MigrationContext;
+import com.swirlds.state.spi.Schema;
+import com.swirlds.state.spi.StateDefinition;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Set;
+
+public class V0540PlatformStateSchema extends Schema {
```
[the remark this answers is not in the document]
imalygin: Added javadoc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -0,0 +1,195 @@
...
+     * but can be used to change and access any part of state that does not require the full {@link SoftwareVersion}.
+     * @param writableStates the writable states
+     */
+    public WritablePlatformStateStore(@NonNull final WritableStates writableStates) {
+        super(writableStates);
+        this.writableStates = writableStates;
+        this.state = writableStates.getSingleton(V0540PlatformStateSchema.PLATFORM_STATE_KEY);
+    }
+
+    public void setAllFrom(@NonNull final PlatformStateAccessor accessor) {
```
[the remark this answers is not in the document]
imalygin: Added missing javadoc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/impl/PbjConverter.java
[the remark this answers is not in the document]
imalygin: Added missing `final` modifier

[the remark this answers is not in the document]
imalygin: Added private constructor to the class

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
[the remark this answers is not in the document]
imalygin: Fair enough. I removed `VERSION_2` constant. Now there is `CURRENT_VERSION` only.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -365,7 +375,24 @@ public void preHandle(@NonNull final Event event) {
...
+            }
+            preV054PlatformState = platformState;
+            logger.info("Found pre-0.54 PlatformState, will migrate to State API singleton");
```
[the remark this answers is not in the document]
imalygin: Added `STARTUP` marker to this log statement

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/impl/BlockRecordManagerImpl.java
imalygin: Fixed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -537,17 +538,17 @@ public PlatformComponentBuilder buildComponentBuilder() {
...
-                            addressBookInitializer.getCurrentAddressBook().copy());
-
-            state.getPlatformState()
-                    .setPreviousAddressBook(
-                            addressBookInitializer.getPreviousAddressBook() == null
-                                    ? null
-                                    : addressBookInitializer
-                                            .getPreviousAddressBook()
-                                            .copy());
+            PlatformStateAccessor platformState = state.getPlatformState();
```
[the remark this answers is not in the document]
imalygin: Added `final` modifier

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateAccessor.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformState.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: Sure. I created standalone `PlaformStateValueAccumulator` class

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: Added javadoc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: Now `PlaformStateValueAccumulator` has a boolean flag per each field. If a field is updated, a corresponding flag is set to true. Then, there is an overloaded version of `PbjConverter#toPbjPlatformState` that checks the flags before update the corresponding fields.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: I added unit tests for this code

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/PlatformStateValueAccumulator.java
```diff
@@ -0,0 +1,484 @@
...
+
+    public boolean isAddressBookUpdated() {
+        return addressBookUpdated;
+    }
+
+    public boolean isPreviousAddressBookUpdated() {
+        return previousAddressBookUpdated;
+    }
+
+    public boolean isRoundUpdated() {
```
litt3, line 436: `isRoundUpdated` and `isConsensusTimestampUpdated` are both unused.

What is going on with these values? They are being set and used in different places throughout the code base, but aren't included in the protobuf?

litt3 (commented):
The size of this PR has me concerned. Personally, it's to the point where I'm lacking confidence in my ability to review such large and consequential changes thoroughly.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/PlatformStateValueAccumulator.java
imalygin: That's actually was the reason for a test failure. I fixed that.

litt3 (approved):

edward-swirldslabs (dismissed):

edward-swirldslabs (dismissed):

tinker-michaelj (dismissed):
LGTM, tyvm @imalygin !

artemananiev (approved):

litt3 (approved):

mxtartaglia-sl (approved):

MiroslavGatsanoga (dismissed):

rbarkerSL (approved):

mxtartaglia-sl (dismissed):

litt3 (approved):

Neeharika-Sompalli (approved):

tinker-michaelj (approved):

mxtartaglia-sl (approved):

rbarkerSL (approved):

PR #14560, opened 2024-08-02
chore: ensure events with no transactions are provided to the app

### LINKED ISSUE #14384: Ensure events with no transactions get passed to the app
Block streams need all events, even those with no transactions. We must ensure that all events gets passed to the app, even if there are no transactions in them.

lpetrovic05:
closes #14384


edward-swirldslabs (approved):
Is this PR just adding a test to ensure we detect when this is violated?  Or is there missing production code that turns on the capability?

lpetrovic05:
> Is this PR just adding a test to ensure we detect when this is violated? Or is there missing production code that turns on the capability?

Just ensuring that this happens, it was already the case that this was true

PR #14561, opened 2024-08-02
chore: Add EventCore and Signature to Event

### LINKED ISSUE #13998: Provide API for apps to consume `EventCore` and `Signature`
depends on #13861

The `Event` interface should include getters for `EventCore` and `signature` so that this data can become part of the block stream.

timo0:
Fix #13998


lpetrovic05 (approved):

tinker-michaelj (approved):
LGTM, ty @timo0 !

PR #14566, opened 2024-08-02
chore: Add hash to TransactionWrapper

### LINKED ISSUE #14385: Include hash of payload in payloadWrapper
Since Hedera will need to hash payloads/transaction, and the platform does this when hashing events, we should include the hash of the payload in the payload wrapper. This way, Hedera can avoid hashing it again, thus saving CPU time.

Must be done after [this task](https://github.com/hashgraph/hedera-services/issues/14426), because the payload wrapper does not currently store the `EventPayload`, just a `OneOf`.

timo0:
Fix #14385


david-bakin-sl (approved):
Not sure why smart contracts is affected here, but ok.  LGTM

lpetrovic05 (approved):

PR #14575, opened 2024-08-02
feat: added some failure tests for airdrop token to invalid address a…

povolev15:
Description:

Airdrop a token to an invalid address

Attempt to airdrop a token from an invalid address

EOA airdrops an invalid token should fail

EOA airdrops multiple duplicated NFT

EOA with enough token balance signs and pays for a tokenAirdropTrsnsaction when they have no HBAR to pay for the transaction

EOA airdrops more than 10 tokens to multiple accounts

Related issue(s):

Fixes https://github.com/hashgraph/hedera-services/issues/14528
Fixes https://github.com/hashgraph/hedera-services/issues/14529
Fixes https://github.com/hashgraph/hedera-services/issues/14530
Fixes https://github.com/hashgraph/hedera-services/issues/14531
Fixes https://github.com/hashgraph/hedera-services/issues/14577
Fixes https://github.com/hashgraph/hedera-services/issues/14576

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
@@ -18,7 +18,6 @@ package com.hedera.node.app.service.token.impl.validators;
 
 import static com.hedera.hapi.node.base.ResponseCodeEnum.AMOUNT_EXCEEDS_ALLOWANCE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INSUFFICIENT_TOKEN_BALANCE;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_AMOUNTS;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_NFT_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TOKEN_ID;
--- a/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenAirdropHandlerTest.java
+++ b/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenAirdropHandlerTest.java
@@ -18,7 +18,6 @@ package com.hedera.node.app.service.token.impl.test.handlers;
 
 import static com.hedera.hapi.node.base.ResponseCodeEnum.EMPTY_TOKEN_TRANSFER_ACCOUNT_AMOUNTS;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_AMOUNTS;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TRANSACTION_BODY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.OK;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_NOT_ASSOCIATED_TO_ACCOUNT;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED;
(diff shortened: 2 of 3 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -809,6 +811,153 @@ final Stream<DynamicTest> spenderHasMoreAllowancesThatTheOwner() {
...
+                            .treasury(ALICE)
+                            .tokenType(FUNGIBLE_COMMON)
+                            .initialSupply(15L),
+                    tokenAirdrop(moving(10L, FUNGIBLE_TOKEN_A).between(ALICE, "0.0.999999999999999"))
+                            .signedByPayerAnd(ALICE)
+                            .hasKnownStatus(INVALID_ACCOUNT_ID));
+        }
+
+        @HapiTest
+        @DisplayName("transfer fungible token to from account")
```
david-bakin-sl, line 832: "to from account" -> "from invalid account" ?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -809,6 +811,153 @@ final Stream<DynamicTest> spenderHasMoreAllowancesThatTheOwner() {
...
+                    cryptoCreate(ALICE).balance(ONE_HUNDRED_HBARS),
+                    tokenAssociate(ALICE, NON_FUNGIBLE_TOKEN),
+                    tokenAirdrop(TokenMovement.movingUniqueWithAllowance(NON_FUNGIBLE_TOKEN, 1L)
+                                    .between(ALICE, "0.0.999999999999999"))
+                            .signedByPayerAnd(ALICE, OWNER)
+                            .hasKnownStatus(INVALID_ACCOUNT_ID));
+        }
+
+        @HapiTest
+        @DisplayName("transfer NFT to from account")
```
david-bakin-sl, line 861: and here too

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiPropertySource.java
```diff
@@ -225,6 +225,12 @@ static AccountID asAccount(String v) {
                 .build();
     }
 
+    static AccountID asAliasAccount(String v) {
+        return AccountID.newBuilder()
+                .setAlias(ByteString.copyFrom(v.getBytes()))
```
david-bakin-sl, line 230: I don't get it.  `String.getBytes` doesn't take the hex value and turn it into a number.  It just takes the characters of the string.  So that must mean that having an alias that consists not of bytes but of the hex-encoded bytes, since what is passed here (in the test below) is a string that looks like `0x0000000000000000000000691752902764108185`.  Am I understanding this right?

david-bakin-sl (commented):
Have a question - something I don't understand.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -809,6 +811,153 @@ final Stream<DynamicTest> spenderHasMoreAllowancesThatTheOwner() {
...
+                                                .setTokenNum(5555555L)
+                                                .build());
+                    }),
+                    tokenAirdrop(moving(50L, FUNGIBLE_TOKEN_A).between(ALICE, BOB))
+                            .signedByPayerAnd(ALICE)
+                            .hasKnownStatus(INVALID_TOKEN_ID));
+        }
+
+        @HapiTest
+        @DisplayName("transfer fungible token to incorrect alias")
```
ibankov, line 930: Display name seems wrong

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiPropertySource.java
david-bakin-sl: well ... maybe it's just me ... but maybe a _different_ way of writing an invalid alias such that it doesn't _look_ like you're trying hard to do a hex-encoding->binary conversion?

david-bakin-sl (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiPropertySource.java
david-bakin-sl: ok.  then, final suggestion if you'd like: Put a comment somewhere to say what kind of misuse of bytes is going on here so that the alias is invalid?  Or whatever is actually happening here?

david-bakin-sl: I mean, now I'm confused as to _why_ this alias is invalid.  Is it becauase it is an ascii string representing a hex encoding, or is that the way it is done and the invalid part is that the alias specified `0x0000000000000000000000691752902764108185` doesn't exist?

(If this alias field is _supposed_ to be the ascii string which is the hex-encoding of something, where is _that_ specified? I mean, it doesn't seem to be the base32url encoding specified by HIP-32, so what is it?)

vtronkov (approved):

PR #14584, opened 2024-08-03
chore: migrate `StateChangeListener` to States API

### LINKED ISSUE #14580: Migrate `StateChangesListener` to `c.s.state.spi` package
### Problem

The `StateChangesListener` on the preview block streams branch (see [here](https://github.com/hashgraph/hedera-services/pull/14579)) is being managed by the application's `WrappedState` class, which adds a potentially dangerous level of separation between the changes committed to the Merkle tree and the changes observed by the listener.

### Solution

Migrate the `StateChangesListener` in an appropriate form to the `c.s.state.spi` package.

### Alternatives

_No response_

### LINKED ISSUE #14583: `WritableQueueStateBase` misses case of items added and read prior to `commit()`
### Description

Although no production use case exhibits this pattern, we should fix `WritableQueueStateBase#commit()` to handle the case of items that are both added and read from the queue prior to a single call to `commit()`.

### Steps to reproduce

I.e. this fails with `NoSuchElementException`:
```
        @Test
        void commitAfterRemovingAllAddedElementsDoesNotThrow() {
            final var backingList = new LinkedList<String>();
            final var subject = new ListWritableQueueState<>(STEAM_STATE_KEY, backingList);
            subject.add(ART);
            subject.add(BIOLOGY);
            subject.removeIf(s -> true);
            subject.removeIf(s -> true);
            assertThatCode(subject::commit).doesNotThrowAnyException();
        }
```

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #14580 
 - Closes #14583 
 - Migrates the listener API from the preview block streams working branch

### DIFF
--- a/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
+++ b/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
@@ -16,6 +16,7 @@
 
 package com.swirlds.state;
 
+import com.swirlds.state.spi.CommittableWritableStates;
 import com.swirlds.state.spi.ReadableKVState;
 import com.swirlds.state.spi.ReadableStates;
 import com.swirlds.state.spi.WritableKVState;
@@ -48,4 +49,16 @@ public interface State {
      */
     @NonNull
     WritableStates getWritableStates(@NonNull String serviceName);
+
+    /**
+     * Registers a listener to be notified on each commit if the {@link WritableStates} created by this {@link State}
+     * are marked as {@link CommittableWritableStates}. Implementations need not support unregistering listeners, as
+     * there is no real case that a client would want to be notified of only some commits made to the state.
+     *
+     * @param listener The listener to be notified.
+     * @throws UnsupportedOperationException if the state does not support listeners.
+     */
+    default void registerCommitListener(@NonNull final StateChangeListener listener) {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/KVChangeListener.java
+++ b/platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/KVChangeListener.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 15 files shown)


platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
```diff
@@ -48,4 +49,16 @@ public interface State {
...
+    /**
+     * Registers a listener to be notified on each commit if the {@link WritableStates} created by this {@link State}
+     * are marked as {@link CommittableWritableStates}. Implementations need not support unregistering listeners, as
+     * there is no real case that a client would want to be notified of only some commits made to the state.
+     *
+     * @param listener The listener to be notified.
+     * @throws UnsupportedOperationException if the state does not support listeners.
+     */
+    default void registerCommitListener(@NonNull final StateChangeListener listener) {
+        throw new UnsupportedOperationException();
```
hendrikebbers, line 62:
```suggestion
        throw new UnsupportedOperationException("State of type '" + getClass() + "' does not support listener");
```

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/KVChangeListener.java
```diff
@@ -0,0 +1,43 @@
...
+ * @param <V> The type of the value
+ */
+public interface KVChangeListener<K, V> {
+    /**
+     * Called when an entry is added in to a map.
+     *
+     * @param key The key added to the map
+     * @param value The value added to the map
+     */
+    void mapUpdateChange(@NonNull K key, @NonNull V value);
```
hendrikebbers, line 35: Does it make sense to define the methods as empty default methods?

hendrikebbers (approved):

imalygin (approved):

edward-swirldslabs (approved):

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
tinker-michaelj: Ah yes, will add this in next PR. Makes sense.

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/spi/KVChangeListener.java
tinker-michaelj: For the data structure-specific listener types, it seemed any implementation should have a response for each type of change.

PR #14587, opened 2024-08-04
fix: update `TokenReconciliationValidator` for non-fungible types

### LINKED ISSUE #14586: `TokenReconciliationValidator` not correct for non-fungible token types
### Description

`TokenReconciliationValidator` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/TokenReconciliationValidator.java#L57) is incorporating balance adjustments that represent the # of NFTs owned of a deleted token type before dissociation.

_However_, it is not also counting # of owned NFTs based on previous mints, burns, and transfers.

### Steps to reproduce

Run the validator after [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/HRCPrecompileSuite.java#L90) `@HapiTest`. 

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #14586


Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

PR #14588, opened 2024-08-04
chore: Add `TransactionReceipt` cache

### LINKED ISSUE #14541: Reduce size of transaction receipt cache
### Problem

We currently use a `QueueState<TransactionRecordEntry>` in state; but this is significantly more information than is required to classify and handle duplicate transactions deterministically. 

### Solution

Define two new messages,
```
message TransactionReceiptEntry {
     /**
     * The ID of the node that submitted the transaction to consensus. The ID is the ID of the node as known by the
     * address book. Valid node IDs are in the range 0..2^63-1, inclusive.
     */
    int64 node_id = 1;

    /**
     * The id of the submitted transaction.
     */
    TransactionID transaction_id = 2;

    /**
     * The resulting status of handling the transaction.
     */
    ResponseCodeEnum status = 3;
}
```

And,
```
message TransactionReceiptEntries {
    repeated TransactionReceiptEntry entries;     
}
```

Use instead an `QueueState<TransactionReceiptEntries>` where each round accumulates changes for one new `TransactionReceiptEntries`; and we purge entire rounds of receipts at the point the youngest receipt (the latest consensus time) in the round is expired.

### Alternatives

_No response_

Neeharika-Sompalli:
Fixes #14541


hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
```diff
@@ -131,18 +146,18 @@ public RecordCacheImpl(
...
 
     /**
-     * Rebuild the internal data structures based on the current working state. Called during startup and during
-     * reconnect. The amount of time it takes to rebuild this data structure is not dependent on the size of state, but
-     * rather, the number of transactions in the queue (which is capped by configuration at 3 minutes by default).
+     * The data structures can't be created from the state during restart or reconnect.
+     * They are just cleared during restart or reconnect.
      */
-    public void rebuild(final WorkingStateAccessor workingStateAccessor) {
+    public void rebuild() {
```
derektriley, line 160: With the change to the queue being TransactionReceiptEntry (previously TransactionRecordEntry) when rebuilding the queue the asTxnRecord only build TransactionRecord's with the receipt and transactionID. I'm just thinking that we may need to do something to include the other information in the TransactionRecord otherwise it's lost and CryptoGetAccountRecordsQuery and TransactionGetRecordQuery will return mostly empty TransactionRecords. Thinking about this more.

Neeharika-Sompalli:
Opening PR to develop. Closing this.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/RecordCacheImpl.java
Neeharika-Sompalli: As discussed, we will not rely on queries returning correct values as per the issue

PR #14592, opened 2024-08-05
chore: flatten event impl

### LINKED ISSUE #14555: Flatten EventImpl
`EventImpl` has a purpose of wrapping a `PlatformEvent` and adding parent pointers as well as consensus metadata. Its structure should be flattened. `EventMetadata` & `ConsensusData` have fields which should be brought into `EventImpl` and these classes can be deleted.

lpetrovic05:
closes #14555


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/CesEvent.java
```diff
@@ -95,7 +96,7 @@ public static void serialize(
         // some fields used to be part of the stream but are no longer used
         // in order to maintain compatibility with older versions of the stream, we write a constant in their place
 
-        out.writeInt(ConsensusData.CLASS_VERSION);
+        out.writeInt(CONSENSUS_DATA_CLASS_VERSION);
```
litt3, line 99: This is the version number from the old consensus data class, but it's not actually used anywhere, right? Is there any reason not to just write an `UNDEFINED` here, as is done for `ConsensusData.generation` and `ConsensusData.roundCreated`?

litt3 (approved):

hendrikebbers (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/CesEvent.java
lpetrovic05: Probably it would work, but I tried to not change the format of the CES in order to avoid any unforeseen consequences. I don't think it makes much of a difference considering we are hoping to get rid of the CES as soon as we can.

PR #14608, opened 2024-08-05
fix: only skip event from missing creator

### LINKED ISSUE #14607: Removed node events should be skipped, not rounds
### Description

Skipping events from a missing node should be isolated to events, not rounds. (Even if in practice a skipped event implies the whole round will be skipped.)

### Steps to reproduce

Run [this embedded test](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/records/DuplicateManagementTest.java#L158), observe that there is no `endRound()` call in the dynamic context of `handleConsensusRound()` for the transaction submitted with a missing node id. 

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #14607

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/HgcaaLogValidator.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/HgcaaLogValidator.java
@@ -55,6 +55,8 @@ public class HgcaaLogValidator {
         private static final List<List<String>> PROBLEM_PATTERNS_TO_IGNORE = List.of(
                 List.of("not in the address book"),
                 List.of("Specified TLS cert", "doesn't exist"),
+                // Stopping an embedded node can interrupt signature verification of background traffic
+                List.of("Interrupted while waiting for signature verification"),
                 List.of("Could not start TLS server, will continue without it"),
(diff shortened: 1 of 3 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/HgcaaLogValidator.java
```diff
@@ -55,6 +55,8 @@ private static class ProblemTracker {
         private static final List<List<String>> PROBLEM_PATTERNS_TO_IGNORE = List.of(
                 List.of("not in the address book"),
                 List.of("Specified TLS cert", "doesn't exist"),
+                // Stopping an embedded node can interrupt signature verification of background traffic
+                List.of("Interrupted while waiting for signature verification"),
```
david-bakin-sl, line 59: Huh, so that's an example of when and where you see `InterruptedException`.

david-bakin-sl (approved):
LGTM

iwsimon (approved):
LGTM. Thanks @tinker-michaelj

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #14616, opened 2024-08-06
fix: get notification listeners up-to-date before `RECONNECT_COMPLETE`

### LINKED ISSUE #14615: Should register/unregister listeners in `onStateInitialized()`
### Description

As can be seen in [this](https://scans.gradle.com/s/6dzmgl6rkua3i/tests/task/:test-clients:test/details/com.hedera.services.bdd.suites.validation.LogValidationTest/logsContainNoUnexpectedProblems()%5B1%5D?top-execution=1) build scan, it is possible the `ReconnectListener` registered with platform notifications is **not** the listener constructed during the latest call to `SwirldState#init()`. 

(Note this manifested because the `WorkingStateAccessor` did not contain the reconnect state, but a state whose readable states were not initialized; we should actually just use the reconnect state from the `ReconnectCompleteNotification`.) 

### Steps to reproduce

Run `DabEnabledUpgradeTest` many times.

### Additional context

_No response_

### Hedera network

other

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #14615 
 - When constructing the app Dagger component, _if_ it is already non-null, then:
     * First, unregister the old component's listeners with the platform `NotificationEngine`; then,
     * Immediately register the new component's listeners.
 - This ensures that the listener that receives [this](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/ReconnectStateLoader.java#L165) `ReconnectCompleteNotification` is for the latest Dagger component.
 - Also switches `ReconnectListener` to simply use the state from the notification instead of `WorkingStateAccessor`.

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
@@ -82,6 +82,7 @@ import com.swirlds.common.platform.NodeId;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.metrics.api.Metrics;
 import com.swirlds.platform.listeners.PlatformStatusChangeListener;
+import com.swirlds.platform.listeners.PlatformStatusChangeNotification;
 import com.swirlds.platform.listeners.ReconnectCompleteListener;
 import com.swirlds.platform.listeners.StateWriteToDiskCompleteListener;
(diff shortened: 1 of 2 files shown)


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #14632, opened 2024-08-06
test: Token not associated after pending airdrop

### LINKED ISSUE #14631: (AIRDROP - 42) Given a receiver has a pending airdrop when the sender signs a cryptoTransfer to the receiver for the airdropped token then the airdrop should remain in pending and the cryptoTransfer should fail with error message TOKEN_NOT_ASSOCIATED_TO_ACCOUNT

### LINKED ISSUE #14633: (AIRDROP - 47) CryptoTransfer to an account not associated to the token and with no auto association slots left should result in a failing cryptoTransfer and no airdrops should be created.

vtronkov:
**Description**:
Added a test to validate that after a pending airdrop the receiver is not associated with the token

**Related issue(s)**:
Fixes #14631
Fixes #14633


stoyanov-st (approved):
LGTM

PR #14641, opened 2024-08-06
test: airdrop to deleted accounts

### LINKED ISSUE #14640: (AIRDROP - 60) Given sender attempts to airdrop a token when the receiver account is а deleted ЕОА then the tokenAirdropTransaction should fail. - deleted ED25519 account - deleted ECDSA account

vtronkov:
**Description**:
Adding tests airdropping to different types of deleted accounts

**Related issue(s)**:

Fixes #14640


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -894,7 +958,9 @@ private static SpecOperation[] setUpTokensAndAllReceivers() {
                                 ByteString.copyFromUtf8("c"),
                                 ByteString.copyFromUtf8("d"),
                                 ByteString.copyFromUtf8("e"),
-                                ByteString.copyFromUtf8("f"))),
+                                ByteString.copyFromUtf8("f"),
```
david-bakin-sl, line 961: Getting to the point where you could almost consider:

```java
mintToken(NON_FUNGIBLE_TOKEN, 
          "abcdefgh".codePoints()
                    .mapToObj(c->String.valueOf((char)c))
                    .map(ByteString::copyFromUtf8)
                    .toList());
```
Eh, never mind.  Now I look at it isn't clearer.

david-bakin-sl (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
stoyanov-st: Can I suggest this if the list goes on?
```suggestion
                              IntStream.rangeClosed('a', 'h')
                .mapToObj(i -> ByteString.copyFromUtf8(String.valueOf((char) i)))
                .toList();
```

vtronkov: Those are good suggestions, thank you guys! We were thinking about creating a method that will mint a token keep an internal counter and return the serial number as a result. That way you can just call this method and don't worry about the serial number or the mint. We will think about it and we might implement it in a refactoring story and make it more generic for everyone to use.

PR #14647, opened 2024-08-06
feat: create translator for contract create transaction

### LINKED ISSUE #14646: Create `ContractCreateTranslator`

petreze:
**Description**:
Translator for contract create transaction to convert block items to record stream entries. Integrate `TransactionRecordParityValidator` to validate the output

**Related issue(s)**:

Fixes #14646 

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/test/java/com/hedera/services/bdd/utils/test/translators/ContractCreateTranslatorTest.java
```diff
@@ -0,0 +1,110 @@
...
+        MockitoAnnotations.openMocks(this);
+        translator = new ContractCreateTranslator();
+    }
+
+    @Test
+    void testTranslateContractCreate() {
+        final ContractID contractID = ContractID.newBuilder().contractNum(5).build();
+        final Bytes evmAddress = Bytes.fromHex("badcadfaddad2bedfedbeef959feedbeadcafecadecedebeed4acedecada5ada");
+        final List<TransactionSidecarRecord> sidecars = List.of(TransactionSidecarRecord.DEFAULT);
+        // Arrange
```
MiroslavGatsanoga, line 78: **nitpick:** I think we use `given/when/then` in‏ other unit tests in the project and even if both are kind of similar, in order to be consistent I would use `given/when/then` pattern instead.

petreze: Agreed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ContractCreateTranslator.java
```diff
@@ -0,0 +1,49 @@
...
+        final var txnOutput = transaction.output();
+        if (txnOutput != null && txnOutput.hasContractCreate()) {
+            final var contractCreateResult = txnOutput.contractCreate().contractCreateResult();
+            receiptBuilder.contractID(contractCreateResult.contractID());
+            recordBuilder
+                    .receipt(receiptBuilder.build())
+                    .contractCreateResult(contractCreateResult)
+                    .evmAddress(contractCreateResult.evmAddress());
+        }
+        recordBuilder.consensusTimestamp(stateChanges.consensusTimestamp());
```
mhess-swl, line 42: Again, would be good to keep this in `BlockStreamTransactionTranslator`

mhess-swl (approved):
Looks good to me :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ContractCreateTranslator.java
petreze: Done, thank you!

MiroslavGatsanoga (approved):
LGTM

mhess-swl (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ContractCreateTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+
+        final var txnOutput = transaction.output();
+        if (txnOutput != null && txnOutput.hasContractCreate()) {
+            final var contractCreateResult = txnOutput.contractCreate().contractCreateResult();
+            receiptBuilder.contractID(contractCreateResult.contractID());
+            recordBuilder
+                    .receipt(receiptBuilder.build())
+                    .contractCreateResult(contractCreateResult)
+                    .evmAddress(contractCreateResult.evmAddress());
+        }
```
mhess-swl, line 43: Nit: I know this isn't modeled in the base translator, but we probably want an `else` statement here–and for any translator–that isn't able to find the data it needs in order to translate the block items; at the very least a log statement (since this is test code anyway, it won't clutter any prod logs). We use `HederaFunctionality` to determine which translator we need; therefore, if this code doesn't execute, I suspect we'll usually (always?) want to know why those items weren't there

mhess-swl (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ContractCreateTranslator.java
petreze: Addressed and added to other translator, thanks! Please re-review @mhess-swl

Issue #14567, opened 2024-08-02
Interrupt reading from data files gracefully when the data source is closed


artemananiev:
Closing MerkleDb data source should be improved. When it happens during reconnects, various exceptions may be thrown, e.g.

```
2024-08-01 14:48:47.620 6204     INFO  MERKLE_DB        <<reconnect: reconnect-controller>> MerkleDbDataSource: Closing Data Source [TokenService.TOKEN_RELS]
2024-08-01 14:48:47.635 6206     ERROR EXCEPTION        <<merkledb: Compacting #0>> CompactionTask: [TokenService.TOKEN_RELSHashStoreDisk] Compaction failed
java.lang.NullPointerException: Cannot invoke "java.nio.channels.FileChannel.size()" because the return value of "java.util.concurrent.atomic.AtomicReferenceArray.get(int)" is null
```

Sometimes it may result in a JVM crash:

```
Current thread (0x000056128c77ec10):  JavaThread "<merkledb: HalfDiskHashMap Flushing #19>" daemon [_thread_in_Java, id=419, stack(0x00007fb7638f9000,0x00007fb7639fa000) (1028K)]Stack: [0x00007fb7638f9000,0x00007fb7639fa000],  sp=0x00007fb7639f87d0,  free space=1021k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
J 12950 c2 com.swirlds.merkledb.files.DataFileCollection.retryReadUsingIndex(Lcom/swirlds/merkledb/collections/LongList;JLcom/swirlds/base/function/CheckedFunction;)Ljava/lang/Object; (103 bytes) @ 0x00007fc12d7b085c [0x00007fc12d7b0700+0x000000000000015c]
J 21237 c2 com.swirlds.merkledb.files.hashmap.HalfDiskHashMap$$Lambda+0x00007fb8d66f4358.run()V (20 bytes) @ 0x00007fc12cf6df08 [0x00007fc12cf6dde0+0x0000000000000128]
J 13543% c2 java.util.concurrent.ThreadPoolExecutor.runWorker(Ljava/util/concurrent/ThreadPoolExecutor$Worker;)V java.base@21.0.1 (187 bytes) @ 0x00007fc12d8e35bc [0x00007fc12d8e3400+0x00000000000001bc]
j  java.util.concurrent.ThreadPoolExecutor$Worker.run()V+5 java.base@21.0.1
[...]

PR #14654, opened 2024-08-06
fix: 14567: Interrupt reading from data files gracefully when the data source is closed

artemananiev:
Fix summary:

* `MerkleDbDataSource.saveRecords()` now handles dirty and deleted leaves in a separate thread / executor
* On close, MerkleDb data source shuts down this executor and waits for it to complete. This is required to make sure no read / write operations are running, when data files and indices are closed
* Compaction coordinator is also improved to wait till all compaction tasks are fully stopped on shutdown
* Minor improvements like closing data files before the corresponding indices

Testing:

* A new hammer test is introduced

Fixes: https://github.com/hashgraph/hedera-services/issues/14567

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
@@ -57,7 +57,7 @@ class MerkleDbCompactionCoordinator {
     private static final Logger logger = LogManager.getLogger(MerkleDbCompactionCoordinator.class);
 
     // Timeout to wait for all currently running compaction tasks to stop during compactor shutdown
-    private static final long SHUTDOWN_TIMEOUT = 60_000;
+    private static final long SHUTDOWN_TIMEOUT_MILLIS = 60_000;
 
     /**
      * An executor service to run compaction tasks. Accessed using {@link #getCompactionExecutor()}.
@@ -215,11 +215,15 @@ class MerkleDbCompactionCoordinator {
         // Wait till all the tasks are stopped
         final long now = System.currentTimeMillis();
         try {
-            while ((tasksRunning.get() != 0) && (System.currentTimeMillis() - now < SHUTDOWN_TIMEOUT)) {
+            while ((tasksRunning.get() != 0) && (System.currentTimeMillis() - now < SHUTDOWN_TIMEOUT_MILLIS)) {
                 Thread.sleep(1);
             }
         } catch (final InterruptedException e) {
             logger.warn(MERKLE_DB.getMarker(), "Interrupted while waiting for compaction tasks to complete", e);
+        }
+        // If some tasks are still running, there is nothing else to than to log it
+        if (tasksRunning.get() != 0) {
+            logger.error(MERKLE_DB.getMarker(), "Failed to stop all compactions tasks");
         }
     }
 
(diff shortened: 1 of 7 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
```diff
@@ -50,10 +51,14 @@
  * The number of threads in the pool is defined by {@link MerkleDbConfig#compactionThreads()} property.
  *
  */
+@SuppressWarnings("rawtypes")
 class MerkleDbCompactionCoordinator {
 
     private static final Logger logger = LogManager.getLogger(MerkleDbCompactionCoordinator.class);
 
+    // Timeout to wait for all currently running compaction tasks to stop during compactor shutdown
+    private static final long SHUTDOWN_TIMEOUT = 60_000;
```
anthony-swirldslabs, line 60: It's best to include the units of the value in the name of the constant, e.g. by adding a `_MILLIS` suffix.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
```diff
@@ -195,12 +204,22 @@ public void resumeCompaction() throws IOException {
...
+        }
+        // Wait till all the tasks are stopped
+        final long now = System.currentTimeMillis();
+        try {
+            while ((tasksRunning.get() != 0) && (System.currentTimeMillis() - now < SHUTDOWN_TIMEOUT)) {
+                Thread.sleep(1);
+            }
+        } catch (final InterruptedException e) {
+            logger.warn(MERKLE_DB.getMarker(), "Interrupted while waiting for compaction tasks to complete", e);
         }
```
anthony-swirldslabs, line 223: We should also check the main condition (the num of tasks running) after the loop finishes and log some diagnostics (or maybe error out) if it's still not satisfied after the timeout.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -318,8 +318,8 @@ public LongSummaryStatistics getFilesSizeStatistics() {
      */
     @Override
     public void close() throws IOException {
-        bucketIndexToBucketLocation.close();
         fileCollection.close();
+        bucketIndexToBucketLocation.close();
```
anthony-swirldslabs, line 322: If the order of the calls really matters that much, it might be worth adding a comment in code mentioning why.

artemananiev: OK

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
artemananiev: OK, let me add a log message, if some tasks are still running after the loop

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
artemananiev: OK

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbCompactionCoordinator.java
anthony-swirldslabs: Just to be super clear, in this case you want to add it after the try/catch block, so that the swallowed exception doesn't get in the way of the diagnostics.

anthony-swirldslabs (approved):

imalygin (approved):

PR #14662, opened 2024-08-07
feat: Implement a block items transaction record validator

### LINKED ISSUE #14501: Implement a `TransactionRecord` validator

mhess-swl:
Here's a first pass at a transaction record validator that can compare block items to an existing V6 record stream. It does so by leveraging the functionality already built into the `RcDiff` tool.

Note that there are still open questions (marked with `TODO`s) in the code. 

Closes #14501


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
```diff
@@ -34,7 +33,7 @@
  * @param output the output (if any) of processing the user transaction
  */
 public record SingleTransactionBlockItems(
-        @NonNull Transaction txn, @NonNull TransactionResult result, @Nullable TransactionOutput output) {
+        @Nullable Transaction txn, @Nullable TransactionResult result, @Nullable TransactionOutput output) {
```
MiroslavGatsanoga, line 36: All of those nullables would be for the first pass only and in the future we'll need to revert to requiring non-null, am I right? If yes, then adding a TODO here as well might provide useful.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    private static Timestamp fromTimestamp(final com.hedera.hapi.node.base.Timestamp timestamp) {
+        return Timestamp.newBuilder()
+                .setSeconds(timestamp.seconds())
+                .setNanos(timestamp.nanos())
+                .build();
+    }
+
+    private record BlockRecordsInput(List<SingleTransactionBlockItems> txns, List<StateChanges> allStateChanges) {}
```
MiroslavGatsanoga, line 143: Having "Input" in the name got me confused for a moment leading me to think that it will contain only input block items.
Perhaps a name without "Input" in it might be better e.g. BlockItems?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/SingleTransactionBlockItems.java
mhess-swl: I'm not sure. The sample tar file we're given has a bunch of blocks without these values. I'm not sure if these blocks are valid, or if they're just test objects until we generate more correct blocks

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
mhess-swl: Agreed, "Input" in the name is confusing. I went with `BlocksData`, how's that?

MiroslavGatsanoga: Sounds good to me, thanks!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
```diff
@@ -88,72 +112,215 @@ public SingleTransactionRecord translate(
...
+                }
+            }
+            recordBuilder.setMemo(memo);
+        }
+
+        final var txnBytes = toBytesForHash(txn);
+        final var hash = txnBytes != Bytes.EMPTY ? hashTxn(txnBytes) : Bytes.EMPTY;
+        recordBuilder.setTransactionHash(toByteString(hash));
+
+        receiptBuilder.setAccountID(transactionID.getAccountID());
```
MiroslavGatsanoga, line 162: I don't think we need to set the `accountId` for every transaction e.g. I don't see it in the records of a `ConsensusSubmitMessage`.
Apart from `CryptoCreate` and hollow account finalization I don't see it used anywhere else in the codebase.

mhess-swl: My mistake, I was thinking this was the payer ID. Thanks for pointing this out!

PR #14673, opened 2024-08-07
feat: create consensus submit message translator

### LINKED ISSUE #14618: Create `ConsensusSubmitMessageTranslator`
Given a submit message transaction, create a translator class that will be responsible for populating `TransactionReceipt` and `TransactionRecord` from the output of the transaction and the state changes block items.

MiroslavGatsanoga:
**Description**:

Translator from block stream items (produced by consensus submit message transactions) to record stream entries.
In order to validate streams produced by transactions from `SubmitMessageSuite` run the following Gradle command:
```
./gradlew :test-clients:test --tests "com.hedera.services.bdd.suites.consensus.SubmitMessageSuite" --tests "com.hedera.services.bdd.suites.validation.StreamValidationTest"
```

**Related issue(s)**:

Fixes #14618 

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ConsensusSubmitMessageTranslator.java
```diff
@@ -0,0 +1,56 @@
...
+                    final var topic = stateChange.mapUpdate().value().topicValue();
+                    receiptBuilder.topicSequenceNumber(topic.sequenceNumber());
+                    receiptBuilder.topicRunningHash(topic.runningHash());
+                });
+
+        return new SingleTransactionRecord(
+                transaction.txn(),
+                TransactionRecord.newBuilder().receipt(receiptBuilder.build()).build(),
+                List.of(),
+                new SingleTransactionRecord.TransactionOutputs(null));
```
mhess-swl, line 54: Since we're not sure how to construct sidecars, maybe we create a dummy object that does nothing, instead of using `List.of()`. Just a thought. Otherwise, this class makes a lot of sense, well done!

petreze: @mhess-swl 
On this topic, can't we use sidecars populated for certain transaction from [BlockStreamBuilder](https://github.com/hashgraph/hedera-services/blob/1033baa9f6feec07fd58d6c95ff3ea24decb3055/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java#L267-L293)? Later, we can access them by `transaction.output().contractCreate().sidecars()` for example

mhess-swl: I'm not very familiar with sidecar records, so maybe they will come from `output()`. Maybe they also come from state changes. Sounds like a good question to ask.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ConsensusSubmitMessageTranslator.java
```diff
@@ -0,0 +1,60 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class ConsensusSubmitMessageTranslator implements TransactionRecordTranslator<SingleTransactionBlockItems> {
+    @Override
+    public SingleTransactionRecord translate(
+            @NotNull SingleTransactionBlockItems transaction, @Nullable StateChanges stateChanges) {
+        final var receiptBuilder = TransactionReceipt.newBuilder();
+
+        // TODO: Where in state is the topicRunningHashVersion stored? Hardcode to 3 for now.
+        receiptBuilder.topicRunningHashVersion(3L);
```
mhess-swl, line 35: I think this will come from the state changes. I'll add an example of what I _think_ needs to go here

mhess-swl: I was wrong, the `Topic` object doesn't have the running hash _version_ at all, so the state changes won't help :( we're going to need to ask @jsync-swirlds why this property was removed (or never present) in a consensus submit's `TransactionOutput`

jsync-swirlds: The `version` of the topic running hashes never changes, so it was removed from the output as unnecessary repetition.
It's version 3 back to some quite early record streams and very unlikely to change, so it's not included in the block stream submit output because that wastes 3-5 bytes for each HCS submit, which is the most frequent transaction we have.

The assumption is that to change the running hash we have to deploy a new consensus node software version, and run a HIP for the change, so if we ever do make that change, we'll add the version as a separate state singleton (so we only transmit the version when it actually changes).
If the mirror node needs that value repeated for every HCS submit, it can be hardcoded when reading block stream input.

We also have an express assumption that block stream _does not_ need to be convertible to a record stream; the two are _not_ compatible.  There _must_ be at *least* enough data in block stream to support mirror node, but it need not be in a form that makes recreating the record stream reasonable, or even possible.

mhess-swl: Ok, makes sense. If it's not already, we should document this in the protobufs. 

The reason we're working on converting block streams to records is for diff testing: compare current outputs to the new block stream outputs and make sure they are as expected

jsync-swirlds: It is documented in the block stream protobuf (one of the many items that are pending a HIP for a major "specification" update).
The purpose here is understood, and reasonable; it just might require a bit more contortions than might be preferred 🤷.

mhess-swl: > It is documented in the block stream protobuf
Awesome, you're on top of it :)

> [...]so even perfect match in diff testing won't guarantee a fully correct block stream (though it will certainly help)
Agreed, and we won't go crazy making them identical. Thanks!

MiroslavGatsanoga: Remove the `TODO` and went ahead with using the constant value for running hash version.

petreze (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ConsensusSubmitMessageTranslator.java
```diff
@@ -0,0 +1,62 @@
...
+import com.hedera.hapi.block.stream.output.StateChanges;
+import com.hedera.hapi.node.transaction.TransactionReceipt;
+import com.hedera.hapi.node.transaction.TransactionRecord;
+import com.hedera.node.app.state.SingleTransactionRecord;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public class ConsensusSubmitMessageTranslator implements TransactionRecordTranslator<SingleTransactionBlockItems> {
+    private static final long RUNNING_HASH_VERSION = 3L;
```
mhess-swl, line 31: It would be nice to leave a comment here pointing to the protobuf, so that anyone reading the code will know why this is a constant

jsync-swirlds: This link will take one directly to the explanation:
https://github.com/hashgraph/hedera-protobufs/blob/pbj-storage-spec-review/block/stream/output/consensus_service.proto#L6

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
```diff
@@ -85,7 +85,7 @@ public void validateBlockVsRecords(@NonNull final List<Block> blocks, @NonNull f
 
         final var actual = translateAll(inputs);
 
-        final var maxDiffs = 10;
+        final var maxDiffs = 1000;
```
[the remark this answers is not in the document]
MiroslavGatsanoga: I increased the number because I am looking at the diffs from the whole `SubmitMessageSuite` and there were a lot of CryptoCreate ones which are not the focus of the task which is to find and fix the `ConsensusSubmitMessage` ones. 

I can revert it to back to 10 but if any reviewer wants to verify if there are any `ConsensusSubmitMessage` diffs left to be fixed they will need to remember to increase this value otherwise it can be misleading and will look like there are no `ConsensusSubmitMessage` diffs to be fixed in the `SubmitMessageSuite` block stream.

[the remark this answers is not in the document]
MiroslavGatsanoga: Reverted back to 10.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/ConsensusSubmitMessageTranslator.java
MiroslavGatsanoga: Added.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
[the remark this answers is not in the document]
mhess-swl: Ahh I understand, didn't think about that. We should come up with a solution for this 🤔

mhess-swl (approved):
LGTM! Ty @MiroslavGatsanoga!

PR #14676, opened 2024-08-07
feat: token airdrop three additional tests

### LINKED ISSUE #14675: (AIRDROP - 74) Attempt to airdrop a token with more than 3 layers of custom fees

povolev15:
HApi tests for token airdrop
1.Given a sender who is frozen for a FT When he signs a tokenAirdropTransaction for the same token 
Then the transaction should fail
2.Given a paused FT 
When a sender with enough FT balance signs a tokenAirdropTransaction for the same token 
Then the transaction should fail
3.Attempt to airdrop a token with more than 3 layers of custom fees

**Related issue(s)**:

Fixes #14675 
#14667
#14666

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
@@ -193,7 +193,7 @@ public class TokenAirdropValidator {
             final TokenID tokenId,
             final AccountAmount senderAmount,
             final ReadableTokenRelationStore tokenRelStore) {
-        //validate association and account frozen
+        // validate association and account frozen
         final var tokenRel = getIfUsable(senderAccount.accountIdOrThrow(), tokenId, tokenRelStore);
 
         if (senderAmount.isApproval()) {
@@ -221,7 +221,7 @@ public class TokenAirdropValidator {
             @NonNull final ReadableTokenRelationStore tokenRelStore,
             @NonNull final ReadableTokenStore tokenStore,
             @NonNull final ReadableNftStore nftStore) {
-        //validate association and account frozen
+        // validate association and account frozen
         getIfUsable(senderAccount.accountIdOrThrow(), tokenId, tokenRelStore);
 
         final var token = tokenStore.get(tokenId);


vtronkov (changes requested):
I think we should add the validations for **ACCOUNT_FROZEN_FOR_TOKEN** and **TOKEN_IS_PAUSED** in `TokenAirdropValidator.validateSemantics` as well.
Currently, the two errors are thrown by the cryptoTrasnfer but the problem is that before we go to the cryptoTransfer we are creating a record for the airdrop and (possibly) a pending state.

I think we should catch this error early. `TokenAirdropValidator.validateSemantics` should be an ok place

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -1323,6 +1326,105 @@ final Stream<DynamicTest> accountThatSupposedToPayHasNoEnoughTokensForCustomFees
                             .signedByPayerAnd(ALICE)
                             .hasKnownStatus(INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE));
         }
+
+        @HapiTest
+        @DisplayName("account that has a token is frozen supposed to fail")
+        final Stream<DynamicTest> accountThatHasTokenIsForzenSupposedToFail() {
```
vtronkov, line 1332: A typo: it should be frozen instead of forzen

JivkoKelchev (changes requested):
TOKEN_IS_PAUSED is covered by `TokenAirdropValidator#validateSemantics`. There is a check in `final var token = getIfUsable(tokenId, tokenStore);`

But we are missing such for ACCOUNT_FROZEN_FOR_TOKEN in TokenAirdropValidator.

You can update `TokenAirdropValidator#validateFungibleTransfers` and `TokenAirdropValidator#validateNftTransfers`
Replacing this:
```
final var tokenRel = tokenRelStore.get(senderAccount.accountIdOrThrow(), tokenId);
validateTrue(tokenRel != null, TOKEN_NOT_ASSOCIATED_TO_ACCOUNT);
```
with this:
`final var tokenRel = getIfUsable(senderAccount.accountIdOrThrow(), tokenId, tokenRelStore);`

By using **getIfUsable** we will check for association between the token and the account, and if the account is frozen too.

vtronkov (approved):

JivkoKelchev (approved):
LGTM, ty @povolev15

PR #14680, opened 2024-08-08
fix: Add test with deleted sender account

### LINKED ISSUE #14578: (AIRDROP - 59) Given sender is a deleted account when it attempts to sign a tokenAirdropTransaction then the transaction should fail.

JivkoKelchev:
**Description**:

(AIRDROP - 59) Given sender is a deleted account when it attempts to sign a tokenAirdropTransaction then the transaction should fail.

**Related issue(s)**:

Fixes #14578

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -121,8 +121,10 @@ public void validateSemantics(
                         .filter(item -> item.amount() < 0)
                         .findFirst();
                 final var senderId = senderAccountAmount.orElseThrow().accountIDOrThrow();
-                final var senderAccount = accountStore.get(senderId);
-                validateTrue(senderAccount != null, INVALID_ACCOUNT_ID);
+
+                final var senderAccount =
```
david-bakin-sl, line 125: This is no longer a `test:` but a `fix:` (PR title).  (Also:  Good thing the test was in the test plan!)

PR #14157, opened 2024-07-04
test: update unit tests to use temporary folders instead of writing files to the main codebase folders

MiroslavGatsanoga:
**Description**:

Fixes for most of the temp test directories from `.gitignore`; the only one left (`**/swirlds-tmp`) will be addressed in separate PR.

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/6714

**Notes for reviewer**:

In order to verify that the previously used test files and directories are not being used anymore run the following:
```
./gradlew -x :test-clients:test test timingSensitive  --rerun-tasks
```

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

Issue #6714, opened 2023-05-22
Change recently added unit tests to use temporary folders instead of writing files to the main codebase folders


jsync-swirlds:
### Problem

Some unit tests are writing files to the following folders.
```
hedera-node/cli-clients/testFile
hedera-node/hedera-mono-service/src/test/resources/balance*/**
hedera-node/hedera-mono-service/temp/**
hedera-node/hedera-mono-service/swirlds-sst-tmp/**
```

### Solution

These tests need to be changed to use temporary filesystem folders (i.e. use the `@TempDir` annotation in the unit test) so that we are not writing test data to the main codebase and then required to ignore those files via `.gitignore`.

### Alternatives

`.gitignore` is updated currently, but this is only acceptable on a temporary basis.

MiroslavGatsanoga:
All of the folders below were fixed by https://github.com/hashgraph/hedera-services/pull/14157:
```
swirlds-platform-core/data/**
hedera-node/cli-clients/testFile
hedera-node/hedera-mono-service/temp/**
hedera-node/hedera-mono-service/swirlds-sst-tmp/**
hedera-node/data/
hedera-node/settingsUsed.txt
hedera-node/hedera-node/data/
```

Only one left is `**/swirlds-tmp` which is used by `platform-sdk` tests.

PR #14695, opened 2024-08-08
test: use temp dir instead of `**/swirlds-tmp`

MiroslavGatsanoga:
**Description**:

This PR builds upon the changes from https://github.com/hashgraph/hedera-services/pull/14157, fixing for the only one left directory (`**/swirlds-tmp`).

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/6714

**Notes for reviewer**:

The changes introduced are mostly in `platform-sdk` tests, I have opened this as separate PR from https://github.com/hashgraph/hedera-services/pull/14157 in order to make reviewing easier (i.e. smaller diff).
Apart from test files there are also two implementation files where changes were introduced, namely `MerkleDb.java` and `LongListDisk.java`, replacing the usage of the deprecated `LegacyTemporaryFileBuilder` with `Files.createTempDirectory()`.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
@@ -185,7 +185,7 @@ public final class MerkleDb {
         return defaultInstancePath.updateAndGet(p -> {
             if (p == null) {
                 try {
-                    p = LegacyTemporaryFileBuilder.buildTemporaryFile("merkledb");
+                    p = Files.createTempDirectory(null).resolve(MERKLEDB_COMPONENT);
                 } catch (IOException z) {
                     throw new UncheckedIOException(z);
                 }
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
@@ -18,7 +18,6 @@ package com.swirlds.merkledb.collections;
 
 import static java.lang.Math.toIntExact;
 
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
 import com.swirlds.merkledb.utilities.MerkleDbFileUtils;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
@@ -27,6 +26,7 @@ import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.Arrays;
@@ -193,7 +193,7 @@ public class LongListDisk extends AbstractLongList<Long> {
     }
 
     static Path createTempFile(final String sourceFileName) throws IOException {
-        return LegacyTemporaryFileBuilder.buildTemporaryDirectory(STORE_POSTFIX).resolve(sourceFileName);
+        return Files.createTempDirectory(null).resolve(sourceFileName);
     }
 
     /** {@inheritDoc} */
--- a/.gitignore
+++ b/.gitignore
@@ -1043,9 +1043,6 @@ platform-sdk/sdk/metricsDoc.tsv
 # Mermaid Files
 **/platform-components.mermaid
 
-# Temporary Test Folders (Poorly Behaved Tests)
-**/swirlds-tmp
-
 # Preconsensus event stream files
 *.pces
 *.pcesD
--- a/hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/SerializationTest.java
+++ b/hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/SerializationTest.java
@@ -25,10 +25,11 @@ import com.hedera.node.config.data.HederaConfig;
 import com.swirlds.common.constructable.ClassConstructorPair;
 import com.swirlds.common.constructable.ConstructableRegistryException;
 import com.swirlds.common.constructable.RuntimeConstructable;
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
+import com.swirlds.common.test.fixtures.TestFileSystemManager;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.config.extensions.sources.SimpleConfigSource;
 import com.swirlds.config.extensions.test.fixtures.TestConfigBuilder;
+import com.swirlds.merkledb.MerkleDb;
 import com.swirlds.metrics.api.Metrics;
 import com.swirlds.platform.state.MerkleStateLifecycles;
 import com.swirlds.platform.state.MerkleStateRoot;
@@ -53,6 +54,7 @@ import com.swirlds.virtualmap.internal.merkle.VirtualRootNode;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Set;
@@ -60,6 +62,7 @@ import java.util.function.Supplier;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.Mock;
@@ -74,11 +77,18 @@ class SerializationTest extends MerkleTestBase {
     @Mock
     private MerkleStateLifecycles lifecycles;
 
+    @TempDir
+    Path tempDir;
+
+    private TestFileSystemManager testFileSystemManager;
+
     @BeforeEach
     void setUp() throws IOException {
         setupConstructableRegistry();
 
-        this.dir = LegacyTemporaryFileBuilder.buildTemporaryDirectory();
+        this.testFileSystemManager = new TestFileSystemManager(tempDir);
+        this.dir = testFileSystemManager.resolveNewTemp(null);
+        Files.createDirectory(this.dir);
         this.config = new TestConfigBuilder()
                 .withSource(new SimpleConfigSource()
                         .withValue(VirtualMapConfig_.FLUSH_INTERVAL, 1 + "")
@@ -218,7 +228,8 @@ class SerializationTest extends MerkleTestBase {
 
         CRYPTO.digestTreeSync(loadedTree);
         // refreshing the dir
-        dir = LegacyTemporaryFileBuilder.buildTemporaryDirectory();
+        dir = testFileSystemManager.resolveNewTemp(null);
+        Files.createDirectory(dir);
         final byte[] serializedBytesWithCache = writeTree(loadedTree, dir);
 
         // let's load it again and see if it works
@@ -242,6 +253,8 @@ class SerializationTest extends MerkleTestBase {
         final var pair = new ClassConstructorPair(MerkleStateRoot.class, constructor);
         registry.registerConstructable(pair);
 
+        // Restore to a fresh MerkleDb instance
+        MerkleDb.setDefaultPath(testFileSystemManager.resolveNewTemp("merkledb"));
         final MerkleStateRoot loadedTree = parseTree(serializedBytes, dir);
         newRegistry.migrate(
                 loadedTree,
--- a/hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/disk/OnDiskTest.java
+++ b/hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/disk/OnDiskTest.java
@@ -27,8 +27,9 @@ import com.hedera.node.app.state.merkle.MerkleSchemaRegistry;
 import com.hedera.node.app.state.merkle.SchemaApplications;
 import com.hedera.node.config.data.HederaConfig;
 import com.swirlds.common.crypto.DigestType;
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
+import com.swirlds.common.test.fixtures.TestFileSystemManager;
 import com.swirlds.config.api.Configuration;
+import com.swirlds.merkledb.MerkleDb;
 import com.swirlds.merkledb.MerkleDbDataSourceBuilder;
 import com.swirlds.merkledb.MerkleDbTableConfig;
 import com.swirlds.platform.test.fixtures.state.MerkleTestBase;
@@ -45,10 +46,12 @@ import com.swirlds.virtualmap.VirtualMap;
 import com.swirlds.virtualmap.internal.merkle.VirtualRootNode;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Set;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 /**
  * A variety of robust tests for the on-disk merkle data structure, especially including
@@ -64,10 +67,17 @@ class OnDiskTest extends MerkleTestBase {
     private StateDefinition<AccountID, Account> def;
     private VirtualMap<OnDiskKey<AccountID>, OnDiskValue<Account>> virtualMap;
 
+    @TempDir
+    private Path tempDir;
+
+    private TestFileSystemManager testFileSystemManager;
+
     @BeforeEach
     void setUp() throws IOException {
         setupConstructableRegistry();
-        final Path storageDir = LegacyTemporaryFileBuilder.buildTemporaryDirectory();
+        testFileSystemManager = new TestFileSystemManager(tempDir);
+        final Path storageDir = testFileSystemManager.resolveNewTemp(null);
+        Files.createDirectory(storageDir);
 
         def = StateDefinition.onDisk(ACCOUNT_STATE_KEY, AccountID.PROTOBUF, Account.PROTOBUF, 100);
 
@@ -158,13 +168,17 @@ class OnDiskTest extends MerkleTestBase {
         virtualMap.copy(); // throw away the copy, we won't use it
         CRYPTO.digestTreeSync(virtualMap);
 
-        final var snapshotDir = LegacyTemporaryFileBuilder.buildTemporaryDirectory("snapshot");
+        final Path snapshotDir = testFileSystemManager.resolveNewTemp("snapshot");
+        Files.createDirectory(snapshotDir);
         final byte[] serializedBytes = writeTree(virtualMap, snapshotDir);
 
         // Before we can read the data back, we need to register the data types
         // I plan to deserialize.
         final var r = new MerkleSchemaRegistry(registry, SERVICE_NAME, DEFAULT_CONFIG, new SchemaApplications());
         r.register(schema);
+
+        // Restore to a fresh MerkleDb instance
+        MerkleDb.setDefaultPath(testFileSystemManager.resolveNewTemp("merkledb"));
 
         // read it back now as our map and validate the data come back fine
         virtualMap = parseTree(serializedBytes, snapshotDir);
--- a/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbBuilderTest.java
+++ b/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbBuilderTest.java
@@ -22,7 +22,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.swirlds.common.crypto.DigestType;
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
+import com.swirlds.common.test.fixtures.TestFileSystemManager;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.config.extensions.test.fixtures.TestConfigBuilder;
 import com.swirlds.merkledb.config.MerkleDbConfig;
@@ -38,14 +38,19 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 class MerkleDbBuilderTest {
+
+    @TempDir
+    private static Path tempDirectory;
 
     private static Path testDirectory;
 
     @BeforeAll
     static void setup() throws IOException {
-        testDirectory = LegacyTemporaryFileBuilder.buildTemporaryFile("MerkleDbBuilderTest");
+        final var testFileSystemManager = new TestFileSystemManager(tempDirectory);
+        testDirectory = testFileSystemManager.resolve(Path.of("MerkleDbBuilderTest"));
     }
 
     @AfterEach
@@ -74,7 +79,7 @@ class MerkleDbBuilderTest {
         final MerkleDbTableConfig<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> tableConfig =
                 createTableConfig();
         final MerkleDbDataSourceBuilder<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> builder =
-                new MerkleDbDataSourceBuilder<>(tableConfig);
+                new MerkleDbDataSourceBuilder<>(testDirectory.resolve("merkledb"), tableConfig);
         VirtualDataSource<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> dataSource = null;
         try {
             dataSource = builder.build("test1", false);
@@ -95,7 +100,8 @@ class MerkleDbBuilderTest {
         final MerkleDbTableConfig<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> tableConfig =
                 createTableConfig();
         final MerkleDbDataSourceBuilder<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> builder =
-                new MerkleDbDataSourceBuilder<>(tableConfig);
+                new MerkleDbDataSourceBuilder<>(testDirectory.resolve("merkledb"), tableConfig);
+        MerkleDb.setDefaultPath(testDirectory.resolve("merkledb"));
         final MerkleDb defaultDatabase = MerkleDb.getDefaultInstance();
         VirtualDataSource<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> dataSource = null;
         try {
@@ -130,9 +136,9 @@ class MerkleDbBuilderTest {
         final MerkleDbTableConfig<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> tableConfig =
                 createTableConfig();
         tableConfig.preferDiskIndices(true).maxNumberOfKeys(1999).hashesRamToDiskThreshold(Integer.MAX_VALUE >> 4);
-        final MerkleDbDataSourceBuilder<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> builder =
-                new MerkleDbDataSourceBuilder<>(tableConfig);
         final Path defaultDbPath = testDirectory.resolve("defaultDatabasePath");
+        final MerkleDbDataSourceBuilder<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> builder =
+                new MerkleDbDataSourceBuilder<>(defaultDbPath, tableConfig);
         MerkleDb.setDefaultPath(defaultDbPath);
         VirtualDataSource<ExampleLongKeyFixedSize, ExampleFixedSizeVirtualValue> dataSource = null;
         try {
--- a/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbDataSourceMetricsTest.java
+++ b/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbDataSourceMetricsTest.java
@@ -28,7 +28,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import com.swirlds.base.units.UnitConstants;
 import com.swirlds.common.constructable.ConstructableRegistry;
 import com.swirlds.common.crypto.DigestType;
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
+import com.swirlds.common.test.fixtures.TestFileSystemManager;
 import com.swirlds.common.test.fixtures.junit.tags.TestComponentTags;
 import com.swirlds.merkledb.test.fixtures.ExampleByteArrayVirtualValue;
 import com.swirlds.merkledb.test.fixtures.TestType;
@@ -47,6 +47,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 class MerkleDbDataSourceMetricsTest {
 
@@ -54,13 +55,17 @@ class MerkleDbDataSourceMetricsTest {
     // default number of longs per chunk
     private static final int COUNT = 1_048_576;
     private static final int HASHES_RAM_THRESHOLD = COUNT / 2;
+
+    @TempDir
+    private static Path tempDirectory;
+
     private static Path testDirectory;
     private MerkleDbDataSource<VirtualKey, ExampleByteArrayVirtualValue> dataSource;
     private Metrics metrics;
 
     @BeforeAll
     static void setup() throws Exception {
-        testDirectory = LegacyTemporaryFileBuilder.buildTemporaryFile("MerkleDbDataSourceMetricsTest");
+        testDirectory = new TestFileSystemManager(tempDirectory).resolve(Path.of("MerkleDbDataSourceMetricsTest"));
         ConstructableRegistry.getInstance().registerConstructables("com.swirlds.merkledb");
     }
 
--- a/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbDataSourceTest.java
+++ b/platform-sdk/swirlds-jasperdb/src/test/java/com/swirlds/merkledb/MerkleDbDataSourceTest.java
@@ -37,7 +37,7 @@ import com.swirlds.base.function.CheckedConsumer;
 import com.swirlds.base.units.UnitConstants;
 import com.swirlds.common.constructable.ConstructableRegistry;
 import com.swirlds.common.crypto.Hash;
-import com.swirlds.common.io.utility.LegacyTemporaryFileBuilder;
+import com.swirlds.common.test.fixtures.TestFileSystemManager;
 import com.swirlds.common.test.fixtures.junit.tags.TestQualifierTags;
 import com.swirlds.merkledb.test.fixtures.ExampleByteArrayVirtualValue;
 import com.swirlds.merkledb.test.fixtures.TestType;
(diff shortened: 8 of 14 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
```diff
@@ -185,7 +185,7 @@ private static Path getDefaultPath() {
         return defaultInstancePath.updateAndGet(p -> {
             if (p == null) {
                 try {
-                    p = LegacyTemporaryFileBuilder.buildTemporaryFile("merkledb");
+                    p = Files.createTempDirectory(null).resolve(MERKLEDB_COMPONENT);
```
jsync-swirlds, line 188: I _think_ this should actually be using `FileSystemManager.resolveNewTemp(...)` so that temporary files are stored in the correct (not `/tmp`) location for merkldDB.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
```diff
@@ -193,7 +193,7 @@ private ByteBuffer initOrGetTransferBuffer() {
     }
 
     static Path createTempFile(final String sourceFileName) throws IOException {
-        return LegacyTemporaryFileBuilder.buildTemporaryDirectory(STORE_POSTFIX).resolve(sourceFileName);
+        return Files.createTempDirectory(null).resolve(sourceFileName);
```
jsync-swirlds, line 196: I _think_ this should actually be using `FileSystemManager.resolveNewTemp(...)` so that temporary files are stored in the correct (not `/tmp`) location for merkldDB.

jsync-swirlds (commented):
Mostly looks good.
A couple questions on the two non-test files changed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
MiroslavGatsanoga: In order to get the correct location is it ok to rely on the `ConfigurationHolder` (I see it's used to get the `MerkleDbConfig`) in order to get properly configured file manager? Something like this perhaps:
```
    final var config = ConfigurationHolder.getInstance().get();
    p = FileSystemManager.create(config).resolveNewTemp(MERKLEDB_COMPONENT);
```

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
MiroslavGatsanoga: Not sure how to properly initialize the file system manager e.g. using some config or inject is a dependency. 
Will need to spend more time on that. 🤔

lpetrovic05 (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDb.java
jsync-swirlds: Not sure, to be honest.  Perhaps @lpetrovic05 will know the correct usage?

lpetrovic05: ConfigurationHolder is deprecated, but it seems that MerkleDb still uses it. So I would say its ok in this situation to use it for the `FileSystemManager`. ConfigurationHolder can be hopefully removed at some later date

mhess-swl (approved):

Neeharika-Sompalli (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
MiroslavGatsanoga: @jsync-swirlds [the last commit](https://github.com/hashgraph/hedera-services/pull/14695/commits/d4fa04a957be6480ba00027e3057b9f9f944d742) I added today is an attempt to resolve this comment (i.e. to use `FileSystemManager` in `LongListDisk`) and it feels hacky to me, perhaps @lpetrovic05 or @imalygin can advise if the approach makes sense.

Otherwise we can either revert to using `Files.createTempDirectory(null)` (if that's acceptable) or some alternative approach can be tried. Pleas let me know what you think.

lpetrovic05 (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
jsync-swirlds: Looks reasonable to me.
One of our MerkleDB experts (e.g. @imalygin or @artemananiev) should probably check that the behavior of the MerkleDB remains correct.

jsync-swirlds (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
```diff
@@ -180,7 +182,7 @@ protected AbstractLongList(final int numLongsPerChunk, final long maxLongs, fina
      * @param path File to read header from
      * @throws IOException If there was a problem reading the file
      */
-    protected AbstractLongList(final Path path, final long reservedBufferLength) throws IOException {
+    protected void init(final Path path, final long reservedBufferLength) throws IOException {
```
imalygin, line 185: What's the motivation for these method? You made all the fields non-final to make it work. There has to be a critically important reason. What is it?

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -239,15 +241,20 @@ public MerkleDbDataSource(
...
+                    new LongListDisk(dbPaths.pathToDiskLocationInternalNodesFile, fileSystemManager);
         } else if (Files.exists(dbPaths.pathToDiskLocationInternalNodesFile) && !forceIndexRebuilding) {
             pathToDiskLocationInternalNodes = new LongListOffHeap(dbPaths.pathToDiskLocationInternalNodesFile);
         } else {
             pathToDiskLocationInternalNodes = new LongListOffHeap();
         }
         // path to disk location index, leaf nodes
         if (tableConfig.isPreferDiskBasedIndices()) {
-            pathToDiskLocationLeafNodes = new LongListDisk(dbPaths.pathToDiskLocationLeafNodesFile);
+            final FileSystemManager fileSystemManager =
```
imalygin, line 255: What's the reason for having multiple instance of `FileSystemManager`? Given the same instance of `Configuration` they will be identical.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
```diff
@@ -102,12 +110,15 @@ public LongListDisk() {
...
-        this(file, DEFAULT_RESERVED_BUFFER_LENGTH);
+    public LongListDisk(final Path file, FileSystemManager fileSystemManager) throws IOException {
+        this(file, DEFAULT_RESERVED_BUFFER_LENGTH, fileSystemManager);
     }
 
-    LongListDisk(final Path file, final long reservedBufferLength) throws IOException {
-        super(file, reservedBufferLength);
+    LongListDisk(final Path file, final long reservedBufferLength, FileSystemManager fileSystemManager)
+            throws IOException {
+        super();
```
imalygin, line 119: this call is redundant

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
```diff
@@ -192,8 +193,8 @@ private ByteBuffer initOrGetTransferBuffer() {
         return buffer;
     }
 
-    static Path createTempFile(final String sourceFileName) throws IOException {
-        return LegacyTemporaryFileBuilder.buildTemporaryDirectory(STORE_POSTFIX).resolve(sourceFileName);
+    static Path createTempFile(final String sourceFileName, FileSystemManager fileSystemManager) throws IOException {
+        return fileSystemManager.resolveNewTemp(sourceFileName);
```
imalygin, line 197: You no longer use `STORE_POSTFIX` here.  Why is that?

imalygin (changes requested):
1. `AbstractLongList` fields must be final
2. Instead of creating `FileSystemManger` instances over and over again, I suggest creating one on `MerkleDbDatasource` level and then pass it as an argument to the underlying data structures .
3. Avoid using `ConfigurationHolder` completely, consider using `PlatformContext` instead. If it's not possible or greatly increases the scope of the PR, at very least limit lookup of `Configuration` object to `VirtualMap` level and then pass it as an argument

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/LongListHeap.java
```diff
@@ -81,7 +81,8 @@ public LongListHeap(final int numLongsPerChunk) {
      * @throws IOException If there was a problem reading the file
      */
     public LongListHeap(final Path file) throws IOException {
-        super(file, 0);
+        super();
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
MiroslavGatsanoga: I missed that during the lifecycle of a `MerkleDbDatasource` instance the configuration instance will be the same hence the `FileSystemManager` will be identical as well.
Given that I agree with the suggestion from below to have one `FileSystemManager` on `MerkleDbDatasource` level and then pass it as an argument to the underlying data structures.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
MiroslavGatsanoga: The idea was to have `tempFile = createTempFile(file.toFile().getName(), fileSystemManager)` for `LongListDisk` in its constructor and update only this concrete type's constructor to require `fileSystemManager` instead of having it in `AbstractLongList` constructor or as it was initially a side effect of calling the `LongListDisk` override of the  abstract `readBodyFromFileChannelOnInit` method.

MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
MiroslavGatsanoga: Added [this commit](https://github.com/hashgraph/hedera-services/pull/14695/commits/41f0017db88dbfa93ea61a9345b83df1fee2342a) to make the fields final again and instead pass the `fileSystemManager` in `AbstractLongList` constructor.

mhess-swl (dismissed):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
```diff
@@ -126,15 +127,23 @@ public abstract class AbstractLongList<C> implements LongList {
      */
     protected final long reservedBufferLength;
 
+    protected final FileSystemManager fileSystemManager;
```
jsync-swirlds, line 130: Would it make more sense to have this value as part of `LongListDisk` instead of here?
It seems like this is only needed in that particular implementation.

imalygin: @MiroslavGatsanoga Joseph is right. `LongListDisk` is the only implementation of `LongList` that works with disk. Please, move `FileSystemManager` there.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/VirtualRootNode.java
```diff
@@ -188,7 +189,10 @@ public static class ClassVersion {
      * Placeholder (since this is such a hotspot) to hold the results from {@link ConfigurationHolder#getConfigData(Class)}
      * rather than calling that method more than once during the lifecycle of a {@link VirtualRootNode} instance.
      */
-    private final VirtualMapConfig config = ConfigurationHolder.getConfigData(VirtualMapConfig.class);
+    private final Configuration configuration =
```
imalygin, line 192: This should be passes as an argument as well.

imalygin (dismissed):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
MiroslavGatsanoga: Moved `FileSystemManager` to `LongListDisk`

jsync-swirlds (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/merkle/VirtualRootNode.java
MiroslavGatsanoga: Fixed in [this commit](https://github.com/hashgraph/hedera-services/pull/14695/commits/5b936699f0e81d172aeb5021b60e1969eb5feac4), aligning changes with https://github.com/hashgraph/hedera-services/pull/15562.

hedera-node/test-clients/build.gradle.kts
```diff
@@ -315,6 +315,8 @@ val cleanYahCli =
         delete(File(project.file("yahcli"), "yahcli.jar"))
     }
 
+dependencies { implementation(project(":swirlds-merkledb")) }
```
mhess-swl, line 318: Is this necessary? I would think adding it to the module-info file would be sufficient

platform-sdk/swirlds-merkledb/src/timingSensitive/java/com/swirlds/merkledb/MerkleDbSnapshotTest.java
```diff
@@ -110,13 +109,21 @@ void setupTest() throws Exception {
...
                 Duration.ofSeconds(1),
                 "Expected no open dbs. Actual number of open dbs: " + MerkleDbDataSource.getCountOfOpenDatabases());
+
+        if (Files.exists(tempDirectory)) {
+            Files.walk(tempDirectory).map(Path::toFile).forEach(file -> {
+                if (!file.delete()) {
+                    file.deleteOnExit();
+                }
+            });
+        }
```
mhess-swl, line 126: Are there files created by this test that don't get deleted?

mhess-swl (dismissed):

rbarkerSL (approved):
LGTM

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -181,13 +186,18 @@ public MerkleDbDataSource(
...
+            final boolean compactionEnabled,
+            final Configuration configuration)
             throws IOException {
         this.database = database;
         this.tableName = tableName;
         this.tableId = tableId;
         this.tableConfig = tableConfig;
 
+        if (tableConfig.isPreferDiskBasedIndices()) {
+            this.fileSystemManager = FileSystemManager.create(configuration);
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Passing the config further and creating the file system manager where needed instead e.g. `LongListDisk`.

platform-sdk/swirlds-merkledb/src/timingSensitive/java/com/swirlds/merkledb/MerkleDbSnapshotTest.java
MiroslavGatsanoga: This was is an attempt get rid of the `CleanupMode.NEVER` similar to the one mentioned in the [above comment](https://github.com/hashgraph/hedera-services/pull/14695#discussion_r1781565596). Still in progress though.

rbarkerSL (dismissed):
lgtm

platform-sdk/swirlds-merkle/src/timingSensitive/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTestBase.java
```diff
@@ -96,10 +98,15 @@ public class VirtualMapReconnectTestBase {
     protected BrokenBuilder teacherBuilder;
     protected BrokenBuilder learnerBuilder;
 
+    @TempDir(cleanup = CleanupMode.NEVER)
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Removed.

hedera-node/test-clients/build.gradle.kts
MiroslavGatsanoga: It seems it's not needed anymore, I will remove it.

platform-sdk/swirlds-merkle/src/timingSensitive/java/com/swirlds/virtual/merkle/reconnect/VirtualMapReconnectTestBase.java
```diff
@@ -96,10 +98,15 @@ public class VirtualMapReconnectTestBase {
...
+    private static Path tempDirectory;
+
+    private static TestFileSystemManager testFileSystemManager;
+
     VirtualDataSourceBuilder createBuilder() throws IOException {
         // The tests create maps with identical names. They would conflict with each other in the default
         // MerkleDb instance, so let's use a new (temp) database location for every run
-        final Path defaultVirtualMapPath = LegacyTemporaryFileBuilder.buildTemporaryFile();
+        final Path defaultVirtualMapPath = testFileSystemManager.resolveNewTemp("merkledb");
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Added.

jsync-swirlds (approved):

rbarkerSL (dismissed):
LGTM

platform-sdk/swirlds-platform-core/src/testFixtures/java/com/swirlds/platform/test/fixtures/config/ConfigUtils.java
```diff
@@ -0,0 +1,35 @@
...
+import com.swirlds.common.io.config.TemporaryFileConfig;
+import com.swirlds.config.api.Configuration;
+import com.swirlds.config.api.ConfigurationBuilder;
+import com.swirlds.merkledb.config.MerkleDbConfig;
+import com.swirlds.virtualmap.config.VirtualMapConfig;
+
+public final class ConfigUtils {
+    private ConfigUtils() {}
+
+    public static final Configuration CONFIGURATION = ConfigurationBuilder.create()
```
lpetrovic05, line 29: There is no javadoc in this class. Also, it seems to provide functionality for a very specific use case, so I think it needs to be moved and renamed accordingly

MiroslavGatsanoga: This is coming from https://github.com/hashgraph/hedera-services/pull/15562, I picked it here in order to minimize the number of conflicts that would arise when the PRs get merged. 
The configuration related changes here were related to addressing [this comment](https://github.com/hashgraph/hedera-services/pull/14695#discussion_r1782018018) and that has increased the scope (i.e. files changed) of this PR significantly imo.
I would suggest this to be addressed in https://github.com/hashgraph/hedera-services/pull/15562 instead. cc @thenswan

thenswan: I am good to address it in #15562 that I opened

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
```diff
@@ -126,6 +128,9 @@ public abstract class AbstractLongList<C> implements LongList {
      */
     protected final long reservedBufferLength;
 
+    /** Platform configuration */
+    protected Configuration configuration;
```
mhess-swl, line 132: I don't see any added lines of code that use this field. Is it needed?

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListHeap.java
```diff
@@ -78,10 +79,11 @@ public LongListHeap(final int numLongsPerChunk) {
      * Create a {@link LongListHeap} from a file that was saved.
      *
      * @param file the file to read from
+     *             * @param configuration the configuration to use
```
mhess-swl, line 82: Formatting

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -215,6 +215,9 @@ public static void main(final String... args) throws Exception {
         final var merkleCryptography = MerkleCryptographyFactory.create(configuration, cryptography);
         MerkleCryptoFactory.set(merkleCryptography);
 
+        // Register with the ConstructableRegistry classes which need configuration.
+        BootstrapUtils.setupConstructableRegistryWithConfiguration(configuration);
```
mhess-swl, line 219: Does this interfere with the constructable registry call in the `Hedera` class? I'm surprised this is necessary

mhess-swl (commented):
@MiroslavGatsanoga this is the PR that keeps growing, as you know 😓 What do you think of limiting the PR changes only to services classes, and then other platform teams can handle changing the platform code? Is that a possibility?

lpetrovic05 (approved):

MiroslavGatsanoga:
> What do you think of limiting the PR changes only to services classes, and then other platform teams can handle changing the platform code? Is that a possibility?

@mhess-swl Any Services related changes were accidental and probably due to the config related changes that are already being addressed in https://github.com/hashgraph/hedera-services/pull/15562.
The current PR is dealing with `**/swirlds-tmp` which is used by platform tests and platform classes such as `LongListDisk`.
So I don't see value in Services specific PR since I don't think `**/swirlds-tmp` is used by Services tests.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
```diff
@@ -77,15 +82,20 @@ public class LongListDisk extends AbstractLongList<Long> {
     /**
      * Create a {@link LongListDisk} with default parameters.
      */
-    public LongListDisk() {
-        this(DEFAULT_NUM_LONGS_PER_CHUNK, DEFAULT_MAX_LONGS_TO_STORE, DEFAULT_RESERVED_BUFFER_LENGTH);
+    public LongListDisk(final @NonNull Configuration configuration) {
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListHeap.java
MiroslavGatsanoga: Fixed.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
MiroslavGatsanoga: Used in `LongListDisk` which inherits from this abstract class and overrides the `onEmptyOrAbsentSourceFile` and `readBodyFromFileChannelOnInit` methods in order to construct `fileSystemManager` for temp files.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
MiroslavGatsanoga: Shouldn't be interfering with the other call since this sets up some platform class instances which previously had parameterless constructors but now need config param e.g. `VirtualMap`, `VirtualRootNode`, `VirtualNodeCache`.
That's another configuration related change that is initially being addressed in https://github.com/hashgraph/hedera-services/pull/15562.

imalygin (dismissed):

imalygin (dismissed):

thenswan (dismissed):

imalygin (dismissed):

Nana-EC (approved):

thenswan (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/BirthRoundStateMigration.java
```diff
@@ -46,11 +47,13 @@ private BirthRoundStateMigration() {}
...
      * @param ancientMode  the current ancient mode
      * @param appVersion   the current application version
+     * @param platformConfiguration the platform configuration
      */
     public static void modifyStateForBirthRoundMigration(
             @NonNull final SignedState initialState,
             @NonNull final AncientMode ancientMode,
-            @NonNull final SoftwareVersion appVersion) {
+            @NonNull final SoftwareVersion appVersion,
+            @NonNull final Configuration platformConfiguration) {
```
poulok, line 56: Why is this being added if it's not even used?

poulok:
Configuration is now being passed to many files even though it is totally unused in those files. Why were these changes made?

rbarkerSL (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/TransactionHandler.java
```diff
@@ -38,9 +40,14 @@ public class TransactionHandler {
...
 
-    public TransactionHandler(final NodeId selfId, final SwirldStateMetrics stats) {
+    /** Platform configuration. */
+    private final Configuration platformConfiguration;
+
+    public TransactionHandler(
+            final NodeId selfId, final SwirldStateMetrics stats, @NonNull final Configuration platformConfiguration) {
         this.selfId = selfId;
         this.stats = stats;
+        this.platformConfiguration = platformConfiguration;
```
poulok, line 50: same comment, this is not needed here

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/BirthRoundStateMigration.java
MiroslavGatsanoga: Removed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/BootstrapUtils.java
```diff
@@ -150,6 +153,19 @@ public static void setupConstructableRegistry() {
         }
     }
 
+    /**
+     * Add classes to the constructable registry which need the configuration.
+     * @param configuration configuration
+     */
+    public static void setupConstructableRegistryWithConfiguration(Configuration configuration)
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Added.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/TransactionHandler.java
MiroslavGatsanoga: Removed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManagerUtils.java
```diff
@@ -39,12 +40,14 @@ private SwirldStateManagerUtils() {}
...
      * @param softwareVersion the current software version
+     * @param platformConfiguration platform configuration
      * @return the newly created state copy
      */
     public static MerkleRoot fastCopy(
             @NonNull final MerkleRoot state,
             @NonNull final SwirldStateMetrics stats,
-            @NonNull final SoftwareVersion softwareVersion) {
+            @NonNull final SoftwareVersion softwareVersion,
+            @NonNull final Configuration platformConfiguration) {
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Removed.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistryTest.java
```diff
@@ -212,9 +218,17 @@ class MigrationTest {
...
+        @TempDir
+        private Path tempDir;
+
         @BeforeEach
         void setUp() {
 
+            final Configuration configuration = new TestConfigBuilder()
+                    .withValue(FileSystemManagerConfig_.TMP_DIR, tempDir.toString())
+                    .getOrCreateConfig();
+            ConfigurationHolder.getInstance().setConfiguration(configuration);
```
thenswan, line 230: Is there a way to avoid using `ConfigurationHolder` here (as I tried to get rid of such usages in #15562)?

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListHeap.java
```diff
@@ -81,8 +80,8 @@ public LongListHeap(final int numLongsPerChunk) {
      * @param file the file to read from
      * @throws IOException If there was a problem reading the file
      */
-    public LongListHeap(final Path file, final Configuration configuration) throws IOException {
```
thenswan, line 84: cc @artemananiev

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
```diff
@@ -70,6 +70,11 @@ public class SwirldStateManager implements FreezePeriodChecker {
      */
     private final SoftwareVersion softwareVersion;
 
+    /**
+     * The platform context.
+     */
+    private final PlatformContext platformContext;
```
thenswan, line 76: Is it still needed here?

MiroslavGatsanoga: Removed.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/state/merkle/MerkleSchemaRegistryTest.java
MiroslavGatsanoga: Yes, removed it.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
```diff
@@ -168,13 +167,13 @@ protected AbstractLongList(final int numLongsPerChunk, final long maxLongs, fina
...
      * @param path File to read header from
+     * @param reservedBufferLength reserved buffer length that the list should have before minimal index in the list
      * @param configuration platform configuration
      * @throws IOException If there was a problem reading the file
      */
     protected AbstractLongList(
             final Path path, final long reservedBufferLength, @NonNull final Configuration configuration)
             throws IOException {
-        requireNonNull(configuration);
         this.configuration = configuration;
```
artemananiev, line 177:
```
this.configuration = requireNonNull(configuration);
```

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListHeap.java
artemananiev: I see two ways to deal with it:

1. Leave `configuration` in AbstractLongList. When a subclass instance is created, provide a config, even if it isn't used in this subclass
2. Remove `configuration` from AbstractLongList. Use it only in subclasses that really need a config

MiroslavGatsanoga: > 2. Remove configuration from AbstractLongList. Use it only in subclasses that really need a config

I need the config only in `LongListDisk` but adding it only there is not an option since its constructor invokes `super()` which itself calls the overridden methods that need the config to be present. So I will go with the first option and revert it to the way it was i.e. for the other subclasses `AbstractLongList` I'll provide the config even though it's not used in them.

@poulok fyi, since it related to https://github.com/hashgraph/hedera-services/pull/14695#issuecomment-2465588945.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
MiroslavGatsanoga: Updated.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/collections/LongListDisk.java
```diff
@@ -83,19 +84,20 @@ public class LongListDisk extends AbstractLongList<Long> {
...
     }
 
     LongListDisk(
             final int numLongsPerChunk,
             final long maxLongs,
             final long reservedBufferLength,
-            final @NonNull Configuration configuration) {
+            @NonNull final Configuration configuration) {
         super(numLongsPerChunk, maxLongs, reservedBufferLength);
+        this.fileSystemManager = FileSystemManager.create(configuration);
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Here we call the `super()` variant that does not have config as parameter and that variant does not call 
either `onEmptyOrAbsentSourceFile` or `readBodyFromFileChannelOnInit` (which initialize `fileSystemManager` in the case of `LongListDisk`)

thenswan (approved):

platform-sdk/swirlds-common/src/testFixtures/java/module-info.java
```diff
@@ -21,6 +21,7 @@
     requires transitive com.hedera.pbj.runtime;
     requires com.swirlds.config.extensions.test.fixtures;
     requires com.swirlds.logging;
+    requires com.swirlds.merkledb;
```
poulok, line 24: I do not think this dependency should be added. Test fixtures is generic testing utilities. Rather than adding this dependency for the `TestFileSystemManager.resetMerkleDb` method, it can be implemented in `MerkleDbTestUtils` or something similar that is specific to merkle db.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/MerkleDb.java
```diff
@@ -176,16 +175,12 @@ public static MerkleDb getInstance(final Path path, final @NonNull Configuration
...
         requireNonNull(configuration);
 
         return defaultInstancePath.updateAndGet(p -> {
             if (p == null) {
-                try {
-                    p = LegacyTemporaryFileBuilder.buildTemporaryFile("merkledb", configuration);
-                } catch (IOException z) {
-                    throw new UncheckedIOException(z);
-                }
+                p = FileSystemManager.create(configuration).resolveNewTemp(MERKLEDB_COMPONENT);
```
poulok, line 183: Have you verified this works for a local network?

platform-sdk/swirlds-common/src/testFixtures/java/module-info.java
MiroslavGatsanoga: Moved the `resetMerkleDb` method to `MerkleDbTestUtils` in [this commit](https://github.com/hashgraph/hedera-services/pull/14695/commits/98e06917eab4d3172446e580023cfb1d1569e8b7). Let me know if you had that in mind or I need to update it further.

MiroslavGatsanoga:
Closing this since the issue will be moved to Platform team.

PR #14697, opened 2024-08-08
fix: still set record `ethereum_hash` on `CONSENSUS_GAS_EXHAUSTED`

### LINKED ISSUE #14610: `EthereumTransaction` record with `CONSENSUS_GAS_EXHAUSTED` should still get hash in record
### Description

The record of an `EthereumTransaction` resolving to `CONSENSUS_GAS_EXHAUSTED` did not get its `ethereum_hash` field set.

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

_No response_

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #14610 
   * In `DispatchProcessor`, if an `EthereumTransaction` is throttled, ensures the record still has the `ethereum_hash` by calling a new `handleThrottled()` method on the `EthereumTransactionHandler`.

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
@@ -161,6 +161,19 @@ public class EthereumTransactionHandler implements TransactionHandler {
         throwIfUnsuccessful(outcome.status());
     }
 
+    /**
+     * Does work needed to externalize details after an Ethereum transaction is throttled.
+     * @param context the handle context
+     */
+    public void handleThrottled(@NonNull final HandleContext context) {
+        final var component = provider.get().create(context, ETHEREUM_TRANSACTION);
+        final var ethTxData =
+                requireNonNull(requireNonNull(component.hydratedEthTxData()).ethTxData());
+        context.savepointStack()
+                .getBaseBuilder(EthereumTransactionStreamBuilder.class)
+                .ethereumHash(Bytes.wrap(ethTxData.getEthereumHash()));
+    }
+
     @NonNull
     @Override
     public Fees calculateFees(@NonNull final FeeContext feeContext) {
--- a/hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/handlers/EthereumTransactionHandlerTest.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/handlers/EthereumTransactionHandlerTest.java
@@ -207,6 +207,18 @@ class EthereumTransactionHandlerTest {
     }
 
     @Test
+    void setsEthHashOnThrottledContext() {
+        given(factory.create(handleContext, ETHEREUM_TRANSACTION)).willReturn(component);
+        given(component.hydratedEthTxData()).willReturn(HydratedEthTxData.successFrom(ETH_DATA_WITH_TO_ADDRESS));
+        given(handleContext.savepointStack()).willReturn(stack);
+        given(stack.getBaseBuilder(EthereumTransactionStreamBuilder.class)).willReturn(recordBuilder);
+        given(recordBuilder.ethereumHash(Bytes.wrap(ETH_DATA_WITH_TO_ADDRESS.getEthereumHash())))
+                .willReturn(recordBuilder);
+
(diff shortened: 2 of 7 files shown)


Neeharika-Sompalli (approved):
Makes sense to me. If we need to externalize details for other operations when throttled we could move this method to the interface. Thanks @tinker-michaelj

lukelee-sl (approved):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
```diff
@@ -161,6 +161,19 @@ public void handle(@NonNull final HandleContext context) throws HandleException
...
     }
 
+    /**
+     * Does work needed to externalize details after an Ethereum transaction is throttled.
+     * @param context the handle context
+     */
+    public void handleThrottled(@NonNull final HandleContext context) {
+        final var component = provider.get().create(context, ETHEREUM_TRANSACTION);
+        final var ethTxData =
+                requireNonNull(requireNonNull(component.hydratedEthTxData()).ethTxData());
```
david-bakin-sl, line 171: (first time I've seen `requireNonNull` nested like this ... it's allowed (obviously) ... but ugly.  but the need to pass nulls along in a chain of method calls is definitely needed - happens all the time.  Java needs the C# `??` "null-coalescing operator".)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
```diff
@@ -406,22 +410,26 @@ final Stream<DynamicTest> serviceFeeRefundedIfConsGasExhausted() {
```
david-bakin-sl, line 410: There are a bunch of tests here, including this one, that make me wonder: Why are they in `FileUpdateSuite`?  This one, `maxRefundIsEnforced`, `allUnusedGasIsRefundedIfSoConfigured` and so on.  Kind of looks like a copy/paste error - they were meant to go somewhere else?

david-bakin-sl (approved):
LGTM - except my question there about `FileUpdateSuite` - if you agree maybe one of us should file a ticket to refactor that?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/handlers/EthereumTransactionHandlerTest.java
```diff
@@ -206,6 +206,18 @@ void delegatesToCreatedComponentAndExposesEthTxDataCallWithToAddress() {
...
+    @Test
+    void setsEthHashOnThrottledContext() {
+        given(factory.create(handleContext, ETHEREUM_TRANSACTION)).willReturn(component);
+        given(component.hydratedEthTxData()).willReturn(HydratedEthTxData.successFrom(ETH_DATA_WITH_TO_ADDRESS));
+        given(handleContext.savepointStack()).willReturn(stack);
+        given(stack.getBaseBuilder(EthereumTransactionStreamBuilder.class)).willReturn(recordBuilder);
+        given(recordBuilder.ethereumHash(Bytes.wrap(ETH_DATA_WITH_TO_ADDRESS.getEthereumHash())))
+                .willReturn(recordBuilder);
+
+        assertDoesNotThrow(() -> subject.handleThrottled(handleContext));
```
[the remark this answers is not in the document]
tinker-michaelj: Since this test uses strict mocks, if the `given()` wasn't used, it would throw `UnnecessaryStubbingException`.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
tinker-michaelj: Yes that would be nice...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/file/FileUpdateSuite.java
tinker-michaelj: Created [this issue](https://github.com/hashgraph/hedera-services/issues/14723) @david-bakin-sl

david-bakin-sl: Ah, good explanation there about collecting leaky tests together willy-nilly - I remember that era.

PR #14782, opened 2024-08-12
fix: update airdrop configs

vtronkov:
**Description**:
* Renamed maxAllowedAirdrops to maxPendingAllowedAirdrops 
* Updated maxPendingAllowedAirdrops from 1M to 20M


hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TokensConfig.java
```diff
@@ -48,8 +48,8 @@ public record TokensConfig(
                 ScaleFactor nftsMintThrottleScaleFactor,
         @ConfigProperty(value = "nfts.useVirtualMerkle", defaultValue = "true") @NetworkProperty
                 boolean nftsUseVirtualMerkle,
-        @ConfigProperty(value = "maxAllowedAirdrops", defaultValue = "1000000") @NetworkProperty
-                long maxAllowedAirdrops,
+        @ConfigProperty(value = "maxPendingAllowedAirdrops", defaultValue = "1000000") @NetworkProperty
```
Neeharika-Sompalli, line 51: This is needed to be 20M

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/schemas/V0530TokenSchema.java
```diff
@@ -32,7 +32,7 @@
 
 public class V0530TokenSchema extends StakingInfoManagementSchema {
     private static final Logger logger = LogManager.getLogger(V0530TokenSchema.class);
-    private static final long MAX_PENDING_AIRDROPS = 1_000_000_000L;
+    private static final long MAX_PENDING_AIRDROPS = 20_000_000_000L;
```
Neeharika-Sompalli, line 35: This is just used in Schema, that defines max keys possible in State. So, should be same as other states in TokenService

Neeharika-Sompalli (approved):

PR #14784, opened 2024-08-12
feat: add more event hashing options

### LINKED ISSUE #14783: Change event hashing options
The current event hashing option `event.migrateEventHashing` is insufficient for all use cases. It needs to be expanded to cover the situation where an upgrade happens and the events before the upgrade are hashed by the new algorithm. It also must not require further config changes once is set. So a version number denoting the first version of the new hashing algorithm is the best solution.

This needs to be merged to release 53 then ported to develop.

lpetrovic05:
closes #14783

regression PR: https://github.com/swirlds/swirlds-platform-regression/pull/4001


lpetrovic05:
wrong approach

PR #14791, opened 2024-08-12
feat: single transaction hashing for pbj events

### LINKED ISSUE #14790: Single transaction hashing for PBJ events
The current implementation of PBJ hashing for events double hashes all transactions. This might be causing performance problems because of memory allocation.

Create a branch of release 53 where events use PBJ hashing, but do not double hash each transaction.

lpetrovic05:
closes #14790


lpetrovic05:
this is not needed

PR #14799, opened 2024-08-12
fix: immediately finalize transfer lists for scheduled crypto transfer

tinker-michaelj:
**Description**:
 - Without the complete HIP-993 implementation, we need a special case to finalize the transfer list in a scheduled `CryptoTransfer` that triggers an auto-creation.
 - Note this special case was eliminated by the completion of HIP-993 changes in `release/0.53`.


david-bakin-sl (commented):
Without reviewing the code - I have no context for it - I suggest that there be a`FUTURE` comment somewhere - perhaps `TriggeredFinalizeContext.hasChildOrPrecedingRecords`? Or maybe somewhere else - that specifically reminds reader that this is a special case that will not be needed and should be removed once HIP-993 is implemented fully.

Neeharika-Sompalli (approved):
Since in 0.52 the only scheduled operations are `CryptoTransfer,ConsensusSubmitMessage,TokenBurn,TokenMint,CryptoApproveAllowance`, the only transaction that can have child records is `CryptoTransfer`.  

Since we have fix in the next releases, just a special case for this scenario makes sense to me. 

LGTM! Thanks @tinker-michaelj

Issue #14285, opened 2024-07-17
Change virtual data source to work with bytes rather than with objects


artemananiev:
Virtual data source today is strictly typed, it knows what kind of objects (keys and values) are stored, and all its APIs are typed. It may be an issue in the future, when a single virtual map / data source is used to store objects of different types. A solution is to change virtual data source APIs to work with bytes or byte arrays, and serialize/deserialize them at a higher level (in VirtualMap or even in States).

PR #14801, opened 2024-08-12
feat: 14285: Change virtual data source to work with bytes rather than with objects

artemananiev:
Fix summary:

* `VirtualDataSource` interface methods are changed to work with bytes rather than records
* `VirtualHashBytes` and `VirtualLeafBytes` classes are introduced as binary counterparts to the existing `VirtualHashRecord` and `VirtualLeafRecord`
* Conversion from records to bytes and back is done in `VirtualRootNode` and `RecordAccessor`
* `VirtualNodeCache` is not changed at this step
* All `KeySerializer` and `ValueSerializer` usages in MerkleDb are removed, except one place for migration (see below)
* Previously, key and value serializers were passed to `MerkleDbTableConfig` objects, which were then used by data source builders to create MerkleDb data sources. Now the serializers are passed to `VirtualMap` constructors and then passed to `VirtualRootNode` as needed

Testing:

* All existing unit tests are updated to work with bytes
* Crypto transfer and other benchmarks

Migration:

* No data migration is needed
* However, existing virtual maps are loaded from state snapshots rather than created from scratch with key and value serializers provided. In the snapshots, the serializers are stored in MerkleDb. As a temporary solution, the serializers are loaded from disk, but not used in MerkleDb code - only to pass them to virtual maps after deserialization. In the next version, methods like `VirtualDataSource.getKeySerializers()` will be removed

Fixes: https://github.com/hashgraph/hedera-services/issues/14285


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -846,6 +853,7 @@ public void snapshot(final Path snapshotDirectory) throws IOException, IllegalSt
         }
     }
 
+    /*
```
imalygin, line 856: This method is commented out. Do you still need it?

platform-sdk/swirlds-virtualmap/src/test/java/com/swirlds/virtualmap/internal/merkle/VirtualInternalNodeTest.java
```diff
@@ -159,6 +163,8 @@ void getValidChildOnDisk() throws IOException {
         final VirtualRootNode<TestKey, TestValue> root = createRoot();
         root.getState().setFirstLeafPath(6);
         root.getState().setLastLeafPath(12);
+        final KeySerializer<TestKey> keySerializer = new TestKeySerializer();
```
imalygin, line 166: `TestKeySerializer` and `TestValueSerializer` are stateless. Can you make them singletons and use these singletons instead of creating instances in multiple tests?

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/serialize/BaseSerializer.java
```diff
@@ -95,6 +100,19 @@ default int getSerializedSize(@NonNull final T data) {
      */
     void serialize(@NonNull final T data, @NonNull final WritableSequentialData out);
 
+    default Bytes toBytes(@NonNull final T data) {
+        try (final ByteArrayOutputStream bout = new ByteArrayOutputStream()) {
+            final WritableSequentialData out = new WritableStreamingData(bout);
+            serialize(data, out);
+            final byte[] bytes = bout.toByteArray();
+            assert bytes.length == getSerializedSize(data);
+            ;
```
imalygin, line 109: nit: empty line

imalygin (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
artemananiev: Nice catch! This functionality is moved to VirtualMap level, no longer needed in MerkleDb. I'll remove the comment.

platform-sdk/swirlds-virtualmap/src/test/java/com/swirlds/virtualmap/internal/merkle/VirtualInternalNodeTest.java
artemananiev: Makes sense. I introduced `TestKeySerializer.INSTANCE` and `TestValueSerializer.INSTANCE` static final fields and used them all over the tests.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/serialize/BaseSerializer.java
artemananiev: Fixed

imalygin (approved):

edward-swirldslabs (dismissed):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileWriter.java
```diff
@@ -88,24 +84,21 @@ public final class DataFileWriter<D> {
...
             final int compactionLevel)
             throws IOException {
-        this.dataItemSerializer = dataItemSerializer;
         this.path = createDataFilePath(filePrefix, dataFileDir, index, creationTime, DataFileCommon.FILE_EXTENSION);
         metadata = new DataFileMetadata(
                 0, // data item count will be updated later in finishWriting()
                 index,
                 creationTime,
-                dataItemSerializer.getCurrentDataVersion(),
+                0, // backwards compatibility: data version
```
anthony-swirldslabs, line 101: Minor: do we have a plan to remove all the "backwards compatibility" reads/writes that we're still performing in this fix in various places?

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/Bucket.java
```diff
@@ -419,6 +403,15 @@ private FindResult findEntry(final int keyHashCode, final K key) {
...
     }
 
+    private boolean keyEquals(final long pos, final int size, final Bytes key) {
+        for (int i = 0; i < size; i++) {
+            if (bucketData.getByte(pos + i) != key.getByte(i)) {
+                return false;
+            }
+        }
+        return true;
+    }
```
anthony-swirldslabs, line 413: Minor: I'm wondering if this implementation that reads data byte by byte is fast enough, or it may be causing a performance issue of some degree.

platform-sdk/swirlds-jasperdb/src/main/resources/com/swirlds/merkledb/files/hash_record.proto
```diff
@@ -8,7 +8,7 @@ option java_multiple_files = true;
 message HashRecord {
 
     // Virtual node path
-    optional uint64 path = 1;
+    optional fixed64 path = 1;
```
anthony-swirldslabs, line 11: It may not be immediately clear whether using a fixed-length number necessarily improves things w/o knowing the distribution of path values. But that's a minor issue per se.

A major issue here is that this changes the format of the data on the wire (aka in the serialized form), while the PR description claims that no data migration is necessary. If the latter is true, is this protobuf model used nowhere then? If so, why do we have it here? If it is used though, and the type of the field changes, then some data migration may be necessary, no?

The same comment applies to the leaf_record.proto just below.

anthony-swirldslabs (commented):
This PR is very, very large.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/Bucket.java
artemananiev: Yes, I was worried about it, too. Filed a ticket to improve `Bytes.equals()` performance: https://github.com/hashgraph/pbj/issues/278

platform-sdk/swirlds-jasperdb/src/main/resources/com/swirlds/merkledb/files/hash_record.proto
artemananiev: The field has always been `fixed64`. This change is to fix the documentation. No data migration is needed.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileWriter.java
artemananiev: I think this field, `DataFileMetadata.serializationVersion`, can be removed altogether

platform-sdk/swirlds-jasperdb/src/main/resources/com/swirlds/merkledb/files/hash_record.proto
anthony-swirldslabs: I see. It's odd to see a file that seemingly contains real code (~ data definition) which isn't actually a compilable code, but instead is a piece of documentation. I'm unsure why we'd maintain it in a .proto file instead of putting it to javadoc or similar. But it's good to know there's no bugs. Thanks.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/DataFileWriter.java
anthony-swirldslabs: My question is also about a couple of similar places in this fix.

anthony-swirldslabs: We can do the cleanup in a separate PR/project if you like. I'd just want to ensure we do have a plan for the cleanup.

imalygin (approved):

anthony-swirldslabs (approved):

edward-swirldslabs (approved):

mxtartaglia-sl (approved):

tinker-michaelj (approved):
LGTM, tyvm @artemananiev !

Issue #14794, opened 2024-08-12
HIP-904 Reject only `RoyaltyFee` with fallback for airdrops


Neeharika-Sompalli:
1. During the Airdrop transaction reject only the tokens that have RoyaltyFee` with fallback. 
2. Also during `Claim` do the same check if the token has been updated with the RoyaltyFee` with fallback while the airdrop is in pending state. If so, fail the claim.

PR #14806, opened 2024-08-13
fix: reject tokens only with royalty with fallback fee

ibankov:
**Description**:

On token airdrop reject only tokens with royalty with fallback fees, for tokens with custom fees.

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/14794

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
@@ -36,6 +36,7 @@ import com.hedera.hapi.node.base.AccountAmount;
 import com.hedera.hapi.node.base.AccountID;
 import com.hedera.hapi.node.base.NftTransfer;
 import com.hedera.hapi.node.base.TokenID;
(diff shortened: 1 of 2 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -132,21 +131,12 @@ public void validateSemantics(
         }
     }
 
-    private boolean tokenHasNoCustomFeesPaidByReceiver(TokenID tokenId, ReadableTokenStore tokenStore) {
+    private boolean tokenHasNoRoyaltyWithFallbackFee(TokenID tokenId, ReadableTokenStore tokenStore) {
```
MrValioBg, line 134: Fungible tokens have no royalty fee, so you can maybe keep the check for the token type, so you can skip the whole loop, directly, if its not an NFT.

ibankov: I agree, added back the token type check

Neeharika-Sompalli (approved):
LGTM !

PR #14831, opened 2024-08-13
feat: include token memo on balance validation error

### LINKED ISSUE #14830: Correlate token names with the creating `HapiSpec` to simplify debugging
### Problem

If there is a `TokenBalanceValidation` failure like [this one](https://scans.gradle.com/s/6fbigokyqs3xy/tests/task/:test-clients:test/details/com.hedera.services.bdd.suites.validation.StreamValidationTest/streamsAreValid()%5B1%5D?top-execution=1), the artifacts available in CI don't support easy debugging.

### Solution

Automatically correlate every token with its originating spec via its `memo` (unless the spec already set the memo explicitly, which should be enough to correlate).

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14830

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenCreate.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenCreate.java
@@ -329,7 +329,16 @@ public class HapiTokenCreate extends HapiTxnOp<HapiTokenCreate> {
                             supplyType.ifPresent(b::setSupplyType);
                             symbol.ifPresent(b::setSymbol);
                             name.ifPresent(b::setName);
-                            entityMemo.ifPresent(s -> b.setMemo(s));
+                            // If the token doesn't have an explicit memo (which
+                            // should help us correlate any related failures to the
+                            // originating spec), then we set the spec name as the
+                            // memo for this purpose
+                            entityMemo.ifPresentOrElse(
+                                    b::setMemo,
+                                    () -> b.setMemo(spec.getName()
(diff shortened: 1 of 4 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenCreate.java
```diff
@@ -329,7 +329,16 @@ protected Consumer<TransactionBody.Builder> opBodyDef(final HapiSpec spec) throw
                             supplyType.ifPresent(b::setSupplyType);
                             symbol.ifPresent(b::setSymbol);
                             name.ifPresent(b::setName);
-                            entityMemo.ifPresent(s -> b.setMemo(s));
+                            // If the token doesn't have an explicit memo (which
+                            // should help us correlate any related failures to the
+                            // originating spec), then we set the spec name as the
+                            // memo for this purpose
+                            entityMemo.ifPresentOrElse(
```
david-bakin-sl, line 336: Great feature, I bet this'll help

david-bakin-sl (approved):
I like!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/token/HapiTokenCreate.java
mhess-swl: It may be helpful to distinguish auto-set memos with a prefix of some kind. Just a thought

mhess-swl (approved):

Issue #14825, opened 2024-08-13
Design proposal: MerkleDb to store binary data rather than strictly typed objects


artemananiev:
This ticket is to have a design proposal for https://github.com/hashgraph/hedera-services/issues/14285. It's a part of `Virtual Mega Map` epic.

PR #14832, opened 2024-08-13
docs: 14825: Design proposal: MerkleDb to store binary data rather than strictly typed objects

artemananiev:
Fix summary: see proposal.md for details.

Fixes: https://github.com/hashgraph/hedera-services/issues/14825


platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
```diff
@@ -0,0 +1,162 @@
...
+
+`Virtual Mega Map` is an effort to refactor Hedera ledger state to be a single virtual map instead of
+the current N-ary merkle tree. This proposal is focused on the data source (`MerkleDb` data source)
+part of it. The data source is not exposed to the application directly, it's used internally by
+`VirtualMap`, which is in turn accessible from Hedera services code via the States APIs.
+
+## Purpose and Context
+
+Historically, virtual maps and MerkleDb databases were designed to store entities of a particular
+types. In Java code, `VirtualMap` class is defined like this:
```
anthony-swirldslabs, line 15:
```suggestion
Historically, virtual maps and MerkleDb databases were designed to store entities of  particular
types. In Java code, `VirtualMap` class is defined like this:
```

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
```diff
@@ -0,0 +1,162 @@
...
+
+No data migration is expected at this point. Data on disk is stored in the same format (100% byte to
+byte compatibility) as previously.
+
+When a new virtual map is created, it gets key and value serializers from the app. However, there
+is a number of already created maps in the old states. They are not created from scratch, but loaded
+from state snapshots. Where do they get the serializers from? MerkleDb. Despite MerkleDb no longer
+uses serializers, it still loads them from disk (MerkleDb configuration), if available. When a
+virtual map is loaded from a state snapshot, and snapshot is old (before MerkleDb is switched to
+work with bytes), the map will query the data source for key and value serializers to use.
```
anthony-swirldslabs, line 118:
```suggestion
When a new virtual map is created, it gets key and value serializers from the app. However, there
is a number of already created maps in the old states. They are not created from scratch, but loaded
from state snapshots. Such virtual maps rely on the MerkleDb to provide the serializers. Despite MerkleDb no longer
uses serializers, it can still load them from disk (from the MerkleDb configuration), if available. When a
virtual map is loaded from a state snapshot, and the snapshot is old (before MerkleDb is switched to
work with bytes), the map will query the data source for key and value serializers to use.
```

Also, might want to expand on the `if available` piece - it's not clear when this condition is true. Is it typical or rare? Why/when would the serializers be missing from the MerkleDb configuration in the current state?

Also, it's not clear whether the serializers read from the MerkleDb configuration would override the ones specified in the data source or the latter take the precedence. If the latter is true, and I presume a code change would update all the data sources to specify the serializers when this change is rolled out, then it's unclear why we even care about the serializers stored in the MerkleDb configuration at all. It would be nice to clarify these points.

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
```diff
@@ -0,0 +1,162 @@
...
+
+Performance impact has to be measured carefully. There are no flow changes in how data is stored
+to disk or read back. However, there are changes to when and how virtual keys and values are
+converted to bytes and vice versa.
+
+Previously, during flush, an entity (virtual key and value) was stored in the node cache as a
+Java object, passed to MerkleDb, and written to disk using key and value serializers. The
+serializers wrote key and value bytes directly to the file. With the new approach, entities
+are serialized to bytes before flush. It results in extra Java byte array allocations. Then
+these byte arrays are written to the files.
```
anthony-swirldslabs, line 158: The extra heap usage seems a bit concerning. I'm wondering if there's a solution that would still stream the bytes directly from the serializer to the disk, perhaps with a tiny proxy in between, but w/o having to allocate byte arrays for every entity. The latter seems a bit wasteful.

I'm curious to see heap usage comparison w/ and w/o the change, esp. under stress testing and/or longevity testing. I think these numbers belong to this document to better understand the performance impact.

anthony-swirldslabs (commented):
I appreciate your publishing this document. It's important to clearly understand the big picture at the high level before reviewing huge refactoring PRs that are impossible to be split further due to the nature of the refactoring.

imalygin:
Looks good to me

imalygin (dismissed):

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: Fixed

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: "If available" here means "when the snapshot was taken before this feature". In the current `develop`, configuration of any MerkleDb table contains key and value serializers. They are stored in MerkleDb metadata file, `database_metadata.pbj` After this feature is merged, the serializers will not be there.

So the code will be like this:
* When a data source is loaded from a snapshot, it reads its table config from MerkleDb metadata. If key / value serializers are there, they are loaded. If not - no error
* When a virtual map is loaded from a snapshot, it checks its serialization version. Based on the version, it will either deserialize the serializers or not
* If the version is "old", and no serializers are available, the virtual map will get them from the data source, it's guaranteed to have them because of the version (if for some reason there are no serializers, an exception is thrown)
* When a virtual map is created from scratch, it will have the serializers set explicitly, it won't try to load them from the data source

So the short answer is: "if available" means an old version, before this feature is implemented. When a snapshot is taken with a new version, the serializers will not be in the data source, nor VirtualMap will query the data source for them.

I am not sure how to describe it in clear words in the proposal...

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: At the design proposal level, all that can be noted is that there will be more Java allocations. Exact numbers will be only available, after the feature is fully implemented and tested.

artemananiev: The only number I can share so far is from my local branch, where this feature is implemented (not the final version). In crypto transfer benchmark, total Java allocations are about 40% higher than in `develop`. Note that the benchmark may or may not reflect the real Hedera services node object allocation patterns.

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
anthony-swirldslabs: Thanks for applying the change, it looks great.

And thanks for providing the extra details about the "if available" part. I think you could put this comment almost as is in the design document to clarify the plan to deal with the serializers.

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
anthony-swirldslabs: Got it. Thanks! The 40% extra Java allocations still looks somewhat concerning. However, I understand that we won't have actual and complete data until we have the implementation ready. In that case, I suggest to evaluate the heap usage data from various performance tests **before** merging the final implementation into `develop`, and only do so if the performance testing results look acceptable.

IMO, some 10%, or maybe up to 20% of extra heap usage may be acceptable. Anything higher than that would need to be thoroughly studied to ensure the new behavior doesn't negatively impact normal node operations (transactions handling and such) in stress and longevity tests. I suggest to add this remark to this document so that we remember to perform this evaluation while we're implementing and testing this feature, and have a rough idea about what we find acceptable.

OlegMazurov: Let's not mix heap allocations with heap usage together. 40% extra allocations may be OK if they are for short lived objects (as seems to be the case here) and the rate is not too high, i.e. if all is within the GC capacity. 40% extra heap usage (objects surviving multiple garbage collections) would be unacceptable.

OlegMazurov:
I suggest the use of key hash codes in `MerkleDb` (`HalfDiskHashMap`) and their compatibility be explicitly mentioned in this proposal.

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
anthony-swirldslabs: Yes, I understand the difference. Thanks for adding the clarification.

However, my point still stands: the document should mention a need for performing a comprehensive performance analysis using stress and longevity tests, specifically in the area of heap usage, to ensure the normal node operations are not impacted negatively before the code change makes it into production.

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: OK, let me add it to the doc.

artemananiev:
> I suggest the use of key hash codes in `MerkleDb` (`HalfDiskHashMap`) and their compatibility be explicitly mentioned in this proposal.

I'm adding a new paragraph in the "Compatibility" section about it. It says that all data source methods that work with virtual keys accept both key bytes and key hash codes (the methods are listed)

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: Yes, this is what `Performance` section is for. It explicitly lists two areas where Java heap usage and/or allocations may be different from the current implementation. Exact numbers can be provided only after implementation is complete, and a number of real world tests / benchmarks are run. If the numbers were available right away, they would be in the proposal

OlegMazurov (approved):

imalygin (approved):

rbair23 (approved):

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
```diff
@@ -0,0 +1,194 @@
...
+hash codes are used, it would result in data loss.
+
+To fix it, all data source calls that deal with virtual keys accept key hash codes in addition to
+key bytes:
+
+```java
+   VirtualLeafBytes loadLeafRecord(final Bytes keyBytes, final int keyHashCode) throws IOException;
+
+    long findKey(final Bytes keyBytes, final int keyHashCode) throws IOException;
+```
```
anthony-swirldslabs, line 150: Minor: I'm not entirely sure if this is a great idea in general because `Object.hashCode()` isn't generally stable. The values may change between JDK implementations, or even between JVM runs. If a different value returned by the method may result in a data loss, then this is a serious issue per se.

And on top of this, it requires us to pass this extra argument here which doesn't look super pretty.

However, I also understand that changing this behavior is probably outside of scope of this design proposal...

anthony-swirldslabs (approved):

platform-sdk/docs/proposals/virtualmegamap-binary-merkledb/proposal.md
artemananiev: You're 100% right, and we have already had issues with `hashCode()` in the past. In this proposal, however, I just keep whatever is used today to avoid data migration.

One hope was we would be able to hash key bytes rather than use key' `hashCode()`. Once the current proposal is implemented, and we switch to bytes at MerkleDb level, I'll file a ticket about hash code. The chances are we will be able to rehash the whole tree, when the existing merkle tree is migrated to a virtual mega map.

anthony-swirldslabs: Sounds good. Let's hope we get there before we actually lose some data.

jasperpotts:
I am nervous about the performance impact of this it seems like it will cause a bunch for byte[] and Bytes objects to bd GCed. Also don't love the read cache is at bytes level. Because all objects stored are PBJ types and immutable it means we will be doing lots of deserialization and object creation than needed. Some things to keep an eye on as we go forward.

PR #14833, opened 2024-08-13
refactor: Convert `RcDiff` into usable class with cmd wrapper

mhess-swl:
**Description**:
This PR moves the `RcDiff` class from its own module to the `test-clients.main` module, leaving only a new CLI command wrapper class in the `rcdiff` module. This is done so that we can use the `RcDiff` functionality in other test contexts.

Part of #14501. We will use `RcDiff` to diff block stream blocks and record files.


MiroslavGatsanoga (approved):
LGTM

tinker-michaelj (approved):
All makes sense to me, tyvm @mhess-swl !

rbarkerSL (approved):

PR #14837, opened 2024-08-14
feat: simplify `@HapiTest` throttle overrides

### LINKED ISSUE #14836: Simplify testing throttle overrides
### Problem

Currently our tests that override throttle definitions via file `0.0.123` have to explicitly restore the previous throttles, which is awkward.

### Solution

Add a `String throttles` attribute to the `@Leaky*HapiTest` annotations which, if set, is interpreted as a resource to load override throttles from. 

Add a `boolean restoreOriginalThrottles` to `HapiSpec` and snapshot/restore the contents of file `0.0.123` before/after the spec if set.

Refactor tests that override throttles to use this approach instead.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14836


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpec.java
```diff
@@ -688,8 +699,11 @@ private void exec(@NonNull List<SpecOperation> ops) {
         final var shouldPreserveProps = !propertiesToPreserve.isEmpty();
         final Map<String, String> preservedProperties = shouldPreserveProps ? new HashMap<>() : Collections.emptyMap();
         if (shouldPreserveProps) {
-            ops = new ArrayList<>(ops);
-            ops.add(0, remembering(preservedProperties, propertiesToPreserve));
+            ops.addFirst(remembering(preservedProperties, propertiesToPreserve));
+        }
+        if (throttleResource != null) {
+            ops.addFirst(new SysFileOverrideOp(
+                    THROTTLES, () -> throttleResource.isBlank() ? null : resourceAsString(throttleResource)));
```
david-bakin-sl, line 706: can you move the predicate `throttleResource.isBlank()` up to the enclosing `if` at L704?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/queries/crypto/HapiGetAccountBalance.java
```diff
@@ -87,6 +87,7 @@ public class HapiGetAccountBalance extends HapiQueryOp<HapiGetAccountBalance> {
     private boolean assertAccountIDIsNotAlias = false;
     private ByteString rawAlias;
 
+    private boolean includeTokenMemoOnError = false;
```
david-bakin-sl, line 90: I thought the "include token memo on error" was a different PR? (mentioned only here and not in all the files that include bits-and-pieces of the feature)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/TxnUtils.java
```diff
@@ -724,4 +726,22 @@ public static void doIfNotInterrupted(@NonNull final InterruptibleRunnable runna
...
+    /**
+     * Returns the contents of the resource at the given location as a string.
+     *
+     * @param loc the location of the resource
+     * @return the contents of the resource as a string
+     */
+    public static String resourceAsString(@NonNull final String loc) {
+        try {
+            try (final var in = TxnUtils.class.getClassLoader().getResourceAsStream(loc);
+                    final var bridge = new InputStreamReader(requireNonNull(in));
```
david-bakin-sl, line 739: is this the way spotless left the indentation?  odd.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/SysFileOverrideOp.java
```diff
@@ -0,0 +1,143 @@
...
+        public abstract long number();
+    }
+
+    private final Target target;
+    private final Supplier<String> overrideSupplier;
+
+    private boolean autoRestoring = true;
+    private byte[] originalContents;
+
+    /**
```
david-bakin-sl, line 77: Just out of curiosity, when would you use this?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/SysFileOverrideOp.java
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * A utility operation to override the contents of a dynamic system file, remembering the initial contents for
+ * later restoration.
+ */
+public class SysFileOverrideOp extends UtilOp {
+    private static final Logger log = LogManager.getLogger(SysFileOverrideOp.class);
+
+    public enum Target {
+        FEES {
+            @Override
```
david-bakin-sl, line 44: style question: why do you prefer this to having each enum have a `long` field and then a single getter method?  seems like it would be shorter, yet have at least the same readability.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/SysFileOverrideOp.java
```diff
@@ -0,0 +1,143 @@
...
+        this.target = requireNonNull(target);
+        this.overrideSupplier = requireNonNull(overrideSupplier);
+    }
+
+    @Override
+    protected boolean submitOp(@NonNull final HapiSpec spec) throws Throwable {
+        allRunFor(spec, getFileContents("0.0." + target.number()).consumedBy(bytes -> this.originalContents = bytes));
+        log.info("Took snapshot of {}", target);
+        final var styledContents = overrideSupplier.get();
+        if (styledContents != null) {
```
david-bakin-sl, line 101: Would a log be desirable for either `styledContents == null` (or maybe `isBlank` too)?

(Is there a use case for the supplied new contents to be empty or blank?)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/fees/CongestionPricingTest.java
```diff
@@ -65,10 +62,8 @@ public class CongestionPricingSuite {
             requirement = {PROPERTY_OVERRIDES, THROTTLE_OVERRIDES},
             overrides = {"contracts.maxGasPerSec", "fees.percentCongestionMultipliers", "fees.minCongestionPeriod"})
     final Stream<DynamicTest> canUpdateMultipliersDynamically() {
-        var artificialLimits = protoDefsFromResource("testSystemFiles/artificial-limits-congestion.json");
-        var defaultThrottles = protoDefsFromResource("testSystemFiles/throttles-dev.json");
```
david-bakin-sl, line 69: was this file (probably no longer used?) supposed to be `throttles-def{ault}` not `throttles-dev`?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/fees/CongestionPricingTest.java
```diff
@@ -169,7 +154,7 @@ final Stream<DynamicTest> canUpdateMultipliersDynamically2() {
                         "1,7x",
                         "fees.minCongestionPeriod",
                         tmpMinCongestionPeriod),
-                fileUpdate(THROTTLE_DEFS).payingWith(EXCHANGE_RATE_CONTROL).contents(artificialLimits.toByteArray()),
+                overridingThrottles("testSystemFiles/artificial-limits-congestion.json"),
```
david-bakin-sl, line 157: yup, indeed this test looks better now, and also you can't really make mistakes in the override/revert

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/SteadyStateThrottlingTest.java
```diff
@@ -146,13 +141,7 @@ final Stream<DynamicTest> checkBalanceQps() {
     @HapiTest
     @Order(7)
     final Stream<DynamicTest> restoreDevLimits() {
```
david-bakin-sl, line 143: (Hmm, maybe they are "developer" limits not "default" limits after all)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/SteadyStateThrottlingTest.java
```diff
@@ -348,37 +337,4 @@ public Optional<HapiSpecOperation> get() {
             }
         };
     }
-
-    @SuppressWarnings("java:S1144")
```
david-bakin-sl, line 352: This was just dead code?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/UmbrellaRedux.java
```diff
@@ -55,18 +50,15 @@ public class UmbrellaRedux {
     @HapiTest
     @Tag(NOT_REPEATABLE)
     final Stream<DynamicTest> umbrellaRedux() {
```
david-bakin-sl, line 52: I have no idea what an `umbrellaRedux` test is about so I have no idea why it _used to_ update the throttles with the default (now we see that `-dev` **does mean** "default" after all) and now it doesn't have to.  Maybe some javadoc somewhere would help explain what the heck this test is for?

david-bakin-sl (commented):
To the extent I understand it, looks good.

But I'm not sure why the "include token memo on error" changes got into this PR when I thought it looked quite nice in its separate PR.  So just a comment.  Will come back later to give checkmark once I learn that.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/HapiSpec.java
tinker-michaelj: The goal is to only add a `SysFileOverrideOp` if the `HapiSpec#throttleResource` field is non-null.

When we **do** have a non-null `throttleResource`, if it is not blank, then we will automatically override file `0.0.123` with the parsed contents of the given resource. If it is blank, we will just take a snapshot of its initial contents and restore them at the end of the spec.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/queries/crypto/HapiGetAccountBalance.java
tinker-michaelj: Merged `develop`, this diff will be gone now.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/TxnUtils.java
tinker-michaelj: Spotless has a mind of its own. 🤷

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/UmbrellaRedux.java
tinker-michaelj: There are a few of these `RegressionProviderFactory`-based tests...they have some uses, but are quite clumsy (and yes, undocumented!)...created an [issue](https://github.com/hashgraph/hedera-services/issues/14850) to evaluate the ROI of this testing approach.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/SteadyStateThrottlingTest.java
tinker-michaelj: Yes, the `@HapiTest` target networks have higher dev throttle limits.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/SysFileOverrideOp.java
tinker-michaelj: I think copilot just generated this 😂 , clearly the `long` field is more readable; fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/SysFileOverrideOp.java
tinker-michaelj: It's used [here](https://github.com/hashgraph/hedera-services/blob/14836-simple-throttle-overrides/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/SteadyStateThrottlingTest.java#L103) to use the snapshot/restore functionality in this op; but to run several `@HapiTest` between snapshotting and restoring.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/SteadyStateThrottlingTest.java
tinker-michaelj: True.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/fees/CongestionPricingTest.java
tinker-michaelj: It really shouldn't have ever existed; we should always have been taking a snapshot of file `0.0.123` contents and then restoring those like in this PR.

Deleted now!

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

PR #14842, opened 2024-08-14
fix: Precision loss for gas calculation of HTS system contracts

### LINKED ISSUE #10828: Gas calculation for precompiles incorrectly varies with HAPI exchange rate
### Description

Gas calculation appears to vary with hbar exchange rate, compromising Hedera's policy of fixed fee gas costs.

### Steps to reproduce

Run the following unit test in the build:

private static final long GAS_COST_IN_TINYCENT = 852;

    @Test
    void exchangeRateChangesCausesUnstableGasCost() {
        final var mintFeeTinycents = 10000123L;
        final var fixedHbarEquiv = 30000;
        final var pre0900CentEquiv = 269100;
        final var post0900CentEquiv = 269232;
        System.out.println("Pre-09:00 gas cost: " + gasCostGiven(mintFeeTinycents, fixedHbarEquiv, pre0900CentEquiv));
        System.out.println("Post-09:00 gas cost: " + gasCostGiven(mintFeeTinycents, fixedHbarEquiv, post0900CentEquiv));
    }

    private long gasCostGiven(final long tinycentFee, final int hbarEquiv, final int centEquiv) {
        final var rate = ExchangeRate.newBuilder().setHbarEquiv(hbarEquiv).setCentEquiv(centEquiv).build();
        final var gasPriceInTinybar = getTinybarsFromTinyCents(rate, GAS_COST_IN_TINYCENT);
        final var feeInTinbyar = getTinybarsFromTinyCents(rate, tinycentFee);
        final var baseGasCost = (feeInTinbyar + gasPriceInTinybar - 1L) / gasPriceInTinybar;
        return baseGasCost + (baseGasCost / 5L);
    }

with the following exchange rates:

```
At 08:59 CST
curl -X 'GET' \
  'https://mainnet-public.mirrornode.hedera.com/api/v1/network/exchangerate?timestamp=1703343540' \
  -H 'accept: application/json'
and
```

```
At 09:01 CST
curl -X 'GET' \
  'https://mainnet-public.mirrornode.hedera.com/api/v1/network/exchangerate?timestamp=1703343660' \
  -H 'accept: application/json'
```

### Additional context

Proof on mainnet for same function call at two different times (with two different exchange rates): fail https://hashscan.io/mainnet/transaction/1703343586.545013838 success https://hashscan.io/mainnet/transaction/1703344296.632845003
fail gas: 14,232 success gas: 14,226

**Impact**

dApps cannot be certain the gas consumed by a smart contract function call when the gas changes with HAPI exchange rate. This compromises Hedera's stated policy that fees are denominated in USD.

### Hedera network

mainnet

### Version

v0.45.0

### Operating system

None

MrValioBg:
**Description**:
This PR Fixes precision loss which results in different gas, depending on the current exchange rate.

**Related issue(s)**:

Fixes #10828 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.contract.impl.exec.gas;
 
+import static com.hedera.node.app.service.contract.impl.utils.ConversionUtils.FEE_SCHEDULE_UNITS_PER_TINYCENT;
 import static java.util.Objects.requireNonNull;
 
 import com.hedera.hapi.node.base.AccountID;
@@ -166,7 +167,7 @@ public class SystemContractGasCalculator {
     private long asGasRequirement(final long tinybarPrice, final long gasPrice) {
         // We round up to the nearest gas unit, and then add 20% to account for the premium
         // of doing a HAPI operation from inside the EVM
-        final var gasRequirement = (tinybarPrice + gasPrice - 1) / gasPrice;
+        final var gasRequirement = (tinybarPrice + gasPrice - 1) * FEE_SCHEDULE_UNITS_PER_TINYCENT / gasPrice;
         return gasRequirement + (gasRequirement / 5);
     }
 }
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
@@ -111,7 +111,6 @@ public class TinybarValues {
         }
         return asTinybars(
                 childTransactionResourcePrices.basePrices().servicedataOrThrow().gas()
-                        / FEE_SCHEDULE_UNITS_PER_TINYCENT
                         * childTransactionResourcePrices.congestionMultiplier());
     }
 


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
+     * For all view operations, the gas requirement is determined using the canonical gas value
+     * for the TOKEN_INFO dispatch type, as specified in the canonical-prices.json.
+     * The TOKEN_INFO operation is representative of view operations.
      *
-     * @return the gas price for the top-level HAPI operation
+     * @return the gas requirement for a view operation
      */
-    public long topLevelGasPrice() {
-        return tinybarValues.topLevelTinybarGasPrice();
+    public long viewGasRequirement() {
```
david-bakin-sl, line 108: (Now that we have `viewGasRequirement` it seems like `canonicalGasRequirement` should have been called something different that is directly related to and opposite from "view".  But it has a bunch of callers already; not worth changing; just something to remember. Maybe the Javadoc down there at `canonicalGasRequirement` could mention explicitly that it's for the operations that (can) change state?)

david-bakin-sl (commented):
First comments since you've updated it just now.  WIll refresh and continue.

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -83,7 +83,11 @@ public record ContractsConfig(
                 boolean systemContractAccountServiceIsAuthorizedRawEnabled,
         @ConfigProperty(value = "systemContract.updateCustomFees.enabled", defaultValue = "false") @NetworkProperty
                 boolean systemContractUpdateCustomFeesEnabled,
-        @ConfigProperty(value = "evm.version.dynamic", defaultValue = "false") @NetworkProperty
+        @ConfigProperty(value = "systemContract.precisionLossFixForGas.enabled", defaultValue = "true") @NetworkProperty
+                boolean isGasPrecisionLossFixEnabled,
+        @ConfigProperty(value = "systemContract.canonicalViewGas.enabled", defaultValue = "true") @NetworkProperty
+                boolean isCanonicalViewGasEnabled,
+        @ConfigProperty(value = "evm.version.dynamic", defaultValue = "true") @NetworkProperty
```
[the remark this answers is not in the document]
MrValioBg: Oh yes! Great catch. Thank you

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -28,6 +29,7 @@
  */
 public class SystemContractGasCalculator {
     private static final long FIXED_VIEW_GAS_COST = 100L;
+    private static final long FIXED_TINY_CENT_GAS_PRICE_COST = 852_000L;
```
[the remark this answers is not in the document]
MrValioBg: Yeah, I totally agree. I actually had a comment explaining that; it seems like it got lost somehow.
Done!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
+            return gasRequirementOldWithPrecisionLoss(body, payer, minimumPriceInTinyCents);
+        }
+
+        final var nominalPriceInTinyBars = feeCalculator.applyAsLong(body, payer);
+        // For the rare cases where nominalPriceInTinyBars > minimumPriceInTinyCents:
+        // Precision loss may occur as we convert between tinyBars and tinyCents, but it is typically negligible.
+        // The minimal nominal price is > 1e6 tinyCents, ensuring enough precision.
+        // In most cases, the gas difference is zero.
+        // In scenarios where we compare significant price fluctuations (200x, 100x), the gas difference should still be
+        // unlikely to exceed 0 units.
```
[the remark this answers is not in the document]
MrValioBg: It was referring to gas units, however I agree its not very clear. Will change it.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
 
-    /**
-     * Given a transaction body whose dispatch gas requirement must be at least equivalent to a given minimum
-     * amount of tinybars, returns the gas requirement for the transaction to be dispatched.
-     *
-     * @param body the transaction body to be dispatched
-     * @param payer the payer of the transaction
-     * @param minimumPriceInTinybars the minimum equivalent tinybar cost of the dispatch
-     * @return the gas requirement for the transaction to be dispatched
-     */
```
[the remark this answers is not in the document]
MrValioBg: Of course we can :) Will add it.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
```diff
@@ -98,6 +111,30 @@ public long topLevelTinybarGasPrice() {
...
+        return asTinybars(
+                topLevelResourcePrices.basePrices().servicedataOrThrow().gas()
+                        * topLevelResourcePrices.congestionMultiplier());
+    }
+
+    /**
+     * Returns the topLevel gas price cost in tinyCents, without denomination, but with congestion multiplier.
+     * @return the tinyCents gas price
+     */
+    public long topLevelTinyCentsGasPrice() {
```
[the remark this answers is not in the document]
MrValioBg: Done :)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCalculator.java
```diff
@@ -79,8 +79,8 @@ public long logOperationGasCost(
         final var hevmGasCost = gasCostOfStoring(
                 logSize(numTopics, dataLength),
                 lifetime,
-                tinybarValues.topLevelTinybarRbhPrice(),
-                tinybarValues.topLevelTinybarGasPrice());
+                tinybarValues.topLevelTinyCentRbhPrice(),
```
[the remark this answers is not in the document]
MrValioBg: Agreed, changed the javadoc.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
+
+        final var nominalPriceInTinyBars = feeCalculator.applyAsLong(body, payer);
+        // For the rare cases where nominalPriceInTinyBars > minimumPriceInTinyCents:
+        // Precision loss may occur as we convert between tinyBars and tinyCents, but it is typically negligible.
+        // The minimal nominal price is > 1e6 tinyCents, ensuring enough precision.
+        // In most cases, the gas difference is zero.
+        // In scenarios where we compare significant price fluctuations (200x, 100x), the gas difference should still be
+        // unlikely to exceed 0 units.
+        final var priceInTinyCents =
+                Math.max(minimumPriceInTinyCents, tinybarValues.asTinyCents(nominalPriceInTinyBars));
```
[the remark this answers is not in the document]
MrValioBg: Agreed. Done.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
MrValioBg: I've added "Used for non-query operations."

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
-        final var priceInTinybars = Math.max(minimumPriceInTinybars, nominalPriceInTinybars);
-        return asGasRequirement(priceInTinybars);
+    public long gasRequirementWithTinyCents(
+            @NonNull final TransactionBody body, @NonNull final AccountID payer, final long minimumPriceInTinyCents) {
+        // If not enabled, make the calculation using the old method.
+        if (!tinybarValues.isGasPrecisionLossFixEnabled()) {
+            return gasRequirementOldWithPrecisionLoss(body, payer, minimumPriceInTinyCents);
+        }
+
+        final var nominalPriceInTinyBars = feeCalculator.applyAsLong(body, payer);
```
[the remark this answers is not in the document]
MrValioBg: What about "computed" ? I've changed it, maybe it suits it better

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
...
      */
-    public long topLevelGasPrice() {
-        return tinybarValues.topLevelTinybarGasPrice();
+    public long viewGasRequirement() {
+        // isCanonicalViewGasEnabled is a temporary feature flag that will be removed in the future.
+        if (!tinybarValues.isCanonicalViewGasEnabled()) {
+            return FIXED_VIEW_GAS_COST;
+        }
+        final var gasRequirement = gasRequirementFromTinyCents(
+                dispatchPrices.canonicalPriceInTinycents(DispatchType.TOKEN_INFO), FIXED_TINY_CENT_GAS_PRICE_COST);
```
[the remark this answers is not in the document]
MrValioBg: Yeah definitely we could consider some improvements that could clarify the confusion

What are you suggesting about the intrinsic gas? I think its not really related to the `SystemContractGasCalculator`, as its not tied to the System contracts, but rather to a `low level call`, which is concept of EVM/Solidity overall.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -29,6 +30,9 @@
 public class SystemContractGasCalculator {
     private static final long FIXED_VIEW_GAS_COST = 100L;
 
+    // This represents the predefined gas price which is $0.000_000_0852 per unit of gas.
```
lukelee-sl, line 33: maybe a bit more explanation like 852_000 = $0.000_000_0852 * 100(cents in dollars) * 100_000_000 (tiny cents in cents) * 1000 (Fee schedule units per tiny cents) or something like that.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -57,44 +59,60 @@ public long gasRequirement(
             @NonNull final AccountID payer) {
         requireNonNull(body);
         requireNonNull(dispatchType);
-        return gasRequirement(body, payer, canonicalPriceInTinybars(dispatchType));
+        requireNonNull(payer);
+        // isGasPrecisionLossFixEnabled is a temporary feature flag that will be removed in the future.
+        if (!tinybarValues.isGasPrecisionLossFixEnabled()) {
+            return gasRequirementOldWithPrecisionLoss(body, payer, canonicalPriceInTinybars(dispatchType));
+        }
```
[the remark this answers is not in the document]
lukelee-sl: I still don't see why you cannot just call `gasRequirementWithTinyCents` here?  Will this not return the same thing?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -186,8 +187,13 @@ public long lazyCreationCostInGas(@NonNull final Address recipient) {
...
                 .build();
-        final var createFee = gasCalculator.canonicalPriceInTinybars(synthCreation, payerId);
-        return (createFee) / gasCalculator.topLevelGasPrice();
+        final var createFee = gasCalculator.feeCalculatorPriceInTinyBars(synthCreation, payerId);
+
+        // isGasPrecisionLossFixEnabled is a temporary feature flag that will be removed in the future.
+        if (!contractsConfig.isGasPrecisionLossFixEnabled()) {
+            return (createFee) / gasCalculator.topLevelGasPrice();
+        }
+        return (createFee) * FEE_SCHEDULE_UNITS_PER_TINYCENT / gasCalculator.topLevelGasPriceInTinyBars();
```
lukelee-sl, line 196: Do we not want to do some sort of compare here in case the canonical price is more?

lukelee-sl (commented):
Left a couple of questions but overall looks good.  Thank you @MrValioBg

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
MrValioBg: This is how it was before. I guess we could reconsider how the pricing is structured, however I have a gut feeling this could end up costing a looot of gas.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
MrValioBg: Done. I've adjusted it.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
MrValioBg: This is a great point. To address the confusion: previously, the name `canonicalPriceInTinybars` was very misleading as it was not the canonical price. That's why the method now says `feeCalculatorPriceInTinyBars`.

However, I think this should definitely be checked, as from what I debugged, just now, it seems that `feeCalculatorPriceInTinyBars` returns 0, so this may be another flaw in the previous system, as it seems to have returned 0 before as well.

Careful consideration should be given, even though it seems like a flaw, as this may increase the gas by hundreds of thousands if we decide to go with the canonical price.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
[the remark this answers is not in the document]
MrValioBg: The function params are different and wont be that easy to remove it later, imo

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
lukelee-sl: Added this issue to address at a future time.
https://github.com/hashgraph/hedera-services/issues/15337

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -29,6 +30,11 @@
 public class SystemContractGasCalculator {
     private static final long FIXED_VIEW_GAS_COST = 100L;
 
+    // This represents the predefined gas price which is $0.000_000_0852 per unit of gas.
+    // 852_000 = $0.000_000_0852 * 100(cents in dollars) * 100_000_000 (tiny cents in cents) * 1000 (Fee schedule units
+    // per tiny cents). For more info -> https://hedera.com/blog/rolling-smart-contract-hedera-api-fees-into-gas-fees
+    private static final long FIXED_TINY_CENT_GAS_PRICE_COST = 852_000L;
```
tinker-michaelj, line 36: Why do we need to hard code this instead of reading it from the fee schedule?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/CustomGasCalculator.java
```diff
@@ -104,15 +104,15 @@ public long getEdSignatureVerificationSystemContractGasCost() {
 
     /**
      * Logically, would return the gas cost of storing the given number of bytes for the given number of seconds,
-     * given the relative prices of a byte-hour and a gas unit in tinybar.
+     * given the relative prices of a byte-hour and a gas unit in tinyCent.
```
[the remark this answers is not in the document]
stoyanov-st: renamed all occurrences in smart-contracts.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
stoyanov-st: Changed to get gas price from the fee schedule

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/SystemContractGasCalculator.java
```diff
@@ -29,6 +30,11 @@
 public class SystemContractGasCalculator {
     private static final long FIXED_VIEW_GAS_COST = 100L;
 
+    // This represents the predefined gas price which is $0.000_000_0852 per unit of gas.
+    // 852_000 = $0.000_000_0852 * 100(cents in dollars) * 100_000_000 (tiny cents in cents) * 1000 (Fee schedule units
```
david-bakin-sl, line 34: "cents in dollars" -> "cents per dollar"; same for the "tinycents in cents"

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
```diff
@@ -115,17 +152,50 @@ public long childTransactionTinybarGasPrice() {
...
      * {@code service} fee component.
      *
      * @return the tinybar-denominated price of a rbh for the current operation
      */
-    public long topLevelTinybarRbhPrice() {
-        return asTinybars(
-                topLevelResourcePrices.basePrices().servicedataOrThrow().rbh()
-                        / FEE_SCHEDULE_UNITS_PER_TINYCENT
-                        * topLevelResourcePrices.congestionMultiplier());
+    public long topLevelTinycentRbhPrice() {
```
david-bakin-sl, line 175: Name here doesn't match code at L177 or the Javadoc.

(Which kind of makes me wonder how far the mismatch extends and affects things...)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
```diff
@@ -115,17 +152,50 @@ public long childTransactionTinybarGasPrice() {
...
+     * Returns the tinycent gas price for dispatching a child transaction based on the current exchange rate,
+     * Without denomination, but with congestion multiplier, saving the precision.
+     * @return the tinycent gas price
+     */
+    public long childTransactionTinycentGasPrice() {
+        if (childTransactionResourcePrices == null) {
+            throw new IllegalStateException("Cannot dispatch a child transaction from a query");
+        }
+        return childTransactionResourcePrices.basePrices().servicedataOrThrow().gas()
+                * childTransactionResourcePrices.congestionMultiplier();
```
david-bakin-sl, line 165: This class still confuses me.  Why is there a `/ FEE_SCHEDULE_UNITS_PER_TINYCENT` up at L151 but not here?

david-bakin-sl (commented):
Just comments, again.  Questions from parts I still don't understand.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
stoyanov-st: I'll update the javadoc.
The code at l177 is the old code that is producing the precision loss because of the division there. It shall be removed with [#15098](https://github.com/hashgraph/hedera-services/issues/15098) when we verify that the calculations are working as expected.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
stoyanov-st: It is different denomination.
Here we return the price in tynicents while at L151 we return in tinybars.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/gas/TinybarValues.java
stoyanov-st: The conversion tinycents -> tinybars was producing the precision loss because of the divisions and the exchange rate. We actually don't need that conversions to calculate the gas but for backwards compatibility we kept the old code (that uses tinybars to calculate the gas) under the feature flag. Some of the methods are still present and only used by the `SystemContracGasCalculator` methods that are annotated as `Deprecated`.

stoyanov-st:
Closing due to DCO issues.
New PR with changes #15446

PR #14843, opened 2024-08-14
fix: call with amount against HTS contract

### LINKED ISSUE #13397: Executing an internal call with amount against HTS contract returns success
### Description

According to the [in-equvalence doc](https://www.notion.so/limechain/Hedera-Ethereum-precompiles-equivalence-shared-26657bbc66394b4b92622ac99d0b907d#7c606a0951fd49cab6fd268e7eb43d87) the internal calls to the HTS contract with amount should fail with INVALID_FEE_SUBMITTED (except for the token create scenario). Currently, these calls pass with SUCCESS but it doesn't comply to the wanted behaviour.

### Steps to reproduce

To reproduce you can checkout this branch and examine the hapiTest: https://github.com/hashgraph/hedera-services/tree/internal-call-with-amount-against-HTS-returns-success

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.50.0

### Operating system

macOS

mustafauzunn:
**Description**:

Internal calls to the Hedera system contract with amount should fail.

The main logic is in `CustomMessageCallProcessor`. We check if the current call is `createToken`, because this is the only case we allow sending value to Hedera system contracts and if not we fail.

**Related issue(s)**:

Fixes #13397 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/MiscTokenTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/MiscTokenTest.java
@@ -45,7 +45,7 @@ public class MiscTokenTest {
 
     @HapiTest
     @DisplayName("cannot transfer value to HTS")
-    public Stream<DynamicTest> cannotGetNonsenseKeyType() {
+    public Stream<DynamicTest> cannotTransferValueToHts() {
         return hapiTest(internalCall
                 .call("isATokenWithCall", fungibleToken)
                 .sending(100L)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -118,7 +131,14 @@ public void start(@NonNull final MessageFrame frame, @NonNull final OperationTra
         // paid using gas; for example, when creating a new token. But the system contract
         // only diverts this value to the network's fee collection accounts, instead of
         // actually receiving it.
+        // We do not allow sending value to Hedera system contracts except in the case of token creation.
         if (systemContracts.containsKey(codeAddress)) {
+            if (!checkIfCreateScenario(frame)) {
```
lukelee-sl, line 136: Nit: consider renaming `checkIfCreateScenario` to `ifTokenCreation`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -164,6 +184,37 @@ public void start(@NonNull final MessageFrame frame, @NonNull final OperationTra
...
+     * @param frame the message frame to check
+     * @return true if the input data matches any of the known create selectors, false otherwise
+     */
+    private boolean checkIfCreateScenario(MessageFrame frame) {
+        if (frame.getInputData().isEmpty()) {
+            return false;
+        }
+        var selector = frame.getInputData().slice(0, 4).toArray();
+        return Arrays.stream(new byte[][] {
+                    CREATE_FUNGIBLE_TOKEN_V1.selector(),
```
lukelee-sl, line 202: Is there a way to have a single list of selectors to check here and in `CreateTranslator` so that we do not have to remember to maintain both lists if we add supported selectors?

lukelee-sl (dismissed):
LG overall.  A few small comment you can decide to address or not.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
MrValioBg: nit: Or maybe `isTokenCreate` / `isTokenCreation`, so it can be read `if is token create`, instead of `if if token create` 😜

stoyanov-st (approved):
LGTM, ty @mustafauzunn !

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/InternalCallTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.Tag;
+
+@Tag(SMART_CONTRACT)
+@DisplayName("internalCall")
+@SuppressWarnings("java:S1192")
+@HapiTestLifecycle
+public class InternalCallTest {
+
+    @Contract(contract = "InternalCall", creationGas = 1_000_000L)
```
MrValioBg, line 40: Hmm, do we need a separate class for this? I guess if this test does not make sense in other suites, we could maybe rename it to MiscTokenTest or something similar so we could add more tests that "do not fit" in the future.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateTranslator.java
```diff
@@ -118,6 +119,26 @@ public class CreateTranslator extends AbstractCallTranslator<HtsCallAttempt> {
                     + ")",
             "(int64,address)");
 
+    /**
+     * A list of `Function` objects representing various create functions for fungible and non-fungible tokens.
+     * This list is used in {@link com.hedera.node.app.service.contract.impl.exec.processors.CustomMessageCallProcessor}
+     * to determine if a given call attempt is a creation call, because we do not allow sending value to Hedera system contracts
+     * except in the case of token creation
+     */
+    public static final List<Function> CREATE_FUNCTIONS = List.of(
```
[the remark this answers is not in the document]
lukelee-sl: This seems like a good suggestion to me.

MrValioBg (approved):
LGTM! Thank you!

MrValioBg (approved):

PR #14846, opened 2024-08-14
feat: create translator for contract call transaction

### LINKED ISSUE #14845: Create `ContractCallTranslator`

petreze:
**Description**:
Translator for contract create transaction to convert block items to record stream entries

**Related issue(s)**:

Fixes #14845 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java
@@ -83,6 +83,7 @@ public class BlockStreamTransactionTranslator implements TransactionRecordTransl
 
         final var singleTxnRecord =
                 switch (txnType) {
+                    case ContractCall -> new ContractCallTranslator().translate(txnWrapper, stateChanges);
                     case UtilPrng -> new UtilPrngTranslator().translate(txnWrapper, stateChanges);
                     default -> new SingleTransactionRecord(
                             txnWrapper.txn(),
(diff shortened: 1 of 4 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/UtilPrngTranslator.java
```diff
@@ -32,7 +33,7 @@ public class UtilPrngTranslator implements TransactionRecordTranslator<SingleTra
 
     @Override
     public SingleTransactionRecord translate(
-            @NotNull SingleTransactionBlockItems transaction, @NotNull StateChanges stateChanges) {
+            @NotNull SingleTransactionBlockItems transaction, @Nullable StateChanges stateChanges) {
```
iwsimon, line 36: Should use @NonNull above.

petreze: @iwsimon 
`stateChanges` might come as `null` for certain transactions. This is why we mark them with nullable even in the [BlockStreamTransactionTranslator](https://github.com/hashgraph/hedera-services/blob/stream-manager-spike/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BlockStreamTransactionTranslator.java#L71)

iwsimon: I meant using @NonNull (from ackage edu.umd.cs.findbugs.annotations) instead of @NotNull.

mhess-swl (approved):

PR #14861, opened 2024-08-14
feat: Create RosterMap map and RosterState singleton

### LINKED ISSUE #14707: Create RosterMap map and RosterState singleton for States API (java code + register new named entities in States API)
Unsure about the States API-related process to create new named entities, so the estimate may be unclear.

kfa-aguda:
Fixes #14707 

As part of the TSS Roster proposal implementation, a new schema is required to define the `RosterState` singleton and `RosterMap` states API. An associated Service will also be defined. When this change is delivered, the platform will become aware of the new entities as well as an associated service (non-rpc exposing) to manage those entities.


hapi/build.gradle.kts
```diff
@@ -34,7 +34,7 @@ tasks.cloneHederaProtobufs {
     // uncomment below to use a specific tag
     // tag = "v0.53.0"
     // uncomment below to use a specific branch
-    branch = "add-throttle-at-consensus"
```
kfa-aguda, line 37: this should never have made it to develop but I'm adding a note here so mine doesn't get merged until I update back to main.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/services/ServicesRegistryImpl.java
```diff
@@ -65,10 +66,14 @@ public ServicesRegistryImpl(
...
      * Register the given service.
      *
-     * @param service The service to register
+     * @param schemaAware The implementation to register
      */
     @Override
-    public void register(@NonNull final Service service) {
+    public void register(@NonNull final SchemaAware schemaAware) {
+        requireNonNull(schemaAware);
+        if (!(schemaAware instanceof final Service service)) {
```
imalygin, line 74: Why though? There is nothing that prevents this class from handling your new class. It should probably be called `SchemaAwareRegistryImpl` and its interface should be renamed as well. 

Now you see the challenge with renaming of the interface `Service` into `SchemaAware`. There multiple implicit dependencies you discover only after you try this renaming.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/RosterSchemaRegistryImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import com.hedera.node.app.roster.schemas.V0540RosterSchema;
+import com.swirlds.state.spi.SchemaAware;
+import com.swirlds.state.spi.SchemaRegistry;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * A {@link com.hedera.hapi.node.state.roster.Roster} implementation of the {@link SchemaAware} interface.
+ * Registers the roster schemas with the {@link SchemaRegistry}.
+ */
+public class RosterSchemaRegistryImpl implements SchemaAware {
```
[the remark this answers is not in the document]
kfa-aguda: i struggled with that one to be honest. Is `RosterSchemaRegistrar` better? I had called it that before impulsively changing it to this. Initial name I used was `RosterSchemaAwareImpl`. Help!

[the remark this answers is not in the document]
kfa-aguda: OK I've picked `RosterSchemaAwareImpl`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/services/ServicesRegistryImpl.java
kfa-aguda: it does other things at the service level though which i don't want to be visible to (lower level) schema aware.
I didn't set out to rename `Service` to `SchemaAware`, but to extract a lower level `SchemaAware` out of it. 
This option is minimally invasive, and achieves the same purpose. The instanceOf isn't ideal but it's better than casting.

imalygin: My point is that this is the only spot in the code where `MerkleSchemaRegistry` is created. 

Here is how the callstack goes:
- `ServicesRegistry#register` in `com.hedera.node.app.Hedera#Hedera`. 
- `SchemaAware#registerSchemas` in `com.hedera.node.app.services.ServicesRegistryImpl#register`
- `SchemaRegistry#register` in `SchemaAware#registerSchemas`. And this is where an instance of a schema finally gets in the registry 

If you have this condition in place
```
   if (!(schemaAware instanceof final Service service)) {
```
your schema is not going to be registered.

kfa-aguda: Thanks. On a second look, even if i fixed this issue you pointed (thank you), there's much more going on with the classId generation using the service name (which i'd have to stop providing) in `StateUtils`. I conclude this "only services use schemas" assumption is deep-rooted, and can't be removed trivially.

        // NOTE: Once this is live on any network, the formula used to generate this key can NEVER
        // BE CHANGED or you won't ever be able to deserialize an exising state! If we get away from
        // this formula, we will need to hardcode known classId that had been previously generated.

I'd rather avoid this work. I'll have to create a dummy service :(

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V0540RosterSchema.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Set;
+
+/**
+ * Schema for the TSS Base service
+ */
+public class V0540RosterSchema extends Schema {
+    public static final String ROSTER_KEY = "ROSTER";
+    public static final String ROSTER_STATES_KEY = "ROSTER_STATE";
+    //
+    private static final long MAX_ROSTERS = 8L;
```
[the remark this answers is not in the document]
rbair23: IIRC, this number is not possible to increase later, but there isn't much of a cost of making it a fair amount larger to start with. @jasperpotts or @artemananiev would maybe have a better number, but I think you could say 1024 or even 2^16 should be fine and more than enough.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/services/ServicesRegistryImpl.java
rbair23: I don't think we need to introduce a SchemaAware. We have an example of a service that doesn't have a public API in the BlockRecordService.

rbair23 (changes requested):
I think you can just extend `Service` the same way `BlockRecordService` does, I don't think you need to create `SchemaAware`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V0540RosterSchema.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Set;
+
+/**
+ * Roster Schema
+ */
+public class V0540RosterSchema extends Schema {
+    public static final String ROSTER_KEY = "ROSTER";
+    public static final String ROSTER_STATES_KEY = "ROSTER_STATE";
+    /** this can't be increased later so we pick some number large enough, 2^16. */
+    private static final long MAX_ROSTERS = 65536L;
```
anthony-swirldslabs, line 36: I'm a bit concerned about this number. If we consider the DAB is implemented and rosters indeed change once a day, this upper limit allows the map to store ~180 years worth of rosters, which is almost two centuries. I don't believe it's reasonable to store that many entries in this map for all practical purposes. In two centuries, I believe this map implementation will support dynamic resizing (provided it still exists by that time.)

On the flip side, this maximum limit may define sizes of some resources (RAM, disk, and whatnot) being allocated by the implementation today. And this extra space would simply be wasted because there's no way for us to populate that many rosters in any reasonable time.

I suggest to drastically reduce this size. I think that 256, while still pretty large in my opinion, would still be a lot more reasonable than the 64K value you're using here.

kfa-aguda: See https://github.com/hashgraph/hedera-services/pull/14861#discussion_r1720307951

anthony-swirldslabs: Yes, I saw that. But my point still stands: I'm not convinced that 64K is a good number per my arguments above.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/RosterServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import com.swirlds.state.spi.SchemaRegistry;
+import com.swirlds.state.spi.Service;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * A {@link com.hedera.hapi.node.state.roster.Roster} implementation of the {@link Service} interface.
+ * Registers the roster schemas with the {@link SchemaRegistry}.
+ * Not exposed outside `hedera-app`.
+ */
+public class RosterServiceImpl implements Service {
```
lpetrovic05, line 31: why not call it just `RosterService`?

kfa-aguda: no other reason really apart from it just being consistent with other Service implementations.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V0540RosterSchema.java
kfa-aguda: I'm not convinced any number > 0 is a good or bad number until i fully understand how this is used, which i can't seem to find (hence why I blindly accepted @rbair23's suggestion). @artemananiev any pointers to how this number is used?

artemananiev: This `maxNumberOfKeys` hint is used to calculate the size of "keys to paths" bucket index in MerkleDb. If the hint is 64K, index size will be 2K, which is 16K bytes (every index entry is a long)

artemananiev: Note that with Virtual Mega Map effort, all these hints will become obsolete anyway

anthony-swirldslabs: Thanks for the details. While 16KB doesn't sounds like much, I still feel that having an index of 2K entries for this particular collection doesn't seem reasonable based on the arguments that I've provided above.

I'm assuming that we divide the hint value by 32 to determine the size of the index.

With this assumption in mind, I suggest to set the hint to 8192. I think this number is large enough to support any reasonable use-cases for this map. And at the same time it's small enough to result in an index of 256 entries/2KB in size which is more than enough for any immediate and mid-term use-cases.

Since the hints will be deprecated, the value we choose today wouldn't really matter in the long run.

anthony-swirldslabs (dismissed):

anthony-swirldslabs (dismissed):

imalygin (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V0540RosterSchema.java
```diff
@@ -0,0 +1,78 @@
...
+            // This migration code is really, for now, a default value provider.
+            // At genesis we put empty roster and roster state into the state
+            // as serialization/deserialization fails on null.
+            // It is important to distinguish this from the Genesis Roster,
+            // which will be provided by the Execution layer depending on
+            // the desired network mode
+            final var rosterState = ctx.newStates().getSingleton(ROSTER_STATES_KEY);
+            rosterState.put(RosterState.DEFAULT);
+            final var roster = ctx.newStates().get(ROSTER_KEY);
+            roster.put(ProtoBytes.DEFAULT, Roster.DEFAULT);
```
anthony-swirldslabs, line 75: I'm unsure about the line 75. It populates the map with a random entry that is never going to be removed (unless we implement a special logic somewhere, which is ugly.)

Can the map be initially empty? I.e. have no entries at all. If that's impossible, then my suggestion would be to populate it with the actual current roster as found in the PlatformState (i.e. the AddressBook converted to Roster.)

In fact, same applies to the RosterState just above. I think it should point to the actual active roster from the beginning and have a corresponding entry in the list of pairs<round, roster_hash>.

It would be difficult and unintuitive to support the initialization of these data structures somewhere else. I believe this initial initialization belongs here.

kfa-aguda: on revisiting, it makes sense to have an initial empty map, but i wasn't sure of the state behaviour. I just tried with a local test and it shows it's ok to do, so i've removed it. Thank you.

But the rosterstate has to stay, as it has to be initialized. It will be subsequently initialized here depending on the network mode configuration, but the flow to arrive at that configuration is not yet done. For now, a default value provider is sufficient.

anthony-swirldslabs (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V0540RosterSchema.java
kfa-aguda: we agreed to leave this as-is.

rbair23 (approved):

PR #14881, opened 2024-08-15
fix: invalid `feeSchedules.json`

Neeharika-Sompalli:
Fixed invalid feeSchedules.json in both testnet and mainnet directories

### DIFF
--- a/hedera-node/configuration/mainnet/upgrade/feeSchedules.json
+++ b/hedera-node/configuration/mainnet/upgrade/feeSchedules.json
@@ -1687,7 +1687,10 @@
                 "min": 0,
                 "max": 1000000000000000
               }
-            },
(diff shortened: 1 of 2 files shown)


tinker-michaelj (approved):

dalvizu (approved):
Don't want to hold up review but we should definitely have a basic CI check seeing that these files are valid json

Should milestone be 52 and not 54?

Neeharika-Sompalli:
> Don't want to hold up review but we should definitely have a basic CI check seeing that these files are valid json
> 
> Should milestone be 52 and not 54?

Yes agreed! Changed that and we will add CI check. Thanks @dalvizu . Will also cherry-pick this to develop, release/0.53

PR #14889, opened 2024-08-16
feat: characterize HIP-904 interaction with aliases, hollow accounts, receiver sig req

### LINKED ISSUE #14878: Write characterizing tests for alias/receiver sig/hollow account interactions with HIP-904
### Problem

As HIP-904 signing requirements include some subtle variations on the normal `CryptoTransfer` signing requirements, we need to ensure we have tests that characterize the exact behavior in all cases.

### Solution

Add a new test class focused on these HIP-904 interactions.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14878


hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TokensConfig.java
```diff
@@ -48,12 +48,12 @@ public record TokensConfig(
                 ScaleFactor nftsMintThrottleScaleFactor,
         @ConfigProperty(value = "nfts.useVirtualMerkle", defaultValue = "true") @NetworkProperty
                 boolean nftsUseVirtualMerkle,
-        @ConfigProperty(value = "maxPendingAllowedAirdrops", defaultValue = "20000000") @NetworkProperty
-                long maxPendingAllowedAirdrops,
-        @ConfigProperty(value = "maxAllowedPendingAirdropsToCancel", defaultValue = "10") @NetworkProperty
-                long maxAllowedPendingAirdropsToCancel,
+        @ConfigProperty(value = "maxAllowedAirdrops", defaultValue = "20000000") @NetworkProperty
```
Neeharika-Sompalli, line 51: We should probably include `Pending` in the name. I think Atul suggested the name should have pending

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenClaimAirdropHandler.java
```diff
@@ -151,38 +153,42 @@ public void handle(@NonNull HandleContext context) throws HandleException {
...
+                op.pendingAirdrops(),
+                EnumSet.of(AirdropHandlerHelper.IdType.RECEIVER));
+        for (final var airdrop : standardAirdropIds) {
             final var tokenId = airdrop.hasFungibleTokenType()
                     ? airdrop.fungibleTokenTypeOrThrow()
                     : airdrop.nonFungibleTokenOrThrow().tokenIdOrThrow();
             getIfUsable(tokenId, tokenStore);
-            validateTrue(pendingAirdropStore.exists(airdrop), INVALID_PENDING_AIRDROP_ID);
-            validateTrue(
-                    validator.tokenHasNoRoyaltyWithFallbackFee(tokenId, tokenStore),
```
Neeharika-Sompalli, line 183: We still need this check here right ?

tinker-michaelj: Ah great catch, fixed. 🙏

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TokensConfig.java
tinker-michaelj: Agreed, fixed.

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj 🙏

Issue #14852, opened 2024-08-14
Fix `ConsensusSubmitMessage` StreamValidation diffs from `SubmitMessageSuite`


MiroslavGatsanoga:
After https://github.com/hashgraph/hedera-services/pull/14673 gets merged, there are still 4 ConsensusSubmitMessage diffs left, that  are failing with:
```
Mismatched field counts  (found but did not expect [parent_consensus_timestamp] between expected receipt ...
```

The issue seems to be that for the block stream transaction result items which are produced by HapiTests resulting in error status during handle (e.g. `messageSubmissionOverSize`, `chunkTransactionIDIsValidated`, `chunkNumberIsValidated`) then the produced transaction result block stream item is with empty fields e.g. `parent_consensus_timestamp` is set with zero values, status is OK (0), etc.

The behaviour is somewhat flaky and sometimes the correct values are set. Also if only `messageSubmissionOverSize` is run (by commenting out all other HapiTests in the suite) then the correct values are set as well and there are no ConsensusSubmitMessage StreamValidation diffs.

It looks like [here](https://github.com/hashgraph/hedera-services/blob/stream-manager-spike/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java#L237) for some reason we don't produce the transaction result block item correctly.

PR #14893, opened 2024-08-16
fix: ConsensusSubmitMessage StreamValidation diffs from SubmitMessageSuite

MiroslavGatsanoga:
**Description**:

For tests resulting in `HandleException` we'll end up rolling back the save point stack and which invokes `BlockStreamBuilder.nullOutSideEffectFields()`.
Removed the reset for `status` and `parentConsensusTimestamp` from there in order to have the similar behavior as in `RecordStreamBuilder.nullOutSideEffectFields()`.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/14852

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
@@ -1032,8 +1032,6 @@ public class BlockStreamBuilder
         transactionResultBuilder.scheduleRef((ScheduleID) null);
         transactionResultBuilder.automaticTokenAssociations(emptyList());
         transactionResultBuilder.congestionPricingMultiplier(0);
-        transactionResultBuilder.status(ResponseCodeEnum.OK);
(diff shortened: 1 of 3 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
```diff
@@ -1032,8 +1032,6 @@ public void nullOutSideEffectFields() {
         transactionResultBuilder.scheduleRef((ScheduleID) null);
         transactionResultBuilder.automaticTokenAssociations(emptyList());
         transactionResultBuilder.congestionPricingMultiplier(0);
-        transactionResultBuilder.status(ResponseCodeEnum.OK);
-        transactionResultBuilder.parentConsensusTimestamp(Timestamp.DEFAULT);
```
mhess-swl, line 1035: I'm ok with this, as long as we're sure to set these properties in all relevant cases

mhess-swl (approved):

PR #14904, opened 2024-08-18
fix: refactor to use State API idioms

tinker-michaelj:
**Description**:
 - Remove all uses of `PlatformStateAccessor` outside `com.swirlds` packages in favor of State API access to platform state.
 - Remove redundant `PlatformStateAccessorSingleton` from `MerkleStateRoot` (the singleton will be created during migration of the `PlatformServiceState` schema).
 - Extract `PlatformStateService` into a standalone package with conventional structure.
     * In particular, add `ReadablePlatformStateStore` and `WritablePlatformStateStore` for read-only and read-write access to the platform state, encapsulating PBJ translation logic.
 - Add a `MerkleStateLifecycles.initPlatformState(State)` callback to allow `MerkleStateRoot` to reuse the same `Schema` migration logic that all other `Service` implementations use for managing its `getPlatformState()` implementation.
    * Update `EmbeddedHedera` to incorporate this callback.
 - Fix two race conditions in `WritableSingletonStateBase` that had never surfaced because Services does not access any Singleton state concurrently:
     1. Prevent a concurrent `get()` from overwriting the value being `put()` simultaneously with stale value from the backing store.
     2. Prevent a concurrent `get()` from returning `null` after a concurrent `put()` has set the `modified` flag to `true`, but _before_ setting the `value` field.
 - Fix all unit tests.
 - Merge `develop`.


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterValidator.java
```diff
@@ -0,0 +1,107 @@
...
+import com.hedera.hapi.node.state.roster.RosterEntry;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A Roster validator.
+ */
+public final class RosterValidator {
```
anthony-swirldslabs, line 30: This is a part of a separate PR/commit. I think you want to rebase your branch and sync with `develop` properly so that it only includes your actual changes.

imalygin: sure. will do

PR #14964, opened 2024-08-20
chore: Added metric that measures the number of queried account balances

### LINKED ISSUE #14962: Metric for number of queried account-balances
Add a metric that shows the number of account balances that are requested as part of `GetAccountInfo` per second.

netopyr:
**Description**:

This PR adds a metric that measures the number of queried account balances per second.

**Related issue(s)**:

Fixes #14962


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountInfoHandler.java
```diff
@@ -69,18 +73,28 @@
  */
 @Singleton
 public class CryptoGetAccountInfoHandler extends PaidQueryHandler {
+
+    private static final SpeedometerMetric.Config BALANCE_SPEEDOMETER_CONFIG = new SpeedometerMetric.Config(
+                    "app", "queriedAccountBalances")
+            .withDescription("Number of account balances requested in GetAccountInfo queries per second");
```
Neeharika-Sompalli, line 79: Should be token balances in specific?

Neeharika-Sompalli (dismissed):
LGTM except one comment. Thanks @netopyr

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountInfoHandler.java
netopyr: Right, good point. Will fix it.

Issue #13901, opened 2024-06-17
VirtualPipeline back-pressure is too harsh 


OlegMazurov:
### Problem

From `swirlds-vmap.log`:
```
2024-06-13 15:52:17.177 18087708 DEBUG VIRTUAL_MERKLE_STATS <<scheduler consensusRoundHandler>> VirtualPipeline: Total size backpressure: 87025 ms
```
According to profile, the lifecycle thread was idle most of that time.
The long delay was caused by a large accumulated size of `VirtualNodeCache` (~8g), which was due to stalled `flush` waiting for a long snapshot writing indexes to disk (see #11845).

### Solution

* reconsider the formula to compute delay in `VirtualPipeline.calculateFamilySizeBackpressurePause()`
* reconsider back-pressure implementation 
* reduce flush stalls, etc.

PR #15003, opened 2024-08-21
fix: 13901: VirtualPipeline back-pressure is too harsh

artemananiev:
Fix summary: changed `VirtualPipeline.applyFamilySizeBackpressure()` to be a set of small sleeps rather a single long sleep. After each small sleep, it checks the current backpressure (which may have changed, if a virtual map copy is flushed in a parallel thread) and compares it with total time slept so far.

Fixes: https://github.com/hashgraph/hedera-services/issues/13901

### DIFF
--- a/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
+++ b/platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
@@ -232,8 +232,10 @@ public class VirtualPipeline {
         }
 
         try {
+            final long sleepStartTime = System.currentTimeMillis();
             logger.debug(VIRTUAL_MERKLE_STATS.getMarker(), "Flush backpressure: {} ms", sleepTimeMillis);
             MILLISECONDS.sleep(sleepTimeMillis);
+            statistics.recordFlushBackpressureMs((int) (System.currentTimeMillis() - sleepStartTime));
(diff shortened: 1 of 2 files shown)


imalygin (dismissed):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/pipeline/VirtualPipeline.java
```diff
@@ -256,39 +256,55 @@ Duration calculateFlushBackpressurePause() {
         final Duration sleepTime = CompareTo.min(computedSleepTime, maxSleepTime);
         final int sleepTimeMillis = (int) sleepTime.toMillis();
         statistics.recordFlushBackpressureMs(sleepTimeMillis);
```
OlegMazurov, line 258: Should statistics be updated every time we calculate a new value or only when we apply it (similar to `applyFamilySizeBackpressure()`)?
The two backpressure methods should be consistent in that regard.

artemananiev: Good catch. I didn't realize the metrics were reported in the "calculate" method, not when the backressure is actually applied. With my current changes it was fixed for family size backpressure, but not for flush backpressure. I'll change it accordingly

OlegMazurov (approved):

OlegMazurov (approved):

imalygin (approved):

PR #15004, opened 2024-08-21
refactor: remove AbortException

### LINKED ISSUE #15002: Remove AbortException during smart contract processing
### Problem

A custom exception called `AbortException` was introduced in order to duplicate mono service behavior during processing of a smart contracts transaction.  This exception can be removed and replaced with the existing `HandleException`

### Solution

Remove the AbortException class and replace handling with `HandleException`

### Alternatives

_No response_

lukelee-sl:
**Description**:
Remove AbortException and replace with standard HandleException

**Related issue(s)**:

Fixes #15002

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
```diff
@@ -224,7 +223,7 @@ void stillChargesHapiFeesOnAbort() {
                 .willReturn(HEVM_CREATION);
         given(processor.processTransaction(
                         HEVM_CREATION, rootProxyWorldUpdater, feesOnlyUpdater, hederaEvmContext, tracer, CONFIGURATION))
-                .willThrow(new AbortException(INVALID_CONTRACT_ID, SENDER_ID));
+                .willThrow(new HandleException(INVALID_CONTRACT_ID));
```
david-bakin-sl, line 226: Test could (should?) be renamed to note that this is "still charges fees for aborted contract _create_" (distinguished from the case of _call_).

david-bakin-sl (commented):
Still thinking.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
lukelee-sl: I removed the distinction between contract call and other types of transactions.  What I really wanted to do was to see if fees had or had not been collected before the transaction was aborted.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -135,30 +137,15 @@ public CallOutcome call() {
...
-                    result.asProtoResultOf(ethTxDataIfApplicable(), rootProxyWorldUpdater), result);
+        } catch (HandleException e) {
+            final var sender = rootProxyWorldUpdater.getHederaAccount(hevmTransaction.senderId());
+            final var senderId = sender != null ? sender.hederaId() : hevmTransaction.senderId();
+
+            return maybeChargeFeesAndReturnOutcome(
+                    hevmTransaction.withException(e),
+                    senderId,
+                    sender,
+                    hevmTransaction.isContractCall() && contractsConfig.chargeGasOnPreEvmException());
```
david-bakin-sl, line 148: I'm not sure why a config named `chargeGasOnPreEvmException` controls this (emphasis on the **`Pre`**), since this handler encloses `processTransaction` where the EVM is run.

(That gate is up at L113 where I'd expect it to be.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -110,7 +111,8 @@ public CallOutcome call() {
         // the error.
         final var hevmTransaction = safeCreateHevmTransaction();
         if (hevmTransaction.isException() && contractsConfig.chargeGasOnPreEvmException()) {
-            return chargeFeesAndReturnOutcome(hevmTransaction);
+            return maybeChargeFeesAndReturnOutcome(
+                    hevmTransaction, context.body().transactionIDOrThrow().accountIDOrThrow(), null, true);
```
david-bakin-sl, line 115: At this point it might be the case that the sender account is already known to the world updater - maybe?  If so then is it worth trying to look it up in case it's there, instead of just providing `null` as the sender account?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransaction.java
```diff
@@ -127,4 +127,23 @@ public long offeredGasCost() {
     public boolean requiresFullRelayerAllowance() {
         return offeredGasPrice == 0L;
     }
+
+    /**
+     * @return a copy of this transaction with the given {@code exception}
+     */
+    public HederaEvmTransaction withException(@Nullable final HandleException exception) {
```
david-bakin-sl, line 134: I suppose if you pass `null` here you can _remove_ an exception, but is that ever going to happen (intentionally)?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
david-bakin-sl: I thought there still was a distinction at `ContextTransactionProcessor` L148?

Plus, charging fees is dependent on that config variable `chargeGasOnPreEvmException` so shouldn't this be one of the property-overriding tests? (and same for `stillChargesHapiFeesOnHevmException` and `stillChargesHapiFeesOnExceptionThrown`?  (In order to ensure it is set to `true`, at least.)

lukelee-sl: Yes it turned out that this distinction was important for the existing test.  Thinking about it further, I realized that we only really care about `ContractCall` transactions because these are the only ones that have gas only fees charged.  So I take back my earlier comment above 😛.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
lukelee-sl: The config's purpose may have evolved since it was first thought of.  I will changed the name to `chargeGasOnEvmHandleException`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
lukelee-sl: At this point, we did not successfully create the HederaEvmTransaction object so it is not likely that the sender is good.  We only use the sender to return the sender nonce in the results which seems not relevant if we end up here.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/hevm/HederaEvmTransaction.java
lukelee-sl: fixed to be @NullNull

david-bakin-sl (approved):
LGTM (even though one change suggested).

tinker-michaelj (approved):
A clear simplification, but seems to cover all the same bases...

LGTM, tyvm @lukelee-sl !

Issue #14406, opened 2024-07-25
investigate: Gradle Determinism issues


rbarkerSL:
The step gradle determinism is failing: 

```
/home/runner/work/hedera-services/hedera-services/platform-sdk/swirlds-logging/src/main/java/com/swirlds/logging/api/internal/level/HandlerLoggingLevelConfig.java:19: warning: [removal] ConfigUtils in com.swirlds.logging.utils has been deprecated and marked for removal

import static com.swirlds.logging.utils.ConfigUtils.configValueOrElse; 
                                                    ^ 
  error: warnings found and -Werror specified 
  1 error
  1 warning
```

```[tasklist]
### Tasks
```

rbarkerSL:
Disabled warnings as errors on javadoc compiler steps

jjohannes:
We would like to keep warnings-as-errors to not generate "broken" Javadoc and endup with a lot of warnings again.

The original issue is unfortunately a shortcoming in Javadoc itself. In such cases, please import the class instead of the method: `import com.swirlds.logging.utils.ConfigUtils`

Then it no longer complains.

Follow up PR: https://github.com/hashgraph/hedera-services/pull/15012

PR #15012, opened 2024-08-21
build: Javadoc warnings are reported as errors

jjohannes:
- This was originally done in #13645
- And then changed in #14413

This PR solves the original issue (#14406) without keeping any warnings.

**Description**:

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/package-info.java
```diff
@@ -1,6 +1,6 @@
 /**
  * The Hedera implementation of the Hedera Node Schedule Service (as defined in
- * {@link com.hedera.node.app.service.schedule}).
+ * {@code com.hedera.node.app.service.schedule}).
```
jjohannes, line 3: Links to packages/modules do not work here, probably due to how Gradle calls the Javadoc tool. It's unclear what exactly does not work, but I wasn't able to fix this quickly.

Note: It also does not work on `develop` right now, but there we get a warning that is ignored.

rbarkerSL (approved):

mxtartaglia-sl:
@jjohannes What is the impact of these on the warning:
```warning: use of default constructor, which does not provide a comment```
I find that warning supper annoying and impossible to solve. Can that one be added as an exception of the validation?

If not, each inner class is forced to have an unnecessary empty private constructor with an unnecessary comment. 
Given that before it was only a warning, one could ignore that in cases where it did not make sense, but if after this PR it becomes a compilation error, then IMO, it would have an -_-  effect.

mxtartaglia-sl (approved):

jjohannes:
@mxtartaglia-sl yes these warnings will be disabled (again) completely by the `-missing` flag in the Javadoc config.
They were re-enabled by mistake in #14413.

jjohannes:
@tinker-michaelj thanks for merging develop. Checks are passing now. Can you also approve this PR please?

Neeharika-Sompalli (approved):
LGTM!

PR #15016, opened 2024-08-21
fix: change order of descriptor variables

### LINKED ISSUE #15013: EventDescriptor variables order reversed
### Issue

In release `0.53`, the `EventDescriptor` PBJ object is being wrongly populated. The variables `birthRound` and `generation` have been switched. This is happening in this [line of code](https://github.com/hashgraph/hedera-services/blob/3747e61503c22d5424d7cd3ff3d0d04c5909f5f7/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java#L156).

In this release, this data is not used for anything else but hashing. So the impact of this only on the hash of the event. This is not compatible with current develop where the order hash been fixed, so the hash of an event is calculated differrently.

### Solutions

There are 2 possible solutions:
- Make a tiny code change in release `0.53` to reverse the order of these variables
- Change the order of the variables in subsequent releases, so that the protobuf remains the same, but the variables are mapped correctly. EDIT: upon further inspection, `generation` has an ordinal of `17` because there is an intent to remove it. changing the order would mess up this plan.

lpetrovic05:
closes #15013

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java
@@ -153,7 +153,7 @@ public class UnsignedEvent extends AbstractHashable {
                 HapiUtils.asTimestamp(timeCreated),
                 this.allParents.stream()
                         .map(ed -> new com.hedera.hapi.platform.event.EventDescriptor(
-                                ed.getHash().getBytes(), ed.getCreator().id(), ed.getGeneration(), ed.getBirthRound()))
+                                ed.getHash().getBytes(), ed.getCreator().id(), ed.getBirthRound(), ed.getGeneration()))
                         .toList(),
                 softwareVersion.getPbjSemanticVersion());
         this.transactions = transactions.stream().map(PayloadWrapper::new).toList();


netopyr (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java
```diff
@@ -153,7 +153,7 @@ public UnsignedEvent(
                 HapiUtils.asTimestamp(timeCreated),
                 this.allParents.stream()
                         .map(ed -> new com.hedera.hapi.platform.event.EventDescriptor(
-                                ed.getHash().getBytes(), ed.getCreator().id(), ed.getGeneration(), ed.getBirthRound()))
+                                ed.getHash().getBytes(), ed.getCreator().id(), ed.getBirthRound(), ed.getGeneration()))
```
anthony-swirldslabs, line 156: As commented on the GitHub issue separately, it would be ideal to avoid using constructors and instead switch to using builders available in every PBJ model, which could help prevent bugs such as this one.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java
poulok: Perhaps the reason this was not done was to avoid two memory allocations for each event

anthony-swirldslabs: Sounds reasonable, provided we have some benchmark data to support this.

poulok (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/events/UnsignedEvent.java
poulok: @OlegMazurov shared that even without the builder, since we now are double hashing events (block stream redaction purposes) the allocation rate jumped significantly, but not enough to affect overall performance. That being the case, we want to be mindful of the allocations.

anthony-swirldslabs: @OlegMazurov also noted in a separate PR that memory allocations for short-lived objects are okay for the performance. So I'm not sure if that number alone is sufficient to make the right choice here.

anthony-swirldslabs: Ref: https://github.com/hashgraph/hedera-services/pull/14832#event-13948835760

edward-swirldslabs (approved):

PR #15019, opened 2024-08-21
feat: metadata view functions via smart contracts

### LINKED ISSUE #14590: Define and implement HTS function call for view functions

mustafauzunn:
**Description**:

This PR introduces new versions for `getTokenInfoV2`,  `getNonFungibleTokenInfoV2` and `getFungibleTokenInfoV2`.
These new versions support metadata and metadata key

**Related issue(s)**:

Fixes #14590 

**Notes for reviewer**:

**Checklist**


hedera-node/configuration/dev/application.properties
```diff
@@ -2,6 +2,7 @@ balances.exportDir.path=data/accountBalances/
 upgrade.artifacts.path=data/upgrade
 contracts.chainId=298
 contracts.maxGasPerSec=15000000000
+contracts.systemContract.tokenInfo.v2.enabled=true
```
lukelee-sl, line 5: we should probably change these to be false for now.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/ReturnTypes.java
```diff
@@ -103,6 +103,36 @@ private ReturnTypes() {
                 + ",string" // ledgerId
             + ")"
         + ")";
+
```
lukelee-sl, line 106: Can we not share a lot of this with `RESPONSE_CODE_TOKEN_INFO` above? Also for the fungible and not versions below.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/TokenTupleUtils.java
```diff
@@ -296,14 +300,43 @@ private static Tuple hederaTokenTupleFor(@NonNull final Token token) {
                 expiryTupleFor(token));
     }
 
+    @NonNull
+    private static Tuple hederaTokenTupleForV2(@NonNull final Token token) {
```
lukelee-sl, line 304: would it be possible (or wise?) to call `hederaTokenTupleFor` and then just add the metadata key?

lukelee-sl (commented):
Looks good overall.  Just a few questions of code sharing between v1 and v2

hedera-node/configuration/dev/application.properties
stoyanov-st: Agreed, but we could keep dev config so that we won't have to rework the tests?

lukelee-sl: yes, that would be fine.

lukelee-sl (dismissed):
LGTM.   tyvm

lukelee-sl (approved):

Nana-EC (approved):

hedera-node/configuration/previewnet/application.properties
```diff
@@ -7,4 +7,4 @@
 bootstrap.genesisPublicKey=c249a323c878f5b5e2daccda6d731e6fdc32f870228d1cd4fae559d947dbc36c
 contracts.chainId=297
 ledger.id=0x02
-entities.unlimitedAutoAssociationsEnabled=true
+entities.unlimitedAutoAssociationsEnabled=true
```
netopyr, line 10: This change needs to be reverted because it may cause issues. All property-files must end with an empty line.

netopyr (changes requested):
The `application.properties` file must end with an empty line.

hedera-node/configuration/previewnet/application.properties
stoyanov-st: Thank you! Great catch! Added the empty line.

lukelee-sl (approved):
LGTM

netopyr (approved):
LGTM in regards to Services code

PR #15075, opened 2024-08-23
feat: auto post-upgrade system file updates

### LINKED ISSUE #14944: Auto-update system files at upgrade boundaries
### Problem

The upgrade ZIP is uploaded to special file `0.0.150`, and then on a `PREPARE_UPGRADE` freeze transaction, extracted to the filesystem on the node at the relative location found in property [`networkAdmin.upgradeArtifactsPath`](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/NetworkAdminConfig.java#L29) (currently _data/upgrade/current_).

NMT automation then moves any files extracted here under a _data/config/_ subdirectory into  _data/config/_ relative to the node's working directory, so that after the upgrade the node actually has easy access to new sources of configuration.

But we currently make minimal use of this fact. Instead new fee schedules, throttles, are submitted as privileged `FileUpdate` transactions independent of the software upgrade.

### Solution

It would be considerably easier to automate such changes by packaging the **all** the following files in the upgrade ZIP's _data/config/_ directory:
1. _feeSchedules.json_ 
2. _throttles.json_
3. _application.properties_
4. _api-permission.properties_

After upgrade, the node can then parse these files, convert them to appropriate protobuf messages, and dispatch a synthetic `FileUpdate` for each on the first post-upgrade transaction, using `0.0.50` as the payer id with an appropriately offset consensus time preceding the first transaction.

That is, post-upgrade the contents of system files `0.0.111`, `0.0.123`, `0.0.121`, and `0.0.122` respectively will be derived from the contents of these files. (If any of 1-4 is not included in the ZIP, and hence not found in _data/config/_ post-upgrade, the node will just log a warning and skip the corresponding `FileUpdate`.)

ℹ️ It is important to note that in test environments we will commonly want to override properties at startup, independent of any council-approved changes to file `0.0.121`. As a result, we should **simultaneously** introduce a new file named _startup-override.properties_ that can be packaged in the ZIP for testing purposes, but not trigger a `FileUpdate` to `0.0.121`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #14944


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
```diff
@@ -106,13 +122,95 @@ public void doGenesisSetup(@NonNull final Dispatch dispatch) {
...
+                        upgradeFileName,
+                        path.toAbsolutePath());
+                return false;
+            }
+            try (final var fin = Files.newInputStream(path)) {
+                final Bytes bytes;
+                try {
+                    bytes = upgradeFileParser.apply(fin);
+                } catch (Exception e) {
+                    log.warn("Failed to parse upgrade file for {} at {}", upgradeFileName, path.toAbsolutePath(), e);
```
david-bakin-sl, line 199: Failure to parse seems kind of serious.  It really means that a _bad_ file was stuck into the upgrade package, an operational error that should never occur (but, people being people, might just happen).  Anyway, given that the presence of the file in the package would indicate that _some_ kind of change was desired and that at least the `*.properties` files might affect the behavior of the network in important ways, should the system continue to come up if any of these files can't be parsed?  Perhaps starting the network would cause some transactions to be processed incorrectly (wrong properties applied).  It could even be that starting the network would cause some migration to be applied when it shouldn't be.

david-bakin-sl (commented):
Only a comment because I'm not really qualified to review further semantics.

litt3 (approved):
Approved wrt the hashgraph owned file

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
tinker-michaelj: True @david-bakin-sl , but a failure here can be remediated by manually submitting a `FileUpdate` to the network post-upgrade.

So we wouldn't want to fail _hard_ here.

david-bakin-sl: ok Michael! Just wanted to make sure that someone who knew more about node operations than me had/would consider it. TY!

(But it does mean that the node upgrade recipe needs to include a scan of the logs for WARN - or at least this specific message - before going live, if that isn't already done, so that the operators know they need to do the file update.  Is it?  WARN isn't a very high log level.)

litt3 (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
tinker-michaelj: True, promoted these to `ERROR`. 💯

mxtartaglia-sl (dismissed):

litt3 (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
```diff
@@ -106,13 +129,95 @@ public void doGenesisSetup(@NonNull final Dispatch dispatch) {
...
+         * present at the given location and can be parsed with this update's parser.
+         *
+         * @param postUpgradeLoc the location of the post-upgrade files
+         * @param systemContext the system context
+         * @return {@code true} if a synthetic update was dispatched, {@code false} otherwise
+         */
+        boolean tryIfPresent(@NonNull final String postUpgradeLoc, @NonNull final SystemContext systemContext) {
+            final var path = Paths.get(postUpgradeLoc, upgradeFileName);
+            if (!Files.exists(path)) {
+                log.info(
```
Neeharika-Sompalli, line 195: Isn't this a warn or error ?

hedera-node/hedera-file-service-impl/src/main/java/com/hedera/node/app/service/file/impl/handlers/FileUpdateHandler.java
```diff
@@ -228,16 +231,21 @@ private void handleUpdateUpgradeFile(FileUpdateTransactionBody fileUpdate, Handl
...
             builder.keys(op.keys());
         } else {
             builder.keys(file.keys());
         }
-        var contentLength = op.contents().length();
-        if (contentLength > 0) {
-            if (contentLength > fileServiceConfig.maxSizeKb() * 1024L) {
+        final var contentLength = op.contents().length();
+        final var zeroLengthShouldClearTarget =
+                accountsConfig.isSuperuser(payerId) && filesConfig.isOverrideFile(fileId);
```
Neeharika-Sompalli, line 246: Probably add javadoc 😛

tinker-michaelj: Yes Neeha, will do! 😛

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
tinker-michaelj: Used `INFO` because there's no real need to replace these files at most upgrade boundaries.

PR #15079, opened 2024-08-23
feat: add exceed max airdrops validation

### LINKED ISSUE #15076: (AIRDROP - 80) exceed max airdrops in the pending state

vtronkov:
**Description**:
* Adding check if we exceeded the airdrops limits
* Adding HAPI tests covering the validation

**Related issue(s)**:

Fixes #15076

**Checklist**


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -134,6 +142,26 @@ public void validateSemantics(
...
 
+    /**
+     * This is rough estimation of the airdrop count. We take the pessimistic approach here and assume that all
+     * transfers will be created as airdrops. In reality some might be a direct crypto transfer.
+     * If we want to calculate the correct airdrop count we will need to create counters in the handle method.
+     * This will add complications to the handle method and the actual check will need to happened at the end of the
+     * method where if the limit is exceeded - the handle operations we did will be discarded.
+     * The only downside of this approach is if we reach 20m airdrops we might error out a bit early but this
+     * doesn't seem like a big issue.
+     */
```
vtronkov, line 153: thoughts?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -93,10 +96,15 @@ public void validateSemantics(
...
-            @NonNull final ReadableNftStore nftStore) {
+            @NonNull final ReadableNftStore nftStore,
+            @NonNull final ReadableAirdropStore airdropStore) {
         var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
         validateTrue(
                 op.tokenTransfers().size() <= tokensConfig.maxAllowedAirdropTransfersPerTx(), INVALID_TRANSACTION_BODY);
+        validateTrue(
+                airdropStore.sizeOfState() + roughAirdropCount(op.tokenTransfers())
+                        <= tokensConfig.maxAllowedPendingAirdrops(),
+                MAX_ENTITIES_IN_PRICE_REGIME_HAVE_BEEN_CREATED);
```
Neeharika-Sompalli, line 107: We should probably do this after we know pendingAirdrops that will be created.
Probably after `separateFungibleTransfers` and `separateNftTransfers`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
Neeharika-Sompalli: We should probably do this after we know pendingAirdrops that will be created.
Probably after each of these steps `separateFungibleTransfers` and `separateNftTransfers`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
vtronkov: That was an option Ivan suggested 😄 Okay, I will move it there 👍

Neeharika-Sompalli:
> Two of the tests are flaky because sometimes it happens that a suite runs before `TokenAirdropWithOverriddenMaxAllowedPendingAirdropsTest` and fills the store. Is there a way to clean the store when I run the test and run the test in isolation?
> 
> I can think of two options but there might be a better way:
> 
> 1. Run the tests in embedded mode(with EmbeddedHapiTest annotation) and create a new verb to access and clean the store
> 2. Test this validation with unit tests instead of HAPI.
> 
> Any ideas? @Neeharika-Sompalli @tinker-michaelj

@vtronkov  I am fine validating in unit tests. But if we mark the HapiTest as `@LeakyHapiTest` would it not solve the problem ?

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
Neeharika-Sompalli: Cool! Thanks

vtronkov:
> > Two of the tests are flaky because sometimes it happens that a suite runs before `TokenAirdropWithOverriddenMaxAllowedPendingAirdropsTest` and fills the store. Is there a way to clean the store when I run the test and run the test in isolation?
> > I can think of two options but there might be a better way:
> > 
> > 1. Run the tests in embedded mode(with EmbeddedHapiTest annotation) and create a new verb to access and clean the store
> > 2. Test this validation with unit tests instead of HAPI.
> > 
> > Any ideas? @Neeharika-Sompalli @tinker-michaelj
> 
> @vtronkov I am fine validating in unit tests. But if we mark the HapiTest as `@LeakyHapiTest` would it not solve the problem ?

We tried `@LeakyHapiTest(overrides = {"contracts.evm.version"})` but it was not working 🤷

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
vtronkov: Fixed 👍

Neeharika-Sompalli (approved):

PR #15084, opened 2024-08-23
test: Additional airdrop negative tests

### LINKED ISSUE #14912: (CLAIM_47) TokenClaimAirdrop transaction with no pending airdrops entries should fail

### LINKED ISSUE #14913: (CLAIM_48) Attempt to claim an airdrop that was claimed by the same account

### LINKED ISSUE #14914: (CLAIM_49) Attempt to claim an airdrop that was claimed by a different account

### LINKED ISSUE #14915: (CLAIM_50) Attempt to claim an airdrop by the sender account (when the receiver account is different)

### LINKED ISSUE #14916: (CLAIM_51) TokenClaimAirdrop transaction containing pending airdrops entries which do not exist in state should fail

JivkoKelchev:
**Description**:

Add claim airdrop test cases from the test plan:
CLAIM_47
CLAIM_48
CLAIM_49
CLAIM_50
CLAIM_51

**Related issue(s)**:

Fixes #14912
Fixes #14913
Fixes #14914
Fixes #14915
Fixes #14916

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
@@ -813,18 +813,16 @@ public class TokenClaimAirdropTest extends TokenAirdropBase {
 
     @HapiTest
     @DisplayName("Attempt to claim an airdrop by the sender")
-    final Stream<DynamicTest> claimWithDifferentAccounts() {
+    final Stream<DynamicTest> claimWithSenderAccount() {
         var receiver = "receiver";
-        var randomAccount = "randomAccount";
         // CLAIM_50
         return hapiTest(flattened(
                 setUpTokensAndAllReceivers(),
                 cryptoCreate(receiver),
-                cryptoCreate(randomAccount),
                 tokenAirdrop(moving(1, FUNGIBLE_TOKEN).between(OWNER, receiver)).payingWith(OWNER),
                 tokenClaimAirdrop(pendingAirdrop(OWNER, receiver, FUNGIBLE_TOKEN))
-                        .payingWith(randomAccount)
-                        .signedBy(randomAccount)
+                        .payingWith(OWNER)
+                        .signedBy(OWNER)
                         .hasKnownStatus(INVALID_SIGNATURE)));
     }
 
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
@@ -34,6 +34,7 @@ import static com.hedera.services.bdd.spec.transactions.TxnVerbs.cryptoTransfer;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.mintToken;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenAirdrop;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenAssociate;
+import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenCancelAirdrop;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenClaimAirdrop;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenCreate;
 import static com.hedera.services.bdd.spec.transactions.TxnVerbs.tokenDissociate;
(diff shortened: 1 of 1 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
```diff
@@ -112,17 +114,20 @@ public class TokenClaimAirdropTest extends TokenAirdropBase {
...
+                "tokens.airdrops.cancel.enabled", "false"));
         // create some entities with disabled airdrops
         lifecycle.doAdhoc(setUpEntitiesPreHIP904());
         // enable airdrops
         lifecycle.doAdhoc(
                 overriding("entities.unlimitedAutoAssociationsEnabled", "true"),
                 overriding("tokens.airdrops.enabled", "true"),
-                overriding("tokens.airdrops.claim.enabled", "true"));
+                overriding("tokens.airdrops.claim.enabled", "true"),
+                overriding("tokens.airdrops.cancel.enabled", "true"));
```
iwsimon, line 130: Above overrideInClass() call is all the default TokesConfig values. EntitiesConfig also has unlimitedAutoAssociationsEnabled default to true. I think we only need lifecycle.overrideInClass() for "tokens.airdrops.enabled", "true", "tokens.airdrops.claim.enabled", "true","tokens.airdrops.cancel.enabled", "true" here. 
               @tinker-michaelj is the above correct?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
```diff
@@ -765,6 +791,43 @@ final Stream<DynamicTest> hollowNoHbarFail() {
...
+        // CLAIM_50
+        return hapiTest(flattened(
+                setUpTokensAndAllReceivers(),
+                cryptoCreate(receiver),
+                cryptoCreate(randomAccount),
+                tokenAirdrop(moving(1, FUNGIBLE_TOKEN).between(OWNER, receiver)).payingWith(OWNER),
+                tokenClaimAirdrop(pendingAirdrop(OWNER, receiver, FUNGIBLE_TOKEN))
+                        .payingWith(randomAccount)
+                        .signedBy(randomAccount)
+                        .hasKnownStatus(INVALID_SIGNATURE)));
```
iwsimon, line 828: randomAccount is not the sender.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
JivkoKelchev: Here, we override airdrop related properties twice. 
1 - we disabled everything using `overrideInClass`.
This way we can create entities like if they are created pre hip904 in `setUpEntitiesPreHIP904`
For example, a hollow account with 0 auto-associations.
2 - After that we enable again everything related to the airdrops, and then start tests.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
JivkoKelchev: Fixed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
iwsimon: We only need  ```
lifecycle.overrideInClass(Map.of(
                "entities.unlimitedAutoAssociationsEnabled", "false"))
``` instead of 
                l```
ifecycle.overrideInClass(Map.of(
                "entities.unlimitedAutoAssociationsEnabled", "false",
                "tokens.airdrops.enabled", "false",
                "tokens.airdrops.claim.enabled", "false")); 
```since the other two are the default config value.

Neeharika-Sompalli (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
JivkoKelchev: But when we release the airdrops, the default values will be changed to "true". 
Also, in different environments, we can have different values. 
I don't think we should be relying on default values here.

iwsimon: I see. Thanks.

iwsimon (approved):
LGTM, thanks @JivkoKelchev

PR #15094, opened 2024-08-23
feat: prepare PlatformBuilder for protobuf Roster

### LINKED ISSUE #14708: Build a Roster object out of the existing AddressBook object in PlatfromBuilder
- use the roster from protobuf
- remove AddressBookRoster and legacy Roster

kfa-aguda:
- use the new roster from protobuf
- modify `PlatformBuilder` to accept bootstrap roster
- if bootstrap roster is not provided, build one from bootstrap AddressBook
- remove AddressBookRoster and legacy Roster
- update all references of legacy roster to addressBook

Fixes #14708


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -659,6 +683,64 @@ public PlatformComponentBuilder buildComponentBuilder() {
         return new PlatformComponentBuilder(buildingBlocks);
     }
 
+    /**
+     * Creates a new roster from the bootstrap address book.
+     *
+     * @return a new roster
+     */
+    @NonNull
+    public Roster createRoster() {
```
edward-swirldslabs, line 692: This new code seems like it would go in a utility class and having it here as public methods breaks the `Builder` pattern of the `PlatformBuilder`.  These methods on the utility class would be called in a `build()` call, maybe the `buildComponentBuilder`?  

If we're performing inversion of control, then the conversion of the address book to a roster happens completely before and outside of the `PlatformBuilder`?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookDiff.java
```diff
@@ -31,9 +30,9 @@
  * @param removedNodes           the nodes that were removed
  * @param modifiedNodes          the nodes that were modified
  */
-public record RosterDiff(
-        @NonNull UpdatedRoster newRoster,
-        boolean rosterIsIdentical,
+public record AddressBookDiff(
```
edward-swirldslabs, line 33: Neither the `RosterDiff` or the `RosterDiffGenerator` are in use in the rest of the code.  This was added under the old paradigm of doing DynamicAddressBook phase 3.  That's completely out the window in terms of design.  I think this code should be outright deleted, but I will defer to @litt3's opinion

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/builder/PlatformBuilderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test
+    void testCreateRosterFromNonEmptyAddressBook() {
+        final Address address1 = new Address(new NodeId(1), "", "", 10, null, 77, null, 88, null, null, "");
+        final Address address2 = new Address(new NodeId(2), "", "", 10, null, 77, null, 88, null, null, "");
+        addressBook.add(address1);
+        addressBook.add(address2);
+        platformBuilder.withBootstrapAddressBook(addressBook);
+        final Roster roster = platformBuilder.createRoster();
```
edward-swirldslabs, line 56: Yes, this breaks the `PlatformBuilder` builder pattern.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/roster/legacy/AddressBookRosterTests.java
```diff
@@ -1,109 +0,0 @@
...
-import com.swirlds.platform.system.address.AddressBook;
-import com.swirlds.platform.test.fixtures.addressbook.RandomAddressBookBuilder;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.util.Iterator;
-import org.junit.jupiter.api.DisplayName;
-import org.junit.jupiter.api.Test;
-
-public class AddressBookRosterTests {
```
edward-swirldslabs, line 40: Are there equivalent tests that need to be created for your own translation from address book to roster?

edward-swirldslabs (changes requested):
The issue this PR is implementing doesn't mention modifying the platform builder, just writing the translation code.  That code should go in a utility class, not the `PlatformBuilder`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
kfa-aguda: IOC is what the `PlatformBuilder.withBootstrapRoster` is for, as are other methods called from `Hedera` class so I think that bit is consistent. And in the case the platform itself wants to create the `Roster`, it can call this particular method directly. 
But you do have a point in putting it into a utility to preserve the builder pattern.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/roster/legacy/AddressBookRosterTests.java
kfa-aguda: Yes, they're in [PlatformBuilderTest.java](https://github.com/hashgraph/hedera-services/pull/15094/files/0c7b030e95d8ea246a05e7dbd9ad7fd4502890d9#diff-6320b31fa1fb8daaa8d4681c250463aef032b94754387f189338438e86369a12)
Now moved into AddressBookUtilsTest.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookDiff.java
kfa-aguda: removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -557,6 +575,10 @@ public PlatformComponentBuilder buildComponentBuilder() {
             throw new IllegalStateException("The current address book of the initial state is null.");
         }
 
+        if (bootstrapRoster == null) {
+            bootstrapRoster = createRoster(boostrapAddressBook);
```
edward-swirldslabs, line 579: there are code coverage warnings on 404 and 579 for lack of coverage by tests.  Odd that 579 is untested.  Does that mean the null check is not matching in existing tests?

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/event/tipset/TipsetEventCreatorTests.java
```diff
@@ -1025,7 +1025,7 @@ void noStaleEventsAtCreationTimeTest(final boolean useBirthRoundForAncient) {
     /**
      * Checks that birth round on events is being set if the setting for using birth round is set.
      * <p>
-     * FUTURE WORK: Update this test to use RosterDiff instead of EventWindow
+     * FUTURE WORK: Update this test to use AddressBookDiff instead of EventWindow
```
edward-swirldslabs, line 1028: this comment change should probably be reverted or updated to something different.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/AddressBookNetworkUtilsTests.java
```diff
@@ -76,4 +83,53 @@ void ErrorOnInvalidLocalAddress() {
...
+        final PlatformBuilder platformBuilder = PlatformBuilder.create(
+                "name",
+                "swirldName",
+                new BasicSoftwareVersion(1),
+                () -> null,
+                (inputStream, path) -> null,
+                new NodeId(0));
+        final AddressBook addressBook = new AddressBook();
+        platformBuilder.withBootstrapAddressBook(addressBook);
+        final Roster roster = AddressBookUtils.createRoster(addressBook);
```
edward-swirldslabs, line 130: Should this be throwing exceptions?   Does the `createRoster` also need to validate that the roster is valid?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
```diff
@@ -284,4 +294,84 @@ public static void verifyReconnectAddressBooks(
...
+    /**
+     * Given a host and port, creates a {@link ServiceEndpoint} object with either an IP address or domain name
+     * depending on the given host.
+     *
+     * @param host the host
+     * @param port the port
+     * @return the {@link ServiceEndpoint} object
+     */
+    public static ServiceEndpoint endpointFor(@NonNull final String host, final int port) {
+        final var builder = ServiceEndpoint.newBuilder().port(port);
```
edward-swirldslabs, line 307: code coverage is complaining on several lines in this file.

kfa-aguda:
> The issue this PR is implementing doesn't mention modifying the platform builder, just writing the translation code. That code should go in a utility class, not the `PlatformBuilder`.

I wrote the description of that issue, and its title specifically highlights that it should modify the platform builder. I edited the description a little to make it more concise, but its goal remains the same.

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/util/AddressBookNetworkUtilsTests.java
kfa-aguda: I think there's value in separating the creator and validator as that makes it more testable. The core code will create, then validate.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -386,6 +390,20 @@ public PlatformBuilder withBootstrapAddressBook(@NonNull final AddressBook boots
...
 
+    /**
+     * Provide the roster to use for bootstrapping the system. If not provided then the roster is created from the
+     * bootstrap address book.
+     *
+     * @param bootstrapRoster the roster to use for bootstrapping
+     * @return this
+     */
+    @NonNull
+    public PlatformBuilder withBootstrapRoster(@NonNull final Roster bootstrapRoster) {
```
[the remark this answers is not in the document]
kfa-aguda: That work will happen when we are ready for creating various forms of Rosters. We will have to work with devops to get to that point.
All Roster's are 'bootstrap' roster :-) but happy to rename if you feel strongly about it. I'm indifferent.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
kfa-aguda: resolved.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -557,6 +575,10 @@ public PlatformComponentBuilder buildComponentBuilder() {
             throw new IllegalStateException("The current address book of the initial state is null.");
         }
 
+        if (roster == null) {
+            roster = createRoster(boostrapAddressBook);
```
anthony-swirldslabs, line 579: The `boostrapAddressBook` is never used after this line. This means that the AddressBook instance that this variable describes is not useful. It's the `addressBook` defined at line 572 above that is used below. We need a Roster instance of that `addressBook` here.

PR #15095, opened 2024-08-23
feat: Support multiple rounds in a block

Neeharika-Sompalli:
- Add support for multiple rounds in a block
- Adds a Queue for `BlockStreamManager` to store blocks that are waiting for the signature on blockHash.


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/BlockStreamAccess.java
```diff
@@ -202,7 +213,7 @@ private Block blockFrom(@NonNull final Path path) {
                 return Block.PROTOBUF.parse(Bytes.wrap(Files.readAllBytes(path)));
             }
         } catch (IOException | ParseException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException("Failed reading block @ " + path, e);
```
thomas-swirlds-labs, line 216: nit: may be worth while now to normalize and sanitize the input path we are reading from

thomas-swirlds-labs (approved):

Issue #14969, opened 2024-08-20
Introduce utility to get active Roster for a round or a state


anthony-swirldslabs:

anthony-swirldslabs:
Also fix https://github.com/hashgraph/hedera-services/pull/15096#discussion_r1735343278

PR #15096, opened 2024-08-23
feat: introduce RosterRetriever

### LINKED ISSUE #14995: Introduce utility to get Roster from state or fall back to getting AB and converting it to Roster

anthony-swirldslabs:
**Description**:
Introducing the `RosterRetriever` utility class that allows one to retrieve an active Roster for a given State.

This is a prerequisite to enable refactoring of the Platform code to replace all usages of the AddressBook with using the Roster instead. Places in code that currently fetch the AddressBook directly from the PlatformState will instead call this utility to retrieve the Roster.

**Related issue(s)**:

Fixes #14995

**Notes for reviewer**:
Unit tests pass.

**Checklist**


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -0,0 +1,180 @@
...
+import java.security.cert.CertificateEncodingException;
+import java.util.Comparator;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A utility class to help retrieve a Roster instance from the state.
+ */
+public final class RosterRetriever {
```
lpetrovic05, line 43: To me these methods do not seem like they should be static. I do not know what the API for accessing the state is supposed to look like, so I might ask some stupid questions. Are these methods supposed to work on any `State` implementation? If yes, then why not make them part of the state interface?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -0,0 +1,180 @@
...
+                                                                address.getHostnameExternal(),
+                                                                address.getPortExternal()),
+                                                        Pair.of(
+                                                                address.getHostnameInternal(),
+                                                                address.getPortInternal()))
+                                                .stream()
+                                                .filter(pair -> pair.left() != null
+                                                        && !pair.left().isBlank()
+                                                        && pair.right() != 0)
+                                                .distinct()
```
kfa-aguda, line 145: do we really need a distinct call here? Seems like an overkill.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -0,0 +1,180 @@
...
+     * and converting it to a Roster.
+     *
+     * @return an active Roster for the round of the state, or a Roster that represents the current AddressBook in PlatformState
+     */
+    @NonNull
+    public static Roster retrieve(@NonNull final State state) {
+        final Bytes activeRosterHash = getActiveRosterHash(state);
+        if (activeRosterHash != null) {
+            final ReadableKVState<ProtoBytes, Roster> rosterMap =
+                    state.getReadableStates("RosterServiceImpl").get("ROSTER");
```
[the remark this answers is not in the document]
anthony-swirldslabs: I considered that originally. However, this would introduce a dependency of the platform core module on an app module. I don't know whether a circular dependency will be created this way, but even if it won't, it didn't feel right to make the platform core depend on the app.

Do you think it's a good idea?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -0,0 +1,180 @@
...
+     * or the current round of the state predates the implementation of the Roster.
+     *
+     * @param state a state
+     * @return a Bytes object with the roster hash, or null
+     */
+    @Nullable
+    public static Bytes getActiveRosterHash(@NonNull final State state) {
+        final long round = getRound(state);
+        final ReadableSingletonState<RosterState> rosterState =
+                state.getReadableStates("RosterServiceImpl").getSingleton("ROSTER_STATE");
```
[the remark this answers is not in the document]
anthony-swirldslabs: Please see above as well.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
anthony-swirldslabs: Yes, I overheard somewhere (I don't recall the reference) that the existing AddressBook may in fact list the same address for both the internal and external endpoints, so I decided to remove any duplicates just in case. This doesn't seem to have a potential to create any issues, does it?

Also, the expectation is that the platform will migrate over to Rosters after which (a couple of releases or so?) this legacy code path that converts an AddressBook to a Roster could be removed altogether.

Let me know what you think.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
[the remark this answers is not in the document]
kfa-aguda: It should be fine to move it from the schema and service into a class in some module that's common to both.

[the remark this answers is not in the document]
anthony-swirldslabs: Happy to do that. Could you please suggest a place (a specific class, or at least a java package/module)?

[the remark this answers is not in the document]
kfa-aguda: swirlds-common is a module common to both, so i think that's a good place (except someone else has a strong reason for not putting it there)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
kfa-aguda: Some considerations i have here are:
1. The API chaining here seems to be handful. Is it worth breaking this method down for clarity?
2. There are max 2 endpoints per address. If true, the stream, filter, distinct, and map calls are operating on two elements? Is that correct?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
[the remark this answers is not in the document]
anthony-swirldslabs: Cool. Done.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
anthony-swirldslabs: This isn't a two-liner to be sure, but it looks manageable. I'd like to re-iterate that this method will be removed very soon - as soon as we replace the AddressBook with the Roster, which should be completed with the TSS-Roster project as a whole. So I'd prefer to keep it as is, in as little number of methods as possible.

And yes, that stream operates on a maximum of 2 elements, one of which is likely a duplicate of another.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/InvalidAddressBookException.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.platform.roster;
+
+/**
+ * An exception thrown by the RosterValidator when a given Roster is invalid.
+ */
+public class InvalidAddressBookException extends RuntimeException {
```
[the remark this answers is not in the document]
anthony-swirldslabs: No. This exception is thrown specifically when it's impossible to convert an AddressBook instance into a Roster. The Roster object simply doesn't exist at that moment, so it cannot be invalid. It's the AddressBook that is.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
kfa-aguda: ok. If it's going to go away soon, maybe that's fine.

kfa-aguda (dismissed):

kfa-aguda (approved):

kfa-aguda (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/RosterServiceImpl.java
```diff
@@ -29,13 +30,10 @@
...
  */
 public class RosterServiceImpl implements Service {
-    /** The name of this service */
-    public static final String NAME = "RosterService";
-
     @NonNull
     @Override
     public String getServiceName() {
-        return NAME;
+        return RosterStateId.NAME;
```
Neeharika-Sompalli, line 36: nit: Do we need a new class to define these Constants. Are we expecting to add more information to that class ?

anthony-swirldslabs: This was done to address a comment above in this PR. Please see the thread at: https://github.com/hashgraph/hedera-services/pull/15096#discussion_r1731856354

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/InvalidAddressBookException.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.platform.roster;
+
+/**
+ * An exception thrown by the RosterValidator when a given Roster is invalid.
```
Neeharika-Sompalli, line 20: Typo: 
```suggestion
 * An exception thrown by the RosterValidator when a given AddressBook  is invalid.
```

Neeharika-Sompalli (approved):
LGTM except one typo. Thanks @anthony-swirldslabs

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/InvalidAddressBookException.java
anthony-swirldslabs: Nice catch! Thanks! I'll fix this typo in a follow-up PR in the next few days.

anthony-swirldslabs: I'm tracking it here: https://github.com/hashgraph/hedera-services/issues/14969#issuecomment-2316340042

Issue #15036, opened 2024-08-21
IndexOutOfBoundsException when flushing an empty virtual map to MerkleDb


artemananiev:
Stack trace:

```
Caused by: java.lang.IndexOutOfBoundsException: Min valid index -1 must be non-negative
	at com.swirlds.merkledb.collections.AbstractLongList.updateValidRange(AbstractLongList.java:465) ~[swirlds-merkledb-0.52.2.jar:?]
	at com.swirlds.merkledb.files.MemoryIndexDiskKeyValueStore.startWriting(MemoryIndexDiskKeyValueStore.java:121) ~[swirlds-merkledb-0.52.2.jar:?]
	at com.swirlds.merkledb.MerkleDbDataSource.writeLeavesToPathToKeyValue(MerkleDbDataSource.java:1145) ~[swirlds-merkledb-0.52.2.jar:?]
	at com.swirlds.merkledb.MerkleDbDataSource.saveRecords(MerkleDbDataSource.java:523) ~[swirlds-merkledb-0.52.2.jar:?]
	at com.swirlds.virtualmap.datasource.VirtualDataSource.saveRecords(VirtualDataSource.java:91) ~[swirlds-virtualmap-0.52.2.jar:?]
	at com.swirlds.virtualmap.internal.merkle.VirtualRootNode.flush(VirtualRootNode.java:1182) ~[swirlds-virtualmap-0.52.2.jar:?]
	at com.swirlds.virtualmap.internal.merkle.VirtualRootNode.setupWithOriginalNode(VirtualRootNode.java:1429) ~[swirlds-virtualmap-0.52.2.jar:?]
	at com.swirlds.common.merkle.synchronization.task.LearnerPushTask.handleCustomRootInitialLesson(LearnerPushTask.java:122) ~[swirlds-common-0.52.2.jar:?]
	at com.swirlds.common.merkle.synchronization.task.LearnerPushTask.extractNodeFromLesson(LearnerPushTask.java:146) ~[swirlds-common-0.52.2.jar:?]
	at com.swirlds.common.merkle.synchronization.task.LearnerPushTask.run(LearnerPushTask.java:247) ~[swirlds-common-0.52.2.jar:?]
	at com.swirlds.common.threading.pool.StandardWorkGroup.lambda$execute$1(StandardWorkGroup.java:153) ~[swirlds-common-0.52.2.jar:?]
[...]

PR #15097, opened 2024-08-24
fix: 15036: IndexOutOfBoundsException when flushing an empty virtual map to MerkleDb

artemananiev:
Fix summary:

* A new field is introduced to `AbstractLongList` and its subclasses: max valid index. Previously there was only a field for min valid index, and list size was used as max valid index
* Both min and max valid indices can be -1, which indicates that the long list is empty
* Before a long list can be used, `updateValidKeyRange()` method must be called. All updates to the list outside the current range will result in an exception

Testing:

* A new unit test is provided. It failed before the fix, and passes with the fix
* Existing unit tests are updated to call `updateValidKeyRange()` when needed

Fixes: https://github.com/hashgraph/hedera-services/issues/15036

### DIFF
--- a/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
+++ b/platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
@@ -285,6 +285,9 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
             final boolean hashIndexEmpty = pathToDiskLocationInternalNodes.size() == 0;
             final LoadedDataCallback<VirtualHashRecord> hashRecordLoadedCallback;
             if (hashIndexEmpty) {
+                if (validLeafPathRange.getMaxValidKey() >= 0) {
+                    pathToDiskLocationInternalNodes.updateValidRange(0, validLeafPathRange.getMaxValidKey());
+                }
                 hashRecordLoadedCallback = (dataLocation, hashRecord) ->
                         pathToDiskLocationInternalNodes.put(hashRecord.path(), dataLocation);
             } else {
@@ -338,6 +341,10 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
         final LoadedDataCallback<VirtualLeafRecord<K, V>> leafRecordLoadedCallback;
         final boolean needRestorePathToDiskLocationLeafNodes = pathToDiskLocationLeafNodes.size() == 0;
         if (needRestorePathToDiskLocationLeafNodes) {
+            if (validLeafPathRange.getMaxValidKey() >= 0) {
+                pathToDiskLocationLeafNodes.updateValidRange(
+                        validLeafPathRange.getMinValidKey(), validLeafPathRange.getMaxValidKey());
+            }
             leafRecordLoadedCallback = (dataLocation, leafRecord) -> {
                 final long path = leafRecord.getPath();
                 pathToDiskLocationLeafNodes.put(path, dataLocation);
@@ -603,7 +610,8 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
     }
 
     /**
-     * Load a leaf record by path
+     * Load a leaf record by path. This method returns {@code null}, if the path is outside the
+     * valid path range.
      *
      * @param path the path for the leaf we are loading
      * @return loaded record or null if not found
@@ -611,9 +619,12 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
      */
     @Override
     public VirtualLeafRecord<K, V> loadLeafRecord(final long path) throws IOException {
+        if (path < 0) {
+            throw new IllegalArgumentException("Path (" + path + ") is not valid");
+        }
         final KeyRange leafPathRange = validLeafPathRange;
         if (!leafPathRange.withinRange(path)) {
-            throw new IllegalArgumentException("path (" + path + ") is not valid; must be in range " + leafPathRange);
+            return null;
         }
         statisticsUpdater.countLeafReads();
         return pathToKeyValue.get(path);
@@ -660,7 +671,7 @@ public final class MerkleDbDataSource<K extends VirtualKey, V extends VirtualVal
     @Override
     public Hash loadHash(final long path) throws IOException {
         if (path < 0) {
-            throw new IllegalArgumentException("path is less than 0");
+            throw new IllegalArgumentException("Path (" + path + ") is not valid");
         }
 
         // It is possible that the caller will ask for an internal node that the database doesn't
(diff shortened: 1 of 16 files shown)


platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -603,17 +610,21 @@ public VirtualLeafRecord<K, V> loadLeafRecord(final K key) throws IOException {
...
      */
     @Override
     public VirtualLeafRecord<K, V> loadLeafRecord(final long path) throws IOException {
+        if (path < 0) {
+            throw new IllegalArgumentException("Path (" + path + ") is not valid");
+        }
         final KeyRange leafPathRange = validLeafPathRange;
         if (!leafPathRange.withinRange(path)) {
-            throw new IllegalArgumentException("path (" + path + ") is not valid; must be in range " + leafPathRange);
+            return null;
```
artemananiev, line 627: This behavior is changed for consistency with related methods like `loadHash()`, which return null (rather than throw an exception), if a path is outside the current path range

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -1067,13 +1078,22 @@ private void runWithSnapshotExecutor(
...
     private void writeHashes(final long maxValidPath, final Stream<VirtualHashRecord> dirtyHashes) throws IOException {
-        if ((dirtyHashes == null) || (maxValidPath <= 0)) {
+        if (hasDiskStoreForHashes) {
+            if (maxValidPath < 0) {
+                // Empty store
+                hashStoreDisk.updateValidKeyRange(-1, -1);
+            } else {
+                hashStoreDisk.updateValidKeyRange(0, maxValidPath);
+            }
+        }
```
artemananiev, line 1088: This is another problem, although related to the bug. Before this change, if `maxValidPath` was -1 (that is, if the virtual map becomes empty), `hashStoreDisk` and the corresponding index were not updated at all. It led to a few memory chunks to remain unused in the index. Not a memory leak, since these chunks would be reused once the map grows, but still not nice

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/MemoryIndexDiskKeyValueStore.java
```diff
@@ -65,15 +64,6 @@ public class MemoryIndexDiskKeyValueStore<D> implements AutoCloseable, Snapshota
...
     private final String storeName;
 
-    /** The minimum key that is valid for this store. Set in {@link MemoryIndexDiskKeyValueStore#startWriting} to
-     * be reused in {@link MemoryIndexDiskKeyValueStore#endWriting()}
-     */
-    private final AtomicLong minValidKey;
-    /** The maximum key that is valid for this store.  Set in {@link MemoryIndexDiskKeyValueStore#startWriting} to
-     * be reused in {@link MemoryIndexDiskKeyValueStore#endWriting()}
-     */
-    private final AtomicLong maxValidKey;
```
artemananiev, line 75: These fields are not needed. The corresponding values can be read from the index.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -100,11 +100,6 @@ public class HalfDiskHashMap<K extends VirtualKey>
     /** DataFileCollection manages the files storing the buckets on disk */
     private final DataFileCollection<Bucket<K>> fileCollection;
 
-    /**
-     * This is the number of buckets needed to store mapSize entries if we ere only LOADING_FACTOR
-     * percent full
-     */
-    private final int minimumBuckets;
```
artemananiev, line 107: This field is redundant, not used anywhere

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -603,17 +610,21 @@ public VirtualLeafRecord<K, V> loadLeafRecord(final K key) throws IOException {
...
     /**
-     * Load a leaf record by path
+     * Load a leaf record by path. This method returns {@code null}, if the path is outside the
+     * valid path range.
      *
      * @param path the path for the leaf we are loading
      * @return loaded record or null if not found
      * @throws IOException If there was a problem reading record from db
      */
     @Override
```
imalygin, line 620: This method needs `@Nullable` annotation, according to the description

artemananiev: OK, I'm adding `@Nullable` to `loadLeafRecord()`, `loadHash()`, and `findKey()` methods - all of them may return null values

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
```diff
@@ -660,7 +671,7 @@ public long findKey(final K key) throws IOException {
...
     public Hash loadHash(final long path) throws IOException {
         if (path < 0) {
-            throw new IllegalArgumentException("path is less than 0");
+            throw new IllegalArgumentException("Path (" + path + ") is not valid");
```
anthony-swirldslabs, line 674: Minor: it's a bit confusing when one can call `updateValidRange(-1, -1)`, supposedly declaring `-1` valid, and then still throw an exception on such a value.

I'm not suggesting to change this because `-1` path truly looks invalid, and so the logic in the loadHash() method here seems correct. But I'm just pointing out a confusing part of the current implementation which may be difficult to grasp.

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
```diff
@@ -627,8 +644,8 @@ private void checkValue(final long value) {
     /** {@inheritDoc} */
     @Override
     public <T extends Throwable> void forEach(final LongAction<T> action) throws InterruptedException, T {
-        final long max = size();
-        for (long i = minValidIndex.get(); i < max; i++) {
+        final long max = maxValidIndex.get();
+        for (long i = minValidIndex.get(); i <= max; i++) {
             final long value = get(i);
```
anthony-swirldslabs, line 649: BTW, this will spin the loop once in case `min == max == -1`. Not sure if it's intended.

anthony-swirldslabs (approved):

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/MerkleDbDataSource.java
artemananiev: I see your point... `updateValidRange(-1, -1)` actually means "no valid paths" rather than "-1 is a valid path". If paths were Longs, not longs, `null`s would do the trick, but there would be different issue.

On the other hand, methods like this are a part of `VirtualDataSource` interface. At that level, path -1 is invalid by any means, no doubts about that. `updateValidRange()` only exists at MerkleDb level, it's implementation details how long list index classes store their range. So I don't see a big problem here.

artemananiev: Let me add a short comment to `VirtualDataSource.loadLeafRecord(long)` JavaDoc:

```
If the path is outside the current data source's leaf path range, this method returns {@code null}.
```

A similar note is in `loadHash()`

platform-sdk/swirlds-jasperdb/src/main/java/com/swirlds/merkledb/collections/AbstractLongList.java
artemananiev: Indeed. If called for an empty list, this method would throw an exception. I added a check if the list is empty.

anthony-swirldslabs (approved):

imalygin (approved):

PR #15108, opened 2024-08-26
fix: RosterSchema should provide default roster for existing states from previous versions AND genesis.

kfa-aguda:
RosterSchema only provides default for Genesis state. That's a bug, as it must also provide default for existing state created in previous versions. This PR fixes that bug.


anthony-swirldslabs (approved):

tinker-michaelj (approved):
LGTM, ty @kfa-aguda !

PR #15113, opened 2024-08-26
fix: throw correct exception when token claim max amount is out of bounds, add hapi tests

### LINKED ISSUE #15014: (CLAIM_69) [Single Token] Receiver has a pending airdrop then receiver gets deleted and on the same alias another account is created

povolev15:
fix: impl when Max amount is out of boundary with throw the right exeption. Refactor the code and remove the update from the writableairdropstore.  In addition implemented an additional three hapi test

-(CLAIM_69) [Single Token] Receiver has a pending airdrop then receiver gets deleted and on the same alias another account is created
-(AIRDROP - 48) Given an EOA sender airdrops maxLong of a token when the sender airdrops the same token again to the same receiver then the tokenAirdropTransaction should fail.
-(AIRDROP - 21) Given EOA sender has a balance of 1 FT A and signs multiple tokenAirdropTransaction for 1 FT A to the same receiver then this should result in a single pending airdrop with the amount added up.

**Related issue(s)**:
Fixes #15014 
#15068 
#15069

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenAirdropHandlerTest.java
+++ b/hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/handlers/TokenAirdropHandlerTest.java
@@ -93,7 +93,6 @@ class TokenAirdropHandlerTest extends CryptoTransferHandlerTestBase {
     @Mock
     private FeeCalculator feeCalculator;
 
-
     @Mock
     private StoreMetricsService storeMetricsService;
 


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
```diff
@@ -1100,6 +1101,42 @@ final Stream<DynamicTest> duplicateAirdropsFail() {
...
+                        getAccountBalance(ALIAS).hasTokenBalance(FUNGIBLE_TOKEN, 0),
+                        // delete the account
+                        cryptoDelete(ALIAS),
+                        // create new account with the same key
+                        cryptoTransfer(tinyBarsFromAccountToAlias(OWNER, ALIAS, 1)),
+                        withOpContext((spec, opLog) -> updateSpecFor(spec, ALIAS)),
+                        // try to claim
+                        tokenClaimAirdrop(pendingAirdrop(OWNER, ALIAS, FUNGIBLE_TOKEN))
+                                .signedBy(ALIAS, DEFAULT_PAYER)
+                                .hasKnownStatus(INVALID_PENDING_AIRDROP_ID)));
```
iwsimon, line 1137: CLAIM_69 says should fail with INVALID_SIGNATURE. Then CLAIM_36 is the same scenario, expect success. Which one is correct? @Neeharika-Sompalli @anastasiya-kovaliova

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/TokenAirdropHandler.java
```diff
@@ -570,4 +571,39 @@ private int countExistingPendingAirdrops(
...
+        requireNonNull(airdropId);
+        requireNonNull(accountAirdrop);
+        requireNonNull(airdropState);
+
+        if (airdropId.hasFungibleTokenType()) {
+            final var existingAirdrop = requireNonNull(airdropState.getForModify(airdropId));
+            final var existingValue = existingAirdrop.pendingAirdropValue();
+            long newValue;
+            try {
+                newValue = Math.addExact(
```
vtronkov, line 595: Nice, I didn't know we had this method implemented 🔝

vtronkov, line 595: 🔝

vtronkov (approved):

kimbor (approved):
lgtm!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenClaimAirdropTest.java
Neeharika-Sompalli: yes this makes sense, because the AccountID of receiver will be different.

iwsimon: Then we should update the test plan on CLAIM_69 and CLAIM_36. @anastasiya-kovaliova

Issue #15118, opened 2024-08-26
Fix finalizing hollow payer with `unlimitedAutoAssociationsEnabled=true`


Neeharika-Sompalli:
When a hollow account pays for the `CryptoTransfer` transaction that does an `autoAssociation` to itself with `unlimitedAutoAssociationsEnabled=true`, the changes are not committed correctly, and causes a `FAIL_INVALID` record.

PR #15121, opened 2024-08-27
fix: fix hollow payer paying for transaction with unlimitedAssociationsEnabled

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/15118

When hollow payer pays for CryptoTransfer that does an autoAssociation to itself with unlimitedAssociationsEnabled = true the changes to state are not committed correctly. This PR fixes that case that was causing a FAIL_INVALID record.
Added a HapiTest validating the behavior


thomas-swirlds-labs (approved):
LGTM. Thanks @Neeharika-Sompalli 💯

ibankov (approved):

iwsimon (approved):
LGTM, thanks @Neeharika-Sompalli

PR #15136, opened 2024-08-27
feat: HIP-1028 Implement Create and Update calls with metadata

### LINKED ISSUE #14591: Define and implement HTS function call for TokenCreatе

stoyanov-st:
**Description**:

This PR aims to add the implementation for `CreateToken` and `UpdateToken` system contract calls with the metadata field described in [HIP-1028](https://hips.hedera.com/hip/hip-1028) included.
Also provided E2E tests for validation of the newly added flows.
The feature is behind a feature flag `contracts.systemContract.metadataKeyAndFieldSupport.enabled` that is only enable on dev and previewnet envs.

**Related issue(s)**:

Fixes #14591 #15126 #14622 #15237 

**Notes for reviewer**:
! In order to completely support `getInfo` function response types we need to have the System Contract versioning implemented so that we can expose new functions. Thus some of the calls for getInfo are commented for the current state.


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
```diff
@@ -187,7 +187,7 @@ private boolean isTokenCreation(MessageFrame frame) {
             return false;
         }
         var selector = frame.getInputData().slice(0, 4).toArray();
-        return CREATE_FUNCTIONS.stream().anyMatch(s -> Arrays.equals(s.selector(), selector));
+        return decoderMap.keySet().stream().anyMatch(s -> Arrays.equals(s.selector(), selector));
```
lukelee-sl, line 190: `decoderMap` is a little hard to understand.  Can we come up with a different name?  Suggestion: `createFunctionSelectorsMap`

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateDecoder.java
```diff
@@ -122,6 +125,32 @@ public TransactionBody decodeCreateFungibleTokenV3(
...
+    /**
+     * Decodes a call to {@link CreateTranslator#CREATE_FUNGIBLE_TOKEN_WITH_METADATA} into a synthetic {@link TransactionBody}.
+     *
+     * @param encoded the encoded call
+     * @param senderId the sender account ID
+     * @param nativeOperations the native operations
+     * @param addressIdConverter the address ID converter
+     * @return the synthetic transaction body
+     */
+    public TransactionBody decodeCreateFungibleTokenWithMetadata(
```
lukelee-sl, line 137: this class has a lot of static methods in this class now.  Might it be possible to have fewer by checking the number of tuples in the `HederaToken` tuple (i.e. tuple.size())?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateTranslator.java
```diff
@@ -180,36 +213,10 @@ public ClassicCreatesCall callFrom(@NonNull HtsCallAttempt attempt) {
...
-            return decoder.decodeCreateNonFungibleWithCustomFeesV1(
-                    inputBytes, senderId, nativeOperations, addressIdConverter);
-        } else if (attempt.isSelector(CREATE_NON_FUNGIBLE_TOKEN_WITH_CUSTOM_FEES_V2)) {
-            return decoder.decodeCreateNonFungibleWithCustomFeesV2(
-                    inputBytes, senderId, nativeOperations, addressIdConverter);
-        } else {
-            return decoder.decodeCreateNonFungibleWithCustomFeesV3(
-                    inputBytes, senderId, nativeOperations, addressIdConverter);
-        }
+        return decoderMap.entrySet().stream()
```
lukelee-sl, line 216: way way nicer!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyTranslator.java
```diff
@@ -76,6 +81,7 @@ private Key getTokenKey(Token token, int keyType) throws InvalidTransactionExcep
             case 16 -> token.supplyKey();
             case 32 -> token.feeScheduleKey();
             case 64 -> token.pauseKey();
+            case 128 -> metadataSupport ? token.metadataKey() : null;
```
lukelee-sl, line 84: is returning null the right thing here.  if the config is off should we not signal some sort of error?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java
```diff
@@ -110,7 +111,7 @@ private static boolean isKnownImmutable(@Nullable final Token token) {
     public @Nullable TransactionBody decodeTokenUpdateV1(@NonNull final HtsCallAttempt attempt) {
         final var call = UpdateTranslator.TOKEN_UPDATE_INFO_FUNCTION_V1.decodeCall(
                 attempt.input().toArrayUnsafe());
-        return decodeTokenUpdate(call, attempt.addressIdConverter());
+        return decodeTokenUpdate(call, attempt.addressIdConverter(), false);
```
lukelee-sl, line 114: David will likely ask for an enum instead of using a boolean :-).  Is it possible to use an overloaded version of `decodeTokenUpdate` instead of adding these versions passing a `false`?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateTranslator.java
```diff
@@ -83,8 +91,10 @@ private TransactionBody nominalBodyFor(@NonNull final HtsCallAttempt attempt) {
             return decoder.decodeTokenUpdateV1(attempt);
```
lukelee-sl, line 91: It seems using the streams may make this nicer to look at (see CreateTranslator)?

lukelee-sl (commented):
LG overall.  Just a few suggestions.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/tokenkey/TokenKeyTranslator.java
stoyanov-st: As we want to keep the same behaviour of ignoring this case when the flag is disabled, returning null complies with the `default` value of the switch and the token check before it.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateTranslator.java
stoyanov-st: Thanks!

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateTranslator.java
stoyanov-st: Added the stream version for update as well

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java
stoyanov-st: Did a slight rework of the decoding. Please check again and DM me if something seems off

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateDecoder.java
stoyanov-st: I removed the static modifiers as I don't see why we need then in first place.
We have separate methods for metadata calls so that we have completely separate flows (with/without metadata).

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/ReturnTypes.java
```diff
@@ -65,7 +65,7 @@ private ReturnTypes() {
             + ")[],";
     protected static final String EXPIRY_FIELDS =
             // second, autoRenewAccount, autoRenewPeriod
-            "(uint32,address,uint32)";
+            "(int64,address,int64)";
```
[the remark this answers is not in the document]
stoyanov-st: This is the way it is defined in the solidity interface. I guess we haven't used the previous versions to compile for e2e tests.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/processors/CustomMessageCallProcessor.java
stoyanov-st: Renamed

lukelee-sl (dismissed):
LGTM.  Maybe we need more unit test coverage for `UpdateDecoder`?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/ReturnTypes.java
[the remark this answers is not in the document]
Nana-EC: @stoyanov-st and @lukelee-sl is this effectively a version change. Won't this change the return type?

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/ContractsConfig.java
```diff
@@ -81,6 +81,9 @@ public record ContractsConfig(
         @ConfigProperty(value = "systemContract.accountService.isAuthorizedRawEnabled", defaultValue = "true")
                 @NetworkProperty
                 boolean systemContractAccountServiceIsAuthorizedRawEnabled,
+        @ConfigProperty(value = "systemContract.metadataKeyAndFieldSupport.enabled", defaultValue = "false")
```
Nana-EC, line 84: Please update the PR description to note that we are enabling this in localnet and previewnet only

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateDecoder.java
```diff
@@ -53,6 +55,7 @@ public class CreateDecoder {
     private static final int FRACTIONAL_FEE = 4;
     private static final int NFT_FIXED_FEE = 1;
     private static final int NFT_ROYALTY_FEE = 2;
+    private static final int METADATA = 9;
```
Nana-EC, line 58: nit: we should give these more descriptive names. 
In use I wouldn't know what this referred to except for coming back and reading the comment a few lines up.
Maybe
```suggestion
    private static final int TOKEN_METADATA_TUPLE_INDEX = 9;
```
or
```suggestion
    private static final int TOKEN_METADATA_INDEX = 9;
```
or something better

Nana-EC (commented):
Nice work.
Some suggestions and questions

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateDecoder.java
stoyanov-st: We have a comment explaining the constants here and imo it makes it more readable when you use it -> `call.get(METADATA);` to `call.get(TOKEN_METADATA_INDEX);`

stoyanov-st:
> LGTM. Maybe we need more unit test coverage for `UpdateDecoder`?

We can add more coverage in the future when we include the versioning

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/ReturnTypes.java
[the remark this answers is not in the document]
lukelee-sl: Good point, @stoyanov-st, will not existing contracts break because of this change?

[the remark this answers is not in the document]
stoyanov-st: This is curious as the expiry fields have been declared as `int64` since 2022. I haven't seen it represented in the response signature which is even more curious. I'll try some tests with the old version to see if it is a problem but I'm not sure how would we support both return types with the same function signature atm.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java
```diff
@@ -200,16 +218,28 @@ public TransactionBody decodeTokenUpdateExpiryV2(@NonNull final HtsCallAttempt a
...
 
-        try {
-            return bodyWith(tokenKeys, txnBodyBuilder);
-        } catch (IllegalArgumentException ignore) {
-            return null;
+    public TokenUpdateTransactionBody.Builder decodeUpdateWithMeta(
+            @NonNull final Tuple call, @NonNull final AddressIdConverter addressIdConverter) {
+        final var tokenUpdateTransactionBody = decodeTokenUpdate(call, addressIdConverter);
+        final var hederaToken = (Tuple) call.get(HEDERA_TOKEN);
+        final Bytes tokenMetadata = hederaToken.size() > 9 ? Bytes.wrap((byte[]) hederaToken.get(9)) : null;
```
lukelee-sl, line 229: maybe define a constant than using 9?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoder.java
```diff
@@ -200,16 +218,28 @@ public TransactionBody decodeTokenUpdateExpiryV2(@NonNull final HtsCallAttempt a
...
+        if (tokenMetadata != null && tokenMetadata.length() > 0) {
+            tokenUpdateTransactionBody.metadata(tokenMetadata);
         }
+        final List<TokenKeyWrapper> tokenKeys = decodeTokenKeys(hederaToken.get(7), addressIdConverter);
```
lukelee-sl, line 233: and 7 here

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/update/UpdateDecoderFunction.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.update;
+
+import com.hedera.hapi.node.transaction.TransactionBody;
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.HtsCallAttempt;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+@FunctionalInterface
+public interface UpdateDecoderFunction {
```
lukelee-sl, line 24: Maybe some java doc to define what this used for?

lukelee-sl (approved):
Generally looks good.  A few not important comments.  You can choose to fix in another pr if you want.  Thank you @stoyanov-st!

stoyanov-st:
> Generally looks good. A few not important comments. You can choose to fix in another pr if you want. Thank you @stoyanov-st!

Thank you!
Considering the timeframe of the PR, I think it would be best to get back to this at a later stage.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/create/CreateDecoder.java
Nana-EC: nit: I don' think the comment is descriptive enough if we're relying on that.
If doesn't convey the logic or assumption of the tuple and index relevance. There's a lot of presume knowledge

Nana-EC (approved):
Majority of my comments are address.
Remaining is a non blocking comment

kimbor (approved):

Issue #14255, opened 2024-07-16
Shared Gradle Convention Plugins for Hiero


jjohannes:
### Problem

Many repositories in the [github.com/hashgraph](https://github.com/hashgraph) organization use Gradle to define their project structure and build (compile, test, check, ...) the code. The currently most sophisticated setup is the one in this repository [hedera-services](https://github.com/hashgraph/hedera-services). In the past, the configuration in this repository was often copied when new repositories were created. This approach has a number of downsides:

- Copying, and then later not fixing, _undiscovered issue_ 
In the past, copying was done at arbitrary points. Possibly copying "undiscovered issue" that were then only fixed here, but not in the copies. There is no mechanism in place to track or even know about the copies.
- Inconsistent structures
Although the Gradle setup in `hedera-services` is "more stable" now, it is still in flux and there are areas to improve it further. In the past, larger changes were made from time to time (e.g. in which folder Gradle config is located). We now have inconsistency in the different repos, where the same things are named differently or are located in different locations.
- Maintainability
Even if we would somehow improve tracking the "copies" it would be quite some work to manually update all places if we make a change
- Ease in setting up new repositories
Right now, setting up a new repo requires someone with a lot of expertise with the Gradle setup and the new setup is already doomed to be "imperfect" from the beginning (e.g. containing, in that context, unused configuration code).

### General solution idea

Gradle plugins can be published to Maven repositories, just as we publish Modules to repositories. Technically, any Maven repository can be used, but there is a "Maven Central for Gradle Plugins" called the "Gradle Plugin Portal" - https://plugins.gradle.org/ - that may be used to publish open-source plugins.

We use the approach of _Convention Plugins_ to configure the Gradle setup already. Technically, Gradle makes no difference between Convention Plugins and other "types" of plugins. Hence, we _could_ publish the plugins we have defined in [gradle/plugins/src/main/kotlin](gradle/plugins/src/main/kotlin) in as they are. Then we can reuse them in other repositories.

Most of the convention plugins are not specific to _hedera-services_ already. Which means they are almost good to be reused in another context/repository.

### Concrete Proposal:

[...]

PR #15142, opened 2024-08-27
chore: move EVM modules to own directory

jjohannes:
**Description**:

In the context of #14255 (common project structures and build configuration across Hedera) we would like to establish a consistent structure. The we don't need to do "special handling" of "special cases" in the build configuration and can share it across repositories.

One such consideration is that if multiple "products" reside in one repository, each is located in a separate **directory**. All modules that have the same **group** belong to a product.

EVM has its own group, but is currently located inside the `hedera-node` directory. This PR moves the two EVM modules to the new `hedera-evm` toplevel directory.

Note: this move is a pure restructuring of the top level directory structure. It does not change anything in the software itself.

**Related issue(s)**:

#14255

**Notes for reviewer**:

**Checklist**


david-bakin-sl (approved):
LGTM.

FYI: This module is becoming less relevant - mainly was used for mono-services and though there are still references to it it's not the way forward.  That is, the way forward includes a different structure of the EVM module so that the mirror node can consume it easier, when that effort (in progress) is complete (by both services and mirror node) then we'll be able to nuke this entirely.

jjohannes:
Superseded by #15271

PR #15150, opened 2024-08-27
feat: add `SwirldState#sealConsensusRound()` lifecycle

### LINKED ISSUE #15139: Add default no-op `SwirldState#sealConsensusRound(Round)`
### Problem

After calling `SwirldState#handleConsensusRound()` [here](https://github.com/hashgraph/hedera-services/blob/stream-manager-spike/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/eventhandling/DefaultTransactionHandler.java#L201), the `DefaultTransactionHandler` makes up to two more changes to state owned by `PlatformStateService`:
  1. The event running hash; and,
  2. The last frozen time.

On the preview block streams branch, this means that the `BlockStreamManager#endBlock(Bytes)` method [here](https://github.com/hashgraph/hedera-services/blob/15073-multiple-blocks/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java#L254) is called slightly too early---this line,
```
        // Flush the boundary state changes as our final hashable block item
        pendingItems.add(boundaryStateChangeListener.flushChanges());
```
will miss the final changes to the `PlatformState` singleton.

We currently still get our `StateChangesValidator` to pass with a hack not worth explaining; but the better approach is ofc to give the application a chance to end its block with all changes made to state.

### Solution

- [ ] Add a `SwirldState#sealConsensusRound(Round)` method with a default no-op implementation.
- [ ] Update `DefaultTransactionHandler` to call this method after it has made all platform state changes.
- [ ] Implement `sealConsensusRound()` in `MerkleStateRoot` via a new lifecycle method that delegates to `BlockStreamManager#endBlock()`.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15139 
 - Fixes a flake in `TokenAirdropTest`.

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
@@ -80,8 +80,11 @@ import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.TestMethodOrder;
 
 @Tag(CRYPTO)
 @HapiTestLifecycle
@@ -378,6 +381,7 @@ public class TokenAirdropTest extends TokenAirdropBase {
     // custom fees
     @Nested
     @DisplayName("with custom fees for")
+    @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
     class AirdropTokensWithCustomFees {
 
         private static long hbarFee = 1_000L;
@@ -391,6 +395,7 @@ public class TokenAirdropTest extends TokenAirdropBase {
 
         @HapiTest
         @DisplayName("fungible token with fixed Hbar fee")
+        @Order(1)
         final Stream<DynamicTest> airdropFungibleWithFixedHbarCustomFee() {
             final var initialBalance = 100 * ONE_HUNDRED_HBARS;
             return defaultHapiSpec(" sender should prepay hbar custom fee")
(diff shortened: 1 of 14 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -80,8 +80,11 @@
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Nested;
```
david-bakin-sl, line 84: I see no reason this file is in this PR?

david-bakin-sl (commented):
Just a comment.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
tinker-michaelj: It's flaky without sequential execution, fixed here. 😵‍💫

david-bakin-sl: Ah.  So you can add ordering _without_ any of our other annotations, like `LeakyHapiTest` or `OrderedInIsolation`?

tinker-michaelj: Yes, it's a JUnit Jupiter feature, you just need to annotate the test class with,
```
@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```

david-bakin-sl: Yes ... I guess I thought that to need an order dependency they must be leaky.  But I think now (rereading javadoc on our annotations) leaky means it changes properties (or other things enumerated in `ContextRequirement` but _does not_ include things like a test creates an account or leaves a balance in an account that is used by another test.  (That's just bad practice, I would suppose.). (I'll ask in another venue what other reasons there are for needing ordering that _aren't_ leaky or just mistakes.)

edward-swirldslabs (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #15160, opened 2024-08-27
docs: tss block signing proposal

### LINKED ISSUE #14268: TSS-Block-Signing Proposal: Platform design for signing a message with TSS
This ticket encompasses the work for getting the TSS Block Signing proposal to the point where it is ready for voting. Getting the proposal through the voting process is covered in a different ticket.

edward-swirldslabs:
This PR creates the TSS-Block-Signing proposal

Fixes #14268


platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
+# Proposal Title Goes Here
```
anthony-swirldslabs, line 1: Nit: seems like it's gone elsewhere :)

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+- Signing blocks and construction of block proofs with TSS signatures is unblocked.
+
+### Requirements
+
+The core requirements governing the TSS proposals are stated in `TSS-Ledger-ID`.
+
+Blocks Signing Specific Requirements:
+
+1. A block is considered signed if its root hash has a ledger signature or a following block's root hash has a
+   ledger signature.
```
anthony-swirldslabs, line 91: Why not simplify and only require a signature on the following block at all times?

Conversely, why not harden the requirement and require a signature on the very block that is to be considered signed at all times?

If both the options are viable, then this needs to be explained further and in greater details as to why, and under what conditions the options are used.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+     */
+    void requestLedgerSignature(@Nonnull byte[] messageHash);
+
+    /**
+     * Registers a consumer of pairs where the first element is the message hash and the second element is the
+     * ledger signature on the message hash.
+     *
+     * @param consumer the consumer of ledger signatures on message hashes.
+     */
+    void registerLedgerSignatureConsumer(@Nonnull final Consumer<Pair<Bytes, PairingSignature>> consumer);
```
anthony-swirldslabs, line 163: Minor: might consider using a `BiConsumer<Bytes, PairingSignature>` instead. This simplifies the API by eliminating the "Pair", as well as reduces memory footprint by not having to allocate that Pair.

Also, the calling code would have to keep its `byte[]` instance, and then compare the arrays byte-by-byte to figure out which signature signs which message. I'm unsure if this is the most convenient/optimal API. If both the request and register methods accepted a `Bytes`, then perhaps its hashCode() could be used instead to identify the reply?

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+
+    ...
+
+    /**
+     * Requests a ledger signature on a message hash.  The ledger signature is computed asynchronously and returned
+     * to all consumers that have been registered through {@link #registerLedgerSignatureConsumer(Consumer)}.
+     *
+     * @param messageHash The hash of the message to be signed by the ledger.
+     */
+    void requestLedgerSignature(@Nonnull byte[] messageHash);
```
anthony-swirldslabs, line 155: Minor: why not `Bytes`?

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+     *
+     * @param consumer the consumer of ledger signatures on message hashes.
+     */
+    void registerLedgerSignatureConsumer(@Nonnull final Consumer<Pair<Bytes, PairingSignature>> consumer);
+}
+```
+
+#### System Transactions
+
+The `TssShareSignatureTransaction` is a new system transaction that is only generated by the `TssBaseService` and
```
anthony-swirldslabs, line 169: "System transactions" are currently implemented as "high priority" transactions which are handled before regular transactions. And currently, there's only one type of transactions that's handled this way - the `StateSignatureTransaction`. Additionally, this is a platform-level concept as of today.

Contrary to that, the TssBaseService is a service (a part of the "app", rather than "paltform"). While the speed of blocks signing matters, it's unclear why it should have the same priority as the state signing transaction.

So I'm wondering what would go wrong if TSS transactions were regular transactions? How bad would that be?

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+   * The share signature produced by the share.
+   */
+  bytes share_signature = 4;
+}
+```
+
+### Architecture and/or Components
+
+In addition to the extension of the `TssBaseService` interface, the service's `TransactionHandler` is extended to
+handle `TssShareSignatureTransaction` in the `pre-handle` phase.
```
anthony-swirldslabs, line 203: I'm curious if this can be implemented in the TssBaseService itself w/o touching any other components? Cannot a regular service pre-handle transactions w/o having to modify the TransactionHandler?

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,286 @@
...
+  generating share signature transactions, and handling share signature transactions do not modify the state and
+  should not execute on the `transaction handling thread`.
+* As soon as the ledger signature for a message can be constructed, all registered consumers are notified with the
+  message hash and the ledger signature. If there are multiple consumers registered, whether they are given the
+  signature serially or in parallel makes no difference as the data is immutable and consumption is expected to be
+  idempotent.
+
+Ledger Signature Generation:
+
+* The storage of `TssShareSignatureTransaction` is only in memory and expires after a configured amount of time.
```
anthony-swirldslabs, line 219: If I understand the description of the proposed implementation correctly, then we expect the code on every node to call the request method at approximately the same time, and that would cause each node to generate a `TssShareSignatureTransaction` and send it to every other node, thus allowing each node to collect the full set of the signatures.

In other words, there's not a way for a single node to call the request method and get a random message signed with the ledger id because there's not a way for that node to request all the other nodes to produce their signatures. So this means that we wouldn't be able to expose the signing method as an API on a node.

Is it good or is it bad?

edward-swirldslabs: Right, this is an internal API that has to be called deterministically on all nodes.  If we wanted to open up the service, like a Notary that signs user/customer messages, then we would need to create a new NotaryService that is user facing and create a new user generated transaction request for that.  It would get gossiped to all nodes and the NotaryService would handle the new transaction after consensus in a deterministic call to create the ledger signature by registering a ledger signature consumer with the TssBaseService and invoking the `requestLedgerSignature` method.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,309 @@
...
+5. The number of consecutive blocks that are not signed is minimized.
+
+The rationale for these requirements is that we can
+construct [Merkle Proofs](https://ethereum.org/en/developers/tutorials/merkle-proofs-for-offline-data-integrity/) to
+extend the signature on Block N to Blocks N-1, N-2, ... all the way back to Block 1. Each block contains the root
+hash of the previous block as its first element with 1 internal merkle node between the root hash of a block and
+the root hash of the previous block. The length of a Merkle Proof extending the signature from Block N to Block
+(N-Y) is 3Y+1 hashes.
+
+To construct a TSS ledger signature, at least 1/2 the shares must sign the same message. These share signatures
```
litt3, line 104: Here, you explicitly say 1/2, but elsewhere you discuss a configurable threshold

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,309 @@
...
+  about the failure to produce a ledger signature.
+  * This infrequent failure to produce a ledger signature is expected to be rare and tolerated by ledger signature
+    consumers.
+  * Failures will be logged at an `INFO` level and metrics will be kept to track their occurrence.
+
+### Configuration
+
+New Configuration:
+
+* `tss.signatureLivenessPeriod` - The amount of time a share signature is held in memory before being discarded.
```
litt3, line 256: I'm not sure we need this configuration value. Shouldn't it be enough to keep share signatures until either

1. an aggregate signature is produced for the block
2. the block is emitted without a ledger signature

and immediately delete once either of these conditions are met? This guarantees we will have share signatures as long as they are required, and no longer.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,309 @@
...
+     */
+    void requestLedgerSignature(@Nonnull byte[] messageHash);
+
+    /**
+     * Registers a consumer of pairs where the first element is the message hash and the second element is the
+     * ledger signature on the message hash.
+     *
+     * @param consumer the consumer of ledger signatures on message hashes.
+     */
+    void registerLedgerSignatureConsumer(@Nonnull final BiConsumer<byte[], PairingSignature> consumer);
```
tinker-michaelj, line 184: For completeness I suggest a,
```
    /**
     * Unregisters a consumer of pairs where the first element is the message hash and the second element is the
     * ledger signature on the message hash.
     *
     * @param consumer the consumer of ledger signatures on message hashes to unregister
     */
    void unregisterLedgerSignatureConsumer(@Nonnull final BiConsumer<byte[], PairingSignature> consumer);
```

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: This assumes that block signing will be the only use case for ledger signatures.  I would rather not create a call back mechanism where block signing has to notify the `TssBaseService` that a block was sent without a signature.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: You are correct.  In the TSS-Ledger-ID is it explicitly `TS +2/2` which is > 1/2 and not configuration dependent.   It is dependent on the number of shares in the candidate roster.  I will update the text for greater clarity.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: I went ahead and added it since I was editing the proposal.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
litt3: It wouldn't need to be a call explicitly saying "block sent without a signature". It could just be an additional method saying `nevermindAboutThisSigningRequest`

There a couple reasons I think this would be a preferable solution:

1. it's simpler to implement than a time based retention mechanism
2. it minimizes number of share signatures that must be kept
3. This alternate solution is equally compatible with future ledger signature use cases: simply keep share signatures either until the aggregate is produced, or `nevermindAboutThisSigningRequest` is called

litt3: It wouldn't need to be a call explicitly saying "block sent without a signature". It could just be an additional method saying `nevermindAboutThisSigningRequest`

There a couple reasons I think this would be a preferable solution:

1. it minimizes number of share signatures that must be kept
2. going this route would simplify the initial implementation, since block signatures wouldn't need to have a time-based cleanup. if a time-based cleanup is ever required for a future use-case, it could be implemented at that time

edward-swirldslabs: Suppose the call back never occurs and the signature is not created because a threshold of share signatures were not aggregated.  That's a memory leak at that point without a timeout.   I don't see how to safely avoid not needing a timeout.   We could infer a timeout instead of having a configuration value based on later signature requests completing and their share signatures being garbage collected.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
```diff
@@ -0,0 +1,317 @@
...
+     * @param messageHash The hash of the message to be signed by the ledger.
+     */
+    void requestLedgerSignature(@Nonnull byte[] messageHash);
+
+    /**
+     * Registers a consumer of the message hash and the ledger signature on the message hash.
+     *
+     * @param consumer the consumer of ledger signatures and message hashes.
+     */
+    void registerLedgerSignatureConsumer(@Nonnull final BiConsumer<byte[], PairingSignature> consumer);
```
tinker-michaelj, line 183: Do we actually need a new type `PairingSignature`? Why not just `byte[]` or `Bytes`?

edward-swirldslabs: It will not be a PBJ object unless someone else converts it to one because they need it in that form.    @mxtartaglia-sl , @AlfredoG87  What form should the ledger signature be communicated in?   byte[]?  This is communication to the block stream code for adding it to the to block.  I'm assuming that when we use the block signature, we'd be using the TSS-Library and have to convert it back to a PairingSignature in order to verify it with the library?

AlfredoG87: I agree with @tinker-michaelj that an byte[] should be good.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
anthony-swirldslabs: Thanks for the extra details and wording in the doc. That helps.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
anthony-swirldslabs: Fair. However, I'm not sure if this particular API is going to be invoked so very frequently that an extra object allocation would present a performance issue.

The reason why `Bytes` might be preferred is because it represents an **immutable** array of bytes as opposed to a plain `byte[]` object.

Also, I'm unsure if an underlying low-level library requirements should define the high-level API input to this extent.

Having said that, I don't know which objects the caller handles. If they start with `byte[]`, then indeed, this might be the way to go. However, if they have a `Bytes` and then have to allocate a new bytes array just to pass it to this API, then perhaps we should reconsider.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
anthony-swirldslabs: I see. In that case I suggest to document this reasoning in the design doc so that we have a clear justification for choosing system transactions over regular transactions.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
anthony-swirldslabs: Got it. Thanks for the clarification!

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
anthony-swirldslabs: Makes sense. Thanks for the additional details!

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
mxtartaglia-sl: >@mxtartaglia-sl , @AlfredoG87 What form should the ledger signature be communicated in? byte[]?

@edward-swirldslabs, what are the expectations for the consumer? Should it write the signature to the block and be done with it, or should it do something else with it?

It is still unclear to me how external parties will read the signatures from the block. The block node is covered with a byte array if it uses the TSS library.

I prefer byte[] for all the reasons I explained in the meeting (an opaque byte array provides flexibility and allows curve adjustment or group assignment changes with less impact). Still, I am unsure if that would bring extra difficulties for external block consumers.

But if we use `PairingSignature`, it should be a protobuf type. Given that If the intention for the consumer is only to write the signature, then we are unnecessarily coupling the library to the consumer where they won't need to use it other than for calling the #toByteArray() method on the signature.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: Even if we input in `Bytes`, the underlying library API is in `byte[]`, so the conversion has to happen anyway.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: Per our discussion yesterday, the ledger signature and the ledger id will be `byte[]` in the form that best fits consensus node and `TSS-Library` needs.   We'll publish a byte format for parsing the byte array to extract components by interested parties.   But for the block stream and consensus node, it's just a block of bytes.

platform-sdk/docs/proposals/TSS-Block-Signing/tss-block-signing.md
edward-swirldslabs: I've updated the description with the desired reasoning.

anthony-swirldslabs (approved):
+1

mxtartaglia-sl (approved):

kfa-aguda (approved):
+0.8

Issue #15091, opened 2024-08-23
Design proposal: migrate Queue states from FCQueue to VirtualMap


artemananiev:
This ticket is to create a design proposal for https://github.com/hashgraph/hedera-services/issues/15090

Issue #15090, opened 2024-08-23
Create `QueueNode` backed by `VirtualMap` instead of `FCQueue`


artemananiev:
Queue states are currently implemented as merkle nodes with two children: a label (`StringLeaf`) and a queue (`FCQueue`). All Queue state APIs are mapped 1-1 to the underlyng `FCQueue`.

There are no plans to support `FCQueue` in the future, for a good number of reasons. Instead, Queue states may be implemented on top of `VirtualMap`.

artemananiev:
This task is blocked by https://github.com/hashgraph/hedera-services/issues/15448. Without that feature, these QueueNode changes cause about 15% performance regression in NFT transfer benchmark.

artemananiev:
This ticket was discussed with the team, and a decision was made to close it at this point. Some changes from the PR will be included to #14803, in particular how queue states are migrated to Virtual Mega Map.

PR #15165, opened 2024-08-27
docs: 15091: Design proposal: migrate Queue states from FCQueue to VirtualMap

artemananiev:
Design proposal for https://github.com/hashgraph/hedera-services/issues/15090.

Fixes: https://github.com/hashgraph/hedera-services/issues/15091
Signed-off-by: Artem Ananev


platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/map-keys.svg
OlegMazurov, line 1: `FCQueue` adds at the tail and removes from the head.
Would be good to preserve for consistency.

artemananiev: Head/tail are implementation details, not a part of queue API. Since I haven't worked with `FCQueue` much, using head to add new elements and tail to remove looked more natural to me, but I'm fine with changing it to match `FCQueue` @jasperpotts @rbair23 what do you think?

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+
+Migration code will be in `QueueNode.migrate()` This method will get its right child (which is
+`FCQueue`), iterate over it, insert all elements in the right order to a virtual map, then set
+the map as the new right child. The left child will be changed from a simple string leaf to a
+`QueueNodeState` leaf node.
+
+## Testing
+
+This feature doesn't add any new functionality. All existing Hedera unit tests should pass with
+the new queue state implementation. Unit tests specific to `QueueNode` may need to be updated.
```
OlegMazurov, line 134: `FCQueue` allows for concurrent updates (`add`/`remove`). I believe that is tested. Either that requirement should be dropped (as with `VirtualMap`) or a layer of synchronized methods on top of `VirtualMap` should be added. The latter will have performance impact.

artemananiev: The way how `FCQeueue` is thread safe is all its methods like add() or remove() are synchronized. As the simplest solution, similar synchronization can be added on top of `VirtualMap`, with similar performance impact.

However, first of all it would be great to understand how queue states are used today. Are they really updated from multiple threads in parallel? @tinker-michaelj @Neeharika-Sompalli your input would be much appreciated here

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Migration testing
+
+Migration testing will use a real mainnet state generated with a previous Hedera version. This
+state will be used to start a node with the latest version that includes queue state changes.
+
+## Performance impact
+
+**Queue operations (add, remove)**: both `FCQueue` and `VirtualMap` add and remove elements fast.
+No performance impact is expected.
```
anthony-swirldslabs, line 144: Could you please expand on the "fast" part? Specifically, I'd like to see the performance of all the operations expressed using the big-O notation for both the old and the new implementations.

Additionally, it would be helpful to provide sizes of existing queues in the current state as well as describe (/speculate about) their future growth to help understand if the new implementation's performance characteristics support the future growth.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+
+**Hashing**: `FCQueue` has a very efficient hashing algorithm, it hashes only newly inserted
+elements and "unhashes" removed elements. It's done on a single thread, however. `VirtualMap`
+needs to rehash more entities because of its tree structure, but it's done in parallel on
+many threads. No significant performance impact is expected.
```
anthony-swirldslabs, line 149: I'm unsure about the "no significant" part. You're saying the FCQueue's hashing algorithm is optimized to rehash the deltas. Additionally, the FCQueue lives in memory today. This sounds pretty fast to me.

On the other hand, you're saying the VM has to hash more entities (i.e. no support for optimized delta hashing), and on top of this, VMs live on disk. All of this sounds pretty slow to me.

So I'd expect some degradation of the hashing performance. It would be interesting to understand how large it would be.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+`QueueNode`:
+
+* No longer marked as `@Deprecated`
+* Has two children: `QueueNodeState` (was: `StringLeaf`) and `VirtualMap` (was: `FCQueue`)
+
+`QueueNodeState`:
+
+* A simple leaf node similar to `StringLeaf`, but contains two long values in addition to
+  a string value
+* These two longs are the head and the tail of the queue - see below
```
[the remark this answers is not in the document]
artemananiev: When the whole state tree is migrated to a mega map, tree structure will be completely different. There will be no state labels (those `StringLeaf`s above), no `SingletonNode` or `QueueNode`, and much more. This is beyond the scope of the current proposal. At this point, only queue state implementation (`QueueNode` and below) is changed, while the rest of the tree is unaffected

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
artemananiev: "Fast" here means O(1), both for FCQueue and VirtualMap.

Existing queue sizes: several queues with just a few (under a hundred, I guess elements) and one queue with about 1.8M elements (transaction receipts at 10K TPS). The latter shouldn't grow over time, it's to keep transaction receipts from the last 3 mins. The size of this queue is more or less constant, as older receipts are removed from the queue automatically

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
artemananiev: It's slightly more complicated than "in memory" vs "on disk". When a virtual map copy is hashed, all its dirty data (recently changed entities) is in the virtual node cache. Only after a copy is fully hashed, it can be flushed to disk. Some unchanged elements along the path to the root of the virtual tree may be loaded from disk, but not dirty leaf nodes. Given that virtual map is hashed on multiple threads, I assume there will be no significant impact on hashing time. Exact numbers can be provided only after this proposal is implemented.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+### Java classes
+
+No changes to `ReadableQueueState` and `WritableQueueState` interfaces.
+
+No changes to `ReadableQueueStateImpl` and `WritableQueueStateImpl` classes.
+
+`QueueNode`:
+
+* No longer marked as `@Deprecated`
+* Has two children: `QueueNodeState` (was: `StringLeaf`) and `VirtualMap` (was: `FCQueue`)
```
rbair23, line 56: VirtualMap is `labeled`, right? Do we need the other child?

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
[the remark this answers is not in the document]
rbair23: How far off is mega-map? Is it worth doing the interim step?

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
rbair23: They couldn't possibly be updated from multiple threads, as that would lead to non-determinism in the state. It must be that it is read from multiple threads but only updated by one.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
anthony-swirldslabs: Got it. Thanks. I'd guess the transaction receipts queue is going to grow as the network grows, but I agree it's not going to be an explosive growth.

Regardless, as long as all the operations are O(1) for both the implementations, I think we should be good as far as scaling is concerned.

I suggest to include the big-O and queue sizes in the document for future reference.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
anthony-swirldslabs: That still sounds a bit tricky, but it seems that this is because the hashing implementation in virtual maps isn't exactly trivial, which is fine.

One more question on this topic: why couldn't we add a third number to the `QueueNodeState` to facilitate delta hashing for the queue as a whole, basically retaining the existing hashing algorithm for queues? Could we consider this option in the design document and provide pros/cons to decide if we want to do that?

artemananiev: It can be done, perhaps. I mean, maintain the running hash code at `QueueNode` level and ignore the hash code from the underlying virtual map. What's the benefit, though? This is extra hashing code, extra value to store somewhere, some additional tricks to disable virtual map hashing, etc.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+`QueueNodeState` leaf node.
+
+## Testing
+
+This feature doesn't add any new functionality. All existing Hedera unit tests should pass with
+the new queue state implementation. Unit tests specific to `QueueNode` may need to be updated.
+
+### Migration testing
+
+Migration testing will use a real mainnet state generated with a previous Hedera version. This
```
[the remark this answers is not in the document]
artemananiev: I don't think so. "Migration" in MTT is confusing: it does not test migrations like FCQueue to VirtualMap, it only checks that a state serialized in an old version can be loaded and used in a new version.

MTT today uses a state with two merkle nodes: a merkle map and a virtual map. First, the merkle map doesn't make any sense, it isn't used any longer in Hedera. Second, after this feature, queue states migration to virtual maps, FCQueues won't be used anywhere, too. What would be the reason to check that FCQueues can be used across software versions, if there are no apps using them?

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+lots of elements, and queue hashes must be computed fast.
+
+There are several drawbacks in `FCQueue`, too:
+
+* All data is in memory
+* When a state snapshot is taken, all queue elements are serialized to disk one by one
+* During reconnects, the whole queue is serialized and sent to the learner
+
+Instead, all the queue APIs can be implemented on top of a `VirtualMap`. It's more memory friendly,
+optimizied for reconnects and snapshots.
```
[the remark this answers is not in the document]
artemananiev: OK, I'll add a section about alternative (and rejected) approaches

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
anthony-swirldslabs: Sounds good.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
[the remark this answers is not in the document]
anthony-swirldslabs: The process can be split the other way around: introduce the mega-map first, and then get back to refactoring the queues after the mega-map exists. I believe this would require less of throw-away work, so I disagree with your argument.

But if you insist on delivering the work in the current sequence, then please feel free to resolve this conversation.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
anthony-swirldslabs: Removing the synchronization from the existing implementation sounds risky to me.

Also, I'm not at all clear as to how the current queues achieve determinism. Synchronized `add`/`remove` methods alone won't do the trick because parallel threads can obtain the lock in random order. I'd suggest to study the current usages of the `add`/`remove` methods to understand this aspect (and/or to bug Services people about it), to ensure the new implementation doesn't break existing logic w.r.t. parallel execution.

lpetrovic05 (dismissed):

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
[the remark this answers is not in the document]
artemananiev: I don't see how it can be done the other way round. When a single virtual map is implemented, all data should be in it. All existing states (singletons, queues, maps) will be migrated to that single map. It means, there will be migration code for queues. There must be an answer to the question how to map queue elements to key/values. This is what covered in this design proposal. The idea is to have two smaller steps rather than a single larger one, but no code will be thrown away

imalygin (dismissed):

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
OlegMazurov: No, `synchronized` methods do not provide determinism. It is achieved by the mode of use (all updates are done by a single thread, namely, transaction handling thread), not by implementation.
The `FCQueue` class, unlike `VirtualMap`, was initially designed as a general purpose data structure to be used in multithreaded environment (and that was tested by unit tests). This requirement, however, is not needed for Services code where `FCQueues` are part of the state. Is there any other use? If yes (which I doubt), that should be left alone and not converted to `VirtualMap`.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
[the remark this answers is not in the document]
anthony-swirldslabs: The code (and disk artifacts) that create and support a dedicated virtual map will be a throw away code. Similarly, the code that's going to use the map is going to have to migrate to using the States API constructs (or whatever API the mega-map will expose), so at the low level it will also be a throw away code.

Generally, the assumption was that the migration to the mega-map would be gradual - one or a few data structures at a time. Your last comment seems to suggest that it's going to be a big-bang with migrating everything at once. I find this approach risky, so again, I cannot fully agree with it.

But like I said, if you prefer to deliver this work this way, please feel free to resolve this conversation.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
anthony-swirldslabs: Just to clarify: are you saying that the transaction handling thread handles all transactions in the consensus (or any other deterministic) order? This seems to be the only way to achieve determinism for the queues.

If so, then yes, I agree that we don't want to implement synchronized methods on the virtual map since they won't really be useful.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Java classes
+
+No changes to `ReadableQueueState` and `WritableQueueState` interfaces.
+
+No changes to `ReadableQueueStateImpl` and `WritableQueueStateImpl` classes.
+
+`QueueNode`:
+
+* No longer marked as `@Deprecated`
```
kfa-aguda, line 55: to clarify, `QueueNode` will be undeprecated but `FCQueue` will now be deprecated?

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
OlegMazurov: That's right, all state changes are performed by the transaction handling thread when processing transactions in the consensus order.

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
kfa-aguda: +0.8

litt3 (approved):

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/proposal.md
artemananiev: Correct

platform-sdk/docs/proposals/queuestate-fcqueue-to-virtualmap/map-keys.svg
artemananiev: I found this JavaDoc in `WritableQueueState`:

```java
    /**
     * Retrieves and removes the element at the head of the queue, or returns null if the queue is empty.
     *
     * @return The element at the head of the queue, or null if the queue is empty.
     */
    @Nullable
    default E poll() {
        return removeIf(e -> true);
    }
```

So elements are added to the tail and removed from the head. I'll update the diagrams accordingly.

jasperpotts (approved):
Looks good, only question is around if this step is needed between where we are today and one big map. It seems like some level of wasted work.

imalygin (approved):

rbair23 (approved):

lpetrovic05 (approved):

anthony-swirldslabs (approved):

PR #15223, opened 2024-08-28
fix: add background traffic to `DabEnabledUpgradeTest`

### LINKED ISSUE #15220: Upgrade HAPI tests should use background traffic
### Description

`@HapiTest`'s run with an artificial 1 minute staking period; and the long sleeps in `DabEnabledUpgradeTest` can lead to a period passing without a single transaction.

Because this is an impossible condition in production with 24 hour staking periods, pending reward management does not take into account skipped periods and this leads to errors in the test logs.

### Steps to reproduce

Run `DabEnabledUpgradeTest` with `@Order(MAX_VALUE - 2)` so `LogValidationTest` comes next.

### Additional context

_No response_

### Hedera network

_No response_

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #15220

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
@@ -37,7 +37,6 @@ import static com.hedera.services.bdd.suites.HapiSuite.ONE_BILLION_HBARS;
 import static com.hedera.services.bdd.suites.HapiSuite.ONE_MILLION_HBARS;
 import static com.hedera.services.bdd.suites.hip869.NodeCreateTest.generateX509Certificates;
 import static com.hedera.services.bdd.suites.regression.system.LifecycleTest.configVersionOf;
-import static java.lang.Integer.MAX_VALUE;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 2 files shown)


Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
```diff
@@ -88,7 +87,7 @@
  * the config version is still zero.
  */
 @Tag(UPGRADE)
-@Order(MAX_VALUE)
+@Order(Integer.MAX_VALUE - 2)
```
david-bakin-sl, line 90: This strange expression `Integer.MAX_VALUE - 2` appears in no other `@Order` so isn't obvious.  Why not leave a comment here explaining what it says in the underlying issue?  And is it `- 2` instead of `- 1` so that it preceeds `StreamValidationTest` in addition to `LogValidationTest`?

david-bakin-sl (approved):
LGTM with one suggestion for adding a comment to the code.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/regression/system/DabEnabledUpgradeTest.java
tinker-michaelj: Ah yes I should add this detail @david-bakin-sl , will do in next PR. 🙏

PR #15177, opened 2024-08-28
chore: Sync latest block stream proto changes

mhess-swl:
* Create `roster` and `roster_state` protos
* Various property updates to `block_stream_info`

PR #15225, opened 2024-08-28
chore: Mirroring state changes update to block stream for roster values

jsync-swirlds:
* Same changes made in the hedera-protobufs repo to add roster and roster_state to the block stream state_changes proto.

This replaces PR #15177, which unintentionally included an "undo" of other recent changes.


tinker-michaelj (approved):
LGTM, ty @jsync-swirlds !

mhess-swl (approved):

PR #15233, opened 2024-08-29
feat: Various block stream modifications

mhess-swl:
More preparation to merge `stream-manager-spike` into `develop`. The following changes are contained:

* Add overload of hash by block number
* Add exchange rate and memo to staking period updater
* Minor update to roster service name
* System setup to use exchange rate; initialize treasury clones earlier
* Roster state key update
* Incorporate exchange rate into node stake updates
* Use exchange rate and user txn commit only in handle workflow
* Minor docs, test default


tinker-michaelj (approved):
LGTM, tyvm @mhess-swl !

PR #15256, opened 2024-08-29
feat: enable PR check preview block streams

### LINKED ISSUE #15254: Enable preview block streams in PR checks
### Problem

We want to test with `blockStream.streamMode=BOTH` in all `@HapiTest` PR checks, including the `StateChangesValidator` and `TransactionRecordParityValidator`.

### Solution

Set this property [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/configuration/dev/application.properties).

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15254 
 - Adds final feature flags around block stream-specific callbacks.
 - Removes `StreamMode#BLOCKS` enum value because we haven't integrated block-item-to-record translation, so we don't fully support an exclusive block stream yet.
 - Adds a few fields omitted from `ContractService` record stream items for mono-service diff testing:
     * The `TransactionBody#memo` for the synthetic `CryptoCreate` that creates a hollow account.
     * The `TransactionRecord#evm_address` for the record of a successful hollow account creation.
     * The `ContractFunctionResult#contractID` (if available) in the result of a failed contract operation.
 - Including these fields simplifies the task of translating back from block items only to the current `RecordStreamItem`s.


hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -48,36 +53,36 @@ option java_multiple_files = true;
...
+     * <p>
+     * If this message is for block number N, then the earliest available
+     * hash SHALL be for block number N-256.<br/>
+     * The latest available hash SHALL be for block N-1.<br/>
+     * This is REQUIRED to implement the EVM `BLOCKHASH` opcode.
      */
-    bytes trailing_block_hashes = 5;
-}
+    bytes trailing_block_hashes = 4;
+}
```
jsync-swirlds, line 88: Need a newline at EOF.

hapi/src/main/java/module-info.java
```diff
@@ -56,12 +56,14 @@
...
-    exports com.hedera.hapi.platform.state;
-    exports com.hedera.hapi.node.state.roster;
     exports com.hedera.hapi.block.stream;
     exports com.hedera.hapi.block.stream.input;
     exports com.hedera.hapi.block.stream.output;
+    exports com.hedera.hapi.platform.state;
+    exports com.hedera.hapi.node.state.roster;
     exports com.hedera.hapi.block.stream.schema;
+    exports com.hedera.hapi.block.protoc;
+    exports com.hedera.hapi.block.stream.protoc;
```
jsync-swirlds, line 66: Is there a chance we could avoid the protoc versions and use the PBJ versions only?

hedera-node/configuration/dev/application.properties
```diff
@@ -9,3 +9,4 @@ staking.fees.stakingRewardPercentage=10
 # Needed for Restart and Reconnect HapiTests that run many transactions of each type
 bootstrap.throttleDefsJson.resource=genesis/throttles-dev.json
 ledger.id=0x03
+blockStream.streamMode=BOTH
```
jsync-swirlds, line 12: Do we need a newline at EOF here?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,129 @@
...
+        asyncStub = createBlockNodeAsyncStub();
+
+        StreamObserver<PublishStreamRequest> requestObserver =
+                asyncStub.publishBlockStream(new StreamObserver<PublishStreamResponse>() {
+                    @Override
+                    public void onNext(PublishStreamResponse value) {
+                        // TODO (driley) Handle EndofStream responses, and if Single Item Ack is enabled via config then
+                        // implement that as well.
+                        // However that would require keeping track of the hashes of the items sent and then verifying
+                        // before
```
jsync-swirlds, line 69: Note, itemAck is a debug option, and consensus node is free to accept-then-ignore single item acknowledgement.
End of Stream and Block acknowledgement _must_ be implemented, however.
The most complex option is probably "BEHIND".
Protocol is not fully set (suggestions welcome), but an initial description is in [this story](https://github.com/hashgraph/hedera-block-node/issues/120) to design the final protocol.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
```diff
@@ -72,10 +72,10 @@ public boolean appliesTo(@NonNull final HapiSpec spec) {
 
     public static void main(String[] args) throws IOException {
         final var input =
-                "/Users/michaeltinker/AlsoDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/block-streams/block-0.0.3/";
+                "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/block-streams/block-0.0.3/";
         final var blocks = BlockStreamAccess.BLOCK_STREAM_ACCESS.readBlocks(Paths.get(input));
         final var loc =
-                "/Users/michaeltinker/AlsoDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/recordStreams/record0.0.3";
+                "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/recordStreams/record0.0.3";
         final var records = RecordStreamAccess.RECORD_STREAM_ACCESS.readStreamDataFrom(loc, "sidecar");
```
jsync-swirlds, line 79: Commented this in a prior PR too.
This could be done with no placeholder as follows:
```java
        final Path nodeZeroDataFolder =
                WorkingDirUtils.workingDirFor(0, "hapi").resolve("data").toAbsolutePath().normalize();
        final String blockFolder = nodeZeroDataFolder.resolve("block-streams/block-0.0.3").toString();
        final String recordFolder = nodeZeroDataFolder.resolve("recordStreams/record0.0.3").toString();
```
Then use the two strings to get `blocks` and `records`.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
```diff
@@ -145,13 +146,15 @@ public class StateChangesValidator implements BlockStreamValidator {
...
-                        "/Users/michaeltinker/AlsoDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/output/swirlds.log"),
+                        "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/output/swirlds.log"),
                 Paths.get(
-                        "/Users/michaeltinker/AlsoDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/config.txt"));
+                        "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/config.txt"),
+                Paths.get(
+                        "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/config/application.properties"));
         final var input =
-                "/Users/michaeltinker/AlsoDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/block-streams/block-0.0.3/";
+                "/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/data/block-streams/block-0.0.3/";
```
jsync-swirlds, line 157: All of the strings here could be made less sensitive (so we don't have to change the code for each different engineer or clone location) by replacing the node data folder prefix `"/Users/michaeltinker/YetAnotherDev/hedera-services/hedera-node/test-clients/build/hapi-test/node0/"` with the following code.
```java
        final Path nodeZeroBaseFolder =
                WorkingDirUtils.workingDirFor(0, "hapi").toAbsolutePath().normalize();
```

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -37,7 +37,12 @@ option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
 
 /**
- * A message stored in state to maintain block stream parameters.
+ * A message stored in state to maintain block stream parameters.<br/>
+ * Nodes use this information for three purposes.
+ * 1. To maintain hash chain continuity at restart and reconnect boundaries.
+ * 1. To store historical hashes for implementation of the EVM `BLOCKHASH`
+ *    and `PREVRANDAO` opcodes.
+ * 1. To track the amount of consensus time that has passed between blocks.
```
anthony-swirldslabs, line 45: Minor: you may want to renumber the items in the list, or use an unnumbered list instead. Using the same number looks confusing for anyone reading this source file directly.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -48,36 +53,36 @@ option java_multiple_files = true;
...
-     * A block hash value.<br/>
-     * This is the hash of the last completed block.
-     */
-    bytes last_block_hash = 2;
-
-    /**
-     * A consensus time for the last completed block.
-     * This is used to determine if this block was the first across an
+     * A consensus time for the current block.<br/>
+     * This is the _first_ consensus time in the current block, and
```
anthony-swirldslabs, line 62: It's a bit unclear what "the first consensus time in the current block" actually means. Is it the consensus time of the first (earliest?) event in this block, and there's some other events with later (newer) consensus times in this block? Or is it the time when the network reached a consensus on this block for the first time? Or...?

I suggest to expand this doc to clarify what this timestamp means exactly.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -48,36 +53,36 @@ option java_multiple_files = true;
...
-    proto.Timestamp last_block_time = 3;
+    proto.Timestamp block_time = 2;
 
     /**
      * A concatenation of hash values.<br/>
      * This combines several trailing output block item hashes and
-     * is used as a seed value for a pseudo-random number generator.
+     * is used as a seed value for a pseudo-random number generator.<br/>
+     * This is also requiried to implement the EVM `PREVRANDAO` opcode.
      */
```
anthony-swirldslabs, line 75: It may be worth specifying the length requirements and constraints for this field, to ensure that algorithms consuming it can actually do their work with the data available.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -48,36 +53,36 @@ option java_multiple_files = true;
...
     /**
      * A concatenation of hash values.<br/>
-     * This combines the last 256 trailing block hashes, and is required to
-     * support the EVM `BLOCKHASH` opcode.
+     * This field combines up to 256 trailing block hashes.
+     * <p>
+     * If this message is for block number N, then the earliest available
+     * hash SHALL be for block number N-256.<br/>
+     * The latest available hash SHALL be for block N-1.<br/>
+     * This is REQUIRED to implement the EVM `BLOCKHASH` opcode.
```
anthony-swirldslabs, line 85: Same as above: it would be nice to know the min and max allowed/required length for this field, for the same reason.

anthony-swirldslabs (commented):
Platform Data changes look good to me. I posted just a few minor comments regarding the docs.

hedera-node/hedera-app/src/main/java/module-info.java
```diff
@@ -52,6 +52,7 @@
     requires static com.github.spotbugs.annotations;
     requires static com.google.auto.service;
     requires static java.compiler;
+    requires static org.jetbrains.annotations;
```
mhess-swl, line 55: `edu.umd...` instead of `org.jetbrains...`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
derektriley: Thanks for the additional information! Will take a look at the initial description.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/stack/SavepointStackImpl.java
```diff
@@ -486,7 +486,6 @@ public HandleOutput buildHandleOutput(@NonNull final Instant consensusTime) {
             }
             switch (streamMode) {
                 case RECORDS -> records.add(((RecordStreamBuilder) builder).build());
-                case BLOCKS -> requireNonNull(blockItems).addAll(((BlockStreamBuilder) builder).build());
```
Neeharika-Sompalli, line 489: Intentionally removed?

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
```diff
@@ -25,25 +27,23 @@
...
         @ConfigProperty(defaultValue = "true") @NetworkProperty boolean compressFilesOnCreation,
         @ConfigProperty(defaultValue = "1") @NetworkProperty int roundsPerBlock) {
     public boolean streamBlocks() {
-        return streamMode == StreamMode.BLOCKS || streamMode == StreamMode.BOTH;
+        return streamMode == BOTH;
     }
 
     public boolean streamRecords() {
-        return streamMode == StreamMode.RECORDS || streamMode == StreamMode.BOTH;
+        return true;
```
Neeharika-Sompalli, line 47: This is changed to be safe, that no one changes it to BLOCKS by mistake ?

Neeharika-Sompalli (commented):
Overall LGTM, other than the review comments to be addressed. HapiTests seem to be failing.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/TransactionRecordParityValidator.java
tinker-michaelj: Ah good point, fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
tinker-michaelj: Done.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: I think the repeated `1.` is the convention @jsync-swirlds is using in the protobufs.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Clarified as the round timestamp of the first round in the block.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Clarified this requires at least 256 bits of entropy.

hedera-node/hedera-app/src/main/java/module-info.java
tinker-michaelj: Fixed.

hedera-node/configuration/dev/application.properties
tinker-michaelj: Added.

hapi/src/main/java/module-info.java
tinker-michaelj: Yes, the only offender was `GrpcBlockItemWriter` and since it was an unused WIP, just removed it along with these `exports`.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Fixed.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Not sure I fully understand this request...do you have suggested wording?

Our block hashes are SHA-384 hashes, though we only use the last 32 bytes to match the length of an Ethereum block hash.

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
anthony-swirldslabs: I see. That's a bit strange as it makes it confusing when editing those files in a text editor. But okay.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
anthony-swirldslabs: Thanks!

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
anthony-swirldslabs: Thanks!

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
anthony-swirldslabs: It's unclear from the specification of the field what is the minimum size of the `bytes` field, and what is the maximum size of the field. For example, similar to a field just above this one, if it has to be at least 256 bits (aka 32 bytes) long, then it's worth documenting it, so that consumers of this field can rely on the data in the field.

anthony-swirldslabs (dismissed):

mhess-swl (approved):
LGTM 🎉 tyvm @tinker-michaelj!

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
jsync-swirlds: Perhaps something like this?
```suggestion
     * This field combines up to 256 trailing block hashes.
     * <p>
     * If this message is for block number N, then the earliest available
     * hash SHALL be for block number N-256.<br/>
     * The latest available hash SHALL be for block N-1.<br/>
     * This is REQUIRED to implement the EVM `BLOCKHASH` opcode.
     * <p>
     * <h5>Field Length</h5>
     * Each hash value SHALL be the trailing 265 bits of a SHA2-384 hash.<br/>
     * The length of this field SHALL be an integer multiple of 32 bytes.<br/>
     * This field SHALL be at least 32 bytes.<br/>
     * The maximum length of this field SHALL be 8192 bytes.
```

jsync-swirlds (dismissed):

anthony-swirldslabs (approved):

mhess-swl (approved):

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

jsync-swirlds (approved):

lukelee-sl (approved):
LGTM.  Looked over the changes in the smart contracts repo only.

PR #15258, opened 2024-08-29
chore: Schedule creation of the `release/0.54` branch

mhess-swl:
This change _schedules_ the creation of the `release/0.54` branch.


rbarkerSL (approved):

.github/workflows/config/node-release.yaml
```diff
@@ -1,11 +1,11 @@
...
+      time: "17:00:00"
     schedule:
-      - on: "2024-07-26"
-        name: release/0.53
+      - on: "2024-08-30"
+        name: release/0.54
         initial-tag:
           create: false
-          name: v0.53.0-alpha.0
+          name: v0.54.0-alpha.0
```
kimbor, line 10: Since the initial-tag create value is `false`, the `v0.54.0-alpha.0` value is not used. Is this what you intended?

mhess-swl: You're right, I read this incorrectly. For some reason I was thinking it would create the branch from the given tag, but that's not what this does.

kimbor (approved):

rbarkerSL (dismissed):

rbarkerSL (approved):

PR #15281, opened 2024-08-30
feat: HIP-632 alias-to-accounts and back

### LINKED ISSUE #4622: [HIP 632][HIP 631] Create class to handle address translation functionality for Hedera Account Service
### Problem

Epic: #4619 

There is no implementation of a class to handle address translation functionality for the Hedera Account Service.

### Solution

Create and unit test a new class - `AddressTranslatePrecompile` to handle the following calls to the Hedera Account Service.

| hash | signature | return                       | description |
| --- | --- |------------------------------| --- |
|e9482d42| getVirtualAddresses(address) | address[]                    | returns an array of virtual addresses for a given Hedera Account ID  |
|a4e310ba| getHederaAddress(address) | (responseCode, address) | returns the top level Hedera Account ID if applicable |
|d501235a| isVirtualAddress(address) | bool                         | true if valid virtual address, false if long-zero or non existing account |

Create additional wrapper classes and utility functionality as needed to be consistent with the implementation of precompile contract handlers in the rest of the system

### Alternatives

_No response_

david-bakin-sl:
**Description**:

Implement the alias<->account methods of HIP-632:

* `getEvmAddressAlias`
* `getHederaAccountNumAlias`
* `isValidAlias`

**Related issue(s)**:

Fixes #4622 

**Notes for reviewer**:

**Checklist**


hedera-node/test-clients/src/main/resources/contract/contracts/HRC632Contract/HederaAccountService.json
```diff
@@ -0,0 +1 @@
+[]
```
lukelee-sl, line 1: Did this actually compile?  The abi seems to be empty.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip632/IsAuthorizedTest.java
```diff
@@ -663,6 +664,7 @@ record TestCase(long gasAmount, ResponseCodeEnum status) {}
                     }));
             final var hapiSpec = testCase.status() == SUCCESS
                     ? throughWhen.then(getTxnRecord(recordName)
+                            .logged()
```
lukelee-sl, line 667: do we want to keep this in?  the output will be large but I guess no real harm.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/getevmaddressalias/EvmAddressAliasCall.java
```diff
@@ -0,0 +1,86 @@
...
+        super(attempt.systemContractGasCalculator(), attempt.enhancement(), true);
+        this.address = requireNonNull(address);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public @NonNull PricedResult execute() {
+        // Prepare the short-circuit error status returns
```
lukelee-sl, line 54: does this comment apply?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/hederaaccountnumalias/HederaAccountNumAliasCall.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    @Override
+    public @NonNull PricedResult execute() {
+
+        final long accountNum = accountNumberForEvmReference(address, nativeOperations());
+        if (accountNum < 0) {
+            // Invalid: not an alias for any account
+            return gasOnly(fullResultsFor(INVALID_SOLIDITY_ADDRESS, ZERO_ADDRESS), INVALID_SOLIDITY_ADDRESS, true);
+        }
+        final var account = requireNonNull(enhancement.nativeOperations().getAccount(accountNum));
```
lukelee-sl, line 56: For this and the next line would it not be better to check for the null or non existent account and return a response code?  As it is, this will throw exceptions and revert the call.

lukelee-sl (commented):
Generally looks god but we do need some BDD tests.

hedera-node/test-clients/src/main/resources/contract/contracts/HRC632Contract/HederaAccountService.json
david-bakin-sl: Yes, and the contracts which _import_ `HederaAccountService.sol` work fine.  I _think_ (but I'm not sure) that this is because this is an _abstract_ contract - in a sense it has no ABI since it can't be instantiated, only used as a base for another contact, and there's no "sharing" of code - that derived contract's binary needs to have all the methods generated right there so that's where the ABI is ...

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip632/IsAuthorizedTest.java
david-bakin-sl: Removed it (but it was probably harmless as the side-effect on the logs isn't prohibitive).

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/getevmaddressalias/EvmAddressAliasCall.java
david-bakin-sl: Removed - copy/paste error

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/hederaaccountnumalias/HederaAccountNumAliasCall.java
david-bakin-sl: Hmm. I copied this from elsewhere (e.g., see `HbarAllowanceCall.getAllowance` @L72).  Should that be fixed too?

david-bakin-sl: (answered: yes) See #15314

david-bakin-sl: Now I remember what happened.  Turns out the call to `accountNumberForEvmReference` method called at L52 above this _does_ the null check, and a null account there causes the `INVALID_SOLIDITY_ADDRESS` at L55.  The `requireNonNull` was is to shut IntelliJ up, because it doesn't know that.

But in fact since it confused you, and confused me too when I went back to look at it, I'll put an explicit (though redundant) check here so the question never again arises.

lukelee-sl (approved):
LGTM

PR #15300, opened 2024-09-02
fix: complete block -> record translation for all PR checks

### LINKED ISSUE #15285: Fully characterize block stream -> record stream backward translation
### Problem

Because the preview block stream includes all state changes, in a sense it clearly contains all information from the current record stream (and more).

However, some of the information in the record stream is less about _what_ changed and more about the reason _why_ it changed. For example,
 - The [`TransactionRecord#assessed_custom_fees`](https://github.com/hashgraph/hedera-protobufs/blob/main/services/transaction_record.proto#L108) field explains why certain token balances changed.
 - The [`TransactionRecord#paid_staking_rewards`](https://github.com/hashgraph/hedera-protobufs/blob/main/services/transaction_record.proto#L136C28-L136C49) explains why some HBAR balances changed.
 - And so on.

Some context beyond pure state changes is needed to reproduce these explanatory fields.

We want to know exactly what context beyond the current items in the preview block stream is needed to recreate the current record stream; and then decide when that context would be overly burdensome for mirror nodes to maintain.

When maintaining the context is too hard, we can add a [`TransactionOutput` item](https://github.com/hashgraph/hedera-protobufs/blob/main/block/stream/output/transaction_output.proto#L113) to supplement the `StateChanges` item.

### Solution

Finish a translator from the preview block stream that can recompute all record streams produced in PR checks.

Document the required context the translator must keep beyond a single transactional unit's block items.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15285 
 - Includes some fixes and tweaks in `BlockStreamBuilder`:
     * Sort automatic token associations in the `TransactionResult` item to match current behavior.
     * Remove redundant or unused fields `transactionOutputBuilder`, `pendingAirdropRecords`, `transactionBytes`, `isCreateResult`, `tokenType`.
     * Ensure appropriate `TransactionOutput` items are written for every `ContractFunctionResult`.
 - Takes some small steps to standardize record stream creation:
     * Does not set a redundant `evm_address` field on the record of a hollow account creation internal to the EVM. 
  - 👉 Fixes a small bug in `WritableAccountStore#summarizeContractChanges()` that failed to detect a hollow account finalized as a contract as a new contract---c.f. [this](https://github.com/hashgraph/hedera-services/blob/6a6096f334679874f5e57cd5bce149100dee3d90/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/opcodes/Create2OperationSuite.java#L507) added assertion which fails on `develop`.
  - Completes all `BlockItem`-to-`RecordStreamItem` translators.
  - "Activates" the `TransactionRecordParityValidator` by failing a PR check if there are non-zero diffs in the generated and translated record streams.


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BaseTranslator.java
```diff
@@ -153,8 +179,39 @@ public void prepareForUnit(@NonNull final BlockTransactionalUnit unit) {
...
+        knownAssociations.add(new TokenAssociation(tokenID, accountID));
+    }
+
+    /**
+     * Tracks the dissociation of a token from an account.
+     *
+     * @param tokenID the token to track
+     * @param accountID the account to track
+     */
+    public void trackDissociation(@NonNull final TokenID tokenID, @NonNull final AccountID accountID) {
```
derektriley, line 208: Is this necessary? Doesn't seem to be utilized in `TokenDissociateTranslator `and `TransactionRecord `doesn't seem to have a field for disassociations.

derektriley (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/BaseTranslator.java
tinker-michaelj: It's used [here](https://github.com/hashgraph/hedera-services/blob/15285-blk-rec-translator/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/translators/impl/TokenDissociateTranslator.java#L42); we need to keep up-to-date on what associations exist so we can classify automatic token associations correctly.

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
```diff
@@ -122,9 +121,6 @@ public void setNonce(final long contractNumber, final long nonce) {
                     context.payer(),
                     HandleContext.ConsensusThrottling.ON);
             childRecordBuilder.memo(LAZY_CREATION_MEMO);
-            if (childRecordBuilder.status() == SUCCESS) {
```
lukelee-sl, line 125: I see the recordBuilder's evm address will be set when dispatching the synthetic transactions above.

lukelee-sl (approved):
Smart contracts module changes LGTM.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java
tinker-michaelj: True, this information is already available in the synthetic transaction. 💯

PR #15302, opened 2024-09-03
feat: Remove default memo for lazy created accounts and auto created accounts

### LINKED ISSUE #14064: Update memo for "lazy-created account" to "auto-created account"
### Problem

Memos appended to auto created accounts, per the [HIP-583](https://hips.hedera.com/hip/hip-583) spec, are incorrectly labeled as "lazy-created" whereas the spec calls this "auto-created". 

Here's an example: https://hashscan.io/mainnet/account/0.0.6271006?ph=1&pt=1&p1=1&k1=1719413387.478646003

### Solution

Update memo on transactions which auto-create an account. 

from: `lazy-created account`
to: `auto-created account`

### Alternatives

_No response_

netopyr:
**Description**:

This PR removes the default memos for lazy-created accounts and auto-created accounts.

**Related issue(s)**:

Fixes #14064


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/utils/SynthTxnUtilsTest.java
```diff
@@ -44,6 +43,9 @@
 import org.junit.jupiter.api.Test;
 
 class SynthTxnUtilsTest {
+
+    private static final String LAZY_CREATION_MEMO = "";
```
thomas-swirlds-labs, line 47: This static variable may not be required any more ?

netopyr: It is used in the test below. It replaces the static field that was previously imported.

thomas-swirlds-labs (dismissed):
LGTM! Thanks @netopyr

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/utils/SynthTxnUtilsTest.java
Nana-EC, line 47: nit: similar to Thomas wondering if it's worth keeping this around.
If there needs to be a static empty memo field I'd imagine it already exists and would be applicable across services.
For instance do we check for empty token memo in this way and if so why not use that field, or do we do a memo.isEmpty().
Non blocking food for thought for a future refactor.

Nana-EC (approved):
LG, minor consideration for the future noted

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/utils/SynthTxnUtilsTest.java
netopyr: Good question. I don't think we have an empty memo field, because this is the default if you do not set anything. I think most tests do not check default values of newly created objects. Definitely makes sense to check and fix.

PR #15322, opened 2024-09-03
chore: Use 0.54 protobufs on release branch

mhess-swl:
Setting the release branch to use the specific protobufs 0.54 tag, so as not to depend on the head of `main`.


mhess-swl:
Opened to wrong branch

PR #15324, opened 2024-09-03
chore: Integrate protobufs to build

tinker-michaelj:
**Description**:
 - Integrates `hedera-services` repo to build.
 - Fixes out-of-sequence `h5` tag usage in one file which causes errors in generated javadoc.


hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -85,7 +85,7 @@ message BlockStreamInfo {
      * The latest available hash SHALL be for block N-1.<br/>
      * This is REQUIRED to implement the EVM `BLOCKHASH` opcode.
      * <p>
-     * <h5>Field Length</h5>
+     * <b>Field Length</b>
```
jsync-swirlds, line 88: This is not correct.  We'll need to figure out how to make javadoc relax about the header tag, as a simple bold doesn't produce nearly the correct output in markdown (which is what these are for; we shouldn't even be running javadoc on the PBJ generated files).

hapi/build.gradle.kts
```diff
@@ -29,27 +29,19 @@ tasks.withType<JavaCompile>().configureEach {
...
+            srcDir(layout.projectDirectory.dir("hedera-protobufs/platform"))
         }
         proto {
-            srcDir(tasks.cloneHederaProtobufs.flatMap { it.localCloneDirectory.dir("services") })
-            srcDir(tasks.cloneHederaProtobufs.flatMap { it.localCloneDirectory.dir("streams") })
-            srcDir(tasks.cloneHederaProtobufs.flatMap { it.localCloneDirectory.dir("platform") })
-            srcDir(tasks.cloneHederaProtobufs.flatMap { it.localCloneDirectory.dir("block") })
+            srcDir(layout.projectDirectory.dir("hedera-protobufs/services"))
+            srcDir(layout.projectDirectory.dir("hedera-protobufs/streams"))
+            srcDir(layout.projectDirectory.dir("hedera-protobufs/block"))
```
jsync-swirlds, line 43: This line might need to wait a bit.
The block stream HIP isn't even published yet, and block streams protobufs may change substantially.

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
jsync-swirlds: Javadoc, very incorrectly, requires headings be _nested_ and in order (starting from 2).
This can be replaced with the following for now (yes, it's a bit ugly, but so is javadoc's standard doclet):
```suggestion
     * <div style="display: block;font-size: .83em;margin-top: 1.67em;margin-bottom: 1.67em;margin-left: 0;margin-right: 0;font-weight: bold;">Field Length</div>
```

hapi/build.gradle.kts
tinker-michaelj: Makes sense...for now, since Services is aiming to ship preview block streams to previewnet in `0.54`, for convenience can we keep these here until their final destination is known?

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Awesome. 💯

hapi/build.gradle.kts
jsync-swirlds: We likely will need to, as I don't know when we'll have a solution for packaging protobufs from other repos.
We may have a few mainnet releases between now and when we all work out a functional solution for the need to split ownership of protobufs among multiple code repos.

jsync-swirlds (approved):
Approved from an integration standpoint.
Other notified teams should _definitely_ review this change prior to merge, as it can have fairly broad implications across teams.
platform-base and platform-data, in particular should review this PR.

Neeharika-Sompalli (approved):
LGTM ! tyvm @tinker-michaelj

edward-swirldslabs (approved):

rbarkerSL (approved):

PR #15338, opened 2024-09-04
chore: Record cache updates for Block Streams

### LINKED ISSUE #14428: Converting Record Cache (TransactionID map) froms values being records to values being list of BlockItems

derektriley:
**Description**:

Adds a method `addBlockItems` to `HederaRecordCache` which will take the BlockItem's (that may include including preceding transactions, user transactions, and child transactions), convert them to `SingleTransactionRecord`'s using existing translators which have been moved and add the records to the cache. This allows for queries from `CryptoGetAccountRecordsHandler, NetworkTransactionGetReceiptHandler,  NetworkTransactionGetRecordHandler` to be answered properly if we were to only produce the block stream.

**Related issue(s)**:

Fixes #14428 

**Notes for reviewer**:

I had thought about moving the translators to hapi-utils but it would introduce a few circular dependencies. Let me know if anyone has other thoughts on where to move the translators.

**Checklist**

Signed-off-by: Derek Riley <derek.riley@swirldslabs.com>Signed-off-by: Derek Riley <derek.riley@swirldslabs.com>


hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/translators/BlockTransactionalUnitTranslator.java
```diff
@@ -0,0 +1,164 @@
...
+    public List<SingleTransactionRecord> translate(@NonNull final BlockTransactionalUnit unit) {
+        requireNonNull(unit);
+        baseTranslator.prepareForUnit(unit);
+        final List<StateChange> remainingStateChanges = new LinkedList<>(unit.stateChanges());
+        final List<SingleTransactionRecord> translatedRecords = new ArrayList<>();
+        for (final var blockTransactionParts : unit.blockTransactionParts()) {
+            final var translator = translators.get(blockTransactionParts.functionality());
+            if (translator == null) {
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.warn(
```
derektriley, line 152: Perhaps we should throw an exception here, if no translator is known for the `HederaFunctionality` then we would be unable to add the approriate `TransactionRecord`'s to the cache.

PR #15340, opened 2024-09-04
fix: allow absent agreement keys on disk and in state

### LINKED ISSUE #15332: remove dependence on pre-existence of the public agreement key in key loader.
This line of code requires the existence of the public agreement key on all nodes that do not have the private agreement key. The mTLS handshaking will transfer the public agreement key. Having it ahead of time and storing it on disk is unnecessary.

remove this line of code.

https://github.com/hashgraph/hedera-services/blob/3f7390a8862073f9aef8ed2f6b25ce87efc6aa99/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java#L401

Add unit tests to validate the public keys are not needed.

edward-swirldslabs:
**Description**:
Dynamic Address Book allows for the rotation of the RSA gossip signing key.  This key functions as the root certificate authority for mutual TLS on the network connection between peer nodes.  Nodes will share their public agreement X509 certificate with each other during mTLS handshaking and do not need to share this key ahead of time. The particulars of the agreement key do not matter so long as the signing key is its root certificate. 

In the DAB implementation, the signing key is being rotated but not the agreement key.  The bug this PR fixes is that after a software upgrade changes the signing key, the loaded agreement key from disk has the old signing key as the root certificate.   This was causing mTLS errors.  

The code change in this PR eliminates the need for the agreement key to be loaded from disk and stops serializing the agreement key in the platform address book.   The agreement keys may be present, but they are only loaded for local nodes.  If the agreement keys are absent for a local node, a new agreement key is created with the gossip signing key as its root certificate.  

**Related issue(s)**:

Fixes #15332 

**Notes for reviewer**:

This code change was developed and tested locally using SOLO from genesis with no agreement key material present.  Because the code behavior is keyed off of `isLocal()` on the network stack, it is impossible to test this code change with unit tests or HAPI tests where all nodes are local.


JeffreyDallas (approved):

iwsimon (dismissed):

jeromy-cannon (approved):

lpetrovic05 (dismissed):

edward-swirldslabs:
I updated the unit test to pass under these new conditions and modified the address book to contain 1 non-local address.

JeffreyDallas (approved):

lpetrovic05 (approved):

iwsimon (approved):

Issue #15301, opened 2024-09-03
Extend test AIRDROP - 27 with max transfers, max tokens, max receivers, max fee collectors so that we ensure we won't hit the transaction limit


anastasiya-kovaliova:
### Problem

Extend the following test:
(Multiple Tokens) Airdrop 10 tokens, each with max layers of custom fees (that should be paid by the sender), each airdrop to a differrent account, each fee should have a different fee collector account (no airdrops to fee collectors)

### Solution

Add a HAPI test

PR #15348, opened 2024-09-05
test: New HAPI test for TokenAirdrop transaction

Evdokia-Georgieva:
**Description**:
This PR adds one test case to TokenAirdropTest.java

The test case extends AIRDROP_27 with max transfers, max tokens, max receivers, max fee collectors in order to ensure that the transaction limit is not reached.

This PR also modifies TokenAirdropBase.java -> createAccountsAndTokensWithAllCustomFees() is added

**Test Plan:**
https://www.notion.so/TokenAirdropTransaction-Test-Plan-919ed1800ee845378c5c625648d05ebc?pvs=4

**Resolves**
1. https://github.com/hashgraph/hedera-services/issues/15301


tinker-michaelj (changes requested):
Very systematic treatment @Evdokia-Georgieva , nice work!

Our [style guide](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/README.md#style-guide) for `@HapiTest`'s does recommend splitting large tests into multiple smaller tests that focus on very specific scenarios. 🙏

Evdokia-Georgieva:
> Very systematic treatment @Evdokia-Georgieva , nice work!
> 
> Our [style guide](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/README.md#style-guide) for `@HapiTest`'s does recommend splitting large tests into multiple smaller tests that focus on very specific scenarios. 🙏

@tinker-michaelj in my latest commit I did my best to change the test according to the style guide, unfortunately I could not strictly fulfil the rule for maximum of 7 operations per test. I had to include a little bit more checks as in the test case I need to perform airdrop transaction with the maximum allowed 10 transfers and each transfer needs to have different fee collector account. 
In order to shorten the test, I removed the checks for the receiver balances as they only confirm that the receiver still has none of the airdropped tokens. I assume that as we confirm that the pending airdrops are created then the receiver balance will be correct, but I think that it is important to check the fee collectors account balances with the transferred fees and this adds few more operations. 
I hope that for this case we can make an exception and allow it to have these few more operations. Please let me know if you agree or I should find another solution.
Thank you in advance for your help!

kimbor (approved):
lgtm; I don't see a way to split this large test up into smaller tests, as the whole point of this test is to test a multiple transfer scenario

tinker-michaelj (approved):
LGTM, tyvm @Evdokia-Georgieva !

JivkoKelchev (approved):

JivkoKelchev:
LGTM, thank you @Evdokia-Georgieva

Neeharika-Sompalli (approved):
LGTM! Thanks @Evdokia-Georgieva

vtronkov (approved):

PR #15364, opened 2024-09-05
chore: cleanup exception behavior and add javadocs

lukelee-sl:
**Description**:
Clean up exception handling behavior and add javadocs


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/getevmaddressalias/EvmAddressAliasCall.java
```diff
@@ -61,7 +62,7 @@ public EvmAddressAliasCall(@NonNull final HasCallAttempt attempt, final Address
         final var accountNum = numberOfLongZero(explicitAddress);
         final var account = enhancement.nativeOperations().getAccount(accountNum);
         // If the account is null or does not have an account id then return bail
-        if (account == null || !account.hasAccountId()) {
+        if (account == null) {
```
david-bakin-sl, line 65: `// If the account is null or does not have an account id then return bail` -> `// if the account is null then tell user there's no such account` or something like that

david-bakin-sl (approved):
LGTM (with one code comment to possibly fix).

(Offline discussion for clarification here: In the cases where tests were _eliminated_ for presence of account num, account id: Those are _impossible to happen_ cases where we should _fail fast_ (e.g., w/ NPE) if in fact they happen.)

PR #14026, opened 2024-06-25
feat: move Google protobuf generated code to test fixtures scope

jjohannes:
**Description**:

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #15374, opened 2024-09-06
build: grpc update / avoid mergeJar

jjohannes:
**Description**:

Right now, we need to merge these two Jars because of the overlapping `org.javax.annotations` package:
- `com.google.code.findbugs:jsr305` - this is needed at annotation processing time, because is is a transitive dependency of libraries we currently use (like Guava) that uses annotation from there.
- `javax.annotation:javax.annotation-api` - this is only needed because the protobuf GRPC plugin (NOT our own PBJ) is using `javax.annotation.Generated`. This PR remove this!

By updating Protobuf GRPC to the latest version, we can omit using the annotation by `option("@generated=omit")`.

See:
- https://github.com/grpc/grpc-java/issues/9179
- https://github.com/DanielLiu1123/grpc-starter/pull/54/files#diff-b6f0dd6ddbdb3c9e619b2f7e79181d562aa8379c06880bf5257b629bf8c7109c
- https://github.com/grpc/grpc-java/pull/11086

**Note:** This generated code is intended to eventually only be used for testing. Right now, there are still some references in the production code that needs adjustment. See: https://github.com/hashgraph/hedera-services/pull/14026

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
@@ -74,8 +74,6 @@ jvmDependencyConflicts {
     }
 }
 
-configurations.javaModulesMergeJars { extendsFrom(configurations["internal"]) }
-
 tasks.buildDependents { setGroup(null) }
 
 tasks.buildNeeded { setGroup(null) }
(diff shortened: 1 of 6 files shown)


edward-swirldslabs (approved):

rbarkerSL (approved):
LGTM

PR #15387, opened 2024-09-07
chore: implement and test indirect block proofs

### LINKED ISSUE #15379: Test production of indirect block proofs
### Problem

If it happens that the ledger signature for block `N` becomes available before the signature for blocks `N-2` and `N-1`, we want to use the signature on block `N` to then provide "indirect" proofs for the preceding blocks using the `BlockProof#sibling_hashes` list.

This functionality should be tested "end-to-end" via a `@HapiTest`. 

### Solution

- [x] Add a placeholder `TssBaseService` to `hedera-app` based on the API [here](https://github.com/hashgraph/hedera-services/pull/15160).
- [x] Bind a placeholder `TssBaseService` implementation that does the same simulated async signing that is currently in `BlockStreamManagerImpl` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java#L202).
    * Inject this implementation into the `Hedera` constructor via a `Supplier<TssBaseService>`.
- [x] In `AbstractEmbeddedHedera`, inject a mock `TssBaseService` whose behavior can be configured by a `@HapiTest` author.
- [x] Add `SpecOperation` helpers to easily configure the behavior of this mock at test execution time.
- [x] Add block stream equivalents to e.g. `streamMustIncludePassFrom(Function<HapiSpec, RecordStreamAssertion>)`.
- [x] Add a `@RepeatableHapiTest` that configures the mock `TssBaseService` to skip scheduling signatures for two blocks before then signing the next one; and asserts the resulting `BlockProof`s look as expected.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15379


Neeharika-Sompalli (approved):
Makes sense to me @tinker-michaelj ! Thanks

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/tss/RepeatableTssTests.java
```diff
@@ -0,0 +1,104 @@
...
+                    final var items = block.items();
+                    final var proofItem = items.getLast();
+                    assertTrue(proofItem.hasBlockProof(), "Block proof is expected as the last item");
+                    final var proof = proofItem.blockProofOrThrow();
+                    if (remainingIndirectionLevels.get() == 0) {
+                        assertTrue(
+                                proof.siblingHashes().isEmpty(),
+                                "No sibling hashes should be present on a direct proof");
+                        return true;
+                    } else {
```
mhess-swl, line 83: I wonder about hiding this complexity in a modified StreamFileAlterationListener, one that keeps remainingIndirectionLevels, to make this more approachable. On the other hand, I'm not sure what the consequences of stuffing assertions into a Listener would be yet. Anyway, just a thought.

tinker-michaelj: Agreed this isn't readable; extracted to a top-level `IndirectProofsAssertion` that might be reused in places.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/fakes/FakeTssBaseService.java
```diff
@@ -0,0 +1,106 @@
...
+public class FakeTssBaseService implements TssBaseService {
+    private static final Logger log = LogManager.getLogger(FakeTssBaseService.class);
+
+    /**
+     * Copy-on-write list to avoid concurrent modification exceptions if a consumer unregisters
+     * itself in its callback.
+     */
+    private final List<BiConsumer<byte[], byte[]>> consumers = new CopyOnWriteArrayList<>();
+
+    private boolean ignoreRequests = false;
```
thomas-swirlds-labs, line 52: could change to AtomicBoolean to make it fully thread-safe

thomas-swirlds-labs (approved):
LGTM. Thanks @tinker-michaelj 🎉

edward-swirldslabs (approved):
LGTM

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/fakes/FakeTssBaseService.java
tinker-michaelj: Ah I should document why this isn't thread-safe; it should only be used in `@RepeatableHapiTest`'s, which disable all concurrent execution.

Will do in follow-up PR. 👍

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/BlockStreamAssertion.java
```diff
@@ -0,0 +1,49 @@
...
+public interface BlockStreamAssertion {
+    /**
+     * Updates the assertion's state based on a relevant {@link Block}, throwing an {@link AssertionError} if a
+     * failure state is reached; or returning true if the assertion has reached a success state.
+     *
+     * @param block the block to test
+     * @throws AssertionError if the assertion has failed
+     * @return true if the assertion has succeeded
+     */
+    default boolean test(@NonNull final Block block) throws AssertionError {
```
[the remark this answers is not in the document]
tinker-michaelj: True, the explicit `BlockStreamAssertion#toString()` override doesn't add much...will change in a follow-up PR.

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #15389, opened 2024-09-07
fix: 15385 Used `MerkleStateRoot.getReadablePlatformState` where possible to prevent race conditions

### LINKED ISSUE #15385: Unable to get `PlatformStateAccessor` due to race condition
### Description

In some cases `MerkleStateRoot.getPlatformState` fails with the following exception

```
node0	14m 50.100s	ERROR	AbstractTaskSchedulerBuilder:	Uncaught exception in scheduler ConsensusEventStream
com.swirlds.base.state.MutabilityException: This operation is not permitted on an immutable object.
	at com.swirlds.base.state.Mutable.throwIfImmutable(Mutable.java:55) ~[swirlds-base-0.54.0.jar:?]
	at com.swirlds.base.state.Mutable.throwIfImmutable(Mutable.java:46) ~[swirlds-base-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.getWritableStates(MerkleStateRoot.java:354) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.writablePlatformStateStore(MerkleStateRoot.java:1034) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.getPlatformState(MerkleStateRoot.java:1004) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.SwirldStateManager.isInFreezePeriod(SwirldStateManager.java:216) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.builder.PlatformComponentBuilder.lambda$buildConsensusEventStream$2(PlatformComponentBuilder.java:591) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.event.stream.DefaultConsensusEventStream.lambda$addEvents$0(DefaultConsensusEventStream.java:225) ~[swirlds-platform-core-0.54.0.jar:?]
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) ~[?:?]
```

This happens despite the fact that `getPlatformState` is implemented as follows:
```
    @Override
    public PlatformStateAccessor getPlatformState() {
        return !isImmutable() ? writablePlatformStateStore() : readablePlatformStateStore();
    }
```

This exception suggests that the state was updated to be immutable by some other state. In this particular case, it doesn't even have to be writable because ultimately, we only need to read two parameters: `platformState.getFreezeTime()` and `platformState.getLastFrozenTime()`.

To fix this, we need to use the read-only State API where possible instead of accessing it via the ambiguous `PlatformStateAccessor`.

### Steps to reproduce

This test can be reproduced with `Crypto-Update-Jar-1.5k-25m`.

### Additional context

_No response_

### Hedera network

other

### Version

v0.54

### Operating system

None

imalygin:
**Description**:

This PR prevents possible error that can happen if one thread gets an instance of writable `PlatformStateAccessor` while another thread creates a copy of the state which makes usage of the writable state invalid which ultimately results in an exception: 

```
node0	14m 50.100s	ERROR	AbstractTaskSchedulerBuilder:	Uncaught exception in scheduler ConsensusEventStream
com.swirlds.base.state.MutabilityException: This operation is not permitted on an immutable object.
	at com.swirlds.base.state.Mutable.throwIfImmutable(Mutable.java:55) ~[swirlds-base-0.54.0.jar:?]
	at com.swirlds.base.state.Mutable.throwIfImmutable(Mutable.java:46) ~[swirlds-base-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.getWritableStates(MerkleStateRoot.java:354) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.writablePlatformStateStore(MerkleStateRoot.java:1034) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.MerkleStateRoot.getPlatformState(MerkleStateRoot.java:1004) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.state.SwirldStateManager.isInFreezePeriod(SwirldStateManager.java:216) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.builder.PlatformComponentBuilder.lambda$buildConsensusEventStream$2(PlatformComponentBuilder.java:591) ~[swirlds-platform-core-0.54.0.jar:?]
	at com.swirlds.platform.event.stream.DefaultConsensusEventStream.lambda$addEvents$0(DefaultConsensusEventStream.java:225) ~[swirlds-platform-core-0.54.0.jar:?]
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) ~[?:?]
```

**Related issue(s)**:

Fixes #15385 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
@@ -236,7 +236,8 @@ public class StateChangesValidator implements BlockStreamValidator {
         final var currentVersion = new ServicesSoftwareVersion(servicesVersion, configVersion);
         final var lifecycles = newPlatformInitLifecycle(bootstrapConfig, currentVersion, migrator, servicesRegistry);
         state = new MerkleStateRoot(lifecycles, version -> new ServicesSoftwareVersion(version, configVersion));
-        state.getPlatformState();
+        // initialize the platform state
+        state.getReadablePlatformState();
         migrator.doMigrations(
                 state,
                 servicesRegistry,
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
@@ -237,7 +237,7 @@ public class StateChangesValidator implements BlockStreamValidator {
         final var lifecycles = newPlatformInitLifecycle(bootstrapConfig, currentVersion, migrator, servicesRegistry);
         state = new MerkleStateRoot(lifecycles, version -> new ServicesSoftwareVersion(version, configVersion));
         // initialize the platform state
-        state.getReadablePlatformState();
+        state.getWritablePlatformState();
         migrator.doMigrations(
                 state,
                 servicesRegistry,
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
@@ -379,6 +379,7 @@ public final class EventRecoveryWorkflow {
         final PlatformEvent lastEvent = ((CesEvent) getLastEvent(round)).getPlatformEvent();
         new DefaultEventHasher().hashEvent(lastEvent);
 
+        final PlatformStateAccessor newReadablePlatformState = newState.getReadablePlatformState();
         final PlatformStateAccessor newWritablePlatformState = newState.getWritablePlatformState();
         final PlatformStateAccessor previousReadablePlatformState =
                 previousState.get().getState().getReadablePlatformState();
@@ -402,10 +403,9 @@ public final class EventRecoveryWorkflow {
         final boolean isFreezeState = isFreezeState(
                 previousState.get().getConsensusTimestamp(),
                 currentRoundTimestamp,
-                newState.getReadablePlatformState().getFreezeTime());
+                newReadablePlatformState.getFreezeTime());
         if (isFreezeState) {
-            newState.getWritablePlatformState()
-                    .setLastFrozenTime(newState.getReadablePlatformState().getFreezeTime());
+            newWritablePlatformState.setLastFrozenTime(newReadablePlatformState.getFreezeTime());
         }
 
         final ReservedSignedState signedState = new SignedState(
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
@@ -153,7 +153,7 @@ public final class EventRecoveryWorkflow {
                         platformContext, signedStateFile, SignedStateFileUtils::readState)
                 .reservedSignedState()) {
             StaticSoftwareVersion.setSoftwareVersion(
-                    initialState.get().getState().getPlatformState().getCreationSoftwareVersion());
+                    initialState.get().getState().getReadablePlatformState().getCreationSoftwareVersion());
 
             logger.info(
                     STARTUP.getMarker(),
@@ -314,7 +314,7 @@ public final class EventRecoveryWorkflow {
                 .init(
                         platform,
                         InitTrigger.EVENT_STREAM_RECOVERY,
-                        initialState.get().getState().getPlatformState().getCreationSoftwareVersion());
+                        initialState.get().getState().getReadablePlatformState().getCreationSoftwareVersion());
 
         appMain.init(platform, platform.getSelfId());
 
@@ -379,32 +379,32 @@ public final class EventRecoveryWorkflow {
         final PlatformEvent lastEvent = ((CesEvent) getLastEvent(round)).getPlatformEvent();
         new DefaultEventHasher().hashEvent(lastEvent);
 
-        final PlatformStateAccessor platformState = newState.getPlatformState();
+        final PlatformStateAccessor platformState = newState.getReadablePlatformState();
 
         platformState.bulkUpdate(v -> {
             v.setRound(round.getRoundNum());
             v.setLegacyRunningEventHash(getHashEventsCons(
-                    previousState.get().getState().getPlatformState().getLegacyRunningEventHash(), round));
+                    previousState.get().getState().getReadablePlatformState().getLegacyRunningEventHash(), round));
             v.setConsensusTimestamp(currentRoundTimestamp);
             v.setSnapshot(SyntheticSnapshot.generateSyntheticSnapshot(
                     round.getRoundNum(), lastEvent.getConsensusOrder(), currentRoundTimestamp, config, lastEvent));
             v.setCreationSoftwareVersion(
-                    previousState.get().getState().getPlatformState().getCreationSoftwareVersion());
+                    previousState.get().getState().getReadablePlatformState().getCreationSoftwareVersion());
         });
 
         applyTransactions(
                 previousState.get().getSwirldState().cast(),
                 newState.getSwirldState().cast(),
-                newState.getPlatformState(),
+                newState.getWritablePlatformState(),
                 round);
 
         final boolean isFreezeState = isFreezeState(
                 previousState.get().getConsensusTimestamp(),
                 currentRoundTimestamp,
-                newState.getPlatformState().getFreezeTime());
+                newState.getReadablePlatformState().getFreezeTime());
         if (isFreezeState) {
-            newState.getPlatformState()
-                    .setLastFrozenTime(newState.getPlatformState().getFreezeTime());
+            newState.getWritablePlatformState()
+                    .setLastFrozenTime(newState.getReadablePlatformState().getFreezeTime());
         }
 
         final ReservedSignedState signedState = new SignedState(
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
@@ -379,17 +379,18 @@ public final class EventRecoveryWorkflow {
         final PlatformEvent lastEvent = ((CesEvent) getLastEvent(round)).getPlatformEvent();
         new DefaultEventHasher().hashEvent(lastEvent);
 
-        final PlatformStateAccessor platformState = newState.getReadablePlatformState();
+        final PlatformStateAccessor newWritablePlatformState = newState.getWritablePlatformState();
+        final PlatformStateAccessor previousReadablePlatformState =
+                previousState.get().getState().getReadablePlatformState();
 
-        platformState.bulkUpdate(v -> {
+        newWritablePlatformState.bulkUpdate(v -> {
             v.setRound(round.getRoundNum());
-            v.setLegacyRunningEventHash(getHashEventsCons(
-                    previousState.get().getState().getReadablePlatformState().getLegacyRunningEventHash(), round));
+            v.setLegacyRunningEventHash(
+                    getHashEventsCons(previousReadablePlatformState.getLegacyRunningEventHash(), round));
             v.setConsensusTimestamp(currentRoundTimestamp);
             v.setSnapshot(SyntheticSnapshot.generateSyntheticSnapshot(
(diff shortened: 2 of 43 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
```diff
@@ -236,7 +236,8 @@ public StateChangesValidator(
         final var currentVersion = new ServicesSoftwareVersion(servicesVersion, configVersion);
         final var lifecycles = newPlatformInitLifecycle(bootstrapConfig, currentVersion, migrator, servicesRegistry);
         state = new MerkleStateRoot(lifecycles, version -> new ServicesSoftwareVersion(version, configVersion));
-        state.getPlatformState();
+        // initialize the platform state
+        state.getReadablePlatformState();
```
tinker-michaelj, line 240: ⚠️ This needs to be `getWritablePlatformState()` so the validator initializes its platform state in the same relative order as the nodes in the test network.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1005,8 +1005,20 @@ public SwirldState getSwirldState() {
      */
     @NonNull
     @Override
-    public PlatformStateAccessor getPlatformState() {
-        return !isImmutable() ? writablePlatformStateStore() : readablePlatformStateStore();
+    public PlatformStateAccessor getReadablePlatformState() {
```
tinker-michaelj, line 1008: How much work would it be to make this method return `ReadablePlatformStateStore` and `getWritable()` below return `WritablePlatformStateStore`?

imalygin: `PlatformStateAccessor` is an abstraction layer that hides the implementation details of the platform state. This interface is implemented by two classes - `com.swirlds.platform.state.State` and `com.swirlds.platform.state.MerkleStateRoot`. In case of the latter, the platform state is backed by `ReadablePlatformStateStore` and `WritablePlatformStateStore`, in case of the former it is not. 

That is, the question is not in the amount of work. The question is in the possibility of this option. 

I also considered a different approach (as we discussed) - accessing the platform state via `com.swirlds.state.State` -> `com.swirlds.state.spi.ReadableStates#getSingleton`. But the solution turned out to be rather ugly. In each case I have to explicitly case and instance of `MerkleStateRoot` to `com.swirlds.state.State` because throughout the code we work with via `com.swirlds.platform.state.MerkleRoot` interface. 

I thought that introducing two methods `getReadablePlatformState` and `getWritablePlatformState` to `PlatformStateAccessor` is a lesser evil. However, I'm open to be persuaded otherwise.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
imalygin: Great! Thank you for the tip, I fixed that.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
tinker-michaelj: Ah good point. 💯

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -244,7 +253,7 @@ public String getInfoString(final int hashDepth) {
     @Override
     public String toString() {
         return new ToStringBuilder(this)
-                .append("platformState", getPlatformState())
+                .append("platformState", getWritablePlatformState())
```
jsync-swirlds, line 256: Why is a `toString` method accessing _writable_ state?  Should this be a `readable` state?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -33,12 +33,23 @@ public interface MerkleRoot extends MerkleInternal {
...
-     * Get the platform state.
+     * Get readable platform state.
+     * Works on both - mutable and immutable {@link MerkleRoot} and, therefore, this method should be preferred.
      *
-     * @return the platform state
+     * @return immutable platform state
      */
     @NonNull
-    PlatformStateAccessor getPlatformState();
+    PlatformStateAccessor getReadablePlatformState();
```
jsync-swirlds, line 42: Just a thought (possibly a future PR?).
Would it make sense to split the platform state accessor interface?
`PlatformStateAccessor` might be the `get` methods, and `PlatformStateMutator extends PlatformStateAccessor` could then add the `set` methods.
That way we get more compiler support when code inadvertently tries to change a "readable" state.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -139,14 +139,23 @@ public void setSwirldState(final SwirldState state) {
...
     }
 
+    /**
+     * Immutable platform state is not supported by this class.
+     */
+    @NonNull
+    @Override
+    public PlatformStateAccessor getReadablePlatformState() {
+        throw new UnsupportedOperationException(
+                "com.swirlds.platform.state.State doesn't support read-only mode for PlatformState");
```
jsync-swirlds, line 149: This seems surprising.  Why wouldn't `State` allow a read-only platform state (even if backed by a writable platform state)?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManagerUtils.java
```diff
@@ -52,7 +52,7 @@ public static MerkleRoot fastCopy(
 
         // Create a fast copy
         final MerkleRoot copy = state.copy();
-        final var platformState = copy.getPlatformState();
+        final var platformState = copy.getWritablePlatformState();
```
jsync-swirlds, line 55: Another idea for a future PR:
Split fast copy to readable/writable versions.  I have the impression (which may be incorrect) that a lot of fast copies are for purely read-only purposes, so perhaps adding a read-only version of this method would enable "safer" usage in those cases.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/TransactionHandler.java
```diff
@@ -56,7 +56,7 @@ public void handleRound(final ConsensusRound round, final MerkleRoot state) {
             final Instant timeOfHandle = Instant.now();
             final long startTime = System.nanoTime();
 
-            state.getSwirldState().handleConsensusRound(round, state.getPlatformState());
+            state.getSwirldState().handleConsensusRound(round, state.getReadablePlatformState());
```
jsync-swirlds, line 59: Perhaps I'm missing something, but wouldn't handling a consensus round require a _writable_ state?

jsync-swirlds (commented):
Mostly some questions where the code might have possible improvements, or it's a bit confusing why we chose readable or writable.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: It's mostly because `getReadablePlatformState` throws an exception in this particular class. However, I agree - it looks confusing, so I changed implementation of `getReadablePlatformState`. Now it has exactly the same implementation as `getWritablePlatformState`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: Yes, you're right. I'll change it

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Sure, here it is https://github.com/hashgraph/hedera-services/issues/15413

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManagerUtils.java
imalygin: Not sure if I fully understand how this change is supposed to look like. Could you please elaborate?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/TransactionHandler.java
imalygin: Surprisingly enough, it's nod needed for consensus itself. However, you intuition is correct in case of tools. I'll update it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/BirthRoundStateMigration.java
```diff
@@ -64,7 +64,7 @@ public static void modifyStateForBirthRoundMigration(
         }
 
         final MerkleRoot state = initialState.getState();
-        final PlatformStateAccessor platformState = state.getPlatformState();
+        final PlatformStateAccessor platformState = state.getWritablePlatformState();
```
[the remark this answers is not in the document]
imalygin: It would. Fixed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/recovery/EventRecoveryWorkflow.java
```diff
@@ -379,32 +379,32 @@ private static ReservedSignedState handleNextRound(
...
         final boolean isFreezeState = isFreezeState(
                 previousState.get().getConsensusTimestamp(),
                 currentRoundTimestamp,
-                newState.getPlatformState().getFreezeTime());
+                newState.getReadablePlatformState().getFreezeTime());
         if (isFreezeState) {
-            newState.getPlatformState()
-                    .setLastFrozenTime(newState.getPlatformState().getFreezeTime());
+            newState.getWritablePlatformState()
+                    .setLastFrozenTime(newState.getReadablePlatformState().getFreezeTime());
```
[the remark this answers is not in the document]
imalygin: Sure. Done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManagerUtils.java
jsync-swirlds: Basically, create a `fastImmutableCopy` and a `fastMutableCopy` (perhaps `fastCopy` calls `fastMutableCopy` method initially).
Change calls, where appropriate, to call the `fastImmutableCopy` whenever possible so the calling component obtains an immutable state rather than a mutable one.

Not sure if that actually works, but it _seems_ like a possibly useful approach to reduce the number of mutable copies in the system.

tinker-michaelj (approved):
LGTM, tyvm @imalygin !

I feel like we should circle back and find the exact points in Platform lifecycle where we must guarantee the `PlatformState` state definition is initialized in `MerkleRoot`, and have a method that does that explicitly, instead of relying on the side effects of calling `getWritablePlatformState()`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManagerUtils.java
jsync-swirlds: Ahh, thanks.  I was not aware the original object becomes immutable.
In that case, you're correct and my suggestion is moot.

jsync-swirlds (approved):

mxtartaglia-sl (approved):

anthony-swirldslabs (approved):

kfa-aguda (approved):
lgtm. Mostly straightforward changes making the context of platformstate explicit, which is awesome.

PR #15392, opened 2024-09-08
chore: default `TransactionExecutor` simulator to no-op system contract authorization checks

### LINKED ISSUE #15391: Support constructing a standalone `TransactionExecutor` with custom system contract authorization strategy
### Problem

For some Mirror Node use cases, we want to simulate contract transactions while assuming all required permissions are available.

The default system contract authorization strategy does not support this, ofc.

### Solution

Support constructing a `ContractServiceImpl` whose default `VerificationStrategies` can be replaced with a custom implementation.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15391


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/ContractServiceImpl.java
```diff
@@ -40,18 +43,21 @@ public class ContractServiceImpl implements ContractService {
...
+            @Nullable final VerificationStrategies verificationStrategies,
+            @Nullable final Supplier<List<OperationTracer>> addOnTracers) {
         requireNonNull(appContext);
         this.component = DaggerContractServiceComponent.factory()
                 .create(
                         appContext.instantSource(),
                         // (FUTURE) Inject the signature verifier instance into the IsAuthorizedSystemContract
                         // C.f. https://github.com/hashgraph/hedera-services/issues/14248
                         appContext.signatureVerifier(),
+                        Optional.ofNullable(verificationStrategies).orElseGet(DefaultVerificationStrategies::new),
```
david-bakin-sl, line 60: (This is a clever way to get a concise syntax, but is it really worth it?)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/standalone/impl/NoopVerificationStrategies.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hyperledger.besu.datatypes.Address;
+
+/**
+ * Implements a no-op {@link VerificationStrategies} for use in standalone contract execution.
+ * The strategy considers all keys to have valid signatures under all conditions.
+ */
+public enum NoopVerificationStrategies implements VerificationStrategies {
+    NOOP_VERIFICATION_STRATEGIES;
+
+    private static final VerificationStrategy NOOP_VERIFICATION_STRATEGY = key -> Decision.VALID;
```
david-bakin-sl, line 33: Should class `VerificationStrategy` have the `@FunctionalInterface` annotation?

david-bakin-sl (approved):
LGTM.  

I guess I'm OK with having `NoopVerificationStrategy` here (instead of the mirror node) because it is pretty much scoped to being an "impl" of a "standalone" workflow.

hedera-node/hedera-app/src/main/java/module-info.java
```diff
@@ -48,6 +48,7 @@
     requires io.netty.transport.classes.epoll;
     requires io.netty.transport;
     requires org.apache.commons.lang3;
+    requires org.hyperledger.besu.datatypes;
```
lukelee-sl, line 51: Why did this have to be added?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/ContractServiceImpl.java
tinker-michaelj: Haha I don't know, I just got used to this pattern for some reason. 🤷

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/standalone/impl/NoopVerificationStrategies.java
tinker-michaelj: Ah yes would be appropriate, I don't think we need any more non-`default` methods on this interface...will add in a future PR. 💯

Neeharika-Sompalli (approved):
LGTM!

PR #15407, opened 2024-09-09
test: Add BlockSignature validations

### LINKED ISSUE #15384: Add `BlockStreamValidator` that validates direct and indirect block proofs in CI PR checks
### Problem

### Solution

Add this validator.

### Alternatives

_No response_

Neeharika-Sompalli:
Fixes #15384 

- Adds validations for `BlockSignature` to `StateChangesValidator`
- The logic is implemented in https://github.com/hashgraph/hedera-services/pull/15387


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/NaiveStreamingTreeHasher.java
```diff
@@ -0,0 +1,84 @@
...
+        final var hasher = new NaiveStreamingTreeHasher();
+        for (final var item : leaves) {
+            hasher.addLeaf(item);
+        }
+        return hasher.rootHash().join();
+    }
+
+    @Override
+    public void addLeaf(@NonNull final Bytes leaf) {
+        if (rootHashRequested) {
```
mhess-swl, line 47: Just checking, `rootHashRequested` doesn't need to be thread-safe, right? Seems like concurrent calls to `addLeaf` _after_ `rootHash()` would be possible in some situations, but I doubt `StateChangesValidator` uses multiple threads

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
```diff
@@ -245,16 +251,28 @@ public StateChangesValidator(
...
     @Override
     public void validateBlocks(@NonNull final List<Block> blocks) {
         logger.info("Beginning validation of expected root hash {}", expectedRootHash);
+        var previousBlockHash = INITIAL_BLOCK_HASH;
+
         for (final var block : blocks) {
+            StateChanges lastStateChanges = null;
+            // get the state hash before applying the state changes from current block
+            //            CRYPTO.digestTreeSync(state);
+            final var startOfBlockStateHash = Bytes.wrap(new byte[48]); // state.getHash();
```
mhess-swl, line 266: Not sure I understand the comment here...meant to be this?
```suggestion
            // (FUTURE) use `state.getHash()` instead of an empty byte array
            final var startOfBlockStateHash = Bytes.wrap(new byte[48]); 
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/NaiveStreamingTreeHasher.java
```diff
@@ -0,0 +1,84 @@
...
+import com.hedera.node.app.blocks.StreamingTreeHasher;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+
+public class NaiveStreamingTreeHasher implements StreamingTreeHasher {
```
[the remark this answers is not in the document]
Neeharika-Sompalli: This is an already existing class Michael added in JMH module. I copied to test-clients @mhess-swl

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/NaiveStreamingTreeHasher.java
Neeharika-Sompalli: No This is already used class in JMH benchmarks. So used the same.

mhess-swl (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
```diff
@@ -246,16 +252,29 @@ public StateChangesValidator(
...
         logger.info("Registered all Service and migrated state definitions to version {}", servicesVersion);
     }
 
     @Override
     public void validateBlocks(@NonNull final List<Block> blocks) {
         logger.info("Beginning validation of expected root hash {}", expectedRootHash);
+        var previousBlockHash = INITIAL_BLOCK_HASH;
+
         for (final var block : blocks) {
+            StateChanges lastStateChanges = null;
```
MiroslavGatsanoga, line 264: As per the issue description we need to validate that `the final state change in the block is a BlockStreamInfo with previous block hash that matches the observed hash for block N-1 in the stream`.

thomas-swirlds-labs (approved):
LGTM. Thanks! @Neeharika-Sompalli 🎉

Neeharika-Sompalli:
Closing this in favor of this PR https://github.com/hashgraph/hedera-services/pull/15425

PR #15417, opened 2024-09-10
feat: migrate event serialization to protobuf

### LINKED ISSUE #15202: Migrate PCES stream to protobuf
Events serialized to PCES should be serialized via protobuf. The PCES file version should be bumped up.

For one release, the code must support both versions for deserializing, but only support serializing the protobuf version. The version number in each PCES file identifies the version of the events it contains, which should be used to determine which method of deserialization to use (this requires the version of PCES file version to be increased). 

It needs to be checked that upon restart, the node does not append to a pre-existing PCES file. Doing so would mix event versions in a single file which will cause problems.

Once this ticket is complete, all events being written to PCES should be serialized as protobuf.

This feature must be migration tested. The MigrationTestingTool test should be sufficient.

### LINKED ISSUE #15205: Migrate event gossiping to protobuf
This ticket is complete when events are serialized in protobuf for gossip. There are many other pieces of data that are sent in gossip that are not being converted to protobuf in this ticket. To find where events are sent, follow the `SyncProtocol` class. The `SyncUtils` class contains the serialization of events for gossip. As part of this ticket, any new safety checks required for protobuf must be considered. There may be new checks that need to be added once protobuf events are gossiped. Checks should include maximum level of object deserialization and maximum field length.

lpetrovic05:
closes #15205
closes #15202


edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
```diff
@@ -586,4 +580,28 @@ private static <T extends SelfSerializable> T registryConstructor(final long cla
...
+     * @param codec the codec to use to parse the record
+     * @param <T>   the type of the record
+     * @return the parsed record
+     * @throws IOException if an IO error occurs
+     */
+    public @NonNull <T extends Record> T readPbjRecord(@NonNull final Codec<T> codec) throws IOException {
+        final int size = readInt();
+        readableSequentialData.limit(readableSequentialData.position() + size);
+        try {
+            return codec.parse(readableSequentialData);
```
anthony-swirldslabs, line 596: Which code is responsible for resetting the limit on the stream? Or the caller is supposed to close the stream after reading a single record from it? This isn't clear from the javadoc, and there's not a `finally{}` block that would undo the limit (or close the stream, for that matter.)

platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventBenchmarks.java
```diff
@@ -96,8 +96,8 @@ public void serializeDeserialize(final Blackhole bh) throws IOException {
         //
         // Benchmark                                (seed)   Mode  Cnt    Score    Error   Units
         // EventSerialization.serializeDeserialize       0  thrpt    3  962.486 ± 29.252  ops/ms
-        EventSerializationUtils.serializePlatformEvent(outStream, event, true);
-        bh.consume(EventSerializationUtils.deserializePlatformEvent(inStream, true));
+        outStream.writePbjRecord(event.getGossipEvent(), GossipEvent.PROTOBUF);
```
anthony-swirldslabs, line 99: Note that the `writePbjRecord` API is currently error-prone if someone passes an incorrect Codec for a given record. PBJ records have a public static final PROTOBUF field that refers to their codec, but given that the `writePbjRecord` accepts a `T extends Record`, one would have to use reflection to fetch the PROTOBUF value.

I understand that using Java reflection may be undesirable (for performance reasons, or complexity), but at the same time this makes it easy to introduce a bug by supplying an incorrect codec to the API.

Any thoughts on this?

mxtartaglia-sl (approved):

platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventBenchmarks.java
lpetrovic05: the method `writePbjRecord()` uses generics to indicate that the instance and codec should be the same type. I would consider this to be sufficient. But if we want to make it safer and simpler, I don't think reflections is the way to go. We should make all PBJ records implement an interface with a method `getProtoCodec()`, that way you get the codec directly from the instance you are serializing.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
lpetrovic05: The limit is set every time this method is called and the stream is never used outside of this method. So no limit reset is necessary. The stream should not be closed because this assumes that there will be a continuous stream from which you can read any number of PBJ records.

mxtartaglia-sl (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
anthony-swirldslabs: So if you call this method once, it reads an integer with the size of the record and then sets a limit on the stream and reads this record.

Suppose you then call this method the second time to read another record from the same stream. The very first call to readInt() will fail now because it will hit a limit set by a previous call to this method.

How is this supposed to be handled?

platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventBenchmarks.java
anthony-swirldslabs: Got it. I think that the type constraints imposed by the generic is a good first step to prevent most of the compile-time errors.

But I'd really like to see a runtime protection as well which may be important when writing a generic code that would use this method to write arbitrary records. I totally like the idea of an interface with a `getProtoCodec()` method! However, I suppose that adding it is outside of scope of this particular PR. I'd appreciate it if you could file an issue against PBJ to implement this interface.

edward-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
```diff
@@ -586,4 +581,32 @@ private static <T extends SelfSerializable> T registryConstructor(final long cla
...
+    /**
+     * Reads a PBJ record from the stream.
+     *
+     * @param codec the codec to use to parse the record
+     * @param <T>   the type of the record
+     * @return the parsed record
+     * @throws IOException if an IO error occurs
+     */
+    public @NonNull <T extends Record> T readPbjRecord(@NonNull final Codec<T> codec) throws IOException {
+        final int size = readInt();
```
artemananiev, line 594: This is a fixed int, 4 bytes. Why not a var int (and probably a protobuf tag, too), to make it a valid length-delimited protobuf field?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataOutputStream.java
```diff
@@ -295,4 +286,22 @@ protected void writeClassIdVersion(final SerializableDet serializable, final boo
...
+     * 		the codec to use to write the record
+     * @param <T>
+     * 		the type of the record
+     * @throws IOException
+     * 		thrown if any IO problems occur
+     */
+    public <T extends Record> void writePbjRecord(@NonNull final T record, @NonNull final Codec<T> codec)
+            throws IOException {
+        writeInt(codec.measureRecord(record));
+        codec.write(record, writableSequentialData);
```
artemananiev, line 305: The length is known, so a check can be added how many bytes have been actually written and compared against the length. A similar check is there when objects are read from input streams.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesBirthRoundMigration.java
```diff
@@ -224,9 +224,9 @@ private static void migrateEvents(
         final Path temporaryFile = LegacyTemporaryFileBuilder.buildTemporaryFile("new-pces-file");
         final SerializableDataOutputStream outputStream = new SerializableDataOutputStream(
```
artemananiev, line 225: There is a tiny step to make this file fully protobuf compatible: field tags. Every event should be written as a length-encoded protobuf object. Another field is needed for the version. This is it

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
lpetrovic05: the call to `readInt` does not go through the `readableSequentialData`, so it is not affected by the limit

platform-sdk/swirlds-platform-core/src/jmh/java/com/swirlds/platform/core/jmh/EventBenchmarks.java
lpetrovic05: https://github.com/hashgraph/pbj/issues/287

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
lpetrovic05: Simple reason, I did not know how to do it 😄 . Are there any examples for this that I can look at?

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataOutputStream.java
lpetrovic05: I would expect that codec would throw if it can't write the whole record, is that not the case?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/event/preconsensus/PcesBirthRoundMigration.java
lpetrovic05: Are there any examples for this I can look at? (might do it in a separate PR)

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
anthony-swirldslabs: Oh, I didn't realize that. Looks good then. Thanks for the clarification!

anthony-swirldslabs (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataOutputStream.java
artemananiev: Such checks (there are a lot in other parts of the repo) are to catch bugs in codecs, when `measureRecord()` and `write()` are not consistent with each other. It's better get an exception on writing than to debug why reading fails much later.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
artemananiev: Reading and writing var ints is as simple as `readableSequentialData.readVarInt(false)` and `writableSequentialData.writeVarInt(value, false)`. Such a change does not make a lot of sense in itself, but if a protobuf tag is written / read before that, it would become a valid length-delimited protobuf message.

To work with tags, a `FieldDefinition` object is needed. Its name can be any, type is `FieldType.BYTES` or `FieldType.MESSAGE`, repeated=true, optional=true, oneof=false, number can be any. Then writing code will look like this:

```java
ProtoWriterTools.writeDelimited(writableSequentialData, fieldDefinition, codec.measureRecord(record), out -> codec.write(record, out));
```

It can be done in a separate PR.

artemananiev (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
lpetrovic05: I created a ticket for it https://github.com/hashgraph/hedera-services/issues/15538

PR #15425, opened 2024-09-10
feat: Use real state hash and add block proof validations

### LINKED ISSUE #15384: Add `BlockStreamValidator` that validates direct and indirect block proofs in CI PR checks

Neeharika-Sompalli:
Fixes #15384 

- Adds validations for `BlockSignature` to `StateChangesValidator`
- The logic is implemented in https://github.com/hashgraph/hedera-services/pull/15387
- Also uses real start of state hash in the `BlockProof`


mhess-swl (approved):

tinker-michaelj (approved):

PR #15432, opened 2024-09-11
chore: testnet event hashing

### LINKED ISSUE #15100: Ensure 0.53 testnet is using the new hashing algorithm
Once testnet gets updated to `0.53`, ensure that the new algorithm is being used and that the hash of an event can be recomputed with current develop. This can be done by downloading an event from the network and adding it to the `EventMigrationTest`.

lpetrovic05:
closes #15100


Neeharika-Sompalli (approved):
LGTM! Thanks @lpetrovic05

Issue #15438, opened 2024-09-11
Eliminate busy loop in HalfDiskHashMap.endWriting()


artemananiev:
`HDHM.endWriting()` has a busy loop. It was originally introduce to maximize thread throughput, while it was handling multiple HDHM buckets in parallel. However, it can be implemented more efficiently using tasks.

alex-kuzmin-hg:
It's how endWriting() looks at ENGNET2 Tom's test (15 mins) in v0.53.4:
![image](https://github.com/user-attachments/assets/693f4186-5c28-44eb-b330-a1f64c724adf)

alex-kuzmin-hg:
Now on performance Eng site: https://perf.analytics.eng.hashgraph.io/engnet2-20240906_175600/reports/com.swirlds.merkledb.files.hashmap.HalfDiskHashMap.html#448

PR #15439, opened 2024-09-11
fix: 15438: Eliminate busy loop in HalfDiskHashMap.endWriting()

artemananiev:
Fix summary:

* `HDHM.endWriting()` is reimplemented using fork-join pool and tasks
* No new unit tests are introduced, since `endWriting()` is well covered by existing tests

Fixes: https://github.com/hashgraph/hedera-services/issues/15438


platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +456,198 @@ public DataFileReader endWriting() throws IOException {
...
-                                    fileCollection.storeDataItem(bucket::writeTo, bucket.sizeInBytes());
-                            // update bucketIndexToBucketLocation
-                            bucketIndexToBucketLocation.put(bucketIndex, bucketLocation);
-                        }
-                    } finally {
-                        ++processed;
-                    }
+                // Wait until all tasks are completed. The latch is set in the end of the last
+                // executed "store bucket" task or when an exception is occurred
+                finishWritingLatch.await();
```
anthony-swirldslabs, line 485: Minor: If the iterator above is empty, then the latch may never be released. And if the tasks block for whatever reason, it also may not be released. Should this `await()` call use a timeout and error out if one occurs?

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +456,198 @@ public DataFileReader endWriting() throws IOException {
...
+                if (exceptionOccurred.get() != null) {
+                    throw exceptionOccurred.get();
                 }
+                // close files session
+                dataFileReader = fileCollection.endWriting(0, numOfBuckets);
+                // we have updated all indexes so the data file can now be included in merges
+                dataFileReader.setFileCompleted();
```
anthony-swirldslabs, line 492: Minor: I'm unsure about the logic that calls this method, but these finalization tasks won't be executed if an exception is thrown above. Should these go to a `finally{}` block?

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +456,198 @@ public DataFileReader endWriting() throws IOException {
...
-            final Iterator<IntObjectPair<BucketMutation>> iterator =
-                    oneTransactionsData.keyValuesView().iterator();
-
-            // read and update all buckets in parallel, write sequentially in random order
-            fileCollection.startWriting();
-            int processed = 0;
-            int inFlight = 0;
-            while (processed < size) {
-                // submit read tasks
-                while (inFlight < MAX_IN_FLIGHT && iterator.hasNext()) {
```
imalygin, line 442: `MAX_IN_FLIGHT` constant is no longer in use.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +456,198 @@ public DataFileReader endWriting() throws IOException {
...
+    }
+
+    public class StoreBucketTask extends AbstractTask {
+
+        private final Bucket bucket;
+
+        private StoreBucketTask next;
+
+        StoreBucketTask(final ForkJoinPool pool, final Bucket bucket) {
+            super(pool, 2);
```
imalygin, line 616: It's worth clarifying that this configuration will result in sequential execution of these tasks.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Correct, this is why this code is only run if `size > 0`, there is a check about it above

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: If an exception is thrown, it isn't recoverable, it doesn't make sense to pretend that the file has been written successfully. This is no different from the current implementation.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Indeed, it is not. However, it will be used again in the next version of the fix

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Oh, I was going to provide some JavaDoc for `StoreBucketTask`, but forgot about it... I'll describe this task and its dependencies shortly

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
anthony-swirldslabs: Sounds good.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
anthony-swirldslabs: Sounds good as long as this doesn't lead to files corruption on disk.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -116,35 +114,58 @@ public class HalfDiskHashMap implements AutoCloseable, Snapshotable, FileStatist
...
+     * compared against {@link #numBuckets} to manage the first and the last "sture bucket"
+     * task dependencies
+     */
+    private final AtomicInteger storeBucketTasksCreated = new AtomicInteger();
+
+    /**
+     * A latch used in {@link #endWriting()} to wait for all submitted tasks to complete. It
+     * is set in the last store task.
+     */
+    private final AtomicReference<AbstractTask> notifyTaskRef = new AtomicReference<>();
```
anthony-swirldslabs, line 147: Nit: s/latch/task/

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +456,186 @@ public DataFileReader endWriting() throws IOException {
...
+        private void createAndScheduleStoreTask(final Bucket bucket) {
+            // Create a subsequent "store bucket" task for the bucket
+            final StoreBucketTask storeTask = new StoreBucketTask(getPool(), bucket);
+            // The last created "store bucket" task. storeTask above will be set as an
+            // output dependency for that task to make sure tasks are running only one at
+            // a time. See StoreBucketTask for details
+            final StoreBucketTask prevTask = lastStoreTask.getAndSet(storeTask);
+            if (prevTask != null) {
+                // This will trigger prevTask execution as soon as its prev task is complete
+                prevTask.setNext(storeTask);
```
anthony-swirldslabs, line 529: Minor: just for my education: since the prevTask already exists and may have started on a separate thread already, is it technically possible for the prevTask to finish before this `setNext` call is performed here? What happens in that case? Will the `storeTask` ever be executed then?

anthony-swirldslabs (approved):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: The latest version of the fix respects `MAX_IN_FLIGHT` as the maximum number of buckets to be processed in parallel to minimize resource usage (every bucket has a heap byte buffer)

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Fixed

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: No this is not possible. For a task to run, both its dependencies must be set. One of the dependencies is "next", which is set above. It means, prevTask may not be completed (or even running) here

anthony-swirldslabs (approved):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +465,242 @@ public DataFileReader endWriting() throws IOException {
...
+        }
+
+        @Override
+        protected boolean exec() {
+            // The next submit task to run after the current one. It will only be run, if
+            // this task doesn't schedule tasks for all remaining buckets, and at least one
+            // bucket is completely processed while this method is running
+            final SubmitTask nextSubmitTask = new SubmitTask(getPool(), it, 2);
+            final boolean newSubmitTaskSet = currentSubmitTask.compareAndSet(this, nextSubmitTask);
+            assert newSubmitTaskSet;
```
OlegMazurov, line 532: `currentSubmitTask` does not need to be atomic, not even `volatile`. JMM ensures its visibility in all threads due to atomic updates of `bucketPermits` (that probably needs to be explicitly stated in comments).

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -425,93 +465,242 @@ public DataFileReader endWriting() throws IOException {
...
+                lastStoreTask.get().setNext(notifyTaskRef.get());
+            }
+        }
+
+        @Override
+        protected boolean exec() {
+            try {
+                BufferedData bucketData =
+                        fileCollection.readDataItemUsingIndex(bucketIndexToBucketLocation, bucketIndex);
+                // The bucket will be closed by WriteBucketTask
```
OlegMazurov, line 599: s/WriteBucketTask/StoreBucketTask/

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -116,35 +114,66 @@ public class HalfDiskHashMap implements AutoCloseable, Snapshotable, FileStatist
...
+     * compared against {@link #numBuckets} to manage the first and the last "store bucket"
+     * task dependencies
+     */
+    private final AtomicInteger storeBucketTasksCreated = new AtomicInteger();
+
+    /**
+     * A task used in {@link #endWriting()} to wait for all submitted tasks to complete. It
+     * is set as a "next" dependency for the last created "store bucket" task.
+     */
+    private final AtomicReference<AbstractTask> notifyTaskRef = new AtomicReference<>();
```
OlegMazurov, line 155: Does not have to be atomic or even volatile.

OlegMazurov (dismissed):
Pending performance testing.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: When an exception occurs, this task is notified to complete exceptionally. It happens without any changes to bucketPermits. I'd leave `notifyTaskRef` an atomic reference.

OlegMazurov: `notifyTaskRef` is set before any task that might complete it exceptionally is created and passed to another thread for execution. There is enough "happens-before" to ensure its visibility.

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Discussed it with Oleg offline, this is not critical (and the current version is more readable)

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
artemananiev: Noted. I'll fix this comment in some future fixes.

imalygin (approved):

platform-sdk/swirlds-merkledb/src/main/java/com/swirlds/merkledb/files/hashmap/HalfDiskHashMap.java
```diff
@@ -116,35 +114,66 @@ public class HalfDiskHashMap implements AutoCloseable, Snapshotable, FileStatist
...
-                                    .setThreadName("HalfDiskHashMap Flushing")
-                                    .setExceptionHandler((t, ex) -> logger.error(
-                                            EXCEPTION.getMarker(), "Uncaught exception during HDHM flushing", ex))
-                                    .buildFactory());
-                    flushExecutor = exec;
+                pool = flushingPool;
+                if (pool == null) {
+                    final MerkleDbConfig vmConfig = ConfigurationHolder.getConfigData(MerkleDbConfig.class);
+                    final int hashingThreadCount = vmConfig.getNumHalfDiskHashMapFlushThreads();
+                    pool = new ForkJoinPool(hashingThreadCount);
```
OlegMazurov, line 171: There is a subtle difference between the proposed configuration and the previous one: the calling thread in `endWriting()` participated in processing by performing writes in a busy loop; now, it's waiting in `join()`.
I suggest increasing `merkleDb.percentHalfDiskHashMapFlushThreads` from 50% to 75% to compensate for that (actually, improve on that). `VirtualMapBench` shows 5% improvement.

anthony-swirldslabs (approved):

OlegMazurov (approved):

imalygin (approved):

PR #15444, opened 2024-09-11
fix: recreate block hash from state

### LINKED ISSUE #15384: Add `BlockStreamValidator` that validates direct and indirect block proofs in CI PR checks

### LINKED ISSUE #15431: Add sufficient context in state to recompute block hash on restart
### Problem

To produce the correct hash for block `N+1`, a node that restarted from a state `S` including all transactions up to block `N` must know the hash of block `N`. Call this `H[N]`.

Now, `S` cannot directly _contain_ `H[N]`, because every change to `S` generates a `StateChange` block item in the block `N` "output tree"---which is itself an input to `H[N]`.

In particular, we have,
```
H[N] = sha384(
    sha384(H[N-1], <input tree root hash for block N>),
    sha384(<output tree root hash for block N>, H[S-1]))
```
where `H[S-1]` denotes the hash of the state at the start of block `N`.

### Solution

Add to the [`BlockStreamInfo` singleton](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto#L53) five new fields,
```
/**
 * The root hash of the input item tree for this block.
 */
bytes input_tree_root_hash = 5;

/**
 * The state hash at the start of this block.
 */
bytes start_of_block_state_hash = 6;

/**
 * The number of output block items preceding the state change that updates this
 * singleton (which MUST be the last output block item in the block)
 */
uint32 num_preceding_output_items = 7;

/**
 * The concatenation of the rightmost hashes at each level of the output tree as it
 * is constructed from all output items preceding the state change updating this
 * singleton (which MUST be the last output block item in the block)
 */
bytes rightmost_preceding_output_tree_hashes = 8;

/**
 * The last-used consensus time in this block.
 */
proto.Timestamp block_end_time = 9;
```

A node restarting from `S` can now serialize and hash the `StateChange` implied by these fields to complete the output tree hash for block `N`; and then combine that with the given `input_tree_root_hash`, `start_of_block_state_hash`, and `H[N-1]` already present in the `BlockStreamInfo` to recover `H[N]` and initialize its `BlockStreamManager` to pick up as if it never restarted.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15384
 - Closes #15431
 - 📗 _Important note to reviewer:_ The best way to understand this PR is to look at how much stricter it makes the `StateChangesValidator`:
     * _Before this PR_ - To pass the validator, the block stream only had to,
         1. Include a sequence of `StateChanges` that, when applied in order to an initialized genesis state, recovered the root hash of the `@HapiTest` network's final freeze state.
    * _After this PR_ - To pass the validator, every block must **also** end with a valid `BlockProof`, which means:
        1. The `StateChanges`, when applied in order, must recover the root hash of the `@HapiTest` network's state at every block boundary, not just the final freeze state.
        2. After each restart or reconnect in the `@HapiTest`, each node must have correctly initialized its previous block hash to the hash of the last block included in the state it restarted from. (Otherwise its proof would no longer match the `previous_block_root_hash` known to the validator.)

This PR makes three main categories of changes  to pass the much stricter `StateChangesValidator`:

1️⃣ **Extends the `BlockStreamInfo` singleton to let nodes recover the hash of the last block included in a saved state.**

Of the four ingredients in a block hash for block `N`, three can trivially be recovered from state: The start-of-block state hash, the previous `N-1` block hash, and the input tree root hash for block `N`. The output tree root hash for block `N` is, however, impossible to put into the state that ends with block `N`, since every act of changing state also changes the containing block's output tree hash.

We can work around this in three steps:
  1. Guarantee the the final `StateChanges` block item contains **only** an update to the `BlockStreamInfo` singleton (which means this last output item can be reconstructed from the value of the singleton in state).
  2. Extend the `BlockStreamInfo` with the summary of the penultimate output tree hashes.
  3. Add logic in `Hedera` on restart to combine (1) and (2) to recover the ultimate output tree hash for the last block in the saved state.

The first step we accomplish by making `BlockStreamManagerImpl` always put the `BlockStreamInfo` update as the final output `BlockItem`. The second step we accomplish by giving `ConcurrentStreamingTreeHasher` a `status()` method that returns the number of leaves in the penultimate tree, as well as the rightmost hashes at every level of that tree. The final step we accomplish by adding [this](https://github.com/hashgraph/hedera-services/blob/15431-restart-block-hash-mgmt/hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java#L918) `startBlockHashFrom(State)` method to `Hedera`. See the image below.

recover-output-tree-hash-from-state

2️⃣ **Updates `BlockStreamManager` with a `Map` of round number to end-of-round hash `Future`**

Hashing states can be very time-consuming, so we delay blocking on the start-of-state hash future for block `N` until the very last moment when we must compute its block hash. This future is completed either immediately on restart (or genesis) by the `ServicesMain` entry point; or after reconnect by a `ReconnectCompleteNotification` with a hashed state; or during normal operations by a `StateHashedNotification`.

3️⃣ **Fixes some misc existing behaviors that were incompatible with `BlockProof` production or outright bugs**

There were a few items that were innocent enough with the record stream, but fundamentally incompatible with block streams. These are the following,
  1. Immediately externalize any migration state changes in the first block after restart instead of delaying until the first user transaction (any delay means the validator will not be able to recover the correct start-of-block state hashes).
  2. Attach any post-upgrade `StateChanges` to staking info for any added and deleted nodes to a new synthetic `NodeStakeUpdate` transaction instead of doing this in `StakingInfoManagementSchema#restart()` (which is an incorrect usage of that API, anyways.)
  3. Clear all pending rewards for a newly deleted node; and alway assign zero reward rate to deleted nodes.


hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -852,6 +869,59 @@ private void initializeDagger(
...
+        // the block was devoted to putting the BlockStreamInfo itself into the state
+        final var outputTreeRootHash = outputTreeRootHashFrom(blockStreamInfo);
+        final var rightParent = combine(outputTreeRootHash, blockStreamInfo.startOfBlockStateHash());
+        return combine(leftParent, rightParent);
+    }
+
+    /**
+     * Given a {@link BlockStreamInfo} context, computes the output tree root hash that must have been
+     * computed at the end of the block that the context describes, assuming the final output block item
+     * was the state change that put the context into the state.
```
thomas-swirlds-labs, line 902: Are there any conditions under which that assumption might not hold -- or will it always definitely be the last output item ?

tinker-michaelj: No exceptions...the `BlockStreamInfo` singleton update will always be the last block output item---see [here](https://github.com/hashgraph/hedera-services/blob/15431-restart-block-hash-mgmt/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java#L224).

thomas-swirlds-labs: Awesome! Thanks @tinker-michaelj  😎

edward-swirldslabs (dismissed):

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
jsync-swirlds, line 1: Recommended modification/addition between lines 51 and 52 (github will not allow comment that far from changed lines)
```java
 * immediately prior completed block.<br/>
 * The state change to update this singleton MUST be the last "output"
 * block item in this block.
```

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -92,4 +92,32 @@ message BlockStreamInfo {
...
+    /**
+     * The number of output block items preceding the state change that updates this
+     * singleton (which MUST be the last output block item in the block)
+     */
+    uint32 num_preceding_output_items = 7;
+
+    /**
+     * The concatenation of the rightmost hashes at each level of the output tree as it
+     * is constructed from all output items preceding the state change updating this
+     * singleton (which MUST be the last output block item in the block)
```
jsync-swirlds, line 115:
```suggestion
     * A concatenation of SHA2-384 hash values.<br/>
     * This is the "rightmost" values of the "output" subtree.
     * <p>
     * The subtree containing these hashes SHALL be constructed from all "output"
     * `BlockItem`s in this block that _precede_ the update to this singleton.
```

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
```diff
@@ -92,4 +92,32 @@ message BlockStreamInfo {
...
+     */
+    repeated bytes rightmost_preceding_output_tree_hashes = 8;
+
+    /**
+     * The last-used consensus time for the current block.<br/>
```
jsync-swirlds, line 120:
```suggestion
     * A block-end consensus time stamp.
     * <p>
     * This field SHALL hold the last-used consensus time for
     * the current block.
```

tinker-michaelj:
> HIP-1037 PR

Ah nice @jsync-swirlds , will read those guidelines carefully, tyvm!

hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto
tinker-michaelj: Added this.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
tinker-michaelj: Agreed @thomas-swirlds-labs , @jsync-swirlds also suggested this---the [header comments for this message](https://github.com/hashgraph/hedera-services/blob/15431-restart-block-hash-mgmt/hapi/hedera-protobufs/services/state/blockstream/block_stream_info.proto#L52) now state it must come last:
```
 * The state change to update this singleton MUST be the last "output"
 * block item in this block.
```

jsync-swirlds (approved):
Approved with respect to the proto file update.

Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj 🙌

edward-swirldslabs (approved):

PR #15454, opened 2024-09-12
chore: 15437 Added log statement to analyze

### LINKED ISSUE #15437: Analyze usage of `State#getWritablePlatformState` to make sure that there no other lurking thread races
### Problem

This is a follow-up for https://github.com/hashgraph/hedera-services/pull/15389
We need to make sure that there are no other thread races when it comes to the platform state

### Solution

One of the solution could be to add logging with a thread name included to see which threads call `getWritablePlatformState` and `getReadablePlatformState`. This would be a good starting point. 

Direct code analysis may work too but it can be quite challenging to trace all the possible calls manually

### Alternatives

_No response_

imalygin:
TBD

Fixes #15437 

**Notes for reviewer**:

**Checklist**


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -356,13 +357,27 @@ public ReadableStates getReadableStates(@NonNull String serviceName) {
...
+        logger.info(STARTUP.getMarker(), "Getting writable states for service {}", serviceName);
+        logLimitedStackTrace();
         throwIfImmutable();
         return writableStatesMap.computeIfAbsent(serviceName, s -> {
             final var stateMetadata = services.getOrDefault(s, Map.of());
             return new MerkleWritableStates(serviceName, stateMetadata);
         });
     }
 
+    private void logLimitedStackTrace() {
```
lpetrovic05, line 369: this could introduce a performance issue, I don't know how often this is called. I think the best way to go is to not allow access to the mutable state by anyone other than the consensus thread

imalygin: This is only for information gathering purposes. This code will not get into prod.

lpetrovic05: there will be a PR to remove it?

lpetrovic05 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: No, this PR is just for experiments :)

PR #15457, opened 2024-09-12
feat: introduce PbjRecordHasher and RosterUtils.hash(Roster)

### LINKED ISSUE #15456: Introduce roster hasher

anthony-swirldslabs:
**Description**:
As a part of a refactoring to replace AddressBook usages with Roster, I'm adding support for hashing Roster objects similar to what we use for events hashing.

I'm introducing a generic `PbjRecordHasher` able to hash an arbitrary PBJ record. And then I use it in a newly introduced RosterUtils.hash(Roster) to compute a Roster hash.

**Related issue(s)**:

Fixes #15456

**Notes for reviewer**:
Unit tests pass.

**Checklist**


platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/roster/RosterUtilsTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import com.hedera.hapi.node.state.roster.Roster;
+import com.hedera.hapi.node.state.roster.RosterEntry;
+import com.swirlds.common.crypto.Hash;
+import org.junit.jupiter.api.Test;
+
+public class RosterUtilsTest {
+    @Test
```
kfa-aguda, line 27: i think we'll need more tests than these - certainly one with a valid roster (one that passes the validity test).

anthony-swirldslabs: Added in the latest commit. Please have a look.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/PbjRecordHasher.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    @NonNull
+    public Hash hash(@NonNull final T record, @NonNull final Codec<T> codec) {
+        try {
+            codec.write(record, stream);
+        } catch (final IOException e) {
+            throw new RuntimeException("An exception occurred while trying to hash a record!", e);
+        }
+
+        return new Hash(digest.digest(), DIGEST_TYPE);
```
jsync-swirlds, line 68: It may not be obvious to a (future) reader of this code that this call _also resets the digest_.  Perhaps a comment would be appropriate?
```suggestion
        // Reminder, MessageDigest.digest resets the digest, so subsequent writes
        // will calculate an independent hash value.
        return new Hash(digest.digest(), DIGEST_TYPE);
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/PbjRecordHasher.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param record a PBJ model
+     * @param codec a codec for the given model
+     * @return a Hash object
+     */
+    @NonNull
+    public Hash hash(@NonNull final T record, @NonNull final Codec<T> codec) {
+        try {
+            codec.write(record, stream);
+        } catch (final IOException e) {
+            throw new RuntimeException("An exception occurred while trying to hash a record!", e);
```
edward-swirldslabs, line 65: Is there anything more specific that can be thrown than a generic `RuntimeException`?   If not, then ignore this comment.  I don't think its worth making a new exception over for this instance.  Just surprised there isn't a runtime hash related exception we could throw.

edward-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/PbjRecordHasher.java
anthony-swirldslabs: Good idea. Done.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/util/PbjRecordHasher.java
anthony-swirldslabs: No, not really, at least I cannot find anything close to it. Also, I'm mainly replicating the pattern currently established in the PbjStreamHasher (which is specific for hashing events and it uses a slightly more complicated way to hash them.)

So for now, the RuntimeException it is.

kfa-aguda (approved):
thanks!

PR #15462, opened 2024-09-13
chore: remove snapshot ops

### LINKED ISSUE #15458: Remove snapshot fuzzy-matching from `test-clients`
### Problem

In the early stages of diff testing `v0.49` Services record streams, it was useful to snapshot the records generated by running one or more `@HapiTest`'s against mono-service and then fuzzy-match them against the records produced by mod-service.

This became much less useful once we were close enough to start doing exact matching. Now that `@RepeatableHapiTest`s allow deterministic test execution, fuzzy-matching is even less likely to have long-term value.

So the snapshot/fuzzy-match code is just taking up unused space and adding complexity to the `HapiSpec` framework.

### Solution

Remove this functionality.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #15458


thomas-swirlds-labs (approved):
LGTM. Thanks @tinker-michaelj 🎉

PR #15282, opened 2024-08-30
build: use common Hiero Gradle config

jjohannes:
**Description**:

Before this can be integrated:

- [x] Integrate #15578
- [x] Integrate #15471 
- [x] Publish _hiero-gradle-conventions_ plugins to Gradle Plugin Portal and use from there

**Related issue(s)**:

#14255

PR #15470, opened 2024-09-16
build: cleanup settings.gradle.kts / remove build.gradle.kts

jjohannes:
**Description**:

In preparation for reusable build config (#14255 /  #15282):

- Protobuf versions are now also managed in 'hedera-dependency-versions'
- Root 'build.gradle.kts' removed (config done in plugins)
- Special project – 'hedera-dependency-versions' and 'reports' – are defined in plugins as they'll be the same in all Hedera projects

**Related issue(s)**:

#14255 /  #15282

**Notes for reviewer**:

**Checklist**


.github/workflows/node-zxc-compile-application-code.yaml
```diff
@@ -516,7 +516,9 @@ jobs:
             ) &&
             !cancelled()
           }}
-        run: sed -i 's/^org.gradle.configuration-cache=.*$/org.gradle.configuration-cache=false/' gradle.properties
+        run: |
+          sed -i 's/^org.gradle.configuration-cache=.*$/org.gradle.configuration-cache=false/' gradle.properties
+          touch build.gradle.kts
```
jjohannes, line 521: Create empty `build.gradle.kts` in root, because Snyk right now expects a Gradle project to have a this. Which is not a requirement for a Gradle project, only `settings.gradle.kts` is required.

gradle/plugins/src/main/kotlin/com.hedera.gradle.lifecycle.gradle.kts
```diff
@@ -23,12 +23,6 @@ plugins {
 // tasks from the 'build' group
 defaultTasks("tasks")
 
-tasks.named<TaskReportTask>("tasks") {
-    if (!isDetail) {
-        displayGroup = "build"
-    }
-}
-
```
jjohannes, line 31: When I added this, the idea was to only show the most important tasks when users do use CLI and run `./gradlew`, but this does not work as expected. `idDetail` is always false, even if you run with `--detail`. Lets remove this cusotmization. If you use Gradle through the IDE it does not matter at all.

hedera-dependency-versions/build.gradle.kts
```diff
@@ -22,7 +22,9 @@ dependencies {
     api(enforcedPlatform("io.netty:netty-bom:4.1.110.Final"))
 
     // forward logging from modules using SLF4J (e.g. 'org.hyperledger.besu.evm') to Log4J
-    runtime(javaModuleDependencies.gav("org.apache.logging.log4j.slf4j2.impl"))
+    runtime("org.apache.logging.log4j:log4j-slf4j2-impl") {
+        because("org.apache.logging.log4j.slf4j2.impl")
+    }
```
jjohannes, line 27: Changed to be consistent with the other entries.

hapi/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ plugins {
     id("com.hedera.gradle.protobuf")
     id("com.hedera.gradle.services-publish")
     id("com.hedera.gradle.feature.test-fixtures")
-    alias(libs.plugins.pbj)
+    id("com.hedera.pbj.pbj-compiler") version "0.9.2"
```
jjohannes, line 21: Because this is a plugin version (which Gradle needs early during configuration). This cannot go into `hedera-dependency-versions`. Since this is something **only** used in the `hapi` project, I think it is best to define the version here.

gradle/plugins/src/main/kotlin/com.hedera.gradle.protobuf.gradle.kts
```diff
@@ -23,31 +23,23 @@ plugins {
...
-        // For dependencies of protobuf compilation use versions from 'hedera-dependency-versions',
-        // but not 'runtime' dependencies of the platform (JAVA_API instead of JAVA_RUNTIME).
-        dependencies {
-            compileProtoPath(project(":hedera-dependency-versions")) {
-                attributes {
-                    attribute(Usage.USAGE_ATTRIBUTE, objects.named(Usage.JAVA_API))
-                    attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category.REGULAR_PLATFORM))
-                }
-            }
+configurations.configureEach {
```
jjohannes, line 36: Need to do some trickery here due to how the Protobuf Gradle Plugin searches for dependencies to the "Tools" (Protobuf Compiler) it uses. This will go into the separate Gradle Config repo, in the next step, when we move this over.

rbarkerSL (approved):

PR #15471, opened 2024-09-16
build: generalize publishing setup

jjohannes:
**Description**:

Do not have 'hedera-services' specifics in plugins so that they may be reused without modification in other repositories.

For this, we introduce two files in the folder for each "product"
- developers.properties
- description.txt (optional)

See changes in the PR for what these files contain.

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14255 / https://github.com/hashgraph/hedera-services/pull/15282

**Notes for reviewer**:

The information moved into the `developers.properties` files end up in the Metadata of the published Jars. E.g. here:
https://repo1.maven.org/maven2/com/hedera/hashgraph/hapi/0.53.5/hapi-0.53.5.pom

Please share if you have alternative/better suggestion for where we should maintain this information.

**Checklist**


developers.properties
```diff
@@ -0,0 +1,6 @@
+# This file is here for 'hapi', because that "product" is missing the toplevel folder
+# (it contains the 'hapi' module directly)
+hedera-base@swirldslabs.com=Hedera Base Team
+hedera-services@swirldslabs.com=Hedera Services Team
+hedera-smart-contracts@swirldslabs.com=Hedera Smart Contracts Team
+release-engineering@swirldslabs.com=Release Engineering Team
```
jjohannes, line 6: For symmetry, I would suggest to move `hapi` one level up into `hapi/hapi`. Even if it is just one "Module" in the "Product". (In a separate PR.)

lpetrovic05, line 6: we are phasing out `swirldlabs.com` and moving to `hashgraph.com`

jjohannes: Would it be the same addresses for the teams, just replace the domain? Then I can do that as part of this PR.

lpetrovic05: I think so, maybe @poulok knows, or @nathanklick

lpetrovic05 (dismissed):

thomas-swirlds-labs (approved):
LGTM. Thanks! @jjohannes

artemananiev (approved):
Looks good to me from platform-data perspective

PR #15515, opened 2024-09-17
feat: Add TSS related system transaction and state protobufs (Services) 

### LINKED ISSUE #14744: Create TSS related system transaction and state protobufs

thomas-swirlds-labs:
**Description**:
Adds the following to the services `hapi/hedera-protobufs`: 

- [x] System Transaction: TssMessageTransaction
- [x] System Transaction: TssVoteTransaction
- [x] State Queue: LedgerId
- [x] State Map: TssMessageMapKey
- [x] State Map: TssVoteMapKey

**Related issue(s)**:

Fixes #14744 

**Notes for reviewer**:

**Checklist**


hapi/hedera-protobufs/platform/event/tss_message_transaction.proto
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+option java_package = "com.hedera.hapi.platform.event.legacy";
+// <<<pbj.java_package = "com.hedera.hapi.platform.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * The TssMessageTransaction is used to send a TssMessage to the network for a candidate roster.
+ */
+message TssMessageTransaction {
```
tinker-michaelj, line 38: I think this and the `TssVoteTransaction` should be messages ending with `TransactionBody` that live in the same folder with the other `*TransactionBody` messages (e.g. `CryptoDeleteTransactionBody` [here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/crypto_delete.proto#L36)).

And then we should include these in the `TransactionBody#data` oneof [here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/transaction_body.proto#L133).

thomas-swirlds-labs: Have updated. Thanks @tinker-michaelj 👍

tinker-michaelj: Ah tyvm @thomas-swirlds-labs ...I had only two other comments:
  1. We need corresponding `HederaFunctionality` enum values [here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/basic_types.proto#L1251).
  2. Since TSS is now an application service, would it make more sense to move its state definitions (tss_{vote,message}_map_key.proto) files under a _services/state/tss_ directory?

edward-swirldslabs (commented):
Do the comments by @jsync-swirlds in the protobuf repo apply here as well?

hapi/hedera-protobufs/services/state/roster/ledger_id.proto
```diff
@@ -0,0 +1,106 @@
...
+ */
+message LedgerId {
+
+  /**
+   * A public key.
+   * This key both identifies the ledger and can be used to verify ledger signatures.
+   * <p>
+   * This value MUST be set.
+   * This value MUST NOT be empty.
+   * This value MUST contain a valid public key.
```
jsync-swirlds, line 46: This lacks line breaks; without them the produced documents will be concatenated to a single line due to markdown tables.
```suggestion
   * A public key.<br/>
   * This key both identifies the ledger and can be used to verify ledger signatures.
   * <p>
   * This value MUST be set.<br/>
   * This value MUST NOT be empty.<br/>
   * This value MUST contain a valid public key.
```

hapi/hedera-protobufs/services/state/tss/tss_message_map_key.proto
```diff
@@ -0,0 +1,52 @@
...
+ * ### Keywords
+ * The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+ * "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+ * document are to be interpreted as described in
+ * [RFC2119](https://www.ietf.org/rfc/rfc2119) and clarified in
+ * [RFC8174](https://www.ietf.org/rfc/rfc8174).
+ */
+syntax = "proto3";
+
+package proto;
```
jsync-swirlds, line 13: We should not continue to (mis)use `proto` as a package:
```suggestion
package com.hedera.hapi.platform.tss;
```

hapi/hedera-protobufs/services/transaction_body.proto
```diff
@@ -418,5 +421,15 @@ message TransactionBody {
         * A transaction body for a `claimAirdrop` request.
          */
         TokenClaimAirdropTransactionBody tokenClaimAirdrop = 60;
+
+        /**
+         * A transaction body for a `tssMessage` request.
```
jsync-swirlds, line 426: Are these messages client requests, or internal "system" transactions?

hapi/hedera-protobufs/services/tss_vote_transaction.proto
```diff
@@ -0,0 +1,76 @@
...
+ * ### Keywords
+ * The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+ * "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+ * document are to be interpreted as described in
+ * [RFC2119](https://www.ietf.org/rfc/rfc2119) and clarified in
+ * [RFC8174](https://www.ietf.org/rfc/rfc8174).
+ */
+syntax = "proto3";
+
+package proto;
```
jsync-swirlds, line 13: Package should match the pbj package for new proto files.
```suggestion
package com.hedera.hapi.platform.event;
```

hapi/hedera-protobufs/services/tss_vote_transaction.proto
```diff
@@ -0,0 +1,76 @@
...
+   * sequence number of a `TssMessageTransaction`. The least significant bit
+   * of byte[0] corresponds to sequence number 0 and the most significant bit
+   * of byte[0] corresponds to sequence number 7.  The least significant bit
+   * of byte[1] corresponds to sequence number 8 and the most significant bit
+   * of byte[1] corresponds to the sequence number 15, and so on. If a bit is
+   * set to 1, then the TssMessage for the corresponding TssMessageTransaction
+   * sequence number was received as one of the first Threshold number of
+   * valid TssMessages in consensus order.  If a bit is set to 0, then the
+   * TssMessage was invalid or there was no TssMessageTransaction with a
+   * sequence number corresponding to the bit index.
```
jsync-swirlds, line 73: This is one very long line rendered, and lacks both clarity and express requirements.
```suggestion
   * A bit vector of message statuses.<br/>
   * Each bit in this vector indicates receipt (1) or non-receipt (0) of a
   * _valid_ `TssMessage` for a corresponding `TssMessageTransaction`.
   * <p>
   * <h4>Example</h4>
   * <ul><li>The least significant bit of byte[0] SHALL be sequence 0.</li>
   *     <li>The most significant bit of byte[0] SHALL be sequence 7.</li>
   *     <li>The least significant bit of byte[1] SHALL be sequence 8.</li>
   *     <li>The most significant bit of byte[0] SHALL be sequence 15.</li>
   * </ul>
   * A bit SHALL be set if the `TssMessage` for the `TssMessageTransaction`
   * with a sequence number matching that bit index has been
   * received, and is valid.<br/>
   * A bit SHALL NOT be set if the `TssMessage` has not been received or was
   * received but not valid.
```

jsync-swirlds (commented):
A few category-type suggestions that should be repeated throughout.

hapi/hedera-protobufs/services/transaction_body.proto
thomas-swirlds-labs: For now they are internal system transactions but I believe in the future they could be called by clients also since we are exposing the TSS Service like any other Hedera API

jsync-swirlds: Interesting, I would not think we want external clients involving themselves in the TSS voting process (seems like an invitation to bad behavior).  The whole TSS process should, as I understand it, be much closer to gossip than a client API.

hapi/hedera-protobufs/services/tss_vote_transaction.proto
jsync-swirlds: Ugh.  I wish we could modify the Javadoc processing to stop being so strict about heading ordering; it's a bug in Javadoc dating over 20 years back to obsolete HTML standards.
Then again, having PBJ generate the new "markdown" flavored Javadoc comment style would also remove the issue.

hapi/hedera-protobufs/services/transaction_body.proto
edward-swirldslabs: I do not think the `TssBaseService` should be invokable directly from clients or that any clients should be able to produce any of the transactions that the `TssBaseService` produces.    If we were to open up a signing API to the public, we would create a new service that takes a user created `SignThisData` transaction and invoked the `TssBaseService` API on the back end of it.

jsync-swirlds: In that case we probably do not need (or want) these to be part of the `TransactionBody` definition; they probably fall more into the "system transaction" category and should not be invokable by a client.

edward-swirldslabs: In discussion with @rbair23,  if they were invoked by a user, we would have to filter out the transaction and reject it.  

I am being a little over dramatic.  It will not mess things up to receive it since the TssBaseService should ignore anything that isn't coming from a recognized node.   But it doesn't make sense to me why we would show users a button that no user will ever get to press.

tinker-michaelj: The tradeoffs between (1) building a parallel `SystemHandleWorkflow` that permits only the TSS transactions; and (2) reusing `HandleWorkflow` with an ingest check that prevents TSS transactions from coming through HAPI are _overwhelmingly_ in favor of (2), even though it's a bit unsatisfying.

This PR as written is the most efficient path forward.

tinker-michaelj (dismissed):
LGTM, tyvm @thomas-swirlds-labs !

Neeharika-Sompalli (approved):
LGTM! Needs some unit tests to be fixed

Neeharika-Sompalli (dismissed):
LGTM! tyvm @thomas-swirlds-labs

tinker-michaelj (dismissed):
LGTM, ty @thomas-swirlds-labs !

hapi/hedera-protobufs/services/tss_vote_transaction.proto
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * A signature produced by the node.<br/>
+   * This is produced using the node RSA signing key to sign the ledger_id.<br/>
+   * This signature is used to establish a chain of trust in the ledger id.
+   */
+  bytes node_signature = 4;
+
+  /**
+   * A bit vector of message statuses.<br/>
+   * Each bit in this vector indicates receipt (1) or non-receipt (0) of a
```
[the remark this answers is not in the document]
thomas-swirlds-labs: Agreed and removed. Thanks @edward-swirldslabs 👍

Neeharika-Sompalli (approved):
LGTM!

edward-swirldslabs (approved):

tinker-michaelj (approved):
LGTM, tyvm @thomas-swirlds-labs !

PR #15522, opened 2024-09-17
test: add block contents validator

MiroslavGatsanoga:
**Description**:

Create additional block stream validator that will validate the structure of the contents in each block.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/15474

**Notes for reviewer**:

Used [this doc](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/block/stream/block_item.proto#L61) for defining the checks that the validator makes.

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
```diff
@@ -0,0 +1,123 @@
...
+        // For the first block the `block_header` might be followed by `event_header` or `state_changes`, due to state
+        // migration.
+        if (isFirst) {
+            if (!blockItems.get(1).hasEventHeader() && !blockItems.get(1).hasStateChanges()) {
+                Assertions.fail("First block header not followed by event header or state changes");
+            }
+            return;
+        }
+
+        // TODO: In general, a `block_header` SHALL be followed by an `event_header`, but for hapiTestRestart we get
```
MiroslavGatsanoga, line 89: @tinker-michaelj can you advise if that's acceptable or should we enforce it here and apply a fix to the code that produces the blocks?

tinker-michaelj: Ah in that case the post-restart `State` initialization changes state _before_ any event has reached consensus. 

I think this is fine.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
```diff
@@ -0,0 +1,123 @@
...
+                    .anyMatch(item -> !item.hasEventHeader() && !item.hasStateChanges())) {
+                Assertions.fail(
+                        "Block with no user transactions should contain items of type `block_header`, `event_headers`, `state_changes` or `state_proof`");
+            }
+
+            return;
+        }
+
+        for (int i = 0; i < blockItems.size(); i++) {
+            // TODO: An `event_header` SHALL be followed by one or more `event_transaction` items. -> looks like we can
```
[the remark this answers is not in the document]
tinker-michaelj: Hmm yes I think it's very easy for there to be an `EventHeader` with no following `EventTransaction` because not every round will include a user transaction.

Seems fine to me.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
MiroslavGatsanoga: Updated code comment.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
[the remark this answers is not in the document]
MiroslavGatsanoga: Removed the comment with the commented-out code since we'll not be enforcing the `event_header SHALL be followed by one or more event_transaction` rule.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
```diff
@@ -0,0 +1,121 @@
...
+            try {
+                validate(blocks.get(i), i == 0);
+            } catch (AssertionError err) {
+                logger.error("Error validating block {}", blocks.get(i));
+                throw err;
+            }
+        }
+    }
+
+    private static void validate(Block block, boolean isFirst) {
```
derektriley, line 67: Is the argument `isFirst` necessary? Did you have an idea to use this in the future?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
```diff
@@ -0,0 +1,121 @@
...
+        if (!blockItems.getFirst().hasBlockHeader()) {
+            Assertions.fail("Block does not start with a block header");
+        }
+
+        // A block SHALL end with a `state_proof`.
+        if (!blockItems.getLast().hasBlockProof()) {
+            Assertions.fail("Block does not end with a block proof");
+        }
+
+        // In general, a `block_header` SHALL be followed by an `round_header`, but for hapiTestRestart we get
```
derektriley, line 85: Given this comment, should this validator account for a state change being present?

derektriley (approved):
Looks good as an initial validator for Block contents. See my additional comments

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
MiroslavGatsanoga: Missed to update the comment, this now applies to the round header check.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
MiroslavGatsanoga: Removed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
MiroslavGatsanoga: Updated.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/BlockContentsValidator.java
```diff
@@ -0,0 +1,119 @@
...
+
+        // A block SHALL start with a `header`.
+        if (!blockItems.getFirst().hasBlockHeader()) {
+            Assertions.fail("Block does not start with a block header");
+        }
+
+        // A block SHALL start with a `header`.
+        if (!blockItems.getFirst().hasBlockHeader()) {
+            Assertions.fail("Block does not start with a block header");
+        }
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Yeah, it's a duplicate, will remove it.

MiroslavGatsanoga:
> @tinker-michaelj I lost track of which validations are already running for block streams. Is this validator still relevant and useful, or do we have coverage for this in other validators?

@mhess-swl The configured block stream validators can be found [here](https://github.com/hashgraph/hedera-services/blob/15474-block-contents-validator/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/StreamValidationOp.java#L76).

mhess-swl (approved):

PR #15541, opened 2024-09-18
feat: Implement gRPC publishBlockStream streaming of blocks

### LINKED ISSUE #15526: Implement a gRPC client for publishBlockStream rpc (design for some handshaking)

petreze:
**Description**:

**Related issue(s)**:

Fixes #15526

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java
```diff
@@ -338,8 +341,10 @@ public synchronized void accept(@NonNull final byte[] message, @NonNull final by
                     .siblingHashes(siblingHashes.stream().flatMap(List::stream).toList());
             block.writer()
                     .writeItem(BlockItem.PROTOBUF.toBytes(
-                            BlockItem.newBuilder().blockProof(proof).build()))
-                    .closeBlock();
+                            BlockItem.newBuilder().blockProof(proof).build()));
+            if (streamWriterType == BlockStreamWriterMode.FILE) {
+                block.writer().closeBlock();
+            }
```
mhess-swl, line 347: This feels like a weird place to put this since it only applies to files. Maybe we have a method in this class that the file writer can override instead

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * @param blockStreamConfig the block stream configuration
+     */
+    public GrpcBlockItemWriter(@NonNull final BlockStreamConfig blockStreamConfig) {
+        requireNonNull(blockStreamConfig, "The supplied argument 'blockStreamConfig' cannot be null!");
+
+        channel = ManagedChannelBuilder.forAddress(blockStreamConfig.address(), blockStreamConfig.port())
+                .usePlaintext()
```
mhess-swl, line 84: I'm assuming communication via plaintext is temporary?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriterTest.java
mhess-swl, line 1: I like these test cases :)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,214 @@
...
+                /*, STREAM_ITEMS_SUCCESS, STREAM_ITEMS_BEHIND*/
+                /*
+                    onNext(PublishStreamResponse.newBuilder()
+                            .setStatus(EndOfStream.newBuilder()
+                                    .setStatus(STREAM_ITEMS_UNKNOWN)
+                                    .build())
+                            .build());
+
+                    closeBlock();
+                }*/
```
mhess-swl, line 111: Todo later?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,214 @@
...
+
+                if (streamResponse.hasAcknowledgement()) {
+                    final Acknowledgement acknowledgement = streamResponse.getAcknowledgement();
+                    if (acknowledgement.hasBlockAck()) {
+                        logger.info("PublishStreamResponse: a full block received: {}", acknowledgement.getBlockAck());
+                    } else if (acknowledgement.hasItemAck()) {
```
mhess-swl, line 117: Do we want to keep a series of booleans or something, representing that all block items have been acknowledged? Not sure if that's overkill or not here, just thinking aloud. Thoughts?

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
```diff
@@ -38,7 +38,7 @@
  */
 @ConfigData("blockStream")
 public record BlockStreamConfig(
-        @ConfigProperty(defaultValue = "BOTH") @NetworkProperty StreamMode streamMode,
+        @ConfigProperty(defaultValue = "RECORDS") @NetworkProperty StreamMode streamMode,
```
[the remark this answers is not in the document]
petreze: Forgot to remove it after debugging some issues, thanks for noticing!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
petreze: Those were for testing purposes only, currently I removed them but we will need to implement handlings of different scenarios based on the returned `EndOfStream` from block nodes

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
petreze: Currently, the intention of those are to show that the connection between CN and BN is established and we can receive items from block nodes after they receive a stream. In the future there should be some handling mechanisms on different scenarios based on the handshake approach, which is yet to be finalized

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java
petreze: Method `closeBlocks()` in BlockItemWriter is implemented by both writers (file and grpc). We just want to call them on different occassions. The file writer should be closed when we write the `BlockProof` which is the last `BlockItem` in every written Block. In the case of the grpc writer, we want to call `closeBlock()` whenever we receive `EndOfStream` object from the block node which indicates the stream ended, not immediately on streaming the `BlockProof`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,207 @@
...
+        UNINITIALIZED,
+        /**
+         * The gRPC client is currently open
+         */
+        OPEN,
+        /**
+         * The gRPC client is already closed
+         */
+        CLOSED
+    }
```
mhess-swl, line 77: Perhaps not necessary in this PR, but it would be helpful to document what each of these states mean at some point. For example, a writer in a `CLOSED` state can never transition to any other state.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
```diff
@@ -0,0 +1,207 @@
...
+
+    @Override
+    public void closeBlock() {
+        if (state.ordinal() < GrpcBlockItemWriter.State.OPEN.ordinal()) {
+            throw new IllegalStateException("Cannot close a GrpcBlockItemWriter that is not open");
+        } else if (state.ordinal() == GrpcBlockItemWriter.State.CLOSED.ordinal()) {
+            throw new IllegalStateException("Cannot close a GrpcBlockItemWriter that is already closed");
+        }
+
+        requestObserver.onCompleted();
```
mhess-swl, line 165: I like the request observer, it's cleaner than checking for the `FILE` type and closing inside this class. Well done!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
petreze: Addressed, thank you!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/GrpcBlockItemWriter.java
petreze: @mhess-swl 
I was originating from the communication between the Java SDK and services in [BaseNode.java](https://github.com/hashgraph/hedera-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/BaseNode.java#L281) where we provide a plaintext communication, if we are not using a TLS communication.

rbarkerSL (dismissed):
Review applies to:

- `gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts`
- `hedera-dependency-versions/build.gradle.kts`
- `hedera-node/hedera-app/build.gradle.kts`
- `hedera-node/hedera-app/src/main/java/module-info.java`

LGTM

gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
```diff
@@ -111,6 +111,24 @@ extraJavaModuleInfo {
...
+        requireAllDefinedDependencies()
+        requires("java.logging")
+        patchRealModule()
+    }
+    module("io.helidon.webclient:helidon-webclient-grpc", "io.helidon.webclient.grpc") {
+        exportAllPackages()
+        requireAllDefinedDependencies()
+        requires("java.logging")
+        patchRealModule()
+    }
```
jjohannes, line 131: Are these really needed or is this some kind of leftover?
The modules already have a module descriptor. That's why `patchRealModule()` needed to be added.

Is this because of the `requires("java.logging")`? If so this should also be reported to the Helidon project to be fixed upstream.

rbarkerSL (approved):
Review applies to:

- gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
- hedera-dependency-versions/build.gradle.kts
- hedera-node/hedera-app/build.gradle.kts
- hedera-node/hedera-app/src/main/java/module-info.java

LGTM

rbarkerSL (dismissed):
Review applies to:

- gradle/plugins/src/main/kotlin/com.hedera.gradle.jpms-modules.gradle.kts
- hedera-dependency-versions/build.gradle.kts
- hedera-node/hedera-app/build.gradle.kts
- hedera-node/hedera-app/src/main/java/module-info.java

LGTM

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
```diff
@@ -38,4 +42,6 @@ public record BlockStreamConfig(
         @ConfigProperty(defaultValue = "true") @NetworkProperty boolean compressFilesOnCreation,
         @ConfigProperty(defaultValue = "32") @NetworkProperty int serializationBatchSize,
         @ConfigProperty(defaultValue = "32") @NetworkProperty int hashCombineBatchSize,
-        @ConfigProperty(defaultValue = "1") @NetworkProperty int roundsPerBlock) {}
+        @ConfigProperty(defaultValue = "1") @NetworkProperty int roundsPerBlock,
+        @ConfigProperty(defaultValue = "localhost") String address,
```
mhess-swl, line 46: Nit: we may want to have a more descriptive name for this property, or some further documentation, specifically to address: wow does this differ from the already-configured address and port for the node?

mhess-swl (dismissed):

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/BlockStreamConfig.java
petreze: Addressed, tyvm!

mhess-swl (approved):

hedera-dependency-versions/build.gradle.kts
```diff
@@ -28,6 +28,15 @@ dependencies {
...
 dependencies.constraints {
+    api("io.helidon.common:helidon-common:4.1.1") {
+        because("io.helidon.common")
+    }
+    api("io.helidon.webclient:helidon-webclient:4.1.1") {
+        because("io.helidon.webclient")
+    }
+    api("io.helidon.webclient:helidon-webclient-grpc:4.1.1") {
+        because("io.helidon.webclient.grpc")
+    }
```
jsync-swirlds, line 39: The use of protoc objects and the helidon-grpc libraries (with secondary grpc.io dependencies) _may not_ be necessary.
The PBJ library has a [helidon "plugin"](https://github.com/hashgraph/pbj/tree/main/pbj-core/pbj-grpc-helidon) to replace the helidon grpc plugin that is _nearly_ complete (modulo a [few bug issues](https://github.com/hashgraph/pbj/issues/306)).

It may be worth evaluating the status of that work to determine if a slight delay of this work would enable use of that plugin to avoid introducing protoc-compiled classes and these helidon grpc dependencies.

petreze: Initially, this issue was meant to be a "first pass" API that will give us the possibility to iterate over it and improve it down the line. But this will be something to consider, for certain. Thanks!

timo0 (approved):

andrewb1269hg (approved):
Looks good to me.

PR #15543, opened 2024-09-18
feat: 15154 Added `createSnapshot` method to State API

### LINKED ISSUE #15154: Implement Snapshot API method
There are two tasks for this ticket:
1. Verify the specific API with the intended users and architect (Jasper). The current thought is to put this method on the State. Need to also consider how to specify the location to save the snapshot.
2. Implement and test it.
3. Refactor existing platform code to use this method for creating snapshots

imalygin:
**Description**:

This PR adds an implementation of `com.swirlds.state.State#createSnapshot` method. 

**Related issue(s)**:

Fixes #15154 

**Notes for reviewer**:

**Checklist**


mxtartaglia-sl (approved):

jeromy-cannon (approved):

litt3 (changes requested):
I'm not fully plugged into the problem, but after an initial review, here are my thoughts:

- You're constructing a new signed state, but it's never being registered with the state garbage collector. I don't think creating a snapshot should cause a new signed state to be created
- Your new snapshot method bypasses a lot of the logic that exists in `DefaultStateSnapshotManager`
- I'm lacking context here, that I don't see in the linked issue. What is the motivation to make this change? Is it a prerequisite for something?

I think the design here needs some careful consideration.

1. How does this fit into how the platform currently manages state snapshots?
2. How much of the logic that exists inside the normal flow of platform wiring is going to be moved elsewhere?
3. Who is allowed to call this method, and when?

imalygin:
@litt3 

> You're constructing a new signed state, but it's never being registered with the state garbage collector. I don't think creating a snapshot should cause a new signed state to be created

Agreed. I got rid of this synthetic state. It was unnecessary. However, to achieve that I had to extract signature set in a separate file. Even though `SignedState.swh` is technically no longer a "signed state" file, I decided not to change the name, as there may be multiple scripts relying on this file. 

> What is the motivation to make this change? Is it a prerequisite for something?

Correct. This logic is required by the Block node. It needs to be able to load and save snapshots of the Merkle tree. 

> How does this fit into how the platform currently manages state snapshots?

With the latest commit it fits well - the state snapshot creation logic now uses `State.createrSnapshot` method.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
```diff
@@ -54,7 +55,8 @@
  */
 public class SerializableDataInputStream extends AugmentedDataInputStream {
 
-    private static final int PROTOCOL_VERSION = SERIALIZATION_PROTOCOL_VERSION;
```
lpetrovic05, line 57: this version refers to the serialization protocol version. the signed state should have a separate version number which should be used

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1092,37 @@ public void computeHash() {
...
+
+    private static void writeMerkleRootToFile(final Path directory, final MerkleRoot merkleRoot) throws IOException {
+        writeAndFlush(
+                directory.resolve(SIGNED_STATE_FILE_NAME), out -> writeMerkleRootToStream(out, directory, merkleRoot));
+    }
+
+    private static void writeMerkleRootToStream(
+            final MerkleDataOutputStream out, final Path directory, final MerkleRoot merkleRoot) throws IOException {
+        out.write(VERSIONED_FILE_BYTE);
+        out.writeInt(FILE_VERSION);
```
lpetrovic05, line 1123: this is the file version that should be changed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileUtils.java
```diff
@@ -32,6 +32,8 @@ public final class SignedStateFileUtils {
      */
     public static final String SIGNED_STATE_FILE_NAME = "SignedState.swh";
 
+    public static final String SIGNATURE_SET_FILE_NAME = "signatureSet.bin";
```
lpetrovic05, line 35: separating the signatures will probably have some downstream effects on JRS, devops and so on. These need to be resolved before they can be separated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
```diff
@@ -112,28 +113,23 @@ public static void writeMetadataFile(
...
      * @param out         the stream to write to
-     * @param directory   the directory to write to
      * @param signedState the signed state to write
      */
-    private static void writeStateFileToStream(
-            final MerkleDataOutputStream out, final Path directory, final SignedState signedState) throws IOException {
+    private static void writeSignatureSetToStream(final MerkleDataOutputStream out, final SignedState signedState)
+            throws IOException {
         out.write(VERSIONED_FILE_BYTE);
         out.writeInt(FILE_VERSION);
```
lpetrovic05, line 121: the signature set should have a separate file version from the state file. and I don't think it needs to VERSIONED_FILE_BYTE, I think that is just a legacy thing

lpetrovic05:
@litt3 

> * You're constructing a new signed state, but it's never being registered with the state garbage collector. I don't think creating a snapshot should cause a new signed state to be created

I did not see this part, can you point me to it?

> * Your new snapshot method bypasses a lot of the logic that exists in `DefaultStateSnapshotManager`

I think this could have been bypassed even before with `SignedStateFileWriter`, so I'm not sure this is an issue.

litt3:
@lpetrovic05 

> I did not see this part, can you point me to it?

It was in a previous version of this PR, Ivan fixed it already.

> I think this could have been bypassed even before with `SignedStateFileWriter`, so I'm not sure this is an issue.

Thinking about it a bit more, I agree. If we're ok with the new utility lacking these things, there is no correctness problem

- signature checks before saving
- logging
- metrics
- cleanup

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/state/signed/StartupStateUtilsTests.java
```diff
@@ -128,14 +128,20 @@ private SignedState writeState(
             @NonNull final PlatformContext platformContext,
             final long round,
             @Nullable final Hash epoch,
-            final boolean corrupted)
+            final boolean corrupted,
+            final boolean immutable)
```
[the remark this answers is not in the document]
imalygin: Agreed. I renamed the variable.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
imalygin: Makes sense. I removed it from this file.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileUtils.java
imalygin: Sure, I'll investigate it.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Done

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableDataInputStream.java
imalygin: okay, I no longer use it for these purposes.

imalygin:
@litt3 

> signature checks before saving

This part is no longer relevant to this functionality, as we save a Merkle tree only, not the signed state. 

> logging
> metrics

Having some logs and metrics is a good idea, I'll add it. 

>cleanup

It's not really clear at this point how the clean up should work in this particular context. So, I'll let the caller of this method deal with it.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableStreamConstants.java
```diff
@@ -39,7 +39,7 @@ private SerializableStreamConstants() {}
     public static final int VERSION_BYTES = Integer.BYTES;
 
     /**
-     * The current version of the serialization protocol implemented by {@link SerializableDataOutputStream} and
+     * The previous version of the serialization protocol implemented by {@link SerializableDataOutputStream} and
```
litt3, line 42: Was this doc change intentional, or is it left over from a previous changeset? I'm not understanding why it used to be "current", but would now be "previous"

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -65,6 +71,8 @@ public static List<SavedStateInfo> getSavedStateFiles(
                 .getSavedStateFiles(mainClassName, platformId, swirldName);
     }
 
+    public record StateFileData(MerkleRoot state, Hash hash, SigSet sigSet, int fileVersion) {}
```
litt3, line 74: members missing nullity annotations

litt3, line 74: public record missing javadoc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
-                () -> new BufferedInputStream(new FileInputStream(stateFile.toFile())),
-                (final MerkleDataInputStream in) -> {
-                    readAndCheckVersion(in);
-
-                    final Path directory = stateFile.getParent();
-
-                    try {
-                        final MerkleRoot state = snapshotStateReader.apply(in, directory);
-                        final Hash hash = in.readSerializable();
+        StateFileData data = readStateFileData(stateFile, snapshotStateReader);
```
litt3, line 98: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+
+        final StateFileData normalizedData;
+        if (data.sigSet == null) {
+            File sigSetFile =
```
litt3, line 102: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+                    stateFile.getParent().resolve(SIGNATURE_SET_FILE_NAME).toFile();
+            normalizedData = deserializeAndDebugOnFailure(
+                    () -> new BufferedInputStream(new FileInputStream(sigSetFile)),
+                    (final MerkleDataInputStream in) -> {
+                        int fileVersion = readAndCheckSigSetFileVersion(in);
```
litt3, line 107: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+        newSignedState.setSigSet(normalizedData.sigSet);
 
         returnState = new DeserializedSignedState(
-                newSignedState.reserve("SignedStateFileReader.readStateFile()"), data.hash());
+                newSignedState.reserve("SignedStateFileReader.readStateFile()"), normalizedData.hash);
 
         return returnState;
     }
 
+    public static StateFileData readStateFileData(
```
litt3, line 132: missing doc for public method

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
-
-                    final Path directory = stateFile.getParent();
-
-                    try {
-                        final MerkleRoot state = snapshotStateReader.apply(in, directory);
-                        final Hash hash = in.readSerializable();
+        StateFileData data = readStateFileData(stateFile, snapshotStateReader);
+
+        final StateFileData normalizedData;
+        if (data.sigSet == null) {
```
litt3, line 101: Can this logic be baked into the `readStateFileData` utility method? That method already has a block for when the file version is `SIG_SET_SEPARATE_VERSION`. Wouldn't it make sense to just read in the sig files there?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
```diff
@@ -154,7 +147,14 @@ public static void writeSignedStateFilesToDirectory(
         Objects.requireNonNull(directory);
         Objects.requireNonNull(signedState);
 
-        writeStateFile(directory, signedState);
+        MerkleRoot state = signedState.getState();
+        if (state.isMutable()) {
```
litt3, line 151: Hmm, are we sure this is the right way to go? Would it make more sense to throw if you try writing a mutable state to disk?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -68,4 +69,6 @@ public interface MerkleRoot extends MerkleInternal {
     /** {@inheritDoc} */
     @NonNull
     MerkleRoot copy();
+
+    void createSnapshot(Path targetPath);
```
litt3, line 73: missing doc

litt3, line 73: param final and annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package com.swirlds.platform.state;
+
+import com.swirlds.common.context.PlatformContext;
+import com.swirlds.common.metrics.RunningAverageMetric;
+import com.swirlds.metrics.api.Metrics;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public class MerkleRootSnapshotMetrics {
```
litt3, line 24: class doc

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
             Thread.currentThread().interrupt();
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void createSnapshot(Path targetPath) {
```
litt3, line 1109: param final and nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
...
+        Time time = platformContext.getTime();
+        final long start = time.nanoTime();
+        throwIfMutable();
+        throwIfDestroyed();
+        createSnapshot(this, targetPath);
+        snapshotMetrics.getWriteStateToDiskTimeMetric().update(TimeUnit.NANOSECONDS.toMillis(time.nanoTime() - start));
+    }
+
+    static void createSnapshot(MerkleRoot merkleRoot, Path targetPath) {
```
litt3, line 1118: This file is huge. Might be a good idea to split this (and the other) static helper methods into a separate utility class

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
             Thread.currentThread().interrupt();
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void createSnapshot(Path targetPath) {
+        Time time = platformContext.getTime();
```
litt3, line 1110: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
litt3, line 1118: params final and nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
...
+        final long start = time.nanoTime();
+        throwIfMutable();
+        throwIfDestroyed();
+        createSnapshot(this, targetPath);
+        snapshotMetrics.getWriteStateToDiskTimeMetric().update(TimeUnit.NANOSECONDS.toMillis(time.nanoTime() - start));
+    }
+
+    static void createSnapshot(MerkleRoot merkleRoot, Path targetPath) {
+        long round = merkleRoot.getReadablePlatformState().getRound();
```
litt3, line 1119: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
...
+            logger.error(
+                    EXCEPTION.getMarker(),
+                    "Unable to write a snapshot on demand for round {} to {}.",
+                    round,
+                    targetPath,
+                    e);
+        }
+    }
+
+    private static void writeMerkleRootToFile(final Path directory, final MerkleRoot merkleRoot) throws IOException {
```
litt3, line 1138: param nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1101,51 @@ public void computeHash() {
...
+        writeAndFlush(
+                directory.resolve(SIGNED_STATE_FILE_NAME), out -> writeMerkleRootToStream(out, directory, merkleRoot));
+    }
+
+    private static void writeMerkleRootToStream(
+            final MerkleDataOutputStream out, final Path directory, final MerkleRoot merkleRoot) throws IOException {
```
litt3, line 1144: param nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -246,6 +248,13 @@ public String getInfoString(final int hashDepth) {
         return createInfoString(hashDepth, platformState, getHash(), this);
     }
 
+    @Override
+    public void createSnapshot(Path targetPath) {
```
litt3, line 252: param final and nullity

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
```diff
@@ -96,4 +97,12 @@ default Hash getHash() {
     default void computeHash() {
         throw new UnsupportedOperationException();
     }
+
+    /**
+     * Creates a snapshots for the state. The state has to be hashed and immutable before calling this method.
+     * @param targetPath The path to save the snapshot.
+     */
+    default void createSnapshot(Path targetPath) {
```
litt3, line 105: final, nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -199,6 +200,7 @@ public int getVersion() {
     public MerkleRoot copy() {
         throwIfImmutable();
         throwIfDestroyed();
+        setImmutable(true);
```
litt3, line 203: I'm not understanding why this wasn't necessary before, but is necessary now. Was it a bug before? If not, what about your changes makes this a requirement?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: >Can this logic be baked into the readStateFileData utility method? 

This is how I initially implemented it but then I realized that I will not be able to reuse `SignedStateFileReader#readStateFileData` in this case. Currently, the method may be used in the context of saving the signed state and saving isolated Merkle tree without everything else. If sig set logic was baked in here, I'd have to take care of the fact that sig set files could be absent due to the fact that this is a snapshot of a Merkle tree, not a full signed state. 

I think checking for absence in `readStateFileData` would be a bit more confusing than having this logic defined here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: >Was it a bug before?

Yes. It most definitely was a bug, and we didn't have any side effects of that only because this State is used in the test code, not prod. With the necessity of creating snapshots, it revealed itself. So, I fixed it.

platform-sdk/swirlds-state-api/src/main/java/com/swirlds/state/State.java
imalygin: Fixed

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/io/streams/SerializableStreamConstants.java
imalygin: Forgot to roll it back. Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Added

imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+            Path stateFile, CheckedBiFunction<MerkleDataInputStream, Path, MerkleRoot, IOException> snapshotStateReader)
+            throws IOException {
+        return deserializeAndDebugOnFailure(
+                () -> new BufferedInputStream(new FileInputStream(stateFile.toFile())),
+                (final MerkleDataInputStream in) -> {
+                    int fileVersion = readAndCheckStateFileVersion(in);
+
+                    final Path directory = stateFile.getParent();
+                    if (fileVersion == INIT_FILE_VERSION) {
+                        try {
```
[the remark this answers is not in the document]
imalygin: Agreed. I extracted two helper methods, as you suggested

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileUtils.java
```diff
@@ -44,10 +47,19 @@ public final class SignedStateFileUtils {
...
+    /**
+     * The previous version of the signed state file
+     */
+    public static final int INIT_FILE_VERSION = 1;
+
     /**
      * The current version of the signed state file
      */
-    public static final int FILE_VERSION = 1;
+    public static final int SIG_SET_SEPARATE_VERSION = 2;
```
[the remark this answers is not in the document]
imalygin: Done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileUtils.java
```diff
@@ -44,10 +47,19 @@ public final class SignedStateFileUtils {
...
+
+    public static final Set<Integer> SUPPORTED_STATE_FILE_VERSIONS =
+            Set.of(INIT_FILE_VERSION, SIG_SET_SEPARATE_VERSION);
+    public static final Set<Integer> SUPPORTED_SIGSET_VERSIONS = Set.of(INIT_FILE_VERSION);
```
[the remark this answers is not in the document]
imalygin: I renamed the original constants to make them state file specific and created a separate one for the sig set file.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
imalygin: No, I'm not so sure if it's the right way to go. The absence of this code causes a bunch of tests to fail. But fixing those tests may be the right thing to do, as we should not allow snapshots creation using mutable state.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Added

imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Done, The logic is in `MerkleTreeSnapshotWriter` now

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
         return returnState;
     }
 
+    public static StateFileData readStateFileData(
+            Path stateFile, CheckedBiFunction<MerkleDataInputStream, Path, MerkleRoot, IOException> snapshotStateReader)
+            throws IOException {
+        return deserializeAndDebugOnFailure(
+                () -> new BufferedInputStream(new FileInputStream(stateFile.toFile())),
+                (final MerkleDataInputStream in) -> {
+                    int fileVersion = readAndCheckStateFileVersion(in);
```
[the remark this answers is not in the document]
imalygin: Done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +95,73 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
 
         returnState = new DeserializedSignedState(
-                newSignedState.reserve("SignedStateFileReader.readStateFile()"), data.hash());
+                newSignedState.reserve("SignedStateFileReader.readStateFile()"), normalizedData.hash);
 
         return returnState;
     }
 
+    public static StateFileData readStateFileData(
+            Path stateFile, CheckedBiFunction<MerkleDataInputStream, Path, MerkleRoot, IOException> snapshotStateReader)
```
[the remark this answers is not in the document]
imalygin: Done

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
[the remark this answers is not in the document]
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
```diff
@@ -112,28 +111,22 @@ public static void writeMetadataFile(
...
-     * @param directory   the directory to write to
-     * @param signedState the signed state to write
+     * Write the signature set file.
+     * @param directory the directory to write to
+     * @param signedState the signature set file
      */
-    public static void writeStateFile(final Path directory, final SignedState signedState) throws IOException {
-        writeAndFlush(
-                directory.resolve(SIGNED_STATE_FILE_NAME), out -> writeStateFileToStream(out, directory, signedState));
+    public static void writeSignatureSetFile(Path directory, SignedState signedState) throws IOException {
```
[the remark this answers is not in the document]
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
```diff
@@ -112,28 +111,22 @@ public static void writeMetadataFile(
      * Write a {@link SignedState} to a stream.
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
imalygin: @lpetrovic05 what do you think? Is this approach justified? That is, we can force the state to become immutable, if it comes to creation of a snapshot

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +104,104 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+                (final MerkleDataInputStream in) -> {
+                    final int fileVersion = readAndCheckStateFileVersion(in);
+
+                    final Path directory = stateFile.getParent();
+                    if (fileVersion == INIT_STATE_FILE_VERSION) {
+                        return readStateFileDataV1(stateFile, snapshotStateReader, in, directory, fileVersion);
+                    } else if (fileVersion == SIG_SET_SEPARATE_STATE_FILE_VERSION) {
+                        return createStateFileDataV2(stateFile, snapshotStateReader, in, directory, fileVersion);
+                    } else {
+                        throw new IOException("Unsupported protocol version: " + fileVersion);
```
lpetrovic05, line 164: wrong exception message, should be `Unsupported state file version`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileWriter.java
lpetrovic05: I think that Austin is right. We should not silently create a copy of a state, this can cause all sorts of issues like memory leaks and caller expecting his state to remain mutable. I think fixing those tests should be the way to go.
Also, why is it a problem to write a mutable state to disk?

imalygin: >Also, why is it a problem to write a mutable state to disk?

The state is supposed to be hashed first, as far as I understand. And we can hash only an immutable state.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
litt3: I think it's confusing that `readStateFileData` returns a `StateFileData`, which may or may not be complete, depending in the version of the state file. I think it would be better for `readStateFileData` to always return a complete `StateFileData`, regardless of the version of the state file. 

If we wanted `readStateFileData` to not consider signatures, I think we wouldn't want it to return a `StateFileData` object, which is signature aware. 

(just IMO, up to you, as it isn't a huge deal)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1094,20 @@ public void computeHash() {
             Thread.currentThread().interrupt();
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void createSnapshot(@NonNull Path targetPath) {
```
litt3, line 1102: final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+        writeMerkleRootToDiskTime = metrics.getOrCreate(WRITE_MERKLE_ROOT_TO_DISK_TIME_CONFIG);
+    }
+
+    /**
+     * Get a metric tracking the average time required to write a Merkle tree to disk.
+     *
+     * @return the metric tracking the average time required to write a Merkle tree to disk
+     */
+    @NonNull
+    public RunningAverageMetric getWriteStateToDiskTimeMetric() {
```
litt3, line 50: IMO it would be better to just make a method that accepts a long and updates the metric, as opposed to making the caller access the metric object directly

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -68,4 +69,9 @@ public interface MerkleRoot extends MerkleInternal {
     /** {@inheritDoc} */
     @NonNull
     MerkleRoot copy();
+
+    /* Creates a snapshots for the state. The state has to be hashed and immutable before calling this method.
```
litt3, line 73: incorrect javadoc formatting (missing second *)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -68,4 +69,9 @@ public interface MerkleRoot extends MerkleInternal {
...
+     * @param targetPath The path to save the snapshot.
+     */
+    void createSnapshot(final Path targetPath);
```
litt3, line 76: missing nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1094,20 @@ public void computeHash() {
...
+        throwIfMutable();
+        throwIfDestroyed();
+        final long start = System.nanoTime();
```
litt3, line 1105: we shouldn't use system time, since it's difficult to test

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -144,13 +146,20 @@ public class MerkleStateRoot extends PartialNaryMerkleInternal
 
     private final Function<SemanticVersion, SoftwareVersion> versionFactory;
     private MerkleCryptography merkleCryptography;
+    private PlatformContext platformContext;
```
litt3, line 149: member variable not necessary

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -144,13 +146,20 @@ public class MerkleStateRoot extends PartialNaryMerkleInternal
...
+    private Platform platform;
```
litt3, line 150: member variable not necessary

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotWriter.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+public final class MerkleTreeSnapshotWriter {
+
+    private static final Logger logger = LogManager.getLogger(MerkleTreeSnapshotWriter.class);
+
+    private MerkleTreeSnapshotWriter() {
+        // prevent instantiation
+    }
+
+    static void createSnapshot(@NonNull final MerkleRoot merkleRoot, @NonNull final Path targetPath) {
```
litt3, line 44: missing javadoc

rbarkerSL (dismissed):
Approval applies to `hedera-node/hedera-app/build.gradle.kts` and `platform-sdk/swirlds-common/src/main/java/module-info.java`

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/StateFileManagerTests.java
```diff
@@ -441,4 +449,8 @@ void stateDeletionTest() throws IOException {
             validateSavingOfState(fatalState, fatalDirectory);
         }
     }
+
+    private static void makeImmutable(SignedState signedState) {
```
litt3, line 453: final, nullity

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +104,104 @@
...
+                    } else if (fileVersion == SIG_SET_SEPARATE_STATE_FILE_VERSION) {
+                        return createStateFileDataV2(stateFile, snapshotStateReader, in, directory, fileVersion);
+                    } else {
+                        throw new IOException("Unsupported state file version: " + fileVersion);
+                    }
+                });
+    }
+
+    @NonNull
+    private static StateFileData createStateFileDataV2(
```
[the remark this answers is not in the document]
imalygin: Sure, makes sense. I added javadoc and removed redundant `fileVersion` parameter, as the method itself is version-specific.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -87,43 +104,104 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
+            final Hash hash = in.readSerializable();
+            return new StateFileData(state, hash, null, fileVersion);
+
+        } catch (final IOException e) {
+            throw new IOException("Failed to read snapshot file " + stateFile.toFile(), e);
+        }
+    }
+
+    @NonNull
+    private static StateFileData readStateFileDataV1(
```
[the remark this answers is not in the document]
imalygin: Agreed. I renamed the method.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotWriter.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: Removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Added

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
imalygin: Sure, makes sense. I replaced this method with `updateWriteStateToDiskTimeMetric`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
```diff
@@ -42,12 +43,10 @@ public MerkleRootSnapshotMetrics(@NonNull final PlatformContext platformContext)
...
-     *
-     * @return the metric tracking the average time required to write a Merkle tree to disk
+     * Update the metric tracking the average time required to write a Merkle tree to disk.
+     * @param timeTakenMs the time taken to write the state to disk
      */
-    @NonNull
-    public RunningAverageMetric getWriteStateToDiskTimeMetric() {
-        return writeMerkleRootToDiskTime;
+    public void updateWriteStateToDiskTimeMetric(long timeTakenMs) {
+        writeMerkleRootToDiskTime.update(TimeUnit.MILLISECONDS.toMillis(timeTakenMs));
```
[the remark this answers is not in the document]
imalygin: You're right. It's redundant.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRootSnapshotMetrics.java
```diff
@@ -42,12 +43,10 @@ public MerkleRootSnapshotMetrics(@NonNull final PlatformContext platformContext)
...
-     * Get a metric tracking the average time required to write a Merkle tree to disk.
-     *
-     * @return the metric tracking the average time required to write a Merkle tree to disk
+     * Update the metric tracking the average time required to write a Merkle tree to disk.
+     * @param timeTakenMs the time taken to write the state to disk
      */
-    @NonNull
-    public RunningAverageMetric getWriteStateToDiskTimeMetric() {
-        return writeMerkleRootToDiskTime;
+    public void updateWriteStateToDiskTimeMetric(long timeTakenMs) {
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -235,8 +243,11 @@ public void init(
             @NonNull final Platform platform,
             @NonNull final InitTrigger trigger,
             @Nullable final SoftwareVersion deserializedVersion) {
-        metrics = platform.getContext().getMetrics();
-        merkleCryptography = platform.getContext().getMerkleCryptography();
+        PlatformContext platformContext = platform.getContext();
```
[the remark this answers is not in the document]
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1082,4 +1093,18 @@ public void computeHash() {
...
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void createSnapshot(@NonNull final Path targetPath) {
+        throwIfMutable();
+        throwIfDestroyed();
+        final Long startTime = time == null ? null : time.currentTimeMillis();
```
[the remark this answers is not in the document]
imalygin: I tried to address your comment.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1092,4 +1112,18 @@ public void computeHash() {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void createSnapshot(@NonNull final Path targetPath) {
+        throwIfMutable();
+        throwIfDestroyed();
+        final long startTime = time.currentTimeMillis();
+        MerkleTreeSnapshotWriter.createSnapshot(this, targetPath);
+        if (snapshotMetrics != null && time != null) {
```
litt3, line 1125: IMO we should remove these null checks, and just update the metric. If the metric call fails because the object wasn't correctly initialized, then that's fine. 

Or, the alternative would be to do the null check, and throw an exception explaining what happened if one of these is null. What I don't think we should do is quitely continue operating, when something that definitely shouldn't be null *is* null

lpetrovic05 (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: okay, I removed the null checks from here. Also, I added a no-arg constructor for `MerkleRootSnapshotMetrics`. If the objects is created with it, it's no-op.

litt3 (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1009,6 +1021,24 @@ public PlatformStateAccessor getReadablePlatformState() {
         return readablePlatformStateStore();
     }
 
+    /**
+     * Sets the Snapshot Metrics for this state
+     *
+     * @param snapshotMetrics the metrics to set
+     */
+    public void setSnapshotMetrics(MerkleRootSnapshotMetrics snapshotMetrics) {
```
litt3, line 1029: nullity annotation, final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1026,7 +1026,8 @@ public PlatformStateAccessor getReadablePlatformState() {
      *
      * @param snapshotMetrics the metrics to set
      */
-    public void setSnapshotMetrics(MerkleRootSnapshotMetrics snapshotMetrics) {
+    @NonNull
```
litt3, line 1029: void methods don't get annotations - the needed nullity annotation I was referring to was on the parameter, `snapshotMetrics`

thenswan: oh sure, just updated

litt3 (approved):

jeromy-cannon (dismissed):

lpetrovic05 (approved):

litt3 (approved):

rbarkerSL (approved):
Review covers

- `hedera-node/hedera-app/build.gradle.kts`
- `platform-sdk/swirlds-common/src/main/java/module-info.java`
- `platform-sdk/swirlds-platform-core/src/testFixtures/java/module-info.java`

LGTM

artemananiev (approved):

Neeharika-Sompalli (approved):
LGTM from hedera-base. Thanks @imalygin

PR #15544, opened 2024-09-18
refactor: 15413 Split `PlatformStateAccessor` into two interfaces

### LINKED ISSUE #15413: Split `PlatformStateAccessor` into two interfaces - read-only and writable
### Problem

This [PR](https://github.com/hashgraph/hedera-services/pull/15389) introduces two additional methods - `MerkleRoot#getWritablePlatformState` and `MerkleRoot#getReadablePlatformState` instead of a single `MerkleRoot#getPlatformState` to make sure that the writable state only used when it's really needed to prevent possible thread races described in #15385 

However, the fact that a readable instance of `PlatformStateAccessor` still has modifiers is confusing

Additionally, introduce a new method for initializing the state accessor that is more intuitive. Currently, the method `getWriteablePlatformState()` is called to initialize the state, even though the returned value is not used at all.

### Solution

Split `PlatformStateAccessor` into two interfaces - `PlatformStateAccessor` and `PlaftormStateModifier`, and `PlatformStateModifier` should extend `PlatformStateAccessor`. This way we can make sure that the object is not misused in a compile time (vs runtime, as it's implemented now). 

Create a new `initialize()` method that performs initialization and does not invoke either of the `getPlatformState` methods.

### Alternatives

_No response_

imalygin:
**Description**:

This is a follow-up for https://github.com/hashgraph/hedera-services/pull/15389 This PR introduce a new interface: `PlatformStateModifier`. Now, `PlatformStateAccessor` has only getters and `PlatformStateModifier` has modifying methods. 

Also, this PR introduces `initPlatformState` to get rid of confusing usages of `getWritablePlatformState` for the sake of initialization. 

**Related issue(s)**:

Fixes #15413 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
@@ -33,6 +33,14 @@ public interface MerkleRoot extends MerkleInternal {
     SwirldState getSwirldState();
 
     /**
+     * This method makes sure that the platform state is initialized.
+     * If it's already initialized, it does nothing.
+     */
+    default void initPlatformState() {
+        // do nothing
+    }
+
+    /**
      * Get readable platform state.
      * Works on both - mutable and immutable {@link MerkleRoot} and, therefore, this method should be preferred.
      *
@@ -48,14 +56,14 @@ public interface MerkleRoot extends MerkleInternal {
      * @return mutable platform state
      */
     @NonNull
-    PlatformStateAccessor getWritablePlatformState();
+    PlatformStateModifier getWritablePlatformState();
 
     /**
      * Set the platform state.
      *
      * @param platformState the platform state
      */
-    void updatePlatformState(@NonNull final PlatformStateAccessor platformState);
+    void updatePlatformState(@NonNull final PlatformStateModifier platformState);
 
     /**
      * Generate a string that describes this state.
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
@@ -36,9 +36,7 @@ public interface MerkleRoot extends MerkleInternal {
      * This method makes sure that the platform state is initialized.
      * If it's already initialized, it does nothing.
      */
-    default void initPlatformState() {
-        // do nothing
-    }
+    void initPlatformState();
 
     /**
      * Get readable platform state.
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
@@ -390,7 +390,7 @@ public class MerkleStateRoot extends PartialNaryMerkleInternal
      * {@inheritDoc}
      */
     @Override
-    public void handleConsensusRound(@NonNull final Round round, @NonNull final PlatformStateAccessor platformState) {
+    public void handleConsensusRound(@NonNull final Round round, @NonNull final PlatformStateModifier platformState) {
         throwIfImmutable();
         lifecycles.onHandleConsensusRound(round, this);
     }
@@ -1014,7 +1014,7 @@ public class MerkleStateRoot extends PartialNaryMerkleInternal
      */
     @NonNull
     @Override
-    public PlatformStateAccessor getWritablePlatformState() {
+    public PlatformStateModifier getWritablePlatformState() {
         if (isImmutable()) {
             throw new IllegalStateException("Cannot get writable platform state when state is immutable");
         }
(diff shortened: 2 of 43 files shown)


mxtartaglia-sl (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
```diff
@@ -75,7 +75,7 @@ public WritablePlatformStateStore(@NonNull final WritableStates writableStates)
     /**
      * Overwrite the current platform state with the provided state.
      */
-    public void setAllFrom(@NonNull final PlatformStateAccessor accessor) {
+    public void setAllFrom(@NonNull final PlatformStateModifier accessor) {
```
litt3, line 78: nit: param still called `accessor`, but this is now a `modifier`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
```diff
@@ -32,6 +32,14 @@ public interface MerkleRoot extends MerkleInternal {
     @NonNull
     SwirldState getSwirldState();
 
+    /**
+     * This method makes sure that the platform state is initialized.
+     * If it's already initialized, it does nothing.
+     */
+    default void initPlatformState() {
```
litt3, line 39: My personal preference would be to *not* provide the empty default implementation, and just have an empty implementation for impls that don't need to do anything.

Empty default methods can be bug prone, since any new impls might mistakenly miss the method.

Up to you though

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
```diff
@@ -1014,20 +1014,28 @@ public PlatformStateAccessor getReadablePlatformState() {
...
         return writablePlatformStateStore();
     }
 
     /**
-     * Updates the platform state with the values from the provided instance of {@link PlatformStateAccessor}
+     * {@inheritDoc}
+     */
+    @Override
+    public void initPlatformState() {
+        writablePlatformStateStore();
```
litt3, line 1029: Perhaps it would be worth extracting the relevant logic from `writablePlatformStateStore`. Looking at that method, it appears there is some stuff in the current `writablePlatformStateStore` method which doesn't actually need to be done on init

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateModifier.java
```diff
@@ -0,0 +1,132 @@
...
+     * @param lastFrozenTime the last freezeTime based on which the nodes were frozen
+     */
+    void setLastFrozenTime(@Nullable Instant lastFrozenTime);
+
+    /**
+     * Set the first software version where the birth round migration happened.
+     *
+     * @param firstVersionInBirthRoundMode the first software version where the birth round migration happened
+     */
+    @NonNull
```
litt3, line 109: void method shouldn't have return val nullity annotation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
```diff
@@ -164,14 +164,14 @@ public PlatformState getWritablePlatformState() {
      * @param platformStateAccessor the platform state
      */
     @Override
-    public void updatePlatformState(@NonNull final PlatformStateAccessor platformStateAccessor) {
+    public void updatePlatformState(@NonNull final PlatformStateModifier platformStateAccessor) {
```
litt3, line 167: nit: param name is outdated

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritablePlatformStateStore.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleRoot.java
imalygin: Sure, I removed default empty implementation

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleStateRoot.java
imalygin: I replaced it with more specific call

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/PlatformStateModifier.java
imalygin: Removed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/State.java
imalygin: Fixed

litt3 (approved):
lgtm

mxtartaglia-sl (approved):

litt3 (approved):

mxtartaglia-sl (approved):

Issue #11211, opened 2024-01-26
Permit unpaid queries when executed from localhost


rbair23:
### Problem

A node operator should be able to query the state on their node. It would be very useful to get these queries from the grpc endpoint rather than digging through local state files.

### Solution

If a configuration setting is enabled (which it would be by default), open a third local port (we already open two, one for TLS, one without). The third local port would be blocked at the firewall. The local port will be bound ONLY to 127.0.0.1 (not 0.0.0.0) so it can only be accessed locally. When queried through this port, the query will be free.

Opening an additional port should be pretty easy. The challenge is going to be recognizing in the query workflow what the source of the transaction was. We could update the QueryWorkflow to take a third argument as to whether it should be free, and have a new QueryMethod instantiated with an argument to indicate it is a free query, tied to the new opened local port.

PR #15554, opened 2024-09-19
feat: permit unpaid queries when executed from localhost

MiroslavGatsanoga:
**Description**:

- Update `NettyGrpcServerManager` to run additional localhost-only node operator gRPC server on a separate port that can be configured with `grpc.nodeOperatorPort` config property
- The node operator server will only have query methods registered
- Enabling the node operator server is behind a `grpc.nodeOperatorPortEnabled` feature flag, currently defaulting to `false`
- Update `QueryWorkflow.handleQuery` method to accept additional boolean parameter that indicates if the workflow should charge for handling queries (the node operator server is should not be charged or throttled when making queries)
- Update `test-clients` so that `SubProcessNetwork`'s nodes can have node operator servers listening on different ports

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/11211

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/QueryMethod.java
```diff
@@ -51,6 +51,11 @@ public final class QueryMethod extends MethodBase {
     /** A metric for the calls per second that queries were answered */
     private final SpeedometerMetric queriesAnsweredSpeedometer;
 
+    /**
+     * Indicates if the QueryWorkflow should charge for handling queries.
+     */
+    private final boolean shouldCharge;
```
tinker-michaelj, line 57: Is there a reason to make this a per-`QueryMethod` field instead of using an all-or-nothing config inside `QueryWorkflow`?

(I.e., would it make sense to charge for some methods and not others?)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/GrpcServiceBuilder.java
```diff
@@ -115,14 +116,14 @@ final class GrpcServiceBuilder {
      */
     public GrpcServiceBuilder(
             @NonNull final String serviceName,
-            @NonNull final IngestWorkflow ingestWorkflow,
+            @Nullable final IngestWorkflow ingestWorkflow,
```
tinker-michaelj, line 119: Imo it makes more sense to keep this `@NonNull` since this builder **must** have a non-null `IngestWorkflow` for HAPI and it's harmless to construct the query-only node operator `Server` with a reference to the `IngestWorkflow`

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/NettyGrpcServerManager.java
```diff
@@ -133,9 +154,28 @@ public NettyGrpcServerManager(
...
+                    return builder.build(metrics, true);
                 })
                 .collect(Collectors.toUnmodifiableSet());
+
+        final var grpcConfig = configProvider.getConfiguration().getConfigData(GrpcConfig.class);
+        if (grpcConfig.nodeOperatorPortEnabled()) {
+            // Convert the various RPC service definitions into query endpoints permitting unpaid queries for node
+            // operators
+            nodeOperatorServices = rpcServiceDefinitions
+                    .get()
```
tinker-michaelj, line 166: Could extract this chain to a private method that accepts an extra filter on what methods to register or something.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/props/NodeConnectInfo.java
```diff
@@ -29,6 +29,7 @@ public class NodeConnectInfo {
     public static int NEXT_DEFAULT_ACCOUNT_NUM = 3;
     private static final int DEFAULT_PORT = 50211;
     private static final int DEFAULT_TLS_PORT = 50212;
+    private static final int DEFAULT_NODE_OPERATOR_PORT = 50213;
```
tinker-michaelj, line 32: I would suggest removing changes to this class; although not marked `@Deprecated`, it really should be eliminated in favor of getting node connection info directly from a `HederaNode` object.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/HederaNode.java
```diff
@@ -121,7 +128,8 @@ CompletableFuture<Void> statusFuture(
      * @return this node's HAPI spec identifier
      */
     default String hapiSpecInfo() {
-        return getHost() + ":" + getGrpcPort() + ":0.0." + getAccountId().accountNumOrThrow();
+        return getHost() + ":" + getGrpcPort() + ":" + getGrpcNodeOperatorPort() + ":0.0."
```
tinker-michaelj, line 131: I would revert this change (and the changes to `NodeConnectInfo` below).

tinker-michaelj (dismissed):
Overall LGTM, nice work @MiroslavGatsanoga !

Trying to imagine both, I think an implementation with two `QueryWorkflow`'s (one dedicated to HAPI queries, one dedicated to node operator queries) would be cleaner than using a `@Singleton` workflow and changing its behavior with an extra argument `chargeQueries` flag passed to every response.

But that's more stylistic...if you agree, could be done in a later issue.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/QueryMethod.java
Neeharika-Sompalli: > implementations with @UserQueries and @OperatorQueries qualifiers instead of a @Singleton...might be cleaner, not sure.

Agreed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/NettyGrpcServerManager.java
```diff
@@ -184,14 +229,28 @@ public synchronized void start() {
...
+            try {
+                final var nodeOperatorPort = grpcConfig.nodeOperatorPort();
+                logger.info("Starting node operator gRPC server on port {}", nodeOperatorPort);
+                nettyBuilder = builderFor(nodeOperatorPort, nettyConfig, profile, true);
+                nodeOperatorServer =
+                        startServerWithRetry(nodeOperatorServices, nettyBuilder, startRetries, startRetryIntervalMs);
+                logger.info("Node operator gRPC server listening on port {}", nodeOperatorServer.getPort());
+            } catch (Exception e) {
+                nodeOperatorServer = null;
+                logger.warn("Could not start node operator gRPC server, will continue without it: {}", e.getMessage());
```
Neeharika-Sompalli, line 251: Seems like most of this code is duplicated from above. Can be extracted to a method?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/query/QueryWorkflow.java
```diff
@@ -29,6 +29,7 @@ public interface QueryWorkflow {
      *
      * @param requestBuffer The raw protobuf query bytes. Must be a {@link Query} object.
      * @param responseBuffer The raw protobuf response bytes.
+     * @param shouldCharge If the workflow should charge for handling the query.
      */
-    void handleQuery(@NonNull Bytes requestBuffer, @NonNull BufferedData responseBuffer);
+    void handleQuery(@NonNull Bytes requestBuffer, @NonNull BufferedData responseBuffer, boolean shouldCharge);
```
Neeharika-Sompalli, line 34: yeah agree with Michael, having two workflows makes it clear.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/props/NodeConnectInfo.java
```diff
@@ -52,11 +54,15 @@ public NodeConnectInfo(String inString) {
         } else {
             port = DEFAULT_PORT;
         }
+
+        tlsPort = DEFAULT_TLS_PORT;
```
Neeharika-Sompalli, line 58: Don't you still need this
```
if (ports.length > 1) {
            tlsPort = ports[1];
        } else {
            tlsPort = DEFAULT_TLS_PORT;
        }

Neeharika-Sompalli (approved):
The review comments can be addressed in a later PR if you think that is appropriate. LGTM otherwise. Thanks @MiroslavGatsanoga

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/GrpcServiceBuilder.java
MiroslavGatsanoga: I wanted to make sure that the query-only node operator Server will not be able to execute transactions, but yes now it seems like the HAPI one might accept null `IngestWorkflow`. I will revert it back to `@NonNull`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty/NettyGrpcServerManager.java
MiroslavGatsanoga: Updated.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/HederaNode.java
MiroslavGatsanoga: Reverted.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/props/NodeConnectInfo.java
MiroslavGatsanoga: Reverted the changes and added `@Deprecated` with comment to use `HederaNode` object for  getting node connection info.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/props/NodeConnectInfo.java
MiroslavGatsanoga: Reverted changes in `NodeConnectInfo` as @tinker-michaelj suggested since it's usage is not recommended going forward.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/query/QueryWorkflow.java
MiroslavGatsanoga: Will be addressed as mentioned [above](https://github.com/hashgraph/hedera-services/pull/15554#discussion_r1782821754).

tinker-michaelj (approved):
LGTM, tyvm @MiroslavGatsanoga !

Issue #15448, opened 2024-09-12
In-memory virtual maps support


artemananiev:
`VirtualMap` doesn't flush all the entities to disk immediately. Recently changed (added, updated, or even deleted) entities are kept in `VirtualNodeCache` and flushed to disk once in a while. For some maps, flush periods may be hours. During these periods between flushes, some map entities may be added and then removed. Such entities should never be flushed to disk. If all map entities are like this, there should be no flushes at all!

It may look like a corner case, but there is at least one important use case like this: maps used as backing stores for `QueueNode`s. See https://github.com/hashgraph/hedera-services/issues/15090 for details. It's critical to make sure the corresponding maps are never flushed to disk (unless they are so huge that queue size exceeds flush threshold, despite some elements will be removed later).

This ticket is to implement this in-memory more for virtual maps. A few approaches have been discussed, and the following approach looks the most promising:

* Virtual node cache copies are merged as usually, till total estimated size of the oldest copy exceeds flush threshold
* Instead of just flushing this copy as we do today, the cache can be inspected to contain any redundant mutations. In particular, some entities that are added, then modified, then removed - all between flushes
* With redundant mutations identified, another check for cache's estimated size is made. If it's still more than the threshold, the copy is flushed to disk
* If the new estimation is below the threshold, no flush is made, but the mutations are removed from the cache. Something similar to merge + garbage collect

With this approach, cache copy merges are still fast. Searching for redundant mutations is what we already do before flush. Garbage collection should be faster than actual flushing. Total number of flushes will be reduced (for some maps - to zero).

PR #15575, opened 2024-09-20
feat: 15448: In-memory virtual maps support

artemananiev:
Fix summary:

* In-memory virtual maps support is implemented

Fixes: https://github.com/hashgraph/hedera-services/issues/15448


platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -990,13 +1019,68 @@ public Stream<VirtualHashRecord> dirtyHashesForFlush(final long lastLeafPath) {
...
+     * just removes all filtered mutations using this method.
      */
-    public long estimatedInternalsCount(final long firstLeafPath) {
-        return (dirtyHashes == null) ? 0 : dirtyHashes.size();
+    public void garbageCollect(final long firstLeafPath, final long lastLeafPath) {
+        final Stream<Mutation<Long, Hash>> filteredHashes = dirtyHashes.stream()
+                .filter(mutation -> mutation.key <= lastLeafPath)
+                .filter(mutation -> !mutation.isFiltered())
+                .peek(mutation -> mutation.next = null);
+        dirtyHashes = new ConcurrentArray<>(filteredHashes);
```
anthony-swirldslabs, line 1067: Minor: this is a public method w/o any concurrency contract defined in its specification, and it modifies a field of this class which may cause issues if this field is accessed from other methods, potentially on other threads. Should the method be private? Should it define its threading model? Should the access to this field be protected/synchronized by any means?

Same comment applies to a few other similar assignments below in this same method.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -990,13 +1019,68 @@ public Stream<VirtualHashRecord> dirtyHashesForFlush(final long lastLeafPath) {
...
+     * filtered the size drops below the threshold, the copy is not flushed to disk, but
+     * just removes all filtered mutations using this method.
      */
-    public long estimatedInternalsCount(final long firstLeafPath) {
-        return (dirtyHashes == null) ? 0 : dirtyHashes.size();
+    public void garbageCollect(final long firstLeafPath, final long lastLeafPath) {
+        final Stream<Mutation<Long, Hash>> filteredHashes = dirtyHashes.stream()
+                .filter(mutation -> mutation.key <= lastLeafPath)
+                .filter(mutation -> !mutation.isFiltered())
+                .peek(mutation -> mutation.next = null);
```
anthony-swirldslabs, line 1066: The consequences of this operation is unclear. From the look of it, some sort of a linked list gets cut. However, original objects remain in the original `dirtyHashes` collection. Also, assuming that the original linked list is not circular, it's unclear how the object with `next` cleared here would differ from the actual tail of the original list. I suggest to add a clarifying comment in code to explain what exactly the code is trying to achieve here and where/how the (potentially) nullified `next` value is going to be used in other places.

Same comment applies to a couple of very similar streams below in this same method.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -1305,23 +1394,58 @@ private static <K, V> void purge(final ConcurrentArray<Mutation<K, V>> array, fi
...
+                    if (latestMutationUpToVersion.isDeleted()) {
+                        assert !latestMutationUpToVersion.isFiltered();
+                        // If the latest mutation up to newestVersion is "deleted", and there are no
+                        // newer mutations, the whole entry for the key can be removed from the index.
+                        // It's safe to do so here, as there are no references to copies older than
+                        // newestVersion and there are no mutations in versions newer than newestVersion
+                        if (latestMutation == latestMutationUpToVersion) {
+                            latestMutationUpToVersion.setFiltered();
+                            filteredCounter.incrementAndGet();
+                            index.remove(latestMutation.key);
```
anthony-swirldslabs, line 1442: Minor q: if there're mutations for the same key in versions greater than the `newestVersion`, would removing the key from the index affect those later versions in any way?

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -1305,23 +1394,58 @@ private static <K, V> void purge(final ConcurrentArray<Mutation<K, V>> array, fi
...
+                        }
+                    } else {
+                        // Propagate the "new" flag to the newer mutation
+                        latestMutationUpToVersion.setNew();
```
anthony-swirldslabs, line 1446: Minor: from the code above and the `lookup` behavior, I assume that there may be mutations in between the `nextMutation`/`latestMutation` and the `latestMutationUpToVersion`. The `isNew` flag is going to be propagated from the `nextMutation` to the `latestMutationUpToVersion`, leaving the interim mutations not marked with `isNew`. I've no idea if this may break any logic, today or in the future, but I just want to call this out to check if this has been considered.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -1394,7 +1518,7 @@ private void serializePathToDirtyHashIndex(
             assert mutation.version <= this.fastCopyVersion.get()
                     : "Trying to serialize pathToDirtyInternalIndex with a version ahead";
             out.writeLong(mutation.version);
-            out.writeBoolean(mutation.isDeleted());
+            out.writeByte(mutation.getFlags());
```
anthony-swirldslabs, line 1521: This looks like a serialized data format change. Is this backward compatible? Does this require any migration? Is the data on disk considered "public" to require a formal design review?

anthony-swirldslabs (commented):
Overall the fix looks good. However, I have a very shallow understanding of the implementation details. So it would be good to get another pair of eyes to take a look here.

Just posting a few comments/questions, mostly to clarify things.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: These are all good questions. VirtualNodeCache has a number of methods that are supposed to be called on the virtual lifecycle thread only. Examples are dirtyLeavesForFlush(), release(), and others. Even more important is the phase of the cache. When a cache is just created, it accepts leaf changes, but hashes cannot be modified. Then cache.prepareForHashing() is called, and it switches the gears: now leaf changes are prohibited, but hashes can be updated. Finally, a cache is sealed using seal() method, and from this point on no changes are allowed. In this particular case, the cache must be sealed. I added the corresponding check to the method, as well as a note to the JavaDoc. It will make sure no changes to dirtyHashes / dirtyLeaves / dirtyLeafPaths are done in parallel.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: It's slightly more complicated than just a linked list... All cache instances share a concurrent map `pathToDirtyHashIndex`, which is a map of paths (longs) to linked lists (hash mutations). In addition to this map, every cache instance contains an array `dirtyHashes`, array elements are individual mutations from the linked lists above. Each mutation in the array points to the next mutation of the same hash, but not to the next element in the array. By setting `mutation.next` to null in this code, the linked lists in the map get trimmed, all elements after certain mutations are removed.

I added a short comment to this `peek()` call:

```java
                // Leave only the latest mutation for every hash path by setting next to null
```

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: All these mutations are effectively just dropped.

Here is an example. Assume for a certain path, the entry in the concurrent map (`index` param for this method) contains a linked list of mutations like this:

```
path: (version3, deleted, null value) -> (version2, value2) -> (version1, new, value1) -> null
```

value1 was inserted in version1, then updated in version2, then removed in version 3. The concurrent array will contain these three mutations, not necessarily next to each other. When mutation for version2 is processed, the next mutation in the list (for version1) is marked as filtered. Since this filtered mutation is "new", the first mutation in the whole linked list is checked, for version3. It's marked as "deleted", and therefore mutation3 is filtered, and the map entry for this path is removed. Some time later (or earlier), mutation3 is processed, it will filter out mutation2, this is fine.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: Check the example in the previous comment. If mutation3 is not "deleted", "new" marker from mutation1 will be propagated to it, but not to version2. This is fine, since version2 (and all subsequent versions) will be removed from the linked list anyway.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: Yes, this is compatible, I double checked. Booleans are written as bytes. No migration is needed.

Previously, the byte was either 1 (deleted) or 0 (not deleted). Now this byte is read as "flags", and then "new" and "deleted" are extracted from it:

```java
            final byte flags = in.readByte();
            final boolean isNew = Mutation.getFlag(flags, Mutation.FLAG_BIT_NEW);
            final boolean isDeleted = Mutation.getFlag(flags, Mutation.FLAG_BIT_DELETED);
```

If the byte was 0, then both "new" and "deleted" are false. If the byte was "1", then "new" is false, and "deleted" is true, as expected. `FLAG_BIT_DELETED` above is 0.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
anthony-swirldslabs: Sounds good.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
anthony-swirldslabs: Got it. Thanks for clarifying!

anthony-swirldslabs (approved):

anthony-swirldslabs (approved):

imalygin (dismissed):

anthony-swirldslabs (approved):

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
```diff
@@ -1011,13 +1047,91 @@ public Stream<VirtualHashRecord> dirtyHashesForFlush(final long lastLeafPath) {
...
+    /**
+     * If a cache copy's estimated size exceeded flush threshold, but after mutations are
+     * filtered the size drops below the threshold, the copy is not flushed to disk, but
+     * just removes all filtered mutations using this method.
+     *
+     * <p>This method can only be called on sealed caches.
+     *
+     * @throws MutabilityException if called on a non-sealed cache instance.
+     */
+    public void garbageCollect(final long firstLeafPath, final long lastLeafPath) {
```
OlegMazurov, line 1113: `firstLeafPath`, `lastLeafPath` are not used in the method.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/ConcurrentArray.java
```diff
@@ -142,6 +142,16 @@ final class ConcurrentArray<T> {
         tail = head;
     }
 
+    /**
+     * Create a new concurrent array with {@link #DEFAULT_ELEMENT_ARRAY_LENGTH} subarray capacity
+     * and elements from the given stream. The created array is immutable.
+     */
+    ConcurrentArray(final Stream<T> from) {
+        this();
+        from.parallel().forEach(this::add);
```
OlegMazurov, line 151: This is very expensive: `add()` contains a `synchronized` block and involves atomic updates.
While initializing a new instance of `ConcurrentArray` we are guaranteed to have exclusive access to it so we could do without synchronization.

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/VirtualNodeCache.java
artemananiev: Removed

platform-sdk/swirlds-virtualmap/src/main/java/com/swirlds/virtualmap/internal/cache/ConcurrentArray.java
artemananiev: I refactored this class a little bit. Current code in the synchronized section in `add()` is extracted to a new method, `addImpl()`. This new method is very straightforward and not thread safe. It can be called from the constructor above or from `add()` with proper synchronization.

anthony-swirldslabs (approved):

anthony-swirldslabs (approved):

OlegMazurov (approved):
No negative performance impact on `NftTransferLoadTest`.

PR #15578, opened 2024-09-22
build: update Java Module patching

jjohannes:
**Description**:

- Update extra-java-module-info to 1.9
- Configure a consistent classpath (new feature in 1.9) to no longer do multiple transformations of the same Jar
- For the consistent/global classpath, use the 'gradle/aggregation' project to configure it (renamed from 'gradle/reports')
- Update Guava and related libraries and remove no longer needed patch rules

This fixes the problem that many Jars are transformed several times, which leads to long IntelliJ import times and the Jars showing up many times in IntelliJ (which looks like this BEFORE this change):

363586854-2f8e8fc3-ea3d-44d2-afe4-8a0121081af2

**Related issue(s)**:

 #14255

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/test-clients/src/main/java/module-info.java
+++ b/hedera-node/test-clients/src/main/java/module-info.java
@@ -102,7 +102,7 @@ module com.hedera.node.test.clients {
     requires com.swirlds.platform.core.test.fixtures;
     requires com.fasterxml.jackson.core;
     requires com.fasterxml.jackson.databind;
-    requires com.github.docker.java.api;
+    requires com.github.dockerjava.api;
     requires com.sun.jna;
     requires io.grpc.netty;
     requires io.grpc.stub;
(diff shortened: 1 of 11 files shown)


hedera-node/test-clients/src/main/java/module-info.java
```diff
@@ -102,7 +102,7 @@
     requires com.swirlds.platform.core.test.fixtures;
     requires com.fasterxml.jackson.core;
     requires com.fasterxml.jackson.databind;
-    requires com.github.docker.java.api;
+    requires com.github.dockerjava.api;
```
jjohannes, line 105: This name is changed, because `com.github.docker.java.api` was our own creation. In the new version, `com.github.dockerjava.api` is the official name.

rbarkerSL (approved):
LGTM

rbarkerSL (approved):

PR #15586, opened 2024-09-25
feat: HIP-991 topic fees

### LINKED ISSUE #14170: HIP-991: TopicIDs with Fees
Implement permissionless revenue-generating topics described in [HIP-991](https://hips.hedera.com/hip/hip-991) (currently under review).
```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-services/issues/16854
- [ ] https://github.com/hashgraph/hedera-services/issues/16855
- [ ] https://github.com/hashgraph/hedera-services/issues/16856
- [ ] https://github.com/hashgraph/hedera-services/issues/16867
- [ ] https://github.com/hashgraph/hedera-services/issues/15420
- [ ] https://github.com/hashgraph/hedera-services/issues/15516
- [ ] https://github.com/hashgraph/hedera-services/issues/15584
- [x] Review HIP proposal and protobuf specification
- [ ] https://github.com/hashgraph/hedera-services/issues/14365
- [ ] https://github.com/hashgraph/hedera-services/issues/15144
- [ ] https://github.com/hashgraph/hedera-services/issues/15523
- [ ] https://github.com/hashgraph/hedera-services/issues/15524
- [ ] https://github.com/hashgraph/hedera-services/issues/15403
- [ ] https://github.com/hashgraph/hedera-services/issues/15328
- [ ] https://github.com/hashgraph/hedera-services/issues/15513
- [ ] https://github.com/hashgraph/hedera-services/issues/15327
- [ ] https://github.com/hashgraph/hedera-services/issues/15536
- [ ] https://github.com/hashgraph/hedera-services/issues/15537
- [ ] https://github.com/hashgraph/hedera-services/issues/17086
- [ ] https://github.com/hashgraph/hedera-services/issues/17284
- [ ] https://github.com/hashgraph/hedera-services/issues/17348
- [ ] https://github.com/hashgraph/hedera-services/issues/17038
- [ ] https://github.com/hashgraph/hedera-services/issues/17393
- [ ] https://github.com/hashgraph/hedera-services/issues/17506
```

vtronkov:
**Description**:
Feature branch for all [HIP-991](https://hips.hedera.com/hip/hip-991) changes

**Related issue(s)**:

Fixes #14170 

**Checklist**

### DIFF
--- a/hapi/hedera-protobufs/services/custom_fees.proto
+++ b/hapi/hedera-protobufs/services/custom_fees.proto
@@ -416,7 +416,7 @@ message CustomFeeLimit {
     AccountID account_id = 1;
 
     /**
-     * A custom fee amount limit.
+     * The maximum fees that the user is willing to pay for the message.
      */
-    FixedFee amount_limit = 2;
+    repeated FixedFee fees = 2;
 }
--- a/hapi/hedera-protobufs/services/response_code.proto
+++ b/hapi/hedera-protobufs/services/response_code.proto
@@ -1702,7 +1702,13 @@ enum ResponseCodeEnum {
     DUPLICATE_DENOMINATION_IN_MAX_CUSTOM_FEE_LIST = 385;
 
     /**
+     * The provided max custom fee list contains fees with
+     * duplicate account id.
+     */
+    DUPLICATE_ACCOUNT_ID_IN_MAX_CUSTOM_FEE_LIST = 386;
+
+    /**
      * Max custom fees list is not supported for this operation.
      */
-    MAX_CUSTOM_FEES_IS_NOT_SUPPORTED = 386;
+    MAX_CUSTOM_FEES_IS_NOT_SUPPORTED = 387;
 }
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/TransactionChecker.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/TransactionChecker.java
@@ -444,10 +444,13 @@ public class TransactionChecker {
 
         // check required fields
         for (var maxCustomFee : maxCustomFeeList) {
-            if (maxCustomFee.accountId() == null
-                    || maxCustomFee.amountLimit() == null
-                    || maxCustomFee.amountLimit().amount() < 0) {
+            if (maxCustomFee.accountId() == null || maxCustomFee.fees().isEmpty()) {
                 throw new PreCheckException(ResponseCodeEnum.INVALID_MAX_CUSTOM_FEES);
+            }
+            for (final var fee : maxCustomFee.fees()) {
+                if (fee.amount() < 0) {
+                    throw new PreCheckException(ResponseCodeEnum.INVALID_MAX_CUSTOM_FEES);
+                }
             }
         }
     }
--- a/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
+++ b/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.consensus.impl.handlers;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.DUPLICATE_ACCOUNT_ID_IN_MAX_CUSTOM_FEE_LIST;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.DUPLICATE_DENOMINATION_IN_MAX_CUSTOM_FEE_LIST;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CHUNK_NUMBER;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CHUNK_TRANSACTION_ID;
@@ -52,6 +53,7 @@ import com.hedera.hapi.node.state.consensus.Topic;
 import com.hedera.hapi.node.transaction.AssessedCustomFee;
 import com.hedera.hapi.node.transaction.CustomFeeLimit;
 import com.hedera.hapi.node.transaction.FixedCustomFee;
+import com.hedera.hapi.node.transaction.FixedFee;
 import com.hedera.hapi.node.transaction.TransactionBody;
 import com.hedera.node.app.hapi.utils.CommonPbjConverters;
 import com.hedera.node.app.service.consensus.ReadableTopicStore;
@@ -386,36 +388,39 @@ public class ConsensusSubmitMessageHandler implements TransactionHandler {
      * Validate that each topic custom fee has equal or lower value than the payer's limit
      *
      * @param topicCustomFees The topic's custom fee list
-     * @param payerCustomFeeLimits List with limits of fees that the payer is willing to pay
+     * @param allCustomFeeLimits List with limits of fees that the payer is willing to pay
      */
     private void validateFeeLimits(
             @NonNull final AccountID payer,
             @NonNull final List<FixedCustomFee> topicCustomFees,
-            @NonNull final List<CustomFeeLimit> payerCustomFeeLimits) {
+            @NonNull final List<CustomFeeLimit> allCustomFeeLimits) {
         // Extract the token fees and hbar fees from the topic custom fees
         Map<TokenID, Long> tokenFees = new HashMap<>();
         AtomicReference<Long> hbarFee = new AtomicReference<>(0L);
         totalAmountToBeCharged(topicCustomFees, hbarFee, tokenFees);
+        final var payerLimits = allCustomFeeLimits.stream()
+                .filter(maxCustomFee -> payer.equals(maxCustomFee.accountId()))
+                .map(CustomFeeLimit::fees)
+                .flatMap(List::stream)
+                .toList();
+
         // Validate payer token limits
         tokenFees.forEach((token, feeAmount) -> {
-            final boolean isValid = payerCustomFeeLimits.stream()
-                    .filter(maxCustomFee ->
-                            token.equals(maxCustomFee.amountLimit().denominatingTokenId()))
-                    .filter(maxCustomFee -> payer.equals(maxCustomFee.accountId()))
+            final boolean isValid = payerLimits.stream()
+                    .filter(maxCustomFee -> token.equals(maxCustomFee.denominatingTokenId()))
                     .anyMatch(maxCustomFee -> {
-                        validateTrue(maxCustomFee.amountLimit().amount() >= feeAmount, MAX_CUSTOM_FEE_LIMIT_EXCEEDED);
+                        validateTrue(maxCustomFee.amount() >= feeAmount, MAX_CUSTOM_FEE_LIMIT_EXCEEDED);
                         return true;
                     });
             validateTrue(isValid, NO_VALID_MAX_CUSTOM_FEE);
         });
         // Validate payer HBAR limit
         if (hbarFee.get() > 0) {
-            final var payerHbarLimit = payerCustomFeeLimits.stream()
-                    .filter(maxCustomFee -> !maxCustomFee.amountLimit().hasDenominatingTokenId())
-                    .filter(maxCustomFee -> payer.equals(maxCustomFee.accountId()))
+            final var payerHbarLimit = payerLimits.stream()
+                    .filter(maxCustomFee -> !maxCustomFee.hasDenominatingTokenId())
                     .findFirst()
                     .orElseThrow(() -> new HandleException(NO_VALID_MAX_CUSTOM_FEE));
-            validateTrue(payerHbarLimit.amountLimit().amount() >= hbarFee.get(), MAX_CUSTOM_FEE_LIMIT_EXCEEDED);
+            validateTrue(payerHbarLimit.amount() >= hbarFee.get(), MAX_CUSTOM_FEE_LIMIT_EXCEEDED);
         }
     }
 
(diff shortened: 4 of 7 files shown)


hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusGetTopicInfoHandler.java
```diff
@@ -165,6 +165,9 @@ private Optional<ConsensusTopicInfo> infoForTopic(
             if (!isEmpty(meta.submitKey())) info.submitKey(meta.submitKey());
             info.autoRenewPeriod(Duration.newBuilder().seconds(meta.autoRenewPeriod()));
             if (meta.hasAutoRenewAccountId()) info.autoRenewAccount(meta.autoRenewAccountId());
+            if (meta.hasFeeScheduleKey()) info.feeScheduleKey(meta.feeScheduleKey());
```
Neeharika-Sompalli, line 168: Should we also check if the key is not empty here? Similar to other key checks

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -72,16 +86,30 @@
...
+        requireNonNull(customFeesValidator);
+        this.customFeesValidator = customFeesValidator;
+    }
 
     @Override
     public void pureChecks(@NonNull final TransactionBody txn) throws PreCheckException {
         final ConsensusUpdateTopicTransactionBody op = txn.consensusUpdateTopicOrThrow();
         validateTruePreCheck(op.hasTopicID(), INVALID_TOPIC_ID);
+
+        if (op.feeExemptKeyList() != null) {
```
Neeharika-Sompalli, line 106: Could be `op.hasFeeExemptKeyList`

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -117,6 +145,18 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
                 context.requireKeyOrThrow(autoRenewAccountID, INVALID_AUTORENEW_ACCOUNT);
             }
         }
+
+        // If the fee schedule key is changed then the transaction must also be signed by the new fee schedule key
+        if (op.hasFeeScheduleKey()
+                && !KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
+            context.requireKeyOrThrow(op.feeScheduleKey(), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
```
Neeharika-Sompalli, line 152: Is this a requirement ? Usually we let adminKey sign for any key changes. I believe the same for `submitKey` changes here.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -205,47 +241,51 @@ private void resolveMutableBuilderAttributes(
...
         builder.expirationSecond(resolvedExpiryMeta.expiry());
         builder.autoRenewPeriod(resolvedExpiryMeta.autoRenewPeriod());
         if (op.hasAutoRenewAccount() && designatesAccountRemoval(op.autoRenewAccount())) {
             builder.autoRenewAccountId((AccountID) null);
         } else {
             builder.autoRenewAccountId(resolvedExpiryMeta.autoRenewAccountId());
         }
+        if (op.hasFeeScheduleKey()) {
+            final var newFeeScheduleKey = op.feeScheduleKey();
+            if (newFeeScheduleKey == null || isKeyRemoval(newFeeScheduleKey)) {
```
Neeharika-Sompalli, line 258: This can't be null here

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -305,6 +345,55 @@ private void validateMaybeNewSubmitKey(
         }
     }
 
+    private void validateMaybeNewFeeScheduleKey(
+            @NonNull final AttributeValidator attributeValidator,
+            @NonNull final ConsensusUpdateTopicTransactionBody op,
+            @NonNull final Topic topic) {
+        if (op.hasFeeScheduleKey()) {
+            validateTrue(topic.hasFeeScheduleKey(), FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED);
+            if (!KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
```
Neeharika-Sompalli, line 354: If this is `UNSET` is it not error ?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/validators/ConsensusCustomFeesValidator.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private void validateExplicitTokenDenomination(
+            @NonNull final AccountID feeCollectorNum,
+            @NonNull final TokenID tokenNum,
+            final long feeAmount,
+            @NonNull final ReadableTokenRelationStore tokenRelationStore,
+            @NonNull final ReadableTokenStore tokenStore) {
+        final var denomToken = getIfUsable(tokenNum, tokenStore, REQUIRE_NOT_PAUSED, INVALID_TOKEN_ID_IN_CUSTOM_FEES);
+        validateFalse(
+                TokenSupplyType.FINITE.equals(denomToken.supplyType()) && feeAmount > denomToken.maxSupply(),
```
Neeharika-Sompalli, line 124: It will easier to read 
```
if(denomToken.supplyType().equals(TokenSupplyType.FINITE)){
            validateTrue(feeAmount > denomToken.maxSupply(), AMOUNT_EXCEEDS_TOKEN_MAX_SUPPLY);
        }
```

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
ibankov: This is what the HIP says  - The Fee Schedule Key can be updated according to the same rules that currently apply to the Submit key. In addition, to update the Fee Schedule Key, the new key must sign the transaction.

But I found a conversation in slack where Michael H says - "Only the admin key has to sign if the custom fee key is updated. This also means the new key can be invalid. This is how it currently works for the submit key, and also HIP-540 explicitly states that:
The current policy is to require a new key to sign an update transaction unless it is being changed by a higher-privilege key."

@netopyr thoughts? If we require only the adminKey as it is higher-privilege key I guess we should change the HIP.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
ibankov: Done. Even though Intelij doesn't like it and throws warning that NPE may occur.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusGetTopicInfoHandler.java
ibankov: Done

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
ibankov: Done. Even though Intelij doesn't like it and throws warning that NPE may occur.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
Neeharika-Sompalli: > This also means the new key can be invalid. 

We need to validate the new key if it is valid. But the signing requirement should be only `adminKey` should sign. Similar to other keys, unless the requirement is for new key to sign.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
ibankov: Aren't we supposed to remove the fee_schedule_key, also here we are checking the new key validity.

Neeharika-Sompalli: The key will be removed only if the Key is `IMMUTABILITY_SENTINEL_KEY`, which has a default keyList. The same is checked in `isKeyRemoval`.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
netopyr: It looks like the HIP must be updated. We want the same behavior as for the supply key, i.e. only the `adminKey` has to sign

ibankov: Removed the feeSchedule validation here.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
ibankov: I see, I guess we don't need this if statement at all then because the attributeValidator.validateKey that we are calling inside checks if the key is UNSET and if true throws BAD_ENCODING exception.

andrewb1269hg (dismissed):
Update looks good.

Neeharika-Sompalli (approved):
LGTM!

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
```diff
@@ -75,7 +77,15 @@ message DeleteTopicOutput {}
...
  */
-message SubmitMessageOutput {}
+message SubmitMessageOutput {
+
+    /**
+     * Custom fees assessed during a SubmitMessage.
+     * <p>
+     * These fees SHALL be present in the full transfer list for the transaction.
+     */
+    repeated proto.AssessedCustomFee assessed_custom_fees = 1;
```
steven-sheehy, line 87: This change isn't detailed in the HIP since it doesn't mention block streams. There is an incongruity here.

In record streams, the `max_custom_fees` is in the generic `TransactionBody` and the `assessed_custom_fees` is in the generic `TransactionRecord` due to wanting to expand max custom fees to more types in the future.

But for block streams, the `max_custom_fees` is again in the generic `TransactionBody` but the assessed custom fees is in the specific submit message output. I think since we made the decision to use `TransactionBody` for `max_custom_fees` we should stick with that decision to keep it generic and reusable and move the assessed custom fees to the `TransactionResult`.

hapi/hedera-protobufs/services/custom_fees.proto
```diff
@@ -336,3 +336,87 @@ message AssessedCustomFee {
...
+message CustomFeeLimit {
+    /**
+     * A payer account identifier.
+     */
+    AccountID account_id = 1;
+
+    /**
+     * A custom fee amount limit.
+     */
+    FixedFee amount_limit = 2;
```
steven-sheehy, line 421: This differs from the HIP: `repeated FixedFee fees = 2;`

hapi/hedera-protobufs/services/consensus_create_topic.proto
```diff
@@ -135,4 +136,49 @@ message ConsensusCreateTopicTransactionBody {
...
+
+    /**
+     * A set of keys.<br/>
+     * Keys in this list are permitted to submit messages to this topic without
+     * paying custom fees associated with this topic.
+     * <p>
+     * If a submit transaction is signed by _any_ key included in this set,
+     * custom fees SHALL NOT be charged for that transaction.<br/>
+     * This field MUST NOT contain more than
+     * `MAX_ENTRIES_FOR_FEE_EXEMPT_KEY_LIST` keys.<br/>
```
steven-sheehy, line 160: This variable is internal to consensus nodes and not defined in the protobuf. While viewers of this PR can see its value, later viewers of the protobuf documentation or generated code will not know its value. We should probably replace its value with the hardcoded value here or define it somewhere in the docs globally.

hapi/hedera-protobufs/services/transaction.proto
```diff
@@ -609,4 +610,13 @@ message TransactionBody {
...
     }
+
+    /**
+     * A list of maximum custom fees that the users are willing to pay.
+     * <p>
+     * This field is OPTIONAL.<br/>
+     * If left empty, the users are accepting to pay any custom fee.<br/>
+     * If used with a transaction type that does not support custom fee limits, the transaction will fail.
+     */
+    repeated CustomFeeLimit maxCustomFees = 1001;
```
steven-sheehy, line 621:
```suggestion
    repeated CustomFeeLimit max_custom_fees = 1001;
```

steven-sheehy (changes requested):
Reviewed proto only.

lukelee-sl (approved):
LGTM.  For smart contracts updates which seem very minimal

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
Neeharika-Sompalli: It is mostly a design decision. Usually `TransactionResult` has fields that are set by most of the transactions. It could be that all transactions will set this field in the future. 

If we decide to make that change, probably we need to remove that field from `CryptoTransferOutput` and `TokenAirdropOutput` also and add to the `TransactionResult`.

hapi/hedera-protobufs/services/consensus_create_topic.proto
kimbor: We have many, many examples in the protobuf where we mention configuration properties in the comments. See, for example, `ConsensusCreateTopicTransactionBody#autoRenewAccount` which has comments which contain `The configured MAX_AUTORENEW_PERIOD at the time of automatic renewal.` I don't think that we want to copy all these configuration values to the protobuf documentation.

hapi/hedera-protobufs/services/transaction.proto
kimbor: Fixed.

hapi/hedera-protobufs/services/consensus_create_topic.proto
kimbor: I guess there are also examples where we do list the default value in the comments. See message TransactionBody
```
    /**
     * A short description for this transaction.
     * <p>
     * This value, if set, MUST NOT exceed `transaction.maxMemoUtf8Bytes`
     * (default 100) bytes when encoded as UTF-8.
     */
    string memo = 6;

```

andrewb1269hg (approved):
Reapprove `hedera-node/hedera-consensus-service-impl/build.gradle.kts`

hapi/hedera-protobufs/services/consensus_create_topic.proto
vtronkov: Fixed 👍 Added the constant as a comment.

hapi/hedera-protobufs/services/custom_fees.proto
vtronkov: Fixed 👍

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
netopyr: I agree with @steven-sheehy. As `max_custom_fees` is in `TransactionBody`, it makes sense to put `assessed_custom_fee` in `TransactionResult`.

As block streams are still in preview in 0.59, I think this can be fixed in a separate PR in 0.60.

steven-sheehy (approved):

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
steven-sheehy: Ok, let's make sure a ticket is created to track and link it here

jsync-swirlds: The Block Streams HIP is headed the opposite direction, only adding to TransactionResult if the value applies to _all_ transactions, and adding to specific transactions whenever possible.

andrewb1269hg (approved):
Reapprove `hedera-node/hedera-consensus-service-impl/build.gradle.kts`

PR #15595, opened 2024-09-25
fix: DefaultKycStatusCall correct returned values from System Contract

### LINKED ISSUE #5775: Wrong status from getTokenDefaultKycStatus precompile
### Description

`getTokenDefaultKycStatus` is returning `false `when a token has a KYC key.
`getTokenDefaultKycStatus` is returning `false `when a token does not have a KYC key.

~Expected: getTokenDefaultKycStatus returns `true` when a token has a KYC key.~

Edit:  Expected: getTokenDefaultKycStatus returns `true` when the default KYC status is KycNotApplicable

### Steps to reproduce

~1. Create a token with KYC key~
1. Create a token without KYC key
2. Call getTokenDefaultKycStatus for the created token
3. Response is false

### Additional context

_No response_

### Hedera network

other

### Version

v.0.35
Still present in v0.55

### Operating system

None

stoyanov-st:
**Description**:
This PR aims to fix the way we return the value for DefaultKycStatusCall system contract..
Now returning the value based on kycKey present. If the related token is missing the kycKey the status is `KycNotApplicable` which is represented as a boolean true. We return false when the token has kycKey where the status is `Revoked`.
Adjusted existing tests and added a new one.

**Related issue(s)**:

Fixes #5775 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/defaultkycstatus/DefaultKycStatusCall.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/defaultkycstatus/DefaultKycStatusCall.java
@@ -50,10 +50,8 @@ public class DefaultKycStatusCall extends AbstractNonRevertibleTokenViewCall {
     @Override
     protected @NonNull PricedResult resultOfViewingToken(@Nullable final Token token) {
         requireNonNull(token);
-        return gasOnly(
-                fullResultsFor(SUCCESS, gasCalculator.viewGasRequirement(), token.accountsKycGrantedByDefault()),
-                SUCCESS,
-                true);
+        final boolean kycStatus = !token.hasKycKey() || token.accountsKycGrantedByDefault();
+        return gasOnly(fullResultsFor(SUCCESS, gasCalculator.viewGasRequirement(), kycStatus), SUCCESS, true);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


Nana-EC (dismissed):

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/defaultkycstatus/DefaultKycStatusCall.java
```diff
@@ -50,10 +50,8 @@ public DefaultKycStatusCall(
     @Override
     protected @NonNull PricedResult resultOfViewingToken(@Nullable final Token token) {
         requireNonNull(token);
-        return gasOnly(
-                fullResultsFor(SUCCESS, gasCalculator.viewGasRequirement(), token.accountsKycGrantedByDefault()),
-                SUCCESS,
-                true);
+        final boolean kycStatus = !token.hasKycKey() || token.accountsKycGrantedByDefault();
```
lukelee-sl, line 53: Perhaps I misunderstood but where is this logic?
> If the related token is missing the kycKey the status is KycNotApplicable which is represented as a boolean true. We return false when the token has kycKey where the status is Revoked

stoyanov-st: The return value for this function is based on the presence of the kycKey. Having the key means the default kyc status is Revoked (false)  and missing it means that kyc is not applicable (true) thus the negative `!token.hasKycKey()`.
The expression takes into consideration the value of the `defaultKycStatus` (that is always false) to ensure that if at any moment we decide to change the services behaviour for this field this function will work as expected.

lukelee-sl (approved):
LGTM

PR #15603, opened 2024-09-30
feat: Update ConsensusSubmitMessageHandler with new custom fee functionality

### LINKED ISSUE #15516: Update ConsensusSubmitMessageHandler with new custom fee functionality
Update ConsensusSubmitMessageHandler with new custom fee functionality:

- Adds `max_custom_fees` and `accept_all_custom_fees` fields to `ConsensusSubmitMessageTransactionBody` and validate topic fees against those limits.
- Build and dispatch `CryptoTransfer` child transactions for fee payments.
- Adds some Hapi tests:
TOPIC_FEE_104, TOPIC_FEE_105, TOPIC_FEE_106, TOPIC_FEE_108, TOPIC_FEE_109, TOPIC_FEE_110, TOPIC_FEE_111, TOPIC_FEE_112, TOPIC_FEE_113, TOPIC_FEE_116, TOPIC_FEE_124, TOPIC_FEE_125, TOPIC_FEE_126
[Link to the test plan](https://limechain.notion.site/HIP-991-Test-Plan-a6ecb34ddbcf49cdaf6a3332e25709f1?pvs=4)

### LINKED ISSUE #16854: Add max_custom_fees and accept_all_custom_fees field to ConsensusSubmitMessageTransactionBody and update the handler to delete allowances
* Delete the existing allowance code from `ConsensusSubmitMessageHandler`
* Add the new `max_custom_fees` and `accept_all_custom_fees` fields to `ConsensusSubmitMessageTransactionBody`
* Add the additional logic in `ConsensusSubmitMessageHandler` to support the newly added fields

JivkoKelchev:
**Description**:

Adds topic custom fees payment in submit message transaction handling.

**Related issue(s)**:

Fixes #15516
Fixes #16854

**Notes for reviewer**:

`ConsensusCustomFeeAssessor` is building a list of synthetic crypto transfer bodies **per** topic custom fee instead of one consolidated transaction body. This is done to avoid exceeding balance changes limits when dispatching child crypto transfers.
There is a [PR](https://github.com/hashgraph/hedera-services/pull/17072) that suppresses all child records and populates all assessed fees in the parent transaction record.

**Checklist**


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/transactions/consensus/HapiTopicCreate.java
```diff
@@ -66,7 +66,7 @@ public class HapiTopicCreate extends HapiTxnOp<HapiTopicCreate> {
     private Optional<KeyShape> feeScheduleKeyShape = Optional.empty();
     private final List<Function<HapiSpec, ConsensusCustomFee>> feeScheduleSuppliers = new ArrayList<>();
     private Optional<List<Function<HapiSpec, Key>>> feeExemptKeyNamesList = Optional.empty();
-    private Optional<List<Key>> feeExemptKeyList = Optional.empty();
+    private Optional<List<Key>> freeMesssageKeyList = Optional.empty();
```
iwsimon, line 69: Not sure why freeMesssageKeyList changed from feeExemptKeyList? feeExemptKeyList is used in HIP-991.

JivkoKelchev: You are right. Fixed.
Initially it was `freeMessageKeyList`, and then it was changed to `feeExemptKeyList`. I accidentally reverted it to initial state, while reverting all allowance related features.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
```diff
@@ -128,6 +149,26 @@ public void handle(@NonNull final HandleContext handleContext) {
...
+        /* handle custom fees */
+        if (!topic.customFees().isEmpty() && !isFeeExempted(topic.feeExemptKeyList(), handleContext.keyVerifier())) {
+            // check payer limits or throw
+            if (!op.acceptAllCustomFees()) {
+                validateFeeLimits(topic.customFees(), op.maxCustomFees());
+            }
+            // create synthetic body and dispatch crypto transfer
+            final var syntheticBodies = customFeeAssessor.assessCustomFee(topic, handleContext);
+            for (final var syntheticBody : syntheticBodies) {
+                final var record = handleContext.dispatch(stepDispatch(
```
iwsimon, line 161: As discussed at today's call, we need to add suppressing a separate record for a child dispatch.

ibankov: This is still under discussion and @JivkoKelchev created a seperate [PR](https://github.com/hashgraph/hedera-services/pull/17072) for that.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
```diff
@@ -280,6 +329,103 @@ public static byte[] noThrowSha384HashOf(final byte[] byteArray) {
...
+    private void validateFeeLimits(
+            @NonNull final List<ConsensusCustomFee> topicCustomFees,
+            @NonNull final List<FixedFee> payerCustomFeeLimits,
+            @NonNull final HandleContext context) {
+        // Validate the duplication of payer custom fee limits
+        validateDuplicationFeeLimits(payerCustomFeeLimits);
+
+        // Extract the token fees and hbar fees from the topic custom fees
+        Map<TokenID, Long> tokenFees = new HashMap<>();
+        AtomicReference<Long> hbarFee = new AtomicReference<>(0L);
```
Neeharika-Sompalli, line 366: We could re-use these maps to construct the synthetic crypto transfers right ?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/customfee/ConsensusCustomFeeAssessor.java
```diff
@@ -0,0 +1,134 @@
...
+     * Build and return a list of synthetic crypto transfer transaction bodies, that represents custom fees payments.
+     * It will return one body per topic custom fee.
+     *
+     * @param topic The topic
+     * @param context The transaction handle context
+     * @return List of synthetic crypto transfer transaction bodies
+     */
+    public List<CryptoTransferTransactionBody> assessCustomFee(
+            @NonNull final Topic topic, @NonNull final HandleContext context) {
+        final List<CryptoTransferTransactionBody> transactionBodies = new ArrayList<>();
```
Neeharika-Sompalli, line 61: I think it will be easier to pass in the maps constructed to create the synthetic bodies

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
JivkoKelchev: It is not possible to reuse these maps to build transfer bodies. There is no information about the collectors here, and the amount is the sum of all fees that have the same denomination.
Since we can have two or more custom fees with the same denomination, but with different collectors, we need to know the total amount per token and the total hbar amount. We use these maps to validate if the limits (`max_custom_fee`) that the payer is willing to pay are greater than or equal to the sum we calculate here.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/customfee/ConsensusCustomFeeAssessor.java
JivkoKelchev: I'm not sure if I understand you correctly, but the maps from the previous comment don't contain the necessary information in order to build the crypto transfer bodies. 

Anyway, I did some refactoring. 
Now, before validation of the `max_custom_fees` and synthetic body creations, we filter the list of custom fees and use only chargeable fees. (we will skip the fees where the payer is collector or token treasury)
This way we don't need to check twice, if a given fee should be validated or to create a body with it.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
```diff
@@ -105,6 +128,12 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         if (topic.hasSubmitKey()) {
             context.requireKeyOrThrow(topic.submitKeyOrThrow(), INVALID_SUBMIT_KEY);
         }
+        // add optional fee exempt keys in to the key verifieer
+        // later it will be used to validate if transaction was signed by
+        // any of these keys and based on that, custom fees will be charged or not
+        if (!topic.feeExemptKeyList().isEmpty()) {
+            context.optionalKeys(new HashSet<>(topic.feeExemptKeyList()));
```
Neeharika-Sompalli, line 135: Was discussing with @tinker-michaelj  regarding this. This seems to be a lot of keys to verify, since we don't have limit on number of simple keys in the 10 keys allowed. Either we should add a limit on number of simple keys in `ConsensusCreateTopicHandler` or do this in a different way. @netopyr  thoughts?

netopyr: As discussed, we should do this check as @tinker-michaelj proposed.

JivkoKelchev: Fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip991/TopicCustomFeeBase.java
```diff
@@ -69,6 +90,82 @@ protected static SpecOperation[] setupBaseForUpdate() {
...
+            final var tokenName = TOKEN_PREFIX + i;
+            specOperations.addAll(createTokenWith2LayerFee(owner, tokenName, false));
+        }
+        return specOperations.toArray(new SpecOperation[0]);
+    }
+
+    /**
+     *
+     *
+     *
```
vtronkov, line 130: It looks like you forgot to finish the documentation here.

vtronkov (approved):

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
```diff
@@ -280,6 +330,153 @@ public static byte[] noThrowSha384HashOf(final byte[] byteArray) {
...
+     *
+     * @param topicCustomFees The topic's custom fee list
+     * @param payerCustomFeeLimits List with limits of fees that the payer is willing to pay
+     */
+    private void validateFeeLimits(
+            @NonNull final AccountID payer,
+            @NonNull final List<FixedCustomFee> topicCustomFees,
+            @NonNull final List<CustomFeeLimit> payerCustomFeeLimits) {
+        // Validate the duplication of payer custom fee limits
+        validateDuplicationFeeLimits(payerCustomFeeLimits);
```
Neeharika-Sompalli, line 383: Can we do this check in `pureChecks` ?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
```diff
@@ -128,6 +154,30 @@ public void handle(@NonNull final HandleContext handleContext) {
...
+            final var syntheticBodies = customFeeAssessor.assessCustomFee(feesToBeCharged, handleContext.payer());
+            for (final var syntheticBody : syntheticBodies) {
+                final var record = handleContext.dispatch(stepDispatch(
+                        handleContext.payer(),
+                        TransactionBody.newBuilder()
+                                .cryptoTransfer(syntheticBody)
+                                .build(),
+                        CryptoTransferStreamBuilder.class,
+                        NOOP_TRANSACTION_CUSTOMIZER));
+                validateTrue(record.status().equals(SUCCESS), record.status());
```
Neeharika-Sompalli, line 177: Didn't we decide that we need to set all assessed custom fees in the same recordBuilder? I don't see that happening here. Am I missing something?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/customfee/ConsensusCustomFeeAssessor.java
```diff
@@ -0,0 +1,118 @@
...
+                        AccountAmount.newBuilder()
+                                .accountID(collector)
+                                .amount(fee.amount())
+                                .build())
+                .build();
+    }
+
+    @VisibleForTesting
+    public AccountID getTokenTreasury(TokenID tokenId, ReadableTokenStore tokenStore) {
+        final var token = tokenStore.get(tokenId);
```
Neeharika-Sompalli, line 114: we could use `getIfUsable(tokenId)` here

Neeharika-Sompalli (approved):
Overall LGTM! Thanks @JivkoKelchev. Had two comments

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
JivkoKelchev: Moved.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/customfee/ConsensusCustomFeeAssessor.java
JivkoKelchev: Fixed.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusSubmitMessageHandler.java
JivkoKelchev: Yes, I implemented it when there wasn't a final decision. So it is in another [PR](https://github.com/hashgraph/hedera-services/pull/17072)

PR #15618, opened 2024-09-30
test: Use DAB upgrade test in CI runs 

Neeharika-Sompalli:
Use upgrade test that uses DAB transactions for deleting and updating node instead of the old way for upgrade


iwsimon (approved):
LGTM, thanks @Neeharika-Sompalli

JeffreyDallas (approved):

rbarkerSL (approved):

PR #15622, opened 2024-09-30
feat: Added states for TssService

### LINKED ISSUE #14745: Create the TSS related State using the protobuf definitions and unit test the API.

### LINKED ISSUE #14754: Create feature flag for TSS-Ledger-ID capability, default to off

### LINKED ISSUE #14966: Create `tss.maxSharesPerNode` and `tss.keyActiveRoster` config parameters

Neeharika-Sompalli:
Fixes #14745 
Fixes #14966 
Fixes #14754 

- Added a map state that contain `TssMessageMapKey` to `TssMessageTransactionBody`
- Added a map state that contain `TssVoteMapKey` to `TssVoteTransactionBody`
- Added the configurations `tss.maxSharesPerNode` and `tss.keyActiveRoster` as suggested in the design document


hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TssConfig.java
```diff
@@ -0,0 +1,32 @@
...
+package com.hedera.node.config.data;
+
+import com.hedera.node.config.NetworkProperty;
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+
+/**
+ * Configuration for the TSS service.
+ * @param maxSharesPerNode The maximum number of shares that can be assigned to a node.
+ * @param keyActiveRoster A test-only configuration set to true to key the active roster with TSS key material.
```
mhess-swl, line 26: This is my understanding:
```suggestion
 * @param keyActiveRoster A test-only configuration; set this to true to enable the process that will key the candidate roster with TSS key material.
```
If a roster is active, doesn't it already has the required key material?

thomas-swirlds-labs (dismissed):
LGTM. Thanks @Neeharika-Sompalli 👍

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TssConfig.java
Neeharika-Sompalli: @mhess-swl I think this configuration is to not wait until upgrade boundary like in production envs. As soon as we get candidate roster we will key it just for tests.

MiroslavGatsanoga (dismissed):
LGTM.

petreze (dismissed):

hedera-node/hedera-config/src/main/java/com/hedera/node/config/data/TssConfig.java
mhess-swl: That would make sense. If that's the case, would `keyCandidateRoster` be a better name? Not necessary for this PR, but may be helpful to clarify

tinker-michaelj (approved):

PR #15632, opened 2024-10-01
test: Create Test-Clients setup and first test for Node-Operator Queries

Evdokia-Georgieva:
**NB**
This PR needs to be merged to develop after the PR for 11211-permit-unpaid-localhost-queries branch is merged.

**Description**:

This PR implements changes to Test-Clients so that testing of Node-Operator queries can be possible.

The changes are related to the implementation of a new Node Operator Port, as follows:
* In HAPI tests call .asNodeOperator() void method.
* The .asNodeOperator() void method sets a new field in HapiQueryOp.java which by default is false but when the method is called it will be set to true.
* Once the field in HapiQueryOp.java is set to true, in .submitOp() method we call the .send() method for queries and we need to pass the new parameter isNodeOperator to this .send() method.
* After we pass it, in EmbededNetwork.java we need to setup that this parameter will not be used
* In AbstractGrpcNetwork - we pass the new parameter to GrpcUtils.java
* In HapiClients.java - we will pass the node parameter to the many service stub methods and we need to introduce new field which will contain the NodeOperator stub id-s, the .send() method will have to get the correct id when happy clients is initialised - these id-s are for the nodes created for the free node-operator queries.
* In HapiCLients.java we need to create new method which is similar to ensureChannelStubsInPool but for node operator.

This PR also includes one test for node-operator queries in order to implement the test-cleints changes.

Test Plan: https://limechain.notion.site/Allow-Free-Queries-for-Node-Operators-Test-Plan-10510baa8dcd800eae72ffff6b766c7b

**Resolves:**
#15650


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        @HapiTest
+        final Stream<DynamicTest> nodeOperatorQueryVerifyPayerBalanceForAccountBalance() {
+            // declare payer account balance variables
+            final AtomicLong currentPayerBalance = new AtomicLong();
+            final AtomicLong payerBalanceAfterFirstQuery = new AtomicLong();
+            final AtomicLong payerBalanceAfterSecondQuery = new AtomicLong();
+            return hapiTest(withOpContext((spec, log) -> {
+                final var initialBalance = getAccountBalance(PAYER).exposingBalanceTo(currentPayerBalance::set);
+                // perform paid query, pay for the query with payer account
```
mhess-swl, line 70: Using `withOpContext(...)` for all test operations makes these tests much more difficult to read. Why is the op context necessary? 

Can we do something like this instead?

```
return hapiTest(
    getAccountBalance(NODE_OPERATOR).payingWith(PAYER).signedBy(PAYER),
    sleepFor(3_000),
    getAccountBalance(PAYER).exposingBalanceTo(payerBalanceAfterFirstQuery::set)
    ...
    doAdhoc(() -> {
        Assertions.assertEquals(
                        currentPayerBalance.get(), 
                        payerBalanceAfterFirstQuery.get(), 
                        "Balances are not equal!");
        Assertions.assertEquals(
                        payerBalanceAfterFirstQuery.get(),
                        payerBalanceAfterSecondQuery.get(),
                        "Balances are not equal!");
    })
); 
```

Another thought, aren't the account balances known at the start of the test, since they are set up with `createAllAccountsAndTokens`? If the account balances are known, we could specify an expected amount instead of tracking balances with awkward atomic variables:

```
return hapiTest(
    getAccountBalance(NODE_OPERATOR).payingWith(PAYER).signedBy(PAYER)
          .hasTinyBars(<expected amount>),
    sleepFor(3_000),
    getAccountBalance(NODE_OPERATOR).payingWith(PAYER).asNodeOperator()
          .hasTinyBars(<expected amount>),
    ... 
);
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
```diff
@@ -0,0 +1,144 @@
...
+                // the grpc client performs the query to different ports
+                final var firstQuery =
+                        getAccountBalance(NODE_OPERATOR).payingWith(PAYER).signedBy(PAYER);
+                final var waitingForQueryPaymentToReachConsensus = sleepFor(3_000);
+                final var newBalance = getAccountBalance(PAYER).exposingBalanceTo(payerBalanceAfterFirstQuery::set);
+                // perform free query to local port with asNodeOperator() method
+                final var secondQuery = getAccountBalance(NODE_OPERATOR)
+                        .payingWith(PAYER)
+                        .signedBy(PAYER)
```
mhess-swl, line 79: Each `.signedBy(PAYER)` shouldn't be necessary if I remember right. Essentially, I thought the framework could figure out the need to sign with `PAYER` if `.payingWith(PAYER)` is used.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/NodeOperatorQueriesBase.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return array of operations
+     */
+    protected static SpecOperation[] createAllAccountsAndTokens() {
+
+        final var createOperations = new ArrayList<SpecOperation>(List.of(
+                cryptoCreate(NODE_OPERATOR).balance(ONE_HUNDRED_HBARS),
+                cryptoCreate(OWNER).balance(0L),
+                cryptoCreate(QUERY_TEST_ACCOUNT).balance(ONE_HUNDRED_HBARS),
+                cryptoCreate(PAYER).balance(ONE_HUNDRED_HBARS),
+                //                        .exposingCreatedIdTo(payerId::set),
```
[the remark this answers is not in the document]
Evdokia-Georgieva: Absolutely unintentional, I must have missed it when deleting while testing, sorry, fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
Evdokia-Georgieva: You are right, I changed the tests without using withOpContext(...) for all operations.

Regarding the atomic variables, I think it is better to not track concrete amounts of the Payer balances as it will be really difficult to track that and currently we can reuse the Payer account for many tests and still assert only whether it was or was not charged which is enough for these tests. It will be really difficult to track the concrete amounts in these tests and reuse the account and I suggest we leave the asserts this way.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/grpc/impl/netty/GrpcServiceBuilderTest.java
```diff
@@ -104,7 +104,7 @@ void queryIsBlank(final String value) {
     @Test
     @DisplayName("The build method will return a ServiceDescriptor")
     void serviceDescriptorIsNotNullOnNoopBuilder() {
-        assertNotNull(builder.build(metrics));
+        assertNotNull(builder.build(metrics), "Object is Null!");
```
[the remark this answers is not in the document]
netopyr: We agreed to use `AssertJ` for assertions. While this is not consistently enforced because we never cleaned up existing code, new code should not use other libraries.

Feel free to clean up the other assertions in this class.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/AbstractGrpcNetwork.java
```diff
@@ -58,12 +58,13 @@ protected AbstractGrpcNetwork(
     public @NonNull Response send(
             @NonNull final Query query,
             @NonNull final HederaFunctionality functionality,
-            @NonNull final com.hederahashgraph.api.proto.java.AccountID nodeAccountId) {
+            @NonNull final com.hederahashgraph.api.proto.java.AccountID nodeAccountId,
+            boolean asNodeOperator) {
```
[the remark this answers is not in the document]
netopyr: @Evdokia-Georgieva: has anybody pointed you to our style guide?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
Evdokia-Georgieva: Yes, you are absolutely right, I checked that and will remove all .signedBy

hedera-node/hedera-app/src/test/java/com/hedera/node/app/grpc/impl/netty/GrpcServiceBuilderTest.java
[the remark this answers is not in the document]
Evdokia-Georgieva: Thank you for this comment, I was not aware that we have agreed on AssertJ, I changed it for the new tests.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/AbstractGrpcNetwork.java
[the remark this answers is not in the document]
Evdokia-Georgieva: Fixed.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
mhess-swl: > currently we can reuse the Payer account for many tests

Reusing accounts across tests without resetting their state very likely means 1️⃣ we won't be able to make these new tests run in parallel, 2️⃣ a higher chance of test flakes, and 3️⃣ that these tests will be harder to debug, all because the tests are not independent of each other :( As is the case with some existing hapi tests in our repo, it's possible that a single test failure in these interrelated tests will require running the whole suite, or even multiple suites, to troubleshoot. We've spent a lot of time hunting these sorts of problems in CI. I would definitely prefer that entities (accounts etc.) be re-created in each test to keep the tests unrelated, even if it makes each test slower. 

That said, keeping the atomic vars is an option if other advantages outweigh the cons. I'm sure I haven't thought of everything 🤪 More thoughts?

netopyr: I agree. Tests must be independent. Only immutable objects can be reused.

Evdokia-Georgieva: I absolutely agree that the best approach is to create and clean up the test data for each individual test but I have some concerns in this case.
First one is that by creating the test data for each separate test, in some tests we might exceed the limitation of 7 operations so that they become too many.
Apart from that, even if we only create Payer in each test, wouldn’t that flood the network with too many accounts, when we can reuse one or two accounts instead? In these tests, the only thing that is important is whether the Payer was charged or not - whether his balance at the beginning of the test was changed or not and in most cases it should not be changed as the node operator queries should be free of charge. Keep in mind that even if we create new Payer for each test, we still won’t need to strictly verify the amount of it balance, but only whether it was charged or not and we will still need to use some variables in order to verify that. I was thinking that if during the creation of the tests the Payer balance becomes insufficient, we could increase it so that it works for these specific tests but I might be wrong.
My second concern with the creation of the test data for each test is related to the data cleanup. If we are going to create separate test data for each test, then we will need to be able to properly clean it so that the state is refreshed for each test. If we do such cleanup anyway then it is ok and this one is not valid.
These were my thoughts while I created the tests, if you think that creating separate test data for each happy test case is better, then I will do them this way, just have in mind these details.

netopyr: Is it mandatory to run these tests in isolation and order? In the past, we had huge problems with test tasks that took forever to complete. One reason was that we could not run tests in parallel, which made us very sensitive to the topic.

I do not think you have to worry about the test data. These tests are executed against ad hoc networks, which are discarded after the tests have finished. For this reason, it is probably not even necessary to clean up the data afterward.

Why does it concern you that we may exceed seven operations? Is there a limit?

Evdokia-Georgieva: Regarding the limit, I was referring to this requirement from the style guide checklist: https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/README.md#style-guide:~:text=The%20test%20includes%20no%20more%20than%20seven%20(7)%20SpecOperation%27s.
Regarding the clean up it is great, I was not aware, so this one was not valid.
It was necessary to run the tests in isolation exactly to avoid race conditions because of the reused test data. If I create the test data for each separate test that won't be necessary. 
In regard to that, I have added new commit with only test accounts for payer and node operator created for each test and removed the ordered in isolation tag, please let me know if this is what we are aiming.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
```diff
@@ -0,0 +1,121 @@
...
+        static void beforeAll(@NonNull final TestLifecycle lifecycle) {
+            lifecycle.doAdhoc(createAllAccountsAndTokens());
+        }
+
+        @HapiTest
+        final Stream<DynamicTest> nodeOperatorQueryVerifyPayerBalanceForAccountBalance_WithoutOpContext() {
+            // declare payer account balance variables
+            final AtomicLong currentPayerBalance = new AtomicLong();
+            final AtomicLong payerBalanceAfterFirstQuery = new AtomicLong();
+            final AtomicLong payerBalanceAfterSecondQuery = new AtomicLong();
```
mhess-swl, line 66: I still think the slight flexibility the atomic vars approach offers isn't worth the cost, for the following reasons: 

- We can't actually see an expected amount anywhere in the test, meaning other factors (e.g. a variable exchange rate, or any other network-wide property) could influence the balances and we wouldn't necessarily know, potentially making debugging harder 
- It's awkward to store every query result (e.g. `payerBalanceAfterFirstQuery`...what if there are 18 queries?) We're essentially getting around the intended use of the framework by not using known verbs like `has(accountWith().balance(123L))`
- This is significantly different than how the vast majority of hapi tests are written 

BUT, if this is how the team wants to write tests going forward, I'll be quiet 😂 If that is the case, we should add a ticket to make these types of assertions more 'supported' by the framework.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
```diff
@@ -71,6 +75,10 @@ public class HapiClients {
      * Id of node-{host, port} pairs to use for non-workflow operations using TLS ports
      */
     private final Map<AccountID, String> tlsStubIds;
+    /**
+     * Id of node-{host, port} pairs to use for non-workflow operations using node operator ports
```
mhess-swl, line 79: Nit:

```suggestion
     * Id of node-{host, port} pairs to use for node operator operations
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
```diff
@@ -144,90 +152,252 @@ private void ensureChannelStubsInPool(@NonNull final NodeConnectInfo node, final
...
-        stubIds = nodes.stream().collect(Collectors.toMap(NodeConnectInfo::getAccount, NodeConnectInfo::uri));
-        tlsStubIds = nodes.stream().collect(Collectors.toMap(NodeConnectInfo::getAccount, NodeConnectInfo::tlsUri));
-        nodes.forEach(node -> {
+    private void ensureNodeOperatorChannelStubsInPool(@NonNull final HederaNode node) {
+        requireNonNull(node);
+        final var channelUri = uri(node.getHost(), node.getGrpcNodeOperatorPort());
+        final var existingPool = channelPools.computeIfAbsent(channelUri, COPY_ON_WRITE_LIST_SUPPLIER);
+        if (existingPool.size() < MAX_DESIRED_CHANNELS_PER_NODE) {
+            final var channel = createNettyChannel(false, node.getHost(), node.getGrpcNodeOperatorPort(), -1);
+            requireNonNull(channel, "Cannot continue without Netty channel");
```
mhess-swl, line 161:
```suggestion
            requireNonNull(channel, "FATAL: Cannot continue without additional Netty channel");
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
```diff
@@ -144,90 +152,252 @@ private void ensureChannelStubsInPool(@NonNull final NodeConnectInfo node, final
...
+            existingPool.add(ChannelStubs.from(channel, new NodeConnectInfo(node.hapiSpecInfo()), false));
+        }
+        stubSequences.putIfAbsent(channelUri, new AtomicInteger());
+    }
+
+    private HapiClients(Supplier<List<NodeConnectInfo>> nodeInfosSupplier) {
```
mhess-swl, line 167: I wish we could make spotless do this... 🤔 

```suggestion
    private HapiClients(final Supplier<List<NodeConnectInfo>> nodeInfosSupplier) {
```
Not a big deal though

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
```diff
@@ -144,90 +152,252 @@ private void ensureChannelStubsInPool(@NonNull final NodeConnectInfo node, final
...
-                .map(HederaNode::hapiSpecInfo)
-                .map(NodeConnectInfo::new)
-                .toList());
+        return new HapiClients(network.nodes());
     }
 
-    public FileServiceBlockingStub getFileSvcStub(AccountID nodeId, boolean useTls) {
-        return nextStubsFromPool(stubId(nodeId, useTls))
+    /**
+     * Retrieves a blocking stub for the FIleService based on the provided parameters.
```
mhess-swl, line 212:
```suggestion
     * Retrieves a blocking stub for the FileService based on the provided parameters.
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/utils/GrpcUtils.java
```diff
@@ -109,93 +118,115 @@ public static TransactionResponse submit(
...
-            case ContractUpdate -> clients.getScSvcStub(nodeAccountId, false).updateContract(transaction);
-            case ContractDelete -> clients.getScSvcStub(nodeAccountId, false).deleteContract(transaction);
-            case EthereumTransaction -> clients.getScSvcStub(nodeAccountId, false)
+            case ContractCall -> clients.getScSvcStub(nodeAccountId, false, false)
+                    .contractCallMethod(transaction);
+            case ContractCreate -> clients.getScSvcStub(nodeAccountId, false, false)
+                    .createContract(transaction);
+            case ContractUpdate -> clients.getScSvcStub(nodeAccountId, false, false)
+                    .updateContract(transaction);
+            case ContractDelete -> clients.getScSvcStub(nodeAccountId, false, false)
```
mhess-swl, line 127: For these and all of the following transaction `case` statements, it seems cleaner to add simple method overloads to `HapiClients`:

```java
public FileServiceBlockingStub getFileSvcStub(final AccountID nodeId, final boolean useTls) {
    return getFileSvcStub(nodeId, useTls, false);
}
```

This way we wouldn't have to add another `false` param to a bunch of transaction calls that have nothing to do with node operator queries. Just a thought.

hedera-node/hedera-app/src/test/java/com/hedera/node/app/test/grpc/GrpcTestBase.java
```diff
@@ -244,6 +244,18 @@ protected String send(final String service, final String function, final String
...
+     *
+     * This method constructs a unary call to the specified service and function,
+     * marshaling the request and response as strings. It blocks until the call is
+     * completed and returns the response payload.
+     *
+     * @param service the name of the service to call
+     * @param function the name of the function to invoke within the service
+     * @param payload the request payload to send
+     * @return the response payload received from the service
+     */
```
mhess-swl, line 258: Thank you for adding this!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
```diff
@@ -0,0 +1,121 @@
...
+                    sleepFor(3_000),
+                    getAccountBalance(PAYER).exposingBalanceTo(payerBalanceAfterFirstQuery::set),
+                    // perform free query to local port with asNodeOperator() method
+                    getAccountInfo(NODE_OPERATOR).payingWith(PAYER).asNodeOperator(),
+                    sleepFor(3_000),
+                    // assert payer account balance is not changed
+                    getAccountBalance(PAYER).exposingBalanceTo(payerBalanceAfterSecondQuery::set),
+                    doAdhoc(() -> {
+                        assertThat(payerBalanceAfterFirstQuery.get())
+                                .withFailMessage("Balances are equal!")
```
mhess-swl, line 113:
```suggestion
                                // The balances after the first query should not be equal because the first query is not a node
                                // operator query, and should therefore be charged to `payer`
                                .withFailMessage("Balances are equal, but should not be!")
```

Likewise on the other assertion. It's helpful to somehow describe _why_ the failure is a problem

mhess-swl (approved):
For the most part my comments are tiny doc nits, and they can be done in future PRs (if at all)

I suggest we get input from other team members on the atomic vars discussion, but again, promised to be quiet, so 🤐 😂 

Thanks @Evdokia-Georgieva!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
mhess-swl: Thanks for making these tests independent, @Evdokia-Georgieva!! Our future selves will all thank us 😅

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/NodeOperatorQueriesBase.java
```diff
@@ -0,0 +1,56 @@
...
+    protected static SpecOperation[] createAllAccountsAndTokens() {
+
+        final var createOperations = new ArrayList<SpecOperation>(List.of(
+                cryptoCreate(OWNER).balance(0L),
+                tokenCreate(FUNGIBLE_QUERY_TOKEN)
+                        .treasury(OWNER)
+                        .tokenType(FUNGIBLE_COMMON)
+                        .initialSupply(100L)));
+
+        return createOperations.toArray(new SpecOperation[0]);
```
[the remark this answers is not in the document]
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/NodeOperatorQueriesBase.java
```diff
@@ -0,0 +1,56 @@
...
+    // node operator account
+    protected static final String NODE_OPERATOR = "operator";
+    protected static final String FUNGIBLE_QUERY_TOKEN = "fungibleQueryToken";
+    protected static final String OWNER = "owner";
+    protected static final String PAYER = "payer";
+
+    /**
+     * Create Node Operator account
+     * Create all other accounts
+     * Create tokens
```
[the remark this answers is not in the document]
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/infrastructure/HapiClients.java
Evdokia-Georgieva: Fixed :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/utils/GrpcUtils.java
Evdokia-Georgieva: Can we do that with a separate task?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/test/grpc/GrpcTestBase.java
Evdokia-Georgieva: Thanks :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
Evdokia-Georgieva: Got it, I think I managed to upgrade the comments and assertion messages.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
Evdokia-Georgieva: Matt, I understand your concerns, here is how I can change the tests (still not committed):

// the tests but without the atomic long variables
        @HapiTest
        final Stream<DynamicTest> nodeOperatorQueryVerifyPayerBalanceForAccountBalance_WithoutAtomicLong() {
            return hapiTest(
                    cryptoCreate(NODE_OPERATOR).balance(ONE_HUNDRED_HBARS),
                    cryptoCreate(PAYER).balance(ONE_HUNDRED_HBARS),
                    // perform getAccountBalance() query, pay for the query with payer account
                    // the grpc client performs the query to different ports
                    getAccountBalance(NODE_OPERATOR).payingWith(PAYER),
                    sleepFor(3_000),
                    // assert payer is charged
                    getAccountBalance(PAYER).hasTinyBars(ONE_HUNDRED_HBARS),
                    // perform free query to local port with asNodeOperator() method
                    getAccountBalance(NODE_OPERATOR).payingWith(PAYER).asNodeOperator(),
                    sleepFor(3_000),
                    // assert payer is not charged as the query is performed as node operator
                    getAccountBalance(PAYER).hasTinyBars(ONE_HUNDRED_HBARS));
        }

        @HapiTest
        final Stream<DynamicTest> nodeOperatorQueryVerifyPayerBalanceForAccountInfo_WithoutAtomicLong() {
            return hapiTest(
                    cryptoCreate(NODE_OPERATOR).balance(ONE_HUNDRED_HBARS),
                    cryptoCreate(PAYER).balance(ONE_HUNDRED_HBARS),
                    // perform getAccountInfo() query, pay for the query with payer account
                    // the grpc client performs the query to different ports
                    getAccountInfo(NODE_OPERATOR).payingWith(PAYER),
                    sleepFor(3_000),
                    // assert payer is charged
                    getAccountBalance(PAYER).hasTinyBars(ONE_HUNDRED_HBARS - QUERY_COST),
                    // perform free query to local port with asNodeOperator() method
                    getAccountInfo(NODE_OPERATOR).payingWith(PAYER).asNodeOperator(),
                    sleepFor(3_000),
                    // assert payer is not charged as the query is performed as node operator
                    getAccountBalance(PAYER).hasTinyBars(ONE_HUNDRED_HBARS - QUERY_COST));
        }

   For the first test there is no headache as the query is free of charge in general, but what bothers me is the case when we will be charged for a query. What I understand is that we will have to know the exact amount that will be charged for the query. Can we rely that this amount will be always the same and the tests won't fail for example after some release, because these charges were changed? If we are sure that the amounts are always the same, then it is ok, but if we are not sure, then the risk here will be that potentially these tests might start failing because we are asserting concrete amounts and they will require some maintenance in the future. 
   That's what was bothering me here from the beginning and this is why I created the tests with the variables, I was thinking for some kind of flexibility but of course I will change them right away if you think that is better, that was just my vision at the beginning.

mhess-swl: Query and transaction costs don't change very much, generally speaking. I've never implemented a cost change, but I imagine when we do, we revisit all of the tests that fail to make sure they are still relevant and set up correctly. 

But guess what? I found a test that does something similar [here](https://github.com/hashgraph/hedera-services/blob/2c30a3ac0ddb31a4f2289b33ddc8ac0c20f5bb70/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoTransferSuite.java#L2080)! It uses the `balanceSnapshot(...)` and `changeFromSnapshot(...)` methods to compute expected deltas without the atomic vars 😄 What do you think of something like this?

```java
    @HapiTest
    final Stream<DynamicTest> nodeOperatorQueryVerifyPayerBalanceForAccountInfo_WithoutAtomicLong() {
        return hapiTest(
                cryptoCreate(NODE_OPERATOR).balance(ONE_HUNDRED_HBARS),
                cryptoCreate(PAYER).balance(ONE_HUNDRED_HBARS),
                balanceSnapshot("payerInitialBalance", PAYER),
                // perform getAccountInfo() query, pay for the query with payer account
                // the grpc client performs the query to different ports
                getAccountInfo(NODE_OPERATOR).payingWith(PAYER),
                sleepFor(3_000),
                // assert payer is charged
                getAccountBalance(PAYER).hasTinyBars(changeFromSnapshot("payerInitial", -QUERY_COST)),
                // perform free query to local port with asNodeOperator() method
                getAccountInfo(NODE_OPERATOR).payingWith(PAYER).asNodeOperator(),
                sleepFor(3_000),
                // assert payer is not charged as the query is performed as node operator
                getAccountBalance(PAYER).hasTinyBars(changeFromSnapshot("payerInitial", -QUERY_COST)));
    }
```

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/utils/GrpcUtils.java
mhess-swl: Sure, whatever's easiest. Making this change now would eliminate this entire block of changes (lines 121 - 229), that's mainly why I suggested it, but doing it later is fine too :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/NodeOperatorQueriesBase.java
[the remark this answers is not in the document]
mhess-swl: I'm not seeing this change or the change below in this file, but that's ok. They're tiny suggestions anyway :)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/queries/AsNodeOperatorQueriesTest.java
mhess-swl: I didn't realize that node operator queries were a high priority, so we don't need to spend lots of time on this now. Let's get this PR merged, then we can come back to this.

imalygin (dismissed):

mhess-swl:
Something weird happened to this PR, there are 285 files changed and a _ton_ of reviewers required 😞

mishomihov00 (dismissed):
My review applies only to all `.yaml`, `.sh` and Dockerfiles.

petreze (approved):

PR #15774, opened 2024-10-02
chore: remove unnecessary checks for value xfer to system contracts

### LINKED ISSUE #13286: Consistently handle calls with value to system contracts
### Problem

Right now, there is code to surface an error to the caller (`INVALID_FEE_SUBMITTED`) when a call is made to a system contract with value.  Each system contract is responsible to detecting this error condition and responding.  A universal check was later added but the checks inside the system contract still remain.

### Solution

- Re-examine if `INVALID_FEE_SUBMITTED` is the appropriate error code to return to the user.
- Remove checks for value transfer from within system contracts

### Alternatives

_No response_

lukelee-sl:
**Description**:
There are checks in some of the system contracts which ensure that value is not passed to the contract.  These checks have been made superfluous by this pr #14843 and thus need to be removed.

**Related issue(s)**:

Fixes #13286

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/ExchangeRateSystemContract.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/ExchangeRateSystemContract.java
@@ -63,7 +63,6 @@ public class ExchangeRateSystemContract extends AbstractFullContract implements 
         requireNonNull(messageFrame);
         try {
             validateTrue(input.size() >= 4, INVALID_TRANSACTION_BODY);
-            validateTrue(messageFrame.getValue().getAsBigInteger().equals(BigInteger.ZERO), INVALID_FEE_SUBMITTED);
             gasRequirement = contractsConfigOf(messageFrame).precompileExchangeRateGasCost();
             final var selector = input.getInt(0);
             final var amount = biValueFrom(input);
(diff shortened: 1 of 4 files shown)


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/ExchangeRateSystemContract.java
```diff
@@ -63,7 +63,6 @@ public FullResult computeFully(@NonNull Bytes input, @NonNull MessageFrame messa
         requireNonNull(messageFrame);
         try {
             validateTrue(input.size() >= 4, INVALID_TRANSACTION_BODY);
-            validateTrue(messageFrame.getValue().getAsBigInteger().equals(BigInteger.ZERO), INVALID_FEE_SUBMITTED);
```
david-bakin-sl, line 66: What about L82-83?  Unneeded now?

david-bakin-sl: And see also `ExchangeRateSystemContractTest.valueShouldNotBeSentToThePrecompile` - doesn't seem still applicable?

david-bakin-sl (commented):
And now I'm also wondering if `CustomExceptionalHaltReason.INVALID_FEE_SUBMITTED` is necessary?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/ExchangeRateSystemContract.java
lukelee-sl: good point.  removed both.

david-bakin-sl (approved):
LGTM

PR #15778, opened 2024-10-02
chore: Add missing config files to `previewnet` config dir

### LINKED ISSUE #15777: Standardize previewnet config dir with other envs
The `previewnet` configuration directory is missing some of the files present in other environment config directories, which is causing a devops script to fail. We need to standardize `previewnet` to match the other environments.

mhess-swl:
Closes #15777


Neeharika-Sompalli (approved):
LGTM! Thanks @mhess-swl 

This is to ensure we have same directory folders for previewnet as `mainnet` and `testnet` as devops requested.

Nana-EC (approved):

PR #15782, opened 2024-10-03
15584 consensus update topic handler additional hapi tests

vtronkov:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hapi/hedera-protobufs/services/response_code.proto
+++ b/hapi/hedera-protobufs/services/response_code.proto
@@ -1626,4 +1626,11 @@ enum ResponseCodeEnum {
     * we cannot add it on update.
     */
     FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED = 375;
+
+    /**
+    * If the topic's custom fees are updated the topic SHOULD have a
+    * fee schedule key
+     */
+    FEE_SCHEDULE_KEY_NOT_SET = 376;
+
 }
--- a/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
+++ b/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
@@ -346,7 +346,11 @@ public class ConsensusUpdateTopicHandler implements TransactionHandler {
             @NonNull final Topic topic) {
         if (op.hasFeeScheduleKey()
                 && !KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
-            validateTrue(topic.hasFeeScheduleKey(), FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED);
+            validateTrue(
+                    topic.hasFeeScheduleKey()
+                            && !KeyOneOfType.UNSET.equals(
+                                    topic.feeScheduleKey().key().kind()),
+                    FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED);
             attributeValidator.validateKey(op.feeScheduleKey(), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
         }
     }
--- a/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
+++ b/hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
@@ -20,6 +20,7 @@ import static com.hedera.hapi.node.base.ResponseCodeEnum.AUTORENEW_ACCOUNT_NOT_A
 import static com.hedera.hapi.node.base.ResponseCodeEnum.AUTORENEW_DURATION_NOT_IN_RANGE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.CUSTOM_FEES_LIST_TOO_LONG;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.FEE_SCHEDULE_KEY_NOT_SET;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.FEKL_CONTAINS_DUPLICATED_KEYS;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_AUTORENEW_ACCOUNT;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_CUSTOM_FEE_SCHEDULE_KEY;
@@ -145,10 +146,16 @@ public class ConsensusUpdateTopicHandler implements TransactionHandler {
             }
         }
 
-        // If the fee schedule key is change then the transaction must also be signed by the new fee schedule key
+        // If the fee schedule key is changed then the transaction must also be signed by the new fee schedule key
         if (op.hasFeeScheduleKey()
                 && !KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
             context.requireKeyOrThrow(op.feeScheduleKey(), INVALID_FEE_SCHEDULE_KEY);
+        }
+
+        // If we change the custom fees the topic needs to have a fee schedule key, and it needs to sign the transaction
+        if (op.hasCustomFees()) {
+            validateTruePreCheck(topic.hasFeeScheduleKey(), FEE_SCHEDULE_KEY_NOT_SET);
+            context.requireKey(topic.feeScheduleKey());
         }
     }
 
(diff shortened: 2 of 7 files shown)


hapi/hedera-protobufs/services/response_code.proto
```diff
@@ -1612,12 +1612,18 @@ enum ResponseCodeEnum {
     /**
     * The provided fee schedule key contains an invalid key.
     */
-    INVALID_FEE_SCHEDULE_KEY = 374;
+    INVALID_FEE_SCHEDULE_KEY = 373;
```
kimbor, line 1615: What is the difference between this code and `INVALID_CUSTOM_FEE_SCHEDULE_KEY = 247;`? It seems like this one is duplicative.

vtronkov: Yep, it seems to be duplicative. Removed 👍

hapi/hedera-protobufs/services/response_code.proto
```diff
@@ -1608,16 +1608,17 @@ enum ResponseCodeEnum {
...
-    FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED = 375;
-    
+    FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED = 373;
+
+    /**
+    * If the topic's custom fees are updated the topic SHOULD have a
+    * fee schedule key
+     */
+    FEE_SCHEDULE_KEY_NOT_SET = 374;
+
```
mhess-swl, line 1623: We need to be careful about changing this, since applications could rely on the previous behavior. Do we have approval to add this response cocde?

mhess-swl (commented):
The title on this PR is misleading. This isn't just a PR adding some tests; these are functionality changes

hapi/hedera-protobufs/services/response_code.proto
mhess-swl: Ah, I see now that this PR would merge to `hip-991-topic-fees`, my mistake

vtronkov:
> The title on this PR is misleading. This isn't just a PR adding some tests; these are functionality changes

Changed 👍 

Also, node that the HIP is on freeze at the moment. I will move the branch in draft mode.

mhess-swl:
> > The title on this PR is misleading. This isn't just a PR adding some tests; these are functionality changes
> 
> Changed 👍
> 
> Also, node that the HIP is on freeze at the moment. I will move the branch in draft mode.

Thank you @vtronkov! 🙏

hapi/hedera-protobufs/services/response_code.proto
```diff
@@ -1637,5 +1637,11 @@ enum ResponseCodeEnum {
     * we cannot add it on update.
     */
     FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED = 377;
-
+  
+    /**
+    * If the topic's custom fees are updated the topic SHOULD have a
+    * fee schedule key
```
iwsimon, line 1643: Those * are not aligned correctly. You will have to update before merging to develop.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -146,8 +147,15 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
         }
 
         // If the fee schedule key is changed then the transaction must also be signed by the new fee schedule key
-        if (op.hasFeeScheduleKey()) {
-            context.requireKeyOrThrow(op.feeScheduleKey(), INVALID_FEE_SCHEDULE_KEY);
+        if (op.hasFeeScheduleKey()
+                && !KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
```
iwsimon, line 151: I don't think you need this checking: !KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())). requireKeyOrThrow() takes care of it.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -146,8 +147,15 @@ public void preHandle(@NonNull final PreHandleContext context) throws PreCheckEx
...
+            context.requireKeyOrThrow(op.feeScheduleKey(), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
```
iwsimon, line 152: If modifying feeScheduleKey, don't we also need current feeScheduleKey to sign too?

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -246,7 +254,12 @@ private void resolveMutableBuilderAttributes(
             builder.autoRenewAccountId(resolvedExpiryMeta.autoRenewAccountId());
         }
         if (op.hasFeeScheduleKey()) {
-            builder.feeScheduleKey(op.feeScheduleKey());
+            final var newSubmitKey = op.feeScheduleKey();
```
iwsimon, line 257: Rename newSubmitKey to newFeeScheduleKey so we don't confuse with real submitKey.

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
```diff
@@ -338,7 +351,9 @@ private void validateMaybeNewFeeScheduleKey(
             @NonNull final Topic topic) {
         if (op.hasFeeScheduleKey()) {
             validateTrue(topic.hasFeeScheduleKey(), FEE_SCHEDULE_KEY_CANNOT_BE_UPDATED);
-            attributeValidator.validateKey(op.feeScheduleKey(), INVALID_CUSTOM_FEE_SCHEDULE_KEY);
+            if (!KeyOneOfType.UNSET.equals(op.feeScheduleKey().key().kind())) {
```
iwsimon, line 354: Not needed,validateKey() takes care of it.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip991/TopicCustomFeeUpdateTest.java
```diff
@@ -16,236 +16,1089 @@
...
+
+        @HapiTest
+        @DisplayName("to replace all fee exempt keys")
+        final Stream<DynamicTest> updateToReplaceAllKeys() {
+            final var key = "key";
+            final var key2 = "key2";
+            final var key3 = "key3";
+            final var key4 = "key4";
+
+            return hapiTest(flattened(
```
iwsimon, line 409: You don't need all those "flattened".

hapi/hedera-protobufs/services/response_code.proto
vtronkov: Yeah, when we merge this PR to `hip-991-topic-fees` and we finalize it we can align the proto messages 👍

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
vtronkov: You can see for example `updateTheFeeScheduleKeyToEmpty` test. It uses `withEmptyFeeScheduleKey` that sets an empty Key. When we send it like this the `op.hasFeeScheduleKey()` is **true** and we try to verify the key but it doesn't exist. We need this additional check so we know that if there is a key it's not UNSET

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
vtronkov: As far as I understand it - no. It's enough to be signed by the admin key and the new feeScheduleKey

hedera-node/hedera-consensus-service-impl/src/main/java/com/hedera/node/app/service/consensus/impl/handlers/ConsensusUpdateTopicHandler.java
vtronkov: validateKey will throw an error if it's UNSET. We want to call validateKey only if it's an actual Key.

kimbor (approved):
lgtm!

Issue #15542, opened 2024-09-18
EventDescriptorWrapper should not copy Bytes


lpetrovic05:
`EventDescriptorWrapper` copies the hash bytes in the constructor. This is unnecessary, the `Hash` class should have a constructor that accepts `Bytes` and no copying should be done.

PR #15783, opened 2024-10-03
chore: Add constructor for `Bytes` in `Hash` to avoid copying

timo0:
Fix [#15542](https://github.com/hashgraph/hedera-services/issues/15542)

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
@@ -48,17 +48,31 @@ public class Hash implements Comparable<Hash>, SerializableWithKnownLength, Seri
     }
 
     /**
-     * Same as {@link #Hash(byte[], DigestType)} but with an empty byte array.
+     * Same as {@link #Hash(Bytes, DigestType)} but with an empty byte array.
      */
     public Hash(@NonNull final DigestType digestType) {
-        this(new byte[digestType.digestLength()], digestType);
+        this(Bytes.wrap(new byte[digestType.digestLength()]), digestType);
     }
 
     /**
-     * Same as {@link #Hash(byte[], DigestType)} but with a digest type ({@link DigestType#SHA_384})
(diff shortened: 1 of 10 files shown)


lpetrovic05 (approved):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -48,19 +48,33 @@ public Hash() {
     }
 
     /**
-     * Same as {@link #Hash(byte[], DigestType)} but with an empty byte array.
+     * Same as {@link #Hash(Bytes, DigestType)} but with an empty byte array.
      */
     public Hash(@NonNull final DigestType digestType) {
-        this(new byte[digestType.digestLength()], digestType);
+        this(Bytes.wrap(new byte[digestType.digestLength()]), digestType);
```
OlegMazurov, line 54: Wrapping in Bytes will affect hashing which instantiates Hashes orders of magnitude more than the rest of the system.
The performance impact needs to be evaluated.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
```diff
@@ -70,16 +84,16 @@ public Hash(@NonNull final byte[] value) {
...
         Objects.requireNonNull(digestType, "digestType");
 
-        if (value.length != digestType.digestLength()) {
-            throw new IllegalArgumentException("value: " + value.length);
+        if ((int) value.length() != digestType.digestLength()) {
+            throw new IllegalArgumentException("value: " + value.length());
         }
 
         this.digestType = digestType;
-        this.bytes = Bytes.wrap(value);
```
timo0, line 82: @OlegMazurov I just changed the point where the bytes got wrapped.

tinker-michaelj (approved):
Makes sense to me, ty @timo0 !

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/crypto/Hash.java
timo0: please see my other comment

OlegMazurov (approved):

PR #15791, opened 2024-10-03
chore: remove nextNodeId from config.txt

### LINKED ISSUE #15215: Remove the nextNodeId line from the config.txt
The last line, nextNodeId, # in config.txt is not used by services. 
Since the next node id is now kept in the application state, there is no need to keep it in config.txt. 
Platform code checks this line. This line should be removed.

edward-swirldslabs:
**Description**:
This PR removes the use of `nextNodeId` from config.txt and removes it from use in validating `AddressBook` data structures.   The getter and setter and internal field are left in the code for compatibility with the protobuf serialization scheme.  

A new method `getNextAvailableNodeId` has been added to replace usage in the code where the value was retrieved for creating new addresses with node id of value 1+ the highest used in the address book.  This method is only used in unit tests. 

Fixes #15215 

This PR needs to make sure the nightly JRS tests and HAPI pass related to updating the address book. 

must be merged in tandem with: https://github.com/swirlds/swirlds-platform-regression/pull/4017


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBook.java
```diff
@@ -92,6 +92,10 @@ private static class ClassVersion {
     private long round = UNKNOWN_ROUND;
 
     /**
+     * DEPRECATED FIELD as of v0.56.0 It remains for compatibility with protobuf serialization.  It's value will always
```
anthony-swirldslabs, line 95:
```suggestion
     * DEPRECATED FIELD as of v0.56.0 It remains for compatibility with protobuf serialization.  Its value will always
```

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/network/TopologyTest.java
```diff
@@ -126,7 +126,8 @@ void testFullyConnectedTopology(final int numNodes, final long ignoredSeed) {
         final AddressBook addressBook =
                 RandomAddressBookBuilder.create(randotron).withSize(numNodes).build();
         for (int thisNode = 0; thisNode < numNodes; thisNode++) {
-            final NodeId outOfBoundsId = addressBook.getNextNodeId();
+            final NodeId outOfBoundsId =
+                    addressBook.getNodeId(addressBook.getSize() - 1).getOffset(1);
```
anthony-swirldslabs, line 130: Is there a reason not to use the `getNextAvailableNodeId` here as well? That would be more consistent with the rest of the tests, and would simplify any future refactoring involving this method.

anthony-swirldslabs (approved):

platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/test/java/com/swirlds/platform/test/network/TopologyTest.java
edward-swirldslabs: agreed

kfa-aguda (dismissed):
lgtm

anthony-swirldslabs (approved):

kfa-aguda (commented):
Question - I've just noticed `platform_state.proto` (validly outside of this PR) has next node id definition in it. Is the plan to remove that as part some separate protobuf cleanup work in a future PR?

anthony-swirldslabs (approved):

edward-swirldslabs:
> Question - I've just noticed `platform_state.proto` (validly outside of this PR) has next node id definition in it. Is the plan to remove that as part some separate protobuf cleanup work in a future PR?

Given that `AddressBook` is not long for this world, I don't think it's worth going through the headache of updating the proto and having to perform state migrations with the changed proto definition.  I left things so that serialization and deserialization with proto and the legacy serialization schemes will still work.

mhess-swl (approved):
Looks good for services 👍

kfa-aguda (approved):

rbair23 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBook.java
```diff
@@ -423,24 +440,26 @@ private void updateAddress(@NonNull final Address address) {
...
      *
      * @param address the address to add
      */
     private void addNewAddress(@NonNull final Address address) {
         final NodeId addressNodeId = address.getNodeId();
-        final int nodeIdComparison = addressNodeId.compareTo(nextNodeId);
+        final int addressBookSize = getSize();
+        final NodeId nextAvailable = addressBookSize == 0
+                ? NodeId.FIRST_NODE_ID
+                : getNodeId(getSize() - 1).getOffset(1);
```
iwsimon, line 452: I am wondering if this nextAvailable is correct.

edward-swirldslabs: We discussed on a zoom call, this is only preventing the insert of a new address between existing address and not trying to to model `nextNodeId` with multiple deletes at the end of a previous address book.

iwsimon (approved):

PR #15793, opened 2024-10-03
feat: Regenerate keys and update node names for tests

### LINKED ISSUE #15789: Regenerate keys and update node names for tests
The refactoring to replace AB with Roster forces the code to use the new `node<id+1>` names for nodes, which breaks existing tests that use pre-generated keys/certs with names of form `aaaa`, `aaab`, etc. Additionally, these pre-generated keys and certs currently use an old .pfx file format and a legacy key loading mechanism.

This story covers:
1. Updating the node names in the tests' AB to follow the new pattern of `node<id+1>`.
2. Updating the pre-generated keys to use the new node names, and making use of the new key loading mechanism from .pem files instead of the legacy .pfx files.

anthony-swirldslabs:
**Description**:
1. Replacing legacy .pfx keys/certs with newly generated .pem files.
2. Making unit tests that use these keys load them via the `EnhancedKeyStoreLoader`.
3. Updating node names in these tests to follow the new implied pattern of `node<id+1>` instead of the old custom `aaaa`, `aaab`, etc.

**Related issue(s)**:

Fixes #15789

**Notes for reviewer**:
Automated tests should pass.

**Checklist**


platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/crypto/CryptoArgsProvider.java
```diff
@@ -41,8 +47,7 @@ public class CryptoArgsProvider {
     private static final char[] PASSWORD = "password".toCharArray();
 
     /**
-     * @return 3 sets of arguments, 1 generated, 1 loaded from files, and 1 that generates agreement keys if they are
-     * not loaded from files.
+     * @return 2 sets of arguments, 1 generated, 1 loaded from files.
```
anthony-swirldslabs, line 50: That's because the new keys in the .pem files only contain the signing keys. IIRC, the agreement keys are no longer relevant, but if they are needed anywhere, I presume that the key loader will generate them.

So there's only 2 sets of arguments now.

edward-swirldslabs (approved):

PR #15834, opened 2024-10-06
fix: detect post-upgrade txn in presence of pre-upgrade events

### LINKED ISSUE #15833: (0.54+) Post-upgrade boundary should not be disrupted by previous version events
### Description

We have seen the first post-upgrade transaction frequently fail to be detected in environments where it is common for events with an earlier software version to reach consensus post-upgrade. 

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

testnet

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #15833 
 - Instead of unconditionally marking post-upgrade done in `BlockRecordManager#advanceConsensusClock()` (⚠️ which must _also_ called for a skipped pre-upgrade event to avoid an ISS on reconnect in this post-upgrade period), only marks the post-upgrade work done in the `HandleWorkflow` branch for transactions of type `POST_UPGRADE_TRANSACTION`.
 - 👉 Adds and uses a `runBackgroundTrafficUntilFreezeComplete()` spec operation to ensure pre-upgrade events are likely to reach consensus after every upgrade in `DabEnabledUpgradeTest`.
 - Simplifies classification of the `POST_UPGRADE_TRANSACTION` to a single condition: Whether the `BlockInfo#migrationRecordsStreamed` flag is `false`.
 - ✔️ Fixes a previously unnoticed race condition in `FeeManager` that could cause fees to be miscomputed by ingest/query threads operating concurrently with a fee schedule update.


thomas-swirlds-labs (approved):
LGTM ty @tinker-michaelj 👍

iwsimon (approved):
LGTM, thanks @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/FeeManager.java
```diff
@@ -135,9 +135,11 @@ public ResponseCodeEnum update(@NonNull final Bytes bytes) {
...
         }
 
-        // Populate the map of HederaFunctionality -> FeeData for the current schedule
-        this.currentFeeDataMap = new HashMap<>();
-        populateFeeDataMap(currentFeeDataMap, currentSchedule.transactionFeeSchedule());
+        // Populate the map of HederaFunctionality -> FeeData for the current schedule, but avoid mutating
+        // the active one in-place as other threads may be using it for ingest/query fee calculations
+        final var newCurrentFeeDataMap = new HashMap<Entry, FeeData>();
+        populateFeeDataMap(newCurrentFeeDataMap, currentSchedule.transactionFeeSchedule());
+        this.currentFeeDataMap = newCurrentFeeDataMap;
```
rbair23, line 142: This is interesting. It looks like the issue is that FeeManager needs to be threadsafe, but is not. I don't think this is sufficient to actually ensure thread safety, because the fields themselves are not either `AtomicReference` or `transient`, or guarded by synchronization. It is possible two CPUs could see different references to the hashmaps for an unspecified amount of time (some memory barrier must be crossed forcing the CPUs to flush their local caches and re-read the shared state).

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -587,7 +587,8 @@ private List<StateChanges.Builder> onMigrate(
         migrationStateChanges.addAll(migrationChanges);
         kvStateChangeListener.reset();
         boundaryStateChangeListener.reset();
-        if (isUpgrade && !trigger.equals(RECONNECT)) {
+        // For specifically a non-genesis upgrade, set in state that post-upgrade work is pending
+        if (isUpgrade && trigger != RECONNECT && trigger != GENESIS) {
```
rbair23, line 591: Was the non-genesis case actually needed or was it just an oversight and this new code is "tighter"?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/records/schemas/V0490BlockRecordSchema.java
```diff
@@ -67,7 +67,8 @@ public void migrate(@NonNull final MigrationContext ctx) {
         final var isGenesis = ctx.previousVersion() == null;
         if (isGenesis) {
             final var blocksState = ctx.newStates().getSingleton(BLOCK_INFO_STATE_KEY);
-            final var blocks = new BlockInfo(-1, EPOCH, Bytes.EMPTY, EPOCH, false, EPOCH);
+            // Note there is by convention no post-upgrade work to do if starting from genesis
+            final var blocks = new BlockInfo(-1, EPOCH, Bytes.EMPTY, EPOCH, true, EPOCH);
```
rbair23, line 71: This fix also I assume is just better, but didn't actually contribute to the fix of the bug that triggered this PR? (It is great to include this, I just want to make sure I understand what is happening)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/fees/FeeManager.java
tinker-michaelj: True, because fee schedules updates are _extremely_ infrequent, I didn't want to add any synchronization for the general case of safe concurrent reads from ingest/query threads and the handle thread.

This change just ensures that _if_ handle is changing the fee schedules while an ingest/query thread reading `FeeData`, the ingest/query thread will at least see a consistent set of schedules instead of discovering an empty `HashMap` with completely missing `FeeData` and logging a scary warning.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/UserTxn.java
```diff
@@ -263,24 +261,14 @@ public StreamBuilder baseBuilder() {
...
      */
     private static boolean isUpgradeBoundary(@NonNull final State state) {
-        final var platformState = state.getReadableStates(PlatformStateService.NAME)
-                .<PlatformState>getSingleton(V0540PlatformStateSchema.PLATFORM_STATE_KEY)
+        final var blockInfo = state.getReadableStates(BlockRecordService.NAME)
+                .<BlockInfo>getSingleton(BLOCK_INFO_STATE_KEY)
                 .get();
-        requireNonNull(platformState);
-        if (platformState.freezeTime() == null
-                || !platformState.freezeTimeOrThrow().equals(platformState.lastFrozenTime())) {
```
mhess-swl, line 275: IIRC, this is being removed so we can depend solely on the `migrationRecordsStreamed` property?

tinker-michaelj: True @mhess-swl .

mhess-swl (approved):

mhess-swl (approved):

iwsimon (approved):

mhess-swl (approved):

PR #15836, opened 2024-10-06
fix: (0.54) post-upgrade txn detection in presence of pre-upgrade events

tinker-michaelj:
**Description**:
 - Cherry-pick #15834


tinker-michaelj:
Not urgent for 0.54.

PR #15837, opened 2024-10-07
refactor: gossip modules creation

### LINKED ISSUE #15646: Create gossip modules in consensus layer package
Create 2 new modules in the consensus package:
- consensus-gossip
- consensus-gossip-impl

Follow this doc for the structure of the modules:
 https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/modules.md

mustafauzunn:
**Description**:

Initial gossip modules creation

**Related issue(s)**:

Fixes #15646 

**Notes for reviewer**:

**Checklist**


consensus/consensus-gossip/src/main/java/com/hedera/service/gossip/DefaultGossipService.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.hedera.service.gossip;
+
+/*
+ *
+ * Default mock gossip service
+ * */
+public interface DefaultGossipService {}
```
rbair23, line 23: I think the name here should be just "GossipService" instead of "DefaultGossipService"

settings.gradle.kts
```diff
@@ -35,6 +35,13 @@ javaModules {
         module("swirlds-unit-tests/core/swirlds-platform-test") // nested module is not found automatically
     }
 
+    directory("consensus") {
+        group = "com.hedera.hashgraph"
+
+        module("consensus-gossip") { artifact = "app-consensus-gossip" }
```
rbair23, line 41: What is the artifact line doing? Does this mean the jar file will be called "app-consensus-gossip"?

rbair23:
For now, can we put consensus-gossip and consensus-gossip-impl into platform-sdk instead of a new directory called "consensus"? I think for the refactoring period, we keep the modules close to each other (all the swirlds-* and the new ones) and then in the end we either rename `platform-sdk` to `consensus`, or what I really think we should do is create a new top level directory called `modules` and we have two sub directories called `consensus` and `execution` and then we put our modules in those as appropriate. But rather than doing all this moving around right now, I'm thinking we can fit within the current organization until we have finished the changes. Maybe just as an easier first step.

platform-sdk/consensus-gossip-impl/src/main/java/module-info.java
```diff
@@ -0,0 +1,8 @@
+module com.hedera.consensus.gossip.impl {
+    exports com.hedera.service.gossip.impl;
```
rbair23, line 2: Do we need to export this? Or is implementing the GossipServiceImpl (`provides` clause below) sufficient?

platform-sdk/consensus-gossip/src/main/java/com/hedera/service/gossip/GossipService.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.service.gossip;
+
+/*
+ *
+ * Mock gossip service
+ * */
```
rbair23, line 22: Can we fix the comment formatting?

rbarkerSL (approved):
LGTM

IvanKavaldzhiev (approved):
LGTM

rbarkerSL (approved):

PR #15931, opened 2024-10-09
chore: remove unused wiring options

### LINKED ISSUE #15630: Remove unused wiring options
Some config options for wiring are no longer used and should be removed, as well as all code associated with it. The config options are:
- `PlatformSchedulersConfig.hashCollectorEnabled`
- `EventConfig.useOldStyleIntakeQueue`

lpetrovic05:
closes #15630

regression PR: https://github.com/swirlds/swirlds-platform-regression/pull/4020


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/sync/protocol/SyncProtocol.java
```diff
@@ -175,11 +167,6 @@ private boolean shouldSync() {
             return false;
         }
 
-        if (intakeIsTooFull.getAsBoolean()) {
```
litt3, line 178: I think that this change exceeds a simple removal of an unused option. Even without the "legacy intake queue", if I'm reading things correctly, it appears that this check prevents a node from syncing if the front of the intake pipeline gets too full.

There may be an argument for letting the health monitor be entirely responsible for turning off the gossip faucet, but it should at least be noted that this is a change which could have an effect on gossip behavior.

lpetrovic05: `intakeIsTooFull` is a lambda in `SyncGossip:347`:
```
() -> intakeQueueSizeSupplier.getAsLong() >= eventConfig.eventIntakeQueueThrottleSize()
```

`intakeQueueSizeSupplier` is created in two places, depending on the config
`SwirldsPlatform:265`:
```
        if (useOldStyleIntakeQueue) {
            final SyncGossip gossip = (SyncGossip) builder.buildGossip();
            intakeQueueSizeSupplier = () -> gossip.getOldStyleIntakeQueueSize();
        } else {
            intakeQueueSizeSupplier = platformWiring.getIntakeQueueSizeSupplier();
        }
```

`PlatformWiring:1051`:
```
        if (hashCollectorEnabled) {
            return () -> postHashCollectorWiring.getScheduler().getUnprocessedTaskCount();
        } else {
            return () -> 0;
        }
```

So, with our current settings, this is not being used at all. If we are removing settings that invoke this code, then I think the code should be removed as well.

litt3: Makes sense

litt3 (approved):

PR #15933, opened 2024-10-09
fix: Airdrop transfer list size validation

### LINKED ISSUE #15906: TokenAirdrop missing semantic validations from CryptoTransfer logic
### Description

The token airdrop validations are missing a semantic check to the amount of `fungibleTransfers` and `nftTransfers` inside a TransferList element while the CryptoTransfer validator is checking that amount and limits it by `ledger.tokenTransfers.maxLen`  and `ledger.nftTransfers.maxLen` configuration properties. 
As per HIP-904 both transaction should use or reuse validation logic.

### Steps to reproduce

1) Sender account tries to execute TokenAirdrop, sending 1 FT each to 10 receiving account. 
2) This will result in 1 TransferList with 11 elements of account amounts. 
Expected result: The transaction should fail with `TOKEN_TRANSFER_LIST_SIZE_LIMIT_EXCEEDED` or a relative error.
Actual result: The transaction is successful 

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.55

### Operating system

None

JivkoKelchev:
**Description**:

Replaces the current transfer list size validation with one that complies with the CryptoTransfer behaviour.

**Related issue(s)**:

Fixes #15906

**Notes for reviewer**:

PR, that replaces the current transfer list size validation with one that complies with the CryptoTransfer behavior :
1. Allowing no more than 10 fungible token **balance adjustments**.
- In case of one fungible token airdrop, it is possible up to 9 receivers (1 adjustment for the sender and 9 for receivers)
- In case of multiple fungible tokens, it is possible to make up to 5 token transfers (5 x 2 adjustments of the sender and the receiver balances)
2. Allowing no more than 10 NFT transfers

**Checklist**

### DIFF
--- a/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
+++ b/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
@@ -16,6 +16,7 @@
 
 package com.hedera.node.app.service.token.impl.validators;
 
+import static com.hedera.hapi.node.base.ResponseCodeEnum.BATCH_SIZE_LIMIT_EXCEEDED;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.EMPTY_TOKEN_TRANSFER_BODY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INSUFFICIENT_TOKEN_BALANCE;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_ACCOUNT_ID;
@@ -24,7 +25,7 @@ import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TOKEN_ID;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.INVALID_TRANSACTION_BODY;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.SENDER_DOES_NOT_OWN_NFT_SERIAL_NO;
 import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_AIRDROP_WITH_FALLBACK_ROYALTY;
-import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED;
+import static com.hedera.hapi.node.base.ResponseCodeEnum.TOKEN_TRANSFER_LIST_SIZE_LIMIT_EXCEEDED;
 import static com.hedera.node.app.service.token.impl.util.TokenHandlerHelper.getIfUsable;
 import static com.hedera.node.app.service.token.impl.util.TokenHandlerHelper.getIfUsableForAliasedId;
 import static com.hedera.node.app.service.token.impl.validators.CryptoTransferValidator.validateTokenTransfers;
@@ -48,7 +49,7 @@ import com.hedera.node.app.service.token.impl.WritableAccountStore;
 import com.hedera.node.app.service.token.impl.handlers.transfer.customfees.CustomFeeExemptions;
 import com.hedera.node.app.spi.workflows.HandleContext;
 import com.hedera.node.app.spi.workflows.PreCheckException;
-import com.hedera.node.config.data.TokensConfig;
+import com.hedera.node.config.data.LedgerConfig;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.List;
 import javax.inject.Inject;
@@ -97,11 +98,9 @@ public class TokenAirdropValidator {
             @NonNull final ReadableTokenStore tokenStore,
             @NonNull final ReadableTokenRelationStore tokenRelStore,
             @NonNull final ReadableNftStore nftStore) {
-        var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
-        validateTrue(
-                op.tokenTransfers().size() <= tokensConfig.maxAllowedAirdropTransfersPerTx(),
-                TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED);
-
+        var legerConfig = context.configuration().getConfigData(LedgerConfig.class);
+        var totalFungibleTransfers = 0;
+        var totalNftTransfers = 0;
         for (final var xfers : op.tokenTransfers()) {
             final var tokenId = xfers.tokenOrThrow();
             final var token = getIfUsable(tokenId, tokenStore);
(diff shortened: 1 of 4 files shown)


hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -118,6 +117,12 @@ public void validateSemantics(
                 // 1. Validate token associations
                 validateFungibleTransfers(
                         context.payer(), senderAccount, tokenId, senderAccountAmount.get(), tokenRelStore);
+                totalFungibleTransfers += xfers.transfers().size();
+
+                // Verify that the current total number of (counted) fungible transfers does not exceed the limit
+                validateTrue(
```
david-bakin-sl, line 123: Can we have a `//FUTURE` comment or better an issue in this repo to refactor this and `CryptoTransferValidator.validateSemantics` so that they use common code for these common checks.?

(There are differences in the specific validations done for crypto xfer vs token airdrop, but these _structural_ checks should be common code.  IMO.)

mhess-swl: Agreed. Since this is a release-blocking issue I suspect we should not refactor this here, but I'd also like to see the common validations extracted

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -97,11 +98,9 @@ public void validateSemantics(
             @NonNull final ReadableTokenStore tokenStore,
             @NonNull final ReadableTokenRelationStore tokenRelStore,
             @NonNull final ReadableNftStore nftStore) {
-        var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
-        validateTrue(
-                op.tokenTransfers().size() <= tokensConfig.maxAllowedAirdropTransfersPerTx(),
-                TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED);
-
```
mhess-swl, line 104: Isn't this check still needed? The check below is using `ledgerConfig.tokenTransfersMaxLen()` and `nftTransfersMaxLen()`, which are both different than `maxAllowedAirdropTransfersPerTx`

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -137,14 +142,11 @@ public void validateSemantics(
...
-                        tokenId,
-                        xfers.nftTransfers(),
-                        tokenRelStore,
-                        tokenStore,
-                        nftStore);
+                validateNftTransfers(senderAccount, tokenId, xfers.nftTransfers(), tokenRelStore, tokenStore, nftStore);
+
+                totalNftTransfers += xfers.nftTransfers().size();
+                // Verify that the current total number of (counted) nft transfers does not exceed the limit
+                validateTrue(totalNftTransfers <= legerConfig.nftTransfersMaxLen(), BATCH_SIZE_LIMIT_EXCEEDED);
```
mhess-swl, line 149: Curious, why use `BATCH_SIZE_LIMIT_EXCEEDED` as the error here? I would've expected `TOKEN_TRANSFER_LIST_SIZE_LIMIT_EXCEEDED` (again)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
```diff
@@ -2183,6 +2166,44 @@ final Stream<DynamicTest> airdropTo0x0Address() {
...
+        @DisplayName("pending airdrop 1 token to more than 10 accounts")
+        final Stream<DynamicTest> pendingAirdropOneTokenToMoreThan10Accounts() {
+            final var accountNames = generateAccountNames(10);
+            return hapiTest(flattened(
+                    // create 10 accounts with 0 auto associations
+                    createAccounts(accountNames, 0),
+                    tokenAirdrop(distributeTokens(FUNGIBLE_TOKEN, OWNER, accountNames))
+                            .payingWith(OWNER)
+                            .hasKnownStatus(TOKEN_TRANSFER_LIST_SIZE_LIMIT_EXCEEDED)));
+        }
```
mhess-swl, line 2180: We have a fungible token test here, and happen to have an NFT test in `TokenAirdropHandlerTest` (via calling the `transactionBodyAboveMaxTransferLimit()` method). It would be good to call out that we're testing NFT coverage specifically in `TokenAirdropHandlerTest`.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
JivkoKelchev: I'm wondering about the same thing, but since this is the response code returned from the crypto transfer, I decided to use it here too.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
```diff
@@ -97,11 +98,9 @@ public void validateSemantics(
             @NonNull final ReadableTokenStore tokenStore,
             @NonNull final ReadableTokenRelationStore tokenRelStore,
             @NonNull final ReadableNftStore nftStore) {
-        var tokensConfig = context.configuration().getConfigData(TokensConfig.class);
-        validateTrue(
-                op.tokenTransfers().size() <= tokensConfig.maxAllowedAirdropTransfersPerTx(),
-                TOKEN_REFERENCE_LIST_SIZE_LIMIT_EXCEEDED);
-
+        var legerConfig = context.configuration().getConfigData(LedgerConfig.class);
```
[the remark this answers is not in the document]
JivkoKelchev: Fixed

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip904/TokenAirdropTest.java
JivkoKelchev: The unit test initially was design to test both FT and NFT, since there was one common validation of the size of the transfers. It was modified to pass and to work at least with NFTs.
I added a second HAPI test for NFT airdrop here too.

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/validators/TokenAirdropValidator.java
JivkoKelchev: I agree with both of you.
I will create the following issue for our happy Thursday.

mhess-swl (approved):
LGTM! Thank you @zhpetkov!

kimbor (approved):
lgtm!

vtronkov (approved):

PR #15943, opened 2024-10-10
fix: ensure configuration loading and name resolution is resilient

### LINKED ISSUE #15921: Improve Platform SDK configuration loading and initialization mechanism resilience
### Problem

In a kubernetes environment DNS resolution and persistent/projected volumes might not be immediately available during the first few seconds of container boot up. Due to brittle nature of the Platform configuration loading and DNS resolution implementations, this causes intermittent failures when using the deterministic container image in a K8S environment. 

### Solution

- Enhance the actual Platform SDK initialization mechanisms to be more durable and resilient to running in kubernetes environment.

### Alternatives

- Not Applicable

nathanklick:
## Description

- Adds a reusable `Retry` utility class to the `swirlds-base` module.
- Adds a reusable `FileSystem` utility class to `swirlds-base` module.
- Adds a reusable `Network` utility class to the `swirlds-base` module.
- Implements a `Network.isNameResolvable` which supports retryable DNS lookups to be resilient to DNS delays.
- Implements a `FileSystem.waitForPathPresence` which supports retryable file/directory/symlink existence checks.
- Ensures the Platform/Services bootstrap is resilient to Kubernetes edge cases where DNS resolution and file mounts may be slightly delayed.

### Related Issues 

- Closes #15921

### DIFF
--- a/platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/FileSystem.java
+++ b/platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/FileSystem.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.base.utility;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Provides file system-related utility methods.
+ */
+public final class FileSystem {
+    /**
+     * Private constructor to prevent utility class instantiation.
+     */
+    private FileSystem() {}
+
+    /**
+     * Evaluates a given {@code path} to ensure it exists.
+     *
+     * <p>
+     * This method includes retry logic to handle files which may not yet exist but will become available within a
+     * given threshold. Examples of these cases are Kubernetes projected volumes and persistent volumes.
+     *
+     * <p>
+     * This overloaded method uses a default of 20 attempts with a 2-second delay between each attempt.
+     *
+     * @param path the path to check for existence.
+     * @return true if the file exists before the retries have been exhausted; otherwise false.
+     * @throws NullPointerException if the {@code path} argument is a {@code null} value.
+     */
+    public static boolean waitForPathPresence(@NonNull Path path) {
+        return waitForPathPresence(path, 20, 2_000);
+    }
+
+    /**
+     * Evaluates a given {@code path} to ensure it exists.
+     *
+     * <p>
+     * This method includes retry logic to handle files which may not yet exist but will become available within a
+     * given threshold. Examples of these cases are Kubernetes projected volumes and persistent volumes.
+     *
+     * @param path        the path to check for existence.
+     * @param maxAttempts the maximum number of retry attempts.
+     * @param delayMs     the delay between retry attempts.
+     * @return true if the file exists before the retries have been exhausted; otherwise false.
+     * @throws IllegalArgumentException if the {@code maxAttempts} argument is less than or equal to zero (0) or the
+     *                                 {@code delayMs} argument is less than zero (0).
+     * @throws NullPointerException    if the {@code path} argument is a {@code null} value.
+     */
+    public static boolean waitForPathPresence(@NonNull Path path, int maxAttempts, int delayMs) {
+        Objects.requireNonNull(path, "path must not be null");
+        try {
+            return Retry.check(FileSystem::waitForPathPresenceInternal, path, maxAttempts, delayMs);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return false;
+        } catch (ExecutionException e) {
+            return false;
+        }
+    }
+
+    /**
+     * Internal method to evaluate a given {@code path} to ensure it is present.
+     *
+     * @param path the path to check for existence.
+     * @return true if the file, directory, block device, or symlink exists and is not empty if the path references a
+     * file; otherwise false.
+     */
+    private static boolean waitForPathPresenceInternal(@NonNull final Path path) {
+        // If the path does not exist, then we should consider it not present
+        if (!Files.exists(path)) {
+            return false;
+        }
+
+        if (Files.isRegularFile(path)) {
+            try {
+                // If the path exists and is a file, then we must check the size to ensure it is not empty
+                return Files.size(path) > 0;
+            } catch (IOException ignored) {
+                // If an exception occurs while checking the file size, then we should consider the file not present
+                return false;
+            }
+        }
+
+        // If the path exists and is a directory, block device or symlink, then we can consider it present
+        return true;
+    }
+}
--- a/platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/Network.java
+++ b/platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/Network.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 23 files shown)


platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/Network.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Internal implementation used by the {@link #isNameResolvable(String, int, int)} method.
+     *
+     * @param name the domain name fragment, fully qualified domain name (FQDN), or IP address to be resolved.
+     * @return true if the name resolves to one or more valid IP addresses; otherwise false.
+     */
+    private static boolean isNameResolvableInternal(@NonNull final String name) {
+        Objects.requireNonNull(name, "name must not be null");
+        try {
+            final InetAddress[] addresses = InetAddress.getAllByName(name);
```
leninmehedy, line 88: Just wondering if `InetAddress.getByName` would suffice here instead of  `InetAddress.getAllByName(name)`

Also I can see `InetAddress.getByName()` is used in the code (in config.txt parser) to resolve to an IP, so it would make the code consistent.

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/Retry.java
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+        } catch (final Throwable e) {
+            // If an exception other than UnknownHostException or InterruptedException is thrown, then we should
+            // immediately skip any further retry attempts.
+            throw new ExecutionException("Unhandled exception thrown during a retryable operation attempt", e);
+        }
+
+        // If we failed to resolve the name to an IP address within the maximum retry threshold, then we should fall
+        // through to here to fail.
```
leninmehedy, line 85: The comment in the body (line 79-80, 84-85) needs to be polished a bit without referring to resolving an IP or UnknownHostException

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
```diff
@@ -178,11 +177,8 @@ public static Address parseAddressText(@NonNull final String addressText) throws
         }
         // FQDN Support: The original string value is preserved, whether it is an IP Address or a FQDN.
         final String internalHostname = parts[5];
-        try {
-            // validate that an InetAddress can be created from the internal hostname.
-            InetAddress.getByName(internalHostname);
-        } catch (UnknownHostException e) {
-            throw new ParseException("Cannot parse ip address from '" + parts[5] + "'", 5);
+        if (!isNameResolvable(internalHostname)) {
+            throw new ParseException("Cannot parse ip address from '" + internalHostname + "'", 5);
```
leninmehedy, line 181: nitpick: `Cannot parse` -> `Cannot resolve`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
```diff
@@ -192,11 +188,8 @@ public static Address parseAddressText(@NonNull final String addressText) throws
         }
         // FQDN Support: The original string value is preserved, whether it is an IP Address or a FQDN.
         final String externalHostname = parts[7];
-        try {
-            // validate that an InetAddress can be created from the external hostname.
-            InetAddress.getByName(externalHostname);
-        } catch (UnknownHostException e) {
-            throw new ParseException("Cannot parse ip address from '" + parts[7] + "'", 7);
+        if (!isNameResolvable(externalHostname)) {
+            throw new ParseException("Cannot parse ip address from '" + externalHostname + "'", 7);
```
leninmehedy, line 192: nitpick: `Cannot parse` -> `Cannot resolve`

leninmehedy (dismissed):
Awesome, looks good. thank you.
I've added some minor comments.

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/FileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Provides file system-related utility methods.
+ */
+public final class FileSystem {
```
lpetrovic05, line 29: I dont think `FileSystem` is an accurate name for this utility, maybe `FileSystemUtils`

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/FileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @param path        the path to check for existence.
+     * @param maxAttempts the maximum number of retry attempts.
+     * @param delayMs     the delay between retry attempts.
+     * @return true if the file exists before the retries have been exhausted; otherwise false.
+     * @throws IllegalArgumentException if the {@code maxAttempts} argument is less than or equal to zero (0) or the
+     *                                 {@code delayMs} argument is less than zero (0).
+     * @throws NullPointerException    if the {@code path} argument is a {@code null} value.
+     */
+    public static boolean waitForPathPresence(@NonNull Path path, int maxAttempts, int delayMs) {
```
lpetrovic05, line 68: Rather then using `int maxAttempts, int delayMs`, I suggested just a single parameter `Duration waitTime` with possibly a `Duration retryTime` as well, but I think this second one is not needed, we can just use a sane default. Using an `int` to represent milliseconds can be more confusing

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/FileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Internal method to evaluate a given {@code path} to ensure it is present.
+     *
+     * @param path the path to check for existence.
+     * @return true if the file, directory, block device, or symlink exists and is not empty if the path references a
+     * file; otherwise false.
+     */
+    private static boolean waitForPathPresenceInternal(@NonNull final Path path) {
```
lpetrovic05, line 87: I would remove the wait from this method name since it doesnt actually wait

platform-sdk/swirlds-base/src/main/java/com/swirlds/base/utility/Retry.java
```diff
@@ -0,0 +1,88 @@
...
+            for (int i = 0; i < maxAttempts; i++) {
+                if (checkFn.apply(value)) {
+                    return true;
+                }
+
+                if (delayMs > 0) {
+                    TimeUnit.MILLISECONDS.sleep(delayMs);
+                }
+            }
+        } catch (final Throwable e) {
```
lpetrovic05, line 78: this seems to suppress `InterruptedException`. Also, I don't see why we should catch unchecked exceptions and rethrow them as checked

platform-sdk/swirlds-config-extensions/src/main/java/com/swirlds/config/extensions/sources/PropertyFileConfigSource.java
```diff
@@ -55,6 +57,9 @@ public PropertyFileConfigSource(@NonNull final Path filePath, final int ordinal)
      */
     @NonNull
     protected BufferedReader getReader() throws IOException {
+        if (!FileSystem.waitForPathPresence(filePath)) {
+            throw new FileNotFoundException("File not found: " + filePath);
+        }
```
lpetrovic05, line 62: I feel like this wait should happen before this instance is constructed

rbarkerSL (approved):

andrewb1269hg (dismissed):

kimbor (dismissed):
I'm not clear why I was asked to review this. When I filter on 'Only files owned by you' I see only `hedera-node/infrastructure/docker/containers/production-next/consensus-node/entrypoint.sh`. I don't know why this file would be owned by the services team.

In my extremely limited knowledge of this area, it lgtm.

leninmehedy (dismissed):

kimbor (approved):

rbarkerSL (dismissed):
Lgtm

platform-sdk/swirlds-config-extensions/src/main/java/com/swirlds/config/extensions/sources/PropertyFileConfigSource.java
lpetrovic05: I still think none of these config classes should be waiting, we should do the wait before we try to initialize the config

nathanklick: If I move the logic, it results in duplicated code in multiple locations. Should we disregard being `DRY` simply to avoid the configuration logic from waiting for file existence?

nathanklick: I've gone ahead and moved the logic to the 4 known immediate callers of `PropertyFileConfigSource` not including test code locations.

Neeharika-Sompalli (approved):
LGTM from hedera-base changes! Thanks @nathanklick

lpetrovic05 (approved):

rbarkerSL (approved):

andrewb1269hg (approved):
Reviewed the following files:
`gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts`
`hedera-node/infrastructure/docker/containers/production-next/consensus-node/entrypoint.sh`
`platform-sdk/swirlds-base/build.gradle.kts`
`platform-sdk/swirlds-config-impl/build.gradle.kts`

PR #15952, opened 2024-10-10
feat: introduce NodeId.of(long)

### LINKED ISSUE #15939: Introduce `NodeId.of()` and cache values with the same id

anthony-swirldslabs:
**Description**:
Introducing a `NodeId.of(long)` static factory method for NodeId objects that uses a pre-built cache implemented in the `NodeIdCache` class which pre-constructs NodeId objects for node ids from 0 to 63. For node ids outside of this range, new NodeId objects will be created at all times.

This optimization is required to support a refactoring to replace `AddressBook` usages with `Roster` in Platform code. The `Roster` uses simple `long` values to represent node ids, and the code that uses the Roster and then needs to interact with the existing code that currently uses `NodeId` objects has to keep creating new `NodeId` objects over and over again. Having a cache helps avoid these unnecessary duplicate objects for the same, most frequently used node id values.

**Related issue(s)**:

Fixes #15939

**Notes for reviewer**:
There's no any logic changes (other than the construction of NodeId objects.) Tests should pass.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/platform/NodeIdCache.java
+++ b/platform-sdk/swirlds-common/src/main/java/com/swirlds/common/platform/NodeIdCache.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.common.platform;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * A cache for NodeId objects.
+ *
+ * It's useful to support code that uses long values to identify nodes (e.g. the Roster)
+ * and needs to interact with code that uses NodeId objects for the same purpose
+ * as it helps prevent creating many duplicate NodeId instances for the same long id value.
+ */
+final class NodeIdCache {
+    private NodeIdCache() {}
+
+    /** Minimum node id to cache. MUST BE non-negative. */
+    private static final int MIN = 0;
+    /** Maximum node id to cache. MUST BE non-negative, &gt;= MIN, and reasonably small. */
+    private static final int MAX = 63;
+
+    private static NodeId[] CACHE = new NodeId[MAX - MIN + 1];
+
+    static {
+        for (int id = MIN; id <= MAX; id++) {
+            CACHE[id - MIN] = new NodeId(id);
+        }
+    }
+
+    /**
+     * Fetch a NodeId value from the cache, or create a NodeId.of object.
+     * The caller MUST NOT mutate the returned object even though the NodeId class is technically mutable.
+     *
+     * @param id a node id value
+     * @return a NodeId object
+     */
+    @NonNull
+    static NodeId getOrCreate(final long id) {
+        if (id >= MIN && id <= MAX) {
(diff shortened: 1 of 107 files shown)


platform-sdk/swirlds-common/src/main/java/com/swirlds/common/platform/NodeIdCache.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+final class NodeIdCache {
+    private NodeIdCache() {}
+
+    /** Minimum node id to cache. MUST BE non-negative. */
+    private static final int MIN = 0;
+    /** Maximum node id to cache. MUST BE non-negative, &gt;= MIN, and reasonably small. */
+    private static final int MAX = 63;
+
+    private static NodeId[] CACHE = new NodeId[MAX - MIN + 1];
```
kfa-aguda, line 36: nit: final qualifier

anthony-swirldslabs: Nice catch! Done.

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/platform/NodeIdCache.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param id a node id value
+     * @return a NodeId object
+     */
+    @NonNull
+    static NodeId getOrCreate(final long id) {
+        if (id >= MIN && id <= MAX) {
+            return CACHE[(int) id - MIN];
+        }
+        return new NodeId(id);
```
kfa-aguda, line 56: This class seem to define a boundary as invariant but doesn't honor it. I suspect that's done to allow tests to arbitrarily create IDs larger than the boundary? If so, I don't think we should allow this. The ID boundary should either be present and enforced, or not present at all. Did you encounter a reason why we must allow larger node ids?

anthony-swirldslabs: The boundary isn't really defined as a part of the specification. The only public specification for this behavior is defined by the `NodeId.of()` which clearly states that it returns a **potentially** cached object, but doesn't guarantee that.

The idea of a cache with a finite number of pre-built items was suggested by @rbair23 and follows the patter established by a well known `java.lang.Integer.valueOf()` behavior and specifically, the JDK internal `Integer.IntegerCache` implementation which implements roughly the same approach.

The idea behind 63 (64, really, when counted as 63-0+1) is to keep the cache reasonably small and cover all the existing mainnet nodes, and have a little room for immediate growth.

There's indeed tests that use larger node id values. Additionally, there's testing tools (hapi, JRS, Solo, etc.) which allow users to create arbitrarily large networks. For this reason, we cannot really "enforce" the boundary. At the same time, we don't want the complexity of a dynamic cache with a Map and dynamic object construction as that would require extra synchronization and therefore would defeat the purpose of a "performance optimization" that this fix is trying to achieve.

Hence the cache for 0..63 with new objects for any other ids.

kfa-aguda: Yeah I'm not questioning the choice of the cache implementation (saw the thread yesterday). I just wasn't super sure why we'd allow uncached nodes with arbitrary large node IDs but looks like my suspiction of it being a test requirement is correct, in which case I accept this is ok. Do you mind adding a note in the javadoc for it?

anthony-swirldslabs: As indicated in the first paragraph, the javadoc for `NodeId.of` already mentions it. I don't think it would be correct to add references to constants defined in a package-private class to a javadoc of a public method though, that's the reason why that mentions goes like "potentially" w/o any further details. If you have a specific suggestion, could you please post it as a suggested change here in this PR? I'd be happy to check it out and apply it if we both agree on the wording.

anthony-swirldslabs: Done.

kfa-aguda (approved):

Neeharika-Sompalli (approved):
LGTM from `hedera-base` changes. Thanks @anthony-swirldslabs

Issue #15955, opened 2024-10-10
Ensure prepare-xts branch launches when node-deploy-production-build finishes


rbarkerSL:
The `zxf-prepare-extended-test-suite.yaml` file is currently skipping the tag-for-xts job. Need to determine why this job is not triggering appropriately when `"ZXF: Deploy Production Release"` completes with a status of success

PR #15967, opened 2024-10-11
ci: Updated XTS job to check statuses and added gpg key to prepare XTS

rbarkerSL:
**Details**:
    
Adds needs checks for `zxcron-extended-test-suite.yaml`
`tag-for-promotion` step.

Adds import GPG step to `zxf-prepare-extended-test-suite.yaml`

**Related Issue(s)**:

Relates to #15955


andrewb1269hg (approved):
Looks good to me. Do we need to write an issue for the TODO added in this PR?

PR #15969, opened 2024-10-11
fix: Ensure `getAccountInfo` returns correct EVM address

### LINKED ISSUE #15954: Fix ContractAccountId field in AccountInfoQuery
### Description

GetAccountInfo query is returning the key-derived EVM address for every account with an ECDSA key, even if it does not have an alias.

- This differs from the pre-0.49 mono code, which only returned a key-derived EVM address for accounts with an alias.

But the actual state of the ledger is:

- An account 0.0.X created with ECDSA key but no alias still has only long-zero 0x00...0X EVM address.
- The record stream item contains this information, so mirror node's evm_address field is correct.
https://swirldslabs.slack.com/archives/C06UM5GEY9Z/p1720630768839249?thread_ts=1719960535.199049&cid=C06UM5GEY9Z

The proposed solution based on precedence from the field and clarification from [HIP 583](https://hips.hedera.com/hip/hip-583) my recommendation would be

- AccountInfoQuery should return the following format for ContractAccountId
- ECDSA account w/o alias => Hedera Account Num Alias (aka long-zero) format e.g. [0x000000000000000000000000000000000044ed9f](https://hashscan.io/testnet/account/0.0.4517279?pt=1&ph=1&pn=1&pc=1&p1=1)
- ECDSA account w alias => EVM Address Alias format e.g. [0x04a317f16ab69349738f6351e07581bc5c2506d1](https://hashscan.io/testnet/account/0.0.121917?app=false&p2=1&pn=1&pt=1&ph=1&pc=1)
- ED account w or w/o alias=> Hedera Account Num Alias format e.g. [0x0000000000000000000000000000000000000425](https://hashscan.io/testnet/account/0.0.1061?p1=1&k1=1719498470.257510003&pt=1&ph=1&pn=1&pc=1)
https://swirldslabs.slack.com/archives/C06UM5GEY9Z/p1720646289865449?thread_ts=1719960535.199049&cid=C06UM5GEY9Z

### Steps to reproduce

See description in slack thread..

### Additional context

See all details in this thread:
https://swirldslabs.slack.com/archives/C06UM5GEY9Z/p1719960535199049

### Hedera network

mainnet

### Version

Post-Modularization releases

### Operating system

None

tinker-michaelj:
**Description**:
 - Fixes #15954 
 - Adds `CryptoUpdateSuite#keyRotationDoesNotChangeEvmAddress()` HAPI test to validate both exported records and query responses after key rotation on accounts created with vanilla `CryptoCreate`, HIP-32 auto-creation, hollow account creation, and HIP-583 explicit EVM address aliasing.


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoUpdateSuite.java
```diff
@@ -132,6 +163,122 @@ final Stream<DynamicTest> idVariantsTreatedAsExpected() {
                         .newStakedAccountId("0.0.21")));
     }
 
+    private static final String[] ACCOUNTS_TO_HAVE_KEYS_ROTATED = {
+        "longZero", "autoCreated", "hollowAccount", "explicitAlias"
+    };
+    private static final UnaryOperator<String> ROTATION_TXN = account -> account + "KeyRotation";
```
david-bakin-sl, line 169: In a future refactor: This would make a fine HAPISpec utility: The four different ways of creating an account w/ EC key, assembled together and put in the registry.  I bet it could be used in multiple other tests to easily provide completeness.

david-bakin-sl (approved):
I understand what's happening in `AccountSummariesApi.hexedEvmAddressOf` - follows the issue description.

HAPISPec additions should prove valuable in the future.

Also, Javadoc format sux.

Neeharika-Sompalli (approved):
LGTM Thanks @tinker-michaelj !

Neeharika-Sompalli (approved):
LGTM!

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/CryptoUpdateSuite.java
tinker-michaelj: Ah yes, makes sense. 💯

PR #15970, opened 2024-10-11
ci: Add message parameter to forced tag step

rbarkerSL:
**Description**:

Adds the `--messsage` parameter to the tag step in prepare XTS flow

**Related issue(s)**:

Relates to 15955


andrewb1269hg (approved):
Looks good.

PR #15986, opened 2024-10-15
refactor: extract IntakeEventCounter and FallenBehindManager to gossip module

mustafauzunn:
**Description**:
 
Extraction of IntakeEventCounter and FallenBehindManager interfaces to gossip module.
Extraction of DefaultIntakeEventCounter to gossip-impl module.
Move Address, AddressBook, AddressBookIterator and SerializableX509Certificate to swirlds-common module.

**Related issue(s)**:

Fixes 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-addressbook-service-impl/build.gradle.kts
```diff
@@ -36,5 +36,6 @@ testModuleInfo {
     requires("org.junit.jupiter.api")
     requires("org.mockito")
     requires("org.mockito.junit.jupiter")
+    requires("com.swirlds.common")
```
rbarkerSL, line 39: Is this for a new publish group? I didn't think we were publishing anything new to com.swirlds.*

platform-sdk/consensus-gossip-impl/src/main/java/module-info.java
```diff
@@ -1,5 +1,9 @@
 module com.hedera.consensus.gossip.impl {
+    exports com.hedera.service.gossip.impl;
+
+    requires transitive com.swirlds.common;
```
rbarkerSL, line 4: Is this for a new publish group? I didn't think we were publishing anything new to com.swirlds.*

platform-sdk/consensus-gossip/src/main/java/module-info.java
```diff
@@ -1,3 +1,6 @@
 module com.hedera.service.gossip {
     exports com.hedera.service.gossip;
+
+    requires transitive com.swirlds.common;
```
rbarkerSL, line 4: Is this for a new publish group? I didn't think we were publishing anything new to com.swirlds.*

rbarkerSL (changes requested):

platform-sdk/swirlds-common/src/main/java/com/swirlds/common/AddressBook.java
```diff
@@ -603,7 +603,30 @@ public Set<NodeId> getNodeIdSet() {
      */
     @NonNull
     public String toConfigText() {
-        return AddressBookUtils.addressBookConfigText(this);
+        return addressBookConfigText(this);
+    }
+
+    public static String addressBookConfigText(@NonNull final AddressBook addressBook) {
```
litt3, line 609: Is the idea that this should be replaced with a call to `AddressBookUtils` after that file has also been moved? If so, this method should probably be private, and have a FUTURE WORK note explaining the future intention

litt3 (dismissed):
lgtm

rbarkerSL (approved):
Review applies to

- hedera-node/hedera-addressbook-service-impl/build.gradle.kts
- platform-sdk/consensus-gossip-impl/src/main/java/module-info.java
- platform-sdk/consensus-gossip/src/main/java/module-info.java
- platform-sdk/swirlds-platform-core/src/main/java/module-info.java
- platform-sdk/swirlds-platform-core/src/testFixtures/java/module-info.java
- platform-sdk/swirlds-unit-tests/core/swirlds-platform-test/src/main/java/module-info.java

LGTM

jeromy-cannon (approved):

poulok (changes requested):
I have a couple questions/concerns:

1. What issue is this fixing?
2. Why is this moving the AddressBook class that is currently being removed from the platform?
3. I think this will conflict with a large refactor PR from @anthony-swirldslabs that is critical to the TSS Roster project. Is it necessary that this is done now?

poulok:
>This PR is part of the platform modularization effort

Ah, understood. We usually keep a 1:1 ratio of PRs to tickets, but if the decision is to depart from that approach, please at least reference the ticket this PR supports in the description to provide more context to reviewers.

>Moved Address, AddressBook, AddressBookIterator and SerializableX509Certificate to the swirlds-common module because they are currently in the platform-core module.
DefaultIntakeEventCounter we move to new module that needs AddressBook for example and if we don't move AddressBook to some other module we will have circular dependencies. We want small PRs and this is the approach I am using

Makes sense, except for the point below.

>I think AddressBook is used in many places currently and I think we can't modularise the codebase without these classes.

There is [this open PR](https://github.com/hashgraph/hedera-services/pull/16026) that goes a very long way to eliminating AddressBook from the platform. More PRs will follow to cleanup the remaining references. We must deconflict these PRs. In my opinion, these other PRs are higher priority as they are required to get TSS Rosters delivered, which unblocks the TSS Base Service integration testing work. All of this is required to deliver Block Nodes.

IvanKavaldzhiev:
Closing since this PR is now obsolete.

Issue #15994, opened 2024-10-15
Need more logging in AbstractHashListener


artemananiev:
It would be great to log flush durations and the number of hashes/leaves flushed in `AbstractHashListener`

PR #15995, opened 2024-10-16
fix: 15994: Need more logging in AbstractHashListener

artemananiev:
Fix summary:
* Two log statements are added to `AbstractHashListener`: to log flush time and to log leaves/hashes counters
* Flush duration is also reported as a virtual map metric

Fixes: https://github.com/hashgraph/hedera-services/issues/15994


alex-kuzmin-hg (approved):
Looks reasonable

anthony-swirldslabs (approved):

PR #15435, opened 2024-09-11
docs: Provide Design document for HIP-904 System Contracts flows

stoyanov-st:
**Description**:

This PR proposes a design document for the upcoming implementation of the HIP-904 system contracts functions and logic.

**Related issue(s)**:

[932](https://github.com/hashgraph/hedera-smart-contracts/issues/932)
Fixes #

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #15996, opened 2024-10-16
feat: HIP-904 Implement Token Cancel Airdrop System Contract

### LINKED ISSUE #15589: HIP-904: Token Cancel Airdrop System Contract
Implement the Token Cancel Airdrop System Contract described in https://github.com/hashgraph/hedera-services/pull/15435

- [x] HSCS System Contract Classes
- [x] Feature flag
- [x] E2E tests

stoyanov-st:
**Description**:

This PR introduces the implementation of `TokenCancelAirdrop` System contact described in https://github.com/hashgraph/hedera-services/pull/15435/files?short_path=03ab1e7#diff-03ab1e73f0ded9a04739c3b1a68c13f779dce416cd2f7b6637b03d8e2e41e105

**Related issue(s)**:

Fixes #15589 

**Notes for reviewer**:
This feature is only enabled for dev and previewnet envs

**Checklist**


lukelee-sl (approved):
Thanks for doing this @stoyanov-st.  Looks good to me.  I have no comments.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+        return TransactionBody.newBuilder()
+                .tokenCancelAirdrop(
+                        TokenCancelAirdropTransactionBody.newBuilder().pendingAirdrops(pendingAirdrops))
+                .build();
+    }
+
+    public TransactionBody decodeCancelAirdropFT(@NonNull final HtsCallAttempt attempt) {
+        final var call = TokenCancelAirdropTranslator.HRC_CANCEL_AIRDROP_FT.decodeCall(attempt.inputBytes());
+
+        final var senderAddress = attempt.senderId();
```
david-bakin-sl, line 93: `senderAddress` -> `senderId`?  (see `receiverId` below which is also an `AccountID`). Similarly `token` -> `tokenId`?

(It's hard enough to keep track of _having_ addresses, accounts, and entity ids in the same place even without misleading names ...)

(Also below at `decodeCancelAirdropNFT`)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.HtsCallAttempt;
+import com.hedera.node.config.data.TokensConfig;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.Arrays;
+import javax.inject.Inject;
+
+public class TokenCancelAirdropDecoder {
+
+    // Tuple indexes
```
david-bakin-sl, line 43: Consider additional comments to break these indexes into sections - one for each method - or some other annotation here.

(This is an area where the headlong library could be improved: There should be a way to tell it what the names of arguments are so you didn't have to keep track of individual argument indexes yourself; it's errorprone.)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Inject
+    public TokenCancelAirdropDecoder() {
+        // Dagger2
+    }
+
+    public TransactionBody decodeCancelAirdrop(@NonNull final HtsCallAttempt attempt) {
+        final var call = TokenCancelAirdropTranslator.CANCEL_AIRDROP.decodeCall(attempt.inputBytes());
+        final var maxPendingAirdropsToCancel =
+                attempt.configuration().getConfigData(TokensConfig.class).maxAllowedPendingAirdropsToCancel();
```
david-bakin-sl, line 60: Consider moving above L58 or maybe to a separate (tiny!) method.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+        validateFalse(((Tuple[]) call.get(0)).length > maxPendingAirdropsToCancel, PENDING_AIRDROP_ID_LIST_TOO_LONG);
+
+        final var transferList = (Tuple[]) call.get(TRANSFER_LIST);
```
david-bakin-sl, line 63: If L63 was above L61 you wouldn't have to do the `(cast)call.get()` twice ... (and note too that at L61 you didn't use the named constant ...)

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+        final var pendingAirdrops = new ArrayList<PendingAirdropId>();
+        Arrays.stream(transferList).forEach(transfer -> {
```
david-bakin-sl, line 65: I like the use of the stream at L65 below ... so ... you could _collect_ into a list instead of adding to it (via map + collect rather than foreach) ... it's a matter of taste how far into streams you want to go but actually there's no advantage to `Arrays.stream(...).forEach(...)` than the basic for loop if you're just adding to the result list ...

david-bakin-sl (commented):
(Have to interrupt this review to go to the dentist.  Did all the production code.  Will come back for all the unit tests/HapiSpec tests.  But looking good so far: Just a bunch of style comments which you can evaluate then take or leave.)

lukelee-sl (dismissed):
LGTM

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
```diff
@@ -0,0 +1,327 @@
...
+        lenient().when(attempt.addressIdConverter()).thenReturn(addressIdConverter);
+    }
+
+    @Test
+    void cancelAirdropDecoder1FTTest() {
+        // given:
+        given(attempt.configuration()).willReturn(configuration);
+        given(attempt.enhancement()).willReturn(enhancement);
+        given(enhancement.nativeOperations()).willReturn(nativeOperations);
+        given(configuration.getConfigData(TokensConfig.class)).willReturn(tokensConfig);
```
david-bakin-sl, line 94: These look like candidates to move to `setup()` (probably with `lenient()`)

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
```diff
@@ -0,0 +1,327 @@
...
+        given(attempt.configuration()).willReturn(configuration);
+        given(configuration.getConfigData(TokensConfig.class)).willReturn(tokensConfig);
+        given(tokensConfig.maxAllowedPendingAirdropsToCancel()).willReturn(10);
+
+        final var encoded = Bytes.wrapByteBuffer(CANCEL_AIRDROP.encodeCall(Tuple.singleton(new Tuple[] {
+            Tuple.of(
+                    asHeadlongAddress(SENDER_ID.accountNum()),
+                    OWNER_ACCOUNT_AS_ADDRESS,
+                    FUNGIBLE_TOKEN_HEADLONG_ADDRESS,
+                    0L),
```
david-bakin-sl, line 136: Could, if you wanted, just put this in a local and then put that same instance into the array 11 times.  Nobody is deduping the array ....

Could also reduce the limit below 10 ...

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
```diff
@@ -0,0 +1,327 @@
...
+            Tuple.of(
+                    asHeadlongAddress(SENDER_ID.accountNum()),
+                    OWNER_ACCOUNT_AS_ADDRESS,
+                    FUNGIBLE_TOKEN_HEADLONG_ADDRESS,
+                    0L)
+        })));
+        given(attempt.inputBytes()).willReturn(encoded.toArrayUnsafe());
+
+        assertThrows(HandleException.class, () -> subject.decodeCancelAirdrop(attempt));
```
david-bakin-sl, line 190: Is it valuable to check the code is `PENDING_AIRDROP_ID_LIST_TOO_LONG`, or unnecessary?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
```diff
@@ -0,0 +1,327 @@
...
+                    OWNER_ACCOUNT_AS_ADDRESS,
+                    FUNGIBLE_TOKEN_HEADLONG_ADDRESS,
+                    0L)
+        })));
+        given(attempt.inputBytes()).willReturn(encoded.toArrayUnsafe());
+
+        assertThrows(
+                HandleException.class,
+                () -> subject.decodeCancelAirdrop(attempt),
+                PENDING_AIRDROP_ID_LIST_TOO_LONG.protoName());
```
david-bakin-sl, line 218: Should this be `PENDING_AIRDROP_ID_LIST_TOO_LONG` -> `INVALID_TOKEN_ID` and if so why doesn't this test fail?

And the answer why it doesn't fail is:  the `message` argument to `assertThrows` is _not_ a matcher against the message of the exception, instead it is the message for Junit to report for the test failure.  (I didn't realize this until I debugged into it.)

Suggest using instead AssertJ's `assertThat(thrown).isInstanceOf(...).hasMessage(...)` or, to match this style here (which I prefer)

```
assertThatExceptionOfType(HandleException.class)
   .isThrownBy(() -> ...some lambda body here...)
   .withMessage("...");
```

Same problem appears in following tests.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslator.java
```diff
@@ -0,0 +1,81 @@
...
+import com.hedera.node.app.service.contract.impl.exec.systemcontracts.hts.HtsCallAttempt;
+import com.hedera.node.app.service.contract.impl.hevm.HederaWorldUpdater;
+import com.hedera.node.config.data.ContractsConfig;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import javax.inject.Inject;
+
+public class TokenCancelAirdropTranslator extends AbstractCallTranslator<HtsCallAttempt> {
+
+    public static final Function CANCEL_AIRDROP =
+            new Function("cancelAirdrops((address,address,address,int64)[])", "(int32)");
```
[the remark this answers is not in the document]
stoyanov-st: Added a comment to specify the PendingAirdorp struct

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslator.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public boolean matches(@NonNull final HtsCallAttempt attempt) {
+        final var cancelAirdropEnabled =
+                attempt.configuration().getConfigData(ContractsConfig.class).systemContractCancelAirdropsEnabled();
+        return attempt.isTokenRedirect()
+                ? attempt.isSelectorIfConfigEnabled(HRC_CANCEL_AIRDROP_FT, cancelAirdropEnabled)
+                        || attempt.isSelectorIfConfigEnabled(HRC_CANCEL_AIRDROP_NFT, cancelAirdropEnabled)
+                : attempt.isSelectorIfConfigEnabled(CANCEL_AIRDROP, cancelAirdropEnabled);
```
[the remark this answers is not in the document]
stoyanov-st: I was planning to refactor that method with a separate PR but will include it here

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslator.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public Call callFrom(@NonNull final HtsCallAttempt attempt) {
+        return new DispatchForResponseCodeHtsCall(
+                attempt,
+                attempt.isSelector(CANCEL_AIRDROP) ? decoder.decodeCancelAirdrop(attempt) : bodyForHRC(attempt),
+                TokenCancelAirdropTranslator::gasRequirement);
+    }
+
+    private TransactionBody bodyForHRC(@NonNull final HtsCallAttempt attempt) {
+        return attempt.isSelector(HRC_CANCEL_AIRDROP_FT)
```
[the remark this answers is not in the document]
stoyanov-st: Moved those to a single method. Seems more like a styling preference than readability imo.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
stoyanov-st: Thanks. I missed to rename those.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
stoyanov-st: Added comments with function and exact signature

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
stoyanov-st: Must've left it like that when I was including the fail fast mechanisms

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoder.java
stoyanov-st: Yes, thanks. Reworked to map and collect

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    @Test
+    void failsIfTokenIsNullHRCFT() {
+        final var encoded = Bytes.wrapByteBuffer(HRC_CANCEL_AIRDROP_FT.encodeCallWithArgs(OWNER_ACCOUNT_AS_ADDRESS));
+        given(attempt.inputBytes()).willReturn(encoded.toArrayUnsafe());
```
david-bakin-sl, line 224: I think this test isn't testing what you think it is testing, IMO.  Looking at `HtsCallAttempt.redirectTokenId` we see that it is supposed to fail (throw exception) if the call attempt is _not_ a redirect.  It is _not_ a redirect if the redirect address is null.  Our `attempt` here is a mock, there are no givens for `isRedirect` or `redirectAddress`, so you'd think it would be null, thus throw an IllegalStateException, but it doesn't.  It continues (and eventually throws a `HandleException(INVALID_TOKEN_ID)`.  But why did it continue, i.e., why did it think it _was_ a redirect case?  Something appears fishy here.  (Whatever the resolution is for this test applies to the next test too.)

david-bakin-sl (commented):
Working through tests now and found some issues that will need correction.  Still going through it....

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslatorTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+    private TokenCancelAirdropTranslator subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new TokenCancelAirdropTranslator(decoder);
+    }
+
+    @Test
+    void matchesHTSCancelAirdropEnabled() {
```
david-bakin-sl, line 97: How about at least one test for an non-matching selector, when enabled?  (Obviously can't be comprehensive, just one mismatch will do.)

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslatorTest.java
```diff
@@ -0,0 +1,301 @@
...
+                addressIdConverter,
+                verificationStrategies,
+                gasCalculator,
+                configuration);
+
+        // when:
+        var call = subject.callFrom(attempt);
+
+        // then:
+        assertEquals(DispatchForResponseCodeHtsCall.class, call.getClass());
```
david-bakin-sl, line 255: Only verification is that you get back the right class ... seems insufficient given `TokenCancelAirdropDecoder` L61-71 has that choice of calling three different decoder methods which return three different (though similar!) things.  _That's_ what this unit test should be testing: That the right decoder was called.

In fact since the decoder used here is a mock I'm not even sure _what's_ being called.  I mean, `callFrom` is calling into the decoder: what's happening when those calls are made to the mock?  I thought mockito complained when you called a method you didn't supply a given for (to just call the real method I thought you must use `thenCallRealMethod()` - but maybe not.

Anyway, maybe these tests can use `verify` to check that at least the right decoder was called?

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
stoyanov-st: Even if I mock `isRedirect` it still skips it? Maybe I'll remove the tests

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslatorTest.java
stoyanov-st: Added verify checks

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropTranslatorTest.java
stoyanov-st: Added

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
stoyanov-st: Adjusted

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
stoyanov-st: Done

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/hts/cancelairdrops/TokenCancelAirdropDecoderTest.java
stoyanov-st: Yes, added check

david-bakin-sl (approved):
LGTM! (And TY for considering those change suggestions!)

PR #15997, opened 2024-10-16
fix: adding support for chain ids where the value requires a sign bit…

### LINKED ISSUE #15953: WRONG CHAIN_ID error when executing pre-EIP-1559 transactions.
### Description

When submitting a pre-EIP-1559 Ethereum transaction on the Hedera network, the `chain_id` is incorrectly converted. Specifically, an additional `0` is added at the beginning of the `chain_id`, causing a mismatch during the chain ID validation.

For example:
- Chain ID `11155111` is incorrectly converted to `00aa36a7` (with an extra leading `0`) for pre-EIP-1559 transactions.
- For post-EIP-1559 transactions, the same chain ID is correctly converted to `aa36a7`.

This issue causes the [matchesChainId](https://github.com/hashgraph/hedera-services/blob/43df250316934b7b129f022259eb293d2dc5e823/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java#L342) method in Hedera to fail validation, leading to a `WRONG CHAIN_ID` error when executing pre-EIP-1559 transactions.

Example pre-EIP-1559 Ethereum transaction:  
[Sepolia Etherscan Example](https://sepolia.etherscan.io/tx/0x1009199f804a4c75d266627548dc154989eec8385c7b3a67e2bd50a5db5063c9)
Example post-EIP-1559 Ethereum transaction:  
[Sepolia Etherscan Example](https://sepolia.etherscan.io/tx/0x1009199f804a4c75d266627548dc154989eec8385c7b3a67e2bd50a5db5063c9)

When comparing the two chain IDs as byte arrays:
- Pre-EIP-1559: `[0, -86, 54, -89]`
- Post-EIP-1559: `[-86, 54, -89]`

This causes the following validation in Hedera to return false:
```java
public boolean matchesChainId(final byte[] hederaChainId) {
    return chainId == null || chainId.length == 0 || Arrays.compare(chainId, hederaChainId) == 0;
}
```

### Steps to reproduce

1. Submit a pre-EIP-1559 Ethereum transaction on the Hedera network.
2. Attempt to execute the transaction.
3. The matchesChainId method will fail to validate the chain ID, resulting in a WRONG CHAIN_ID error.

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet, other

### Version

v0.1.0

### Operating system

None

arianejasuwienas:
… to be stored in a separate byte within the byte array returned by the biginteger type (#15953)

**Description**:
When using a `CHAIN_ID` represented as a BigInteger, an additional byte is required to store the sign information (indicating whether the value is positive or negative), if there is no free bit available for this information, as in values like 11155111.

As a result, the previous code caused BigInteger.toByteArray to return a value with an unexpected extra byte, leading to validation failures and other issues. For example, the message payload contained this extra byte,  which caused attempts to derive the signer from the message and signature to return an incorrect signer ID instead of the correct one.

**Related issue(s)**:

Fixes #15953 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
+++ b/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
@@ -387,7 +387,7 @@ public record EthTxData(
             // after EIP155 the chain id is equal to
             // CHAIN_ID = (v - {0,1} - 35) / 2
             chainId = BigIntegers.asUnsignedByteArray(
-                vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
+                    vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
         } else if (isLegacyUnprotectedEtx(vBI)) {
             // before EIP155 the chain id is considered equal to 0
             chainId = new byte[0];
--- a/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
+++ b/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
@@ -31,6 +31,7 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import org.apache.commons.codec.binary.Hex;
 import org.bouncycastle.jcajce.provider.digest.Keccak;
+import org.bouncycastle.util.BigIntegers;
 
 public record EthTxData(
         byte[] rawTx,
@@ -385,7 +386,7 @@ public record EthTxData(
         if (vBI.compareTo(BigInteger.valueOf(34)) > 0) {
             // after EIP155 the chain id is equal to
             // CHAIN_ID = (v - {0,1} - 35) / 2
-            chainId = vBI.subtract(BigInteger.valueOf(35)).shiftRight(1).toByteArray();
+            chainId = BigIntegers.asUnsignedByteArray(vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
         } else if (isLegacyUnprotectedEtx(vBI)) {
             // before EIP155 the chain id is considered equal to 0
             chainId = new byte[0];
--- a/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
+++ b/hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
@@ -386,7 +386,8 @@ public record EthTxData(
         if (vBI.compareTo(BigInteger.valueOf(34)) > 0) {
             // after EIP155 the chain id is equal to
             // CHAIN_ID = (v - {0,1} - 35) / 2
-            chainId = BigIntegers.asUnsignedByteArray(vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
+            chainId = BigIntegers.asUnsignedByteArray(
+                vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
         } else if (isLegacyUnprotectedEtx(vBI)) {
             // before EIP155 the chain id is considered equal to 0
             chainId = new byte[0];
--- a/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
+++ b/hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
@@ -35,6 +35,7 @@ import com.swirlds.common.utility.CommonUtils;
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
+import org.bouncycastle.util.BigIntegers;
 import org.bouncycastle.util.encoders.Hex;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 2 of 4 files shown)


arianejasuwienas:
Failing jobs are unrelated to the changes itself...
Merging this PR will fix them:
https://github.com/hashgraph/hedera-services/pull/16006

arianejasuwienas:
Replaced with:

https://github.com/hashgraph/hedera-services/pull/16034

I no longer have the access right to write in this repository.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -601,4 +604,18 @@ void bigPositiveValueWithDifferentTypes(EthTransactionType type) {
 
         assertEquals(bigValue, populateEthTxData.value());
     }
+
+    @Test
+    void thatPassesOnBigIntegerByteArray() {
```
Nana-EC, line 609: nit: function name should be more descriptive to test case
```suggestion
    void populateEthTxDataWithLargeSignedChanId() {
```

Nana-EC (commented):
Nice find, minor readability suggestion

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/leaky/LeakyEthereumTestsSuite.java
```diff
@@ -91,6 +91,46 @@ final Stream<DynamicTest> legacyUnprotectedEtxBeforeEIP155() {
...
 
+    /**
+     * test unprotected legacy ethereum transactions before EIP155
+     * When using a `CHAIN_ID` represented as a BigInteger, an additional byte is required
+     * to store the sign information (indicating whether the value is positive or negative),
+     * if there is no free bit available for this information, as in values like 11155111.
+     */
+    @LeakyHapiTest(overrides = {"contracts.chainId"})
+    /* default */ final Stream<DynamicTest> legacyUnprotectedEtxBeforeEIP155WithChainIdHavingExtraByteForSign() {
+        final String DEPOSIT = "deposit";
```
lukelee-sl, line 102: nit: it is better to use `var` if the type can be inferred.

lukelee-sl (approved):
LGTM.  Small nits if you choose to fix.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
```diff
@@ -601,4 +604,18 @@ void bigPositiveValueWithDifferentTypes(EthTransactionType type) {
...
+        final var subject = EthTxData.populateEthTxData(Hex.decode(RAW_TX_TYPE_0_WITH_CHAIN_ID_11155111));
+        byte[] passingChainId = BigIntegers.asUnsignedByteArray(BigInteger.valueOf(11155111L));
+        assertEquals(Hex.toHexString(subject.chainId()), Hex.toHexString(passingChainId));
+    }
+
+    @Test
+    void thatFailsOnBigIntegerByteArray() {
```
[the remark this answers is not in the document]
david-bakin-sl: except `ChainId` please ...

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
```diff
@@ -385,7 +386,8 @@ private static EthTxData populateLegacyEthTxData(RLPItem rlpItem, byte[] rawTx)
         if (vBI.compareTo(BigInteger.valueOf(34)) > 0) {
             // after EIP155 the chain id is equal to
             // CHAIN_ID = (v - {0,1} - 35) / 2
-            chainId = vBI.subtract(BigInteger.valueOf(35)).shiftRight(1).toByteArray();
+            chainId = BigIntegers.asUnsignedByteArray(
+                    vBI.subtract(BigInteger.valueOf(35)).shiftRight(1));
```
david-bakin-sl, line 390: Consider leaving a comment to explain the reason here is that signed/unsigned high order byte thing of `BigInteger`.  It won't be obvious to someone coming across this later.  A pointer to the issue #15953 would be fine.

david-bakin-sl (approved):
LGTM

david-bakin-sl:
P.S. This is the real PR now - the _other_ one is closed.

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
arianejasuwienas: Finding a better name for this test is a great idea! However, I don’t want to focus solely on the id size, as that's not the only issue. The problem could occur even with smaller numbers. A more detailed explanation can be found here:

https://github.com/hashgraph/hedera-services/issues/15998

What do you think about names:
```
populateEthTxDataComparedToSignedByteArrayExtraByteAdded
populateEthTxDataComparedToUnsignedByteArrayNoExtraByteAdded
```
for these tests?

@david-bakin-sl 
@Nana-EC

hedera-node/hapi-utils/src/test/java/com/hedera/node/app/hapi/utils/ethereum/EthTxDataTest.java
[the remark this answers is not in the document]
arianejasuwienas: https://github.com/hashgraph/hedera-services/pull/15997#discussion_r1821395388

hedera-node/hapi-utils/src/main/java/com/hedera/node/app/hapi/utils/ethereum/EthTxData.java
arianejasuwienas: @david-bakin-sl Will this description be sufficient? https://github.com/hashgraph/hedera-services/pull/15997/files#diff-ce266e5cc986c3cb7c3505dd7203a63ffacfe9df6b36cac3f80dca625380ab56R389

david-bakin-sl: Yes! Except at L389 I think you meant `The BigIntegers library` with an s ....

david-bakin-sl (approved):
LGTM

lukelee-sl (approved):
LGTM

kimbor (approved):

PR #16006, opened 2024-10-16
fix: stabilize `keyRotationDoesNotChangeEvmAddress()`

### LINKED ISSUE #16005: Test `keyRotationDoesNotChangeEvmAddress` is flaky
### Description

The record stream validation in this test is timing out intermittently.

### Steps to reproduce

Run repeatedly.

### Additional context

_No response_

### Hedera network

_No response_

### Version

multiple

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #16005
 - Ensures `HapiGetTxnRecord` does not expose a synthetic system account creation record as a creation "owned" by a transaction.
 - Adds javadoc to `AssertionResult` and `EventualAssertionResult`.
 - Includes more failure information in the message of the `AssertionError` thrown by a `AbstractEventualStreamAssertion`.
 - Increases the allowed timout for the flaky test to 20s.


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/AbstractEventualStreamAssertion.java
```diff
@@ -82,11 +82,18 @@ public void assertHasPassed() {
             final var eventualResult = result.get();
             unsubscribe();
             if (!eventualResult.passed()) {
-                Assertions.fail(eventualResult.getErrorDetails());
+                Assertions.fail(assertionDescription() + " ended with result: " + eventualResult.getErrorDetails());
             }
         } catch (final InterruptedException e) {
+            unsubscribe();
```
david-bakin-sl, line 88: Given L83 exists too - does it matter exactly _when_ `unsubscribe` is called?  If not, could be done in a `finally`.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/EventualAssertionResult.java
```diff
@@ -24,36 +24,72 @@
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
+/**
+ * Represents the possibly pending result of an assertion. The result will be pending for at most the given timeout
+ * duration, after which the result will be completed as either a success or a failure based on the value of the
+ * {@code hasPassedIfNothingFailed} flag.
```
david-bakin-sl, line 30: `hasPassedIfNothingFailed` -> `passAfterTimeout`?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/EventualAssertionResult.java
```diff
@@ -24,36 +24,72 @@
...
+     */
     public AssertionResult get() throws InterruptedException {
         if (!ready.await(timeout.toMillis(), TimeUnit.MILLISECONDS)) {
-            if (hasPassedIfNothingFailed && result == null) {
-                return AssertionResult.success();
+            if (passAfterTimeout && result == null) {
+                return AssertionResult.newSuccess();
             } else {
-                return AssertionResult.timeout(timeout);
+                return AssertionResult.newTimeout(timeout);
```
david-bakin-sl, line 73: you can get here even if `result != null`, that doesn't seem right.   (There's a race where `pass()` could be called after the timeout before you get here; I think it is possible.)

People disparage `synchronized` methods and blocks as if they're _always_ bad and _always_ impact performance and affect concurrency, but on a very-low rate non-real-time critical class like this, why not use it?

david-bakin-sl (dismissed):
A few comments you could consider, but I'm good.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/assertions/AbstractEventualStreamAssertion.java
tinker-michaelj: Moved to `finally`.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/streams/EventualAssertionResult.java
tinker-michaelj: True, it's conceivable that `pass()` could be called just after the timeout period expires, but these timeouts are always set very liberally so won't be an issue in practice.

We can't make this method and the `fail()`/`pass()` methods all synchronized because then we couldn't call `pass()` once we started waiting.

david-bakin-sl: oh nuts you're right about synchronized.  And darn it that doesn't even rise to the level of "concurrency is hard".   So NVM that!

iwsimon (approved):
LGTM, thanks @tinker-michaelj

PR #15941, opened 2024-10-09
chore: migrate cryptography from .pfx files to .pem files

edward-swirldslabs:
**Description**:
This PR migrates  the cryptography from pfx files to pem files.    Merge targeting .55,  may need to go to develop instead since DAB was delayed to 56.  

Behavior Of Note: 
* This runs at every startup before loading cryptography from disk. 
* If successful, pfx files are moved to an `OLD_PFX_KEYS` subdirectory. 
* Success is determined by loading the PEM keys and comparing their encoding to the PFX keys
* If not successful, all created PEM files are deleted (no pre-existing PEM files are deleted) and pfx files remain unmoved. 
* All agreement key material is permanently deleted regardless of success or failure. 

Fixes #15827 

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #16025, opened 2024-10-17
chore: migrate cryptography from .pfx to .pem files

### LINKED ISSUE #15827: Migrate the private keys in pfx files to private keys in pem files.
For DAB support we need to replace the usage of pfx files for storing private keys with PEM files.   

If the public/private PFX file has been loaded, 
* generate the PEM files
* load the PEM files

Updated plan:  Instead of the following deletes, move the files to a sub-directory `OLD_PFX_KEYS` where they will not be read from. 
* Upon successful read of PEM files, delete the PFX files. 
* public PFX should be deleted last. 
* delete any agreement key material from disk. 

This needs unit testing explicitly.

edward-swirldslabs:
This PR has been created to target develop instead of release 55.   The current state of the code addresses all the comments located on the old PR: https://github.com/hashgraph/hedera-services/pull/15941

**Description**:
Behavior Of Note:

This runs at every startup before loading cryptography from disk.
If successful, pfx files are moved to an OLD_PFX_KEYS subdirectory.
Success is determined by loading the PEM keys and comparing their encoding to the PFX keys
If not successful, all created PEM files are deleted (no pre-existing PEM files are deleted) and pfx files remain unmoved.
All agreement key material is permanently deleted regardless of success or failure.

**Related issue(s)**:
Fixes #15827 

**Testing**: 

I've added the migration step to the positive case of the `EnhancedKeyStoreLoaderTest.java` and to the negative case 1 without problems.   

Adding the migration action to `keyStoreLoaderNegativeCase2Test` has complicating interactions with the design of the unit test.  Performing the migration does not unpack the agreement keys from the pfx files because we're deleting them and generating the agreement keys automatically with the signing key as the root certificate authority.    I've added the agreement key generation step with expected failures.   The injection of the keys into the address book is now failing due to the agreement keys being missing.  I've added special casing for when the injection step fails.   I suspect the reason that `legacy-invalid-case-2` does not cause failure on injection is that this case was using a degenerate agreement key to fail instead of a degenerate signing key, and the degenerate agreement key is being thrown away in the migration step and replaced with a good agreement key in the generation step.


iwsimon (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+    /**
+     * Move the PFX files to the OLD_PFX_KEYS sub-directory.
+     *
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private void cleanupMovePfxFilesToSubDirectory() throws KeyStoreException, KeyLoadingException {
+        AtomicLong cleanupErrorCount = new AtomicLong(0);
+
+        final String now = DateTimeFormatter.ofPattern("yyyy-MM-dd-HH:mm:ss").format(LocalDateTime.now());
```
anthony-swirldslabs, line 1447: Minor: it may make it difficult to work with files/directories that have `:` in their name because this character needs to be escaped. You may want to consider replacing `:` with, say, `_`, or instead add `h`, `m`, and `s` after the time numbers.

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@
...
+            // roll back due to errors.
+            // this deletes any pem files created, but leaves the agreement keys deleted.
+            logger.error(STARTUP.getMarker(), "Due to {} errors, reverting pem file creation.", errorCount);
+            rollBackSigningKeysAndCertsChanges(pfxPrivateKeys, pfxCertificates);
+        } else {
+            // cleanup pfx files by moving them to sub-directory
+            cleanupMovePfxFilesToSubDirectory();
+        }
+
+        logger.info(STARTUP.getMarker(), "Completed migration of key store.");
```
litt3, line 1223: Slightly misleading to log "completed migration", even in cases where migration failed. This log message could be moved into the `else` block above to remedy

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+    }
+
+    /**
+     * Delete any agreement keys from the key store directory.
+     */
+    private void deleteAgreementKeys() {
+        // delete any agreement keys of the form a-*
+        final File[] agreementKeyFiles = keyStoreDirectory.toFile().listFiles((dir, name) -> name.startsWith("a-"));
+        if (agreementKeyFiles != null) {
+            for (File agreementKeyFile : agreementKeyFiles) {
```
litt3, line 1234: `agreementKeyFile` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                if (agreementKeyFile.isFile()) {
+                    try {
+                        Files.delete(agreementKeyFile.toPath());
+                        logger.debug(STARTUP.getMarker(), "Deleted agreement key file {}", agreementKeyFile.getName());
+                    } catch (IOException e) {
```
litt3, line 1239: `e` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+     * @param pfxCertificates the map of extracted certificates.
+     * @return the number of errors encountered during the extraction process.
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private long extractPrivateKeysAndCertsFromPfxFiles(
+            final Map<NodeId, PrivateKey> pfxPrivateKeys, final Map<NodeId, Certificate> pfxCertificates)
+            throws KeyStoreException, KeyLoadingException {
+        final KeyStore legacyPublicStore = resolveLegacyPublicStore();
+        AtomicLong errorCount = new AtomicLong(0);
```
litt3, line 1263: `errorCount` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                        errorCount.incrementAndGet();
+                    } else {
+                        logger.trace(
+                                STARTUP.getMarker(),
+                                "Writing private signing key for node {} to PEM file {}",
+                                nodeId,
+                                sPrivateKeyLocation.getFileName());
+                        try {
+                            writePemFile(true, sPrivateKeyLocation, privateKey.getEncoded());
+                        } catch (IOException e) {
```
litt3, line 1294: `e` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                                    sPrivateKeyLocation.getFileName());
+                            errorCount.incrementAndGet();
+                        }
+                    }
+                }
+            }
+
+            // extract certificates for all nodes
+            final Path sCertificateLocation = keyStoreDirectory.resolve("s-public-" + nodeAlias + ".pem");
+            Path ksLocation = legacyCertificateStore();
```
litt3, line 1308: `ksLocation` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private long extractPrivateKeysAndCertsFromPfxFiles(
+            final Map<NodeId, PrivateKey> pfxPrivateKeys, final Map<NodeId, Certificate> pfxCertificates)
+            throws KeyStoreException, KeyLoadingException {
+        final KeyStore legacyPublicStore = resolveLegacyPublicStore();
+        AtomicLong errorCount = new AtomicLong(0);
+
+        iterateAddressBook(addressBook, (i, nodeId, address, nodeAlias) -> {
+            if (isLocal(address)) {
```
litt3, line 1266: This lambda is very long, in an already long method. Consider extracting into a static method

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                }
+            }
+
+            // extract certificates for all nodes
+            final Path sCertificateLocation = keyStoreDirectory.resolve("s-public-" + nodeAlias + ".pem");
+            Path ksLocation = legacyCertificateStore();
+            if (!Files.exists(sCertificateLocation) && Files.exists(ksLocation)) {
+                logger.trace(
+                        STARTUP.getMarker(),
+                        "Extracting signing certificate for node {} from file {} ]",
```
litt3, line 1312: is the trailing `]` in this message intentional?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                            errorCount.incrementAndGet();
+                        }
+                    }
+                }
+            }
+
+            // extract certificates for all nodes
+            final Path sCertificateLocation = keyStoreDirectory.resolve("s-public-" + nodeAlias + ".pem");
+            Path ksLocation = legacyCertificateStore();
+            if (!Files.exists(sCertificateLocation) && Files.exists(ksLocation)) {
```
litt3, line 1309: Some of these new migration methods have very deep nesting, which makes following the code a bit difficult.

Consider following alternative structures:

```
if (boolean) {
    // do nested stuff
}
```

becomes

```
if (!boolean) {
    return;
}

// do stuff
```

and

```
if (boolean) {
    // do stuff
} else {
    // do other stuff
}
```

becomes

```
if (boolean) {
    // do stuff
    return;
}

// do other stuff
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                    errorCount.incrementAndGet();
+                } else {
+                    logger.trace(
+                            STARTUP.getMarker(),
+                            "Writing signing certificate for node {} to PEM file {}",
+                            nodeId,
+                            sCertificateLocation.getFileName());
+                    try {
+                        writePemFile(false, sCertificateLocation, certificate.getEncoded());
+                    } catch (CertificateEncodingException | IOException e) {
```
litt3, line 1333: `e` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
         void apply(int index, NodeId nodeId, Address address, String nodeAlias)
                 throws KeyStoreException, KeyLoadingException;
     }
+
+    /////////////////////////////////////////////////////////////////////////////////////////////////
+    //////////////////////////////////// MIGRATION METHODS //////////////////////////////////////////
```
litt3, line 1180: Considering these are relatively self contained methods, consider creating a `migration utils` class, rather than tacking them on to this already long file

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+     * @param pfxPrivateKeys  the map of extracted private keys.
+     * @param pfxCertificates the map of extracted certificates.
+     * @return the number of errors encountered during the validation process.
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private long validateKeysAndCertsAreLoadableFromPemFiles(
+            final Map<NodeId, PrivateKey> pfxPrivateKeys, final Map<NodeId, Certificate> pfxCertificates)
+            throws KeyStoreException, KeyLoadingException {
+        AtomicLong errorCount = new AtomicLong(0);
```
litt3, line 1359: `errorCount` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+        iterateAddressBook(addressBook, (i, nodeId, address, nodeAlias) -> {
+            if (isLocal(address) && pfxCertificates.containsKey(nodeId)) {
+                // validate private keys for local nodes
+                Path ksLocation = privateKeyStore(nodeAlias, KeyCertPurpose.SIGNING);
```
litt3, line 1363: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                                pemPrivateKey.getEncoded(),
+                                pfxPrivateKeys.get(nodeId).getEncoded())) {
+                    logger.error(ERROR.getMarker(), "Private key for node {} does not match the migrated key", nodeId);
+                    errorCount.incrementAndGet();
+                }
+            }
+
+            // validate certificates for all nodes PEM files were created for.
+            if (pfxCertificates.containsKey(nodeId)) {
+                Path ksLocation = certificateStore(nodeAlias, KeyCertPurpose.SIGNING);
```
litt3, line 1376: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Extracts the private keys and certificates from the PFX files and writes them to PEM files.
+     *
+     * @param pfxPrivateKeys  the map of extracted private keys.
```
litt3, line 1253: The javadocs for `pfxPrivateKeys` and `pfxCertificates` are a bit misleading, since the implication based on the wording is that these maps *already* contain extracted material. When, in reality, these are empty maps that will be populated by this method.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                            || !Arrays.equals(
+                                    pemCertificate.getEncoded(),
+                                    pfxCertificates.get(nodeId).getEncoded())) {
+                        logger.error(
+                                ERROR.getMarker(),
+                                "Certificate for node {} does not match the migrated certificate",
+                                nodeId);
+                        errorCount.incrementAndGet();
+                    }
+                } catch (CertificateEncodingException e) {
```
litt3, line 1389: `e` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+     * @param pfxPrivateKeys  the map of extracted private keys.
+     * @param pfxCertificates the map of extracted certificates.
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private void rollBackSigningKeysAndCertsChanges(
+            final Map<NodeId, PrivateKey> pfxPrivateKeys, final Map<NodeId, Certificate> pfxCertificates)
+            throws KeyStoreException, KeyLoadingException {
+
+        AtomicLong cleanupErrorCount = new AtomicLong(0);
```
litt3, line 1410: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+        iterateAddressBook(addressBook, (i, nodeId, address, nodeAlias) -> {
+            // private key rollback
+            if (isLocal(address) && pfxPrivateKeys.containsKey(address.getNodeId())) {
+                try {
+                    Files.deleteIfExists(privateKeyStore(nodeAlias, KeyCertPurpose.SIGNING));
+                } catch (IOException e) {
```
litt3, line 1416: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                    cleanupErrorCount.incrementAndGet();
+                }
+            }
+            // certificate rollback
+            if (pfxCertificates.containsKey(address.getNodeId())) {
+                try {
+                    Files.deleteIfExists(certificateStore(nodeAlias, KeyCertPurpose.SIGNING));
+                } catch (IOException e) {
```
litt3, line 1424: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+    }
+
+    /**
+     * Move the PFX files to the OLD_PFX_KEYS sub-directory.
+     *
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private void cleanupMovePfxFilesToSubDirectory() throws KeyStoreException, KeyLoadingException {
+        AtomicLong cleanupErrorCount = new AtomicLong(0);
```
litt3, line 1445: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+        }
+    }
+
+    /**
+     * Move the PFX files to the OLD_PFX_KEYS sub-directory.
+     *
+     * @throws KeyStoreException   if the underlying method calls throw this exception.
+     * @throws KeyLoadingException if the underlying method calls throw this exception.
+     */
+    private void cleanupMovePfxFilesToSubDirectory() throws KeyStoreException, KeyLoadingException {
```
litt3, line 1444: Did you intend to name this method "cleanup move..."? It seems the verb should be *either* "cleanup", or "move".

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+
+        logger.info(STARTUP.getMarker(), "Cryptography Migration Cleanup: Moving PFX files to {}", pfxDateDirectory);
+
+        if (!Files.exists(pfxDateDirectory)) {
+            try {
+                if (!Files.exists(pfxArchiveDirectory)) {
+                    Files.createDirectory(pfxArchiveDirectory);
+                }
+                Files.createDirectory(pfxDateDirectory);
+            } catch (IOException e) {
```
litt3, line 1459: `e` may be `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                logger.error(
+                        ERROR.getMarker(),
+                        "Failed to create [.archive/yyyy-MM-dd_HH-mm-ss] sub-directory. Manual cleanup required.");
+                return;
+            }
+        }
+        iterateAddressBook(addressBook, (i, nodeId, address, nodeAlias) -> {
+            if (isLocal(address)) {
+                // move private key PFX files per local node
+                File sPrivatePfx = legacyPrivateKeyStore(nodeAlias).toFile();
```
litt3, line 1469: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1165,4 +1175,340 @@ private interface AddressBookCallback {
...
+                if (sPrivatePfx.exists()
+                        && sPrivatePfx.isFile()
+                        && !sPrivatePfx.renameTo(
+                                pfxDateDirectory.resolve(sPrivatePfx.getName()).toFile())) {
+                    cleanupErrorCount.incrementAndGet();
+                }
+            }
+        });
+        File sPublicPfx = legacyCertificateStore().toFile();
```
litt3, line 1478: `final`

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: updated to remove colons

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: so moved.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: making final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: making final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: making final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: found the IDE settings to turn this detection on.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: nope, removed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: the lambda is the only logic in the method containing it.  There would not be any code improvement from doing that.  Doing the embedded lambda is necessary to get the error count given the API for calling the lambda.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: making final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: I think this is somewhat arbitrary.  The last job I worked at was a defense contractor which required explicit else statements on all ifs to avoid the ambiguity of whether or not an else statement was left off on purpose or if it was a coding mistake and not considered as a case.    

Another coding philosophy that I have seen enforced at some places is to have a single return statement, which would force explicit casing for code paths.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: I placed it here in this way thinking the code would get deleted at some point when the migration was no longer relevant.    The `migrate()` method is not static, but called from the instance.  It makes sense as a preemptive phase in the sequence of 
```
migrate()
scan()
generateIfNecessary() 
verify()
injectInAddressBook()
```

The methods are not self-contained, but use static private variables and make calls to private static methods.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: updated wording

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: colons are not allowed in method names. :-).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: made final

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
litt3: At the end of the day, code with blocks nested 5 deep is very difficult to understand, check for bugs, and maintain. My suggestion is just one way to avoid such deep nesting. Another way would be to make more static helper methods, which can be individually considered.

litt3 (dismissed):

anthony-swirldslabs (approved):

litt3 (approved):

PR #16028, opened 2024-10-18
chore: use `0s` as `@RepeatableHapiTest` valid start offset

### LINKED ISSUE #16027: Use zero `validStart` offset in `RepeatableEmbeddedHedera` to simplify testing receipt expiration
### Problem

[This](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/TxnRecordRegression.java#L121) spec passes even with a much shorter sleep than `181s` because the `validStart` of the transaction whose receipt is being purged is being offset by `42s` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/hedera/embedded/AbstractEmbeddedHedera.java#L81).

Hence it expires after any sleep closer to `180s - 42s = 138s`, which makes the test less precise.

### Solution

Change `RepeatableEmbeddedHedera` to use a `0s` offset for its `validStart` times.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #16027


david-bakin-sl (approved):
I get this change - LGTM - what I don't get is what was magic about the 42s wait that had been coded in (and is still used for `ConcurrentEmbeddedHedera`).

PR #16031, opened 2024-10-18
chore: use `fireAndForget()` for freeze period background traffic

### LINKED ISSUE #16030: `@HapiTest` freeze boundary traffic sometimes logging too much
### Description

E.g. [here](https://github.com/hashgraph/hedera-services/runs/31647093635) we see the Restart check produced so many logs that the JUnit XML could not be parsed, even though the test passed.

The problem is that [this](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/UtilVerbs.java#L542) op may end up submitting lots of transactions that don't reach consensus before the freeze round.

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

_No response_

### Version

develop

### Operating system

None

tinker-michaelj:
**Description**:
 - Closes #16030


Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

PR #16048, opened 2024-10-18
chore: tolerate nextNodeId field in config.txt

### LINKED ISSUE #15934: Allow config.txt to have `nextNodeId` but ignore it.
https://github.com/hashgraph/hedera-services/issues/15215 removes all dependency on next node id and fails if the value is present in the file.   What is needed to reduce complexity of deployment of this change through our release pipeline is that the field is allowed to stay in the file, but its value is not read or used.   Once the CI/CD pipelines are updated, the field can be dropped from the file.

edward-swirldslabs:
**Description**:
In PR https://github.com/hashgraph/hedera-services/pull/15791, `nextNodeId` was completely removed from config.txt.  DevOps indicated that a hard removal of the field would make things more complicated as they switch back and forth between software version in test environments.  The decision was made to allow the field to be present until all test environments and CI/CD pipelines were on a version of the software that did not require the field to be present, then it could be removed from the format. 

Fixes #15934 

A new issue was created to undo the effects of this PR when we can completely remove the field.  https://github.com/hashgraph/hedera-services/issues/16049


anthony-swirldslabs (approved):

iwsimon (approved):
LGTM

poulok (approved):

PR #16063, opened 2024-10-21
test: add test to validate null admin key for contract can still xfer value

### LINKED ISSUE #15982: create test to check if contract with null admin key can still change state through system contracts
### Problem

A contract with a null admin key should still be able to update its state via system contracts.  This would be consistent with the security model v2.

### Solution

- create a contract with a null admin key
- try altering its state via system contract via a HAPI test

### Alternatives

_No response_

lukelee-sl:
**Description**:
Contracts that have a null admin key should still be able to transfer value that it owns via system contracts.  Add a test to validate.

**Related issue(s)**:

Fixes #15982

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AtomicCryptoTransferHTSSuite.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AtomicCryptoTransferHTSSuite.java
@@ -1437,4 +1437,42 @@ public class AtomicCryptoTransferHTSSuite {
                                         .contractCallResult(
                                                 htsPrecompileResult().withStatus(SPENDER_DOES_NOT_HAVE_ALLOWANCE)))));
     }
+
+    @HapiTest
+    final Stream<DynamicTest> nullContractAdminKeyTransfer() {
+        final var nullAdminKeyXferTxn = "nullAdminKeyXferTxn";
+        return hapiTest(
+                cryptoCreate(RECEIVER),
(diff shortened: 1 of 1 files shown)


kimbor (approved):

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/AtomicCryptoTransferHTSSuite.java
```diff
@@ -1437,4 +1437,42 @@ final Stream<DynamicTest> blockCryptoTransferForPermittedDelegates() {
...
+                            spec,
+                            contractCall(
+                                            CONTRACT,
+                                            TRANSFER_MULTIPLE_TOKENS,
+                                            transferList()
+                                                    .withAccountAmounts(
+                                                            accountAmount(contract, -amountToBeSent, false),
+                                                            accountAmount(receiver, amountToBeSent, false))
+                                                    .build(),
+                                            EMPTY_TUPLE_ARRAY)
```
david-bakin-sl, line 1463: I see this pattern in several tests in this file - but I don't get why there's both a `transferList()...build()` argument and an `EMPTY_TUPLE_ARRAY` argument.  The method in the contract takes a single argument of type `TokenTransferList[]`.  How do these two arguments "coaelsce" or something into one `TokenTransferList[]`?

david-bakin-sl (commented):
Except for my question for my own education, this test looks fine.

stoyanov-st (approved):

Issue #15989, opened 2024-10-15
TSS Roster Epic Tickets Update


poulok:
This ticket is to track the effort to update all tickets in alignment including details to ensure that all team members have the same understanding of the work to be done and the acceptance criteria. The team's agreed upon details are documented in the team slack channel.

This includes the updated TSS Roster proposal being accepted.

edward-swirldslabs:
The epic has been revised to approvable standards.

PR #14045, opened 2024-06-25
docs: TSS Rosters design proposal

kfa-aguda:
**Description**:

This design doc provides the specification for `Rosters` - a new Data Structure introduced for enabling the Threshold Signature Scheme. The proposal details the data structure and lifecycle of `Rosters`, as well as the changes needed from Services and DevOps. 

fixes #13893

PR #15442, opened 2024-09-11
feat: Implement APIs for storing and retrieving rosters to/from state

kfa-aguda:
In this PR
1. The `ReadableRosterStore` and `WritableRosterStore` have been introduced as Roster stores.
2. During software upgrade, the candidate roster, if found in the state, is unconditionally retrieved from the rosterstore and adopted as the active roster
3. the adopted (previously candidate roster) is removed from the state.
4. during normal restart (not a software upgrade), an active roster from the state is used if present. Otherwise, one is constructed from the AB. 

**Edit**
There has been a slight change in the design of the feature this PR was originally intended to address. The introduction of Consensus vs Execution brings some responsibility changes as it relates to which component(?s) will determine the active roster. The Tss Base Service will need the history of Rosters whilst ServicesMain will now be responsible for introducing the logic to arrive at the active Roster the platform will use. Therefore, instead of this PR including all of that logic to adopt a candidate roster, it will only be responsible for introducing the APIs needed to interact with Rosters and state. The other points will be addressed in a future work.
Main classes that need to determine the active roster may choose to call the APIs in this PR to set and adopt a roster or derive another roster through some other mechanism.

More explicitly, this PR addresses the following:

1. Points 1, 3 & 4 above
2. Changes required to register Services/Schema/State/Migration registrations.

Point #3 will be completed in a follow-up PR later.

Closes #14711

PR #16067, opened 2024-10-21
docs: Update the TSS Roster Proposal

### LINKED ISSUE #16066: Update the TSS Roster Proposal

kfa-aguda:
**Description**:
This proposal modifies the TSS Roster Design Proposal specified in #14045 in order to provide Rosters for TSS Ledger ID effort.

There have been some updates and refinements of components since both the previous Roster ID and TSS Ledger ID proposals got accepted. This effort is an attempt to revise some of the work items in the Roster epic and ensure alignment and acceptance by all teams involved.

The following updates have been made:
1. Add the introduction of Roster stores, why and what new stores have been introduced, with their expected usage.
2. Specify that the genesis roster, override etc will all be loaded from their equivalent config.txt files and a new determine will be introduced in ServicesMain which takes all of these rosters and returns an active roster history (usually 2, previous and current)
3. The freeze upgrade transaction will be a trigger for creating and setting the candidate roster in the state.
4. Specify that a new feature flag will be introduced to gate this feature.
5. Explain round number determination - either latest state round + 1 or just 1, and why.
6. Update diagrams and decision trees (2 items).
7. Explain the flow the very first time roster feature flag is enabled.
8. Replace existing TSS Roster Lifecycle diagram with the latest diagram.

**Related issue(s)**: #15989 #14045

Fixes #16066


platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -18,61 +18,86 @@ components within the platform.
...
+The steps to validate and add the Roster to the state will be executed within the same thread that submits the request.
+A caught exception will be thrown if the Roster is invalid. The durability of the submitted Roster is ensured by storing it in the state.
 
 ## Data Structure
 
-The App is already responsible for managing the address book. We propose that it continues to do so, and when it
-deems necessary (such as when it receives a HAPI transaction), creates a candidate Roster object and passes it to
-the platform via the new API. The platform will then validate the roster and store it in the state.
+The App is already responsible for managing the address book. We propose that it continues to do so, and when it deems necessary (such as during a freeze upgrade), creates a candidate `Roster` object and passes it to the platform via the new API.
+The platform will then validate the roster and store it in the state.
```
anthony-swirldslabs, line 100: This looks a bit inaccurate. The Roster Service is an app/services-level entity. It has nothing to do with the "platform". Further, the "platform" (whatever that means nowdays) doesn't even need to know about candidate rosters. It will be told what Roster to use as an active roster (and what prior history of rosters is) on startup.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -278,10 +299,9 @@ A Roster is considered valid if it satisfies the following conditions:
 6. All ServiceEndpoint/s must have a valid IP address or domain name (mutually exclusive), and port.
 7. The roster must have a unique NodeId for each RosterEntry.
 
-On the submission of a new Candidate Roster, the platform will validate the roster against these conditions. Note
-that a constructed `Roster` can be valid, but not accepted by the platform. For example, if a new candidate roster
-is set via the API, but its hash evaluates to the hash of the existing candidate roster, the new roster will be
-discarded. That is, the operation has no effect.
+On the submission of a new Candidate Roster, the platform will validate the roster against these conditions.
+Note that a constructed `Roster` can be valid, but not accepted by the platform.
```
anthony-swirldslabs, line 303: Similar to above, I suggest to re-evaluate usages of the word "platform" in this document to align it with the current (and future) state of the architecture.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -293,173 +313,157 @@ the state:
...
-A quick note on PCES. The software version is used to determine which roster (active or previous) should be used to
-validate the signatures of events in the PCES. If the PCES event version is less than the current software version,
-it indicates that the event was signed with the old roster, and hence, the old roster is used for validation. If the
-PCES event version matches the current software version, the active roster is used. This behavior will be
-maintained until Birth Rounds or Dynamic Address Book is implemented.
+A quick note on Pre-Consensus Event Stream (PCES).
+The software version is used to determine which roster (active or previous) should be used to validate the signatures of events in the PCES.
+If the PCES event version is less than the current software version, it indicates that the event was signed with the old roster, and hence, the old roster is used for validation.
+If the PCES event version matches the current software version, the active roster is used.
+This behavior will be maintained until Birth Rounds or Dynamic Address Book is implemented.
```
anthony-swirldslabs, line 372: I think we need to expand this section to avoid the usage of the "previous roster". Note that the "previous AddressBook" is okay as it describes the current behavior, but there's not a "previous roster" - there's just an entry in the history with an older round number.

Also, and most importantly, we need to describe those artificially added records to the history in case the old and the current roster are the same (or there's not an old roster at all.)

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -293,173 +313,157 @@ the state:
...
+Now might be a good time to revisit the startup decision tree diagram for reference.
 
 The next round number for this network will continue from the last round number in the provided state.
-The trigger for adopting the `Override Roster` will be the presence of an Override Roster at start-up with a State
-existing on disk. Bear in mind the Network Transplant process is designed primarily with test networks in mind.
-Upon adoption of the Override Roster, the Override Roster should be deleted from disk. In test networks where this
-process is designed to be used, the Override Roster can be easily accessed and deleted by DevOps.
+Note that the Network Transplant process is designed primarily with test networks in mind.
+Upon adoption of the `Override Roster`, the `override-config.txt` should be deleted/renamed from disk.
+In test networks where this process is designed to be used, the `override-config.txt` file can be easily accessed and deleted by DevOps.
```
anthony-swirldslabs, line 383: The wording here is a bit ambiguous. The first sentence says "it should be deleted". A reader would assume something will implement something to delete it.

Then the second sentence says "it can be deleted"... manually. Okay.

Then what happens if it could be deleted, but actually wasn't deleted? Does this cause an error? Is this idempotent? Or....?

I suggest to rephrase this to clarify the procedure and/or mechanism.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -18,61 +18,86 @@ components within the platform.
...
 ## Architecture
 
-###### Roster Lifecycle
+### Roster Types.
+
+This proposal introduces four types of `Roster`s - candidate `Roster`, active `Roster`, genesis `Roster`, and override `Roster`.
 
-![](TSS%20Roster%20Lifecycle.drawio.svg)
+1. A Candidate `Roster` is a Roster that is currently being considered for adoption to become the active `Roster`.
+2. An Active `Roster` is a Roster currently being used for signing of blocks.
```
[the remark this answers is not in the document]
kfa-aguda: I think the dilema when I was writing this is ... do we frame this in the end goal context vs today's? I went for the end goal but I'm not sure. If we do use today's framing, it will have to carry the same definition as a the active address book which itself too may be confusing.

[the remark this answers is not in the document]
kfa-aguda: Will "An Active Roster is the roster currently being used to determine consensus" work?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
kfa-aguda: Thanks. I've further revised the use of the word "platform".

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
kfa-aguda: good catch. Clarified.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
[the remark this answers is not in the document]
anthony-swirldslabs: Yes, this latter phrasing looks perfect to me. Thanks!

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
anthony-swirldslabs: Not quite accurate. The round number in the Roster history tells us at which round a particular Roster has become active. For example, the history could look like `[{100, hash1}, {200, hash2}]`, while currently we may be processing the round number 250. Per this history object, even though the hash2 has a round number that is less than the current round number, the Roster with `hash2` is still considered active as of now.

Regardless though, the updated wording in the document looks fine. Thanks!

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -18,61 +18,86 @@ components within the platform.
...
+The `WritableRosterStore#setCandidateRoster()` method will be used to set a candidate `Roster` in the state as follows:
+
+1. Validate the candidate `Roster`.
+2. Store the candidate `Roster` hash in the `RosterState`, and the Roster itself in the `RosterMap`.
+
+The `WritableRosterStore#setActiveRoster()` method will be called to set a new active `Roster` in the state. A call to `setActiveRoster` will infer the following operations:
+
+1. Invalidate and remove any existing candidate `Roster` present in the state
+2. Rotate the current active `Roster` to the previous active `Roster`
+3. Set the new Roster as the active `Roster` in the state
```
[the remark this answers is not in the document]
anthony-swirldslabs: The updated version looks good to me. Thanks for addressing this, @kfa-aguda .

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
kfa-aguda: I think my use of the word "consensus round" is inaccurate. The point stands though that there's an existence of a "previous roster", and it is the roster that was used to sign events in rounds older than the starting round number of the active roster. One may or may not be present in the state. It won't be present if in Genesis or in the edge cases described in this proposal (which we have a solution to artificially create one). 
I know this comment is resolved, but just adding this comment for completeness.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -524,16 +604,23 @@ Some of the obvious test cases to be covered in the plan include validating one
...
+### DevOps Changes, in summary
+
+1. DevOps will decide whether to keep the `config.txt` model and if so, work with Services on its translation into a
+   Roster.
+2. DevOps will be responsible for providing the `config.txt`, `genesis-config.txt` or `override-config.txt` as files in
+   the desired location.
+3. DevOps will be responsible for managing the lifecycle of the files on disk that are used in creating the Roster,
+   including the new TSS encryption key.
+4. DevOps will be responsible for cleaning up the `override-config.txt` file after the network transplant process is
+   complete.
```
anthony-swirldslabs, line 626: No strong opinion here, but we could consider the `.archive/<date_time>/` approach here to simplify DevOps' life. If we don't, then this may cause issues when e.g. one restarts a node twice with an override roster present - they'll end up completely forgetting the actual previous roster.

Speaking of which, did we consider all the edge-cases for PCES replay when an override roster is provided? Are we sure we can safely forget the actual previous roster in any and all cases?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
+### Design Decisions
+
+The roster data has been reduced to the bare essentials required by the platform.
+
+1. A `Roster` is a list of `RosterEntry` in order of NodeId.
+2. A `RosterEntry` contains the following fields for a consensus node in the network:
+   1. nodeId
+   2. consensus weight
+   3. gossip_ca_certificate - containing the gossip signing public key
+   4. tss_encryption_key - not used until TSS is implemented
```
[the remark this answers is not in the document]
edward-swirldslabs: removing

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
+than storing it in the state. When TSS is implemented, there will be TSS related key material stored in the state,
+and it is imperative that this key material stay in sync with the `Roster` data. Leaving the content as a modifiable
+file on disk would pose risks of introducing errors.
+
+## Changes
+
+The platform `AddressBook` lifecycle is being replaced with a new `roster` lifecycle. A new feature flag
+`addressbook.useLegacyConfigTxt` will be used to toggle between the current `config.txt` based address book
+lifecycle and the new `roster` lifecycle. The feature flag will be set to `true` by default. The new `roster`
+lifecycle will be enabled by setting the flag to `false`.
```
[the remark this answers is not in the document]
edward-swirldslabs: changing name of flag to reflect flip on values.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
+Conditions:
+
+1. A state is loaded from disk
+2. No software upgrade is happening
+
+Ignored Files: (Log warning if present)
+
+1. `config.txt`
+2. `genesis-config.txt`
+3. `override-config.txt`
```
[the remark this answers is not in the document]
edward-swirldslabs: The warning that an override can't be applied when the use is trying to trigger it makes sense to me.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
+
+Startup Tasks:
+
+1. If there exists active rosters in the roster state.
+   1. Read the active rosters and construct the existing rosterHistory from roster state
+   2. Start the platform with the existing rosterHistory.
```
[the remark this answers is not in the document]
edward-swirldslabs: Until the code is migrated away from a class saying `Platform` and `PlatformBuilder`, I disagree.  It is cognitively confusing to anyone who has not been part of the design conversations with Richard.    I prefer to write proposals to the lowest common denominator of knowledge.  This proposal will be implemented long before the term `Platform` can be retired.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Agreed, this was a section I hadn't yet got to updating.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -192,27 +432,18 @@ message RosterEntry {
      * SHALL NOT supply both values for the same endpoint.<br/>
      * This list SHALL NOT be empty.<br/>
      */
-    repeated proto.ServiceEndpoint gossip_endpoint = 5;
+    repeated proto.ServiceEndpoint gossip_endpoint = 4;
```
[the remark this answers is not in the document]
edward-swirldslabs: Fair.  I made the field name and the number reserved.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
+2. If there is no roster state content, this is a fatal error: The migration did not happen on software upgrade.
+
+##### MODE: Network Transplant
+
+Conditions:
 
-The steps to validate and add the Roster to the state will be executed within the same thread that submits the request.
-A caught exception will be thrown if the Roster is invalid. The durability of the submitted Roster is ensured by storing it in the state.
+1. A state is loaded from disk
+2. A software upgrade is happening
```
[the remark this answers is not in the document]
edward-swirldslabs: The practice right now is to do a configuration based software upgrade.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,93 +19,353 @@ components within the platform.
...
 
-## Data Structure
+Ignored Files: (Log warning if present)
+
+1. `genesis-config.txt`
+
+Startup Tasks:
+
+1. If there is a candidate roster in the state (Non-Migration Software Upgrade)
+   1. candidateRoster := read the candidate roster from the roster state.
```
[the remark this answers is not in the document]
edward-swirldslabs: This will possibly change pending feedback from DevOps and I will updated when we know more.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -115,9 +114,9 @@ file on disk would pose risks of introducing errors.
 ## Changes
 
 The platform `AddressBook` lifecycle is being replaced with a new `roster` lifecycle. A new feature flag
-`addressbook.useLegacyConfigTxt` will be used to toggle between the current `config.txt` based address book
-lifecycle and the new `roster` lifecycle. The feature flag will be set to `true` by default. The new `roster`
-lifecycle will be enabled by setting the flag to `false`.
+`addressbook.useRosterLifecycle` will be used to toggle between the current `config.txt` based address book
+lifecycle and the new `roster` lifecycle. The feature flag will be set to `fakse` by default. The new roster
```
anthony-swirldslabs, line 118: Nit: `s/fakse/false/`

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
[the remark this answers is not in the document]
anthony-swirldslabs: Let's hear what DevOps say, but as mentioned in another thread, I believe an override shouldn't require a software upgrade - I don't see any objective reasons why it should - in which case the presence of this file should trigger the override, and hence there's not a need for any warnings. To put it differently: if we detected that the file has been present, then we're not in the "Normal Restart" branch already, so we'll never need to emit the warning here specifically.

But let's confirm with DevOps just to see if there's indeed a good reason to require a software upgrade for this.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -582,45 +586,93 @@ public class WritableRosterStore extends ReadableRosterStore {
...
+
+#### Software Upgrade (No Transplant, Subsequent Upgrades)
+
+1. There must be a loadable state present on disk in the standard location.
+2. There must be a software upgrade happening.
+3. Ensure that `config.txt`, `genesis-config.txt`, and `override-config.txt` are not present on disk.
+
+#### Software Upgrade Network Transplant
+
+1. Copy the PCES files from a single node in the source network and replicate to all nodes in the target network.
```
anthony-swirldslabs, line 673: Minor q: does this also assume that a state **from a single node** (probably, the same single node) will be replicated to all the nodes as well? If so, let's add a point about this for clarity.

If not, then I'm wondering how reliable is it to use PCES files from a single node with potentially different states on different nodes? I'd guess this may cause some flakiness, and I'm not sure if there's a good reason to tolerate this flakiness.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: ack!

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: If the states are different, an ISS would result.   There is a known event branching issue when we copy multiple PCES files from different nodes.   But I will add the text about using the same state file.

anthony-swirldslabs: Sounds good. Thanks for adding the text!

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -210,60 +210,52 @@ Startup Tasks:
...
-   8. set (currentRoster, currentRound) as the active roster in the roster state.
-   9. Start the platform with the new rosterHistory.
+1. Read the current AddressBooks from the platform state.
+2. configAddressBook := Read the address book in config.txt
+3. previousRoster := translateToRoster(currentAddressBook)
+4. currentRoster := translateToRoster(configAddressBook)
+5. currentRound := state round +1
+6. rosterHistory := `[(currentRoster, currentRound), (previousRoster, 0)]`
+7. set (previousRoster, 0) as the active roster in the roster state.
+8. set (currentRoster, currentRound) as the active roster in the roster state.
```
anthony-swirldslabs, line 239: Minor: basically, we want to perform 7 and 8 in order to build the 6. I'd suggest to either make the 7 and 8 sub-points of the 6, or put 6 after the 7 and 8 and state that this history state is the result of the previous two actions. That would look more logical compared to the current way the points are presented.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -668,11 +679,11 @@ Manual Testing Plan:
...
+1. Copy the state and PCES files from a single node in the source network and replicate them to all nodes in the target
+   network.
 2. There must be a loadable state present on disk in the standard location.
-3. There must be a software upgrade happening.
-4. Provide a `override-config.txt` file on disk in the same directory that `config.txt` is normally provided.
-5. Do not provide a `genesis-config.txt` or `config.txt` file on disk.
-6. After startup the `override-config.txt` file will be moved to a subdirectory `archive/yyyy-MM-dd_HH-mm-ss/`.
+3. Provide a `override-config.txt` file on disk in the same directory that `config.txt` is normally provided.
+4. Do not provide a `genesis-config.txt` or `config.txt` file on disk.
+5. After startup the `override-config.txt` file will be moved to a subdirectory `archive/yyyy-MM-dd_HH-mm-ss/`.
```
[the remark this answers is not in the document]
edward-swirldslabs: updated

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: reordered

anthony-swirldslabs (dismissed):

anthony-swirldslabs (approved):
+1

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
+during a software upgrade. Discrepancy in the information present between the `Roster` data structure and the
+`AddressBook` data structure make config.txt a poor medium for carrying out `Roster` updates. Writing the updated
+roster out to disk is less secure than storing it in the state. When TSS is implemented, there will be TSS related
+key material stored in the state, and it is imperative that this key material stay in sync with the `Roster` data.
+
+After this proposal is implemented, new rosters will be created and stored in the state for adoption at a software
+upgrade boundary. It has been common practice in testing to use manually created config.txt files to start new
+networks from genesis and transplant merkle state to a different new network from an existing network. To continue to
+allow these processes, alternate mechanisms of providing a genesis roster and a network overriding roster will be
+introduced.
```
poulok, line 44: I think it would be helpful to state explicitly that `config.txt` will no longer be used at all, and that in normal production operation, the `genesis.json` file will be used once at genesis and then no on-disk non-state representations of the address book or roster will be used again.

This is a non-material update, should you agree and decide to add this.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
+
+### Dependencies, Interactions, and Implications
+
+The implementation of the TSS-Roster proposal must come after the completion of DAB Phase 2 and must be completed
+before TSS-Ledger-ID.
+
+This change in the mechanics of how the roster is updated, new networks are started, and how merkle state is
+transplanted will impact current DevOps practices significantly. SOLO and all testing environments will need to be
+updated as well. This change includes the inversion of control of the startup process where in the responsibility
+for reading the state, address book, and configuration from disk is moved from the platform to `ServicesMain`.
```
poulok, line 54: Not sure if this is the right place, but it would be nice to explicitly call out that the implementation of much of this feature lies within the bounds of services code as things are currently defined.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
+
+This roster life-cycle change has been carved off from the TSS effort as a separate proposal to reduce the size and
+scope of changes per release and allow the changes entailed by TSS to be more incremental.
+
+### Requirements
+
+1. Immutability: Rosters must be immutable to protect the integrity of its computed hash.
+2. Efficient Storage: Roster data stored in the state should minimize overhead and unnecessary data copying.
+3. Clear API: To create and submit candidate rosters, and to manage the lifecycle of submitted rosters.
```
poulok, line 63: Also for the application to provide the necessary rosters to the platform on startup

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
-- Address book paradigm discontinued within the platform codebase in favor of `Roster`s
+The use of config.txt for starting new networks from genesis and transplanting merkle state to a new network will
+be replaced and require the presence of specific files on disk to initiate these scenarios. A `genesis-config.txt`
+file will be used to start a new network from genesis, and an `override-config.txt` file will be used to transplant
+merkle state to a new network. Initiating these scenarios through distinct files will make the logic for
+determining which roster to use at startup more straightforward.
 
-## Architecture
+To be consistent with future consensus design, the running platform code will not directly read from or write to the
+merkle state. The appropriate roster data will be provided to the platform during platform construction.
```
poulok, line 103: >platform code will not directly read from or write to the
merkle state

This is true for the roster data, but is still true for other platform data. I suggest `...platform code will not directly read from or write to the merkle state when it comes to roster data.`

poulok (commented):
Still going through, wanted to leave these here in the interim. None of my comments are material.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
+   signatures. The active roster with latest round number is used to determine consensus
+   * The `Roster History` is a list of active rosters paired with the rounds they became active. At this time the
+     history length is at most 2.
+     * `currentRoster` - the active roster in the history paired with the highest round number.
+     * `previousRoster` - the active roster in the history paired with the lowest round number.
+     * In this document we denote the roster history as an ordered list of pairs by descending round
+       number. Example:  [(currentRoster, 2), (previousRoster, 1)].
+2. A `Genesis Roster` is loaded from disk at the genesis of a new network and becomes the first active roster.
+3. The `Candidate Roster` is a roster that will become active at the next software upgrade.
+4. An `Override Roster` is loaded from disk and unconditionally set as the latest active roster. This is used to
```
poulok, line 146: Could the override roster also be used in production in case of an emergency, for example if the roster was corrupted and the process to update it had a bug that prevented it from being fixed with transactions?

edward-swirldslabs: yes, but this will break TSS once we have TSS.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: `genesis.json` is a new file name and I assume format to me.    This proposal does not suggest switching to a JSON format and this is why we're leaving it with the `X-config.txt` name since we're using the format in `config.txt`.

edward-swirldslabs: updated wording to reflect your suggestion

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Added comment here.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Adjusted wording to indicate its the application doing the management.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: made the constraint explicitly scoped to roster data.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,366 @@ components within the platform.
...
+This proposal introduces four types of `Roster` - `active`, `candidate`, `genesis`, and `override`.
+
+1. `Active Rosters` are paired with the round numbers for when they became active and are used to validate event
+   signatures. The active roster with latest round number is used to determine consensus
+   * The `Roster History` is a list of active rosters paired with the rounds they became active. At this time the
+     history length is at most 2.
+     * `currentRoster` - the active roster in the history paired with the highest round number.
+     * `previousRoster` - the active roster in the history paired with the lowest round number.
+     * In this document we denote the roster history as an ordered list of pairs by descending round
+       number. Example:  [(currentRoster, 2), (previousRoster, 1)].
```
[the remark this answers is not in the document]
edward-swirldslabs: I think this is one of those things were the telephone game mutates what was said and the new category took on a life of its own in too many minds that I decided it wasn't worth fighting.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -144,7 +151,8 @@ This proposal introduces four types of `Roster` - `active`, `candidate`, `genesi
 2. A `Genesis Roster` is loaded from disk at the genesis of a new network and becomes the first active roster.
 3. The `Candidate Roster` is a roster that will become active at the next software upgrade.
 4. An `Override Roster` is loaded from disk and unconditionally set as the latest active roster. This is used to
-   transplant the state to a different network.
+   transplant the state to a different network.  Until the TSS effort is delivered and the network is given a ledger
+   id, this override mechanism can be used to manually correct the roster / address book on the current network.
```
anthony-swirldslabs, line 155: Minor: I'd suggest to note that using the override roster would/should still be possible as an emergency mechanism, even after the TSS is enabled, with an understanding that the ledger id cannot be preserved in that case.

anthony-swirldslabs (dismissed):
+1

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: I think that is the domain of a later TSS Proposal and needs to be thought through more thoroughly than what we can say here.

mxtartaglia-sl (dismissed):

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -214,13 +493,7 @@ message RosterState {
      */
     repeated RoundRosterPair round_roster_pairs = 2;
```
Neeharika-Sompalli, line 494: Since the proposal mostly mentions this list as `rosterHistory`, we should consider renaming the `round_roster_pairs` to `roster_history`

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,374 @@ components within the platform.
...
+
+1. `config.txt`
+2. `override-config.txt`
+
+Startup Tasks:
+
+1. Translate `genesis-config.txt` to a genesisRoster.
+2. Create a genesis state with the genesisRoster.
+3. Set (genesisRoster, 0) ase the new active roster in the roster state.
+4. rosterHistory := `[(genesisRoster, 0)]`
```
netopyr, line 216: I wonder if the following statement would simplify the implementation.

```
rosterHistory := [(genesisRoster, 0), (genesisRoster, 0)]
```

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -96,22 +404,18 @@ message Roster {
...
 /**
  * A single roster entry in the network state.
  *
- * Each roster entry in SHALL encapsulate the elements required
- * to manage node participation in the Threshold Signature Scheme (TSS).<br/>
  * All fields are REQUIRED.
  */
 message RosterEntry {
 
+    // The tssEncryptionKey is no longer in the roster.
```
netopyr, line 415: Is this permanently removed or just moved to another proposal?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
-
-The logic to determine whether to invoke this method could be determined by some mechanism Services deems appropriate.
-
-#### Override Roster
+public interface ReadableRosterStore {
+    /**
+     * Gets the candidate roster if found in state or null otherwise.
+     * Note that state commits are buffered,
+     * so it is possible that a recently stored candidate roster is still in the batched changes and not yet committed.
+     * Therefore, callers of this API must bear in mind that an immediate call after storing a candidate roster may return null.
```
netopyr, line 534: I still do not understand why we have to insist on this behavior. It breaks with the commonly expected behavior of getters and setters and differs from all other stores that exist in consensus node.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
-a specific state, with a specific set of nodes, such as during a Network Transplant Process. A common example of
-this is Mainnet State being transplanted into a testing network.
+    /**
+     * Gets the active roster.
+     * Returns the active roster iff:
+     *      the roster state singleton is not null
+     *      the list of round roster pairs is not empty
+     *      the first round roster pair exists
+     *      the active roster hash is present in the roster map
+     * otherwise returns null.
```
netopyr, line 548: If the roster is valid, most of these conditions appear to be met. Is it necessary to list them here? (same below)

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
+    @Nullable
+    List<RoundRosterPair> getRosterHistory();
 
-Therefore, Provided an existing State, an `Override Roster` and a network upgrade (which could be a simple
-config-only upgrade if a software version upgrade isn't required), the platform will adopt the provided Override Roster.
+    /**
+     * Get the roster based on roster hash
+     *
+     * @param rosterHash The roster hash
+     * @return The roster.
```
netopyr, line 570: I assume this method returns `null` if the roster is not found. We should add this in the JavaDoc.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
+public class WritableRosterStore extends ReadableRosterStore {
 
-#### New Transplant Procedure
+    /**
+     * Sets the candidate roster in state.
+     * Setting the candidate roster indicates that this roster should be adopted as the active roster when required.
+     *
+     * @param candidateRoster a candidate roster to set. It must be a valid roster.
+     */
+    public void putCandidateRoster(@NonNull final Roster candidateRoster);
```
netopyr, line 592: Why `putCandidateRoster()` and not `setCandidateRoster()`?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
 
-The App will decide a network transplant sequence based on the following heuristics:
+    /**
+     * Sets the Active roster.
+     * This will be called to store a new Active Roster in the state.
+     *
+     * @param roster an active roster to set
+     * @param round the round number in which the roster became active.
+     *              It must be a positive number greater than the round number of the current active roster.
```
netopyr, line 600: What happens if it is not greater? Ignored, `Exception`, something else?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -245,291 +518,180 @@ message RoundRosterPair {
...
+     */
+    public void putActiveRoster(@NonNull final Roster roster, final long round);
```
netopyr, line 602: I associate `put` with `Set` and `Map`. Is this intended? Otherwise, I would choose another verb (set, add, ...).

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,374 @@ components within the platform.
...
+4. An `Override Roster` is loaded from disk and unconditionally set as the latest active roster. This is used to
+   transplant the state to a different network.  Until the TSS effort is delivered and the network is given a ledger
+   id, this override mechanism can be used to manually correct the roster / address book on the current network.
+
+### Core Behaviors: Roster Lifecycle
+
+The following diagram depicts the new roster life-cycle where the `Roster History` provided to the platform is
+determined at startup and committed to the merkle state before handing it to the platform.
+
+On software upgrade, during normal execution, if a candidate roster is present in the state, it will be unconditionally
```
MiroslavGatsanoga, line 162: What should be the behavior if the candidate roster is not present in the state?
By "unconditionally" is it meant that we should not check if we can recreate the ledger id from the candidate roster i.e. that has already happened at some earlier moment?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,374 @@ components within the platform.
...
+1. `genesis-config.txt`
+
+Startup Tasks:
+
+1. Read the current AddressBooks from the platform state.
+2. configAddressBook := Read the address book in config.txt
+3. previousRoster := translateToRoster(currentAddressBook)
+4. currentRoster := translateToRoster(configAddressBook)
+5. currentRound := state round +1
+6. set (previousRoster, 0) as the active roster in the roster state.
```
MiroslavGatsanoga, line 245: Why the round number is 0? Shouldn't it be `currentRound - 1` which equals the `state round`?

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Is it not too late in the cycle to make these field name changes?    I would prefer it be called `rosterHistory` as well.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Probably eliminates any special casing to have both instantiated, sure.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: It is removed from the roster, but still needs to exist.  We'll need to keep the public keys in the state as part of the TssBaseService data.   And if we automate it, we'll need a transaction that publishes the public key to update the table.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: I did not write that text.   I would be happy to update it to whatever it needs to be.   I think I copied it from the code base, which has @derektriley as the committer.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: Adding documentation that it returns null.  I won't resolve this comment until I create a PR that corrects the javadoc in the code.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: @derektriley  will have to explain that one. :-). https://github.com/hashgraph/hedera-services/pull/15442/commits/9771b715022758d8fb75ed420dac885f7e9f6191

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: The full answer requires some context.  As part of our initial Dynamic Address Book effort, where we want to support the ability to switch address books (rosters) without a system restart, we planned to use an event's `birthRound` as the means of determining which roster to use to find the public key to validate the signature on the event.  The `birthRound` is the latest consensus round number when the event is created.   The rosters are associated with the round number that it becomes active and governs consensus.  We use the roster with the highest round in becoming active that is below or equal to the event's birth round to validate the event's signature.   

As of right now, we have not completed the switch to using `birthRound` in this process.  The old code doesn't store round numbers and uses the address book by position to determine "current" or "previous".  Right now we only update the roster at software upgrade boundaries.  After a software upgrade, the old events from the previous software version still need to be validated.  We use the previous roster for this.  In this migration, the round numbers do not matter, only the position / order that determined current and previous roster.   Going forward, when rosters become active in the state, they are associated with the round number they become active.  When we switch to fully dynamic address books / dynamic rosters, the round numbers will matter and we will have accurate round numbers there. 

I chose 0 because we don't know the actual number.  While it is somewhat arbitrary since we are selecting it by position, there will be events in the pre-consensus event stream that are older than `currentRound -1`.    We don't know how many rounds back the PCES will have events for.  0 gives a guarantee that it is prior to all possible events.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: This is the TSS-Roster proposal, which should deliver before TSS-Ledger-ID and there is no concept of a ledger id in this proposal.   This was meant to adapt the Dynamic Address Book Phase 2 life-cycle to using rosters.   The config.txt address book from disk is adopted unconditionally at software upgrade boundaries.  

If there is no candidate roster in the state, and its a software upgrade, and there is an active roster in the roster state, then this is an error scenario.  The changes implied by TSS-Roster have not been implemented properly.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
edward-swirldslabs: The API on the roster state only allows setting the "current" active roster.   When two rosters are set, the first roster is rotated out to become the "previous" or second roster.   The instruction here to set (previous, 0) as the active roster is so that it can be pushed back into the second position when we set (currentRoster, currentRound) as the active roster.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
anthony-swirldslabs: > If there is no candidate roster in the state, and its a software upgrade, and there is an active roster in the roster state, then this is an error scenario.

Why does this have to be an error? Why cannot a software upgrade happen w/o any changes to the roster? Why cannot the system simply continue to use the existing active roster in this case?

edward-swirldslabs: Config.txt is written to disk unconditionally.   Similarly a candidate roster should be written to state unconditionally. 

It is easier to debug scenarios when it is known that there will always be a next roster as an artifact than to try to figure out why a next roster was not generated.

anthony-swirldslabs: That seems a bit strict a requirement, but okay.

platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md
```diff
@@ -17,68 +19,374 @@ components within the platform.
...
-  Event Validation etc.) must be adapted to use the new `Roster` object instead.
-- Address book paradigm discontinued within the platform codebase in favor of `Roster`s
+#### Alternatives Considered
 
-## Architecture
+Continuing to use config.txt as the medium for updating the platform `AddressBook` was considered. However, the
+discrepancy in the information present between the `Roster` data structure and the `AddressBook` data structure make
+config.txt a poor medium for carrying out `Roster` updates. Writing the updated roster out to disk is less secure
+than storing it in the state. When TSS is implemented, there will be TSS related key material stored in the state,
+and it is imperative that this key material stay in sync with the `Roster` data. Leaving the content as a modifiable
```
thomas-swirlds-labs, line 117: This text is already exists verbatim in the Purpose and Context section. Consider deduplicating or re-wording?

edward-swirldslabs:
The state of this PR is very stale and like divergent from the current code base after the roster lifecycle was turned over to Services.  I am closing this PR.

PR #16074, opened 2024-10-22
feat: introduce RosterEntry.internal_gossip_endpoint

### LINKED ISSUE #16073: Introduce `internal_gossip_endpoint`
**Background**
The `AddressBook` used to differentiate between `internal` and `external` endpoints and treat them rather differently, e.g. imposing additional conditions on when/how the internal endpoint is used. The new `Roster` paradigm didn't consider all the use-cases and instead introduced a flat, properties-less list of all endpoints, intending to try using all the endpoints one-by-one until one succeeds to connect to a node. This simple approach could cause issues when a private network of a node has a clash of IP addresses with another node's internal endpoint.

**Acceptance criteria**
Introduce `RosterEntry.internal_gossip_endpoint`

anthony-swirldslabs:
**Description**:
The `AddressBook` used to differentiate between `internal` and `external` endpoints and treat them rather differently, e.g. imposing additional conditions on when/how the internal endpoint is used. The new `Roster` paradigm didn't consider all the use-cases and instead introduced a flat, properties-less list of all endpoints, intending to try using all the endpoints one-by-one until one succeeds to connect to a node. This simple approach could cause issues when a private network of a node has a clash of IP addresses with another node's internal endpoint.

In order to avoid the complexity of building additional heuristics to try and "guess" if a given point is internal or external, I'm introducing a dedicated `internal_gossip_endpoint` field in the RosterEntry to clearly separate the internal and external endpoints.

Additionally, I'm completely removing the `NodeInfo.gossipEndpoints()` because it's unused in app/services, and therefore we can avoid adding an extra complexity of trying to reconcile the external and internal endpoints when populating this field.

**Related issue(s)**:

Fixes #16073

**Notes for reviewer**:
Unit tests pass.

**Checklist**


hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
```diff
@@ -101,7 +101,6 @@ public void migrate(@NonNull final MigrationContext ctx) {
                     .accountId(nodeInfo.accountId())
                     // Default node description hard coded to the values used currently
                     .description(formatNodeName(nodeInfo.nodeId()))
-                    .gossipEndpoint(nodeInfo.gossipEndpoints())
```
Neeharika-Sompalli, line 104: Other option would be to add `internal` and `external` gossipEndpoints into `NodeInfo`. Since we discussed `gossipEndpoints ` are not used anywhere other than `AddressBookSchema` I am okay removing them, since by the time we enable DAB, we could iterate through `Roster` instead of `AddressBook`

Neeharika-Sompalli: But Genesis startup will run this migrate 🤔 and in future PRs we also need to sue this in TSS while constructing new candidate Rosters and we will not have enough information with this removed. Is it possible to add both `internal` and `external` gossip endpoints to NodeInfo @anthony-swirldslabs ?

iwsimon: You can not remove gossipEndpoint here. gossipEndpoint is node's important attribute and is used. When DAB enabled, we will still need another migration to get this attribute.

anthony-swirldslabs:
We've discussed this and decided to abandon using internal IP addresses in the Roster, so I'm closing this PR.

PR #16095, opened 2024-10-22
feat: align state and records for self managed contract keys on create

### LINKED ISSUE #16057: Assign self manage contract keys when creating contracts and externalizing results.
Currently, the admin key used for the state and for externalization are not consistent .  Create contracts with a self managed admin key in state and in the externalized record for hollow account flow and non HAPI contract create flows

### LINKED ISSUE #16507: Update slack reporting from XTS to include additional workflow information
The Slack reporting step in `zxcron-extended-test-suite.yaml` should be updated to include the below additional information on failure:

- workflow run id
- workflow run URL
- commit author

lukelee-sl:
**Description**:
Update hollow account completed as contract flow and non HAPI contract create so that the admin key will be a contract key with the newly created account number to imply self management in both the saved state and the externalized record.

**Related issue(s)**:

Fixes #16057 

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/scope/HandleHederaOperationsTest.java
```diff
@@ -417,7 +417,9 @@ void createContractWithSelfAdminParentDispatchesAsExpectedThenMarksCreated() thr
         final var pendingId = ContractID.newBuilder().contractNum(666L).build();
         final var synthContractCreation = synthContractCreationFromParent(pendingId, parent)
                 .copyBuilder()
-                .adminKey((Key) null)
+                .adminKey(Key.newBuilder()
```
david-bakin-sl, line 420: Sad that this is the only unit test affected by this change, and that only incidently.

**RANT ON**

The base class `HandleHederaOperations` is basically untestable via unit tests - notwithstanding that this test file has 30 separate tests. Let me give you an example - here's a call chain:

* `public createContract` -> **`private`** `dispatchAndMarkCreation` -> **`private`** `contractBodyCustomizerFor` -> **`private`** `standardized` -> **`private`** `needsStandardization`

10 separate conditional choices are in that path (`if` statements, two fluent `fooOrElse()` methods, a trinary op, and a boolean expression with 3 parts separated by `||`).

Overuse of `private` probably indicates some design-in-the-local problem in this class, huh?

**RANT OFF**

(This isn't a _Hedera_ problem.  It's an industry problem.  I mean: I've seen it elsewhere.  A lot.)

hedera-node/hedera-token-service-impl/src/test/java/com/hedera/node/app/service/token/impl/test/api/TokenServiceApiImplTest.java
```diff
@@ -203,7 +203,7 @@ void finalizesHollowAccountAsContractAsExpected() {
         assertEquals(1, accountStore.sizeOfAccountState());
         final var finalizedAccount = accountStore.getContractById(CONTRACT_ID_BY_NUM);
         assertNotNull(finalizedAccount);
-        assertEquals(STANDIN_CONTRACT_KEY, finalizedAccount.key());
+        assertEquals(Key.newBuilder().contractID(CONTRACT_ID_BY_NUM).build(), finalizedAccount.key());
```
david-bakin-sl, line 206: Test here is missing a check that `maxAutoAssociations` is set correctly.  (I mean, missing this check independent of your change, which is fine.)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
```diff
@@ -151,7 +153,8 @@ final Stream<DynamicTest> hollowAccountCompletionWithEthereumTransaction() {
                             .via(TRANSFER_TXN_2);
 
                     final var op3 = getAliasedAccountInfo(evmAddress)
-                            .has(accountWith().key(SECP_256K1_SOURCE_KEY).noAlias());
+                            .has(accountWith().key(SECP_256K1_SOURCE_KEY).noAlias())
+                            .logged();
```
david-bakin-sl, line 157: (what does this additional `.logged()` check?  was it just omitted before or is there something new to check?)

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
```diff
@@ -557,6 +560,12 @@ final Stream<DynamicTest> hollowAccountCompletionWithEthereumContractCreate() {
                             getTxnRecord(TRANSFER_TXN_2).andAllChildRecords().logged();
 
                     allRunFor(spec, op2, op3, hapiGetSecondTxnRecord);
+                    // ensure that the finalized contract has a self management key
```
david-bakin-sl, line 563: Does a check of this kind need to be made above in `hollowAccountCreationWithContractCreate` - it's a HAPI contract create but an admin key is specified - or do just assume that works as it is tested elsewhere?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
```diff
@@ -481,8 +471,7 @@ private ContractCreateTransactionBody standardized(
         }
     }
 
-    private boolean needsStandardization(
-            @NonNull final ContractCreateTransactionBody op, @Nullable final Key standardAdminKey) {
-        return op.hasInitcode() || op.gas() > 0L || op.initialBalance() > 0L || standardAdminKey != op.adminKey();
+    private boolean needsStandardization(@NonNull final ContractCreateTransactionBody op) {
```
david-bakin-sl, line 474: Now that the argument `standardAdminKey` is gone - isn't clear to me what `needsStandardization` means.  Perhaps change the name of the method or add javadoc?

david-bakin-sl (commented):
Some comments.  The "meat" of this PR looks good.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
lukelee-sl: `hollowAccountCreationWithContractCreate` does not finalize the hollow account through the same flow that results in a contract admin key

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
lukelee-sl: removed.  just wanted to make sure that the account had the right key.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/crypto/HollowAccountFinalizationSuite.java
david-bakin-sl: ok, so since not hollow account finalization then not a proper subject for this Suite

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaOperations.java
lukelee-sl: added javadoc

david-bakin-sl (approved):
LGTM

hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/api/TokenServiceApiImpl.java
```diff
@@ -178,7 +176,9 @@ public void finalizeHollowAccountAsContract(@NonNull final AccountID hollowAccou
         }
         final var accountAsContract = hollowAccount
                 .copyBuilder()
-                .key(STANDIN_CONTRACT_KEY)
+                .key(Key.newBuilder()
+                        .contractID(ContractID.newBuilder().contractNum(hollowAccountId.accountNumOrElse(0L)))
```
tinker-michaelj, line 180: It seems like this should be,
```suggestion
                        .contractID(ContractID.newBuilder().contractNum(hollowAccountId.accountNumOrThrow()))
```
because we only use this [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/HandleHederaNativeOperations.java#L138) and it explicitly converts to a `0.0.X` id,
```
final var hollowAccountId = requireNonNull(accountStore.getAccountIDByAlias(evmAddress));
```

tinker-michaelj (dismissed):
One nit to `OrThrow()` instead of `OrElse()` where we should have an enforceable invariant, but LGTM, tygm @lukelee-sl !

tinker-michaelj (approved):
LGTM, ty @lukelee-sl !

david-bakin-sl (approved):
LGTM

PR #16099, opened 2024-10-22
chore: Updates to match recent changes to the Block Stream definition

jsync-swirlds:
* Updated the block service definitions
   * Split one monolithic service into a few modular services
   * Modified the _publish_ service to match agreed protocol.
   * Changed the publish call to send repeated block items,
     rather than one-at-a-time.
* Merged a few other small changes

> Note.
>> This change includes a small section in the Block Node API definition that is
   removed (due to multiple PBJ bugs).  This will be restored once the relevant
   bug (`#280`) in the PBJ library is fixed.

### DIFF
--- a/hapi/hedera-protobufs/block/stream/output/consensus_service.proto
+++ b/hapi/hedera-protobufs/block/stream/output/consensus_service.proto
@@ -6,7 +6,11 @@
  * ### Topic Running Hash Calculation
  * Submitted messages include a topic running hash. This value has changed
  * over time, with the known versions detailed in the `RunningHashVersion`
- * enumeration.
+ * enumeration. The running hash version SHALL NOT be transmitted in the
+ * Block Stream, however, as it is a fixed value that can only be changed
+ * with a new release of consensus node software following a HIP to update
+ * the specification, so repeating that fixed value in the block stream is
+ * wasteful and unnecessary.
  *
  * ### Keywords
  * The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
@@ -185,6 +185,10 @@ public class BlockStreamBuilder
     @Nullable
     private HederaFunctionality functionality;
     /**
+     * The exchange rate set to add to the record stream transaction receipt.
+     */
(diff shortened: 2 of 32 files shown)


mxtartaglia-sl (approved):

timo0 (approved):

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
```diff
@@ -6,7 +6,11 @@
  * ### Topic Running Hash Calculation
  * Submitted messages include a topic running hash. This value has changed
  * over time, with the known versions detailed in the `RunningHashVersion`
- * enumeration.
+ * enumeration. The running hash version SHALL NOT be transmitted in the
+ * Block Stream, however, as it is a fixed value that can only be changed
+ * with a new release of consensus node software following a HIP to update
+ * the specification, so repeating that fixed value in the block stream is
+ * wasteful and unnecessary.
```
mhess-swl, line 13: Does the change of not providing this running hash version itself require a hip?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamBuilder.java
```diff
@@ -840,7 +844,10 @@ public BlockStreamBuilder contractID(@Nullable final ContractID contractID) {
     @NonNull
     @Override
     public BlockStreamBuilder exchangeRate(@Nullable final ExchangeRateSet exchangeRate) {
-        transactionResultBuilder.exchangeRate(exchangeRate);
+        // Block Stream doesn't include exchange rate in output (it's in state
+        // changes when it is updated), so store exchange rate for the
+        // translation context
```
mhess-swl, line 849: 😳 How have our block <-> record comparisons been passing if the exchange rate hasn't been included?

jsync-swirlds: The exchange rate _was_ included in the block stream, but is removed in the latest version (as presented in this PR), as it's unnecessary extra data once block nodes exist (it can be provided from state changes much more efficiently).

hapi/hedera-protobufs/block/stream/output/consensus_service.proto
jsync-swirlds: It has to be detailed in the Block Stream HIP (and is).  The "change" is not a change so much as the running hash version was (deliberately) never included in the Block Stream.

mhess-swl (approved):

PR #16130, opened 2024-10-23
feat: throttled tx metrics

### LINKED ISSUE #16035: Metric for tracking throttled transactions
Create metrics (`Counter`) for each transaction type and query type that track how often transactions of that type were throttled.

kimbor:
**Description**:
This PR creates a new set of metrics to track when transactions and queries are throttled.

`Counter` is used as the metric type. This type of metric will only ever increase. Calculation will be done on the Grafana side to translate this number into events per second.

**Related issue(s)**:
Fixes #16035


hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/OpWorkflowMetrics.java
```diff
@@ -77,15 +83,20 @@ public OpWorkflowMetrics(@NonNull final Metrics metrics, @NonNull final ConfigPr
                     .withDescription("The average duration of a " + name + " transaction in nanoseconds")
                     .withUnit("ns");
             final var avgMetric = metrics.getOrCreate(avgConfig);
-            transactionMetrics.put(functionality, new TransactionMetric(maxMetric, avgMetric));
+            transactionDurationMetrics.put(functionality, new TransactionMetric(maxMetric, avgMetric));
+
+            // initialize the transaction throttle metrics
+            final var throttledConfig = new Counter.Config("app", name + "ThrottledTps")
+                    .withDescription("The number of " + name + " transactions that failed due to throttling");
```
mhess-swl, line 90: Just a suggestion for precision
```suggestion
            final var throttledConfig = new Counter.Config("app", name + "ThrottledTxns")
                    .withDescription("The number of " + name + " transactions that were throttled");
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
```diff
@@ -193,10 +198,9 @@ public TransactionInfo runAllChecks(
         // 4. Check throttles
         assertThrottlingPreconditions(txInfo, configuration);
         final var hederaConfig = configuration.getConfigData(HederaConfig.class);
-        if (hederaConfig.ingestThrottleEnabled()) {
-            if (synchronizedThrottleAccumulator.shouldThrottle(txInfo, state)) {
-                throw new PreCheckException(BUSY);
-            }
+        if (hederaConfig.ingestThrottleEnabled() && synchronizedThrottleAccumulator.shouldThrottle(txInfo, state)) {
+            workflowMetrics.incrementThrottled(functionality);
+            throw new PreCheckException(BUSY);
```
mhess-swl, line 203: Is it important to differentiate between 'frontend' and 'backend' throttles, or is it sufficient for the requirements to use the same metric for both?

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/DispatchProcessorTest.java
```diff
@@ -414,6 +437,7 @@ void consGasExhaustedWaivesServiceFee() throws ThrottleException {
         verify(recordBuilder).status(CONSENSUS_GAS_EXHAUSTED);
         verify(feeAccumulator).chargeFees(PAYER_ACCOUNT_ID, CREATOR_ACCOUNT_ID, FEES);
         verify(feeAccumulator).chargeFees(PAYER_ACCOUNT_ID, CREATOR_ACCOUNT_ID, FEES.withoutServiceComponent());
+        verify(opWorkflowMetrics).incrementThrottled(CONTRACT_CALL);
```
mhess-swl, line 440: This test and `consGasExhaustedForEthTxnDoesExtraWork()` are the only tests I see in here that are actually testing `incrementThrottled`, and they're both contract-related. It would be nice to see other more 'native' transaction types also work. The majority of the lines added to this file read 
```
verify(opWorkflowMetrics, never()).incrementThrottled(any());
```
i.e. verifying the throttle _isn't_ incremented, which is important, but may not be sufficient.

Likewise for the IngestCheckerTest, it would be nice to see more assertions about actually incrementing the throttle appropriately.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/OpWorkflowMetrics.java
kimbor: I agree with the "ThrottledTps" -> "ThrottledTxns" change, thank you.
For the "transactions that failed due to throttling" -> "transactions that were throttled" change, I disagree. I think that "transactions that were throttled" sounds like we're counting the txs that are subject to throttling (i.e. all the transactions received that don't fail due to ingest checks), not just the txs that we rejected due to throttle limits.

kimbor: I updated this, see if you like my wording @mhess-swl

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/ingest/IngestChecker.java
kimbor: Requirements do not include differentiating between frontend and backend throttles

hedera-node/hedera-app/src/test/java/com/hedera/node/app/workflows/handle/DispatchProcessorTest.java
kimbor: I added a new test in DispatchProcessorTest for a native transaction type.
The other files already have 100% test coverage on the patch. I think there could be better test coverage of throttle logic esp in OpWorkflowMetricsTest. But, that is outside the scope of this PR.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/OpWorkflowMetrics.java
mhess-swl: Looks great, thanks!

mhess-swl (dismissed):
LGTM! Tyvm @kimbor, I've wanted more insight into this for awhile! 🙏

mhess-swl (approved):

Issue #14748, opened 2024-08-09
implement `setCandidateRoster` method call and update state


edward-swirldslabs:
Invoke the `setCandidateRoster` method on the service interface and implement the back end with the `TssStateManager` modifying the state and then passing forward the candidate roster to the `TssCryptographyManager`. 

This full business logic is defined [here](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/TSS-Ledger-Id/TSS-Ledger-Id.md?plain=1#L899) in the TSS-Ledger proposal.

1️⃣ **Construct and accept/ignore the new candidate roster**
 - [x] Add `tssEncryptionKey` to `Node{Create,Update}TransactionBody`.
 - [x] Add `tssEncryptionKey` to `Node` state object.
 - [x] Iterate through the `AddressBookService`'s `Node` K/V state to construct a _new_ candidate `Roster`.
 - [x] Compare the hash to the existing active and candidate rosters; abort if it is identical to either.
 - [x] If it is not identical, set as the `RosterState#candidate_roster_hash` via a `WritableRosterStore`.

2️⃣ **Compute the share counts per node for the candidate roster**
EDIT: This is already done in `TssMessageHandler`, since this logic is needed there
 - [x] Run [this algorithm](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/TSS-Ledger-Id/TSS-Ledger-Id.md?plain=1#L280) from the TSS Ledger ID proposal using the candidate roster's weights.

PR #16131, opened 2024-10-23
feat: Set TSS candidate roster logic

mhess-swl:
This PR includes the following:

* Addition of the `tss_encryption_key` field to the `Node`, `NodeCreate`, and `NodeUpdate` protobuf messages
* A first pass at implementing `setCandidateRoster(...)` as detailed in point 1 in #14748
* The wiring to appropriately call `TssService#setCandidateRoster` from the daily staking update

Part of #14748


hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
```diff
@@ -77,4 +85,21 @@ protected <T extends ReadableKVState<EntityNumber, Node>> T nodesState() {
     public Iterator<EntityNumber> keys() {
         return nodesState().keys();
     }
+
+    private Roster constructFromNodesState(@NonNull final ReadableKVState<EntityNumber, Node> nodesState) {
+        final var rosterEntries = new ArrayList<RosterEntry>();
+        for (final Iterator<EntityNumber> it = nodesState.keys(); it.hasNext(); ) {
+            final var nodeNumber = it.next();
+            final var nodeDetail = nodesState.get(nodeNumber);
+            final var entry = RosterEntry.newBuilder()
```
jsync-swirlds, line 94: Do we need a test for `if(!nodeDetail.isDeleted())` here so we don't include deleted nodes in the roster?

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
```diff
@@ -77,4 +85,21 @@ protected <T extends ReadableKVState<EntityNumber, Node>> T nodesState() {
...
+                    .nodeId(nodeDetail.nodeId())
+                    .weight(nodeDetail.weight())
+                    .gossipCaCertificate(nodeDetail.gossipCaCertificate())
+                    .gossipEndpoint(nodeDetail.serviceEndpoint())
```
Neeharika-Sompalli, line 98: Should be `nodeDetail.gossipEndpoint` here

jsync-swirlds (commented):
Proto files look good.
One question on the candidate roster generation method.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -114,10 +115,14 @@ public void bootstrapLedgerId(
...
 
     @Override
-    public void setCandidateRoster(@NonNull final Roster roster, @NonNull final HandleContext context) {
-        requireNonNull(roster);
-        // (TSS-FUTURE) https://github.com/hashgraph/hedera-services/issues/14748
-        tssSubmissions.submitTssMessage(TssMessageTransactionBody.DEFAULT, context);
+    public void setCandidateRoster(@NonNull final Roster newCandidate, @NonNull final HandleContext context) {
+        requireNonNull(newCandidate);
+
+        final var rosterStore = context.storeFactory().writableStore(WritableRosterStore.class);
```
Neeharika-Sompalli, line 121: While testing, I found that this place will be called by `NodeStakeUpdate` and the `writableStoreFactory` here will be from `TokenService`. So here, we cant use the same context to construct rosterStore. 
We can move thos logic to `nodeStakeUpdates` and do like this. 
```
final var rosterStore = writableStoreFactory.getStore(WritableRosterStore.class);
        if (!Objects.equals(newCandidateRoster, rosterStore.getCandidateRoster())
                && !Objects.equals(newCandidateRoster, rosterStore.getActiveRoster())) {
            rosterStore.putCandidateRoster(newCandidateRoster);
        }
```

Neeharika-Sompalli (dismissed):
Have two comments. These issues I found while testing to enable the repeatable tests. Otherwise LGTM ! Thanks @mhess-swl

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
mhess-swl: Makes sense to me. I'll ask to be sure

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
mhess-swl: Done

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
mhess-swl: Thanks! fixed

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
mhess-swl: Yes we want to check for deleted nodes, and not include them. Fixed.

jsync-swirlds (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -146,27 +148,34 @@ public void setCandidateRoster(@NonNull final Roster roster, @NonNull final Hand
...
+                            .sourceRosterHash(activeRosterHash)
+                            .targetRosterHash(candidateRosterHash)
+                            .shareIndex(shareIndex.getAndAdd(1))
+                            .tssMessage(Bytes.wrap(msg.bytes()))
+                            .build();
+                    tssSubmissions.submitTssMessage(tssMessage, context);
+                });
+            }
+        } else {
+
```
Neeharika-Sompalli, line 178: Why is this else empty ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -146,27 +148,34 @@ public void setCandidateRoster(@NonNull final Roster roster, @NonNull final Hand
...
-                final var tssMessage = TssMessageTransactionBody.newBuilder()
-                        .sourceRosterHash(activeRosterHash)
-                        .targetRosterHash(candidateRosterHash)
-                        .shareIndex(shareIndex.getAndAdd(1))
-                        .tssMessage(Bytes.wrap(msg.bytes()))
-                        .build();
-                tssSubmissions.submitTssMessage(tssMessage, context);
-            });
+        // https://github.com/hashgraph/hedera-services/issues/16166
+        final boolean hasKeyMaterial = false;
```
Neeharika-Sompalli, line 152: I thought we are opening this PR only for happy path. This is always true right ?

Neeharika-Sompalli (dismissed):
Have two comments otherwise LGTM!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
mhess-swl: My mistake, this should be `true` for the happy path.

I asked yesterday, this is _not_ always true. Key material can be missing from an _active_ roster IFF there is no existing ledger Id, i.e. in the genesis or new network scenarios. So it's _almost_ always true.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
mhess-swl: I forgot to add a link to the issue that will implement this, thanks!

mhess-swl: I ended up getting rid of the if/else; it's more confusing than helpful. We can implement it in another issue, and just focus on happy path right now

Neeharika-Sompalli (approved):
LGTM! Thanks @mhess-swl

hapi/hedera-protobufs/services/node_create.proto
```diff
@@ -137,4 +137,27 @@ message NodeCreateTransactionBody {
...
+     * if the Ethereum ecosystem creates precompiles for BLS12_381,
+     * we may switch to that curve.<br/>
+     * This value SHALL be specified according to EIP-196 and EIP-197 standards.
+     * See [EIP-196](https://eips.ethereum.org/EIPS/eip-196#encoding) and
+     * (EIP-197](https://eips.ethereum.org/EIPS/eip-197#encoding)<br/>
+     * This field is OPTIONAL (that is, it can initially be null),
+     * but once set, it MUST NOT be null.<br/>
+     * If this field is set:
+     * <ul>
+     *   <li>This field MUST contain the bytes of a standard ALT_BN128 key value.</li>
```
mxtartaglia-sl, line 156: @mhess-swl If we are not doing any translation between what the library produces and the bytes stored here, this comment is not accurate. 
Internally, we decided not to store them in the same format as Ethereum does.
We had an extensive discussion about this topic with @edward-swirldslabs , @Nana-EC , @david-bakin-sl, @AlfredoG87 and @tinker-michaelj where we came to the conclusion that the internal storage and how it will be exposed to smart contracts can differ.
A brief explanation of the format is exposed in the tss-library proposal.
By the way, there is no `standard ALT_BN128 key value.` each library seems to be defining its own.

PR #16133, opened 2024-10-24
fix: LegacyConfigPropertiesLoader should not suppress ParseException

### LINKED ISSUE #16132: LegacyConfigPropertiesLoader should not suppress ParseException
#### Problem
1. Currently LegacyConfigPropertiesLoader suppress `ParseException` when it fails to parse an `address` line in the config.txt: https://github.com/hashgraph/hedera-services/blob/b44714686870de6909c4103729368db735b14316/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java#L125

The consequence of this is that the node starts with incomplete address-book and does not load all the keys of the nodes. Therefore it rejects other nodes that try to communicate with it and fails to gossip with other nodes. 

2. It logs an incorrect error message when it fails to parse the address for incorrect internal/external address: `***** 'address' needs a minimum of 7 parameters *****`.
3. The error logging does not work since it appears logger is not initialized: https://github.com/hashgraph/hedera-services/blob/b44714686870de6909c4103729368db735b14316/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java#L118

#### Proposal
1. `LegacyConfigPropertiesLoader` should throw the `ConfigurationException` wrapping the `ParseException` when it fails to parse an `address` line
2. The error message needs to be more accurate and should include the `address` line info for clarity
3. It should not attempt to resolve the FQDN internal/external addresses and just start the node. This ensures the node starts with all the keys and will be able to accept connection from other nodes. Otherwise, it may block nodes to start even if one single node's FQDN is not resolvable.

### Steps to reproduce

1. Provide an invalid `internalAddress` like `INVALID.DOMAIN` in one of the `address` line in the config.txt
2. Start the platform software
4. Notice a message (although incorrect): `***** 'address' needs a minimum of 7 parameters *****`
5. Notice that despite this error, it starts the node
6. This node would then fail to communicate with the node that corresponds to that invalid `address` when the other node is up and tries to communicate with this one

### Additional context

I noticed it while developing `solo-operator` where nodes uses FQDN for internal and external addresses and the platform fails to lookup the IP corresponding to the FQDN address because of DNS lookup delay in a kubernetes platform. 

A sample config.txt that the network was using is as below:

```
swirld, 298
app, HederaNode.jar
address, 0, node0, node0, 1, node0.consensus, 50111, node0-svc, 50111, 0.0.3
address, 1, node1, node1, 1, node1.consensus, 50111, node1-svc, 50111, 0.0.4
address, 2, node2, node2, 1, node2.consensus, 50111, node2-svc, 50111, 0.0.5
```

### Hedera network

other

### Version

v0.56.0-develop.x4b7c054

### Operating system

Linux

leninmehedy:
**Description**:

**Related issue(s)**:

Fixes #16132 
Fixes https://github.com/hashgraph/solo-operator/issues/233

**Notes for reviewer**:

During boot time, platform shouldn't try to resolve the FQDN addresses (internal or external) in the config.txt as other nodes may not be up by the time a node starts parsing config.txt. So this PR removes FQDN resolution at config parsing time.

Also it throws exception if `config.txt` parsing fails for other reasons at the boot time.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/legacy/LegacyConfigPropertiesLoader.java
@@ -56,7 +56,7 @@ public final class LegacyConfigPropertiesLoader {
     public static final String ERROR_MORE_THAN_ONE_APP =
             "config.txt had more than one line starting with 'app'. All but the last will be ignored.";
     public static final String ERROR_NO_PARAMETER = "%s needs a parameter";
-    public static final String ERROR_ADDRESS_NOT_ENOUGH_PARAMETERS = "'address' needs a minimum of 7 parameters";
+    public static final String ERROR_ADDRESS_COULD_NOT_BE_PARSED = "'address' could not be parsed";
     public static final String ERROR_PROPERTY_NOT_KNOWN =
             "'%s' in config.txt isn't a recognized first parameter for a line";
     public static final String ERROR_NEXT_NODE_NOT_GREATER_THAN_HIGHEST_ADDRESS =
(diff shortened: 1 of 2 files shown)


edward-swirldslabs (approved):

litt3 (commented):
`NetworkUtils` has no remaining usages, and may be deleted. Otherwise LGTM

leninmehedy:
Just holding of merging it since this PR may not be sufficient to fix all the issues. I have documented my observations, current workarounds and related problems here for reference and offline discussions: https://www.notion.so/swirldslabs/Problem-DNS-lookup-issue-during-Platform-boot-1297c9ab259180fdbc2fe97019d7d7fd?pvs=4

We'll come back to it once I get confirmations from others.

lpetrovic05 (approved):

PR #16077, opened 2024-10-22
feat: SCSC metrics for #TXs that fail pureCheck

david-bakin-sl:
**Description**:

SCSC metrics for #TXs that fail pureCheck:
- counts for CONTRACT_CREATE / CONTRACT_CALL / ETHEREUM_TRANSACTION that fail pureCheck for any reason
- and for those that fail pureCheck because they don't even have the intrinsic gas amount
- and for Type 3 format (BLOB) Ethereum transactions (that we don't support)

**Related issue(s)**:

Fixes #16076 

**Notes for reviewer**:

- Key class is `ContractMetrics` to encapsulate all SCSC metrics; and currently it is used in `ContractCreateHandler`, `ContractCallHandler`, and `EthereumTransactionHandler`
- Added a `Supplier<Metrics>` to the `AppContext` so smart contract service (and others) can get the metrics instance from the app.
  - There are quite a few unit tests that needed to be modified for this change as they create an `AppContext` out of thin air
  - This may not be necessary.  I was unable to _inject_ a `Metrics` object where needed but apparently token service can in `CryptoGetAccountBalanceHandler` - I need to find out how
  - Plus it is somewhat awkward as the `AppContext` is created _before_ the `Metrics` instance is gotten from the platform
  - And this is why I have a locked "once only" method to initialize `ContractMetrics` - I'll be asking people about how to do that better as well

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #16137, opened 2024-10-24
feat: Create tss metrics

### LINKED ISSUE #14760: Create Metrics for TssMessageTransaction, TssVoteTransaction, and any others required
Create the indicated TSS metrics and a TSS dashboard that shows: 
1. TSS Candidate Roster Lifecycle
1. TssMessageTransaction Count / Status
1. TssVoteTransaction Count / Status
1. Aggregation Calculation Time

petreze:
**Description**:
Introduce `TssMetrics` class to be injected in the `TssBaseService` to support metrics for some of the TSS functionalities.
The final version of the PR aims to provide metrics on the following:

- TSS Candidate Roster Lifecycle
- `TssMessageTransaction` count per candidate roster
- `TssVoteTransaction` count per candidate roster
- Aggregation Calculation Time (the time it takes to compute shares from the key material)

**Related issue(s)**:

Fixes #14760 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/ReadableRosterStore.java
```diff
@@ -0,0 +1,49 @@
...
+package com.hedera.node.app.roster;
+
+import com.hedera.hapi.node.state.roster.Roster;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+@Deprecated
+// This is just added to unblock services PRs until this PR https://github.com/hashgraph/hedera-services/pull/15442
+// is merged
+public interface ReadableRosterStore {
```
mhess-swl, line 26: This is in `develop` now, so shouldn't be needed

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssMetrics.java
```diff
@@ -0,0 +1,104 @@
...
+
+        if (functionality == TSS_MESSAGE) {
+            tssMessageTxCounter.increment();
+        } else if (functionality == TSS_VOTE) {
+            tssVoteTxCounter.increment();
+        }
+    }
+
+    /**
+     * Track when the candidate roster is set/
```
mhess-swl, line 88:
```suggestion
     * Track when the candidate roster is set
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssMetrics.java
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @param lifecycle the time at which the candidate roster is set
+     */
+    public void updateCandidateRosterLifecycle(long lifecycle) {
```
mhess-swl, line 92: Nit: `final long lifecycle`

Is it worth checking that this value is > 0? (in this and the next method)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/NodeStakeUpdates.java
```diff
@@ -173,14 +175,17 @@ public static boolean isNextStakingPeriod(
...
-            @NonNull final HandleContext handleContext, @NonNull final WritableRosterStore rosterStore) {
+            @NonNull final HandleContext handleContext,
+            @NonNull final WritableRosterStore rosterStore,
+            @NonNull final TssMetrics tssMetrics) {
         final var nodeStore = handleContext.storeFactory().readableStore(ReadableNodeStore.class);
         final var newCandidateRoster = nodeStore.newRosterFromNodes();
 
         if (!Objects.equals(newCandidateRoster, rosterStore.getCandidateRoster())
                 && !Objects.equals(newCandidateRoster, rosterStore.getActiveRoster())) {
+            tssMetrics.updateCandidateRosterLifecycle(Instant.now().getEpochSecond());
```
mhess-swl, line 186: We should use `InstantSource` for the timestamp here, since we have EmbeddedHedera in the test framework now

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/NodeStakeUpdates.java
```diff
@@ -173,14 +175,17 @@ public static boolean isNextStakingPeriod(
     }
 
     private void keyNewRoster(
-            @NonNull final HandleContext handleContext, @NonNull final WritableRosterStore rosterStore) {
+            @NonNull final HandleContext handleContext,
+            @NonNull final WritableRosterStore rosterStore,
+            @NonNull final TssMetrics tssMetrics) {
```
mhess-swl, line 180: Does it make sense to pass `TssMetrics` into `TssBaseService`, so we don't have to pass it through here? 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -143,32 +146,45 @@ public void setCandidateRoster(@NonNull final Roster roster, @NonNull final Hand
                 context.storeFactory().readableStore(ReadableRosterStore.class).getActiveRoster();
         final var activeRosterHash = RosterUtils.hash(sourceRoster).getBytes();
         final var candidateRosterHash = RosterUtils.hash(roster).getBytes();
+
+        final var aggregationCalculationStart = Instant.now();
```
mhess-swl, line 150: `InstantSource` instead of `Instant.now()` in this class too

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -143,32 +146,45 @@ public void setCandidateRoster(@NonNull final Roster roster, @NonNull final Hand
...
         final var candidateRosterParticipantDirectory = computeTssParticipantDirectory(roster, maxSharesPerNode, (int)
                 context.networkInfo().selfNodeInfo().nodeId());
+        final var aggregationCalculationEnd = Instant.now();
+        final var durationOfAggregation = Duration.between(aggregationCalculationStart, aggregationCalculationEnd)
+                .toSeconds();
+        tssMetrics.updateAggregationTime(durationOfAggregation);
+
+        // https://github.com/hashgraph/hedera-services/issues/16166
+        final boolean hasKeyMaterial = false;
+        if (!hasKeyMaterial) {
```
mhess-swl, line 162: This whole `if` block is a little outdated. Check `develop` for the most recent

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
```diff
@@ -143,32 +146,45 @@ public void setCandidateRoster(@NonNull final Roster roster, @NonNull final Hand
...
         final var activeRosterHash = RosterUtils.hash(sourceRoster).getBytes();
         final var candidateRosterHash = RosterUtils.hash(roster).getBytes();
+
+        final var aggregationCalculationStart = Instant.now();
         final var tssPrivateShares =
                 getTssPrivateShares(sourceRoster, maxSharesPerNode, tssStore, candidateRosterHash, context);
         final var candidateRosterParticipantDirectory = computeTssParticipantDirectory(roster, maxSharesPerNode, (int)
                 context.networkInfo().selfNodeInfo().nodeId());
+        final var aggregationCalculationEnd = Instant.now();
+        final var durationOfAggregation = Duration.between(aggregationCalculationStart, aggregationCalculationEnd)
```
mhess-swl, line 156: @edward-swirldslabs is this the aggregation you want to see? Only the time spent in `getTssPrivateShares` and `computeTssParticipantDirectory`? Or should we also be including the time it takes to generate or process the messages?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -353,6 +359,9 @@ private void handlePlatformTransaction(
             handleOutput.blockRecordSourceOrThrow().forEachItem(blockStreamManager::writeItem);
         }
         opWorkflowMetrics.updateDuration(userTxn.functionality(), (int) (System.nanoTime() - handleStart));
+        if (userTxn.functionality() == TSS_MESSAGE || userTxn.functionality() == TSS_VOTE) {
+            tssMetrics.updateTssTransactionMetrics(userTxn.functionality());
```
mhess-swl, line 363: Actually, I was just contemplating a mechanism for service-specific metrics (see PR #16077, which is injecting metrics for contract service). I'm curious if we can get something like that to work instead, so we don't have to have this in `HandleWorkflow`. It's just an idea at this point, though

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
petreze: Oh, good catch. It's a leftover after fixing conflict, ty

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/ReadableRosterStore.java
petreze: Leftover from fixing conflicts, thanks for noticing!

edward-swirldslabs (changes requested):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java
petreze: Got it, thanks a lot!

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/NodeStakeUpdates.java
petreze: I've updated the branch in that way you can have a look @mhess-swl

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * A class to track all the metrics related to TSS functionalities.
+ */
+@Singleton
+public class TssMetrics {
+    private static final String TSS_CANDIDATE_ROSTER_LIFECYCLE = "tss_candidate_roster_lifecycle";
+    private static final String TSS_CANDIDATE_ROSTER_LIFECYCLE_DESC = "the lifecycle of the current candidate roster";
```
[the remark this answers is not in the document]
petreze: I'm not entirely sure on this one - if the start or end time of the target roster's lifecycle can be measured within `TssCryptographyManager`. I think the start of the candidate roster's lifecycle might be introduced in [setCandidateRoster()](https://github.com/hashgraph/hedera-services/blob/14760-add-tss-metrics/hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseServiceImpl.java#L149). On the end time, if I understood it correctly, the voting is closed, when the threshold for this particular candidate roster is already met and we do not accept further votes - my thought process led me to `handle()` in [TssVoteHandler](https://github.com/hashgraph/hedera-services/blob/14760-add-tss-metrics/hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/handlers/TssVoteHandler.java#L83). Any thoughts? @edward-swirldslabs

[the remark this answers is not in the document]
edward-swirldslabs: That would be fine.  My feedback here was in reference to the design I wrote, but the implementation is organizing things differently than the design.   Agreed the start time is in the same thread of execution that `setCandidateRoster` is called and the end time is determined on the `TssVoteHandler` thread when it closes voting.

edward-swirldslabs (dismissed):
The issues I raised were addressed, thank you.  :-)

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssCryptographyManager.java
```diff
@@ -129,6 +139,14 @@ private CompletableFuture<LedgerIdWithSignature> computeAndSignLedgerIdIfApplica
                     final var signature = gossip.sign(ledgerId.publicKey().toBytes());
 
                     final BitSet tssVoteBitSet = computeTssVoteBitSet(validTssOps);
+
+                    // calculate the time it took to aggregate and compute ledger id, public shares and private shares
```
edward-swirldslabs, line 143: The private shares are not being computed here, we can drop it from the comment.   Measuring the other two aggregation operations are enough to give a feel for how expensive aggregation is.

petreze: Good catch, a bit misleading! The comment is updated, ty!

edward-swirldslabs (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
mhess-swl: We won't get to implementing anything like what I had in mind for awhile, so we shouldn't wait for it

mhess-swl (dismissed):

kimbor (dismissed):

mhess-swl (approved):

edward-swirldslabs (approved):

PR #16053, opened 2024-10-19
chore: standardize and simplify `ScheduleService`

tinker-michaelj:
**Description**:
 - Closes #16023 
 - Closes #16044 
 - Closes #14125
 - Greatly streamlines the `ScheduleService` business logic by simply updating `signatories` with relevant signers from the set of cryptographic keys with valid signatures returned by `KeyVerifier#signingCryptoKeys()`. That is, a cryptographic key `K` is added to a schedule's signatories iff,
     1. `K` had a valid signature on the `ScheduleCreate` or `ScheduleSign` transaction.
     2. `K` was not already in the set of signatories.
     3. `K` was a constituent of a key required to sign the schedule's transaction at the time the `ScheduleCreate` or `ScheduleSign` reached consensus.
 - Changes the coding style in the `ScheduleService` to be more consistent with the rest of the application. For example,
   - Uses `validateTrue(condition, STATUS)` instead of `if (!condition) { throw new HandleExeption(STATUS); }`
   - Uses `var` instead of redundant type declarations.
   - Does not duplicate invariant checks (especially, null checks) that are already guaranteed by workflows.

PR #16144, opened 2024-10-24
feat: Long-term scheduled transactions State API

### LINKED ISSUE #16059: Define Schema for block stream-friendly state organization

### LINKED ISSUE #16068: Add migrations from existing state

### LINKED ISSUE #16069: Update readable/writable stores to use new states

iwsimon:
**Description**:

Define Schema for block stream-friendly state organization.
Add migrations from existing state.
Update readable/writable stores to use new states

**Related issue(s)**:

Fixes #16059 
Fixes #16068 
Fixes #16069

**Notes for reviewer**:

**Checklist**


hapi/hedera-protobufs/services/state/schedule/schedule.proto
```diff
@@ -146,3 +146,14 @@ message ScheduleList {
...
   repeated Schedule schedules = 1;
 }
+
+/**
+ * A message for storing a list of schedule_ids in state.
+ * This is used to store lists of schedule_id that expire at a particular time.
+ */
+message ScheduleIdList {
+    /**
+     * a list of schedule_ids, in no particular order.
```
jsync-swirlds, line 156:
```suggestion
     * A list of schedule_ids, in no particular order.
```

hapi/hedera-protobufs/services/state/schedule/schedule.proto
```diff
@@ -146,3 +146,14 @@ message ScheduleList {
    */
   repeated Schedule schedules = 1;
 }
+
+/**
+ * A message for storing a list of schedule_ids in state.
+ * This is used to store lists of schedule_id that expire at a particular time.
```
jsync-swirlds, line 152:
```suggestion
 * This is used to store lists of ScheduleID values.
 * One example is all schedules that expire at a particular time.
```

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ReadableScheduleStoreImpl.java
```diff
@@ -41,8 +42,8 @@ public class ReadableScheduleStoreImpl implements ReadableScheduleStore {
             "Null states instance passed to ReadableScheduleStore constructor, possible state corruption.";
 
     private final ReadableKVState<ScheduleID, Schedule> schedulesById;
-    private final ReadableKVState<ProtoLong, ScheduleList> schedulesByExpirationSecond;
-    private final ReadableKVState<ProtoBytes, ScheduleList> schedulesByStringHash;
+    private final ReadableKVState<ProtoLong, ScheduleIdList> scheduleIdsByExpirationSecond;
+    private final ReadableKVState<ProtoBytes, Schedule> schedulesByStringHash;
```
jsync-swirlds, line 46: String hash values can and will collide (it's only a partial hash, not comprehensive).  Changing this to a single schedule will _miss_ duplications.  That's why we stored the whole schedule originally; storing a list of ID values might be acceptable as an alternative to storing the whole schedule.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/ScheduleCreateHandler.java
```diff
@@ -250,15 +249,12 @@ public Fees calculateFees(@NonNull final FeeContext feeContext) {
         return scheduleOpsUsage.scheduleCreateUsage(txn, sigUsage, lifetimeSecs);
     }
 
-    private @Nullable Schedule maybeDuplicate(
-            @NonNull final Schedule schedule, @Nullable final List<Schedule> duplicates) {
-        if (duplicates == null) {
+    private @Nullable Schedule maybeDuplicate(@NonNull final Schedule schedule, @Nullable final Schedule duplicate) {
```
jsync-swirlds, line 252: This should still handle hash collisions, the hash implementation is not robust enough to guarantee no collisions, so it is entirely likely that multiple schedules will match the hash, and need to be checked for duplication.

hedera-node/hedera-schedule-service/src/main/java/com/hedera/node/app/service/schedule/ReadableScheduleStore.java
```diff
@@ -56,11 +56,11 @@ public interface ReadableScheduleStore {
      * @return a {@code List<Schedule>} of entries that have the same hash as the provided schedule
      */
     @Nullable
-    List<Schedule> getByEquality(@NonNull Schedule scheduleToMatch);
+    Schedule getByEquality(@NonNull Schedule scheduleToMatch);
```
jsync-swirlds, line 59: This should still be a list (even if it's changed to just be ID values).
The "duplication" hash of the schedule is not enough to know if it's a duplicate, and multiple "candidate" duplicate schedules may have the same "duplication" hash value.

jsync-swirlds (commented):
I don't think the change from a list of candidate duplicate schedules to a single candidate duplicate schedule is correct.
The partial "hash" used is not robust, and collisions are quite likely, so we should still have a list (perhaps of ID values instead of whole schedules) keyed to the hash value, and check all of them when a new schedule _might_ be a duplicate.

Edit: The hash function used is better than I thought, so this is incorrect, but see below for an even more effective alternative that avoids needing a hash function entirely.

iwsimon:
> I don't think the change from a list of candidate duplicate schedules to a single candidate duplicate schedule is correct. The partial "hash" used is not robust, and collisions are quite likely, so we should still have a list (perhaps of ID values instead of whole schedules) keyed to the hash value, and check all of them when a new schedule _might_ be a duplicate.

Thanks for reviewing this PR @jsync-swirlds. After @tinker-michaelj PR #16053, schedule service code has been updated.
`scheduleByEqualityMutable` is only updated during `WritableScheduleStoreImpl.put()`, which are called in `ScheduleCreateHandler.handle()` and `ScheduleSignHandler.handle()`. In `ScheduleSignHandler`, we handle the same Schedule. In `ScheduleCreateHandler`, we use `maybeDuplicate() `to double check with the same attributes as in `ScheduleStoreUtility.calculateBytesHash() `of a Schedule, if it's duplicate it will throw HandleException, otherwise it gets put in `scheduleByEqualityMutable`. You will never have more than one Schedule with same ProtoBytes in `scheduleByEqualityMutable`.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
iwsimon: Updated.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
iwsimon: Updated.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/ReadableScheduleStoreImpl.java
iwsimon: The reason is explained below.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/ScheduleCreateHandler.java
iwsimon: The reason is explained below.

hedera-node/hedera-schedule-service/src/main/java/com/hedera/node/app/service/schedule/ReadableScheduleStore.java
iwsimon: The reason is explained below.

jsync-swirlds:
> > I don't think the change from a list of candidate duplicate schedules to a single candidate duplicate schedule is correct. The partial "hash" used is not robust, and collisions are quite likely, so we should still have a list (perhaps of ID values instead of whole schedules) keyed to the hash value, and check all of them when a new schedule _might_ be a duplicate.
> 
> Thanks for reviewing this PR @jsync-swirlds. After @tinker-michaelj PR #16053, schedule service code has been updated. `scheduleByEqualityMutable` is only updated during `WritableScheduleStoreImpl.put()`, which are called in `ScheduleCreateHandler.handle()` and `ScheduleSignHandler.handle()`. In `ScheduleSignHandler`, we handle the same Schedule. In `ScheduleCreateHandler`, we use `maybeDuplicate() `to double check with the same attributes as in `ScheduleStoreUtility.calculateBytesHash() `of a Schedule, if it's duplicate it will throw HandleException, otherwise it gets put in `scheduleByEqualityMutable`. You will never have more than one Schedule with same ProtoBytes in `scheduleByEqualityMutable`.

That's kind of my point.
If there is only ever one schedule that _might_ be a duplicate, you have a high probability of allowing a duplicate to be added.
Example:
 * Schedule a is added, it has properties "f", "g", "h" and hash 75
 * Schedule b is added, it has properties "f", "q", "h" and hash 75
    * Schedule b is not a duplicate, but will _replace_ schedule A in `scheduleByEqualityMutable` because they have the same key (despite not being duplicates).
 * Schedule c is added, it has properties "f", "g", "h" and hash 75
    * Schedule c is not _considered_ a duplicate because it does not match Schedule b, but it _is_ a duplicate of Schedule a, which is not represented in `scheduleByEqualityMutable`.

I believe this can still happen with the latest changes.

iwsimon:
> > > I don't think the change from a list of candidate duplicate schedules to a single candidate duplicate schedule is correct. The partial "hash" used is not robust, and collisions are quite likely, so we should still have a list (perhaps of ID values instead of whole schedules) keyed to the hash value, and check all of them when a new schedule _might_ be a duplicate.
> > 
> > 
> > Thanks for reviewing this PR @jsync-swirlds. After @tinker-michaelj PR #16053, schedule service code has been updated. `scheduleByEqualityMutable` is only updated during `WritableScheduleStoreImpl.put()`, which are called in `ScheduleCreateHandler.handle()` and `ScheduleSignHandler.handle()`. In `ScheduleSignHandler`, we handle the same Schedule. In `ScheduleCreateHandler`, we use `maybeDuplicate() `to double check with the same attributes as in `ScheduleStoreUtility.calculateBytesHash() `of a Schedule, if it's duplicate it will throw HandleException, otherwise it gets put in `scheduleByEqualityMutable`. You will never have more than one Schedule with same ProtoBytes in `scheduleByEqualityMutable`.
> 
> After multiple discussions, and a discussion in slack about how degenerate inputs are handled by various hash functions, a better approach presents itself (thanks to a suggestion from Leemon).
> 
> We can take the three fields we compare, and just concatenate those values together for the `ProtoBytes` _key_, then make ScheduleID the _value_. That eliminates the secondary comparison (if the key matches, the values are identical), and the scheduleID in the value is exactly what we need to add in the Block Stream/Record Receipt for the "duplicate schedule" response. Overall we should end up saving quite a bit of space on average, and even worst-case we'll be saving more than 50 bytes for each map entry.

We have Schedule's 5 fields: memo, admin key, scheduled transaction, expiration time and wait for expiry flag. As memo could be up to 100 Utf8Bytes, we should still use sha256() to hash them as `ProtoBytes` Key.

jsync-swirlds:
> > > > I don't think the change from a list of candidate duplicate schedules to a single candidate duplicate schedule is correct. The partial "hash" used is not robust, and collisions are quite likely, so we should still have a list (perhaps of ID values instead of whole schedules) keyed to the hash value, and check all of them when a new schedule _might_ be a duplicate.
> > > 
> > > 
> > > Thanks for reviewing this PR @jsync-swirlds. After @tinker-michaelj PR #16053, schedule service code has been updated. `scheduleByEqualityMutable` is only updated during `WritableScheduleStoreImpl.put()`, which are called in `ScheduleCreateHandler.handle()` and `ScheduleSignHandler.handle()`. In `ScheduleSignHandler`, we handle the same Schedule. In `ScheduleCreateHandler`, we use `maybeDuplicate() `to double check with the same attributes as in `ScheduleStoreUtility.calculateBytesHash() `of a Schedule, if it's duplicate it will throw HandleException, otherwise it gets put in `scheduleByEqualityMutable`. You will never have more than one Schedule with same ProtoBytes in `scheduleByEqualityMutable`.
> > 
> > 
> > After multiple discussions, and a discussion in slack about how degenerate inputs are handled by various hash functions, a better approach presents itself (thanks to a suggestion from Leemon).
> > We can take the three fields we compare, and just concatenate those values together for the `ProtoBytes` _key_, then make ScheduleID the _value_. That eliminates the secondary comparison (if the key matches, the values are identical), and the scheduleID in the value is exactly what we need to add in the Block Stream/Record Receipt for the "duplicate schedule" response. Overall we should end up saving quite a bit of space on average, and even worst-case we'll be saving more than 50 bytes for each map entry.
> 
> We have Schedule's 5 fields to compare: memo, admin key, scheduled transaction, expiration time and wait for expiry flag. As memo could be up to 100 Utf8Bytes, we should still use sha256() to hash them as `ProtoBytes` Key.

If we store the entire schedule as a value, we will always store more data than the 5 fields, regardless of length, as all 5 fields will be in the value along with other fields.
Example:
* Schedule A has 386 bytes in memo, admin_key, scheduled, expiration, wait_for_expiry.
* Schedule A has another 36 bytes in all other fields.
   * Alternative A
      * We hash the 5 fields, get 32 bytes of hash, and store the 32 bytes as key and 422 bytes as value (the full size of the schedule).
      * Total storage = 454 bytes
   * Alternative B
      * We store 386 bytes as key, and <=7 bytes as value (for the ID)
      * Total storage = 393 bytes

By storing them in the key and just an ID in the value, we will always save space compared to the full schedule in value and a 256-bit hash.  We would also avoid the cost of the hash function, which isn't trivial...
I believe we should choose the path that stores less data, but I hold that opinion lightly.

iwsimon:
> Member

Alternative C
We hash the 5 fields, get 32 bytes of hash, and store the 32 bytes as key and <=7 bytes as value (for the ID)
Total storage = 39 bytes.

jsync-swirlds:
> > Member
> 
> Alternative C We hash the 5 fields, get 32 bytes of hash, and store the 32 bytes as key and <=7 bytes as value (for the ID) Total storage = 39 bytes.

That's definitely an option.  If we do not mind the lookup to get the full schedule value when needed, then that's an even more space efficient option.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/WritableScheduleStoreImpl.java
```diff
@@ -161,16 +160,18 @@ private Schedule markDeleted(final Schedule schedule, final Instant consensusTim
...
-                    schedulesByIdMutable.remove(schedule.scheduleIdOrThrow());
-
-                    final ProtoBytes hash = new ProtoBytes(ScheduleStoreUtility.calculateBytesHash(schedule));
-                    schedulesByEqualityMutable.remove(hash);
-                    logger.info("Purging expired schedule {} from state.", schedule.scheduleIdOrThrow());
+            final var scheduleIdList = scheduleIdsByExpirationMutable.get(second);
+            if (scheduleIdList != null) {
+                for (final var scheduleId : scheduleIdList.scheduleIds()) {
+                    final var schedule = schedulesByIdMutable.get(scheduleId);
+                    if (schedule != null) {
```
tinker-michaelj, line 167: If this _was_ `null` it would be a serious invariant failure, and we should log an event at `ERROR` I think.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/WritableScheduleStoreImpl.java
```diff
@@ -161,16 +160,18 @@ private Schedule markDeleted(final Schedule schedule, final Instant consensusTim
...
+                        final ProtoBytes hash = new ProtoBytes(ScheduleStoreUtility.calculateBytesHash(schedule));
+                        scheduleIdByEqualityMutable.remove(hash);
+                    }
+                    schedulesByIdMutable.remove(scheduleId);
+                    logger.info("Purging expired schedule {} from state.", scheduleId);
```
tinker-michaelj, line 172: We should probably either remove this or make it a `DEBUG` level event.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/V0570ScheduleSchema.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public V0570ScheduleSchema() {
+
+        super(VERSION);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @NonNull
+    @Override
+    public Set<StateDefinition> statesToCreate() {
```
tinker-michaelj, line 69: In this schema we can also add a `statesToRemove()` override that gets rid of the now-unused state keys from the `V0490ScheduleSchema`---see [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-app/src/main/java/com/hedera/node/app/state/recordcache/schemas/V0540RecordCacheSchema.java#L60) for an example of using `statesToRemove()`.

tinker-michaelj (dismissed):
Awesome work @iwsimon , this all makes sense to me! 🎉 

I added a few comments that could be done in a separate PR.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
```diff
@@ -146,3 +146,15 @@ message ScheduleList {
    */
   repeated Schedule schedules = 1;
 }
+
+/**
+ * A message for storing a list of schedule_ids in state.
+ * This is used to store lists of ScheduleID values.
+ * One example is all schedules that expire at a particular time.
```
jsync-swirlds, line 153:
```suggestion
 * A message for storing a list of schedule identifiers in state.<br/>
 * This is used to store lists of `ScheduleID` values.
 * One example is all schedules that expire at a particular time.
```

hapi/hedera-protobufs/services/state/schedule/schedule.proto
```diff
@@ -146,3 +146,15 @@ message ScheduleList {
...
+ */
+message ScheduleIdList {
+   /**
+    * A list of schedule_ids, in no particular order.
+    */
+    repeated ScheduleID schedule_ids = 1;
```
jsync-swirlds, line 159:
```suggestion
    /**
     * A list of schedule identifiers, in no particular order.
     * <p>
     * While the order is not _specified_, it MUST be deterministic.
     */
    repeated ScheduleID schedule_ids = 1;
```

jsync-swirlds (dismissed):
LGTM, just two minor updates to protobuf specification for clarity.

Note also, the `state_changes.proto` modifications impact the in-progress Block Streams HIP, and must be reflected there.

In the somewhat near future that HIP will be finalized, and further changes to that file will require a separate HIP.

tinker-michaelj (dismissed):

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/WritableScheduleStoreImpl.java
iwsimon: Added error log.

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/WritableScheduleStoreImpl.java
iwsimon: Changed to debug level log.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
```diff
@@ -148,13 +148,15 @@ message ScheduleList {
 }
 
 /**
- * A message for storing a list of schedule_ids in state.
- * This is used to store lists of ScheduleID values.
+ * A message for storing a list of schedule identifiers in state.<br/>
+ * TThis is used to store lists of `ScheduleID` values.
```
jsync-swirlds, line 152: TT -> T

hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/schemas/V0570ScheduleSchema.java
iwsimon: Added. I was going to ask how to remove the old schema. This answered. Thanks.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
iwsimon: Updated.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
iwsimon: Updated.

iwsimon:
> LGTM, just two minor updates to protobuf specification for clarity.
> 
> Note also, the `state_changes.proto` modifications impact the in-progress Block Streams HIP, and must be reflected there.
> 
> In the somewhat near future that HIP will be finalized, and further changes to that file will require a separate HIP.

Thanks for pointing this out. I will create another PR on protobuf repository for *.proto file changes here.

hapi/hedera-protobufs/services/state/schedule/schedule.proto
iwsimon: Done. Thanks.

jsync-swirlds (approved):

jsync-swirlds (approved):

jsync-swirlds (dismissed):

jsync-swirlds (approved):

tinker-michaelj (approved):

PR #16178, opened 2024-10-25
feat: counting get balance throttle

### LINKED ISSUE #16177: Implement counting GetBalance throttle
Implement a throttle that takes the number of balances into account when throttling GetBalance-queries

netopyr:
**Description**:

This PR adds the possibility to consider the number of balances of an account when throttling.

**Related issue(s)**:

Fixes #16177 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/throttle/ThrottleAccumulator.java
```diff
@@ -197,13 +200,35 @@ public boolean checkAndEnforceThrottle(
...
         }
         return false;
     }
 
+    private int getBalanceCount(@NonNull final Query query, @NonNull final ReadableAccountStore accountStore) {
+        final var accountID = query.cryptogetAccountBalanceOrThrow().accountID();
+        if (accountID != null) {
+            final var account = accountStore.getAccountById(accountID);
+            if (account != null) {
+                return account.numberPositiveBalances();
```
tinker-michaelj, line 226: An account with lots of associated `0` token balances will still be expensive to query; I would suggest,
```
return Math.min(account.numberAssociations(), tokenConfig.maxRelsPerInfoQuery());
```
(C.f. [here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/state/token/account.proto#L132) and [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-token-service-impl/src/main/java/com/hedera/node/app/service/token/impl/handlers/CryptoGetAccountBalanceHandler.java#L186)).

netopyr: Right, good catch! I will fix it.

Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr. Should we add an end-to-end test for it, can be done in a separate PR.

tinker-michaelj (approved):
LGTM, ty @netopyr !

PR #16180, opened 2024-10-25
feat: withAndWithoutLongTermEnabled mod implementation

### LINKED ISSUE #16010: Rehabilitate the mono withAndWithoutLongTermEnabled() HapiTest modifier

vtronkov:
**Description**:
1. Converting `defaultHapiSpec` to `hapiTest`*
2. Implementing `AllTestsWithLongTermFlagEnabledTest` and `AllLeakyTestsWithLongTermFlagEnabledTest` that consist of `ALL_TEST` list of test classes. In these new test classes we are getting all tests methods from `ALL_TEST`, we enable the `scheduling.longTermEnabled` flag and run the tests. That way we are not having a separate CI step and we are testing the scheduled tests with the flag enabled**
3. Ignoring `signFailsDueToDeletedExpiration` test since it works only with `longTermEnabled=false`

* *This was relevant for the first implementation but not anymore. I will leave the change because we are using `hapiTest` more frequently.
* **I separated the tests to `@HapiTests` and `@LeakyHapiTests` because the leaky tests are running in series. In order to make it work I need to annotate `runAllTests` with `@LeakyHapiTest` which makes all ~110 tests running in series because of 5 tests that are leaky. If I combine all the tests it takes about 1:20 minutes(as embedded test) to run all tests. If they are separated it takes around 30 seconds.

**Related issue(s)**:

Fixes #16010 

**Checklist**

###### (just for reference/history, not required to read)
1. My first idea was to make it work as _Eth suffix is working but this wasn't exactly how the mono's `withAndWithoutLongTermEnabled` was working. So I put it aside.
2. Writing a jUnit extension to turn on the flag - I was unable to figure out a way to insert `SpecOperation` into the currently executed method. I didn't spend too much time here because I didn't quite like this solution
4. Wrote my own `HapiSpec.hapiTest` method 
3.1 In The first version I was building two separate `DynamicTest` - one with the flag turned on and one with the flag turned off. Then I put the two tests into the returned `Stream<DynamicTest>`. This was kinda working but I had problems with the state and things didn't work the way I expected.
3.2 In the second version I was building a single `DynamicTest` where I put the operation twice - with a flag operation turned off at the beginning and a flag operation turned on between the tests. Again I had some store problems, the ids were duplicated and I got errors along the lines of "the scheduled transaction was already executed". I tried fixing these problems by changing the names of the operations but in order to do that I had to make deep copies of the objects and it was getting more complicated than needed.
5. In the end, I returned to the original idea - using the file name as an indicator that the flag should be turned on(and eventually adding additional logic).
6. Because of the discussion with Michael(see below) and since we decided not to add the additional complex logic there is no need to implement the suffix solution. Thus the `AllTestsWithLongTermFlagEnabledTest` and `AllLeakyTestsWithLongTermFlagEnabledTest` was implemented(see details above)

My questions to @tinker-michaelj are:
1. ~~Are you okay with this approach or you had something else in mind? If so I'm guessing I will try to integrate this with CI just like `_Eth` was working~~
2. ~~Currently all tests from the schedule package instead of 2 are passing. In the beginning, we were talking about making this logic smart so it can detect that the flag is turned on and if needed to sleep until a scheduled transaction is executed. It looks like most of the tests are testing the transactions themselves and there are only a few end-to-end tests where this additional logic will be helpful. Maybe it's an overkill to add it at that point and it would be easier to write tests that are only for long-term transactions. Or maybe I'm missing something here?~~

**Discussed this with Michael** - we don't need the additional "understanding that the flag is on and make the test sleep until the transaction is executed" is not needed because the tests are not written in that way


hedera-node/test-clients/src/main/java/module-info.java
```diff
@@ -94,7 +94,6 @@
     requires com.hedera.node.app.service.token.impl;
     requires com.hedera.node.app.service.token;
     requires com.swirlds.base.test.fixtures;
-    requires com.swirlds.config.extensions.test.fixtures;
```
vtronkov, line 97: Since I deleted `ScheduleExecutionTest.getTestConfig` method(it was not used) we no longer use `import com.swirlds.config.extensions.test.fixtures.TestConfigBuilder;`. Thus this package not needed

tinker-michaelj:
I think we need to disable some of these---note multiple failures in the (Misc) check [here](https://scans.gradle.com/s/tfklzbf3g2apw).

vtronkov:
> I think we need to disable some of these---note multiple failures in the (Misc) check [here](https://scans.gradle.com/s/tfklzbf3g2apw).

I think 3 of them are flaky and are not related to the PR. The other two I've written about in the description above. Hopefully, when Jivko is done with this PR(the one this branch is branched off) they will be fixed. If not - we can disabled them.

tinker-michaelj (approved):
Nicely executed @vtronkov ! LGTM, tyvm 🙏

PR #16188, opened 2024-10-26
ci: disable release 0.53 regression

JeffreyDallas:
**Description**:

0.54 is on mainnet , no longer need to run regression for release 0.54


san-est (approved):
LGTM

PR #16218, opened 2024-10-29
fix: isAuthorizedRaw fails when given an auto-created account with EC…

### LINKED ISSUE #15920: bug: `isAuthorizedRaw` fails when given an auto-created account with EC key
Or, any account with an EC key, auto-created or not, it's just that auto-creation is an easy way to get one.

It's a bug at https://github.com/hashgraph/hedera-services/blob/dd730bfb7603d4a40d3e73307a68e89c456f4d76/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCall.java#L151 - that `false` should be checking the key type, as in the `ED` case on the next line.

(Does not affect what is probably the most common use case: lazy created accounts.)

david-bakin-sl:
**Description**:
Fix bug where `isAuthorizedRaw` fails when given an auto-created account with EC… key

Fix bug for auto-created account.  Add a HapiSpec DSL method for creating accounts all four ways they can be created with EC keys.
* The four ways are: explicitly create with EC key, auto create, completed hollow account, long-zero

**Related issue(s)**:
Fixes #15920

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
@@ -174,16 +174,6 @@ public class ConversionUtils {
     }
 
     /**
-     * Given an entity number, returns it as a headlong address.
-     *
-     * @param anEntityNum a number (typically, last element of a shard.realm.num triplet)
-     * @return the argument represented as a Headlong address
-     */
-    public static com.esaulpaugh.headlong.abi.Address headlongAddressOf(final long anEntityNum) {
-        return asHeadlongAddress(asEvmAddress(anEntityNum));
-    }
-
-    /**
      * Given an account, returns its "priority" address as a Besu address.
      *
      * @param account the account
--- a/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
+++ b/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
@@ -174,6 +174,16 @@ public class ConversionUtils {
     }
 
     /**
+     * Given an entity number, returns it as a headlong address.
+     *
+     * @param anEntityNum a number (typically, last element of a shard.realm.num triplet)
+     * @return the argument represented as a Headlong address
+     */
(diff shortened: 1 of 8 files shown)


hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/systemcontracts/has/isauthorizedraw/IsAuthorizedRawCallTest.java
```diff
@@ -83,6 +87,23 @@ void revertsWhenEcdsaIsNotEvmAddress() {
...
+    @ParameterizedTest
+    @ValueSource(longs = {0L, 1L, 10L, 100L, 1_000_000_000_000L, Long.MAX_VALUE})
+    void anyNonNegativeAccountValidIfED(final long account) {
+        final var result = getSubject(mock(Address.class)).isValidAccount(account, SignatureType.ED);
+        assertTrue(result);
+    }
+
+    @ParameterizedTest
+    @CsvSource({})
+    void noLongZeroAddressesValidIfEC(final long account, final boolean expected) {}
```
stoyanov-st, line 105: Missing this test body

david-bakin-sl: yes, TY

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/utils/ConversionUtils.java
```diff
@@ -173,6 +173,16 @@ public static com.esaulpaugh.headlong.abi.Address headlongAddressOf(@NonNull fin
         return asHeadlongAddress(asEvmAddress(tokenId.tokenNum()));
     }
 
+    /**
+     * Given an entity number, returns it as a headlong address.
```
lukelee-sl, line 177: Is this version of the overloaded method ever used/called?

david-bakin-sl: I think I used it at one time, not used now.   Removed it.

kimbor (approved):
lgtm!

PR #16220, opened 2024-10-29
feat: 15155 Implemented `loadSnapshot` method

### LINKED ISSUE #15155: Implement Load Snapshot method
This ticket does not include supporting multiple states in memory concurrently.

- [ ] Determine where the load method will be
- [ ] Extract code that iterates over all states on disk and loads a state to the State API module
- [ ] Refactor platform code to use this "new" API method

imalygin:
**Description**:

This PR adds `loadSnapshot` method to `State` and implementation to `MerkleStateRoot`. 

Misc:
- moved snapshot-related constants to `MerkleTreeSnapshotReader`
- inlined usages of `SignedStateFileUtils::readState`

**Related issue(s)**:

Fixes #15155 

**Checklist**


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
```diff
@@ -0,0 +1,166 @@
...
+        @NonNull
+        public Hash hash() {
+            return hash;
+        }
+
+        @Override
+        @Nullable
+        public SigSet sigSet() {
+            return sigSet;
+        }
```
anthony-swirldslabs, line 82: A `record` should provide these getters out of the box. Can we remove the manual overrides here?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Reads a SignedState from disk using the provided snapshot reader function.
```
anthony-swirldslabs, line 86: This looks like a copy of an older javadoc because a reference to a "provided snapshot reader function" doesn't seem to make sense for this method. The only thing that is provided to it is a path. Could you please rephrase the javadoc?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
```diff
@@ -0,0 +1,166 @@
...
+     * Read the version from a signed state file and check it
+     *
+     * @param in the stream to read from
+     * @throws IOException if the version is invalid
+     * @return the protocol version
+     */
+    private static int readAndCheckStateFileVersion(@NonNull final MerkleDataInputStream in) throws IOException {
+        final byte versionByte = in.readByte();
+        if (versionByte != VERSIONED_FILE_BYTE) {
+            throw new IOException("File is not versioned -- data corrupted or is an unsupported legacy state");
```
anthony-swirldslabs, line 156: The javadoc for `VERSIONED_FILE_BYTE` says `The signed state file was not versioned before, this byte was introduced to mark a versioned file`.

I'm wondering if the `before` here refers to the v1 or something prior to the v1? If it does refer to the v1, then I don't think we should throw an exception here. If it refers to something before the v1 even, then I suggest to articulate this more clearly in the javadoc for this constant.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -72,50 +65,36 @@ public static List<SavedStateInfo> getSavedStateFiles(
...
-     * @param state the Merkle tree state
-     * @param hash the hash of the state
-     * @param sigSet the signature set
-     * @param fileVersion the version of the file
-     */
-    public record StateFileData(
-            @NonNull MerkleRoot state, @NonNull Hash hash, @Nullable SigSet sigSet, int fileVersion) {}
-
     /**
      * Reads a SignedState from disk using the provided snapshot reader function. If the reader throws
```
anthony-swirldslabs, line 69: There's not a reader function provided anymore, too. Please update the javadoc accordingly.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
imalygin: >I'm wondering if the before here refers to the v1 or something prior to the v1? 

I think it was prior to v1. This particular byte has been there for many releases. 

>then I suggest to articulate this more clearly in the javadoc for this constant.

how would you recommend to rephrase it?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
imalygin: Fixed

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
anthony-swirldslabs: Something like:

```
Prior to v1, the signed state file was not versioned. This byte was introduced in v1 to mark a versioned file.
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
anthony-swirldslabs: I don't believe this particular one has been fixed yet.

imalygin: Now it is.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/MerkleTreeSnapshotReader.java
imalygin: Thank you. I updated javadoc as you suggested.

anthony-swirldslabs (dismissed):

lpetrovic05 (dismissed):

timo0 (approved):

anthony-swirldslabs (approved):

lpetrovic05 (approved):

PR #16237, opened 2024-10-30
Per service metrics poc

### LINKED ISSUE #16224: POC for enabling services to manage their own metrics
Many of the services-related node metrics are metrics generated for each transaction type, independent of the service that produces them. However, as part of an EVM metric requirements effort, we want to provide a way for a service to manage a set of its own metrics. A first pass has been done [here](https://github.com/hashgraph/hedera-services/pull/16077). This ticket is for exploring the possibility of creating a more generic solution, and to demonstrate the concept with a POC. 

The services layer follows a typical pattern of injecting needed system functionality via context objects, while also restricting service access to each service's own scope (in the majority of cases). For example, the `com.hedera.node.app.spi.store.StoreFactory` interface is used to create restricted wrappers around a service's portions of state. Having a similar mechanism for creating and managing _metrics_ would give each service a built-in way to focus on their own metrics logic while avoiding the hassle of integrating said metrics into the internals of the consensus node. However, if this approach will not work, we likely want to find another solution.

mhess-swl:
**Description**:

**Related issue(s)**:

Closes #16224 

**Notes for reviewer**:


mhess-swl:
Closing for now as there doesn't seem to be much interest in this feature

PR #16313, opened 2024-10-30
chore: HIP-796: Add token lock, partition, move keys (Services)

### LINKED ISSUE #16310: Add `HIP-796` Protobuf Key Definitions
Addition of `lock_key` , `partition_key` and `partition_move_key` definition to Protobufs in support of HIP-796

thomas-swirlds-labs:
**Description**:

Addition of lock_key , partition_key and partition_move_key definition to Protobufs in support of HIP-796

**Related issue(s)**:

Fixes #16310 

**Reviewer Notes**
Note this PR is only for adding values to this query and unblock dev work on feature branch


hapi/hedera-protobufs/services/token_get_info.proto
```diff
@@ -209,6 +209,31 @@ message TokenInfo {
      * (token definition and individual NFTs).
      */
     Key metadata_key = 28;
+
+    /**
+     * The key which can lock tokens in an account or account partition.
```
jsync-swirlds, line 214: nitpick:
The summary line should be answering a question, not assigning a purpose.
That is: <dl><dt>What is this?</dt><dd>A thing or subcategory of thing.</dd></dl>
That answer is then followed by any explanatory detail (after a line break).

Here:
```
A function-specific account key.<br/>
This key authorizes transactions to lock tokens in an account or
account partition.
```

hapi/hedera-protobufs/services/token_get_info.proto
jsync-swirlds, line 1: Is this PR _only_ for adding values to this query, or is there intention to add keys in state and/or transactions as well?

A bit more detail in the PR description (there is a heading for "reviewer notes" in the template, for instance), and possibly removing the extraneous items, like checklists, would help quite a bit here.

jsync-swirlds (commented):
A question about the intended scope of this PR, and
A mild recommendation for the specification text

hapi/hedera-protobufs/services/token_get_info.proto
thomas-swirlds-labs: So to generalize should all fields of type Key follow the same pattern -- all should begin with the same summary line? - in this case - `A function-specific account key.<br/>`

hapi/hedera-protobufs/services/token_get_info.proto
thomas-swirlds-labs: The development for `HIP-796` has been taking place on a long running feature branch in `Services` (and a matching feature branch in the `hedera-protobufs` repo). This work started long before we began the process of making the proto changes in `Services` and after approval syncing those changes with `hedera-protobufs`. We intend to integrate these commits with develop at a later stage which would include all the other state and transaction changes but for now this is just to unblock dev work

hapi/hedera-protobufs/services/token_get_info.proto
jsync-swirlds: All of the function-specific keys (e.g. metadata, kyc, pause, etc...) probably should, yes.
Many summary lines will be similar across messages or fields.  It's the description (after the line break) that gives added detail when the summary is repeated or not sufficiently clear on its own.

jsync-swirlds (approved):

PR #16400, opened 2024-11-04
chore: consistently use findbugs annotation

jjohannes:
**Description**:

Do not use 'org.jetbrains.annotations'.

See this for getting IntelliJ to be more helpful there:
https://github.com/hashgraph/hedera-services/blob/1fe6a2de8b797eec943a45420c48a0d59f9f3271/docs/intellij-quickstart.md#spotbugs-annotations 

This should not have been added: https://github.com/hashgraph/hedera-services/pull/15908/files#diff-b387a7cf1f02bb7e5aeb741f42e0f851f454dd6f5f48beab877b433b3cb46e6cR55

Please give some thought to `module-info` additions. In particular when it is about 3d party modules that have not been used before.

**Related issue(s)**:

Follow up to #15908

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/pairings/FakeFieldElement.java
+++ b/hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/pairings/FakeFieldElement.java
@@ -18,7 +18,6 @@ package com.hedera.node.app.tss.pairings;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.math.BigInteger;
-import org.jetbrains.annotations.NotNull;
 
 public class FakeFieldElement implements FieldElement {
 
@@ -28,7 +27,7 @@ public class FakeFieldElement implements FieldElement {
         this.value = value;
     }
 
-    @NotNull
+    @NonNull
     @Override
     public BigInteger toBigInteger() {
         return value;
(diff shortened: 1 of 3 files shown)


hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/PlaceholderTssLibrary.java
```diff
@@ -58,60 +58,60 @@ public PlaceholderTssLibrary() {
...
-    @NotNull
+    @NonNull
     @Override
-    public TssMessage generateTssMessage(@NotNull TssParticipantDirectory tssParticipantDirectory) {
+    public TssMessage generateTssMessage(@NonNull TssParticipantDirectory tssParticipantDirectory) {
         return null;
     }
 
-    @NotNull
+    @NonNull
```
jjohannes, line 67: It states `@NonNull` here, but then the method returns `null`. This does not look correct to me.
There are more cases like this in this class (see warnings in IDE).

iwsimon: This is wrong.

mxtartaglia-sl: Hey @jjohannes, As far as I understand, this was created as a temporary workaround.
All the uses of this class should be replaced with the library in `hedera-cryptography`.
cc: @MiroslavGatsanoga and @Neeharika-Sompalli

mxtartaglia-sl (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/PlaceholderTssLibrary.java
jjohannes: This has been resolved in #16406

mxtartaglia-sl (approved):

mhess-swl (approved):

iwsimon (approved):

PR #16407, opened 2024-11-04
chore: Implement `RekeyScenarioOp`

### LINKED ISSUE #16309: Add embedded `@HapiTest` support for simulating TSS transactions from other nodes
### Problem

In most TSS test scenarios, we need to be able to simulate the participation of nodes other than embedded `node0`. This includes, at minimum,
 - [ ] Submitting valid `TssMessageTransaction`s for one or more `shareId`s.
 - [ ] Submitting invalid `TssMessageTransaction`s for one or more `shareId`s.
 - [ ] Submitting `TssVoteTransaction`s for selected subsets of `shareId`s.

### Solution

Add `SpecOperation`s that simplify the task of simulating such behavior.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #16309


jeromy-cannon (dismissed):

jeromy-cannon (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/handlers/TssUtils.java
```diff
@@ -123,13 +125,27 @@ public static List<TssMessage> getTssMessages(List<TssMessageTransactionBody> va
      */
     public static Map<Long, Long> computeNodeShares(
             @NonNull final List<RosterEntry> rosterEntries, final long maxSharesPerNode) {
+        final var weights = new LinkedHashMap<Long, Long>();
+        rosterEntries.forEach(entry -> weights.put(entry.nodeId(), entry.weight()));
+        return computeSharesFromWeights(weights, maxSharesPerNode);
```
anthony-swirldslabs, line 130: Minor:
1. The new map consumes extra memory and doesn't seem to provide any extra benefit as opposed to using the `RosterEntry::weight` directly as the old code did.
2. Modifying external collections from lambdas is generally an anti-pattern because the code would break the moment someone makes the stream parallel.

Both the above issues would be amplified if/when the size of the roster grows significantly as more nodes join the network.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/StakePeriodChanges.java
```diff
@@ -172,15 +179,14 @@ public static boolean isNextStakingPeriod(
...
-
-        if (!Objects.equals(newCandidateRoster, rosterStore.getCandidateRoster())
-                && !Objects.equals(newCandidateRoster, rosterStore.getActiveRoster())) {
-            rosterStore.putCandidateRoster(newCandidateRoster);
-            tssBaseService.setCandidateRoster(newCandidateRoster, handleContext);
+        final var roster = nodeStore.snapshotOfFutureRoster();
+        if (!Objects.equals(roster, rosterStore.getCandidateRoster())
+                && !Objects.equals(roster, rosterStore.getActiveRoster())) {
+            rosterStore.putCandidateRoster(roster);
+            tssBaseService.setCandidateRoster(roster, handleContext);
```
anthony-swirldslabs, line 189: I'm wondering if the first operation should trigger the second one automatically? Also, I'm wondering if the TSS Base Service is supposed to read the actual candidate roster from the roster store instead of receiving a random roster as an input argument?

The code above doesn't seem robust enough as one can forget to rekey a new candidate roster if they miss the second call, or they can initiate a rekeying of a roster that isn't a candidate (or, in an edge case of a new network, the current active) roster.

anthony-swirldslabs (approved):

Neeharika-Sompalli (approved):
Most changes are making the code more readable. Thanks for the comments in the staking code. 
LGTM! Thanks @tinker-michaelj

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/handlers/TssUtils.java
tinker-michaelj: Thanks Anthony, re (1) I wanted to reuse the node's precise shares-from-weights calculation [here](https://github.com/hashgraph/hedera-services/blob/16309-tss-txn-sim/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/spec/utilops/tss/RekeyScenarioOp.java#L420) when setting the test expectations; re (2) this isn't a stream, just a `forEach()`.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/steps/StakePeriodChanges.java
tinker-michaelj: Agreed Anthony, this doesn't seem like _quite_ the right place to orchestrate things...will keep thinking about this. 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/handlers/TssUtils.java
anthony-swirldslabs: Fair.

PR #16427, opened 2024-11-04
feat: introduce RosterHistory

### LINKED ISSUE #16353: Create a `RosterHistory` class able to return current and previous rosters and lookup by round number.
**Background**

In the TSS-Roster proposal we explicitly pass a `RosterHistory` to the `PlatformBuilder` which contains the `current` and `previous` roster.   While we are refactoring, it may be convenient to also initialize this class with `AddressBook` and return the `current` and `previous` AddressBook.  Once we are done refactoring, the code related to `AddressBook` would be removed. 

Constructors: 
```
new RosterHistory(currentRoster, currentRound, previousRoster, previousRound) 
new RosterHistory(List<RoundRosterPair>, Map<RosterHash, Roster>)
```

Roster History API: 
```
Roster       getCurrentRoster()
Roster       getPreviousRoster()
Roster       getRosterForRound(long round)
```

**Acceptance Criteria** 

* Create unit tests that validate the behavior of the method calls for each constructor. 

**Dependencies**
 
 None

anthony-swirldslabs:
**Description**:
Introducing a `RosterHistory` class to encapsulate the information about active/previous (and in the future, any historical) rosters necessary for the consensus node to function. Specifically, this object is going to be a `PlatformBuilder` input instead of the currently used `withAddressBook`/`withRoster` and friends. The refactoring of the PlatformBuilder, as well as the code in ServicesMain that instantiates it, will be performed in a separate PR in the future.

**Related issue(s)**:

Fixes #16353

**Notes for reviewer**:
Unit tests.

**Checklist**

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterHistory.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterHistory.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.swirlds.platform.roster;
+
+import com.hedera.hapi.node.state.roster.Roster;
+import com.hedera.hapi.node.state.roster.RoundRosterPair;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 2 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterHistory.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A Roster History object that encapsulates information about the current active roster
+ * and the previous active roster, and their corresponding starting round numbers.
+ */
+public class RosterHistory {
+    private final List<RoundRosterPair> history;
+    private final Map<Bytes, Roster> rosters;
```
edward-swirldslabs, line 33: The hash values are not relevant at this point.  Since the data size is small, all algorithms are effectively the same, near constant.   But a slightly more efficient implementation would be `long[]` containing the round numbers in sorted order and `Roster[]` where the round number and its paired roster share the same index.

edward-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterHistory.java
anthony-swirldslabs: I suppose this optimization could save a few bytes consumed by the List/Map data structures, but it would require a tiny amount of extra CPU cycles to process the inputs and convert them to this "optimized" form. Further, if in the future the size of the history grows (e.g. in the fully dynamic case where the AB can change any time), then the extra CPU cycles here will increase, possibly defeating the extra few bytes we save.

I suggest we consider the optimization if/when we get some data pointing to this piece of code being a bottleneck for the system performance.

PR #16430, opened 2024-11-05
feat: generalize use of spec.autoScheduledTxns for rapid coverage

### LINKED ISSUE #16011: Evaluate ways to generalize use of `spec.autoScheduledTxns` for rapid coverage

vtronkov:
**Description**:
* Creating a `TestsWithAutoScheduledEnabledTest` class that can run all tests defined inside with `spec.autoScheduledTxns` enabled

**Related issue(s)**:

Fixes #16011

**Notes for reviewer**:
This PR is extending the changes in https://github.com/hashgraph/hedera-services/pull/16180
Inside the `TestsWithAutoScheduledEnabledTest` you can define all the test classes you want to execute and all operations you want to switch to scheduled.

For now the list of test classes is empty we need to figure out what tests we want to include.

Currently, most of the tests are failing because of the assert. When we schedule a transaction(for example CryptoCreate) we are not updating the registry. So later on when we try to get the entity by name from the registry we can't find it and the tests fail.
It's possible that I'm missing something, if not we can create a separate PR to fix this issue.

**Checklist**


tinker-michaelj (changes requested):
I think in hindsight, my attempt to add auto-scheduling as a `HapiSpec`-level property was probably misguided and we should just delete the `spec.autoScheduledTxns` property.

In general, if we find ourselves wanting to run `@HapiTest`s from multiple test classes with shared overrides, we should probably accomplish that with a JUnit Jupiter feature instead of writing `HapiSpec` extensions.

Maybe this PR can be converted to just deleting `spec.autoScheduledTxns` and any associated code in `HapiSpec`.

vtronkov:
> I think in hindsight, my attempt to add auto-scheduling as a `HapiSpec`-level property was probably misguided and we should just delete the `spec.autoScheduledTxns` property.
> 
> In general, if we find ourselves wanting to run `@HapiTest`s from multiple test classes with shared overrides, we should probably accomplish that with a JUnit Jupiter feature instead of writing `HapiSpec` extensions.
> 
> Maybe this PR can be converted to just deleting `spec.autoScheduledTxns` and any associated code in `HapiSpec`.

Thanks for the review! I will close this PR so we can save it for audit purposes and will create a new issue to delete the `spec.autoScheduledTxns` flag.

vtronkov:
The flag is deleted in this PR: https://github.com/hashgraph/hedera-services/pull/16816

PR #16444, opened 2024-11-05
fix: move `public.pfx` to correct subdirectory

### LINKED ISSUE #16439: fix cryptography migration of public.pfx file to correct subdirectory.
**Background** 

The `public.pfx` file should be moving to `.archive/date` sub-folder.   Instead it is only moving to `.archive`.   The problem is that if the public.pfx is migrated multiple times, it will overwrite with the latest version.  

**Acceptance Criteria**

The public.pfx file is moved to the dated sub folder, just like the other private pfx files. 

**Notes**

This may need to be cherry picked into 0.56 release.

edward-swirldslabs:
**Description**:

When I implemented the migration, I used the wrong sub-directory for moving `public.pfx` to.  This corrects the mistake to put the file in the dated sub-directory.  

**Related issue(s)**:

Fixes #16439 

**Notes for reviewer**:

I also fixed a bug in the Browser start which needed to add the cryptography to the address book in the call to GuiEventStorage.


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
```diff
@@ -1476,7 +1476,7 @@ private void cleanupByMovingPfxFilesToSubDirectory() throws KeyStoreException, K
         if (sPublicPfx.exists()
                 && sPublicPfx.isFile()
                 && !sPublicPfx.renameTo(
-                        pfxArchiveDirectory.resolve(sPublicPfx.getName()).toFile())) {
+                        pfxDateDirectory.resolve(sPublicPfx.getName()).toFile())) {
```
litt3, line 1479: I notice that the variable name is incorrect, `pfxDateDirectory` instead of `pfxDataDirectory`

Up to you if you want to fix it in this PR

litt3 (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/crypto/EnhancedKeyStoreLoader.java
edward-swirldslabs: Date directory is correct.  The directory is literally the date.

poulok:
I believe you have an open PR to cherry pick the original change back into 56. Please remember this one as well. I'm assuming it's needed.

edward-swirldslabs:
> I believe you have an open PR to cherry pick the original change back into 56. Please remember this one as well. I'm assuming it's needed.

The cherry pick of this PR's commit is included in:  https://github.com/hashgraph/hedera-services/pull/16457

anthony-swirldslabs (approved):

PR #16466, opened 2024-11-06
feat: Removing tssEncryptionKey from Roster state

### LINKED ISSUE #16348: Update the protobuf definition of roster.
**Background** 

* There is an existing protobuf definition of roster [located here](https://github.com/hashgraph/hedera-services/blob/develop/hapi/hedera-protobufs/services/state/roster/roster.proto). 
* The `tssEncryptionKey` needs to be removed and is no longer being stored in the roster.   Since the protobuf definition is on mainnet, it is safest to simply reserve the number tag and field name to prevent its instantiation in future protobuf.  

**Acceptance Criteria** 

1. Unit Test: load a saved serialized instance of a roster in the old format and ensure that it can be deserialized without error in the new format.   
2. Fix any code that might be dependent on the existence of this field.  

**Dependencies** 

none

thomas-swirlds-labs:
**Description**:

- [ ] Removes `tssEncryptionKey` from the roster protobuf 
- [ ] Adds a unit test that loads a saved serialized instance of a Roster in the old format and ensures that it can be deserialized without error in the new format.
- [ ] Adds another test that verifies the contents of the Roster are correct after deserialization.
- [ ] Fixes code that depended on `tssEncryptionKey`.  

**Related issue(s)**:

Fixes #16348 

**Notes for reviewer**:
Serialized instance of Roster containing RosterEntries added to `old_roster.dat`

**Checklist**


jsync-swirlds (dismissed):

timo0 (dismissed):

hapi/hedera-protobufs/services/state/roster/roster.proto
jsync-swirlds, line 1: Not sure what setting does this (I suspect a default in IntelliJ), but the indentation for `.proto` files should be **4** spaces, not **2**.

jsync-swirlds (approved):
The `.proto` change looks good.

hapi/hedera-protobufs/services/state/roster/roster.proto
```diff
@@ -45,10 +45,16 @@ message Roster {
...
  * Each roster entry SHALL encapsulate the elements required
  * to manage node participation in the Threshold Signature Scheme (TSS).<br/>
- * All fields except tss_encryption_key are REQUIRED.
+ * All fields are REQUIRED.
  */
 message RosterEntry {
 
+    // The tssEncryptionKey field has been reserved because it is no longer
+    // required to be stored in the Roster. The public portion of this key will
+    // continue to be stored in the Roster state, but will, however, be stored
```
edward-swirldslabs, line 54: The public portion of the `tssEncryptionKey` will need to be stored in `TssBaseService`, not the `Roster state`

hapi/hedera-protobufs/services/state/roster/roster.proto
```diff
@@ -45,10 +45,16 @@ message Roster {
...
+    // just as raw bytes.
+    reserved 4;
```
edward-swirldslabs, line 56: Is it too late to simply remove the field and re-number the following elements?   I know we're wanting to ship block streams as clean as possible without deprecated content in day one.

hedera-node/hedera-app/src/test/resources/old_roster.dat
```diff
@@ -0,0 +1,6 @@
+
+d
+2
+9
+gossipCaCertificate"encryptionKey*�
+domain.com
```
edward-swirldslabs, line 6: The content of this file looks bizarre, is this correct? 
```
d
2
9
...
```

thomas-swirlds-labs: It is just a serialized data file for testing. I used the old protobuf definition which had the `tssEncryptionKey` still in it to create an instance of the Roster with the RosterEntry's and serialized it to this file. Then I read this file in the unit test and deserialized it into the an instance of the new protobuf definition to ensure it still works.  
```
 Roster.PROTOBUF.write(roster, out);
```

hapi/hedera-protobufs/services/state/roster/roster.proto
thomas-swirlds-labs: Makes sense now. Have updated

hapi/hedera-protobufs/services/state/roster/roster.proto
```diff
@@ -103,5 +88,5 @@ message RosterEntry {
      * SHALL NOT supply both values for the same endpoint.<br/>
      * This list SHALL NOT be empty.<br/>
      */
-    repeated proto.ServiceEndpoint gossip_endpoint = 5;
+    repeated proto.ServiceEndpoint gossip_endpoint = 4;
```
jsync-swirlds, line 91: Renumbering a field is the most incompatible change you can make to a protobuf definition, and cannot be "fixed" with migration.
Are we certain there is no data anywhere that would require migration for this change?

edward-swirldslabs: @jsync-swirlds, All writes of the roster data structure to state are gated behind feature flags that have not been turned on by default yet.

jsync-swirlds: And they are not turned on in testnet, or previewnet?
Just making sure we don't create a situation where we parse data in a way that _silently_ corrupts data somewhere.

thomas-swirlds-labs: Reverting to reserving the field as per our discussion, thanks @edward-swirldslabs !

Neeharika-Sompalli (approved):
LGTM! Thanks @thomas-swirlds-labs

jsync-swirlds (approved):

edward-swirldslabs (approved):

Issue #16436, opened 2024-11-05
Remove `MerkleRoot` interface


thenswan:
After #15638 there will be no need to keep `MerkleRoot` interface. This issue also depends on #16322. The execution plan:
   - `SwirldState getSwirldState()` can be superseded by lifecycle methods in `MerkleStateRoot`;
   - `getReadablePlatformState`, `getWritablePlatformState` and `updatePlatformState` will be redundant after #16352, so it should be easy to simplify things;
   - `getInfoString` should be moved to the `MerkleInternal`;
   -  `State` interface also has `copy` and `createSnapshot`, so they can be simply deleted.

Issue #15574, opened 2024-09-20
Self-contained `State` module


imalygin:
This is the logical continuation of the Public Merkle API initiative (#10988). Also, it overlaps with #14772.

There are two main goals of this Epic:
1. Remove all platform specifics from the State module
2. Add state lifecycle management to the State module in an generic (non-platform specific) way

Currently, a Block node would have to not only use the `swirlds-state-api` module but also `swirlds-platform-core` as a dependency, which, in turn, contains functionality that is not relevant in the context of the state (consensus, gossip, transaction handling).

We need a State module that has not only a set of API interfaces but also all the necessary implementations to represent the full Merkle state.

### Platform Dependency Removal
- [ ] https://github.com/hashgraph/hedera-services/issues/16221
- [ ] https://github.com/hashgraph/hedera-services/issues/16380
- [ ] https://github.com/hashgraph/hedera-services/issues/16193
- [ ] https://github.com/hashgraph/hedera-services/issues/15638
- [ ] https://github.com/hashgraph/hedera-services/issues/16322
- [ ] https://github.com/hashgraph/hedera-services/issues/16223
- [ ] https://github.com/hashgraph/hedera-services/issues/16436
- [ ] https://github.com/hashgraph/hedera-services/issues/16569
- [ ] https://github.com/hashgraph/hedera-services/issues/16589

### Lifecycle Management
- [ ] https://github.com/hashgraph/hedera-services/issues/15620
- [ ] https://github.com/hashgraph/hedera-services/issues/17023

poulok:
We discussed at DevCon that this State Module may also need to include the management of states similarly to the `SwirldStateManager` which keeps a thread-safe reference to the mutable state and the latest immutable state, but more than that. It could include logic that does things like ensure that only a single loaded state is mutable and that it is always the latest one even if the user of the module loads another, older state from disk.

thenswan:
Current plan of moving forward with this epic.

1. Remove the `State` class:
   - Issue: #15638. It should be broken down into 2 separate issues: the original should be about removing the `State` class and follow up should be about removing `MerkleRoot` interface.
   - PR: #16352
2. #16322
3. Remove `MerkleRoot` interface.
   - `SwirldState getSwirldState()` can be superseded by lifecycle methods in `MerkleStateRoot`;
   - `getReadablePlatformState`, `getWritablePlatformState` and `updatePlatformState` will be redundant after #16352, so it should be easy to simplify things;
[...]

PR #16477, opened 2024-11-07
fix: 16322 Extracted platform-related functionality from `MerkleStateRoot` into `PlatformMerkleStateRoot`.

### LINKED ISSUE #16322: Refactor `MerkleStateRoot` - extract lifecycle-related code to a separate class
Currently `MerkleStateRoot` has two responsibilities: 
- it's an entry point to get access to the state (represented by `com.swirlds.state.State` interface)
- it's a bridge between the platform and the app (represented by 
`com.swirlds.platform.system.SwirldState`)

These two functionalities can and should be in the separate classes. In fact, it's a pre-requisite for #16223 Without this separation it will not be possible to define clear boundaries between `swirlds-state-impl` module and the rest of the platform.

So, the final goal of this ticket is to move the implementations of `com.swirlds.platform.system.SwirldState` methods to a separate class.

imalygin:
**Description**:

This PR splits `MerkleStateRoot` into two classes `MerkleStateRoot` and `PlatformMerkleStateRoot`.

Here is how the class diagram looks like:
image

All functionality related to the interaction between the platform and the app has been migrated to `PlatformMerkleStateRoot`. This class also implements the `MerkleRoot` interface, which will eventually be removed (https://github.com/hashgraph/hedera-services/issues/16436).

`MerkleStateRoot` is now focused solely on state interactions. It implements the `com.swirlds.state.State` interface, which serves as a gateway to the state content, and the `com.swirlds.common.merkle.MerkleInternal` interface, indicating that this is a Merkle tree implementation.

Also, `MerkleStateRoot` is now an abstract class because, to determine the current round, we need to extract it from the consensus snapshot, which resides in the platform context. As the name suggests, the platform context belongs to the `swirlds-platform-core` module. Therefore, I had to defer that logic to the concrete implementation.

`TestMerkleStateRoot` is used to test purposes and its implementation of `getCurrentRound()` always return 0. 

**Related issue(s)**:

Fixes #16322 

**Notes for reviewer**:

**Checklist**


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/snapshot/SignedStateFileReader.java
```diff
@@ -103,7 +104,7 @@ public static List<SavedStateInfo> getSavedStateFiles(
         final SignedState newSignedState = new SignedState(
                 configuration,
                 CryptoStatic::verifySignature,
-                normalizedData.state(),
+                (MerkleRoot) normalizedData.state(),
```
lpetrovic05, line 107: why is this being cast? does the signed state need to change?

imalygin: This is a temporary fix. We have a ticket for a final solution about `MerkleRoot` interface - https://github.com/hashgraph/hedera-services/issues/16436. In this case we need `MerkleRoot` instance and `PlatformMerkleStateRoot` _is_ an instance of `MerkleRoot`, so I don't anticipate any issues here.

lpetrovic05 (approved):

lpetrovic05 (approved):

artemananiev (dismissed):

lpetrovic05 (approved):

rbarkerSL (approved):
Approval applies to
- hedera-node/hedera-app/src/main/java/module-info.java
- hedera-node/test-clients/src/main/java/module-info.java
- platform-sdk/swirlds-platform-core/build.gradle.kts
- platform-sdk/swirlds-platform-core/src/main/java/module-info.java
- platform-sdk/swirlds-state-impl/src/main/java/module-info.java

lpetrovic05 (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/store/ReadableStoreFactory.java
```diff
@@ -136,7 +136,7 @@ private static Map<Class<?>, StoreEntry> createFactoryMap() {
      */
     public ReadableStoreFactory(@NonNull final State state) {
         this.state = requireNonNull(state, "The supplied argument 'state' cannot be null!");
-        if (state instanceof MerkleStateRoot merkleStateRoot) {
+        if (state instanceof PlatformMerkleStateRoot merkleStateRoot) {
             this.versionFactory = merkleStateRoot.getVersionFactory();
```
mhess-swl, line 140: Why was `getVersionFactory()` removed from `MerkleStateRoot`?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
```diff
@@ -231,7 +232,7 @@ public StateChangesValidator(
                         metrics),
                 DiskStartupNetworks::new,
                 NodeId.of(0L));
-        this.state = (MerkleStateRoot) hedera.newMerkleStateRoot();
+        this.state = (PlatformMerkleStateRoot) hedera.newMerkleStateRoot();
```
mhess-swl, line 235: I'm confused why there are so many changes to the `MerkleStateRoot` interface. Doing so requires changes like we see here, where we go from an interface to something _more_ specific, despite this not being a platform test of any kind. What is the advantage of these changes?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/store/ReadableStoreFactory.java
imalygin: Because this functionality is needed by the platform only. It doesn't have to be a part of the state.

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/junit/support/validators/block/StateChangesValidator.java
imalygin: The fundamental problem with the current `MerkleStateRoot` implementation is that it plays two roles:
- a gateway between the app and the platform
- Merkle-tree-based implementation of the State API

This PR partially addresses that - it separates platform-related code from the state implementation. Note that there is a follow-up ticket to refactor `swirlds-platform-core` further and get rid of `PlatformMerkleStateRoot`
https://github.com/hashgraph/hedera-services/issues/16568. 

Once this PR is merged and this , Block Node will be able to use `swirlds-state-api` and `swirlds-state-impl` to work with the state, and there will be no need to depend on `swirlds-platform-core`. This PR is a part of a broader effort - https://github.com/hashgraph/hedera-services/issues/15574 - which has it as the end goal.

mhess-swl (approved):
Looks good for services ✅

thenswan (approved):

lpetrovic05 (approved):

PR #16484, opened 2024-11-07
feat: add scaling multiplier to the ScheduleCreate fee calculations

### LINKED ISSUE #16455: Consider using a simpler cost scaling strategy instead of adding `SCHEDULED` subtypes

vtronkov:
**Description**:
* Adding a simple scale config that scales the fees based on the scheduled type

Fixes #16455 

**Notes for reviewer**:
* Adding a scale config file(currently we are scaling by a factor of 1, essentially not scaling). We will update that when we have the actual scaling values
* Changing the dispatch fee logic to scale the fees after they are calculated if the transaction is a ScheduleCreate.
* To get the multiplier we unpack the scheduleCrate transaction to get the inner transaction and by its type, we determine what the scale factor would be.


hedera-node/hedera-app-spi/build.gradle.kts
```diff
@@ -29,7 +29,5 @@ testModuleInfo {
     requires("org.assertj.core")
     requires("org.junit.jupiter.api")
     requires("org.junit.jupiter.params")
-    requires("org.mockito")
-    requires("org.mockito.junit.jupiter")
```
vtronkov, line 33: I got an error on the **Module Info** GitHub check:
```
* What went wrong:
Execution failed for task ':app-spi:checkModuleDirectivesScope'.
> /home/runner/_work/hedera-services/hedera-services/hedera-node/hedera-app-spi/build.gradle.kts
  
  Please remove the following requires directives (or change to runtimeOnly):
    testModuleInfo {
      requires("org.mockito")
      requires("org.mockito.junit.jupiter")
    }

```

I'm not 100% sure why I got it but my hypothesis is that someone has deleted a code that was depending on these packages and pushed the PR with a failing **Module Info** step. Because I've made some changes in the **spi** package GitHub marked that as a problem on my PR.

Either way, it seems that we do not require these packages anymore so I deleted them.

andrewb1269hg (dismissed):
Review applies to `hedera-node/hedera-app-spi/build.gradle.kts`

Approved.

hedera-node/hedera-app-spi/build.gradle.kts
andrewb1269hg: This may no longer be an issue since we reverted a gradle action version roll that went in yesterday morning.

vtronkov: Yep, I don't see it in the diff anymore. Thanks :)

tinker-michaelj (changes requested):
The basic idea looks reasonable (assuming we do need this); but we have a strict policy of not using floating-point arithmetic anywhere in transaction handling.

So we should either change the scaling factors to whole numbers; or possibly fractions. But just whole numbers to start probably.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/dispatcher/TransactionDispatcher.java
```diff
@@ -124,12 +132,55 @@ public Fees dispatchComputeFees(@NonNull final FeeContext feeContext) {
...
+
+        return fees.multiplyBy(multiplier);
+    }
+
+    // Getting the scheduled multiplier based on what transaction type is scheduled
+    private static int getScheduleMultiplier(TransactionBody transactionBody) {
+        try {
+            HederaFunctionality functionality = getScheduledHederaFunctionality(transactionBody);
+            var multiplier =
+                    new AssetsLoader().loadScheduledTransactionMultipliers().get(functionality);
```
netopyr, line 161: Is this reading a file from disk on the handle thread every time we encounter a ScheduleCreate?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/dispatcher/TransactionDispatcher.java
```diff
@@ -124,12 +132,55 @@ public Fees dispatchComputeFees(@NonNull final FeeContext feeContext) {
...
 
+    // Scaling the fees if certain conditions are met.
+    // E.g. when we do a ScheduleCreate we want to scale the fees based on what transaction type is scheduled
+    private static Fees scale(TransactionBody transactionBody, Fees fees) {
+        int multiplier = 1; // by default, we are not scaling
+
+        final var functionality = transactionBody.data().kind();
+        // We are only scaling the ScheduleCreate. In the future we can add more functionalities.
+        switch (functionality) {
+            case SCHEDULE_CREATE -> multiplier = getScheduleMultiplier(transactionBody);
```
netopyr, line 150: This is the wrong concept. We do not want transaction-specific code in the Hedera app if we can avoid it.

Why is the multiplier applied here? Is this a generic concept that will be used for multiple transaction types? I would have expected this to happen in `ScheduleCreateHandler.calculateFees()`

vtronkov: `Is this a generic concept that will be used for multiple transaction types?`
Yes, I though that we will have scaling for other types as well, but I might be wrong. At the beginning it was implemented inside `ScheduleCreateHandler.calculateFees()` but then I moved it here.

Let see if this PR is needed at all and then I will resolve your comments 👍

vtronkov:
After analyzing the HIP and the current implementation - this PR is not needed

Issue #16491, opened 2024-11-07
Extend _config.txt_ (or replacement format) with `adminKey` override support


tinker-michaelj:
### Problem

When starting a test network from a production state, operations cannot test `NodeUpdate` transactions without having the node `adminKey`s in that production state.

### Solution

Extend the file format that provides the genesis/override rosters on disk to include an optional override `adminKey`.

PR #16496, opened 2024-11-08
chore: support node admin key overrides at startup

tinker-michaelj:
**Description**:
 - Addresses #16491 for 0.56


timo0 (approved):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/record/SystemSetup.java
```diff
@@ -149,42 +156,67 @@ public void doPostUpgradeSetup(@NonNull final Dispatch dispatch) {
...
             }
         });
+        final var autoNodeAdminKeyUpdates = new AutoEntityUpdate<Map<Long, Key>>(
+                (ctx, nodeAdminKeys) -> nodeAdminKeys.forEach(
+                        (nodeId, key) -> ctx.dispatchAdmin(b -> b.nodeUpdate(NodeUpdateTransactionBody.newBuilder()
+                                .nodeId(nodeId)
+                                .adminKey(key)
+                                .build()))),
+                adminConfig.upgradeNodeAdminKeysFile(),
+                SystemSetup::parseNodeAdminKeys);
```
iwsimon, line 192: This kind of adminKey update looks dangerous to me. If we have some key in this node-admin-keys.json file, the node adminKey will be updated. This should be guarded carefully.

iwsimon (approved):
LGTM. Have some concern comments. For 0.56 to cover state migration test, it's Ok. In long term, we might want to think something more carefully. Thanks. @tinker-michaelj .

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
```diff
@@ -90,12 +105,16 @@ public void migrate(@NonNull final MigrationContext ctx) {
...
 
-        Key finalAdminKey = adminKey == null || adminKey.equals(Key.DEFAULT)
+        final var defaultAdminKey = adminKey == null || adminKey.equals(Key.DEFAULT)
                 ? Key.newBuilder().ed25519(bootstrapConfig.genesisPublicKey()).build()
                 : adminKey;
         NodeAddress nodeDetail;
         final var addressBook = networkInfo.addressBook();
         for (final var nodeInfo : addressBook) {
+            final var nodeAdminKey = nodeAdminKeys.getOrDefault(nodeInfo.nodeId(), defaultAdminKey);
+            if (nodeAdminKey != defaultAdminKey) {
```
jsync-swirlds, line 115: Should this be `==`?
From how I read this, to log _when the override is used_ it would need to be `==`.
As it stands, it logs here when the admin key from the `nodeInfo` is used and the "default" is not applied.

anthony-swirldslabs (approved):

Issue #16499, opened 2024-11-08
Read addressBook from disk instead of rosterState to construct NetworkInfo


Neeharika-Sompalli:
While constructing a `StateNetworkInfo` we are currently retrieving roster entries from state. This will work when DAB is enabled and node additions happen using DAB transactions. 

**Solution:**
Read addressBook from disk instead of rosterState to construct NetworkInfo

PR #16500, opened 2024-11-08
fix: Use disk address book to construct network Info

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/16499

Use disk address book to construct StateNetworkInfo instead of using roster entries until Roster proposal and DAB are enabled


tinker-michaelj (approved):

mhess-swl (approved):

anthony-swirldslabs (approved):

PR #16504, opened 2024-11-08
fix: User Roster from disk addressBook for StateNetworkInfo

### LINKED ISSUE #16499: Read addressBook from disk instead of rosterState to construct NetworkInfo

Neeharika-Sompalli:
Fixes #16499 

- When `keyCandidateRoster` is false, always use the diskAddressBook for constructing Roster used in `StateNetworkInfo`


anthony-swirldslabs (approved):

tinker-michaelj (approved):

tinker-michaelj (approved):

PR #16509, opened 2024-11-08
fix: set admin key to self managed key for EthereumTransaction contract deployments

### LINKED ISSUE #16508: Externalized record for contract create via `EthereumTransaction` has admin key equal to null
### Description

pr #16095 was attempting to fix the issue where contracts created via `EthereumTransactions` are externalizing a null admin key to the mirror node.  the code does not seem to work as intended and should be fixed.

### Steps to reproduce

1. Deploy a contract through the relay
2. see if the new contract has a self managing admin key

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.56.2

### Operating system

None

lukelee-sl:
**Description**:
Contracts deployed via `EthereumTransaction` are setting the externalized record to the mirror node as null.  Update to externalize a self managing contract key

**Related issue(s)**:

Fixes #16508


Nana-EC (approved):

lukelee-sl:
Needs a bed test.  Will add in follow up pr.

PR #16515, opened 2024-11-08
chore: support extra dispatch authorizations

### LINKED ISSUE #16503: Provide a way to submit signatories for authorize a schedule sign transaction via the dispatch mechanism
### Problem

Currently, there is no way to provide a set of keys to sign a schedule sign transaction via the dispatch mechanism for synthetic child transactions.  

### Solution

When handling a `ScheduleSign` transaction, the list of signatories is provided by the `KeyVerifier` via the `signingCryptoKeys` method.  Provide a way for dispatch child transactions to have the correct set of signing keys to provide via the `KeyVerifier` as currently the `DefaultKeyVerifier` only allows for using the empty map for this purpose.

### Alternatives

_No response_

tinker-michaelj:
**Description**:
 - Closes #16503 
 - Unblocks HIP-755 by providing an [overload of `SystemContractOperations.dispatch()`](https://github.com/hashgraph/hedera-services/blob/16503-support-extra-dispatch-authorizations/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/scope/SystemContractOperations.java#L83) that accepts a `Set<Key>` of authorizing keys that the `ScheduleSignHandler` will incorporate. 
 - Makes two `ScheduleService` changes to support authorizing `Key{contractID=0.0.X}` and `Key{delegatable_contract_id=0.0.X}` keys in a schedule's signatories list:
   1. The `ScheduleService` handlers now [appends both forms of numeric contract id keys](https://github.com/hashgraph/hedera-services/blob/16503-support-extra-dispatch-authorizations/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/AbstractScheduleHandler.java#L158) to the schedule signatories list in state.
   2. The [`Predicate<Key>` used to test key authorizations](https://github.com/hashgraph/hedera-services/blob/16503-support-extra-dispatch-authorizations/hedera-node/hedera-schedule-service-impl/src/main/java/com/hedera/node/app/service/schedule/impl/handlers/AbstractScheduleHandler.java#L298) now incorporates direct and "delegatable" contract id authorizations.

As an implementation detail,
 - Consolidates all `HandleContext.dispatch*()` variants into a single `Dispatch` method receiving a [`DispatchOptions`](https://github.com/hashgraph/hedera-services/blob/16503-support-extra-dispatch-authorizations/hedera-node/hedera-app-spi/src/main/java/com/hedera/node/app/spi/workflows/DispatchOptions.java#L36) with four convenience factories:
     1. `independentDispatch()` - for dispatches whose effects should be committed independent of the parent's success or failure.
     4. `setupDispatch()` - for dispatches that are part of the parent's transactional unit but logically perform a setup function prior to the parent's business logic.
     5. `subDispatch()` - for dispatches that are a logical sub-transaction within the parent's transactional unit, and should have their results externalized no matter if the parent succeeds or fails.
     6. `stepDispatch()` - for dispatches that are a logical step in the parent's business logic, and should only be externalized if the parent succeeds.


Neeharika-Sompalli (approved):
LGTM! tyvm @tinker-michaelj

lukelee-sl (approved):
LGTM.  tyvm!

Neeharika-Sompalli (changes requested):
LGTM!

Issue #16516, opened 2024-11-09
Increase Account and NFT maximums to 100M each


poulok:
Increase Account and NFT maximums to 100M each

PR #16517, opened 2024-11-09
feat: increase maximums to 100M each by default

poulok:
**Description**:
Increases the account and NFT maximums to 100M each.

**Related issue(s)**:

Addresses #16516 in release 57

**Checklist**


OlegMazurov:
What about `TokensConfig.maxAggregateRels`? Should it be proportionally increased from 50M as well?

PR #16523, opened 2024-11-11
feat: Put candidate roster in state on freeze upgrade

### LINKED ISSUE #16371: Write candidate roster to state on freeze upgrade action instead of writing config.txt if flag is turned on.
**Background**

With the release of DAB Phase 2, a new config.txt is being written to disk along with the public gossip keys triggered by a freeze upgrade transaction.  Instead of writing config.txt to disk, we need to write a new candidate roster into the state.  The gossip public keys will still be written to disk.  

Check the `AddressBook.useRosterLifecycle` feature flag at [this code location](https://github.com/hashgraph/hedera-services/blob/9cc0ab85e50337000a406aeb51d00dfb523f8034/hedera-node/hedera-network-admin-service-impl/src/main/java/com/hedera/node/app/service/networkadmin/impl/handlers/ReadableFreezeUpgradeActions.java#L275).  If the flag is false, write config.txt and PEM files with the existing implementation.   If the flag is true, save a new candidate roster to the roster state.  

**Acceptance Criteria** 

* Manually test DAB HAPI or JRS tests to validate execution with the feature flag set to true. 

**Dependencies** 
https://github.com/hashgraph/hedera-services/issues/16357

petreze:
**Description**:
This PR adds the candidate roster to state on freeze upgrade.
This is _**only**_ needed when tss is not yet enabled (if `tss.keyCandidateRoster = false`) and we decide to ship Roster proposal before TssLedgerId.

**Related issue(s)**:

Fixes #16371 

**Notes for reviewer**:

**Checklist**


edward-swirldslabs (approved):
I'm not a code owner, so my approval doesn't account for much. :-)

Neeharika-Sompalli (approved):
LGTM! Thanks @petreze

PR #16533, opened 2024-11-11
feat: change the staking defaults

### LINKED ISSUE #16532: change the staking fraction defaults
for the November mainnet update, will we be changing the setting so that the fractions of transaction fees split between 98/800/801 as 80% / 10% / 10%, rather than the current 90% / 10% / 0% ?

povolev15:
change the nodeRewardPercentage from 0 to 10

Fixes #16532


netopyr (approved):
LGTM - thanks @povolev15

jsync-swirlds (approved):

netopyr (approved):
LGTM - thanks @povolev15

mxtartaglia-sl (approved):

PR #16534, opened 2024-11-11
feat: change stacking nodeRewardPercentage

### LINKED ISSUE #16532: change the staking fraction defaults

povolev15:
fix cheery pick into 56: https://github.com/hashgraph/hedera-services/pull/16533

**Related issue(s)**:

Fixes #16532


netopyr (approved):
LGTM - thanks @povolev15

PR #16562, opened 2024-11-12
docs: TSS Encryption Key Automation Proposal

### LINKED ISSUE #16529: Create Design Proposal for Automated TSS Encryption Key Lifecycle.
**Background** 

Create a design proposal for the automated generation and lifecycle maintenance of the TssEncryptionKey.  

It was previously thought the Node Operators could manually generate the key and we would store the key in the Roster.  There is a lot of friction and change management entailed by adding responsibility to Node Operators to generate the key.  Automating it will make deployment of this capability simpler and faster. 

Since TSS has been moved to a service, the `tssEncryptionKey` has been removed from the Roster and will now need to be stored in `TssBaseService`'s state. 

This new proposal will describe the new transactions and state required for the automated `tssEncryptionKey` lifecycle.  

**Acceptance Criteria**

The proposal is in a reviewable state and ready for voting.

**Dependencies**

None

edward-swirldslabs:
**Description**:

This PR is a platform design proposal describing the process for automatically generating the `tssEncryptionKey`. 

**Related issue(s)**:

Fixes #16529


platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,231 @@
...
+   1. If there is no public key in the state, or the public key in the state does not matching the loaded
+      private key, a new public/private `tssEncryptionKey` will need to be generated through a call back to
+      `CryptoStatic.java` to generate the key and persist the new private key to disk. The new public key will need
+      to be distributed to the rest of the network as in `1.1` above.
+   2. If the public key in the state matches the loaded private key, all is good.
+
+At the end of the `TssBaseService` initialization there will be two outcomes:
+
+1. The public key in the state matches the private key and the `TssBaseService` is ready to execute normally.
+2. At the start of `TssBaseService` execution, it needs to updated the public key on the network and cannot perform
```
rbair23, line 132:
```suggestion
2. At the start of `TssBaseService` execution, it needs to update the public key on the network and cannot perform
```

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,231 @@
...
+
+1. The public `tssEncryptionKey` of a node must be recorded in the state and publicly known by other nodes.
+2. The private `tssEncryptionKey` of a node must be kept private to that node.
+3. The lifecycle of the `tssEncryptionKey` should be automated.
+
+### Design Decisions
+
+* The private `tssEncryptionKey` will be stored in the same directory as the node's private gossip signing key.
+* The public `tssEncryptionKey` will be stored in the `TssBaseService` state.
+* A new `TssEncryptionKeyTransaction` will be, gossiped, come to consensus, and handled by the `TssBaseService` to
```
rbair23, line 58:
```suggestion
* A new `TssEncryptionKeyTransaction` will be gossiped, come to consensus, and handled by the `TssBaseService` to
```

rbair23 (dismissed):
Love it.

derektriley (approved):
Looks good to me. Thanks! @edward-swirldslabs

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Changes
+
+The following changes to the system are required.
+
+### Architecture and/or Components
+
+The `TssBaseService` already exists with its own state.
+
+* This proposal adds a new `Map<Long, Bytes>` to the state for storing the public key of each node.
```
anthony-swirldslabs, line 81: Please update the doc to clarify if this is a virtual map, or a protobuf map.

If it is the latter, I recall that a decision was made to never use protobuf maps in the state because of their non-deterministic nature when it comes to serialization (which doesn't necessarily apply to PBJ in our own implementation, but may apply to 3rd-party implementations.)

If you still see this as a protobuf map, I'd like @jsync-swirlds to weigh in.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+2. The private and public `tssEncryptionKey` were generated, are present, and match each other.
+
+#### TssBaseService Initialization
+
+At the initialization of the `TssBaseService` the loaded public/private `tssEncryptionKey` will be provided. The
+private key will always be provided, but the public key may be null. There are three scenarios:
+
+1. The public `tssEncryptionKey` is provided indicating a new key-pair was generated.
+   1. Once execution is live, a `TssEncryptionKeyTransaction` will need to be sent periodically until the state
+      reflects the correct public key.
```
anthony-swirldslabs, line 123: Does the "sent periodically" part essentially implement a retry mechanism? Should there be a limit on the number of retries to avoid flooding the network with these transactions?

If it's not a retry mechanism, then why does this have to be done multiple times? Why a single transaction not sufficient?

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+2. The public `tssEncryptionKey` is not provided indicating only the private key was loaded from disk.
+   1. If there is no public key in the state, or the public key in the state does not matching the loaded
```
anthony-swirldslabs, line 125:
```suggestion
   1. If there is no public key in the state, or the public key in the state does not match the loaded
```

mxtartaglia-sl (approved):
+1
Makes a lot of sense.

As we envisioned the BlsKeyGen tool to be a cli app, we would need to include a minor change in the library code (exporting packages and opening some visibilities so they can be reused)

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+
+The `TssBaseService` state will be extended with a new map of the public `tssEncryptionKey` for each node.
+
+1. The keys of the map are Longs/UInt64 for the `nodeId` of the nodes
+2. The values of the map are the latest `TssEncryptionKeyTransaction` from each node.
+
+Data Lifecycle:
+
+* When a node id is not present in any of the `active rosters` or a `candidate roster`, the entry for that node id
+  should be removed. This cleanup task should trigger any time there is a candidate or active roster state change.
```
anthony-swirldslabs, line 186: Does the removal of entries need any synchronization and/or transactions between nodes to ensure they end up with identical lists of public keys?

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
jsync-swirlds: A Virtual Map is good, but @anthony-swirldslabs is correct that we should not use protocol buffer maps in state due to the specified non-deterministic characteristic of protocol buffer map iteration.

edward-swirldslabs: virtual map. will update.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
edward-swirldslabs: retry until the public key is committed to state.   

There are edge cases where events might become ancient before they come to consensus and their transactions are not processed. 

Avoid flooding by making the timeout period large, like 5 minutes.

anthony-swirldslabs (approved):
+1

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,238 @@
...
+The `TssBaseService` state will be extended with a new map of the public `tssEncryptionKey` for each node.
+
+1. The keys of the map are Longs/UInt64 for the `nodeId` of the nodes
+2. The values of the map are the latest `TssEncryptionKeyTransaction` from each node.
+
+Data Lifecycle:
+
+* When a node id is not present in any of the `active rosters` or a `candidate roster`, the entry for that node id
+  should be removed. This cleanup task could be eagerly triggered any time there is a roster state change or lazily
+  triggered whenever a candidate roster is set. If this cleanup happens at startup, it must happen through a schema.
```
mhess-swl, line 190:
```suggestion
  triggered whenever a candidate roster is set. If this cleanup happens at startup, it must happen through a schema in order for the corresponding state changes to be made visible in the output stream.
```

mhess-swl (commented):
Overall LGTM! Just a couple minor comments

jsync-swirlds (commented):
Just a few minor questions.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+
+At the initialization of the `TssBaseService` the loaded public/private `tssEncryptionKey` will be provided. The
+private key will always be provided, but the public key may be null. There are three scenarios:
+
+1. The public `tssEncryptionKey` is provided indicating a new key-pair was generated.
+   1. Once execution is live, a `TssEncryptionKeyTransaction` will need to be sent periodically until the state
+      reflects the correct public key.
+2. The public `tssEncryptionKey` is not provided indicating only the private key was loaded from disk.
+   1. If there is no public key in the state, or the public key in the state does not matching the loaded
+      private key, a new public/private `tssEncryptionKey` will need to be generated through a call back to
```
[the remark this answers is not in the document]
edward-swirldslabs: If the public key is not in the state, then no one on the network has the public key.   We could use the TSS-Library to generate the public key and that would allow us to use the existing private key on disk.   But whether we commit to the private key on disk or generate a new one, it will have the same impact to the network and other nodes.  We will need to gossip a transaction to communicate the public key to the network.  

The decision about whether or not to use the private key on disk would be a security based decision.  Rotating the key might be the safest thing to do and avoid the risk of adopting a compromised key.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+* Normal execution can resume when private shares can be decrypted from the `TssMessages` belonging to the
+  latest `active roster`.
+
+### Public API
+
+#### TssEncryptionKeyTransaction Protobuf
+
+The `TssEncryptionKeyTransaction` will only appear in events created from the node sending the transaction. The
+signature on the event and the creator identifier of the event are sufficient to provide authenticity of the
+transaction.
```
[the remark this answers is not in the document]
edward-swirldslabs: I will update for clarity

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+
+1. The private `tssEncryptionKey` is loaded from disk and there is no public `tssEncryptionKey`.
+2. The private and public `tssEncryptionKey` were generated, are present, and match each other.
+
+#### TssBaseService Initialization
+
+At the initialization of the `TssBaseService` the loaded public/private `tssEncryptionKey` will be provided. The
+private key will always be provided, but the public key may be null. There are three scenarios:
+
+1. The public `tssEncryptionKey` is provided indicating a new key-pair was generated.
```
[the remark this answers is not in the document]
edward-swirldslabs: Storing the public key on disk will break the symmetry with the RSA key.  The RSA public certificates are only stored and used from the Hedera Address Book on the network.  

We have a way to generate the public BLS key from the private BLS key.  I will modify the proposal so that if the private key is loaded from disk, the public key is created from it.   https://github.com/hashgraph/hedera-cryptography/blob/b518f2cfb3e7d664037542a45ad79e820ffcdc11/cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/bls/BlsPrivateKey.java#L100

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -0,0 +1,233 @@
...
+2. Send (and periodically resend) a `TssEncryptionKeyTransaction` until the public key is updated in the state.
+3. Normal execution of the `TssBaseService` cannot resume until a new candidate roster has been keyed and adopted
+   that provides private shares to this node that can be decrypted using the new `tssEncryptionKey`.
+
+New `TssEncryptionKeyTransaction` handle workflow:
+
+* Upon receipt of a `TssEncryptionKeyTransaction`:
+  1. Add the public key to the `TssBaseService` state.
+  2. Disable any mechanism created for resending the transaction.
+  3. Check condition for resuming normal execution.
```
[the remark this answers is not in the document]
edward-swirldslabs: I will update the wording for clarity.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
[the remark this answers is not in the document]
edward-swirldslabs: After Jsync's feedback below, we'll render the public key from the loaded private key.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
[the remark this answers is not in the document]
mxtartaglia-sl: This makes sense.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -178,17 +173,18 @@ message TssEncryptionKeyTransaction {
 
 #### TssBaseService New State
 
-The `TssBaseService` state will be extended with a new map of the public `tssEncryptionKey` for each node.
+The `TssBaseService` state will be extended with a new virtual map of the public `tssEncryptionKey` for each node.
 
-1. The keys of the map are Longs/UInt64 for the `nodeId` of the nodes
+1. The keys of the map are UInt64 for the `nodeId` of the nodes
```
anthony-swirldslabs, line 178: Minor: while personally I like longs/uint64 too, I've heard many people find it useful and more type-safe to actually use a "node id" abstraction for representing node ids. So if we already have a protobuf that represents a node id, we might as well use it here, too.

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
```diff
@@ -198,16 +194,17 @@ No new configuration is needed.
...
 A health metric should be created to indicate the state the `TssBaseService` is in:
 
-1. distributing the public `tssEncryptionKey` to the network.
-2. public `tssEncryptionKey` is good, waiting for the next candidate roster to be keyed and adopted.
-3. Able to participate in signing with the active roster. (Private shares are decryptable with current
-   `tssEncryptionKey`.)
+1. State 1: distributing the public `tssEncryptionKey` to the network.
+2. State 2: the latest `tssEncryptionKey` is in the state, waiting to receive private shares.
+3. State 3: Private shares exist for this node in the latest `active roster` and the `TssBaseService` can execute
+   normally.
```
anthony-swirldslabs, line 200: Minor: "shares exist ... in ... roster" sounds a bit confusing because the roster doesn't actually contain any shares. I suggest to rephrase this for clarity and perhaps refer to a specific store/state where the private shares are actually stored by the TSS service.

anthony-swirldslabs (approved):
+0.(9)

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
jsync-swirlds: Node ID is a `uint64` everywhere.

anthony-swirldslabs: Good to know. Thanks for confirming.

mxtartaglia-sl (approved):

platform-sdk/docs/proposals/TSS-Encryption-Key-Automation/TSS-Encryption-Key-Automation.md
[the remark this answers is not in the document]
mhess-swl: Ah I was incorrectly assuming we would generate the public key from the loaded private key, which we will do, but wasn't a valid assumption when I first read this 😅

anthony-swirldslabs (approved):
+1

mxtartaglia-sl (approved):

Issue #16483, opened 2024-11-07
ETH Transaction Zero HAPI fees on success: Add test coverage


stoyanov-st:
Add tests to verify that we no longer charge the relay accounts on successful EthereumTransactions.
Add tests to verify the we still charge the relay accounts on unsuccessful EthereumTransactions.

PR #16579, opened 2024-11-13
fix: Zero HAPI fees for successful Ethereum Transaction relayers

### LINKED ISSUE #16482: Implement the change for zero charge to the relayer
In order to skip the charge for the relayer account we need to change the way we calculate the fee for Ethereum transactions. 
In the `calculateFees()` method of the `EthereumTransactionHandler` we are charging the payer account ( in this case this is the relayer account) with HAPI fee for the network. We want to skip this charge when the transaction is successful and only charge it for aborted/failed ones.

stoyanov-st:
**Description**:

This PR aims to remove the HAPI fee that the relayers pay for transactions.
We are now charing only if the transaction failed.
Note that the new logic is gated behind a feature flag.

**Related issue(s)**:

Fixes #16482 
#16483 

**Notes for reviewer**:
This fix is applied only for the dev env.

**Checklist**


hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
```diff
@@ -179,6 +180,12 @@ public void handleThrottled(@NonNull final HandleContext context) {
     @Override
     public Fees calculateFees(@NonNull final FeeContext feeContext) {
         requireNonNull(feeContext);
+        final var zeroHapiFeeEnabled =
+                feeContext.configuration().getConfigData(ContractsConfig.class).evmEthTransactionZeroHapiFeesEnabled();
+        return zeroHapiFeeEnabled ? Fees.FREE : getLegacyCalculateFees(feeContext);
```
david-bakin-sl, line 185: I think it's unfortunate we can't use the "proper" designed fee mechanism because it doesn't handle different costs for success vs failed. 

Anyway, this line with `Fees.FREE` make it clear the fee isn't charged _here_ (when the config option is set).  Consider a comment here pointing off to `ContextTransactionProcessor` esp. the method `chargeOnFailedEthTxn` which is where the fee is charged in case of failure.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
```diff
@@ -236,7 +254,44 @@ void stillChargesHapiFeesOnAbort() {
     }
 
     @Test
-    void stillChargesHapiFeesOnHevmException() {
+    void chargeHapiFeeOnFailedEthTransaction() {
```
david-bakin-sl, line 257: Tests here don't seem to test with the new flag _both_ disabled and enabled.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/handlers/EthereumTransactionHandlerTest.java
```diff
@@ -67,6 +67,7 @@
 import com.hedera.node.app.spi.workflows.PreHandleContext;
 import com.hedera.node.config.data.ContractsConfig;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.config.api.Configuration;
```
david-bakin-sl, line 70: Could there be a test (or pair of tests) for `calculateFees` both ways, new flag set and not set?

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
```diff
@@ -227,9 +229,8 @@ final Stream<DynamicTest> baseRelayerCostAsExpected() {
                         .maxPriorityGas(2_000_000L)
                         .gasLimit(1_000_000L)
                         .sending(depositAmount),
-                // The relayer's cost to transmit a simple call with sufficient gas allowance is ≈ $0.0001
-                getAccountInfo(RELAYER)
-                        .has(accountWith().expectedBalanceWithChargedUsd(ONE_HUNDRED_HBARS, 0.0001, 0.5)));
+                // The relayer's cost is no longer applicable for successful transactions.
```
david-bakin-sl, line 232: Shouldn't these tests here be contingent on the new flag being set?  Which you need to specify explicitly in the test?   And contriwise - shouldn't there still be a test for the new flag _not_ set?

david-bakin-sl (commented):
Unit and BDD testing don't seem complete because (if I'm reading this correctly) they only now test the case where the zero hapi fees config is _enabled_, not for it being _disabled_.    ??

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
stoyanov-st: Considering that we will be removing the flag when we verify that this is the behaviour we want, I don't see point in investing more time to create more tests here that would be removed in the near future.

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/handlers/EthereumTransactionHandlerTest.java
stoyanov-st: Added the tests

hedera-node/hedera-smart-contract-service-impl/src/test/java/com/hedera/node/app/service/contract/impl/test/exec/ContextTransactionProcessorTest.java
stoyanov-st: extended existing tests to verify this depending on the config

hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/ethereum/EthereumSuite.java
david-bakin-sl: ok

david-bakin-sl (dismissed):
lgtm

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/handlers/EthereumTransactionHandler.java
stoyanov-st: Added Javadoc explaining the situation and pointing to the method. Thanks!

hedera-node/configuration/previewnet/application.properties
```diff
@@ -7,6 +7,7 @@
 bootstrap.genesisPublicKey=c249a323c878f5b5e2daccda6d731e6fdc32f870228d1cd4fae559d947dbc36c
 contracts.chainId=297
 contracts.systemContract.updateNFTsMetadata.enabled=true
+contracts.evm.ethTransaction.zeroHapiFees.enabled=true
```
lukelee-sl, line 10: you might want to check to see if this is the appropriate place to set this to true if your intention is to only turn this on for local node.

Nana-EC: I think this was supposed to be the dev file so that it's only on for local node and not yet on for any public env

Nana-EC (approved):
Signing off on config and call logic change.

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -185,6 +191,8 @@ private CallOutcome maybeChargeFeesAndReturnOutcome(
             gasCharging.chargeGasForAbortedTransaction(
                     senderId, hederaEvmContext, rootProxyWorldUpdater, hevmTransaction);
         }
+
+        chargeOnFailedEthTxn(hevmTransaction);
```
lukelee-sl, line 195: it seems that `gasCharging.chargeGasForAbortedTransaction` has the potential to charge the relay intrinsic gas if the transaction is an `EthereumTransaction`.  Given that right now `chargeGas` is only true if the transaction is a `ContractCall`, should we consider removing that clause from the if statement?  

I kind of do not like the fact that we now have two places where we charge someone fees (although the amounts are different - intrinsic gas vs $0.0001 converted to hbars).  Can the logic of the two calls be consolidated so that there is only one method that charges fees to the caller or the relay?

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
```diff
@@ -145,6 +147,10 @@ public CallOutcome call() {
                 result = result.withSignerNonce(sender.getNonce());
             }
 
+            if (!result.isSuccess()) {
```
lukelee-sl, line 150: I wonder if this if statement can/should be combined with the if statement below?

stoyanov-st: Makes sense but I think we should first verify that this behaviour is the one we are targeting and get back to it to refactor it properly.

lukelee-sl: yes, agreed

hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/ContextTransactionProcessor.java
stoyanov-st: For the first question I think the if statement you're mentioning here is out of scope for this task but we can definitely revisit to make sure it is correctly charging for gas.

For the second question - we can add a follow-up PR to address this. I'm tilting towards refactoring it but we should make sure that it is super clear that we might be charging for gas AND hapi fees.

lukelee-sl (approved):
Let's see if it makes sense to do some refactoring later.  LGTM

tinker-michaelj (approved):
LGTM, tyvm @stoyanov-st !

PR #16581, opened 2024-11-13
feat: Network and NodeMetadata Protobuf 

### LINKED ISSUE #16549: Create `Network` and `NodeMetadata` protobuf representation
**Background**
In order for config.txt to be replaced, a new representation in protobuf format should be defined which includes all "biographical information" on each node that is needed by a Hedera network with _TSS fully operational_.

See [Roster Lifecycle Implementation](https://www.notion.so/swirldslabs/Roster-lifecycle-implementation-13b7c9ab25918069a0afcc44713015b9) for the same details listed below.

```java
/**
 * The full information needed to bootstrap a network with TSS enabled
 * and the legacy address book completely replaced by rosters.
 */
message Network {
  /**
   * The metadata of the nodes in the network.
   */
  repeated NodeMetadata node_metadata = 1;

  /**
   * If set, a list of TSS messages that have been generated by the nodes in
   * a target network that is receiving a "transplant state". This lets us
   * preserve the ledger id of the target network even if the transplant
   * state has no keys (or even roster entries) in common with the target.
   * This must be exactly the ordered list of TSS messages used to generate
   * the target network's ledger id; and must generate the same ledger id
   * given below.
   */
  repeated com.hedera.hapi.services.auxiliary.tss.TssMessageTransactionBody tss_messages = 2;

  /**
   * If set, the expected ledger id of the network.
   */
  bytes ledger_id = 3;
}

/**
 * The full information needed for a node.
 */
message NodeMetadata {
  /**
   * The metadata of the node.
   */
  Node node = 1;
  
  /**
   * If set, the public part of a prescribed TSS encryption key for the node.
   */
  bytes tss_encryption_key = 2;
}

``` 

**Acceptance Criteria**
- Protobuf definitions added to `hedera-services`

**Dependencies**
None

thomas-swirlds-labs:
**Description**:

- Defines a protobuf-based network object to replace the legacy `config.txt` file, fully encapsulating the information required to bootstrap a network with Threshold Signature Scheme (TSS) enabled and with the legacy address book completely replaced by Roster

**Design and Motivations**:

- Eliminates the requirement for separate overrides or genesis configurations, providing a standardized format for storing metadata about all nodes in a network, including `TssEncryptionKeys`, Roster entries, and `ledgerIds`. 
- **Note:** this is a format for data that exists on disk, not in state or block streams, and can be regenerated during state freeze operations for network resets.
- Each node in the network is represented by `NodeMetadata`, which includes fields such as the `TssEncryptionKey` (now removed from the Roster) and other information previously found in the Address Book. 
- This ensures compatibility for both Genesis startup (networks starting without state) and Network Transplant (starting from a previously saved state from another network). 
- `tssEncryptionKey` is required here because it has been removed from the Roster and will need to be create here with Tss Key Material 
- `LedgerId` is required, because it needs to be validated against when you start the new network

**Related issue(s)**:

Fixes #16549


hapi/hedera-protobufs/services/state/network/network.proto
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * The full information needed for a node.
+ */
+message NodeMetadata {
+  /**
+   * The metadata of the node.
+   */
+  com.hedera.hapi.node.state.addressbook.Node node = 1;
```
jsync-swirlds, line 78: Instead of the node itself, wouldn't it be more efficient to include the node _ID_ here?

hapi/hedera-protobufs/services/state/network/network.proto
```diff
@@ -0,0 +1,84 @@
...
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.network">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+
+
+/**
+ * The full information needed to bootstrap a network with TSS enabled
+ * and the legacy address book completely replaced by rosters.
```
jsync-swirlds, line 46: Style is off here.  We're falling back into "describe" habit rather than "specify" habit.
In addition to _adding_ requirements below this, this description should be reworded
```suggestion
 * A full description of the Network of Consensus Nodes.<br/>
 * This records the full information needed to bootstrap a network
 * with TSS enabled and the legacy address book completely replaced
 * by rosters.
 *
 * TBD Need requirements detailed here, including block stream effects.
```

hapi/hedera-protobufs/services/state/network/network.proto
jsync-swirlds, line 1: All of the messages and fields need full specification, rather than just brief description.
In particular the specification should clarify why we need more data stored in state and added to the block stream for these rather than adding the 1 or two additional fields to existing messages.

jsync-swirlds (commented):
It isn't entirely clear (even after reading the linked issue and further linked Notion page) what these messages are for.

General _suggestions_:
1. Detail in the PR description all of "what", "why", and "how" of the PR.
   * What is being done here, design/feature wise
   * Why is this needed, and why is it designed _in this way_
   * How is this to be used and how does it meet future needs
1. Make sure that `proto` file specification always explains all of the relevant requirements for each message and field.
1. In a proto _file_ comment, add any relevant additional detail, possibly including the "what/why/how" from above, as applied to that particular file.
1. Writing too much description is, in my experience, almost never a problem. The _common_ problem is PRs, specifications, design documentation (i.e. code files), and related text that is woefully short and imprecise.

hapi/hedera-protobufs/services/state/network/network.proto
thomas-swirlds-labs: Good point. Have updated

hapi/hedera-protobufs/services/state/network/network.proto
```diff
@@ -52,33 +63,53 @@ message Network {
...
+    * An elliptic curve public encryption key.<br/>
+    * If set, the public part of a prescribed TSS encryption key for the node.
+    * This is currently an ALT_BN128 curve, but the elliptic curve
+    * type may change in the future. For example,
+    * if the Ethereum ecosystem creates precompiles for BLS12_381,
+    * we may switch to that curve.
+    * <p>
+    * This value SHALL be specified according to EIP-196 and EIP-197 standards,
+    * See <a href='https://eips.ethereum.org/EIPS/eip-196#encoding'>EIP-196</a> and
+    * <a href='https://eips.ethereum.org/EIPS/eip-197#encoding'>EIP-197</a><br/>
```
jsync-swirlds, line 110: I believe this has changed.  TSS library team is saying their keys are raw binary as used by the arkworks rust code and not encoded according to the EIP standards.

hapi/hedera-protobufs/services/state/network/network.proto
thomas-swirlds-labs: Thanks @jsync-swirlds. Have updated with more detailed specifications now 👍

thomas-swirlds-labs:
> It isn't entirely clear (even after reading the linked issue and further linked Notion page) what these messages are for.
> 
> General _suggestions_:
> 
> 1. Detail in the PR description all of "what", "why", and "how" of the PR.
>    
>    * What is being done here, design/feature wise
>    * Why is this needed, and why is it designed _in this way_
>    * How is this to be used and how does it meet future needs
> 2. Make sure that `proto` file specification always explains all of the relevant requirements for each message and field.
> 3. In a proto _file_ comment, add any relevant additional detail, possibly including the "what/why/how" from above, as applied to that particular file.
> 4. Writing too much description is, in my experience, almost never a problem. The _common_ problem is PRs, specifications, design documentation (i.e. code files), and related text that is woefully short and imprecise.

@jsync-swirlds. Agreed. I've updated the description of the PR to include more detailed description of the motivation and design and give some more background on the what/why/how. Hopefully that should be a little clearer now?

petreze (approved):

hapi/hedera-protobufs/services/state/network/network.proto
jsync-swirlds, line 1: A few suggestions:
1. If this is purely for on-disk data, it doesn't belong in the `state` folder.  Placing it in state implies it will be stored in state; and it is not (cannot be) designed properly for both uses.
1. Why are we even using protobuf for this?  On-disk startup files should be human readable and human editable to support operations teams, and protobuf is not a good format for that purpose.  I would think a YAML file would be the best option, closely followed by JSON.
1. It should be clearly documented in these files that this is an on-disk configuration file definition.
1. The HAPI documentation guidelines _do not apply_ to on-disk data formats; but that requires this definition to be well separated from the HAPI files (not in the `hapi` module).

edward-swirldslabs: Can we specify the structure of the file in protobuf and still have a JSON representation on disk?     Could it even be YAML, translated to JSON, then parsed by protobuf?

jsync-swirlds: It is _possible_ to output JSON from a protobuf specification, but it's not a _production_ option; that format is intended for debugging and development.  In particular it's not guaranteed to round-trip deterministically.

It might be much easier to just create a Java `record` class and have that output and input as YAML, using whichever YAML processor we prefer.

Protobuf-JSON format _might_ work; I cannot say for certain either way, but it's definitely not what I would _recommend_.

One other thought, this is an internal private detail of the consensus node.
That _should not_ be published in the HAPI API or presented to SDKs in the `hedera-protobufs` repo.  As such (and as noted above) this definition probably belongs in a different module (hedera-app, or perhaps the network service?).

edward-swirldslabs: If we are producing this file by hand, that may be problematic.   It's meant to be an automated export (probably at startup) and consumed as an `override-network` in the transplant and reset  use case.    With the TSS information, there's a lot that can go wrong.  It's all binary data and doesn't go well for by-hand assembly.

jsync-swirlds: The TSS data is binary, and probably should be retained as such (e.g. base64 encoded).
The Node data, however, very likely needs to be edited every time (because the destination network won't have the same nodes), so the overall file format still needs to be editable.

Neeharika-Sompalli: Yeah the reason I remember, we added a protobuf object was to be able to easily parse TSS key data. @tinker-michaelj  thoughts ?

thomas-swirlds-labs: Addressed in latest commits. Since moving to another module outside of `hapi` is non-trivial in terms of the impact to `gradle config` and `pbj plugin` setup, for now we have decided to move to a new `internal-protobufs` folder. The implementation of the file format on disk will be done in a follow up PR once that decision has been finalized

edward-swirldslabs (dismissed):

jsync-swirlds (dismissed):
Well executed and thorough.
Thanks.

andrewb1269hg (approved):
Review applies to:
`hapi/build.gradle.kts`
`hedera-node/hedera-app-spi/build.gradle.kts`

jsync-swirlds (approved):

andrewb1269hg (approved):

Issue #16568, opened 2024-11-12
Remove dependency on `swirlds-state-impl` from `swirlds-platform-core`


imalygin:
This is a follow-up for https://github.com/hashgraph/hedera-services/issues/16322

`PlatformMerkleStateRoot` extends `MerkleStateRoot` but it shouldn't really be the case because it unnecessarily creates a dependency on the implementation of the `com.swirlds.state.State` interface. 
What we _do_ need is an implementation of `com.swirlds.platform.system.SwirldState` that contains an instance of the `com.swirlds.state.State` and instance of `com.swirlds.platform.state.MerkleStateLifecycles`. Then, it should have the implementation of `SwirldState` interface methods that looks as follows:

```
    void preHandle(final Event event) {
        lifecycles.onPreHandle(event, state);
    }

    void handleConsensusRound(final Round round, final PlatformStateModifier platformState) {
         throwIfImmutable();
         lifecycles.onHandleConsensusRound(round, this);
    }

    void sealConsensusRound(@NonNull final Round round) {
            requireNonNull(round);
            throwIfImmutable();
           lifecycles.onSealConsensusRound(round, state);
    }
```

and so on. Then this implementation should replace usages of `PlatformMerkleStateRoot`, and `PlatformMerkleStateRoot`, in its turn, should be removed. 

This refactoring will allow to get rid of `swirlds-state-impl` dependency. The common wisdom is that whilst app can depend on the implementation details of the state (and therefore the module), the platform has absolutely to reason for this dependency.

### Tasks
- https://github.com/hashgraph/hedera-services/issues/17353
- https://github.com/hashgraph/hedera-services/issues/17357
- https://github.com/hashgraph/hedera-services/issues/17868

imalygin:
Blocked by https://github.com/hashgraph/hedera-services/issues/16322

imalygin:
To complete `PlatformMerkleStateRoot` removal this task -  https://github.com/hashgraph/hedera-services/issues/16436 - has to be completed first.

PR #16592, opened 2024-11-13
docs: 16193 Lifecycle management for the State

### LINKED ISSUE #16193: Create a design doc outlining all the neccesary changes required for this module
The design doc can be in notion since we do not expect it to be long living. It should include:

1. Descriptions of all Jasper's wishes and use cases for it
2. A clear list of all changes that are necessary to achieve 1 including which packages need to move
3. State machine diagram, or similar, showing the lifecycle of a state and what methods transition them
4. Decide and document if there should be two modules or one (impl & api)

[Working diagram for context](https://drive.google.com/file/d/10N4V4Jr61B4OhbdSIMk5_futD8b8HlqL/view?usp=drive_link)

Jasper should review this document and approve before this ticket is closed.

imalygin:
**Description**:

This PR adds a design proposal for the state lifecycle management

**Related issue(s)**:

Fixes #16193 

**Notes for reviewer**:

**Checklist**


platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+These methods correspond to the state lifecycle as follows:
+
+[![State lifecycle](state-lifecycle.svg)](state-lifecycle.svg)
+
+Note that the `loadSnapshot` method doesn't really belong to `com.swirlds.state.State`, as it's a method of the class, and in some cases, we may need to load the snapshot before we have an instance of the state. 
+This method should be moved to a separate class that will be responsible for managing the state lifecycle.
+
+- `SwirldStateManager` needs refactoring. The current implementation has two sets of responsibilities:
+  - Keeping track of the references to the latest immutable state and the latest mutable state. This includes creating a copy of the state and updating the references.
```
artemananiev, line 40: This responsibility and related methods should be a part of `swirlds-state-api` module. The other one will be kept in `swirlds-platform-core`

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+  - Handling Platform-related events:
+    - Handling consensus rounds.
+    - Sealing consensus rounds.
+    - Keeping the `freeze time` parameter up-to-date.
+    - Providing state for signing.
+    - Initializing the state by extracting it from the signed state.
+
+These two sets of responsibilities should be separated. The `swirlds-state-impl` should not have any details related to the platform. However, it should provide all necessary levers for the platform to interact with the state and its lifecycle.
+
+Therefore, `swirlds-state-impl` needs a `StateManager` class that will have the following responsibilities:
```
artemananiev, line 50: Does it make sense to have a state manager interface in the `swirlds-state-api` module, and one more class that implements the interface in `swirlds-state-impl`?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+Additionally, there should be a configuration parameter that limits the number of states kept in memory. When the limit is reached, the oldest state should be evicted.
+
+The interface should look as follows:
+
+```java
+public interface StateManager {
+    State getLatestImmutableState();
+    
+    State getLatestMutableState();
```
artemananiev, line 64: It can be just `getMutableState()`, since there is only one mutable state and it's always the latest.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+    
+    State getState(long round);
+    
+    void evictState(long round);
```
artemananiev, line 68: This one is questionable... In the current implementation, states are released using ref counting mechanism. Will it be different in `swirlds-state-impl`?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+Therefore, `swirlds-state-impl` needs a `StateManager` class that will have the following responsibilities:
+- Keeping track of the references to the latest immutable state and the latest mutable state.
+- Tracking states that existed before the latest immutable state. These should be available for lookup by their round number.
+- Evicting old states that are no longer needed.
+- Loading snapshots from the disk.
+
+Additionally, there should be a configuration parameter that limits the number of states kept in memory. When the limit is reached, the oldest state should be evicted.
+
+The interface should look as follows:
```
poulok, line 58: Something we've learned in these proposals is that we should be flexible with the exact APIs. I suggest changing this sentence to this to make it more flexible so that if something small changes, you don't need to come back and change this section:
```suggestion
The interface will look something like this:
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
poulok: Yes, we want to get rid of the reference counting all together. The state module will retain a state in memory until it is told to release it, or possibly until it reached the configured maximum age, although i'm not sure about this second part.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Design & Architecture
+
+### Java classes
+
+- `com.swirlds.state.State` has the following lifecycle-related methods:
+  - `State copy()`: creates a copy of the state, which is mutable. The source of the copy becomes immutable.
+  - `void computeHash()`: computes the hash of the immutable state.
+  - `void createSnapshot(Path targetPath)`: creates a snapshot of the state at the specified path.
+  - `State loadSnapshot(Path sourcePath)`: loads the snapshot from the specified path.
```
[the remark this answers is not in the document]
poulok: I think it can be static, at least for now. When we get to supporting multiple states from different state families in memory at the same time, we may decide to change it.

platform-sdk/docs/proposals/state-lifecycle/state-lifecycle.svg
poulok, line 1: How is a genesis state provided?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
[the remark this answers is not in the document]
imalygin: As of now, this method is not static. Kelly is right, technically it can be static. However, it would be suboptimal because we'd have to move `MerkleTreeSnapshotReader` class to `swirlds-state-api`. As the name suggest, it knows about the implementation details of the `State` and therefore it doesn't belong to `swirlds-state-api`. 

I'd like to move this method to the `StateLifecycleManager` because we don't want this logic to require an instance of `State` object.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Sure. I'll mention that explicitly.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+These methods correspond to the state lifecycle as follows:
+
+[![State lifecycle](state-lifecycle.svg)](state-lifecycle.svg)
+
+Note that the `loadSnapshot` method doesn't really belong to `com.swirlds.state.State`, as it's a method of the class, and in some cases, we may need to load the snapshot before we have an instance of the state. 
+This method should be moved to a separate class that will be responsible for managing the state lifecycle.
+
+- `SwirldStateManager` needs refactoring. The current implementation has two sets of responsibilities:
```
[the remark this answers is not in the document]
imalygin: >The Swirld prefix (e.g. SwirldState, SwirldStateManager) is confusing. My understanding is it actually means "platform". It would be great to rename the classes accordingly in the future

Yes, I agree. Whenever I move the functionality to the `swirlds-state-api` module, I drop this prefix. Generally, I agree - we should get rid of this prefix in the `swirlds-platform-core` as well

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: The idea was to have an implementation only, but you're right - we'd better have `StateManager` as an interface and `StateManagerImpl` as an implementation

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Agreed

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I'll explicitly mention that `evictState` is there as a replacement for the reference counting mechanism

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: :) fair enough. I'll rephrase it

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+    
+    State getState(long round);
+    
+    void evictState(long round);
+    
+    State loadSnapshot(Path sourcePath);
+}
+```
+
+Note that the loaded snapshot should be treated as one of the immutable states and should be available via `getState`.
```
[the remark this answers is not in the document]
imalygin: >Should a call to evictState() on the mutable state should fail?

Yes, absolutely. I can't imagine a scenario where it would be appropriate. 

>How will reconnect work with this lifecycle manager?

I think it should work similarly to how it works now. @artemananiev  correct me if I'm wrong, but here is how I understand it: 
1. We load the latest snaphot
2. We create a copy of it to get a mutable instance
3. We apply a set of changes to the state as a part of the reconnect process. 

If I'm not mistaken, we don't need to set the state explicitly from the outside. `StateManager` can take care of it.

platform-sdk/docs/proposals/state-lifecycle/state-lifecycle.svg
imalygin: Good point. Probably, it should be a part of the `StateManager` functionality.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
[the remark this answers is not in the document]
poulok: Yes, you are correct for how it works. I was not thinking about it correctly.

platform-sdk/docs/proposals/state-lifecycle/state-lifecycle.svg
poulok: How does it work now? Is there application specific logic for determining what the genesis state contains? Is the state module capable of creating it itself? Or does the State Module create a totally empty one and provide it to the app for filling in the initial details? If it's the latter, i think it should be called `createEmptyState()` or similar.

imalygin: >Is there application specific logic for determining what the genesis state contains? 

The creation of the genesis state doesn't depend on the app or even the platform. However, there is post-creation initialization logic, and that logic belongs to the app. This logic involves registering services in the state, and only the app knows what services to register. 

So, technically the state will be able to create itself after this refactoring is complete https://github.com/hashgraph/hedera-services/issues/16568

>i think it should be called createEmptyState() or similar.

Agreed.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,98 @@
...
+- Tracking states that existed before the latest immutable state. These should be available for lookup by their round number.
+- Restricts mutability of the state to a single object. That is, only one state object should be mutable only once.
+- Evicting old states that are no longer needed. `evictState` method *must* fail if the state is still mutable.
+- Loading snapshots from the disk.
+
+Additionally, there should be a configuration parameter that limits the number of states kept in memory. When the limit is reached, the oldest state should be evicted.
+Currently, we rely on the implicit garbage collection. We have a reference count mechanism to achieve this.
+We want to get rid of this mechanism and rely on the explicit eviction mechanism.
+
+The following diagram illustrates the proposed illustrate management:
```
poulok, line 64:
```suggestion
The following diagram illustrates an example of the states in memory throughout their lifecycles:
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,98 @@
...
+    - Handling consensus rounds.
+    - Sealing consensus rounds.
+    - Keeping the `freeze time` parameter up-to-date.
+    - Providing state for signing.
+    - Initializing the state by extracting it from the signed state.
+
+These two sets of responsibilities should be separated. The `swirlds-state-impl` should not have any details related to the platform. However, it should provide all necessary levers for the platform to interact with the state and its lifecycle.
+
+Therefore, `swirlds-state-api` needs a `StateManager` interface with an implementation `StateManagerImpl` in `swirlds-state-impl`  that will have the following responsibilities:
+- Genesis state initialization.
```
poulok, line 53:
```suggestion
- Create an empty state
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,98 @@
...
+- Keeping track of the references to the latest immutable state and the latest mutable state.
+- Tracking states that existed before the latest immutable state. These should be available for lookup by their round number.
+- Restricts mutability of the state to a single object. That is, only one state object should be mutable only once.
```
poulok, line 56:
```suggestion
- Restricts mutability to a single state. That is, only one state object should be mutable at any given time.
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,98 @@
...
+- Evicting old states that are no longer needed. `evictState` method *must* fail if the state is still mutable.
```
poulok, line 57:
```suggestion
- Evicting old states from memory that are no longer needed. `evictState` method *must* fail if the state is still mutable.
```

poulok (dismissed):
+1. I left a few minor wording clarifications, but looks great! Be sure to get @jasperpotts to take a look before merging.

Please also ensure that someone from the Block Node team reviews and understands since they are another stakeholder.

poulok (dismissed):

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
+**Imporatant**: It's up to the client code to make sure that the state is released once it's no longer needed.
+Without this, the state will not be evicted from memory and eventually the application will run out of memory.
 
 The following diagram illustrates an example of the states in memory throughout their lifecycles:
 
-[![Multistate management](multi-state-management.png)](multi-state-management.png)
+[![Multistate management](multi-states.svg)](multi-states.svg)
+
+Also, `swirlds-state-api` needs a `SnapshotManager` interface with an implementation `SnapshotManagerImpl` in `swirlds-state-impl` that will have the following responsibilities:
+- Load snapshots from the disk. A state loaded from a snapshot must be immutable and available for look up by the round number.
```
poulok, line 77:
```suggestion
- Load snapshots from the disk. A state loaded from a snapshot is immutable and available for look up by the round number.
```
The method to get a state by round number was removed which I think is correct. This verbage needs to be updated to say that the snapshot loaded from disk will be fast copied. The immutable version loaded from disk will be available via the call to `getLatestImmutableState()` and the mutable copy will be available via `getMutableState()`

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
+- Evict the loaded state by the round number when the state is no longer needed.
```
poulok, line 78: This has been removed, so this line should also be removed

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
-    void evictState(long round);
-    
+public interface SnapshotManager {
+
+    State getLoadedState(long round);
+
     State loadSnapshot(Path sourcePath);
+
+    void evictLoadedState(long round);
 }
```
poulok, line 95: I don't agree with the splitting of methods this way. The idea of the snapshot manager was to deal with instances of `Snapshot` which will be a new class that loads minimal information about a state from disk like the hashes that can be used to produce a merkle proof, but not load the entire state into memory. Loading an entire state into memory and providing it as a `State` object is a function of a `StateManager` or `States` object.

The `loadSnapshot` method should probably be `loadStateFromDisk` to differentiate loading the minimal data from loading the full state into memory.

Also, I don't see a reason for the `evict` or `getLoadedState` method anymore. When a state's references go to zero, it will should be GCed. And if the user of the module need to hang onto a state for collecting signatures or anything like that, it's going to have to keep track of the reference itself to prevent it from being GCed, therefore this method is not needed.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
 
-Additionally, there should be a configuration parameter that limits the number of states kept in memory. When the limit is reached, the oldest state should be evicted.
-Currently, we rely on the implicit garbage collection. We have a reference count mechanism to achieve this.
-We want to get rid of this mechanism and rely on the explicit eviction mechanism.
+    State getMutableState();
+}
+```
+
+**Imporatant**: It's up to the client code to make sure that the state is released once it's no longer needed.
+Without this, the state will not be evicted from memory and eventually the application will run out of memory.
```
poulok, line 70: Two comments:
1. Typo in "Important"
2. Nothing about reservations has been explained in this document so far so this note is not meaningful. You don't need to go into the details of how it works, but a short paragraph explaining that reference counts are used to determine when a state is eligible for GC would be helpful. Then you could also state that the state module will ensure the mutable and latest immutable states are held onto, but if the user of the library needs to keep a state around longer than that, it is their responsibility to ensure proper use of additional reservations. The CN will do with with the existing class structures of SignedState and ReservableSignedState that sit on top of these state reservations.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
poulok, line 78: I don't think we need this anymore

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
+Therefore, `swirlds-state-api` needs a `StateLifecycleManager` interface with an implementation `StateLifecycleManagerImpl` in `swirlds-state-impl`  that will have the following responsibilities:
+- Take initial state as a starting point. It can be a genesis state or a state loaded from a snapshot.
+- Keep track of the references to the latest immutable state and the latest mutable state.
+- Restrict mutability to a single state. That is, only one state object should be mutable at any given time.
+
+The interface will look something like this:
+
+```java
+public interface StateLifecycleManager {
+    void setInitialState(State initialState);
```
poulok, line 61: In the normal startup of a CN that has a state on disk, will this be how the state lifecycle manager gets initialized?

1. Load state from disk (I think this should be in the StateLifecycleManager, explained in a [comment](https://github.com/hashgraph/hedera-services/pull/16592/files/160388010a7ac16edfd958e431815dd40bb26080..5b058e400180e8d1a7fcc8c032de6f2072b56b7d#r1853936664) below
2. Use that state to invoke `setInitialState()` which then performs a fast copy, setting both the latest immutable state and the mutable state

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
-Note that the loaded snapshot should be treated as one of the immutable states and should be available via `getState`.
-
 ### Metrics
 
-We need to monitor the number of states that are kept in memory to ensure there is no memory leak.
+We need to monitor the number of states that are kept in memory by `SnapshotManager` to ensure there is no memory leak.
 
 ### Performance impact
 
 There is only one important performance consideration—the number of states that are kept in memory. However, this can be mitigated by the configuration parameter that limits the number of states kept in memory. In the absence of this parameter, it should be carefully tested and monitored to prevent memory leaks.
```
poulok, line 104: I think this part is now obsolete

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -33,66 +33,76 @@ These methods correspond to the state lifecycle as follows:
...
+    void evictLoadedState(long round);
 }
 ```
 
-Note that the loaded snapshot should be treated as one of the immutable states and should be available via `getState`.
-
 ### Metrics
 
-We need to monitor the number of states that are kept in memory to ensure there is no memory leak.
+We need to monitor the number of states that are kept in memory by `SnapshotManager` to ensure there is no memory leak.
```
poulok, line 100: I think this can be removed

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Sure, I will add information about reservations.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Yes, but we need self-management for the number of states in memory. I'll add more details.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: `StateLifecycleManager` will maintain a list of states previously loaded with `loadSnapshot` method. We need to monitor the size of this list to make sure that we don't have unused state accumulating in memory or they may be all in use but it's just too many of them which may cause OOM. 

Alternatively, we can get rid of `State getLoadedState(long round)` method and use `loadSnapshot` whenever we need to access a state on the disk.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Yes, this makes sense. In this case there is no need to expose `setInitialState`, as it should be a responsibility of the `StateLifecycleManager` to deal with it

imalygin: I'll update the doc accordingly

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Right now we don't maintain a list of the loaded states. There is no such functionality - we don't have a class that has `State getLoadedState(long round)` method. If we intend to have such a method, there should be some way to get rid of the states that are no longer needed, manual or automatic.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+     * @return
+     */
+    int getReservationCount();
+}
+
+```
+
+We're going to rely on this mechanism to manage the number of states that are kept in memory. 
+
+**Important**: It's up to the client code to make sure that the state is released once it's no longer needed.
```
poulok, line 113: Will the State Module ensure that a state is at least hashed before it is garbage collected even if it's reservations are at 0? Or is it entirely on the client to maintain reservations in all cases?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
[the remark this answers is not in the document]
poulok: Should this method still be here then? I agree, it should be in another class. `StateLifecycleManager` is good.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+
+- `com.swirlds.state.State` has the following lifecycle-related methods:
+  - `State copy()`: creates a copy of the state, which is mutable. The source of the copy becomes immutable.
+  - `void computeHash()`: computes the hash of the immutable state.
+  - `void createSnapshot(Path targetPath)`: creates a snapshot of the state at the specified path.
+  - `State loadSnapshot(Path sourcePath)`: loads the snapshot from the specified path.
+
+These methods correspond to the state lifecycle as follows:
+
+[![State lifecycle](state-lifecycle.svg)](state-lifecycle.svg)
```
poulok, line 127: This diagram and the classes now seem so to be a bit contradictory. Is describes a `State` class that has methods like `copy()` and `computeHash()`, yet these operations are described/depicted as part of the lifecycle of the state. Then there is the `StateLifecycleManager` that has `copy()`. In all cases, `copy()` can only be called on the mutable state. So we need to be very clear about which operations are part of the lifecycle and put those in the lifecycle manager and not duplicate the copy method.

poulok: I think that the hash method can only start part of the "lifecycle" if it is done automatically on the immutable state created on copy. However I don't think this will work because hashing is expensive and, at least on the consensus node, should not be performed on the same thread as the copy itself.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+
+There is an initiative to create self-contained State modules (`swirlds-state-api` and `swirlds-state-impl`) that would have minimal dependencies, and specifically, they should not depend on `swirlds-platform-core`. Part of this initiative is the implementation of a state lifecycle management mechanism that could be used by both the Consensus Node and the Block Node.
+
+### Requirements
+
+- Add an interface representing state lifecycle management and its implementation.
+- Update the current State API to be coherent with this State Lifecycle API.
+- Implement this set of APIs and migrate the existing code to use it.
+
+## Design & Architecture
```
poulok, line 22: I find myself wanting some summary of what the new functionality is at a level between the `Purpose and Context` and the `Design & Architecture` levels. Something that describes what this new feature will do before getting into the details like methods. What is meant by "lifecycle management" now that the client must maintain references?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Reservation count mechanism
+
+This mechanism is already implemented as a part of the Merkle tree implementation. However, it's importnant to mention it here
+to provide important context for understanding how exactly the state lifecycle management will work.
+
+Every Merkle tree node implements the following interface:
+
+```java
```
poulok, line 31: I think putting this code block here is not necessary, and is in fact distracting. It's large and takes up a lot of space, and breaks up my mental flow. I think a few sentences describing the basics of the reservations would be much better.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+
+- Add an interface representing state lifecycle management and its implementation.
+- Update the current State API to be coherent with this State Lifecycle API.
+- Implement this set of APIs and migrate the existing code to use it.
+
+## Design & Architecture
+
+### Reservation count mechanism
+
+This mechanism is already implemented as a part of the Merkle tree implementation. However, it's importnant to mention it here
```
poulok, line 26:
```suggestion
This mechanism is already implemented as a part of the Merkle tree implementation. However, it's important to mention it here
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
poulok: Somewhere, I'm not sure where but maybe here, it would be good to describe how the states flow through the CN along its lifecycle at a high level and how things will change with this new State API. For example, the fact that the CN will invoke the State Module to load a state from disk and to make fast copies when a round is done being handled, but still need to hold onto state objects with reservations to collect signatures. Then the equivalent for BN.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,231 @@
...
+
+Additionally, this logic is tightly coupled with other functionalities, such as:
+
+- State signatures
+- Platform service registration
+- Invalid state file cleanup
+- Hash validation
+- State reservation
+
+Invalid state file cleanup and hash validation should be moved to the `StateLifecycleManager` implementation. The remaining responsibilities belong to the platform code.
```
poulok, line 217: What kind of "invalid" checks? Will these checks be common to both CN and BN?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I see two possible approaches to that: 
- It may be entirely up to client to maintain reservations and make sure that they are acquired and released properly. In this case, `StateLifecycleManager` will not track if the state is hashed or not. 
- Reservation maintenance may be responsibility of `StateLifecycleManager`, and direct reservation of `State` instance should not be allowed. In this case, `StateLifecycleManager` has more control over `State` and can make sure that it's hashed before it's released but in this case we would have to maintain a list of `State` instances in memory. 

I'm leaning towards option 1.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Fair enough. I'll remove the code and replace it with the summary of the idea.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I didn't just fixed the typo, I reworded the whole sentence.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Done.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I added a clarification for the "invalid" checks. Yes, they will be common for CN and BN.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I updated an existing diagram mapping the State method names to the part of the lifecycle. I also added a diagram that depicts how the final result is going to look like once we add `StateLifecycleManager`

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,176 @@
...
+public interface StateLifecycleManager {
+  /**
+   * Get the latest immutable state.
+   */
+  State getLatestImmutableState();
+
+  /**
+   * Get the mutable state.
+   */
+  State getMutableState();
```
jsync-swirlds, line 113: A question on this came up in Block Node design discussion.
If I have a `mutable` state, make a large set of changes, then decide to _discard_ those changes, can I discard the `mutable` state, and get a _new_ `mutable` state from the `last immutable` state?

Functionally (using this API), can I efficiently revert back to the last immutable state and then start making changes again?

The use case here is if I make a bunch of state changes in Block Node, but then something causes the block stream to stop and have to retry the current block (network interruption, failed block proof, etc...), I need a way go back to the starting state and begin applying the same block again.

imalygin: >If I have a mutable state, make a large set of changes, then decide to discard those changes, can I discard the mutable state, and get a new mutable state from the last immutable state?

If we use the pattern established for the `MerkleNode` instances and their `copy` methods, we can't. This is one-way street. That is, once you create a copy of a mutable instance it becomes immutable, and only the copy is mutable. The framework doesn't allow to create a copy of an immutable instance
`com.swirlds.state.merkle.MerkleStateRoot#copy`:

```
   public T copy() {
        throwIfImmutable();
        throwIfDestroyed();
        setImmutable(true);
        return copyingConstructor();
    }
```

>Functionally (using this API), can I efficiently revert back to the last immutable state and then start making changes again?

I think we can support this use case but it require a tiny bit more effort as it deviates from the normal usage of a copy method.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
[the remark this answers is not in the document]
imalygin: It should not be there and it should be static and be a part of `StateLifecycleManagerImpl`

platform-sdk/docs/proposals/state-lifecycle/proposal.md
jsync-swirlds: We'll need that revert-to-last-immutable capability in Block Node to handle when a block fails validation or other interruptions, so we would very much appreciate if that could be included in the design for this API.

imalygin: Done

jsync-swirlds: Thank you.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,241 @@
...
+
+The `SwirldStateManager` class requires refactoring to separate its two distinct responsibilities. `swirlds-state-impl` should not include any platform-specific details but should provide the necessary mechanisms for the platform to interact with the state and its lifecycle.
+
+### `StateLifecycleManager` Interface
+
+To achieve this, `swirlds-state-api` will define a `StateLifecycleManager` interface, with an implementation `StateLifecycleManagerImpl` in `swirlds-state-impl`. The interface will handle:
+
+- Creating/loading an initial state. If no snapshot is available, a new state will be created.  
+- Managing references to the latest immutable state and the latest mutable state.
+- Resetting the mutable state to the latest immutable state.
```
lpetrovic05, line 120: I don't understand this resetting functionality, can you add a bit more detail please? During a reset, what happens to the mutable state and to the immutable state?

imalygin: > During a reset, what happens to the mutable state and to the immutable state?

The previous mutable state get discarded from the `StateLifecycleManager` (it's a client's code responsibility to release and destroy it). Current immutable state remains unchanged.

jsync-swirlds: Exactly.  Block Node will need the ability to reset to the last immutable state and begin applying changes again from that point.
Examples for when this might be needed include
1. A block fails verification, so we have to request it be resent or request it from another source.
1. A stream is interrupted mid-block (e.g. by a network fault) and we need to request resend or from another source.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,268 @@
...
+    */
+    State createSnapshot(final @NonNull Path targetPath);
+
+    /**
+     * Creates a mutable copy of the state. The previous mutable state becomes immutable,
+     * replacing the latest immutable state. 
+     * 
+     * @return a mutable copy of the state
+     */
+    State copyMutableState();
```
poulok, line 170: I think this should to return the immutable state that was used as the source of the copy operation. This is currently what the copy method in the `SwirldsStateManager` class returns because that is what is used for the signed state. It is the thing that moves forward in the lifecycle. If it is not returned by this method, the caller code will look something like this:

```
State soonToBeImmutableState = sslm.getMutableState();
sslm.copyMutableState();
SignedState ss = new SignedState(soonToBeImmutableState); // now is immutable
```

If it returns the immutable state, then it looks like this:

```
State completedState = sslm.copyMutableState();
SignedState ss = new SignedState(completedState);
```

lpetrovic05 (dismissed):

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,268 @@
...
+    public static List<Path> discover(Path sourcePath) {
+        // implementation
+    }
+}
+```
+
+
+#### Notes
+
+- The `StateLifecycleManager` implementation will maintain a **1-to-1 relationship** with the state.
```
poulok, line 220: I don't understand the change to being a 1-1 relationship with the state. How will it ensure a single mutable state at a given time if there is one instance per state? The app would have to keep track of the mutable state on it's own, which is one of the reasons we wanted this lifecycle manager.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,268 @@
...
+    }
+}
+```
+
+
+#### Notes
+
+- The `StateLifecycleManager` implementation will maintain a **1-to-1 relationship** with the state. 
+The assumption is that the application will require only one mutable state at a time, and this state should be initialized during the application's startup.  
+- If client code needs a state loaded from an arbitrary snapshot, it can directly use the `StateLifecycleManagerImpl#loadSnapshot` method, which is `static`.
```
poulok, line 222: The client code should have any access to the implementation class, only the interface, so I don't see how this will work

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: Yes, that's a good point. I'll make this change.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: The client code will need to use the implementation classes anyway. 

However, instead of having static method we can have a constructor that takes a path to the snapshot. And this is how the client code will be able to work with the loaded state.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: I'll clarify the wording but what I meant is that this relationship is between the lifecycle manager and a sequence of states. It is **not** a relationship between one instance of the state and one instance of the lifecycle manager. 

Imagine a hypothetical scenario in which we need to support more than one network. We will to have one lifecycle manager per network, and the lifecycle manager will have two state references at a time (current mutable state and the latest immutable state). However, these references will be updated over time and will point to other instances.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+The following design describes a solution for managing the lifecycle of the state object. It covers the following aspects:
+
+- State initialization (loading a snapshot or creating a new state)
+- Maintaining references to the mutable state and the latest immutable state
+- Restricting mutability to a single state object
+- Loading snapshots from disk
+
+## Current State of the Code
+
+### Reservation Count Mechanism
```
jasperpotts, line 35: I really want this to go away. It just feels over complicated. The basic problem is we have a chain of versions starting from the mutable version going back to the oldest known version on disk. We have to explicitly delete a version when we are finished with it so it can be merged in virtual node cache or written to disk and freed in memory. If we make it explict manual in State API to have to call release/delete(version) then the state machine that controls states will have to keep track of if it needs a state or not and call release() when done. This will normally be at the end of processing and writing of blockStream block. But sometimes one will be held longer for state saving. Or coping to teach from. 

Maybe we would have to do this later when we can delete all non-virtual-map merkle nodes as will be very hard to change before that.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+
+Once the reservation count reaches 0, the state becomes eligible for garbage collection.
+
+---
+
+### Java Classes
+
+`com.swirlds.state.State` provides the following lifecycle-related methods:
+
+- `State copy()`: Creates a mutable copy of the state. The source state becomes immutable.
```
jasperpotts, line 76: Don't love the name `copy()` as it implies it is completely separate copy of tree. What this really is doing is creating a new version so maybe calling it `newVersion()` is more obvious.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+- `void computeHash()`: Computes the hash of the immutable state.
+- `void createSnapshot(Path targetPath)`: Creates a snapshot of the state at the specified path.
+- `State loadSnapshot(Path sourcePath)`: Loads a snapshot from the specified path.
+
```
jasperpotts, line 80: should there be a `createNewState()` method to create a brand new state?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: >Don't love the name copy() as it implies it is completely separate copy of tree. 

I was following the pattern that we have for `MerkleNode`, and, correct me if I'm wrong, but it doesn't imply a completely separate copy of the tree. Also, it follows the pattern we use in the other parts of the code - once we create a copy, the source becomes immutable.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: That would be up to `StateLifecycleManager`. It will be responsible for either loading an existing snapshot or for creating an instance of a new state.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+Once the reservation count reaches 0, the state becomes eligible for garbage collection.
+
+---
+
+### Java Classes
+
+`com.swirlds.state.State` provides the following lifecycle-related methods:
+
+- `State copy()`: Creates a mutable copy of the state. The source state becomes immutable.
+- `void computeHash()`: Computes the hash of the immutable state.
```
artemananiev, line 77: Compute the state hash. Can only be called on immutable states

platform-sdk/docs/proposals/state-lifecycle/proposal.md
artemananiev: I am slightly confused about loadSnapshot(). As far as I can see, a State instance is to be created first, then loadSnapshot() is called for this instance, is it correct? Shouldn't it be a factory method somewhere else, e.g. StateLifecycleManager?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+   @NonNull
+   ReadableStates getReadableStates(@NonNull String serviceName);
+
+   @NonNull
+   WritableStates getWritableStates(@NonNull String serviceName);
+
+   default void registerCommitListener(@NonNull final StateChangeListener listener)l;
+
+   @Nullable
+   Future<Hash> getHashFuture();
```
artemananiev, line 123: computeHash(), maybe?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+public interface StateLifecycleManager {
+    /**
+     * Get the mutable state.
+     */
+    State getMutableState();
+
+   /**
+    * Reset the mutable state to the latest immutable state.
+    */
+   void resetMutableState();
```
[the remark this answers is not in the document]
jsync-swirlds: This was discussed in an [earlier conversation](https://github.com/hiero-ledger/hiero-consensus-node/pull/16592#discussion_r1890840031) on this PR.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+- `void createSnapshot(Path targetPath)`: Creates a snapshot of the state at the specified path.
+- `State loadSnapshot(Path sourcePath)`: Loads a snapshot from the specified path.
+
+These methods correspond to the state lifecycle as follows:
+
+[![State lifecycle](state-lifecycle.svg)](state-lifecycle.svg)
+
+---
+
+### `SwirldStateManager`
```
lpetrovic05, line 87: I think there is a PR to delete this class

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
+# Lifecycle Management for the State
+
+## Summary
+
+This proposal describes a possible implementation of state lifecycle management that could work equally well for both the Consensus Node and the Block Node.
```
Nana-EC, line 5:
```suggestion
This proposal describes a possible implementation of state lifecycle management that could work equally well for the Consensus Node, Block Node and any clients that parse network streams and manage state.
```

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+## Design & Architecture
+
+### Summary
+
+The following design describes a solution for managing the lifecycle of the state object. It covers the following aspects:
+
+- State initialization (loading a snapshot or creating a new state)
+- Maintaining references to the mutable state and the latest immutable state
+- Restricting mutability to a single state object
+- Loading snapshots from disk
```
Nana-EC, line 31: This is a repetition of the first bullet point. Should this rather be persistence a snapshot to disk?
I think serializing a snapshot should be a feature

Nana-EC, line 31: Is there feature support for applying changes or is that the State API?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
Nana-EC: For information reason why can it only be called on immutable state?
Is there consideration for being able to compute the hash in the mutable state?
Say I start with State A, do a `copy()`, apply state changes to the mutable state and want to capture the hash of the state at that point.
Is this feasible and supported?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
```diff
@@ -0,0 +1,270 @@
...
+
+---
+
+### Java Classes
+
+`com.swirlds.state.State` provides the following lifecycle-related methods:
+
+- `State copy()`: Creates a mutable copy of the state. The source state becomes immutable.
+- `void computeHash()`: Computes the hash of the immutable state.
+- `void createSnapshot(Path targetPath)`: Creates a snapshot of the state at the specified path.
```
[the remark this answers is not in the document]
Nana-EC: Same note above. Why only on immutable. What about capture the point in time after a select number of changes and persisting that?

platform-sdk/docs/proposals/state-lifecycle/proposal.md
Nana-EC: My reading of this is loadSnapshot() would load state into the immutable version.
So maybe to start you would either loadSnapshot() or create a blank state to start with.
@imalygin is that what you were going for?

If so then similar to @jasperpotts comment one approach might be to utilize a create method.
Maybe it's `create()` & `create(Path sourcePath)`, with the later loading a snapshot from disk.
Or just a `create(Path sourcePath)` where if `sourcePath` is `null` you create an empty state to start with

Nana-EC: `createImmutable(...)` is also another option as it then allows you to rename `copy()` to `createMutable()` which highlights the need for an immutable state first and then you copy from immutable to Mutable state

Nana-EC (commented):
Nice suggestions.
Made it half way through will circle back for the last few sections

poulok:
@rbarker-dev is this PR ready to be reviewed again?

rbarker-dev:
> @rbarker-dev is this PR ready to be reviewed again?

I'm not sure.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: > Shouldn't it be a factory method somewhere else, e.g. StateLifecycleManager?

@artemananiev yes, your assumption is correct. You're commenting on the current state of things. Later in this document I propose moving this functionality to `StateLifecycleManager`

imalygin: @Nana-EC not exactly. `StateLifecycleManager` should be responsible for creation of snapshot and loading state from them.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
imalygin: > For information reason why can it only be called on immutable state?

@Nana-EC the state must not be mutable because any change to the state will make the hash invalid. _Technically,_ we can calculate hash on a mutable state, but it's goes against the very idea of the state hash.

platform-sdk/docs/proposals/state-lifecycle/proposal.md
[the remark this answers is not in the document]
imalygin: @Nana-EC to guarantee consistency of the state snapshot, it must be immutable. 

> What about capture the point in time after a select number of changes and persisting that?

To some extent, it's already happening under the hood of the virtual map - it's versioned, and each version has a set of mutations associated with it. Once a round is closed, we create a next version of the state, and with previous versions we can do various operations like merge, hash and flush to disk, as we know that no other changes to these versions are expected.

PR #16459, opened 2024-11-06
chore: Make test apps use roster

litt3:
- closes #16458

PR #16594, opened 2024-11-13
chore: Add `StartupNetworks` to `MigrationContext`, use in restart-only `RosterService` schema

### LINKED ISSUE #16551: Provide `StartupNetworks` implementation for use in ServicesMain
**Background**
Create an implementation of the `StartupAssets` interface which provides inputs to the startup phase of the Hedera network.

See [Roster Lifecycle Implementation](https://www.notion.so/swirldslabs/Roster-lifecycle-implementation-13b7c9ab25918069a0afcc44713015b9) for details.

**Acceptance Criteria**
- `StartupAssets` interface implementation created
- Unit tests for implementation

**Dependencies**
#16550

tinker-michaelj:
**Description**:
 - Closes #16551 
 - Refactors `Hedera` to use a single call to `OrderedServiceMigrator.doMigrations()` (necessary for correctness when using restart-only `RosterService` schemas).


hapi/hedera-protobufs/services/state/network.proto
jsync-swirlds, line 1: We now have this file in 3 different PRs (all of which need the same or similar changes).  Which is the canonical PR?

mxtartaglia-sl (approved):

hapi/hedera-protobufs/services/state/network.proto
thomas-swirlds-labs: @jsync-swirlds. This one should be the canonical [PR](https://github.com/hashgraph/hedera-services/pull/16581)  
Once that one gets approved we can incorporate here, as it is a prerequisite. Thanks! 👍

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V057RosterSchema.java
```diff
@@ -0,0 +1,120 @@
...
+        requireNonNull(ctx);
+        if (!ctx.configuration().getConfigData(AddressBookConfig.class).useRosterLifecycle()) {
+            return;
+        }
+        final var rosterStore = rosterStoreFactory.apply(ctx.newStates());
+        final var startupNetworks = ctx.startupNetworks();
+        if (ctx.isGenesis()) {
+            setActiveRoster(GENESIS_ROUND_NO, rosterStore, startupNetworks.genesisNetworkOrThrow());
+        } else {
+            final long roundNumber = ctx.roundNumber();
```
MiroslavGatsanoga, line 81: I think according to the proposal `roundNumber` for override and both upgrade cases should be `state round +1` e.g. [see here](https://github.com/hashgraph/hedera-services/blob/16066-tss-roster-update/platform-sdk/docs/proposals/001-tss-rosters-design-proposal/proposal.md?plain=1#L244).

hedera-node/configuration/dev/genesis-network.json
```diff
@@ -0,0 +1,54 @@
...
+      "gossipEndpoint": [{
+        "ipAddressV4": "fwAAAQ==",
+        "port": 35378
```
Neeharika-Sompalli, line 47: Are these correct ports? Usually we use 50204 range right ?

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/AddressBookServiceImpl.java
```diff
@@ -31,6 +31,6 @@ public final class AddressBookServiceImpl implements AddressBookService {
     @Override
     public void registerSchemas(@NonNull SchemaRegistry registry) {
         registry.register(new V053AddressBookSchema());
-        registry.register(new V056AddressBookSchema());
+        registry.register(new V057AddressBookSchema());
```
Neeharika-Sompalli, line 34: Should we rename or add new schema?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -502,27 +537,10 @@ public void initializeStatesApi(
...
-                state,
-                servicesRegistry.subRegistryFor(EntityIdService.NAME, PlatformStateService.NAME, RosterStateId.NAME),
-                deserializedVersion == null ? null : new ServicesSoftwareVersion(deserializedVersion),
-                version,
-                bootstrapConfigProvider.getConfiguration(),
-                platformConfiguration,
-                UNAVAILABLE_NETWORK_INFO,
-                UNAVAILABLE_METRICS);
-        migrationStateChanges = new ArrayList<>();
-        migrationStateChanges.addAll(platformStateMigrations);
```
Neeharika-Sompalli, line 517: Seems like this is dead code, thats why it was removed?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -608,24 +629,26 @@ private List<StateChanges.Builder> onMigrate(
...
+                // (FUTURE) In principle, the FileService could change the active configuration during a
+                // migration, implying we should pass a config provider; but we don't need this yet
                 configProvider.getConfiguration(),
                 platformConfiguration,
                 genesisNetworkInfo,
-                metrics);
-        migrationStateChanges.addAll(migrationChanges);
+                metrics,
+                startupNetworks);
+        PLATFORM_STATE_SERVICE.clearActiveRosterFn();
```
Neeharika-Sompalli, line 650: Why do we need to clear it here ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/ServicesMain.java
```diff
@@ -94,63 +95,51 @@ public class ServicesMain implements SwirldMain {
...
-                new OrderedServiceMigrator(),
-                InstantSource.system(),
-                appContext -> new TssBaseServiceImpl(
-                        appContext,
-                        ForkJoinPool.commonPool(),
-                        ForkJoinPool.commonPool(),
-                        new PlaceholderTssLibrary(),
-                        ForkJoinPool.commonPool(),
-                        new NoOpMetrics()));
+        // No-op, everything must be initialized in the main() entrypoint
```
Neeharika-Sompalli, line 108: This makes much more sense! 🙏

hedera-node/hedera-app/src/main/java/com/hedera/node/app/info/DiskStartupNetworks.java
```diff
@@ -0,0 +1,322 @@
...
+        final var config = configProvider.getConfiguration();
+        final var unscopedNetwork = loadNetwork(config, OVERRIDE_NETWORK_JSON);
+        if (unscopedNetwork.isPresent()) {
+            return unscopedNetwork;
+        }
+        return loadNetwork(config, "" + roundNumber, OVERRIDE_NETWORK_JSON);
+    }
+
+    @Override
+    public void setOverrideRound(final long roundNumber) {
```
Neeharika-Sompalli, line 106: When would we need this? Wouldn't all the cases have a round number they start from?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseService.java
```diff
@@ -29,16 +32,23 @@
...
+ * as a side effect, the ledger id; as this is exactly the same as the TSS
+ * public key.
  */
 public interface TssBaseService extends Service {
+    /**
+     * Since the roster service has to decide to adopt the candidate roster
+     * based on the available key material, the TSS service must be migrated
+     * before the roster service.
+     */
+    int MIGRATION_ORDER = RosterService.MIGRATION_ORDER - 1;
```
Neeharika-Sompalli, line 50: nit: We could probably fix one migration order for all services?  Need not be in this PR, just a thought.
So, it will be clear which service happens in which order

hedera-node/hedera-app/src/main/java/com/hedera/node/app/workflows/handle/HandleWorkflow.java
```diff
@@ -412,25 +403,6 @@ private HandleOutput execute(@NonNull final UserTxn userTxn) {
...
-                        .getConfigData(TssConfig.class)
-                        .keyCandidateRoster();
-                final var useRosterLifecycle = configProvider
-                        .getConfiguration()
-                        .getConfigData(AddressBookConfig.class)
-                        .useRosterLifecycle();
-
-                if (!keyCandidateRoster
-                        && useRosterLifecycle
-                        && userTxn.functionality() == HederaFunctionality.FREEZE) {
```
Neeharika-Sompalli, line 427: Isn't this still needed ? Since we enable Roster we don't need this?

Neeharika-Sompalli: Ah I see you moved to a different place.

Neeharika-Sompalli (commented):
Wow, great work @tinker-michaelj ! 🙌 Overall looks great to me. Have a few questions though.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V057RosterSchema.java
tinker-michaelj: True @MiroslavGatsanoga , great catch---fixed.

hedera-node/configuration/dev/genesis-network.json
tinker-michaelj: For now this is just an asset for an embedded test; once the `PlatformStateService` restart schema is done and we are running against subprocess networks, I will make these ports all `0` in this JSON and then change the code to substitute the actual subprocess `@HapiTest` ports as part of the test network setup.

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/AddressBookServiceImpl.java
tinker-michaelj: The previous `V056AddressBookSchema` is actually deleted; so this PR is adding a new schema.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
tinker-michaelj: I wanted to do this just to emphasize the scope of this `Roster` supplier is very limited---only to the code running inside `Schema`s.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/info/DiskStartupNetworks.java
tinker-michaelj: I _think_ we need this to mark the earliest round number that we used an override network for....but it's hard to say without a live test. Will examine further once `PlatformStateService` restart schema is done. 🤔

hedera-node/hedera-app/src/main/java/com/hedera/node/app/tss/TssBaseService.java
tinker-michaelj: Interesting idea, a global ordering might well be clearer than these local "relative orderings". 💯

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
tinker-michaelj: This was shadowing an instance field with the same name; so just used the field.

Neeharika-Sompalli (dismissed):
LGTM! Thanks @tinker-michaelj

hapi/hedera-protobufs/services/state/network.proto
```diff
@@ -0,0 +1,86 @@
+/**
```
thomas-swirlds-labs, line 1: FYI - This file got moved to another directory outside of `hedera-protobufs` after a lengthy discussion on this [PR](https://github.com/hashgraph/hedera-services/pull/16581). This should hopefully also negate the need to change `module-info` again :)

mxtartaglia-sl (approved):

rbarkerSL (dismissed):

thomas-swirlds-labs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -49,6 +48,7 @@
 public final class RosterRetriever {
     private RosterRetriever() {}
 
+    private static final String ROSTER_SERVICE = "RosterService";
```
anthony-swirldslabs, line 51: The `RosterStateId.NAME` and other constants in that class were introduced for a reason so that both the platform and app code could use them. It's not a good idea to replicate string literals when they need to be used in multiple places.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/PlatformStateService.java
```diff
@@ -54,21 +63,53 @@ public void registerSchemas(@NonNull final SchemaRegistry registry) {
...
+     */
+    public long roundOf(@NonNull final MerkleStateRoot root) {
+        requireNonNull(root);
+        final var platformState = platformStateOf(root);
+        return platformState == null
+                ? 0L
+                : platformState
+                        .consensusSnapshotOrElse(ConsensusSnapshot.DEFAULT)
+                        .round();
+    }
```
anthony-swirldslabs, line 105: This logic is already implemented in `ReadablePlatformStateStore`. And the functionality of getting the round given a `State` (or a `MerkleStateRoot`) has been implemented in the `RosterRetriever` already. I suggest to avoid code duplication. We should either remove that replica and use this one, or not introduce this one and keep using the `RosterRetriever` instead.

Minor: all the `...Of()` methods in this class don't seem to access instance variables and hence can all be made `static` to make them usable in other places.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/schemas/V057PlatformStateSchema.java
```diff
@@ -0,0 +1,52 @@
...
+            SemanticVersion.newBuilder().major(0).minor(57).build();
+
+    /**
+     * A supplier for the active roster.
+     */
+    private final Supplier<Roster> activeRoster;
+
+    public V057PlatformStateSchema(@NonNull final Supplier<Roster> activeRoster) {
+        super(VERSION);
+        this.activeRoster = requireNonNull(activeRoster);
```
anthony-swirldslabs, line 45: We're currently working on removing the AddressBook objects (both current and previous) from the PlatformState, and currently we don't have plans to reintroduce similar concepts for rosters in the PlatformState. The RosterService should be the only source of truth for the information about the active/previous rosters. So this supplier doesn't seem to belong here.

Any code that needs to know the roster(s), should either use the RosterService directly, or use the `RosterRetriever` depending on the context (ideally, we'd deprecate the RosterRetriever eventually and only ever use the RosterService, but currently there's special cases where we need to use the retriever for now - e.g. in https://github.com/hashgraph/hedera-services/pull/16459 ).

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritableRosterStore.java
```diff
@@ -39,6 +38,8 @@
  * Read-write implementation for accessing rosters states.
  */
 public class WritableRosterStore extends ReadableRosterStoreImpl {
+    public static final String ROSTER_KEY = "ROSTERS";
+    public static final String ROSTER_STATES_KEY = "ROSTER_STATE";
```
anthony-swirldslabs, line 42: As mentioned above, we put these constants into the `RosterStateId` to be able to sue them in other places. Please note that the Readable store also uses them. It's not a good idea to duplicate these string literals. I suggest to keep using the `RosterStateId`.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/ReadableRosterStoreImpl.java
```diff
@@ -114,22 +114,16 @@ public Bytes getActiveRosterHash() {
...
-                                    .build())
-                            != null);
-            if (allHashesPresent) {
-                return rosterStateSingleton.roundRosterPairs();
-            }
-        }
-        return null;
+    public Bytes getPreviousRosterHash() {
+        final var rosterHistory = getRosterHistory();
+        return rosterHistory.size() > 1 ? rosterHistory.getLast().activeRosterHash() : null;
```
anthony-swirldslabs, line 121: This code makes a hard assumption that the size of the history is never greater than two. This logic will fail if the history is longer. And we do expect the size of the history to increase in the future when the full DAB is implemented.

Since we define the active roster as the entry at index 0, the previous roster should be defined as the entry at index 1 if it exists, or null otherwise. The logic in this method should never access entries with indices higher than 1.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/ReadableRosterStoreImpl.java
```diff
@@ -114,22 +114,16 @@ public Bytes getActiveRosterHash() {
...
     @Override
-    public List<RoundRosterPair> getRosterHistory() {
-        final RosterState rosterStateSingleton = rosterState.get();
-        if (rosterStateSingleton != null
-                && !rosterStateSingleton.roundRosterPairs().isEmpty()) {
-            boolean allHashesPresent = rosterStateSingleton.roundRosterPairs().stream()
-                    .allMatch(pair -> rosterMap.get(ProtoBytes.newBuilder()
-                                    .value(pair.activeRosterHash())
-                                    .build())
-                            != null);
```
anthony-swirldslabs, line 128: While it's unclear if this validation logic belongs to the readable store, it's still important to have this implemented somewhere. I suggest to keep it for now. If you want to remove it from this method, then the logic should be moved elsewhere as a part of this same fix to ensure we maintain these invariants.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
tinker-michaelj: In general I certainly agree; but these `Service`/`Schema` keys can never change, so there is no risk of them getting out of sync via multiple references, and it's jarring to have a different place for these keys for `RosterService` specifically; so I prefer this.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/PlatformStateService.java
tinker-michaelj: `ReadablePlatformStateStore` requires the States API to have been initialized, and this accessor's purpose is to **setup** the code that initializes the States API.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/schemas/V057PlatformStateSchema.java
tinker-michaelj: True, once this platform migration is done, we may want a different migration here.

But since we hope to merge this PR and then almost immediately enable the `Roster` lifecycle in PR checks, we need a stopgap here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/WritableRosterStore.java
tinker-michaelj: As above.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/ReadableRosterStoreImpl.java
tinker-michaelj: Ah got it, fixed.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/ReadableRosterStoreImpl.java
tinker-michaelj: Makes sense, added filtering to only return `RoundRosterPair`s that reference extant rosters.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/schemas/V057PlatformStateSchema.java
anthony-swirldslabs: Sounds good.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/service/PlatformStateService.java
anthony-swirldslabs: I see. Makes sense.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
anthony-swirldslabs: This approach still establishes a pattern of it being "okay" to replicate these names, and in the future, a typo in one of these or new replicas can easily break things in a way that may be very difficult to debug. So I don't agree with this justification. But if you insist...

anthony-swirldslabs (dismissed):

MiroslavGatsanoga (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

rbarkerSL (approved):

PR #16610, opened 2024-11-14
chore: Modified block stream handling for record file data

jsync-swirlds:
* Adjusted record file item to latest expectation
 * Removed address book version from block header
 * Added a missing service name in `block/stream/output/smart_contract_service.proto` documentation

### Details
During work to migrate record file data into block stream format, Jasper found some issues with the current proto files, and requested some modest changes.
The `address_book_version` field in `BlockHeader` is not needed with TSS.
The `number` and `hash_algorithm` in `RecordFileItem` are duplicative of `BlockHeader`.
A decision was made to add a `BlockHeader` to record file blocks.
A (to be designed) `AddressBookProof` item will also be added to record file blocks in a later PR.

### Implementation Note
The `address_book_version` field was not used anywhere, so removal does not require code changes.
The `RecordFileItem` is not used in consensus code, so changes do not impact code in this repository.

### DIFF
--- a/hapi/hedera-protobufs/block/stream/output/block_header.proto
+++ b/hapi/hedera-protobufs/block/stream/output/block_header.proto
@@ -152,14 +152,4 @@ message BlockHeader {
      * This SHOULD always be `SHA2_384`, currently.
      */
     proto.BlockHashAlgorithm hash_algorithm = 6;
-
-    /**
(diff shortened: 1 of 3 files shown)


mhess-swl (approved):
LGTM

Issue #16494, opened 2024-11-07
Remove .given(), .when(), .then()


mhess-swl:
In an effort to comply with the Hapi test guidelines, and to clear up confusion about its usage, this ticket is for cleaning up any `@HapiTest`s written in the `.given()`, `.when()`, `.then()` format, in favor of the more declarative `hapiTest(...ops...)` approach.

mhess-swl:
There is/has been no interest in prioritizing this recently, and it's not really necessary.

PR #16613, opened 2024-11-15
chore: Refactor hapi tests to use `hapiTest(...)` instead of `defaultHapiSpec(...)` (Part 1)

mhess-swl:
This PR does not change any test conditions; instead, it merely wraps each set of spec operations with the new `hapiTest(...)` method wrapper instead of `defaultHapiSpec(...)`. 

Part of #16494


thomas-swirlds-labs (approved):
LGTM. Thanks @mhess-swl 👍

PR #16619, opened 2024-11-15
docs: add Leemon's glossary

poulok:
**Description**
Leemon put together a Consensus Glossary that we should check into the repo as the source of truth for terminology and definitions. Let's review and discuss if anything should be changed, clarified, etc. before committing.

### DIFF
--- a/platform-sdk/docs/hashgraph_glossary.md
+++ b/platform-sdk/docs/hashgraph_glossary.md
@@ -1,0 +1,250 @@
+# Hashgraph Glossary
+
+<a name="top"/>
+
+- [Rounds](#rounds)
+- [Fields of an event](#fields-of-an-event)
+- [Events](#events)
+- [Event relationships](#event-relationships)
+- [Generations](#generations)
+- [Address Books](#addressbooks)
+- [Parameters](#parameters)
+
+<a name="rounds"/>
+
+## Rounds
+
+- __Pending__ round - the round number of the earliest round that hasn't yet reached consensus. This is the round whose
+  judges are currently being determined. Those consensus calculations use the roster whose roster round is the current
+  `pendingRound`. Any new event is created containing a `birthRound` field equal to the current `pendingRound`. All
+  events currently in the hashgraph should have birth rounds in the range from `pendingRound - roundsAncient` to
+  `pendingRound`, inclusive.
+- __Consensus__ round - the round number for the latest round that has reached consensus. <br>
+    - `consensusRound = pendingRound - 1`
+- __First non-ancient__ round - the oldest round that is not ancient<br>
+    - `firstNonAncientRound = pendingRound - roundsAncient`
+- __First non-expired__ round - the oldest round that is not expired<br>
+    - `firstNonExpiredRound = pendingRound - roundsExpired`
+- __Last roster__ round - Last future round for which there should be a roster in the queue<br>
+    - `lastRosterRound = pendingRound + roundsRoster`
+- __Roster__ round - Any given roster (subset of an address book) is associated with a long, which is its roster round
+  number. The `nodeID`s and consensus weights in a roster should be used when that roster round number equals the
+  pending round. A queue of rosters must be stored in state, one for each round number from `firstNonAncientRound` to
+  `lastRosterRound`, inclusive. Every time a round reaches consensus, the oldest roster can be removed. Any time the
+  transactions in a round are handled, a new roster is added for the future. An event is only allowed to be in the
+  hashgraph when its birth round is the roster round for some roster in the queue. And it must have been created by a
+  node that is listed in that roster. Of course, ancient, non-expired events may still be in memory, to gossip to nodes
+  that are behind this node, but they don't count as being in this node's hashgraph. Events with birth rounds too far in
+  the future should not be accepted during gossip. And a node should not send such events to another node, if it knows
+  that node will just discard them for that reason.
+
+  [top](#top)
+
+<a name="fields-of-an-event"/>
+
+## Fields of an event
+
+- __birthRound__ - set to the creator's pending round at the moment of creation. Is signed and immutable.
+- __votingRound__ - the round for this event, which determines whether it is a witness. It is a witness if its voting
+  round is different from its self-parent's voting round (or it has no self-parent). If it is a witness, then it might
+  be eligible for election as a judge (if it's in the pending round), or it might be an initial voter for judge (if it's
+  in the pending round plus `d`), or it might be a voter and vote collector (if it's later than the pending round plus
+  `d`). Every time the processing of a new round starts, the voting round is recalculated for all events. It can be
+  different than it was in the past because non-ancestors of the latest judges are defined to have a voting round of
+  negative infinity (represented as a 0 in memory), and that can cause all their descendents to change their voting
+  rounds. If a judge in the consensus round is not a descendant of any other judge in the consensus round, then it is
+  guaranteed to have a voting round equal to the consensus round. In the fields listed here, this is the only one that
+  is mutable. All the rest are immutable, and are signed when the event is created.
+- __consensusRound__ - the first round in which this event was an ancestor of all judges.
+- __consensusTimestamp__ - the median timestamp of when it first reached each of the nodes that created judges in its
+  consensus round. This is adjusted by adding nanoseconds to ensure that in consensus order, each transaction is at
+  least 1000 nanoseconds after the previous one. (That ensures they are unique and monotonically increasing, and there
+  are big enough gaps so that synthetic transactions inserted between them can have timestamps that are unique).
+- __consensusOrder__ - an long `N` indicating this is the `N`th event in all of history, according to the calculated
+  consensus order.
+- __selfParent__ - the hash of the self parent (or null if there is none, or the self-parent's birth round is ancient)
+- __selfParentCreator__ - the node that created the self-parent event. If this is incorrect, then it cannot be added to
+  the hashgraph until the self-parent becomes ancient.
+- __selfParentBirthRound__ - the claimed birth round of the self-parent. If this is incorrect, then it cannot be added
+  to the hashgraph until the self-parent becomes ancient.
+- __otherParent__ - the hash of the other-parent (or null if there is none, or the other-parent's birth round is
+  ancient).
+- __otherParentCreator__ - the node that created the other-parent event
+- __otherParentBirthRound__ - the claimed birth round of the other-parent. If this is incorrect, then it cannot be added
+  to the hashgraph until the other-parent becomes ancient.
+
+  [top](#top)
+
+<a name="events"/>
+
+## Events
+
+- __Judge__ event - an event that wins the election to be made a judge. It must be a witness, and it will have tended to
+  have been gossiped to most of the other nodes quickly (otherwise it would have lost the election). An event reaches
+  consensus when it is an ancestor of all judges in a given round. The first round where that happens is its consensus
+  round. It's a math theorem that every round is guaranteed to have at least one judge, and a math conjecture that every
+  round is guaranteed to have judges created by a supermajority of nodes (>2/3 of weight).
+- __Consensus__ event - an event that has reached consensus
+- __Ancient__ event - an event whose birth round is less than `(pendingRound - roundsAncient)`. This event should not be
+  in the hashgraph, though it may still be in memory in order to give to other nodes that are behind this one. During
+  gossip, a node will discard any event that is ancient for it. So no event should be sent to a node if that event will
+  be ancient for that node.
+- __Expired__ event - an event whose birth round is less than `(pendingRound - roundsExpired)`. Expired events should be
+  removed from memory. So if a node is so far behind that the events it needs are expired for its neighbors, then it
+  will not be able to catch up by gossip alone, so it will have to do a reconnect to catch up.
+- __Witness__ event - an event whose voting round is greater than its self-parent's voting round (or which doesn't have
+  a self parent). Only witnesses can be judges or voters.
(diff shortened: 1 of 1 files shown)


platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+- [Parameters](#parameters)
+
+<a name="rounds"/>
+
+## Rounds
+
+- __Pending__ round - the round number of the earliest round that hasn't yet reached consensus. This is the round whose
+  judges are currently being determined. Those consensus calculations use the roster whose roster round is the current
+  `pendingRound`. Any new event is created containing a `birthRound` field equal to the current `pendingRound`. All
+  events currently in the hashgraph should have birth rounds in the range from `pendingRound - roundsAncient` to
```
poulok, line 20: The code currently calls this `roundsNonAncient`, which to me is more clear than this name. However, the code also uses `roundsExpired` which should be `roundsNonExpired` because it is used in the same manner but for a different concept as `roundsNonAncient`. I suggest we make both include "non" because it means the number of rounds that are NOT ancient and NOT expired.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  `pendingRound`, inclusive.
+- __Consensus__ round - the round number for the latest round that has reached consensus. <br>
+    - `consensusRound = pendingRound - 1`
+- __First non-ancient__ round - the oldest round that is not ancient<br>
+    - `firstNonAncientRound = pendingRound - roundsAncient`
+- __First non-expired__ round - the oldest round that is not expired<br>
+    - `firstNonExpiredRound = pendingRound - roundsExpired`
+- __Last roster__ round - Last future round for which there should be a roster in the queue<br>
+    - `lastRosterRound = pendingRound + roundsRoster`
```
poulok, line 29: The terms `first` and `last` are a bit unnatural to me when describing the hashgraph which we think of as growing upward. I can get used to them, but the terms `earliest` and `latest` are more familiar.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+- __Roster__ round - Any given roster (subset of an address book) is associated with a long, which is its roster round
+  number. The `nodeID`s and consensus weights in a roster should be used when that roster round number equals the
+  pending round. A queue of rosters must be stored in state, one for each round number from `firstNonAncientRound` to
+  `lastRosterRound`, inclusive. Every time a round reaches consensus, the oldest roster can be removed. Any time the
+  transactions in a round are handled, a new roster is added for the future. An event is only allowed to be in the
+  hashgraph when its birth round is the roster round for some roster in the queue. And it must have been created by a
+  node that is listed in that roster. Of course, ancient, non-expired events may still be in memory, to gossip to nodes
+  that are behind this node, but they don't count as being in this node's hashgraph. Events with birth rounds too far in
+  the future should not be accepted during gossip. And a node should not send such events to another node, if it knows
+  that node will just discard them for that reason.
```
poulok, line 39: I don't think we should be describing in detail how the rosters will be stored in state. This is a good conceptual explanation, but the actual implementation may be different. Can we just say that each round has a single (but not necessarily unique) roster that is used for verifying the signatures of events with a matching birth round and whose weights are used in calculating judges for that round when it is the pending round?

poulok: There is some very good explanation of how the birth round value will influence various parts of the system in [this doc](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/consensus-layer/Consensus-Layer.md)

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+- [Fields of an event](#fields-of-an-event)
+- [Events](#events)
+- [Event relationships](#event-relationships)
+- [Generations](#generations)
+- [Address Books](#addressbooks)
+- [Parameters](#parameters)
+
+<a name="rounds"/>
+
+## Rounds
```
lpetrovic05, line 15: should there be an intro explaining what rounds are?

platform-sdk/docs/hashgraph_glossary.md
edward-swirldslabs: I agree

platform-sdk/docs/hashgraph_glossary.md
edward-swirldslabs: I agree.  Defining `round` by itself will also make the specialized terms `pending round` `consensus round` and others feel less circular.

platform-sdk/docs/hashgraph_glossary.md
lpetrovic05: the `roundsExpired` is really a sync gossip concept, so maybe it should not be here. if we move to a new gossip algorithm, this might not exist anymore

platform-sdk/docs/hashgraph_glossary.md
lpetrovic05: I agree

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  round is different from its self-parent's voting round (or it has no self-parent). If it is a witness, then it might
+  be eligible for election as a judge (if it's in the pending round), or it might be an initial voter for judge (if it's
+  in the pending round plus `d`), or it might be a voter and vote collector (if it's later than the pending round plus
+  `d`). Every time the processing of a new round starts, the voting round is recalculated for all events. It can be
+  different than it was in the past because non-ancestors of the latest judges are defined to have a voting round of
+  negative infinity (represented as a 0 in memory), and that can cause all their descendents to change their voting
+  rounds. If a judge in the consensus round is not a descendant of any other judge in the consensus round, then it is
+  guaranteed to have a voting round equal to the consensus round. In the fields listed here, this is the only one that
+  is mutable. All the rest are immutable, and are signed when the event is created.
+- __consensusRound__ - the first round in which this event was an ancestor of all judges.
```
lpetrovic05, line 58: we already have a definition for `consensusRound` above, this is confusing. we should probably change it to `roundReachedConsensus`

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  least 1000 nanoseconds after the previous one. (That ensures they are unique and monotonically increasing, and there
+  are big enough gaps so that synthetic transactions inserted between them can have timestamps that are unique).
+- __consensusOrder__ - an long `N` indicating this is the `N`th event in all of history, according to the calculated
+  consensus order.
+- __selfParent__ - the hash of the self parent (or null if there is none, or the self-parent's birth round is ancient)
+- __selfParentCreator__ - the node that created the self-parent event. If this is incorrect, then it cannot be added to
+  the hashgraph until the self-parent becomes ancient.
+- __selfParentBirthRound__ - the claimed birth round of the self-parent. If this is incorrect, then it cannot be added
+  to the hashgraph until the self-parent becomes ancient.
+- __otherParent__ - the hash of the other-parent (or null if there is none, or the other-parent's birth round is
```
edward-swirldslabs, line 70: Does this terminology work for events with multiple other parents?

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  [top](#top)
+
+<a name="addressbooks"/>
+
+## Address Books
+
+- __Address book__ - the current info for all nodes, stored as a file in the Hedera file system, and therefore also in
+  state. This includes public keys, IP addresses, node ID, proxies, consensus weight, and other info.
+- __Roster__ - a subset of the address book, with just the information needed for consensus. This includes the nodeID,
+  public key for signing events, consensus weight, number of consensus shares, and some cryptographic info related to
```
edward-swirldslabs, line 213: `number of consensus shares` - TSS shares are no longer in the roster.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  TSS and state proofs. Each roster is associated with an unsigned long `roster round`. The roster with a given
```
edward-swirldslabs, line 214: `cryptographic info related to TSS and state proofs` is no longer in the roster.

platform-sdk/docs/hashgraph_glossary.md
anthony-swirldslabs: Also, this paragraph talks about `pending`, `roster`, `birth`, and `ancient` rounds all at once.
1. I suggest to have a separate paragraph for each of these.
2. The paragraphs that define these rounds should go in a logical order, so that they don't have forward-references to terms defined below (as much as it's possible, of course.)

platform-sdk/docs/hashgraph_glossary.md
anthony-swirldslabs: +1.

Besides, it would be wasteful to keep a roster per round as rosters don't usually change that often.

I suggest to reword this definition to state that a roster that becomes active at a particular round continues to be active for all future rounds until a new roster becomes active for a future round.

This would closely correspond to the actual implementation we have to day, and probably will have in the foreseeable future.

platform-sdk/docs/hashgraph_glossary.md
anthony-swirldslabs: Also, `d` looks a tad too short for a term. Let's name this constant so that its name is recognizable, and perhaps define it in the document if there's something to define about it.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+- __consensusRound__ - the first round in which this event was an ancestor of all judges.
+- __consensusTimestamp__ - the median timestamp of when it first reached each of the nodes that created judges in its
+  consensus round. This is adjusted by adding nanoseconds to ensure that in consensus order, each transaction is at
+  least 1000 nanoseconds after the previous one. (That ensures they are unique and monotonically increasing, and there
+  are big enough gaps so that synthetic transactions inserted between them can have timestamps that are unique).
+- __consensusOrder__ - an long `N` indicating this is the `N`th event in all of history, according to the calculated
+  consensus order.
+- __selfParent__ - the hash of the self parent (or null if there is none, or the self-parent's birth round is ancient)
+- __selfParentCreator__ - the node that created the self-parent event. If this is incorrect, then it cannot be added to
+  the hashgraph until the self-parent becomes ancient.
```
anthony-swirldslabs, line 67: It's a bit unclear what `If this is incorrect` means exactly. I suggest to reword this somehow differently.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  vote collector.
+- __Initial voter__ event - a witness with a voting round equal to `1 + pendingRound`. For each node, it votes for the
+  witness created by that node in round pendingRound, or votes NULL if it cannot see any witness by that node in that
+  round. (That's ordinary seeing, not strongly seeing).
+- __Vote collector__ event - a witness with a voting round greater than `1 + pendingRound`. It collects votes from all
+  witnesses that it can strongly see in the previous voting round. For each node, it sets its vote for witness in the
+  pending round to be the majority (or plurality) of the votes it collected. In case of tie, it picks the witness with
+  the least signture lexicographically. For a given node, the event it creates in one round might have a different vote
+  than the one it creates in the previous round. So this is like the a node virtually voting in many rounds, repeatedly
+  changing its vote to match the majority of its peers. For any particular election (about whether a particular event is
```
anthony-swirldslabs, line 121:
```suggestion
  than the one it creates in the previous round. So this is like a node virtually voting in many rounds, repeatedly
  changing its vote to match the majority of its peers. For any particular election (about whether a particular event is
```

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+- __Local__ generation - an event is assigned this immutable number when it is added to the hashgraph. It is 1 plus the
+  max of all the local generations of its non-ancient parents. If it has no non-ancient parents, then it has a local
+  generation of 1. For a given event, this number can be different on different nodes. It can even be different on a
+  single node, before and after a restart or reconnect. The purpose of this number is to make it easy to sort a set of
+  events into topological order. (A topological order is any ordering of the events such that an event's parents always
+  precede it in the order).
+- __Consensus__ generation - an event is assigned this immutable number when it reaches consensus. It is 1 plus the max
+  of all the consensus generations of its eligible parents, where where an event is eligible if it is not ancient and
+  has not reached consensus in a previous round. If an event has no eligible parents, then its consensus generation is
+  1. This is deterministic, so an event is guaranteed to have the same consensus generation on all nodes. This is used
```
anthony-swirldslabs, line 197: `its consensus generation is` - is what?

And the numbered item that follows looks a bit out of context.

platform-sdk/docs/hashgraph_glossary.md
```diff
@@ -0,0 +1,250 @@
...
+  add a line of code to this search to do the calculation of consensus generation at the same time, with no additional
+  searching at all.
+
+  [top](#top)
+
+<a name="addressbooks"/>
+
+## Address Books
+
+- __Address book__ - the current info for all nodes, stored as a file in the Hedera file system, and therefore also in
```
anthony-swirldslabs, line 210: I suggest to limit this doc to rosters only, and not even mention the AB.

poulok:
This document is now out of date. A newer version will be opened in the future.

PR #16622, opened 2024-11-15
docs: Removed incorrect specification for node id in receipt

### LINKED ISSUE #16595: TransactionReceipt nodeId property doesn't appear to be set properly for DAB transactions
### Problem

I ran a NodeDelete transaction. I submitted the transaction to node 6 (with account ID 0.0.9). I deleted node 2. The receipt had the nodeId set to 0. After testing a few different times, submitted to different nodes and deleting different nodes, it appears that the nodeId never gets set for the NodeDelete.

### Solution

- NodeDelete and NodeUpdate both need to set the nodeId of the affected node.

### Alternatives

_No response_

jsync-swirlds:
Very small update to fix some documentation-only errors in the protocol buffer specification for Node update/delete and transaction receipt.

Fixes #16595

### DIFF
--- a/hapi/hedera-protobufs/services/node_delete.proto
+++ b/hapi/hedera-protobufs/services/node_delete.proto
@@ -36,9 +36,8 @@ option java_multiple_files = true;
  * - A deleted address book node SHALL be removed entirely from network state.
  * - A deleted address book node identifier SHALL NOT be reused.
  *
- * ### Record Stream Effects
- * Upon completion the "deleted" `node_id` SHALL be in the transaction
- * receipt.
+ * ### Block Stream Effects
+ * None.
  */
 message NodeDeleteTransactionBody {
     /**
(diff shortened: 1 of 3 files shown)


hapi/hedera-protobufs/services/node_delete.proto
```diff
@@ -36,9 +36,8 @@ option java_multiple_files = true;
  * - A deleted address book node SHALL be removed entirely from network state.
  * - A deleted address book node identifier SHALL NOT be reused.
  *
- * ### Record Stream Effects
- * Upon completion the "deleted" `node_id` SHALL be in the transaction
- * receipt.
+ * ### Block Stream Effects
+ * None.
```
Neeharika-Sompalli, line 40: There will still be StateChanges recorded for this.

Neeharika-Sompalli (approved):
LGTM! Thanks @jsync-swirlds. Had some comments that can be fixed later.

Issue #16621, opened 2024-11-15
Change `RepeatableIntegrationTest` to validate the block signature is correct


Neeharika-Sompalli:
Change `RepeatableIntegrationTest` to validate the block signature is correct

Issue #16564, opened 2024-11-12
Add Metrics for Block Signing


Neeharika-Sompalli:
Add Metrics for

- [x] The time between the request of a ledger signature and the generation of it
- [x] The number of consecutive failures to generate a ledger signature.

PR #16632, opened 2024-11-17
chore: Enable block signing test and add Metrics

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/16621
Fixes https://github.com/hashgraph/hedera-services/issues/16564

- Adds Block Signing test to `RepeatableIntegrationTests`. That validates the signature is not the mock signature.
- Added metrics to update the time taken for ledger signature in `TssMetrics`


petreze (approved):
LGTM

derektriley (approved):
LGTM.

PR #16635, opened 2024-11-18
chore: Swap node endpoints

### LINKED ISSUE #16608: Swap order of gossip endpoints when creating rosters.
**Background**

As of DAB Phase 2, there are 2 gossip `ServiceEndpoints` in the `Node.proto`.  The "internal ip address" is at index 0 and the `external ip address" is at index 1 in the list.  

Going forward for the near term, Rosters should have the "external ip address" at index 0 and the platform will only only use the "external" ip address when connecting to peers.  

When [constructing Rosters from the node.proto](https://github.com/hashgraph/hedera-services/blob/release/0.56/hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java#L100), the order of endpoints needs to be switched so that the external ip address is at index 0.  

**Acceptance Criteria**

All places where rosters are constructed using node.proto data have the gossip endpoints swapped so that the external ip address is at index 0. 

**Notes**

In the future, the internal IP address will be taken out of the address book and roster to be provided as node configuration.   This has not been designed yet, so the data remains as content here in this list.

petreze:
**Description**:
As per the DAB HIP, there are 2 `ServiceEndpoints` in `node.proto`.
Currently, the "internal ip" is at index 0 and the "external ip" is at index 1 in the list.
However, platform expects the "external ip" to be at index 0. This is why, we want to swap both endpoints when constructing `RosterEntry`

**Related issue(s)**:

Fixes #16608 

**Notes for reviewer**:

**Checklist**


hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
```diff
@@ -92,12 +93,18 @@ private Roster constructFromNodesState(@NonNull final ReadableKVState<EntityNumb
         for (final var it = nodesState.keys(); it.hasNext(); ) {
             final var nodeNumber = it.next();
             final var node = requireNonNull(nodesState.get(nodeNumber));
+            final var nodeEndpoints = node.gossipEndpoint();
+            // we want to swap the internal and external node endpoints
+            // so that the external one is at index 0
+            if (nodeEndpoints.size() == 2) {
```
petreze, line 99: Hey @edward-swirldslabs, here do we assume that the `gossipEndpoint` list will always contain exactly 2 endpoints (internal and external in that order). I saw that by definition in DAB HIP it is specified like that but are there other cases to consider?

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
```diff
@@ -92,12 +93,18 @@ private Roster constructFromNodesState(@NonNull final ReadableKVState<EntityNumb
...
+                Collections.swap(nodeEndpoints, 0, 1);
+            }
             if (!node.deleted()) {
                 final var entry = RosterEntry.newBuilder()
                         .nodeId(node.nodeId())
                         .weight(node.weight())
                         .gossipCaCertificate(node.gossipCaCertificate())
-                        .gossipEndpoint(node.gossipEndpoint())
+                        .gossipEndpoint(nodeEndpoints)
                         // (TSS-FUTURE) Enable node's TSS encryption key
```
edward-swirldslabs, line 108: Can we remove the comments here about TSS encryption key.  That will never be enabled.

petreze: Removed

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/ReadableNodeStoreImpl.java
petreze: Thanks a lot for the clarification, updated!

edward-swirldslabs (approved):
awesome sauce!

Neeharika-Sompalli (approved):
LGTM!

Issue #16552, opened 2024-11-12
Define restart only `Schema`'s for roster and platform services


derektriley:
**Background**
### define `restart()`-only `Schema`s for roster and platform services that `Hedera` computes from the `StartupAssets` and `InitTrigger`

- The `RosterService` schema’s `restart()` method must ensure the candidate, previous, and current rosters are set correctly from the initial conditions and the init trigger.
- The `PlatformService` schema’s `restart()` method must ensure the `PlatformState` has the correct previous and current `AddressBook`s (or `Roster`s).

Note that since candidate roster adoption depends on `TssBaseService` verdicts, these `Schema`s need to be constructed with something like a `Predicate<Bytes> viableRosterHash` whose implementation is in terms of the `TssBaseService` singleton.

See [Roster Lifecycle Implementation](https://www.notion.so/swirldslabs/Roster-lifecycle-implementation-13b7c9ab25918069a0afcc44713015b9) for details listed below.

**Acceptance Criteria**
- RosterService restart-only schema created
- RosterService restart-only schema unit tests
- PlatformStateService restart-only schema created
- PlatformStateService restart-only schema unit tests

**Dependencies**
#16550

PR #16638, opened 2024-11-18
feat: platform and roster states restart-only schemas

MiroslavGatsanoga:
**Description**:

Adds implementation for `V057PlatformStateSchema` and updates the `V057RosterSchema` according to [latest roster proposal](https://github.com/hashgraph/hedera-services/pull/16067).

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/16552

**Checklist**


hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -633,6 +633,7 @@ private void migrateSchemas(
         }
         blockStreamService.resetMigratedLastBlockHash();
         startupNetworks = startupNetworksFactory.apply(selfNodeId.id(), configProvider, tssBaseService);
+        PLATFORM_STATE_SERVICE.setAppVersionFn(() -> version);
```
tinker-michaelj, line 636: This isn't necessary since the version can be [directly computed](https://github.com/hashgraph/hedera-services/blob/16551-startup-assets/hedera-node/hedera-app/src/main/java/com/hedera/node/app/version/ServicesSoftwareVersion.java#L81) from the `Configuration` on the `MigrationContext`,
```
ServicesSoftwareVersion.from(ctx.configuration())
```

hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/schemas/V057RosterSchema.java
```diff
@@ -81,21 +82,55 @@ public void restart(@NonNull final MigrationContext ctx) {
...
-                setActiveRoster(roundNumber + 1, rosterStore, overrideNetwork.get());
+                // currentRound := state round +1
+                final long currentRound = roundNumber + 1;
+                log.info("Found override network for round {}", currentRound);
+
+                // If there is no active roster in the roster state.
+                if (rosterStore.getActiveRoster() == null) {
+                    //   Read the current AddressBooks from the platform state.
+                    //   previousRoster := translateToRoster(currentAddressBook)
+                    final var platformState = platformStateStoreFactory.apply(ctx.newStates());
```
tinker-michaelj, line 93: ⚠️ This won't work for two reasons:
  1. The `{Readable|Writable}States` returned by `MigrationContext` are filtered to just the states owned by the `Schema`'s service; so this schema can't "see" states owned by `PlatformStateService`.
  2. The `RosterService.MIGRATION_ORDER` is [**before**](https://github.com/hashgraph/hedera-services/blob/16551-startup-assets/hedera-node/hedera-app/src/main/java/com/hedera/node/app/roster/RosterService.java#L42) the `PlatformStateService.MIGRATION_ORDER`, so even if the `PlatformStateService` states were available, their metadata would not be registered.

Instead we should probably pass a more source-agnostic `Supplier<Roster>` into this schema and use it.

PR #16456, opened 2024-11-06
chore: removing notnull annotations and replacing libraries

boooby19:
**Description**:
This PR modifies standardization for code which means we are trying to use just one library for annotations @NonNull and @Nullable - edu.umd.cs.findbugs.annotations
For this standardization it was neccessary to remove @NotNull annotations in the code.

**Related issue(s)**: #15375 

Fixes #

**Notes for reviewer**:
is it possible to do something similar like this? Thanks ! :) @netopyr

PR #16641, opened 2024-11-18
removing notnull annotations and replacing libraries #16456

boooby19:
Description:
This PR modifies standardization for code which means we are trying to use just one library for annotations @nonnull and https://github.com/nullable - edu.umd.cs.findbugs.annotations
For this standardization it was neccessary to remove https://github.com/NotNull annotations in the code.

Related issue(s): https://github.com/hashgraph/hedera-services/issues/15375

Fixes #

Notes for reviewer:
@netopyr second chance :)


boooby19:
> If you were also to run `./gradlew spotlessApply` it would remove the now unnecessary `import` statement in that file ...
> 
> (Good practice before uploading any PR or change to PR.)

yes, sorry, I just removed the library

david-bakin-sl (dismissed):
LGTM

david-bakin-sl:
BTW - Don't know if you get a merge button or not - but don't merge w/o 2 <= TWO approvers.  That's the practice here.  TY for the fix!

boooby19:
@netopyr sorry for problems, but I dont know how to turn on that DCO mark, because at first commit that DCO was goot, at second it throws X, but at first when I forked it I already set these lines from maintener guide : 
$ git config --global user.name "Joe Smith"
$ git config --global user.email "joe.smith@example.com"
$ git commit -s -m 'Update README'

boooby19:
sorry guys, I am going to make a new one..

PR #16647, opened 2024-11-18
fix: correct a missing conversion between `AddressBook` cert hash hex-string-as-bytes and actual SHA2-384 hash bytes for `Node` entries.

jsync-swirlds:
* When migrating from an `AddressBook` to the `Node`s, convert from hex string to hash bits
* When generating an `AddressBook` from the `Node`s, convert from hash bits to hex string
* Fix the two unit tests that assumed both were arbitrary bytes.
* Fix the one HAPI test that interacts with the `101`/`102` file generation.

### DIFF
--- a/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip993/SystemFileExportsTest.java
+++ b/hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip993/SystemFileExportsTest.java
@@ -65,6 +65,7 @@ import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.BUSY;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.SUCCESS;
 import static com.hederahashgraph.api.proto.java.ResponseCodeEnum.UNAUTHORIZED;
 import static com.swirlds.common.utility.CommonUtils.unhex;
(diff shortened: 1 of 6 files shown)


hedera-node/test-clients/src/main/java/com/hedera/services/bdd/suites/hip993/SystemFileExportsTest.java
```diff
@@ -463,6 +466,11 @@ private static VisibleItemsValidator nodeDetailsExportValidator(
         };
     }
 
+    private static byte[] getHexStringBytesFromBytes(final byte[] rawBytes) {
+        final String hexString = HexFormat.of().formatHex(rawBytes);
+        return Normalizer.normalize(hexString, Normalizer.Form.NFD).getBytes(UTF_8);
+    }
+
```
mhess-swl, line 473: Would it make sense to move this to a common module, e.g. hapi-utils?

jsync-swirlds: Possibly.  We're fixing an issue in release 0.56 here, but I fully support a follow-up PR to move this to hapi-utils and add even more thorough unit tests.

mhess-swl (approved):

PR #16656, opened 2024-11-18
feat: overhaul SignedState.getAddressBook()

### LINKED ISSUE #16541: Overhaul `SignedState.getAddressBook()`
**Background**
In order to remove AddressBook from PlatformState we need to:
1. Allow SignedState to use RosterService (and fall back to PlatformState) to read AddressBook
2. Switch to using the RosterService instead of PlatformState's addressBook/previousAddressBook. Note that it would be preferable if the App/Services/Execution is the only entity that actually updates the RosterService data. The Platform/Consensus should only ever need to receive this information via the RosterHistory object on startup, or read it via the RosterService/RosterRetriever, but never update it directly.
3. Actually remove the ABs from the PlatformState

**Acceptance criteria**
This task covers the item 1 - we need to enable the SignedState to access RosterService data if available. If unavailable, it should fall back to reading the data from the PlatformState. RosterRetriever already implements this logic, we just need to make use of it.

anthony-swirldslabs:
**Description**:
This is the first step in a series of fixes to eventually remove `AddressBook` objects from the `PlatformState`. In this fix, the `SignedState` class now uses the `RosterRetriever` to fetch the active roster from the state. The latter automatically falls back to reading the AddressBook from the PlatformState if the RosterService data isn't populated.

Notable changes:
1. The crux of the fix is the `SignedState.getAddressBook()` implementation and the supporting `SignedStateFileReader.registerServiceStates()` that allows the former to access states of the RosterService and the PlatformState via the States API.
2. The `SignedStateFileReader` and the `ReconnectLearner` initialize these states to make the AddressBook/Roster accessible from the loaded/received SignedState.
3. The `FakeMerkleStateLifecycles` is similarly extended to enable accessing these states in tests (both unit, and JRS.)
4. The `RandomSignedStateGenerator` also enables the tests to uses these states, and it also registers all the generated states in ThreadLocal lists, allowing the tests to release them once the tests finish. This is now required to avoid producing OOMs due to a large number of VirtualMap instances created with each generated state.
5. `RosterSchema` is moved back to the platform code, and the `RosterStateId` is resurrected, again, to be accessed from the platform code and eliminate circular dependencies between app and platform.
6. Tests using VirtualMaps now have `MerkleDb.resetDefaultInstancePath()` calls to allow the VirtualMap implementation to create multiple instances of the same map on disk. W/o this call, the tests would clash because each of them would want to create the same ROSTERS map when a SignedState object is initialized in them.

**Related issue(s)**:

Fixes #16541

**Notes for reviewer**:
Tests should pass.

**Checklist**


platform-sdk/swirlds-platform-core/src/testFixtures/java/com/swirlds/platform/test/fixtures/state/RandomSignedStateGenerator.java
```diff
@@ -199,6 +206,9 @@ public SignedState build() {
             v.setConsensusTimestamp(consensusTimestampInstance);
         });
 
+        FAKE_MERKLE_STATE_LIFECYCLES.initRosterState((MerkleStateRoot) stateInstance);
+        // TODO: future work: populate roster history and stop setting AddressBook on platformState above.
```
edward-swirldslabs, line 210: The term `TODO` gets caught by some code scanners.  We (the former platform hashgraph team) have  been using just `Future Work` or other variants to indicate work that will be completed in future PRs and leave `TODO` as something that needs to be completed before a PR can be merged.   Is this needing to be completed before this PR can be merged?

edward-swirldslabs (dismissed):

edward-swirldslabs (approved):

edward-swirldslabs (dismissed):

edward-swirldslabs (dismissed):

tinker-michaelj (approved):
Makes sense to me, tyvm @anthony-swirldslabs !

edward-swirldslabs (approved):

PR #16672, opened 2024-11-20
fix: platform tolerates bad gossip certificates

### LINKED ISSUE #16648: Platform tolerates malformed certificates.
**Background**

For the purposes of this set of code changes, malformed X509Certificates for gossip endpoints will be dropped and the platform will tolerate the absence of the gossip certificate for other nodes.  If a node detects that its own gossip certificate is malformed or missing, this is a fatal error.  

1. [ ] The comments on `roster.proto` need to be updated to allow for a null or missing certificates. 
2. [ ] Nodes should not connect to peers or receive incoming connections from peers with missing or invalid gossip certificates. 
3. [ ] Event signature validation should treat the events as invalidly signed if the public certificate is missing at the time of validation. 
4. [ ] State signatures without public keys for validation will not count towards signing weight. 
5. [ ] Loading malformed X509Certificates from disk should produce a log message and result in a null certificate being added. 

**Acceptance Criteria**

* [ ] Unit tests will be developed for each of these use cases to handle the absence of the X509Certificate and fail appropriately. 
* [ ] Manual testing of handling malformed certificates is required.  
* [ ] One or more issues should be created for creating negative integration test scenarios in relation to malformed or missing X509Certificates on gossip endpoints.

edward-swirldslabs:
**Description**:

This PR allows the platform to tolerate missing or malformed gossip certificates in the address book and roster. 
* The platform will throw a fatal exception if the node itself has a malformed or missing gossip certificate. 
* State Signature Validation will ignore signatures from nodes with missing or malformed gossip certificates. 
* Event Signature Validation will not validate the event with missing or malformed gossip certificates. 
* Outgoing and Incoming connections will not be formed to or by nodes with missing or malformed gossip certificates. 

Fixes #16648 

**Notes for reviewer**:

* The startup logic does not have unit tests and it is beyond the scope of this PR to create them.
   * issue created: https://github.com/hashgraph/hedera-services/issues/16733
* I manually ran the platform with invalid gossip certificates on my local machine and verified desired behavior.


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
```diff
@@ -48,13 +48,11 @@ private Utilities() {}
 
     /**
      * Convert a string to a boolean.
+     * <p>
+     * A false is defined to be any string that, after trimming leading/trailing whitespace and conversion to lowercase,
```
anthony-swirldslabs, line 52: Minor: it would be nice to avoid massive code reformatting in PRs, unless it's enforced by spotless or similar mandatory build logic. This is because massive reformatting makes it more difficult to navigate the git history in the future. Also, it makes it difficult to review such code changes because it unnecessarily enlarges them. If the massive, voluntary (not enforced by a tool) reformatting must be released for whatever reason, this can be done in a separate, dedicated PR.

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -288,6 +293,19 @@ public PlatformBuilder withStaleEventCallback(@NonNull final Consumer<PlatformEv
     public PlatformBuilder withKeysAndCerts(@NonNull final KeysAndCerts keysAndCerts) {
         throwIfAlreadyUsed();
         this.keysAndCerts = Objects.requireNonNull(keysAndCerts);
+        // Ensure that the platform has a valid signing cert that matches the signing private key.
```
lpetrovic05, line 296: this does not check if it matches the private key

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
```diff
@@ -174,7 +179,20 @@ public SyncGossip(
         final ThreadConfig threadConfig = platformContext.getConfiguration().getConfigData(ThreadConfig.class);
 
         final BasicConfig basicConfig = platformContext.getConfiguration().getConfigData(BasicConfig.class);
-        final List<PeerInfo> peers = Utilities.createPeerInfoList(roster, selfId);
+        final RosterEntry selfEntry = RosterUtils.getRosterEntry(roster, selfId.id());
+        final List<PeerInfo> peers;
+        if (RosterUtils.fetchGossipCaCertificate(selfEntry) == null) {
```
lpetrovic05, line 184: why do we even make it this far if we dont have a valid cert? why not die immediately

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkPeerIdentifier.java
```diff
@@ -60,6 +61,11 @@ public NetworkPeerIdentifier(@NonNull final PlatformContext platformContext, @No
         noPeerFoundLogger = new RateLimitedLogger(logger, platformContext.getTime(), Duration.ofMinutes(5));
         this.x501PrincipalsAndPeers = HashMap.newHashMap(peers.size());
         for (final PeerInfo peerInfo : peers) {
+            if (!CryptoStatic.checkCertificate(peerInfo.signingCertificate())) {
```
lpetrovic05, line 64: if a peer does not have a cert, why even add him to the list?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/PeerInfo.java
```diff
@@ -19,18 +19,22 @@
...
+ * A record representing a peer's network information.  If the certificate is not null, it must be encodable as a valid
+ * X509Certificate.  It is assumed the calling code has already validated the certificate.
  *
  * @param nodeId             the ID of the peer
  * @param hostname           the hostname (or IP address) of the peer
- * @param signingCertificate the certificate used to validate the peer's TLS certificate
+ * @param signingCertificate the certificate used to validate the peer's TLS certificate, or null.
  */
-public record PeerInfo(@NonNull NodeId nodeId, @NonNull String hostname, @NonNull Certificate signingCertificate) {
+public record PeerInfo(@NonNull NodeId nodeId, @NonNull String hostname, @Nullable Certificate signingCertificate) {
```
lpetrovic05, line 33: why make it nullable?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -143,14 +146,15 @@ public final class PlatformBuilder {
...
-     * @param selfId              the ID of this node
-     * @param softwareVersion     the software version of the application
-     * @param initialState        the initial state supplied by the application
+     * @param appName                  the name of the application, currently used for deciding where to store states on
+     *                                 disk
+     * @param swirldName               the name of the swirld, currently used for deciding where to store states on
+     *                                 disk
+     * @param selfId                   the ID of this node
+     * @param softwareVersion          the software version of the application
+     * @param initialState             the initial state supplied by the application
```
jsync-swirlds, line 155: Could we not do this alignment of param text (with associated frequent reformats and excessive wrapping)?  It's fine to just have a single space after the parameter name in all cases; this all gets rendered (even IDEs render the HTML view), so there's no reason to make the code "pretty".

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
```diff
@@ -288,6 +293,19 @@ public PlatformBuilder withStaleEventCallback(@NonNull final Consumer<PlatformEv
     public PlatformBuilder withKeysAndCerts(@NonNull final KeysAndCerts keysAndCerts) {
         throwIfAlreadyUsed();
         this.keysAndCerts = Objects.requireNonNull(keysAndCerts);
+        // Ensure that the platform has a valid signing cert that matches the signing private key.
+        // https://github.com/hashgraph/hedera-services/issues/16648
+        if (!CryptoStatic.checkCertificate(keysAndCerts.sigCert())) {
+            throw new IllegalStateException("Starting the platform requires a signing cert.");
```
jsync-swirlds, line 299: nit: cert =>> certificate

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
jsync-swirlds: lines 302-307 signs a test value and verifies it.  Wouldn't that ensure the keys match?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
edward-swirldslabs: It is unfortunate that the formatting by IntelliJ is slightly different.  I unconsciously auto format using keyboard shortcuts.   At this point, if I switch it back and extend the commit history, you'll have the same thin in reverse.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gossip/SyncGossip.java
edward-swirldslabs: In this implementation the platform commits sepuku in the platform builder when it discovers the cert in `KeysAndCerts` is bad.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
anthony-swirldslabs: Not really. The `develop` will only receive a squashed commit. And I only ever care about the commit history in the `develop` branch. So undoing the reformatting in this PR while it's still open wouldn't be a problem.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/NetworkPeerIdentifier.java
edward-swirldslabs: Looking closer:  `fetchGossipCaCertificate` does a `CryptoStatic.decodeCertificate` but not a `CryptoStatic.checkCertificate`.    The `List<PeerInfo>` was constructed with non-null certificates that successfully decode from an encoded form.   

If someone modifies the construction of `List<PeerInfo>`, this check would remain.  It's the latest check that enforces the invariants in the issue.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/PeerInfo.java
edward-swirldslabs: You're right, after examining the code, the entries that are null are dropped from the list and the record doesn't need to support null.

edward-swirldslabs: When I make the field `NonNull`, the IDE complains of possible null because it can't reason about the filter applied earlier in the stream.    Making it nullable also forces the code closest to use to deal with the potential of no certificate or bad certificate.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
jsync-swirlds: Also, IntelliJ formatting _should not_ rewrite the javadoc this way if you have the "official" style settings installed.
Note, github won't allow .xml uploads (even if the content is xml; it requires a `.txt` extension).  This file must be saved with `.xml` extension to import into IntelliJ.
[Hashgraph_Style.xml](https://github.com/user-attachments/files/17835249/Hashgraph_Style.xml.txt)

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
edward-swirldslabs: updated text

anthony-swirldslabs (dismissed):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/Utilities.java
edward-swirldslabs: I have imported the new style, thank you.

edward-swirldslabs: Using the new style, the formatter for intellij still puts the javadoc params on the same line.   I'll undo the style changes to the older javadoc

anthony-swirldslabs (approved):
LGTM

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/builder/PlatformBuilder.java
lpetrovic05: ah, yes, I missed that

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/network/PeerInfo.java
lpetrovic05: I disagree with this decision. Checks should be done once, so that we don't have to re-check lots of times downstream. A list of peers should be ones with valid certificates, so we should ensure that when constructing the list and then trust if from then on. With this approach, we just add a bunch of boilerplate code with no real purpose.

edward-swirldslabs: I have restored it to being `@NonNull`

anthony-swirldslabs (approved):

anthony-swirldslabs (approved):

PR #16679, opened 2024-11-20
build: run validatePomFiles as part of qualityGate and PR checks

### LINKED ISSUE #16649: Add validatePomForMavenPublication task to gradle maven-publish tasklist
Bring in the `io.freefair.maven-central.validate-poms` plugin and use it to validate pom.xml files that are generated.

jjohannes:
**Description**:

Run `validatePomFiles` as part of qualityGate and PR checks. It is not expensive to run this check and this will hopefully reduce the amount of issues we only find very late in the publishing process right now.

**Related issue(s)**:

Resolves #16649
Builds on #16651

**Checklist**

### DIFF
--- a/gradle/plugins/build.gradle.kts
+++ b/gradle/plugins/build.gradle.kts
@@ -32,13 +32,11 @@ dependencies {
     implementation(
         "gradle.plugin.com.google.cloud.artifactregistry:artifactregistry-gradle-plugin:2.2.2"
     )
+    implementation("io.freefair.gradle:maven-plugin:8.11") // for POM validation
     implementation("io.github.gradle-nexus:publish-plugin:1.3.0")
     implementation("me.champeau.jmh:jmh-gradle-plugin:0.7.2")
     implementation("net.swiftzer.semver:semver:1.3.0")
     implementation("org.gradlex:extra-java-module-info:1.9")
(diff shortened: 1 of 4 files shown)


gradle/plugins/build.gradle.kts
```diff
@@ -32,13 +32,11 @@ dependencies {
     implementation(
         "gradle.plugin.com.google.cloud.artifactregistry:artifactregistry-gradle-plugin:2.2.2"
     )
+    implementation("io.freefair.gradle:maven-plugin:8.11") // for POM validation
```
jjohannes, line 35: Use short dependency notation and latest version of plugin

andrewb1269hg (dismissed):

hedera-node/hedera-app-spi/build.gradle.kts
```diff
@@ -29,7 +29,5 @@ testModuleInfo {
     requires("org.assertj.core")
     requires("org.junit.jupiter.api")
     requires("org.junit.jupiter.params")
-    requires("org.mockito")
-    requires("org.mockito.junit.jupiter")
```
jjohannes, line 33: These are not needed. Looks like this is wrong on `develop`. When leaving these in, check is failing on this PR.

andrewb1269hg (approved):

andrewb1269hg (approved):

PR #16708, opened 2024-11-21
fix: Set the designated payer when building an EVM transactions from scheduled transactions

JivkoKelchev:
**Description**:
Updates the `hevmTransactionFactory` to accept the payer instead of extracting it from the transaction ID, while building from hapi transaction. This is needed because scheduled transactions inherited their ID from `scheduleCreate` transactions, and may have a different payer. 

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Nana-EC:
@JivkoKelchev can you add more details to this PR.
Also let's discuss this a bit more, my understanding is this is enabling the scheduling of a ContractCall transaction.
Can we enable this only in dev & previewnet at this juncture while this is explored more.
Same for all Smart contract transactions
FYI @netopyr and @lukelee-sl

JivkoKelchev:
Hi @Nana-EC ,

This change was needed because scheduled transactions inherit their transaction ID from their initial `ScheduleCreate` transaction.

In this [case](https://github.com/hashgraph/hedera-services/pull/16708/commits/0ce191430d9c69f9c57776bfccdbbd813033ceca#diff-969854443032bba31cf9463b5991002f4f342d9c92075888a6395e65964e4987R568-R574), we want a custom payer that will pay the gas. So we need a way to provide the payer (contract call sender) instead of extracting the one from the transaction ID, while creating the `HederaEvmTransaction` [here](https://github.com/hashgraph/hedera-services/pull/16708/commits/0ce191430d9c69f9c57776bfccdbbd813033ceca#diff-48236ed231483234387dba1eacf83a148dd8cc1fabdb1d45a21a8b7519b7f787L150-L151).

However, I’m not really sure if we still need these changes, since @tinker-michaelj did significant refactoring of the long-term schedule implementation. What do you think Michael?

tinker-michaelj (approved):
LGTM, tyvm @JivkoKelchev !

PR #16722, opened 2024-11-21
chore: replace `Platform.getAddressBook()` with `Platform.getRoster()`

### LINKED ISSUE #16717: Replace `getAddressBook()` with `getRoster()` in `SwirldsPlatform.java`
**Background**

In our platform refactor we are working to eliminate usages of `AddressBook` with `Roster`.   #16459 adds `Platform.getRoster()` to the platform interface and replaces usages of `Platform.getAddressBook()` in the platform demo applications.   

This PR is to replace the remaining usages of this method with `getRoster()` and use the `Roster` in the calling code paths.  This PR should be able to replace the remaining usages and remove the `getAddresBook()` method completely from the interface. 

**Acceptance Criteria**

* System works with the `getAddressBook()` method removed. 
* Create new issues whenever we discover that the App/Services is calling `getRoster()`.   They create the roster and give it to the platform.  That is where they need to get their data from. 

**Dependencies**
https://github.com/hashgraph/hedera-services/pull/16459 must merge before this issue can be started.

edward-swirldslabs:
**Description**:
As part of the platform refactor to replace usage of `AddressBook` with `Roster`, this PR replaces the use of `Platform.getAddressBook()` with `Platform.getRoster()`.   The impacted code is updated with equivalent usage for rosters.

**Related issue(s)**:

Fixes #16717

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/internal/WinTabAddresses.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/internal/WinTabAddresses.java
@@ -19,10 +19,11 @@ package com.swirlds.platform.gui.internal;
 import static com.swirlds.platform.gui.GuiUtils.wrap;
 import static com.swirlds.platform.gui.internal.BrowserWindowManager.getPlatforms;
 
+import com.hedera.hapi.node.state.roster.RosterEntry;
 import com.swirlds.platform.gui.GuiUtils;
 import com.swirlds.platform.gui.components.PrePaintableJPanel;
+import com.swirlds.platform.roster.RosterUtils;
 import com.swirlds.platform.system.Platform;
-import com.swirlds.platform.system.address.Address;
 import javax.swing.JTextArea;
 
 /**
@@ -52,13 +53,15 @@ class WinTabAddresses extends PrePaintableJPanel {
         String s = "";
         synchronized (getPlatforms()) {
             for (final Platform p : getPlatforms()) {
-                final Address address = p.getAddressBook().getAddress(p.getSelfId());
-                s += "\n" + address.getNodeId().id() + "   " + address.getNickname()
-                        + "   " + address.getSelfName()
-                        + "   " + address.getHostnameInternal()
-                        + "   " + address.getPortInternal()
-                        + "   " + address.getHostnameExternal()
-                        + "   " + address.getPortExternal();
+                final RosterEntry entry =
+                        RosterUtils.getRosterEntry(p.getRoster(), p.getSelfId().id());
+                final String name = RosterUtils.formatNodeName(entry.nodeId());
+                s += "\n" + entry.nodeId() + "   " + name
+                        + "   " + name
+                        + "   " + RosterUtils.fetchHostname(entry, 1) // internal hostname
+                        + "   " + RosterUtils.fetchPort(entry, 1) // internal port
+                        + "   " + RosterUtils.fetchHostname(entry, 0) // external hostname
+                        + "   " + RosterUtils.fetchPort(entry, 0); // external port
             }
         }
         s += wrap(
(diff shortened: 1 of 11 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/internal/WinTabAddresses.java
```diff
@@ -52,13 +53,15 @@ public void prePaint() {
         String s = "";
         synchronized (getPlatforms()) {
             for (final Platform p : getPlatforms()) {
```
anthony-swirldslabs, line 55: Minor: this is a very poor anti-pattern. If the getter ever becomes dynamic and could return different objects, this synchronization scheme will break apart. A local variable is needed here. This may never become a problem in this particular code, but an engineer seeing this code and then replicating this pattern elsewhere could run into a big trouble...

Neeharika-Sompalli (dismissed):
LGTM! Thanks @edward-swirldslabs

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/address/AddressBookNetworkUtils.java
```diff
@@ -53,21 +56,38 @@ public static boolean isLocal(@NonNull final Address address) {
...
+     * Check if the rosterEntry is local to the machine.
+     *
+     * @param rosterEntry the rosterEntry to check
+     * @return true if the rosterEntry is local to the machine, false otherwise
+     * @throws IllegalStateException if the locality of the rosterEntry cannot be determined.
+     */
+    public static boolean isLocal(@NonNull final RosterEntry rosterEntry) {
+        Objects.requireNonNull(rosterEntry, "The rosterEntry must not be null.");
+        try {
+            final String internalHostname = RosterUtils.fetchHostname(rosterEntry, 1);
```
anthony-swirldslabs, line 83: The current logic in the `buildRoster()` would skip a duplicate service endpoint if both internal and external endpoints in the AddressBook are the same, which is true for the majority of the nodes on the network in mainnet.

Further, the Roster specification doesn't mandate having more than a single endpoint in the list.

So fetching an entry at index 1 unconditionally may and likely will fail.

Additionally, the current logic in the Consensus module would only ever use the entry at index 0 to establish network connections.

With all this in mind, you probably want to run this check against the entry at index 0. Or perhaps remove it altogether, with a note in the "node configuration/endpoint override" story that it needs to be restored when those overrides are implemented.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/address/AddressBookNetworkUtils.java
edward-swirldslabs: Nope.   It has to use the internal ip address until we have an alternate means of determining which node is starting. 

Issue: https://github.com/hashgraph/hedera-services/issues/11751

My PR for addressing this has been blocked for higher priority work and requires coordination across teams. 

PR: https://github.com/hashgraph/hedera-services/pull/12112

edward-swirldslabs: I can modify the implementation to check for size of the list and query appropriately.

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/gui/internal/WinTabAddresses.java
anthony-swirldslabs: Filed https://github.com/hashgraph/hedera-services/issues/16726 .

Neeharika-Sompalli (approved):
LGTM! Thanks @edward-swirldslabs

PR #16732, opened 2024-11-22
chore: replace `AddressBook` with `Roster` in `PlatformBuilder`

### LINKED ISSUE #16728: Replace `AddressBook` with `Roster` in `PlatformBuilder` and `SwirldStateManager`
**Background**

Continuing the the platform refactor to replace `AddressBook` usage with `Roster`,  use `Roster` in the `SwirldStateManager` constructor and remove the creation of address books from rosters in the `PlatformBuilder`. 

**Acceptance Criteria**

Code works and passes all tests.

edward-swirldslabs:
**Description**:

This PR replaces the `AddressBook` with `Roster` in the `PlatformBuilder`

**Related issue(s)**:

Fixes #16728

### DIFF
--- a/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
+++ b/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
@@ -18,16 +18,17 @@ package com.swirlds.platform.state;
 
 import static com.swirlds.platform.state.SwirldStateManagerUtils.fastCopy;
 
+import com.hedera.hapi.node.state.roster.Roster;
 import com.swirlds.common.context.PlatformContext;
 import com.swirlds.common.platform.NodeId;
 import com.swirlds.platform.FreezePeriodChecker;
 import com.swirlds.platform.internal.ConsensusRound;
 import com.swirlds.platform.metrics.SwirldStateMetrics;
+import com.swirlds.platform.roster.RosterRetriever;
 import com.swirlds.platform.state.signed.SignedState;
 import com.swirlds.platform.system.Round;
 import com.swirlds.platform.system.SoftwareVersion;
 import com.swirlds.platform.system.SwirldState;
-import com.swirlds.platform.system.address.AddressBook;
 import com.swirlds.platform.system.status.StatusActionSubmitter;
 import com.swirlds.platform.uptime.UptimeTracker;
 import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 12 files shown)


platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/state/SwirldStateManager.java
```diff
@@ -125,7 +126,9 @@ public void setInitialState(@NonNull final MerkleRoot state) {
     public void handleConsensusRound(final ConsensusRound round) {
         final MerkleRoot state = stateRef.get();
 
-        uptimeTracker.handleRound(round, state.getReadablePlatformState().getAddressBook());
+        uptimeTracker.handleRound(
+                round,
+                RosterRetriever.buildRoster(state.getReadablePlatformState().getAddressBook()));
```
anthony-swirldslabs, line 131: The `handleConsensusRound` is something that executes rather often, while the roster (currently) doesn't change that often. Building a new instance for a (currently) immutable object every time a round is processed seems expensive from the performance perspective.

Let's find a way to avoid having this translation here.

edward-swirldslabs: fair point.  let me look at the best way to avoid this.

lpetrovic05 (dismissed):

anthony-swirldslabs (approved):

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/eventhandling/DefaultTransactionHandlerTests.java
```diff
@@ -119,7 +120,12 @@ void normalOperation(final boolean pcesRound) throws InterruptedException {
                 handlerOutput.reservedSignedState().get().getReservationCount(),
                 "state should be returned with a reservation");
 
-        assertTrue(tester.getSubmittedActions().isEmpty(), "no status should have been submitted");
+        // only the self event reaching consensus should be reported, no freeze action.
+        assertEquals(1, tester.getSubmittedActions().size(), "the freeze status should have been submitted");
```
anthony-swirldslabs, line 124: Minor: the assert error message seems inconsistent with the comment at line 123.

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/eventhandling/DefaultTransactionHandlerTests.java
```diff
@@ -167,8 +173,9 @@ void freezeHandling() throws InterruptedException {
                 1,
                 handlerOutput.reservedSignedState().get().getReservationCount(),
                 "state should be returned with a reservation");
-
-        assertEquals(1, tester.getSubmittedActions().size(), "the freeze status should have been submitted");
+        // In addition to the freeze action, the uptime tracker reports a self event coming to consensus in the round.
+        assertEquals(2, tester.getSubmittedActions().size(), "the freeze status should have been submitted");
```
anthony-swirldslabs, line 177: Minor: I may be unaware of the internal details, but the generation of an extra event seems like a significant change in the behavior.

Why does this extra event get generated with this fix?

edward-swirldslabs: In the code change, the `UptimeTracker.handleRound()` is no longer taking `AddressBook` as a second argument.  This unit testing was indirectly passing a null `AddressBook` and `UptimeTracker.handleRound()` was exiting due to the AB being null.   I removed the AB from the method because there is a roster in the round and the additional `AB` seems superfluous.    In the previous implementation, the `UptimeTracker` was exiting because the unit test is not standing up a complete platform state with a non-null address book.

anthony-swirldslabs: Got it. Thanks for the clarification!

platform-sdk/swirlds-platform-core/src/test/java/com/swirlds/platform/eventhandling/DefaultTransactionHandlerTests.java
edward-swirldslabs: fixed message.   thanks for the catch.

anthony-swirldslabs (approved):

PR #16719, opened 2024-11-21
chore: Modify and mount default block streams output directory

mhess-swl:
This PR modifies the default block streams output directory to `/opt/hgcapp/blockStreams` in order to match the corresponding record streams directory. This directory is also added as a mount in the relevant docker compose files.

Closes #16713

PR #16736, opened 2024-11-22
feat: update production container image to support block streams

### LINKED ISSUE #16735: Production container updated for block stream support
## Description

We need to update the production docker container image to pre-create the `/opt/hgcapp/blockStreams` path. Additionally, we need to ensure the consensus software block stream location is also defaulted to `/opt/hgcapp/blockStreams`. 

### Related Issues 

- swirlds/swirlds-docker#1000

nathanklick:
## Description

* Updates the production container image to support block streams.
* Updates the default block stream storage location in the consensus node software.

### Related Issues

- Closes #16735


netopyr (changes requested):
This seems to be an incomplete duplicate of https://github.com/hashgraph/hedera-services/pull/16719

nathanklick:
> This seems to be an incomplete duplicate of #16719

@netopyr This is the bare minimum required for `0.56` to support block streams on testnet/mainnet. I thought the goal was to minimize changes in `0.56`.

PR #16759, opened 2024-11-22
fix: Rework `GenesisNetworkInfo` to use the genesis address book

### LINKED ISSUE #16643: GenesisNetworkInfo should use AddressBook, not Roster.
**Background**

The current code reads config.txt on genesis, constructs a genesis address book, then constructs a genesis roster from the AB, and finally constructs the GenesisNetworkInfo from the genesis roster. 

https://github.com/hashgraph/hedera-services/blob/acbc5b6db856ed005a0406f1476420e49bba423e/hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java#L607-L613

What should happen is that the roster is not created until just before handing the roster over to the platform.   The `GenesisNetworkInfo` should be constructed using the genesis AddressBook.  

In the `node.proto` in translating from `AddressBook`, the order of gossip endpoints should be internal ip address at index 0 and external ip address at index 1.   

**Acceptance Criteria**

Rosters are not created until we're ready to give it to the platform. 

When constructing a roster, the order of gossip endpoints is reversed.  external at index 0 and internal at index 1.  

**Notes**

mhess-swl:
In addition to replacing the roster with the genesis address book in `GenesisNetworkInfo`, this PR also makes a further attempt to standardize how the code creates/retrieves rosters. 

Closes #16643


mhess-swl:
We're determining if this work is still relevant

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
```diff
@@ -180,18 +181,6 @@ private Map<Long, NodeAddress> getNodeAddressMap(@NonNull final MigrationContext
...
-                (byte) Integer.parseInt(octets[0]),
-                (byte) Integer.parseInt(octets[1]),
-                (byte) Integer.parseInt(octets[2]),
-                (byte) Integer.parseInt(octets[3])
-            }));
-        } else {
-            builder.domainName(host);
-        }
-        return builder.build();
+        return AddressBookUtils.endpointFor(host, port);
```
edward-swirldslabs, line 184: Is a `V053` schema still getting executed?  When would they no longer be executed?

edward-swirldslabs (approved):

hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/schemas/V053AddressBookSchema.java
mhess-swl: This change merely calls the same logic, now contained in `AddressBookUtils`, to construct the service endpoint. 

To answer your question, yes, this schema is still executed for all nodes starting from genesis or < v53 cases.

edward-swirldslabs (dismissed):

edward-swirldslabs (dismissed):

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
```diff
@@ -630,8 +629,7 @@ private void migrateSchemas(
         if (trigger == GENESIS) {
             final var config = configProvider.getConfiguration();
             final var ledgerConfig = config.getConfigData(LedgerConfig.class);
-            final var genesisRoster = buildRoster(requireNonNull(genesisAddressBook));
-            genesisNetworkInfo = new GenesisNetworkInfo(genesisRoster, ledgerConfig.id());
+            genesisNetworkInfo = new GenesisNetworkInfo(requireNonNull(genesisAddressBook), ledgerConfig.id());
```
Neeharika-Sompalli, line 632: It was intentional to use Roster inside `GenesisNetworkInfo`. How does it matter if we are using addressBook or Roster here ?

hedera-node/hedera-app/src/main/java/com/hedera/node/app/info/GenesisNetworkInfo.java
```diff
@@ -38,19 +40,17 @@
...
 
     /**
      * Constructs a new {@link GenesisNetworkInfo} instance.
      *
-     * @param genesisRoster The genesis roster
+     * @param genesisAddressBook The genesis address book
      * @param ledgerId      The ledger ID
      */
-    public GenesisNetworkInfo(@NonNull final Roster genesisRoster, @NonNull final Bytes ledgerId) {
+    public GenesisNetworkInfo(@NonNull final AddressBook genesisAddressBook, @NonNull final Bytes ledgerId) {
```
Neeharika-Sompalli, line 51: I don't understand why it shouldn't be Roster here

hedera-node/hedera-app/src/main/java/com/hedera/node/app/Hedera.java
edward-swirldslabs: The swapping of the order of the ip addresses in the roster vs the hedera address book was how I broke integration two weeks ago.  They were not being swapped when going from roster -> genesis network info.  

The roster was always intended to be derived from Hedera Address Book and not used to define the Hedera Address Book.   The situation will be moot once we adopt the new format for genesis network description `genesis_network.json`.   The roster just doesn't have enough information to properly define the Hedera AB.

hedera-node/hedera-app/src/main/java/com/hedera/node/app/info/GenesisNetworkInfo.java
edward-swirldslabs: Right now there might be so little extra information in the Hedera Address Book that we can default value the content that is missing and extend a roster to become a full address book.  But if the definition of the Hedera Address Book grows, that is more and more content that has to be defaulted in.   Right now our genesis process is only used in test cases and fill-in-the-gaps fill in the gaps with defaults might work.   But very soon with SPNs we need a real genesis behavior and we can't assume a hard-coded default value for things like the account bound to a node.  The future `genesis_network.json` format will have the ability to set a dynamic account value.  The Roster doesn't have that value and can't be an accurate in-memory representation of the genesis network information. 

Roster --> `AddressBook` is only useful to the platform while we are refactoring and incrementally replacing the usage of the AddressBook in the platform.  After the refactor is complete, the only valid translation is to go from `AddressBook`  -->  Roster.    `AddressBook` here in the code should be replaced with whatever format you will use to translate `genesis_network.json` into an in-memory data structure and from there rosters can be created to be given to the platform.

Neeharika-Sompalli (approved):
LGTM! Thanks @mhess-swl

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
```diff
@@ -356,4 +389,60 @@ public static String formatConsensusEventStreamName(
...
+        return signingCertificateBytes;
+    }
+
+    /**
+     * Maps a {@link Network} object to its equivalent {@link Roster}.
+     *
+     * @param network the network to represent as a roster
+     * @return the converted roster
+     */
+    public static Roster fromNetwork(@NonNull final Network network) {
```
MiroslavGatsanoga, line 417: nitpick: I would've looked first for such method (and the other two below) in `RosterUtils` instead of here, we can perhaps move their definition there?

MiroslavGatsanoga (approved):
LGTM!

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/system/address/AddressBookUtils.java
mhess-swl: Good idea, done. Thanks!

Neeharika-Sompalli (approved):
LGTM! Thanks @mhess-swl

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -219,4 +237,9 @@ private static Roster retrieveInternal(@NonNull final State state, @Nullable fin
         }
         return null;
     }
+
+    private static boolean isValidServiceEndpoint(@NonNull final ServiceEndpoint endpoint) {
+        return endpoint.ipAddressV4().length() > 0
```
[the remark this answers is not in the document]
mhess-swl: This was my attempt to remember the following (current) code:
```
                    .gossipEndpoint(List.of(
                                    Pair.of(address.getHostnameExternal(), address.getPortExternal()),
                                    Pair.of(address.getHostnameInternal(), address.getPortInternal()))
                            .stream()
                            .filter(pair -> pair.left() != null && !pair.left().isBlank() && pair.right() != 0) // <----- this
                            .distinct()
```

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -158,7 +162,21 @@ public static RosterEntry buildRosterEntry(
             final long weight,
             @NonNull final Bytes cert,
             @NonNull final List<ServiceEndpoint> endpoints) {
-        final List<ServiceEndpoint> rosterOrdered = endpoints.size() > 1 ? endpoints.reversed() : endpoints;
+        // Filter out any invalid endpoints
+        final List<ServiceEndpoint> filtered = endpoints.stream()
+                .filter(e -> {
```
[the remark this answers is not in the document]
mhess-swl: Probably. What is the expected behavior in that case?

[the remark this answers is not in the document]
edward-swirldslabs: Nope, there is no code to handle that.    Can't connect to anyone that has 0 gossip endpoints.  Its fundamentally bad input to be able to reach this state.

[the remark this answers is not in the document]
edward-swirldslabs: @anthony-swirldslabs  wrote the original.     If we have proper input validation, this filter should never execute.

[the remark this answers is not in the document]
mhess-swl: Both `NodeCreateHandler` and `NodeUpdateHandler` call `AddressBookValidator#validateServiceEndpoint` to verify a valid endpoint ([here](https://github.com/hashgraph/hedera-services/blob/778d2794ca3f60be28a864538484659d348d3afd/hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/handlers/NodeCreateHandler.java#L109) and [here](https://github.com/hashgraph/hedera-services/blob/778d2794ca3f60be28a864538484659d348d3afd/hedera-node/hedera-addressbook-service-impl/src/main/java/com/hedera/node/app/service/addressbook/impl/handlers/NodeUpdateHandler.java#L130)), so we should be good 👍

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -145,59 +145,67 @@ public static long getRound(@NonNull final State state) {
...
-                                    throw new InvalidAddressBookException(e);
-                                }
-                            })
-                            .toList())
-                    .build();
         } catch (CertificateEncodingException e) {
             throw new InvalidAddressBookException(e);
         }
+
+        final List<ServiceEndpoint> endpoints = AddressBookUtils.endpointsFor(address);
```
anthony-swirldslabs, line 207: This `endpointsFor` method returns endpoints in an order that is opposite to what the Roster expects. In the Roster, the external endpoint goes at index 0, followed by an optional internal endpoint. The internal endpoint may be skipped if it's identical to the external one.

The entire platform code relies on this particular order of endpoints in the Roster and will be broken if it's reversed. This is because we're going to deprecate internal endpoints from the Roster eventually and replace them with the node configuration. So we choose to use the entry at index 0 from the start to refer to an external endpoint that the platform normally uses.

It would be best to retain the existing conversion logic in the RosterRetriever class as is.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterUtils.java
```diff
@@ -338,4 +341,42 @@ public static AddressBook buildAddressBook(@NonNull final Roster roster) {
...
+        }
+
+        final var node = metadata.nodeOrThrow();
+        return RosterRetriever.buildRosterEntry(
+                NodeId.of(node.nodeId()),
+                node.weight(),
+                node.gossipCaCertificate(),
+                // If we're reading metadata, the endpoint order SHOULD match the address book, which always has the
+                // internal endpoint at index 0
+                node.serviceEndpoint());
```
anthony-swirldslabs, line 380: Then you cannot put it into the roster w/o flipping. A RosterEntry should always have the external endpoint at index 0.

edward-swirldslabs: Agreed, this needs to be flipped.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
```diff
@@ -145,59 +145,67 @@ public static long getRound(@NonNull final State state) {
...
+                    if (!isValidServiceEndpoint(e)) {
+                        log.warn("Filtering invalid service endpoint {}", e);
+                        return false;
+                    }
+                    return true;
+                })
+                .toList();
+
+        // Order the endpoints in the expected roster order
+        final List<ServiceEndpoint> rosterOrdered = filtered.size() > 1 ? filtered.reversed() : filtered;
```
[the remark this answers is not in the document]
mhess-swl: > I just saw this after posting my original comment. I understand how you work around the issue. However, I don't find this particularly elegant or easy to maintain. I'd still prefer to retain the original code because the logic there is quite straightforward.

I have to disagree. As a newcomer to this code, reading the current implementation–or changing it–were not straightforward. Converting into `Pair` objects and doing inline validation–essentially duplicating existing validation code elsewhere, I might add–is not only clunky, but error-prone. Furthermore, the roster was designed inside the boundaries of the platform code, and yet the roster itself isn't even consistent with platform's own address book ordering. How is unwieldy, error-prone, and inconsistent code straightforward (or elegant)?

> Also, this workaround can in fact still end up filtering out the external endpoint which the platform really needs, and only keeping the internal endpoint. This is an error condition from the platform perspective.

If this is true, then why doesn't the current conversion code account for these potential errors either?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterUtils.java
mhess-swl: This only proves my earlier point: passing around two service endpoints in a list, with no designation of internal vs. external, is _at best_ difficult to account for. I recommend we change these APIs from an arbitrary list of `ServiceEndpoint` to something more explicit.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
anthony-swirldslabs: I fully support the intent of this change per your description. And if this conversion logic must be re-written, that's okay, too.

Speaking of this particular line of code though, since this call occurs in a Roster-related class and we know that the `AddressBookUtils` would return the endpoints in an order that is opposite to the roster specification, it would be a lot more appropriate to flip the entries right here, instead of burying the flipping logic down the call stack in a completely different method. This would help ensure that the rest of the code in this class doesn't need to deal with flipping at all.

However, per a comment below in this PR, it does seem that there's an extra place which requires the flipping logic - in the RosterUtils. It would be ideal to find a way to have this flipping implemented literally once in a single place somehow, to follow the spirit of the intent of this fix.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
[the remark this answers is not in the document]
anthony-swirldslabs: As suggested above, it may make sense to move the flipping logic to where we obtain the list of endpoints in the first place, so that they're always in the correct order everywhere else in this roster-specific class.

As for the exception, that's a good question to which there's not a good answer other than "we missed it when we wrote it originally." Would adding the exception break any unit tests? If not, is there a good reason not to include it with this conversion logic overhaul while we're at it?

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterUtils.java
anthony-swirldslabs: I don't disagree with the sentiment fully. However, as of today, the plan is to use the endpoints in the RosterEntry to represent various external endpoints, such as IP-address based, FQDN-based, and in the future, perhaps IPv6-based. While currently it's unclear how this array of endpoints is going to be actually used, it is certain that we don't have plans to include any internal endpoints in this list because they're generally not useful in the roster as a public data structure.

But as long as we do use this list, we need to have the external endpoint at index zero, so the quoted code needs to flip the endpoints here.

platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/roster/RosterRetriever.java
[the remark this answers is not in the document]
anthony-swirldslabs: Likely, "we just missed it back then". Less likely, but not impossible, some unit tests might use bogus AddressBooks w/o any endpoints at all, in which case this code would fail to support them. If the latter is true, well, then we're good. If not though, I think it wouldn't hurt adding the exception while we're refactoring this code anyway.

mhess-swl:
I found out that services will need to get `genesis-network.json` loaded and working in order for the TSS startup scenarios to be successful. In other words, we can't load from config.txt anyway given the rest of the TSS implementation, so this work is no longer necessary. Instead of using `config.txt` in the relevant spot called out in the ticket, we'll use the genesis network instead. 

I'm going to close this PR unless there are any objections? @anthony-swirldslabs @edward-swirldslabs

edward-swirldslabs:
> I found out that services will need to get `genesis-network.json` loaded and working in order for the TSS startup scenarios to be successful. In other words, we can't load from config.txt anyway given the rest of the TSS implementation, so this work is no longer necessary. Instead of using `config.txt` in the relevant spot called out in the ticket, we'll use the genesis network instead.
> 
> I'm going to close this PR unless there are any objections. @anthony-swirldslabs @edward-swirldslabs

No objections.   So long as the Roster is not being used as the medium of setting up the app Network state.

mhess-swl:
I pushed a small change to the `16839-restart-hapi-tests` branch that eliminates the need to use the genesis address book (as explained above) in the `Hedera` class: 

```
        if (trigger == GENESIS) {
            final var config = configProvider.getConfiguration();
            final var ledgerConfig = config.getConfigData(LedgerConfig.class);
            genesisNetworkInfo = new GenesisNetworkInfo(genesisNetwork.nodeMetadata(), ledgerConfig.id());
        }
```

Turns out the network genesis json info is already passed into this method, but wasn't being used here.

anthony-swirldslabs:
> I'm going to close this PR unless there are any objections

Sounds good to me, too.

mhess-swl:
Closing in favor of using the new network json properties instead

PR #16762, opened 2024-11-24
chore: removing notnull annotations and replacing libraries

boooby19:
Description: This PR modifies standardization for code which means we are trying to use just one library for annotations [@nonnull](https://github.com/nonnull) and https://github.com/nullable - edu.umd.cs.findbugs.annotations For this standardization it was neccessary to remove https://github.com/NotNull annotations in the code.

Related issue(s): #15375 

Fixes #
Notes for reviewer:
@netopyr hopes that this would be good


netopyr:
Hi @boooby19 ! Something is still wrong. It says the commit is unverified. 

image

